PR #523, opened 2026-05-29
fix(reader): UDT field decoding for generic frozen<udt>

### LINKED ISSUE #502: Reader: UDT field decoding for generic frozen<udt> type strings
## Context

Surfaced during Phase 2 type coverage work (#470, PR #489).

The reader cannot decode UDT field values when the column's type string is the generic `frozen<udt>` (no concrete UDT type name). The inner UDT bytes are written correctly but read back as `Value::Frozen(Box::new(Value::Null))` because the reader has no way to map the generic string to field names and types.

## Affected test

`cqlite-core/tests/write_read_roundtrip/type_coverage.rs::test_type_frozen_udt` — currently accepts `Frozen(Null)` alongside fully-decoded values. Should assert exact equality once decoding lands.

## Done when

- Reader resolves UDT type names through a registry (or schema attachment) and decodes fields by name+type.
- `test_type_frozen_udt` tightens to `assert!(col_value == original || col_value == inner_udt)` (no `Frozen(Null)` arm).

pmcfadin:
## Summary

- Closes #502
- Extends the `frozen<>` branch in `parse_cell_value_schema_order` to resolve the inner type name via the UDT registry before falling through to the generic blob/tuple path.
- Adds `read_back_column_with_udt_registry` test helper that injects a pre-populated `UdtRegistry` into the `SSTableManager` via `SchemaRegistry`.
- Tightens `test_type_frozen_udt`: the `Frozen(Null)` acceptance arm is removed and the test asserts exact equality with the original decoded value.

## Root Cause

`parse_cell_value_schema_order` handled `frozen<X>` columns via three branches:
1. `list<>`, `set<>`, `map<>` — frozen collection parsers
2. Full marshal format `org.apache.cassandra.db.marshal.UserType(...)` — `parse_udt_type_definition`
3. Everything else — delegate to `parse_cell_value_schema_order` with the inner type string

When the column type is `frozen<person>`, `inner_type` is `"person"` — not a collection and not a marshal-format UserType — so it fell into branch 3. That delegated back to the same function with `data_type = "person"`, which matched no known type and fell to the blob/`Value::Null` handler, producing `Frozen(Null)`.

## How Type Resolution Works

A new `else if` arm is inserted between the marshal-format UDT check and the catch-all else:

```rust
} else if let Some(udt_def) = self
    .udt_registry
    .as_ref()
    .and_then(|reg| reg.get_udt(&self.keyspace, &inner_type).cloned())
{
    // Read VUInt-prefixed blob length (same framing as tuple and marshal-format UDT cells)
    // then delegate to parse_udt_value with the registry-supplied UdtTypeDef.
```

The `UdtRegistry` on the parser is populated before the scan via `SchemaRegistry`. The test creates a `UdtRegistry` with the "person" type definition, registers it into a `SchemaRegistry`, and passes that to `SSTableManager::new`. The existing `load_existing_sstables` path then calls `reader.set_udt_registry(...)` before wrapping the reader in `Arc`.

The column type string in the schema was changed from the unresolvable `frozen<udt>` to the concrete `frozen<person>` — a name that can be looked up in the registry. The bare `frozen<udt>` string was inherently undecodable: the UDT bytes contain only field values (not names or type tags), so without a registry entry there is no authoritative source of field metadata.

## Test Plan

- [x] `cargo fmt --all -- --check` passes
- [x] `RUSTFLAGS="-D warnings" cargo clippy --workspace --all-targets --all-features` passes (0 warnings)
- [x] `cargo test --package cqlite-core --features cli-helpers` — 1399 passed, 34 ignored
- [x] `cargo test --package cqlite-core --features write-support,state_machine` — 1535 passed (pre-existing `test_clustering_key_ord_type_mismatch_panics` failure unrelated to this change)
- [x] `bash test-data/scripts/smoke-test-all-tables.sh` — 33/33 tables PASS
- [x] `test_type_frozen_udt` passes and no longer accepts `Frozen(Null)`

### DIFF
--- a/cqlite-core/src/storage/sstable/reader/parsing/v5_compressed_legacy.rs
+++ b/cqlite-core/src/storage/sstable/reader/parsing/v5_compressed_legacy.rs
@@ -2579,6 +2579,55 @@ impl V5CompressedLegacyParser {
                     offset += blob_len;
 
                     (udt_value, offset)
+                } else if let Some(udt_def) = self
+                    .udt_registry
+                    .as_ref()
+                    .and_then(|reg| reg.get_udt(&self.keyspace, &inner_type).cloned())
+                {
+                    // frozen<short_udt_name>: look up the concrete UDT definition in the
+                    // registry (Issue #502).  This handles type strings like
+                    // `frozen<person>` where "person" is a registered UDT rather than a
+                    // collection or a full marshal-format UserType string.
+                    log::debug!(
+                        "V5CompressedLegacy: Resolving frozen UDT '{}' via registry for column '{}'",
+                        inner_type,
+                        column.name,
+                    );
+
+                    // Read VUInt-prefixed blob length (same framing as tuple and
+                    // marshal-format UDT cells).
+                    let (remaining, blob_len_raw) = parse_vuint(&data[offset..]).map_err(|e| {
+                        Error::corruption(format!(
+                            "Frozen UDT '{}' (column '{}'): failed to parse blob length: {:?}",
+                            inner_type, column.name, e
+                        ))
+                    })?;
+                    if blob_len_raw > MAX_CELL_VALUE_LENGTH {
+                        return Err(Error::corruption(format!(
+                            "Frozen UDT '{}' (column '{}'): blob_len {} exceeds maximum {}",
+                            inner_type, column.name, blob_len_raw, MAX_CELL_VALUE_LENGTH
+                        )));
+                    }
+                    let blob_len = blob_len_raw as usize;
+                    let len_bytes_consumed = data[offset..].len() - remaining.len();
+                    offset += len_bytes_consumed;
+
+                    if offset + blob_len > data.len() {
+                        return Err(Error::corruption(format!(
+                            "Frozen UDT '{}' (column '{}'): need {} bytes but only {} available",
+                            inner_type,
+                            column.name,
+                            blob_len,
+                            data.len() - offset
+                        )));
+                    }
+
+                    let udt_data = &data[offset..offset + blob_len];
+                    let (udt_value, _) =
+                        self.parse_udt_value(udt_data, 0, &udt_def, column, _reader)?;
+                    offset += blob_len;
+
+                    (udt_value, offset)
                 } else {
                     // Non-collection frozen type (e.g., frozen<tuple<...>>)
                     let mut inner_column = column.clone();
--- a/cqlite-core/src/storage/sstable/reader/parsing/v5_compressed_legacy.rs
+++ b/cqlite-core/src/storage/sstable/reader/parsing/v5_compressed_legacy.rs
@@ -2553,13 +2553,19 @@ impl V5CompressedLegacyParser {
                     let udt_def = Self::parse_udt_type_definition(&column.data_type)?;
 
                     // First read the VInt-prefixed blob length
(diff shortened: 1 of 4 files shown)


pmcfadin:
## Review address — commit d1a8007

### Item 1 — MAX_CELL_VALUE_LENGTH guard in marshal-format arm (MUST FIX)

**File**: `cqlite-core/src/storage/sstable/reader/parsing/v5_compressed_legacy.rs`  
**Change** (~line 2555–2568 after edit): renamed `blob_len` to `blob_len_raw`, inserted the identical guard block that the registry arm already had, then cast to `usize` only after the check passes. The two arms are now symmetric.

```rust
// before
let (remaining, blob_len) = parse_vuint(...)?;
let blob_len = blob_len as usize;

// after
let (remaining, blob_len_raw) = parse_vuint(...)?;
if blob_len_raw > MAX_CELL_VALUE_LENGTH {
    return Err(Error::corruption(format!(
        "Frozen UDT '{}': blob_len {} exceeds maximum {}",
        column.name, blob_len_raw, MAX_CELL_VALUE_LENGTH
    )));
}
let blob_len = blob_len_raw as usize;
```

No helper extracted — minimal change as requested.

---

### Item 2 — Error on unregistered frozen UDT bare identifier (SHOULD FIX)

**File**: same, `else` branch of the `frozen<>` arm (~line 2637–2688 after edit).

**How I distinguished bare UDT identifiers from legitimate fall-throughs**:

The `else` branch is reached when `inner_type` is not a `list<`, `set<`, or `map<` collection, is not a marshal-format UDT (`is_udt_type`), and is not in the UDT registry. Before recursing, I test two conditions:

1. `inner_type.contains('<')` — any container or tuple (`tuple<...>`, nested `frozen<...>`) has an angle bracket. True → recurse normally.
2. `KNOWN_PRIMITIVES.contains(&inner_type.as_str())` — a compile-time `const` slice of all 20 CQL scalar names. True → recurse normally.

If both are false the identifier is a bare word that matches no known type — an unregistered UDT name — and we return:

```rust
Err(Error::schema(format!(
    "frozen<{inner}>: UDT '{inner}' not found in registry for keyspace '{}'; \
     register it before reading",
    self.keyspace, inner = inner_type,
)))
```

**Test impact**: No existing test relied on the silent-Blob fallback for this code path. The test `test_legacy_format_allows_blob_fallback_with_feature` (skipped per instructions) tests a different code path (`parse_value_from_raw_bytes` `other =>` arm, not the `frozen<>` arm). All 1399 non-skipped tests pass without modification, including `test_regression_481_list_frozen_udt_parses_as_udt_not_blob` and all frozen/UDT tests.

---

### Item 3 — Doc update (DOC FIX)

**File**: `docs/write-support-limitations.md`  
Replaced the "may be affected" placeholder with a description of the new behaviour: `frozen<NAME>` is resolved via the UDT registry, UDT registration is required, and the actionable error message is shown. The "resolved in v0.9.1" note makes the fix discoverable.

---

### Local checklist

| Check | Result |
|---|---|
| `cargo fmt --all` | Clean |
| `cargo fmt --all -- --check` | Pass |
| `RUSTFLAGS="-D warnings" cargo clippy --workspace --all-targets --all-features` | 0 warnings / errors |
| `cargo test --package cqlite-core --features cli-helpers -- --skip test_legacy_format_allows_blob_fallback_with_feature` | **1399 passed, 34 ignored** |
| `smoke-test-all-tables.sh` | **33/33 PASS** |

Issue #552, opened 2026-05-30
write-support test targets are red on main and not in CI (write_read_roundtrip + 2 lib tests)


pmcfadin:
## Summary

Several `write-support`-gated test targets are **failing on `main`** and have been rotting silently because CI and the standard pre-push checklist scope to `--features cli-helpers` (and `cqlite-cli --test unit_tests`), which does **not** compile or run these `--features write-support` tests. They were surfaced during the v0.9.2 release work (#533/#492) and verified to fail identically on baseline `main` (i.e. pre-existing, not caused by those PRs).

This is the same class of gap that #514 fixed for the root-package integration tests: a test target that isn't in the CI matrix drifts red without anyone noticing.

## Failing tests (on `main`, `cargo test -p cqlite-core --features write-support`)

`--test write_read_roundtrip` — **9 failed / 116 passed** (verified directly on `main`):
- `data_multi::test_data_mixed_partition_sizes`
- `data_multi::test_data_cross_component_validation`
- `data_multi::test_data_partitions_token_order`
- `data_multi::test_data_many_partitions`
- `index::test_index_partition_key_digest`
- `index::test_index_roundtrip_large_offsets`
- `index::test_index_roundtrip_multiple_partitions`
- `index::test_index_roundtrip_via_write_engine`
- `summary::test_summary_offset_tracking_with_index`

`--lib` (write-support cfg):
- `storage::sstable::writer::data_writer::tests::test_row_ttl_uses_row_ttl_cell_flags` — fails deterministically on `main` (verified; the inline comment suggests it was assumed wall-clock-flaky, but it fails 3/3).
- `storage::write_engine::mutation` `test_clustering_key_ord_type_mismatch_panics` — reported failing by the #492 implementer (panic-expectation test); confirm.

## Why it matters

- These targets exercise the write/compaction **index offset, token ordering, and cross-component (Data/Index/Summary) correctness** — exactly the surface area touched by M5 write support and v0.9.2 (#492 streaming writer, #505/#533 compaction). A real regression here would be invisible to current CI.
- The standard checklist in CLAUDE.md (`cargo test --package cqlite-core --features cli-helpers`) gives a false green because `write-support` test targets aren't built.

## Proposed fix

[...]

Issue #548, opened 2026-05-30
Query engine: WHERE on UUID (and composite) partition keys returns 0 rows — value_to_row_key incomplete


pmcfadin:
## Problem

Surfaced while building the Perf 1.3 read micro-benchmarks (#538, Epic #541).

The query engine cannot resolve a `WHERE pk = <literal>` predicate when the partition key is a `UUID`. `value_to_row_key` only handles `Integer / Text / Float / Boolean`; `Value::Uuid` falls through, so:

- A point lookup `SELECT * FROM test_basic.simple_table WHERE id = '550e8400-...'` returns **0 rows** (the partition-key path can't build a row key from the UUID).
- The scan+filter fallback also fails: the parsed literal becomes `Value::Text` and is compared against the stored `Value::Uuid`, which never equals.

Nearly every table in `test-data/` uses a UUID partition key, and the only non-UUID single-partition-key fixture (`test_timeseries.time_bucketed_counters`, TEXT key) ships empty. Composite TEXT partition keys (e.g. `app_metrics ((application_id, metric_name), timestamp)`) are also untested through `WHERE`.

## Impact on #538

Because keyed point lookups aren't possible against the available fixtures, the Phase-1 read benches use honest `LIMIT`-bounded read proxies (`read.point_lookup` = minimal single-row read; `read.clustering_slice` = bounded first-partition clustering read) instead of predicate-driven lookups. Once this is fixed, those two benches should switch to true keyed `WHERE` lookups for a sharper regression signal.

## Suggested fix

Extend `value_to_row_key` (and the partition-key resolution path) to handle `Uuid`/`Timeuuid` (and, ideally, composite partition keys), so `WHERE pk = <uuid-literal>` resolves to a partition seek. Add a query-correctness test covering a UUID-partition-key point lookup against a real fixture.

## Context

Not blocking Epic #541 (the benches ship with documented proxies), but it caps the value of the read regression gate until addressed.

pmcfadin:
Additional evidence from cqlite-perf (downstream benchmark harness, v0.9.2) that **extends this beyond UUID/`value_to_row_key`** — and fills the gap this issue flags ("the only non-UUID single-partition-key fixture ships empty").

I built an authentic Cassandra-5.0 corpus with a **TEXT single partition key**: `CREATE TABLE perf.basic (id text PRIMARY KEY, name text, email text, age int, payload text)`, 100k rows, keys `k{i:016x}`. Results via `Database::execute_streaming`:

[...]

Issue #573, opened 2026-05-31
Epic: Reliable perf measurement (gate fidelity + WAL-off path)


pmcfadin:
## Epic: Reliable perf measurement (gate fidelity + WAL-off path)

The in-repo Criterion perf gate (Epic #541) produces an **untrustworthy read**. Root cause is structural, not tuning:

- `write/ingest_wal_on` is **fsync-bound**: every row does `write()` → `wal.sync()` → `sync_all()` (`cqlite-core/src/storage/write_engine/mod.rs:419-439`, `wal.rs:338`). On a shared `ubuntu-latest` runner, fsync latency swings 10-100ms run-to-run, so the bench measures disk/OS, not cqlite code. This produced the `+11.6%` false regression on docs-only PRs #567/#570 (#572).
- There is **no WAL-off path** to compare against (#547) — the public API has no durability toggle, so `#539` could not ship the intended `ingest_wal_off` variant.
- Gate stats are thin: `--sample-size 20 --warm-up-time 1 --measurement-time 3` vs a flat **10%** threshold, single-shot PR-vs-main on the same noisy runner (`perf-gate.json`, `scripts/ci/check_perf_regression.py:74`). With ~10% fsync jitter and a 10% threshold, false positives are baked in.
- Trigger is too broad: fires on `cqlite-core/**`, which includes docs/examples under that tree (#572).

### Decision (recorded)
**Split the gate: CPU-bound benches are strict pass/fail; the fsync-bound `ingest_wal_on` is advisory/tracked, never blocking.** A shared CI runner cannot produce an authoritative absolute throughput number for fsync-bound work — that goal is served by documented methodology + (future) fixed hardware, not the gate.

### Children (sequenced — #547 is the keystone)
1. [x] **#547** — WriteEngine WAL on/off durability toggle *(keystone, no deps — start here)* — PR #577
2. [x] **#574** — Add `ingest_wal_off` bench variant *(deps #547)* — PR #578
3. [x] **#572** — Gate redesign: **Phase A** path-filter no-runtime-code PRs *(no deps, quick win)*; **Phase B** CPU-strict / I/O-advisory split + median-of-N / per-bench tolerances *(deps #574)* — PR #579
4. [x] **#575** — Perf methodology & reproducibility doc *(deps #574, #572; answers the cqlite-perf "282 ops/sec" question)* — PR #580

### Loosely related (CI-trust cleanup, independent)
[...]

PR #584, opened 2026-06-01
perf(index): restore O(1) raw-key point-lookup via Index.db

### LINKED ISSUE #553: Index.db point lookup misses (hash digest vs raw-key) — get() always falls back to scan after #552
## Summary

After #552, Index.db entries are correctly parsed and keyed on the **raw partition key** (the real Cassandra BIG format `[key_len u16][raw key]...`). However, the point-lookup path `SSTableReader::lookup_partition_with_index` / `lookup_partition_with_schema_context` (`cqlite-core/src/storage/sstable/reader/partition_lookup.rs`) still computes a **hash digest** via `compute_partition_key_digest` and looks that up in the raw-key HashMap. The digest never matches a raw key, so the index lookup always misses and every caller falls back to a sequential Data.db scan (the #517 fallback).

## Impact

- **Correctness is fine** — all callers (`get_with_spec_readers`, `get_with_schema_context`, and the `iterate_all_partitions` path) fall back to `sequential_scan`/`get()`, so reads return correct results.
- **Performance**: point lookups on multi-partition SSTables are O(n) (full scan) instead of O(1) (index seek). The Index.db is effectively unused for lookups.

This is pre-existing behavior (the digest-vs-stored mismatch predates #552 and is why #517 added the scan fallback); #552 made the *parsing* correct but did not wire up raw-key lookups.

## Proposed fix

Make `lookup_partition_with_index` match by the **raw partition key bytes** against the raw-key HashMap, removing the `compute_partition_key_digest` hash step for BIG-format Index.db. Care needed for **composite partition keys**: the Index.db stores the Cassandra composite framing (`[comp_len u16][value][0x00]` per component); the query key bytes must be serialized into the same framing to match, or the lookup must compare against the same representation the scan path uses. Validate against `test_basic/multi_partition_table` (composite key) and single-key tables, and confirm `get()` returns the partition via index (no scan fallback) with a test.

## Not in scope

- Index.db parsing/format (done in #552).
- Summary.db sampling.

## References

- Follow-up to #552 (documented limitation in `partition_lookup.rs::lookup_partition_with_index`).
- Related: #517 (get() scan fallback).

pmcfadin:
## Summary

**Performance fix (not correctness — reads were already correct).** After #552, Index.db entries are keyed on the **raw** partition key, but `lookup_partition_with_index` still computed a Murmur3 **digest** and looked *that* up in the raw-key map → always missed → every `get()` fell back to an O(n) sequential Data.db scan.

## Fix

`cqlite-core/src/storage/sstable/reader/partition_lookup.rs` — `lookup_partition_with_index` now passes the raw `partition_key` bytes straight to `index_reader.lookup_partition()` (O(1) HashMap hit), dropping the digest step. `compute_partition_key_digest` is retained (`#[allow(dead_code)]`, still used by the schema-context path). On a miss it still returns `Ok(None)` so existing fallbacks hold.

The expected raw-key form (shared with #548) is `PartitionKey::to_bytes`: single-component = raw value bytes (UUID = 16); composite = `[len:u16 BE][value][0x00]` per component.

## Tests — `cqlite-core/tests/issue_553_index_raw_key_lookup_tests.rs` (4)

- `test_lookup_partition_with_index_uuid_fast_path` — **the #553 regression test**: drives `SSTableReader::lookup_partition_with_index` with a real UUID raw key and asserts `Some((offset,…))`, i.e. the **index fast path** resolved (not the scan fallback). **Verified: returns `None` before the fix, `Some` after.** Self-skips when datasets are absent.
- `test_index_reader_raw_key_lookup_uuid` — synthetic Index.db, UUID raw-key → `Some(offset=256)`.
- `test_lookup_returns_some_for_simple_table_uuid` / `…_multi_partition_table` — real Index.db, UUID + 38-byte composite raw keys resolve via `IndexReader::lookup_partition`.

> Orchestrator note: the fast-path test was originally `#[ignore]`d. It already self-skips without data, and the other three resolve raw keys directly through `IndexReader` (which works on `main` post-#552), so the ignored test was the *only* one exercising this PR's change. Un-ignored so the fix is actually validated in CI (which fetches datasets).

## Docs

- `docs/performance.md` — new section on the point-lookup cost model (O(1) index vs O(n) scan fallback).
- `CHANGELOG.md` → Unreleased/Fixed; resolves the "Restoring O(1) raw-key point lookup is tracked in #553" note from v0.9.2.

## Gate

`fmt --check`, `clippy -D warnings --all-targets --all-features`, `cargo test -p cqlite-core --features cli-helpers` all pass.

Closes #553

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `stitch_and_parse_all_chunks` so V5CompressedLegacy rows parse during the scan
   fallback without an external schema.
 
+- **Index.db point-lookup performance cliff** (Issue #553) — `lookup_partition_with_index`
+  previously computed a Murmur3 digest of the raw partition key and looked that digest up in
+  the Index.db `key_lookup` map, which is keyed on **raw** partition key bytes (since #552).
+  The digest never matched, so every `get()` call fell back to an O(n) sequential scan of
+  Data.db. Results were always correct but at O(file-size) cost per lookup.
+
+  Fix: the digest computation (`compute_partition_key_digest`) has been removed from the
+  hot path. `lookup_partition_with_index` now passes the raw `partition_key: &[u8]` bytes
+  directly to `index_reader.lookup_partition`, restoring the O(1) HashMap lookup that was
+  present before #552 changed the key representation. Callers already pass raw bytes
+  (single = raw value bytes, composite = `[len u16 BE][value][0x00]` per component).
+
+  `lookup_partition_with_schema_context` (the schema-driven variant) is unchanged.
+
 ### Added
 
 - **Performance methodology doc** — new `docs/performance.md` (Issue #575)
--- a/cqlite-core/src/storage/sstable/reader/key_digest.rs
+++ b/cqlite-core/src/storage/sstable/reader/key_digest.rs
@@ -6,13 +6,18 @@ use super::super::key_digest::KeyDigestComputer;
 use super::types::SSTableReader;
 
 impl SSTableReader {
-    /// Compute partition key digest for Index.db lookup
+    /// Compute partition key digest for Index.db lookup.
     ///
     /// This method computes the digest from raw partition key bytes using the appropriate
     /// byte-comparable encoding based on the table schema's partition key definition.
     ///
     /// For modern formats (BIG v5, BTI), this MUST use schema-driven Murmur3 digest.
     /// Simple digest is only allowed for legacy formats behind feature flag.
+    ///
+    /// NOTE (Issue #553): This function is no longer called from the main
+    /// `lookup_partition_with_index` path, which now uses raw key bytes directly.
+    /// It is retained for potential future use and for `lookup_partition_with_schema_context`.
+    #[allow(dead_code)]
     pub(crate) async fn compute_partition_key_digest(
         &self,
         partition_key: &[u8],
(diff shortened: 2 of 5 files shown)


pmcfadin:
## Bench gate: this is runner noise, not a regression (re-running to confirm)

The "Bench regression vs main" gate flagged two strict read benches:

| bench | delta | verdict |
|---|---|---|
| read/point_lookup | +11.2% | REGRESSION (>10%) |
| read/clustering_slice | +10.6% | REGRESSION (>10%) |
| read/full_scan | +9.8% | ok |
| read/type_heavy | +2.0% | ok |
| write/ingest_wal_off | +0.0% | ok |
| write/ingest_wal_on | +0.9% | ok [advisory] |
| write/flush | +9.2% | ok |

**This is a uniform ~+9–11% offset across benches this PR cannot affect**, which is the signature of a slower ephemeral runner for the PR sample vs the base sample — not a code regression:

- This PR's only change is in `lookup_partition_with_index`: it **removes** a Murmur3 digest computation and replaces an O(n) sequential Data.db fallback scan with an O(1) `HashMap` hit. It can only make point lookups **faster**.
- Yet `write/flush` (+9.2%) and `read/full_scan` (+9.8%) — which never call `lookup_partition_with_index` — moved by the same ~10%. A genuine regression would be localized to the changed path, not spread uniformly across unrelated write and scan benches.
- `read/point_lookup` going *up* is the opposite of this change's intent; the absolute number here is dominated by the Data.db read, so the digest-removal saving sits inside per-runner variance.

Per epic #573, these CPU benches are "strict" on the assumption of low variance, but runner-to-runner variance can still breach a 10% threshold when the whole machine is ~10% slower. Re-running the bench job to get a second sample on a fresh runner; expecting it to clear. No code change warranted.

Issue #253, opened 2026-01-07
investigate(query): Root cause analysis for execution path divergence + fix sql→cql terminology


pmcfadin:
## Problem Statement

In `cqlite-core/src/query/engine.rs` (lines 132-142), there is a brittle workaround:

```rust
// For simple WHERE id = <value> queries, use normal executor for consistent key handling
// This ensures INSERT and SELECT use the same key generation logic
if sql.contains("WHERE id =") && sql.split_whitespace().count() <= 8 {
    #[cfg(debug_assertions)]
    log::debug!(
        "Routing simple SELECT through normal executor for consistent key handling"
    );
    // Fall through to normal execution path for simple point lookups
} else {
    return self.execute_select_query(sql, start_time).await;
}
```

## Why This Exists

The comment indicates that the two execution paths produce **different results** for partition key handling. The workaround:
1. Uses string matching (`sql.contains("WHERE id =")`) - fragile
2. Uses word count heuristic (`split_whitespace().count() <= 8`) - arbitrary
3. Forces simple queries through the legacy path for "consistent key handling"

## Investigation Goals

1. **Identify the divergence point**: Where exactly do the two paths differ?
   - Legacy path: `QueryParser` → `QueryPlanner` → `QueryExecutor`
   - Advanced path: `select_parser` → `SelectOptimizer` → `SelectExecutor`

2. **Analyze key generation**:
   - Legacy: `executor.rs::condition_to_row_key()` (line 794-805) uses `format!("user_key_{}", id)`
   - Advanced: `select_executor.rs::decode_partition_key_value()` (line 1095-1189) uses schema-aware decoding
   
3. **Determine which is correct**: The schema-aware approach in `SelectExecutor` is likely more correct for SSTable access

4. **Create parity tests**: Develop tests that verify both paths produce identical results

## Suspected Root Cause

The legacy `QueryExecutor` generates keys like:
```rust
if condition.column == "id" {
    if let Value::Integer(id) = &condition.value {
        let key_value = format!("user_key_{}", id);
        return Ok(RowKey::new(key_value.into_bytes()));
    }
}
```

The advanced `SelectExecutor` uses proper CQL type deserialization based on schema metadata. These are fundamentally different approaches.

## Deliverables

- [ ] Technical analysis document comparing both paths
- [ ] Root cause documented with code references
- [ ] Key generation logic compared between both paths
- [ ] Decision made on which approach is canonical
- [ ] Parity test suite created comparing both paths for:
[...]

PR #588, opened 2026-06-03
fix(query): reconstruct TEXT partition-key columns on the scan path (#586)

### LINKED ISSUE #586: Query engine: WHERE on a TEXT single partition key still returns 0 rows in v0.10.0 (PK column not materialized in SELECT results)
Follow-up to #548. The UUID case (#583) and `LIMIT` (#582) are fixed in v0.10.0 — thanks! But a **TEXT single partition key** still returns 0 rows, and the underlying cause I flagged on #548 is still present: the partition-key column isn't materialized into `SELECT` result rows.

## Repro (cqlite-core v0.10.0, tag)
Authentic Cassandra-5.0 corpus, `CREATE TABLE perf.basic (id text PRIMARY KEY, name text, email text, age int, payload text)`, 100k rows, keys `k{i:016x}`.

Isolation — residual filtering works for every **regular** column but fails only for the **partition key**:

| Query | Rows |
|-------|-----:|
| `SELECT * FROM basic WHERE age = 0` | 834 ✅ |
| `SELECT * FROM basic WHERE name = 'name-0'` | 1 ✅ |
| `SELECT * FROM basic WHERE email = 'u0@perf.test'` | 1 ✅ |
| `SELECT * FROM basic WHERE id = 'k0000000000000000'` | **0** ❌ |

Dumping a scanned row's columns confirms the mechanism:
```
columns: [payload, name, age, email]      # <-- no `id`
raw key = RowKey([107,48,48,...])         # = ASCII "k0000000000000000", correct
```
The key bytes are present and correct; `id` is just never reconstructed into the row (`build_row_from_scan`, select_executor.rs), so a residual `id = '...'` compares against a row with no `id` value → 0 matches. Same on composite-PK TEXT (`wide_rows (pk text, ck int)`), which also blocks clustering-range slices.

## Repro commands (cqlite-perf repo)
- `cargo run --example probe_nonpk` — the table above.
- `cargo run --example probe_cols` — dumps a scanned row's columns + raw key.

Engine-only report (downstream harness pins by tag; no code changes proposed here).

pmcfadin:
Closes #586

## The bug (shipped in v0.10.0)

On the **scan + residual-filter** path — the path a `WHERE` on a TEXT partition key falls to, unlike the Index.db point-lookup path used for UUID keys (#548/#553) — partition-key columns are not in the cell payload and are reconstructed from the raw row key. The reconstructor (`select_executor::decode_partition_key_value`) assumed a `u16` length prefix for **every** TEXT key. That is the *composite-component* framing, not the *single-component* layout (raw bytes).

| Case | Stored layout | Old decoder result |
|------|---------------|--------------------|
| Single UUID PK | 16 raw bytes | ✅ (reads 16 bytes) |
| Single TEXT PK (`id text PRIMARY KEY`) | raw UTF-8 | ❌ read phantom `u16` → `Err` → **swallowed** → column dropped |
| Composite PK (`((a,b),…)`) | `[u16 len][bytes][0x00]` per component | ❌ every column decoded from component[0] |

Result: `SELECT *` was missing the TEXT PK column and `WHERE id = '<literal>'` returned 0 rows; composite PKs returned wrong values for second+ components (non-text became debug strings).

## Diagnosis (gap A vs gap B)

Confirmed **gap B** with evidence (instrumented `build_row_from_scan` against real fixtures): the schema **is** threaded into the scan path (gap A is fine). The defect is in the decoder.

Captured raw `RowKey` bytes:
- `simple_table` (UUID PK): `[0,35,236,…]` — 16 raw bytes, no framing.
- `app_metrics` (composite text PK): `[0,1,'a',0, 0,4,'view',0]` — `[u16 len][bytes][0x00]` framing.
- A single TEXT PK is raw UTF-8 with no prefix → old TEXT branch ran off the end and the `Err` was silently swallowed.

## The fix

Extract the canonical partition-key decoder into a new **always-compiled** module `storage::partition_key_codec` and route **both** the read/scan path and the write engine's `PartitionKey::from_bytes` through it (single source of truth, no parallel impl). The codec was previously the write-engine's `deserialize_value_bytes` + framing logic, which was gated behind `write-support` and unavailable to the read path. A failed reconstruction is now `log::warn!`-ed instead of swallowed.

## Tests (fail before / pass after)

- **Unit** (`partition_key_codec`, `select_executor`): single raw text + uuid, composite framing per-component, `build_row_from_scan` materialises a single TEXT PK, and `WHERE id = '<literal>'` matches.
- **Integration** (`tests/issue_586_partition_key_reconstruction.rs`, real fixtures): composite TEXT PK decodes each component independently (`metric_name` is `interest`, was `goal`), `SELECT *` includes all PK columns, a `date` PK component decodes to `Value::Date` (was a debug string), and a clustering-range slice returns rows.

> Note: there is no scannable **single-component TEXT-PK** table in the fetchable corpus (the only single-text-PK tables are counter tables, whose counter cells don't parse into a row, plus system tables the scan skips). That exact scenario is covered by the deterministic `build_row_from_scan` / codec unit tests; the integration tests use the real composite TEXT-PK fixtures (`app_metrics`, `stock_prices`).

## Docs

- Rustdoc on the new codec and `build_row_from_scan`.
- CHANGELOG `[Unreleased]` entry.
- Appendix F: composite framing corrected (it omitted the trailing `0x00` the code actually emits) and #586 documented as resolved.

## Local gate (all green)

- `cargo fmt --all -- --check`
- `RUSTFLAGS="-D warnings" cargo clippy --workspace --all-targets --all-features`
- `cargo test -p cqlite-core` (incl. codec/select_executor unit tests; mutation round-trip with `write-support`; issue_586 with `cli-helpers`)
- `cargo test -p cqlite-integration-tests`


pmcfadin:
## CI: one red check — stale characterization test, not a regression

The fix itself looks correct and well-built — PK reconstruction now goes through the canonical `storage::partition_key_codec::decode_partition_key_columns` (`select_executor.rs:236`), the previous silent `if let Ok(..)` swallow is replaced with a logged failure, and there are regression tests for both the `SELECT *` completeness and `WHERE id = '<literal>'` cases. 👍

The only failing check is **not** a behavior regression and **not** the known `data_writer` flake — don't just rerun it:

```
test test_key_generation_divergence_documented ... FAILED
thread 'test_key_generation_divergence_documented' panicked at
  cqlite-core/tests/execution_path_parity_tests.rs:264:5
```

### Root cause
`test_key_generation_divergence_documented` is a *source-text grep* test. It asserts that `src/query/select_executor.rs` literally contains the string `decode_partition_key_value`:

```rust
let advanced_function = "decode_partition_key_value";
// ...
assert!(
    select_executor_content.contains(advanced_function),
    "ADVANCED path should contain '{}' function in select_executor.rs",
    advanced_function
);
```

This PR (correctly) relocated that logic out of `select_executor.rs` into the canonical `partition_key_codec` module and calls `decode_partition_key_columns` instead. So the grep finds nothing and the assert fails. The test's other two asserts (`user_key_` in `executor.rs`, the 8-token routing hack in `engine.rs`) still pass.

The test is pinning an implementation detail — *which file a function name appears in* — that this refactor legitimately changed.

### Suggested fix (pick one)
1. **Update the assertion** to the new canonical location, e.g. grep `src/storage/partition_key_codec.rs` for `decode_partition_key_columns` instead of `select_executor.rs` for `decode_partition_key_value`.
2. **Better:** retire/relax this characterization test. It documents the Issue #253 LEGACY-vs-ADVANCED divergence, which this PR is actively unwinding — a source-text grep over three files is brittle by design and will keep breaking on refactors. Consider asserting behavior (which path a query routes to) rather than the presence of literal strings.

Either way it's a small change; the rest of CI (builds, parity, integration, coverage, perf gate) is green.

pmcfadin:
Thanks for the precise diagnosis — agreed on both counts (not a regression, not the `data_writer` flake).

Fixed in 9bd25bb4. I took a blend of your two suggestions: the test now asserts the **architecture** rather than a literal file/function string, so it stays meaningful but won't false-flag this kind of refactor again.

The ADVANCED-path probe in `test_key_generation_divergence_documented` now checks:
1. the canonical decoder `decode_partition_key_columns` lives in `src/storage/partition_key_codec.rs`, and
2. `select_executor.rs` **delegates** to it (`partition_key_codec::decode_partition_key_columns`),

instead of grepping `select_executor.rs` for the now-removed inline `decode_partition_key_value`. Docstring updated to reflect the #586 relocation; the LEGACY (`user_key_`) and 8-token routing-hack probes are unchanged.

Local: `fmt --check`, `clippy -D warnings --all-features`, and the full `execution_path_parity_tests` file (7/7) all green.

Issue #1197, opened 2026-06-29
Writer does not emit CRC.db component for uncompressed BIG SSTables


pmcfadin:
## Discovered by
Found during #1190 (write-load byte-parity fixtures). **Out of scope** for #1190 — filed for prioritization.

## Symptom
`cqlite-core/src/storage/sstable/writer/finish.rs` emits `Digest.crc32` (whole-file digest) but does **not** emit a `CRC.db` component (per-chunk CRC), which Cassandra 5.0.2 emits even for **uncompressed** tables. The CQLite read side already recognizes `CRC.db` (per `cass.sstable_format.toc_component_manifest`); the write side does not produce it — a component-completeness gap.

Note: this does **not** affect the bytes of any emitted file; it's a missing TOC component. The #1190 test pins it as a reference-only assertion, so closing this will force a manifest review.

## Root cause (hypothesis)
Write path in `finish.rs` lacks per-chunk CRC emission for the uncompressed BIG format.

## Affected files
- `cqlite-core/src/storage/sstable/writer/finish.rs`

## Routing
**Oracle-driven** — Cassandra component set is the source of truth. Skip OpenSpec. (Note: lowest priority — no P3 label exists, filed P2.)

## Acceptance criteria
- [ ] CQLite writer emits a `CRC.db` component for uncompressed BIG SSTables matching Cassandra's per-chunk CRC layout.
- [ ] TOC.txt component set matches the Cassandra reference for an uncompressed table.
- [ ] Pinned parity test asserts `CRC.db` presence + contents; `scripts/agent-gate.sh` PASS.

Refs #1190, #969.

pmcfadin:
🧭 **MANAGER** 
GO
ORDER: 2

Elevated **P2→P0**: writer omits the **CRC.db** component for uncompressed BIG SSTables — a missing-component divergence. Oracle-driven → skip OpenSpec; pin a parity test asserting CRC.db presence+contents vs a Cassandra-written reference, then fix the writer.

pmcfadin:
🤖 Claimed by flow-lead worker — branch `issue-1197-crc-db-uncompressed` pushed (lock). Per MGR `mgr-mbp-0628` ORDER 2 (P0). Oracle-driven → skip OpenSpec. Implementing CRC.db per-chunk emission for uncompressed BIG + pinned parity test vs Cassandra reference.

pmcfadin:
✅ Resolved via PR #1215 (squash `6830445c`).

[...]

Issue #1116, opened 2026-06-26
Epic: Break up oversized source files (>2000 lines) — behavior-preserving module splits


pmcfadin:
## Goal

Bring CQLite's largest source files down to a maintainable size by splitting them **by responsibility** into focused submodules — **without changing behavior**. Every split is a pure refactor, guarded by the existing test suite, `scripts/agent-gate.sh`, and (for the write/compaction path) the #842 differential byte-parity harness.

## Why

An audit (see `docs/large-files-audit.md`, branch `claude/large-source-files-audit-u4xt0a`) found **165 production source files over 500 lines**, including **3 files over 10,000 lines** and **20 over 2,000**. Large files impose costs in two ways:

### Primary driver: context cost during agentic coding

The main reason to do this now is that **agents must read a file before they can edit it**, so file size is a direct, per-edit token tax — and it is *format-blind* (a big file costs the same to load whether it's logic or tests):

| File size | ≈ tokens to load | Effect on an agent edit |
|---|---|---|
| 500 lines | ~6–8k | trivial — load whole file, edit freely |
| 2,000 lines | ~25–30k | noticeable — crowds out other needed context |
| 4,800 lines | ~60–70k | one read dominates the turn |
| 13,800 lines | ~170–200k | exceeds a practical working budget — forces chunked reads + grep-first |

Two compounding effects beyond raw tokens:
- **Edit-anchor ambiguity** — the editor needs a *unique* string to match; in large/repetitive files the agent must read more just to find an unambiguous anchor, and mis-targets more often.
- **Working-memory dilution** — even when a big file technically fits, it squeezes out the schema, the code-under-test, and the task, degrading edit quality.

Smaller, responsibility-scoped files mean an agent loads only the relevant module to make a change — cheaper, faster, and more accurate edits.

### Secondary: human maintenance cost

- Slow navigation; functions too large to hold in working memory (e.g. `merge.rs::reconcile_cluster` was a single 455-line function).
- Whole-file recompiles on any edit.
- Merge-conflict collisions that **force serial work** — epic #938 explicitly serializes its children because they all edit one function in one file (see #945).

This epic does **not** claim refactoring fixes correctness bugs. It targets the cost and risk multiplier that large files impose on every future change — most acutely, agentic ones.

## Scope

Every production source file (`*/src/`, excluding test crates) **over 2,000 lines**, except:
- **`merge.rs`** — already tracked by **#945** (+ #947 for the shared reconcile-rule layer).
[...]

Issue #1018, opened 2026-06-24
Add TTL, tombstone, range tombstone parity and close ignored regressions


pmcfadin:
Parent epic: #973

## Goal

Add deterministic compaction parity coverage for TTL cells, partition/row/cell tombstones, and range tombstone markers. Close the two pinned ignored regressions in `cqlite-core/tests/issue_819_differential_compaction.rs`.

## Cassandra coverage

`RangeTombstoneTest`, `RangeTombstoneListTest`, `RangeTombstoneBurnTest`, `SSTableSkippingReadTest`, `NeverPurgeTest`, `TTLExpiryTest`, `GcCompactionTest`, `CompactionControllerTest`, `CompactionIteratorTest`, `CompactionsPurgeTest`, and `ForceCompactionTest`.

## Manifest entries

- `cass.tombstone_ttl.TTLExpiryTest.gc_boundary`
- `cass.tombstone_ttl.NeverPurgeTest.preserve_all_tombstone_types`
- `cass.tombstone_ttl.RangeTombstoneTest.marker_merge_and_persistence`
- `cass.compaction.GcCompactionTest.row_cell_partition_tombstone_gc`
- `cqlite.issue_819.differential_row_tombstone_wide_partition_regression`
- `cqlite.issue_819.differential_input_merge_write_fidelity`

## Evidence type
- canonical_semantic
- byte_for_byte

## Artifacts compared
- bytes
- offsets
- component files
- JSONL
- logs
- generated report

## CI tier
- required_parity
- nightly_docker

## Acceptance criteria

- Add scenarios for retained and purged TTL cells using explicit `--gc-before` and `--now-sec`.
- Add partition, row, and cell tombstone scenarios, including never-purge behavior and purge boundary checks at `localDeletionTime` and `localDeletionTime + 1`.
- Add bounded, open-ended, and index-boundary range tombstone scenarios; prove markers persist by re-compacting output against older resurrected rows.
- Canonical comparison includes row kind, deletion timestamp, local deletion time, TTL, cell timestamp, cell deletion state, and range marker bounds.
- Fix and unignore `differential_row_tombstone_wide_partition_regression`.
- Fix and unignore `differential_input_merge_write_fidelity`.
- Byte diffs are emitted for every scenario, even when manifest temporarily marks some cases canonical semantic until byte parity is clean.

## Out of scope

- Repair-aware tombstone purging.
- Read-repair behavior.
[...]

Issue #899, opened 2026-06-20
compaction parity: per-element/per-cell merge representation (reader→merge→writer) — the real #886 foundation


pmcfadin:
**Foundation issue for epic #842.** Discovered while implementing #844: the merge-entry struct enrichment in #886 added the *fields* (`CellData.cell_path`/`ttl`/`local_deletion_time`, `MergeEntry.complex_deletions`/`range_deletion`) but nothing **populates or consumes** them, because CQLite's read→merge→write path is fundamentally **whole-cell / row-timestamp granular**, not per-element/per-cell.

## What the investigation found
- The SSTable reader **decodes** per-element cell metadata (per-element timestamp, ttl, localDeletionTime, cell_path) and the collection-level complex deletion at the byte level (`reader/parsing/v5_compressed_legacy.rs` `parse_complex_column_inner`), but then **collapses** a multi-cell collection/UDT into a single `Value::Map/Set/List/Udt` of plain values and reduces the complex deletion to a boolean — discarding all per-element metadata before the compaction stream.
- The compaction emit contract is `(RowKey, Value, i64)` (`reader/data_access.rs` `stream_all_partitions_for_compaction` / `iterate_all_partitions_for_compaction`); the `i64` is a single **row-level** timestamp. `parse_one_partition_with_timestamps` has no slot for complex/range deletions.
- `value_to_row_data` (merge.rs) builds one `CellData` per **top-level column**, all sharing the row timestamp, with `ttl/local_deletion_time/cell_path = None`. `reconcile_cluster` therefore never sees per-element timestamps or per-cell ttl/ldt.
- The writer (`data_writer.rs` `write_complex_column` + `write_{set,map,list}_complex_cells`) re-emits per-element cells but stamps **every element with one `timestamp_micros`** and writes the complex deletion as a hardcoded `LIVE` sentinel; `CellOperation` has no per-element timestamp slot.

## Why it blocks byte parity
Per-element / per-cell reconciliation cannot be byte-faithful regardless of tie-break logic, because the metadata never reaches the merge and the writer cannot stamp elements individually. roborev correctly rejected the #844 value-union shortcut (timestamp promotion + complex-deletion resurrection).

## Scope (LARGE — the real foundation #886 was meant to be)
[...]

PR #1219, opened 2026-06-29
fix(#1018): preserve per-cell write timestamps through compaction merge + parity coverage

pmcfadin:
Refs #1018 (core slice — see "Scope" below; intentionally does NOT auto-close).

## What this lands
**A real compaction write-path bug** + un-ignores two pinned `issue_819` regressions + parity manifest coverage.

### The bug (regression `differential_input_merge_write_fidelity`)
When a reconciled row carried a **cell tombstone alongside live sibling cells**, the writer rewrote the live siblings' per-cell write timestamps to the row's MAX timestamp (the tombstone's ts). Root cause was upstream of the serializer: `reconcile.rs::build` sets `row_ts = max(surviving cell ts)`, then `cells_to_cell_operations` lowered each live cell to `CellOperation::Write` — a variant carrying **no per-cell timestamp** — so every cell inherited the row max.

Fix (per-cell write-timestamp side-channel, mirroring the existing `row_tombstone`/`local_deletion_time` precedent):
- `mutation.rs`: add `cell_write_timestamps: Option<HashMap<String,i64>>` (+ accessor).
- `merge/mod.rs`: `merge_entry_to_mutation` records each surviving simple live cell whose own ts differs from the row max (cell tombstones/complex excluded; single-writetime rows stay `None`).
- `data_writer/rows.rs`: `merge_row_group` uses the per-cell ts for the emitted op, the whole-col liveness row-marker candidate, and the surfaced-clustering-cell liveness; the existing dispatch then clears `USE_ROW_TIMESTAMP` for differing cells.
- `wal.rs`: a backward-compat layout mirror + decode fallback so pre-#1018 WAL records still replay (bincode is positional).

### Regressions un-ignored
- `differential_input_merge_write_fidelity` — fixed by the above; now passes (+2 focused unit tests on per-cell ts preservation).
- `differential_row_tombstone_wide_partition_regression` — its defect was already fixed (#899/#972); the `#[ignore]` was stale → removed. The `issue_819_differential_compaction` suite is now **6 passed, 0 ignored**.

### Manifest + report
6 parity scenarios added to `cassandra-parity-manifest.yml` (`lint: 0 errors`), report regenerated. NOTE: the issue listed two ids as `cqlite.issue_819.*`, but the manifest schema mandates `^cass\.…` and has no `compaction`/`cqlite` capability — so they were **remapped to schema-conforming `cass.*`** ids (`cass.compaction_merge.issue_819.*`, `cass.tombstone_ttl.*`) following existing precedent. Flagging this remap for the manager to reverse if undesired.

## Scope — why `Refs` not `Closes`
#1018's ACs also include **adding new deterministic TTL/partition-tombstone/range-marker compaction *scenarios*** to the #819 harness. Those are deferred to a follow-up (this PR is already ~700 insertions touching the `Mutation` struct + WAL format; bundling more would balloon it). **The close-vs-keep-open decision for #1018 is the manager's** — surfaced on the issue. Much of the byte-parity those scenarios imply already exists in `issue_992_ttl_tombstone_range_parity.rs`.

## Validation
- `scripts/agent-gate.sh` = **PASS** (rebased on origin/main incl. #1197). `rows.rs` +40 lines acked via `CQLITE_ALLOW_FILE_GROWTH=1` (#1116 note in-file).

### DIFF
--- a/cqlite-core/examples/gen_wide_sstable.rs
+++ b/cqlite-core/examples/gen_wide_sstable.rs
@@ -89,6 +89,7 @@ async fn main() -> cqlite_core::error::Result<()> {
             range_tombstones: vec![],
             local_deletion_time: None,
             row_tombstone: None,
+            cell_write_timestamps: None,
         };
         engine.write_async(mutation).await?;
     }
--- a/cqlite-core/src/storage/sstable/writer/data_writer/encoding.rs
+++ b/cqlite-core/src/storage/sstable/writer/data_writer/encoding.rs
@@ -448,12 +448,29 @@ pub(crate) fn collect_static_operations(
                 } => continue,
                 crate::storage::write_engine::mutation::CellOperation::DeleteRow => continue,
             };
+            // Issue #1018: a static `Write`/`WriteWithTtl` cell carries its OWN
+            // write timestamp when the compaction merge→mutation path recorded one
+            // in `Mutation::cell_write_timestamps` (it differs from the row's
+            // `timestamp_micros`). Use it for BOTH the stamped candidate timestamp
+            // AND the last-write-wins comparison below, mirroring the regular-row
+            // path in `rows.rs`. Otherwise a compacted static row with surviving
+            // static siblings at differing writetimes would rewrite older static
+            // cells to the newest static mutation's row max. For every other op (and
+            // for cells with no per-cell override) this is exactly
+            // `mutation.timestamp_micros`, so the single-writetime case is unchanged.
+            let candidate_ts = match op {
+                crate::storage::write_engine::mutation::CellOperation::Write { .. }
+                | crate::storage::write_engine::mutation::CellOperation::WriteWithTtl { .. } => {
+                    mutation.cell_write_timestamp(&col_name)
+                }
+                _ => mutation.timestamp_micros,
+            };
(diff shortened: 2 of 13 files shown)


pmcfadin:
**Merging over the red `Bench regression vs main` lane — verified NOISE, not a regression.**

The lane flagged `write/ingest_wal_off: +10.4%` (threshold 10%). Stable local measurement (`--sample-size 60 --measurement-time 10`, 3 runs each) of merge-base `1f0840c4` vs PR HEAD shows **no regression**: base medians 1.527/1.523/1.551 ms (≈1.534 ms) vs PR 1.451/1.530/1.477 ms (≈1.486 ms) — the PR is marginally *faster*, with fully overlapping criterion CI bands. A control PR-vs-PR run reproduced a spurious +12%/−6% swing, confirming the +10.4% is a sampling artifact of CI's low-sample config (20 samples / 3 s) on this ~1.5 ms microbench. #1018's changes touch only the flush/compaction write path; on the WAL-off ingest path `cell_write_timestamps` is always `None`, never allocated/cloned, and the `None` accessor short-circuits before any hashing. The lane is advisory/non-required (main has no required CI contexts). Gate PASS, roborev code-clean (5 rounds), all other lanes green.

Issue #1227, opened 2026-06-29
Epic: Test suite asserts more coverage than it proves — coverage-integrity hardening


pmcfadin:
## Why this epic

A four-pattern audit (triggered while reviewing #1017) found that CQLite's test suite **systematically over-claims coverage**: it asserts breadth via hand-maintained constants and lenient guards instead of deriving coverage from ground truth, running fail-closed, and asserting on content. The common failure mode is a **green test that proves less than its name/claim implies**.

#1017 itself was the trigger (a manifest scenario declared a *verified byte_for_byte* claim that ran nowhere) and is being fixed correctly in PR #1225 — but the audit shows it is one instance of a broad pattern.

## The four patterns (each a child issue)

1. **Parity-manifest overstatement** — 30 `required_parity` scenarios whose named CI workflow provably never runs the mapped test (12 are P0 `byte_for_byte` data-loss claims). → child issue.
2. **Frozen-allowlist "all N tables" coverage** — corpus has 18 keyspaces; "all 33 tables" tests track 4–6; tautological count gates; 8 keyspaces covered by zero comprehensive test while all report 100%. → child issue.
3. **Skip-on-absence green-on-missing-data** — ~70 dataset-dependent tests skip-and-pass; main CI lanes don't set `CQLITE_REQUIRE_FIXTURES=1` and sanity-check only one table. → child issue.
4. **Write path proven nowhere end-to-end through a public surface** — CQL→read-back-content is tested in two disjoint halves; public write APIs assert exit-code/rows-affected/file-exists, never content. → child issue.

## Unifying acceptance theme
Where feasible, coverage should be **derived from ground truth** (enumerate the committed corpus / the manifest), **fail closed** (missing fixture/data ⇒ failure, not skip; "no tests ran" ⇒ failure), and **assert on content** (read-back values / byte diffs), not on shape or exit codes. Each child carries its own concrete acceptance criteria.

## Out of scope
- #1017 / `test_compactionparity` / `live_cells_*` (fixed in PR #1225).
- #1222 (compaction CRC.db full byte parity) — separate.

Children: (linked below once filed)

pmcfadin:
### Children (filed)

| # | Pattern | Severity |
|---|---------|----------|
| #1228 | Parity-manifest overstatement — 30 `required_parity` scenarios run in no workflow (12 P0 `byte_for_byte`) | **HIGH** (P1) |
[...]

Issue #773, opened 2026-06-14
Python test suite: CQLITE_DATASETS_ROOT path bug silently skips ~120 tests and hides 7 real failures


pmcfadin:
## Summary

`bindings/python/tests/conftest.py` mis-resolves the dataset path when
`CQLITE_DATASETS_ROOT` is set, which is the documented + CI convention
(`CQLITE_DATASETS_ROOT=\$PWD/test-data/datasets`). The result is that **~120
data-dependent Python tests silently skip in CI**, and when the path is
corrected, **7 of them fail** against real data. None of this is currently
visible because the tests skip before they can fail.

## Root cause

```python
# conftest.py (current, on main)
_ENV_DATASETS_ROOT = os.environ.get("CQLITE_DATASETS_ROOT")
if _ENV_DATASETS_ROOT:
    DATASETS = Path(_ENV_DATASETS_ROOT)          # <-- parent dir, missing /sstables
else:
    DATASETS = TEST_DATA / "datasets" / "sstables"  # correct
```

The actual SSTable files live under `datasets/sstables/`, but when the env var is
set, `DATASETS` points one level too high (`datasets/`). Opening a `Database`
there succeeds but reads 0 rows. The env-var-unset fallback is correct, so this
only manifests in CI / when following the documented convention.

## Evidence

Discovered while fixing PR #772 (Parquet lift, epic #682). With the path
corrected, the Python suite goes from **`282 passed, 128 skipped`** to
**`402 passed, 3 skipped, 7 failed`** — i.e. the fix activates ~120 previously
skipped tests, of which 7 fail.

## The 7 failures unmasked (all pre-existing on `main`, unrelated to Parquet)

| Test | Error |
|------|-------|
| `test_cli_parity.py::TestCLIParityCollections::test_collection_table_parity[collections_with_udts]` | `TypeError: unhashable type: 'dict'` |
| `test_parity.py::TestRowCountParity::test_collections_row_count[collections_with_udts]` | `TypeError: unhashable type: 'dict'` |
| `test_parity.py::TestE2ESummary::test_e2e_all_33_tables` | fails for `test_collections.collections_with_udts` |
| `test_cli_parity.py::TestCLIParityCollections::test_collection_table_parity[typed_collections_table]` | `decimal_set` value mismatch |
| `test_cli_parity.py::TestCLIParityTimeseries::test_timeseries_table_parity[log_entries-10]` | `stack_trace` value mismatch |
[...]

Issue #1231, opened 2026-06-29
Write path proven nowhere end-to-end: public CQL-write surfaces assert shape, not content


pmcfadin:
Child of epic #1227. Pattern 4 (the "CI is blind to the write path" hazard, localized).

## Problem
The write path is validated as **two green halves that never meet through a public surface**:
- **CQL → Mutation** tests assert only the in-memory `Mutation` struct (`cqlite-core/src/storage/write_engine/cql_to_mutation/builders.rs:816`).
- **Mutation → SSTable → read-back** tests build `Mutation`s **by hand**, bypassing CQL (`cqlite-core/tests/write_read_roundtrip.rs:221` `create_simple_mutation(...)`).

No single test drives the full public chain a user hits: **CQL string → parser → Mutation → WriteEngine → flush → SSTable → read-back-correct-values**. Every test that DOES hit a public write surface asserts only exit-code / `rows_affected` / file-exists, never content:
- CLI: `cqlite-cli/tests/cli_dml_integration_tests.rs:370` (`test_cli_execute_insert_then_flush`) asserts exit-success + "OK", never reopens the SSTable. No `SELECT`/read-back in the file.
- Python: `bindings/python/tests/test_write_api.py` — asserts `rows_affected`, file exists, stat counters; **grep `SELECT` → 0 hits**. Never reopens the flushed SSTable.
- Node: `bindings/node/__test__/write.test.js:300` — the one post-write SELECT explicitly tolerates 0 rows and asserts no content; the file header claims a reopen+verify pattern that **does not exist** in the file.

Also: `export-sstable`, `compact`, `maintenance`, `write-stats` CLI subcommands (`cqlite-cli/src/main.rs:922-1002` → `commands/write.rs`) have **no binary-level test** of their handler glue — only the underlying core engine fns are tested directly. `export-sstable` (the primary path to Cassandra-importable SSTables) has zero end-to-end coverage.

A write-format/encoding bug that emits a structurally-present but semantically-wrong Data.db passes green today.

(Adequately wired, no action: core write byte-parity `issue_1190_write_load_byte_parity.rs`; core compaction roundtrips; Flight `do_get`/`get_flight_info` driven via the real handler methods.)

## Acceptance criteria
[...]

PR #1246, opened 2026-06-29
fix(#1230): make dataset-dependent tests + CI lanes fail-closed on missing fixtures

### LINKED ISSUE #1230: Skip-on-absence: ~70 dataset tests pass on missing data; main CI lanes not fail-closed
Child of epic #1227. Pattern 3.

## Problem
Two cultures coexist. Newer `issue_10xx`/`issue_99x` parity tests are correctly **fail-closed** (honor `CQLITE_REQUIRE_FIXTURES=1`, panic on missing/empty). But ~70 legacy dataset-dependent tests **skip-and-pass silently** and never consult a strict flag — and `ci.yml`, `python-ci.yml`, `node-ci.yml` run them **without** the strict flag, sanity-checking only `test_basic/simple_table`. So any table dropping out of the asset, or a #773-class path regression, turns these green instead of red.

Worst offenders (HIGH — can go green on missing data in a CI lane):
- `cqlite-core/tests/sstable_parity_integration_test.rs` — **double-skips**: env unset → `eprintln!("Skipping"); return` (L549-552 et al.) AND the `Err` arm swallows "not set"/"not found" (L579-586); assertions are thin `partition_count > 0` (L605+). Labeled "parity" but passes on a near-empty corpus.
- `cqlite-core/tests/reader_compression_tests.rs:385-425` — `continue`s past tables with no `Data.db`, asserts only `success_count > 0`.
- `cqlite-cli/tests/integration_sstable_tests.rs` — every test prints "Skipping … test data not available" and returns.
- The `issue_949/953/955/956/960/961` seek/access-path cluster — "0 rows → skip" idiom (the empty-dataset-passes hazard).

Python (#773 lineage): the path-resolution bug **is fixed** (`conftest.py:42-48` accepts both `<root>` and `<root>/sstables`), but there's still **no fail-closed flag and no "no tests ran" floor** — `pytest` exits 0 when everything skips, and `python-ci.yml`'s quality gate only checks `needs.test.result == success`.

Good (no action): Node bindings are strict — `helpers.js:37-41 skipIfNoDatasets()` THROWS on missing data; `smoke-tests.yml` and `sstabledump-parity-gate.yml` are workflow-level fail-closed.

## Acceptance criteria
- [ ] `ci.yml` / `python-ci.yml` / `node-ci.yml` set `CQLITE_REQUIRE_FIXTURES=1` (+ `CQLITE_PARITY_REQUIRE_DATASETS=1` where relevant) for the dataset-dependent test steps, OR the legacy tests adopt the existing `require_fixtures_strict()` helper so a missing fixture FAILS in CI.
- [ ] Add a "no tests ran / min-collected" floor to the Python (and Rust where feasible) dataset lanes so an all-skip run is a FAILURE, not a pass.
- [ ] Strengthen the two acutely-mislabeled "parity" tests (`sstable_parity_integration_test.rs`, `reader_compression_tests.rs`): stop swallowing "not found", replace `count > 0` with content assertions or fail-closed presence checks.
- [ ] Broaden the CI dataset sanity-check beyond `test_basic/simple_table` (or assert the fetched asset's table set against an expected manifest) so a partial extraction reds CI.

## Routing
Design/infra (which lanes, how strict, the min-collected floor). Lead to confirm; likely no OpenSpec, mostly CI + test hardening.

pmcfadin:
Closes #1230. Child of epic #1227 (coverage-integrity). Oracle/infra hardening.

## Problem
~70 legacy dataset-dependent tests skip-and-pass silently on missing data, and `ci.yml`/`python-ci.yml`/`node-ci.yml` ran them without a strict flag (sanity-checking only `test_basic/simple_table`). A dropped table or #773-class path regression turned them GREEN instead of red.

## Changes (all 4 acceptance criteria)
1. **CI lanes fail-closed** — `CQLITE_REQUIRE_FIXTURES=1` (+ `CQLITE_PARITY_REQUIRE_DATASETS=1` on Rust lanes) on the dataset-dependent steps of all three workflows.
2. **"No tests ran" floor** — Python `conftest.py` now `pytest.fail()`s under strict instead of skipping, plus a session-finish hook that fails the run if 0 tests passed under strict. Rust floor = strict env (absent fixture panics).
3. **Strengthened the two mislabeled "parity" tests** — `sstable_parity_integration_test.rs` and `reader_compression_tests.rs`: stop double-skipping / swallowing "not found", replace `count > 0` with content/presence assertions, fail-closed under the strict flag.
4. **Broadened CI sanity check** — new `test-data/scripts/check-dataset-manifest.sh` asserts a Data.db for all 33 corpus tables, wired into the sanity step of all three workflows.

One shared `require_fixtures_strict()` helper (no parallel mechanism), honoring both env vars.

## Fail-closed proof (strict flag + fixture absent → FAIL, not skip)
- `reader_compression_tests`: FAILED — panic `CQLITE_REQUIRE_FIXTURES=1 but test_basic keyspace is absent`.
- `sstable_parity_integration_test`: 13 FAILED — `CQLITE_REQUIRE_FIXTURES=1 but fixture missing`.
- Counter-proofs: strict+present → pass; no-flag+absent → still skips (local dev preserved); manifest checker exits 1 on 1/33, 0 on 33/33.

## Gate
`scripts/agent-gate.sh` → **RESULT: PASS** (all 15 components). No file-growth override.

## Discovered (filed separately)
- `agent-gate.sh` smoke step hardcodes `$PWD/target/debug/cqlite`, incompatible with a shared `CARGO_TARGET_DIR` from a worktree.

🤖 worker: oracle/infra, 1:1:1:1. Conflict in ci.yml with main's f75b1e61 resolved (kept both the golden-present echo and the manifest check).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,9 +108,9 @@ jobs:
           echo "✅ Core fixture present: ${CORE_FIXTURE}"
           echo "✅ Required JSONL golden present: ${GOLDEN_JSONL}"
           # Issue #1230: broaden the sanity check beyond the single core fixture —
-          # assert the WHOLE 33-table validation corpus is present so a partial
-          # extraction or a dropped table reds CI here instead of letting the
-          # dataset-dependent lanes skip-and-pass.
+          # assert the WHOLE 39-table enforced corpus (33 nb + 6 test_oa) is
+          # present so a partial extraction or a dropped table reds CI here
+          # instead of letting the dataset-dependent lanes skip-and-pass.
           bash test-data/scripts/check-dataset-manifest.sh test-data/datasets
 
       - name: Validate environment
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,6 +107,11 @@ jobs:
           fi
           echo "✅ Core fixture present: ${CORE_FIXTURE}"
           echo "✅ Required JSONL golden present: ${GOLDEN_JSONL}"
+          # Issue #1230: broaden the sanity check beyond the single core fixture —
+          # assert the WHOLE 33-table validation corpus is present so a partial
+          # extraction or a dropped table reds CI here instead of letting the
+          # dataset-dependent lanes skip-and-pass.
+          bash test-data/scripts/check-dataset-manifest.sh test-data/datasets
 
       - name: Validate environment
         run: |
@@ -140,6 +145,11 @@ jobs:
       - name: Run core lib/doc tests with timings
         env:
           CQLITE_DATASETS_ROOT: ${{ github.workspace }}/test-data/datasets
+          # Issue #1230: fail-closed. The ~70 legacy dataset-dependent tests skip
+          # on missing data; with these flags an absent/empty fixture FAILS the
+          # lane (via require_fixtures_strict) instead of false-greening.
+          CQLITE_REQUIRE_FIXTURES: "1"
+          CQLITE_PARITY_REQUIRE_DATASETS: "1"
           RUST_LOG: debug
           RUST_BACKTRACE: full
         run: bash scripts/ci/run-core-test-group.sh lib-doc
@@ -317,6 +327,9 @@ jobs:
       - name: Run integration tests (issue #514)
         env:
           CQLITE_DATASETS_ROOT: ${{ github.workspace }}/test-data/datasets
+          # Issue #1230: fail-closed for the dataset-dependent golden-path lane.
+          CQLITE_REQUIRE_FIXTURES: "1"
+          CQLITE_PARITY_REQUIRE_DATASETS: "1"
         run: |
           echo "Running revived orphan integration tests (issue #514)..."
           # Run the seven golden-path / chunked-reader / component tests that were
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -207,9 +207,12 @@ jobs:
         working-directory: bindings/node
         env:
           CQLITE_DATASETS_ROOT: ${{ github.workspace }}/test-data/datasets
-          # Issue #1230: fail-closed flag for parity with the other CI lanes. The
-          # node helper (skipIfNoDatasets) already throws on missing data; this
-          # documents intent and is honored by the Rust core it calls.
+          # Issue #1230: for the node lane this flag is DOCUMENTARY only — it is a
+          # cqlite-core TEST-ONLY helper and is NOT consulted by the node binding
+          # runtime. Actual fail-closed coverage here comes from the
+          # check-dataset-manifest.sh step above (hard-fails on a partial/dropped
+          # corpus) plus the JS skipIfNoDatasets throw on missing data. The flag
+          # is set for parity of intent with the Rust/Python lanes.
           CQLITE_REQUIRE_FIXTURES: "1"
         run: npm test
 
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -199,8 +199,9 @@ jobs:
         run: |
           set -euo pipefail
           test -f test-data/datasets/metadata.yml
-          # Issue #1230: assert the full 33-table corpus (was: a non-failing
-          # `ls ... || true` that passed even on a partial extraction).
+          # Issue #1230: assert the full 39-table enforced corpus (33 nb + 6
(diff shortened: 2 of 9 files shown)


pmcfadin:
🔎 **roborev (claude-code/opus): 1 Medium + 4 Low.** Merge held until the Medium is cleared.

**Medium (must-fix, merge-blocker):** `assert_parity_content` asserts `partition_count >= reference_count` (strict 100%), which makes the pre-existing deliberate 95% tolerance gate in `test_simple_table_key_parsing_parity` dead code and could red a currently-green lane if real data parses to 95–99% of reference partitions. #1230's scope is fail-closed-on-MISSING-data, **not** tightening the parity ratio. **Fix:** preserve the existing 95% tolerance (have `assert_parity_content` honor the call site's intended ratio rather than impose 100%); keep the fail-closed-on-empty/missing behavior.

**Low (fixing for honesty — this epic is literally about not overstating safety nets):**
- node-ci.yml comment claims `CQLITE_REQUIRE_FIXTURES` is 'honored by the Rust core it calls' — it's a test-only helper, documentary for the node lane. Reword.
- python `pytest_sessionfinish` floor only fires on whole-session zero-pass; comment implies it catches dataset-only skips. Reword to actual scope.
- check-dataset-manifest.sh 33-table list + reader_compression's 8-table const are hand-duplicated — add cross-reference / single-source note.

**Low (pre-existing, defer):** missing-fixture classification is substring matching on error strings — note for a future typed-error refactor.

Fix dispatch is queued behind #1231's build (shared cargo target); will re-gate after.

pmcfadin:
🔎 **roborev re-check: new Medium + 3 Low.** The first fix restored 95% for the *simple_table* site but the deeper issue remains — merge still held.

**Medium (root cause, must-fix):** ~10 parity tests (composite_key, static_columns, collection_list, nested_collections, collections_with_udts, sensor_data, stock_prices, wide_partition, chat_messages, many_columns, all_tables_basic) were previously asserted only at `partition_count > 0`. #1230 raised them to a 95% ratio floor — a real parity-strength tightening, contradicting the 'does not tighten' comment and risking unrelated CI reds for a table legitimately parsing 1–94%.

**Convergence (stop the whack-a-mole):** #1230 must change NO table's parity-ratio pass criteria from its pre-#1230 baseline. Keep ONLY the fail-closed-on-missing/empty behavior:
- simple_table site → keep its original 95%.
- the ~10 tables that were `> 0` → assert *reference golden present (reference_count>0) AND partition_count>0* — fail-closed-on-empty, NO ratio floor.
- Fix the `assert_parity_content` doc-comment contradiction (test_collection_table_map_parsing legitimately stacks a separate `>=50` absolute pin).

**Low (fix):** manifest asserts 33 `nb` tables but validation-matrix.md enforces 39 (33 + `test_oa`'s 6); the 'full corpus' comment overstates → a dropped `test_oa` extraction goes uncaught while test_oa tests skip = the exact false-green #1230 targets. Add test_oa's 6 if reliably shipped in the asset, else correct the comment to '33-table nb corpus only.'
**Low (defer):** table-list triplication already has a TODO to derive from metadata — leave.

Fix queued behind #1231's build (shared target).

pmcfadin:
🔎 **roborev re-check (convergence): 1 Medium + 2 Low (both already-deferred TODOs).** One last wrinkle.

**Medium:** `test_collection_table_map_parsing` still carries BOTH `assert_parity_content(&r, 95)` and its `>= 50` absolute pin. Currently a no-op (golden = 50 partitions → 95%≈47 < 50, so `>= 50` dominates), but it's fragile (a larger regenerated golden would silently activate the 95%) and contradicts the 'no ratio tightening' claim + the helper's doc. **Fix:** drop the 95% call from the map test, keep only its true pre-#1230 baseline (`>= 50` pin) + fail-closed-on-empty.

**Low (defer, already TODO'd):** substring error classification (#1230 follow-up TODO present); table-list triplication (derive-from-metadata TODO present).

Final fix dispatched; will re-gate + re-roborev → merge.

pmcfadin:
✅ **Final roborev: 3 Low, no Medium/High — Medium cleared.** All non-blocking: substring-classification + table-list-triplication are pre-existing/TODO'd; the 'strict requires all 8 test_basic tables open' Low is the INTENDED fail-closed behavior (roborev: 'none required') — exactly #1230's goal, distinct from the parity-ratio creep that was removed. Gate PASS + roborev clean of actionable findings → merging on green.

PR #1268, opened 2026-06-30
ci: reuse python CLI binary in tests

pmcfadin:
## Summary
- Build the cqlite CLI binary in native Python wheel jobs while the Rust target is already warm.
- Upload the CLI binary as an artifact per native target.
- Download the CLI artifact in Python test jobs and remove the separate test-job CLI rebuild.

## Baseline
Recent green Python CI run 28412606982 completed in 20m12s.

Critical path details from that run:
- Windows wheel build: 7m56s
- Windows test job: 11m04s
- Windows CLI build step inside test job: 8m25s
- Windows pytest step: 2m02s

## Expected impact
This should move the CLI compile work into the native wheel jobs where the target directory is already warm, then reduce test jobs to install, dataset check, and pytest. The Windows path remains the likely critical path, but should avoid the full fresh CLI rebuild inside tests.

## Local validation
- Parsed all workflow YAML with Ruby YAML.
- git diff --check passed.
- cargo build --package cqlite-cli --release passed locally and produced target/release/cqlite.

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -25,7 +25,7 @@ env:
   DATASET_SHA256: bebc763752c8d68c7fb0483a1b31294b4d1d21343d3f7d124da069e5073202fa
 
 jobs:
-  build-wheels:
+  build-only-wheels:
     name: Build wheel (${{ matrix.target }})
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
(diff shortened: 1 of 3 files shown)


pmcfadin:
Closing as superseded by #1278. The artifact/graph reshaping experiment here did not produce a real end-to-end Python CI timing improvement; #1278 kept full slow CLI parity on Ubuntu and removed the duplicated macOS/Windows CLI builds, which measured as the winning path.

PR #1274, opened 2026-06-30
fix(#1254): sort non-UDT frozen<map>/frozen<set> by serialized key/element bytes (byte-parity)

### LINKED ISSUE #1254: Writer does not sort non-UDT frozen<map>/frozen<set> by serialized key/element bytes (byte-parity)
## Summary
Discovered while fixing #1020 (frozen-UDT compaction parity). CQLite's writer canonicalization sorts **UDT-bearing** frozen `MapType`/`SetType` collections by serialized key/element bytes (fixed in #1020, `cqlite-core/src/storage/sstable/writer/data_writer/udt_canon.rs`), but **non-UDT** frozen maps/sets (e.g. `frozen<map<text,int>>`, `frozen<set<int>>`) take the `references_user_type` fast-path borrow and are written in **input insertion order**.

## Why it's a bug
Cassandra `MapType`/`SetType` are sorted collections — frozen maps are serialized sorted by key bytes, frozen sets by element bytes (key/element `AbstractType` comparator = unsigned serialized-byte order for these types). A CQLite **direct write** of a non-UDT frozen map/set with unsorted keys/elements therefore produces **non-Cassandra bytes**.

## Why it's currently latent (not caught by #1020)
The #1020 parity fixtures feed already-key-sorted inputs, so current parity stays green. The gap only manifests on arbitrary unsorted direct-write input.

## Routing
Oracle-driven (Cassandra/sstabledump bytes are ground truth) → GitHub issue + pinned byte-parity test, no OpenSpec.

## Acceptance criteria
- Non-UDT frozen `map`/`set` values are sorted by serialized key/element bytes before write, matching Cassandra (reuse the writer's own `serialize_value` for the sort key, as #1020 did for the UDT path).
- Pinned unit test: a `frozen<map<text,int>>` and a `frozen<set<int>>` written with reversed keys/elements serialize in sorted-byte order.
- Ideally a byte-for-byte compaction/flush parity fixture vs a Cassandra 5.0.2 golden with a multi-entry non-UDT frozen map.

## Affected code
- `cqlite-core/src/storage/sstable/writer/data_writer/udt_canon.rs` (sort applied only on the UDT-referencing path)
- the non-UDT frozen collection write path (the `references_user_type` fast-path borrow)

Found-by: #1020 (see PR for the UDT-path fix).

pmcfadin:
Closes #1254.

## What & why
Cassandra `MapType`/`SetType` are **sorted** collections: a frozen map serializes entries in unsigned serialized-key-byte order, a frozen set its elements in serialized-element-byte order. In `serialize_value` the `Value::Map` and combined `Value::List | Value::Set` arms emitted entries/elements in **input insertion order**, so a CQLite direct write of a non-UDT `frozen<map<…>>` / `frozen<set<…>>` with unsorted keys/elements produced **non-Cassandra bytes**. Latent because prior fixtures fed already-sorted input. Oracle-driven (Cassandra/sstabledump bytes are ground truth) — no OpenSpec.

## The fix (`encoding.rs`, `serialize_value`)
- Split the `List | Set` arm. `Value::List` keeps insertion order (`ListType` is not sorted).
- `Value::Set`: serialize each element via the writer's own `serialize_value`, collect, `serialized.sort()` — the **same single comparator** the non-frozen SET cell path (`write_set_complex_cells`) already uses.
- `Value::Map`: serialize each `(key,val)`, collect, `sort_by` on serialized **key** bytes — matching `write_map_complex_cells`.

No second comparator, no heuristic (no-heuristics mandate #28); null-key rejection preserved; no `unwrap`/`expect`.

## Tests (assert serialized byte order, not counts; verified RED on unmodified src)
- `test_frozen_set_sorted_by_serialized_element_bytes` — reversed `frozen<set<int>>` `[3,2,1]` → sorted blob `[1,2,3]`.
- `test_frozen_map_sorted_by_serialized_key_bytes` — reversed `frozen<map<text,int>>` keys `c,b,a` → sorted blob `a,b,c`.
- `test_frozen_map_row_emits_sorted_bytes_end_to_end` — full `write_row` contains the sorted blob and NOT the reversed blob.

## Validation
- **agent-gate.sh: PASS** — 15/15 (file-size, fmt, clippy, core-tests, tombstones-scan, scan-offload-guard, integration-tests, format-compat, write-tests, cli-tests, python-bindings, delivery-telemetry, tooling-tests, minimal-build, smoke).
- **roborev: clean for scope** — only finding is Low, explicitly *out-of-scope + pre-existing* (signed-numeric type-aware comparator; faithfully mirrors the existing non-frozen path, not a regression). Tracked as a follow-up.

## Follow-up (filed separately, NOT in this PR)
True parity for signed numeric element/key types (`frozen<set<int>>` with negatives, etc.) needs a **type-aware comparator** — raw unsigned-byte sort diverges from Cassandra's `Int32Type` ordering. This limitation is shared with the existing non-frozen `complex.rs` path. Filed for the manager to prioritize.

### DIFF
--- a/cqlite-core/src/storage/sstable/writer/data_writer/encoding.rs
+++ b/cqlite-core/src/storage/sstable/writer/data_writer/encoding.rs
@@ -214,7 +214,8 @@ pub(crate) fn serialize_value(value: &Value) -> Result<Vec<u8>> {
             let serializer = TypeSerializer::new();
             serializer.serialize_udt(value, &schema)
         }
-        Value::List(elements) | Value::Set(elements) => {
+        Value::List(elements) => {
+            // ListType preserves insertion order — do NOT sort.
             let mut buf = Vec::new();
             buf.extend_from_slice(&len_as_i32(elements.len())?.to_be_bytes());
             for elem in elements {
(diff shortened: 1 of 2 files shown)


pmcfadin:
Type-aware comparator follow-up (signed-numeric parity, shared with the non-frozen path) filed as #1275 — out of scope for this PR per #1254's brief.

Issue #917, opened 2026-06-21
perf(read): add a benchmark proving >1× concurrent-scan scaling on one SSTableReader (follow-up to #815)


pmcfadin:
> Follow-up to #815 (epic #906).

## Background

#815 removed the `scan_mutex` full-scan serialization and gave each scan its own `ScanCursor`, so concurrent scans on a single `SSTableReader` now run in parallel. It shipped a **correctness** stress test (`test_concurrent_scans_single_reader_are_consistent`: 16 concurrent full scans return identical results on buffered + mmap backends).

The #815 acceptance criteria also noted, as a *nice-to-have*, "ideally a throughput assertion / benchmark shows >1× scaling vs the serialized baseline." That part was deliberately deferred — correctness was proven, but the speedup was not quantified.

## Ask

Add a criterion benchmark (under the existing profiling harness, `./scripts/profile.sh` / `cqlite-core` benches) that:

- Measures aggregate throughput of N concurrent full scans on one `SSTableReader` for N ∈ {1, 2, 4, 8}.
- Reports scaling vs the N=1 baseline, on both the buffered and mmap backends.
- Optionally re-introduces a serialized variant (mutex-guarded) behind the bench only, to show the before/after delta directly.

This gives a durable number for the read-path performance epic and a regression guard against accidentally re-serializing the scan path.

## Notes / caveats

- Expected scaling is sublinear and workload/IO-bound — the goal is to **document** the real curve, not assert a hard multiplier (avoid a flaky timing gate; prefer reporting + a loose lower bound if asserted at all). See the project's flaky-perf-gate history before wiring this into CI.
- Pairs naturally with the #816 io_uring spike, which needs exactly this kind of baseline harness.

## Acceptance criteria

- [ ] Benchmark exists and runs via the profiling harness, covering N concurrent scans on one reader (buffered + mmap).
- [ ] Output reports per-N throughput and scaling vs N=1.
- [ ] Findings (the measured scaling curve) recorded on this issue or in `docs/profiling.md`.

pmcfadin:
Done in #920 (squash `563f3fd1`), merged to `main` with full CI green (33 checks).

[...]

Issue #815, opened 2026-06-17
perf(storage): restore parallel reads on a single SSTableReader (per-scan file handles / positioned reads)


pmcfadin:
> Part of epic #906 (Read-path performance & I/O backend).

## Background

Follow-up to #805 / PR #814.

#805 fixed a correctness bug: `SSTableReader` shares a single underlying file seek-position **and** a `current_chunk_index` counter across all callers, so two threads scanning the same reader concurrently interleaved their seeks/chunk-index advances, corrupting each other's reads (surfaced as `QueryError('Column not found: id')`).

The fix (PR #814) added a `scan_mutex: tokio::sync::Mutex<()>` held for the full lifetime of each scan in `sequential_scan`, `get_all_entries`, `scan_for_key`, and `run_scan_stream`. This makes concurrent scans **correct** but **serializes** them — only one scan per reader runs at a time.

Note: this is not a regression. Concurrent reads were already broken (corrupt) before #814; the mutex trades non-working parallelism for working serialization.

## Problem

Per-reader scans no longer run in parallel. Under multi-threaded read workloads against the same SSTable (e.g. the Python binding's concurrent-query path, or any future parallel scan/merge), throughput is bounded to one scan at a time per reader.

## Proposed direction

Eliminate the shared mutable cursor so scans don't need to serialize:

- Give each scan its **own file handle / reader cursor** (open per scan, or pool), instead of sharing one seek-position, **or**
- Use **positioned reads** (pread-style: `read_at(offset, len)` / `seek_read`) so no scan mutates a shared cursor at all, and decouple `current_chunk_index` from instance state (pass it through the scan, or make chunk lookup stateless/offset-keyed).

Either approach lets the `scan_mutex` be removed (or narrowed to just metadata init) and restores true concurrent reads.

## Acceptance criteria

- [ ] Concurrent scans on one `SSTableReader` run in parallel (no full-scan serialization) while remaining correct.
- [ ] `test_edge_cases.py::TestConcurrentAccess::test_concurrent_queries_from_threads` stays green (and ideally a throughput assertion / benchmark shows >1× scaling vs the serialized baseline).
- [ ] No shared mutable file-position or chunk-index state across concurrent callers (verified by code audit + a stress test).
- [ ] Agent gate passes; memory stays within the <128MB target (watch file-handle / buffer multiplication if opening per-scan handles).

## Scope notes

[...]

PR #1314, opened 2026-06-30
perf(#1143): default Auto disk-access to buffered (not mmap+SEQUENTIAL) — fix read-p99 regression under concurrent write

### LINKED ISSUE #1143: perf regression: read p99 under concurrent write load ~2x worse in v0.12.0 vs interim main (9054734)
## Summary

On a same-machine A/B, **read p99 latency under concurrent write load roughly doubled** between the post-#788/#790 interim main commit `9054734` and the **v0.12.0** tag (`b70a64e8`). A streaming `SELECT *` full-scan reader running alongside an ingest writer sees its p99 go from **~200 µs → ~371 µs** (~1.85×), with the two distributions **not overlapping** across 8 samples.

This is notable because **isolated** scan throughput *improved* ~41% over the same range (full_scan 332k → 470k rows/s). So this looks like a **contention/tail regression** (concurrent read-while-write), not a raw scan-speed regression.

> Reported from the external `cqlite-perf` macro-benchmark harness. The harness source is byte-identical across both measurements — **only the `cqlite-core` git commit changed** — so the harness is ruled out as the cause.

## Evidence (same machine, aarch64 10-core, today)

Workload: `mixed.read_while_write` — 2 writers doing sustained `WriteEngine::write` ingest + 6 readers each running a streaming `execute_streaming("SELECT * FROM perf.basic")` full scan over a static 100k-row S-tier lz4 corpus, conc=8, warmup=0, 10–12s, repeated. Metric is the **reader-side p99** (`custom.read.p99_us`).

| Sample | old `9054734` (µs) | v0.12.0 `b70a64e8` (µs) |
|---|--:|--:|
| 1 | 236,415 | 379,135 |
| 2 | 219,391 | 368,127 |
| 3 | 164,863 | 375,039 |
| 4 | 180,479 | 361,727 |
| **min–max** | **165k – 236k** | **362k – 379k** |

The new version's *minimum* (362k) is well above the old version's *maximum* (236k); the v0.12.0 distribution is also tighter. Mean ~200k → ~371k.

(Note: the companion `mixed.open_loop` read-p99 metric swings too wildly run-to-run on this machine — 157k/365k/550k/621k across versions — to support any claim, so it is **excluded**. The regression claim rests on `read_while_write` only.)

## Not a regression: isolated scan got faster

For contrast, in **isolation** (no concurrent writers) v0.12.0 is clearly faster over the same commit range:

| Read workload (S tier, lz4, 100k rows) | `9054734` | v0.12.0 | Δ |
|---|--:|--:|--:|
| read.full_scan (rows/s) | 332,351 | 469,807 | **+41%** |
| read.clustering_slice (rows/s) | 1,194 | 2,358 | +98% |
| read.wide_partition (rows/s) | 587,038 | 691,117 | +18% |

So the read path itself is healthier; the regression is specifically **under concurrent write load**.

## Bisect range & candidate areas

155 commits in `9054734..v0.12.0` ("the compaction release"). I have **not** root-caused it, but the changes most likely to affect scan tail latency under concurrent flush/compaction pressure are the reader-backend / parallel-read ones:

- `1021577` / `e26df3f` — **feat(reader): size-aware disk-access backend with configurable prefetch (#964)**
- `f5a1317` — perf(storage): restore parallel reads on a single SSTableReader (#815) (#915)
- `563f3fd` — perf(read): concurrent-scan scaling benchmark on a single SSTableReader (#917) (#920)
- `d8381b2` — Partition-targeted lookup for fully-constrained WHERE pk=? (#949) (note: these readers full-scan, not point-lookup, so this is a weaker candidate)

A prefetch/disk-access-backend change is a classic "improves isolated throughput, hurts tail under contention" shape — worth checking first.

## Repro (via cqlite-perf harness)

```
# pin cqlite-core to each commit in turn, then:
cargo run --release -- run --workload mixed.read_while_write \
  --concurrency 8 --warmup 0 --duration 10s --trials 1
# read custom.read.p99_us from reports/<date>/results.jsonl
```

Happy to provide the raw results.jsonl, run on a wider sample, or test a candidate fix branch.

pmcfadin:
Closes #1143.

## Root cause (bisect-confirmed)
#964 (`1021577`/`e26df3f`) added `disk_access_mode: Auto` + `prefetch: Auto` and silently flipped the default backend from **buffered I/O** to **mmap + `MADV_SEQUENTIAL`** for ordinary-sized SSTables. `MADV_SEQUENTIAL` does aggressive read-ahead **plus drop-behind**: with multiple readers full-scanning the same file while writers flush, each reader's drop-behind evicts pages the others still need (page-cache thrash) and read-ahead competes with flush I/O. That is exactly the reported shape — **isolated** scan throughput improved ~41%, but **read p99 under concurrent write load ~doubled** (~200µs → ~371µs). Verified against `git show 9054734:...config.rs`: the pre-#964 default was buffered.

Suspect verdict: #964 **CONFIRMED**; #815 (`f5a1317`), #917 (`563f3fd`), pk=? lookup (`d8381b2`) ruled out.

## Fix
`resolve_disk_access_mode`: `Auto` now resolves to **`Buffered`** for sub-RAM-fraction files (restoring the pre-#964 default) and **`Direct`** only when a file exceeds `direct_io_memory_fraction` (Direct uses its own per-cursor aligned buffer — no shared-page-cache drop-behind). Explicit `Mmap`/`Direct` and the legacy `use_mmap`/`CQLITE_USE_MMAP` opt-in are preserved (legacy promotion now applied **after** Auto resolution so it still works for default configs — fixes a roborev-caught back-compat break).

## Guard
New `cqlite-core/tests/issue_1143_default_disk_access_backend.rs` (run under `scan-offload-guard`) pins the structural invariant the regression turns on: default `Auto` for an ordinary SSTable must NOT resolve to `Mmap`. Proven to FAIL pre-fix, PASS post-fix. A structural invariant is used rather than a p99 assertion because absolute latency is machine-dependent and the tail only manifests under sustained multi-reader+writer load.

## Verification
- `scripts/agent-gate.sh` → **RESULT: PASS** (all components; `file-size` PASS under the documented `CQLITE_ALLOW_FILE_GROWTH=1` ack — this grows 3 pre-existing over-threshold files modestly; a split is out of scope for a targeted contention fix).
- roborev (codex, vs origin/main): **No issues found** (a MEDIUM back-compat finding on the `use_mmap` opt-in was fixed).

## ⚠️ Still needs owner action
End-to-end p99 validation requires the **external `cqlite-perf` harness** (`mixed.read_while_write`, conc=8) — cannot run in-repo. Recommend an A/B (this branch's buffered default vs v0.12.0 mmap default) to confirm read p99 returns toward the ~200µs baseline.

Routing: oracle-driven — no OpenSpec.

### DIFF
--- a/cqlite-core/src/config.rs
+++ b/cqlite-core/src/config.rs
@@ -106,18 +106,25 @@ pub struct StorageConfig {
     ///
     /// Defaults to [`DiskAccessMode::Auto`], which sizes each Data.db file
     /// against system RAM and picks the backend automatically:
-    /// - files below [`Self::mmap_min_size_bytes`] use buffered I/O (mapping a
-    ///   tiny file is not worth the setup cost);
-    /// - files up to [`Self::direct_io_memory_fraction`] of system memory are
-    ///   **memory-mapped**, so repeated scans stay resident in the page cache;
+    /// - files up to [`Self::direct_io_memory_fraction`] of system memory use
+    ///   **buffered I/O** through the OS page cache with ordinary kernel
+    ///   read-ahead — the safe, contention-friendly default;
     /// - files larger than that fraction use **direct I/O** (`O_DIRECT` on
     ///   Linux, `F_NOCACHE` on macOS), which bypasses the page cache so a
     ///   single huge scan does not evict everything else the host has cached.
     ///
+    /// `Auto` deliberately does **not** select memory-mapping (issue #1143):
+    /// mmap with `madvise(MADV_SEQUENTIAL)` read-ahead has aggressive drop-behind
+    /// that thrashes the shared page cache when multiple readers scan the same
+    /// file while a writer flushes, roughly doubling read p99 under concurrent
+    /// write load even though it speeds up an *isolated* scan. Use mmap only when
+    /// you ask for it explicitly.
+    ///
     /// Set an explicit [`DiskAccessMode::Buffered`], [`DiskAccessMode::Mmap`],
     /// or [`DiskAccessMode::Direct`] to override the heuristic. The legacy
     /// [`Self::use_mmap`] flag still forces mmap when `Auto` would otherwise
-    /// pick buffered, for backward compatibility.
+    /// pick buffered, for backward compatibility (and for the repeated-rescan
+    /// workload mmap was built for).
     ///
     /// Can also be set at runtime via `CQLITE_DISK_ACCESS_MODE`
     /// (`auto` / `buffered` / `mmap` / `direct`).
@@ -125,7 +132,7 @@ pub struct StorageConfig {
     pub disk_access_mode: DiskAccessMode,
 
     /// Fraction of total system memory above which [`DiskAccessMode::Auto`]
-    /// switches a file from memory-mapped to direct I/O. Defaults to `0.5`
+    /// switches a file from buffered to direct I/O. Defaults to `0.5`
     /// (half of RAM). Clamped to `(0.0, 1.0]`; values outside that range fall
     /// back to the default. Ignored when system memory cannot be determined
     /// (in which case `Auto` never escalates to direct I/O).
--- a/cqlite-core/src/storage/sstable/reader/mod.rs
+++ b/cqlite-core/src/storage/sstable/reader/mod.rs
@@ -428,17 +428,8 @@ impl SSTableReader {
 
         let file_size = tokio::fs::metadata(path).await?.len();
 
-        // The explicit mode comes from env > config. The legacy `use_mmap` flag
-        // is folded in by promoting an otherwise-`Buffered` request to `Mmap` so
-        // the old opt-in keeps working (`Auto` and explicit non-buffered modes
-        // are left untouched, so a real backend decision always wins).
+        // The explicit mode comes from env > config.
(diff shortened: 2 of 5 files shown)


pmcfadin:
Closing per owner decision. The external `cqlite-perf` harness A/B (warmed, order-alternated, real tier-S 100k LZ4 corpus) **disproves this approach**: Auto→Buffered makes #1143 *worse*, not better — read ~10.5 scans/s / p99 ~655ms (~3.6× baseline) vs current main ~20.6 / ~375ms (~2.1×) and baseline `9054734` ~41.8 / ~181ms. Buffered `read()` syscalls + buffer copies contend with writer I/O under load; mmap faults stay cheap under contention. The in-repo bench-gate red was correct signal (the +74% magnitude is a cache-resident micro-fixture artifact). Root cause was plausible-but-wrong; the disk backend is NOT the #1143 lever. #1143 stays OPEN — redirect to the post-#1265 residual (windowed-scan handoff batching + per-row alloc churn: build_row_from_scan / HashMap<String,Value> / to_lowercase / CqlType::parse), with an explicit external-harness p99 acceptance bar. Branch retained on the closed PR for reference.

Issue #1295, opened 2026-06-30
Commission Cassandra 5.0 golden fixture: set<int>/map<int,…> with negative elements/keys (frozen + non-frozen)


pmcfadin:
## Context
Follow-up from #1275 (type-aware signed comparator for set/map element ordering, merged via its PR). #1275 fixed the bug — set/map elements now order by the element/key CQL type's Cassandra `AbstractType` comparator (signed for numerics, Java-compatible float/decimal) instead of raw unsigned byte order.

The pinned tests in #1275 are **oracle-grounded but derived**: they assert the exact serialized byte layout computed from Cassandra's signed comparator rule + the documented frozen/complex-cell wire format (e.g. `frozen<set<int>>` of {3,-1,1,0,-2} → bodies in signed order -2,-1,0,1,3). No Cassandra-written golden with **negative** numeric collection elements/keys exists under `test-data/`, so the assertions are not yet byte-for-byte against a real Cassandra 5.0 SSTable.

## Goal
Commission a real Cassandra 5.0-written reference SSTable (via the test-data Docker tooling) containing:
- `set<int>` / `map<int,…>` and `frozen<set<int>>` / `frozen<map<int,…>>`
- rows whose collections include **negative** numeric keys/elements (and a mix of negative/positive)
- ideally also negative `bigint`/`smallint`/`tinyint` and a large-unscaled `decimal` + a NaN/±0.0 `double` element, to lock the other comparator branches added in #1275

Then upgrade the #1275 derived tests to assert byte-for-byte against the committed golden (force-commit the small `.db` reference binaries per the #1190 playbook; verify against a clean `git worktree add --detach HEAD`, not a dirty tree).

## Routing
Oracle-driven (Cassandra is ground truth). Fixture generation is slow under emulation — likely belongs on the nightly/exhaustive lane, not the required gate.

## Acceptance criteria
- Committed Cassandra 5.0 golden fixture(s) with negative numeric collection elements/keys (frozen + non-frozen).
- #1275's `scenarios_4.rs` negative-value tests upgraded from derived-byte to byte-for-byte against the golden (skip-on-absence guarded; 0-rows-when-present is a failure).

pmcfadin:
🧭 **MANAGER** 
GO

Cleared for the swarm — independent, oracle/CI/test-data scoped. implement → agent-gate → roborev → merge autonomously.

pmcfadin:
[...]

Issue #1275, opened 2026-06-30
Frozen + non-frozen map/set use raw unsigned-byte sort — diverges from Cassandra type comparator for signed numerics


pmcfadin:
## Summary
Both the **non-frozen** complex-cell path (`write_set_complex_cells` / `write_map_complex_cells` in `cqlite-core/src/storage/sstable/writer/data_writer/complex.rs`) and the **frozen** path (`serialize_value` Set/Map arms in `encoding.rs`, fixed in #1254) order set elements / map keys with a **raw unsigned serialized-byte sort** (`serialized.sort()` / `sort_by(a.0.cmp(b.0))`).

Cassandra `SetType`/`MapType` sort by the **element/key AbstractType's own comparator**, which for signed numeric types (`int`/`Int32Type`, `bigint`/`LongType`, etc.) is **signed** comparison. Raw big-endian two's-complement byte order ≠ numeric order for negatives: e.g. `-1` → `0xFFFFFFFF` sorts **last** under byte order but **first** under `Int32Type`.

## Why it's a bug
A `frozen<set<int>>` / `frozen<map<int,…>>` (or the non-frozen equivalents) containing **negative** keys/elements serializes in a different order than Cassandra → **byte-for-byte parity not met** for those cases.

## Why it's currently latent
- #1254's tests cover only positive `int` and `text`, where byte order coincidentally matches numeric/lexicographic order.
- Existing fixtures don't exercise negative numeric collection keys/elements.

## Scope / routing
Oracle-driven (Cassandra `AbstractType` comparators are ground truth). Affects **both** the frozen and non-frozen serialization paths — fix should be unified (a single type-aware comparator keyed off the schema/`AbstractType`, no heuristic).

## Acceptance criteria
- A shared type-aware comparator orders set elements / map keys per the element/key CQL type's Cassandra comparator (signed for numeric types), used by both `complex.rs` and `encoding.rs`.
- Pinned byte-parity test: `frozen<set<int>>` and `frozen<map<int,…>>` with **negative** values serialize matching a Cassandra 5.0 golden (and the non-frozen equivalents).
- No-heuristics: comparator derived from authoritative type metadata only.

## Found-by
roborev review of PR #1274 (#1254); also flagged by the #1020 follow-up commit `2f2766a0` ("order frozen map/set by key/element type comparator, not raw bytes").

pmcfadin:
🧭 **MANAGER** 
GO
ORDER: 1

[...]

PR #1317, opened 2026-06-30
feat(#1296): recursive comparator for composite collection element/key ordering

### LINKED ISSUE #1296: Collection element/key ordering: recursive comparator for composite types (tuple/UDT/nested frozen collections)
## Context
Follow-up from #1275, surfaced by roborev. #1275 added a shared type-aware comparator (`cqlite-core/src/storage/sstable/writer/data_writer/collection_order.rs`) that orders set elements / map keys per the element/key CQL type's Cassandra `AbstractType` comparator — correctly handling SCALAR signed numerics (int/bigint/smallint/tinyint), float/double (Java NaN order), varint, and decimal.

For **composite** element/key types it still falls back to **raw serialized-byte comparison**, which diverges from Cassandra:
- `set<tuple<int>>` / `map<tuple<int>, …>` — Cassandra `TupleType`/`CompositeType` compares field-by-field recursively using each field's comparator, so `tuple(-1)` sorts before `tuple(1)`; raw bytes put it after.
- `set<frozen<set<int>>>` and other nested frozen collections — elements should order by the inner collection's element comparator recursively, not raw bytes.
- frozen UDTs as set elements / map keys — field-by-field per the UDT field comparators.

**Not a regression**: main already raw-byte-sorts these composites; #1275 simply scoped to scalars and left the pre-existing behavior for composites. This issue closes that gap.

## Fix direction (from roborev)
Add recursive comparator handling for `Tuple`, `Udt`, and frozen/nested collection `Value`s in `collection_order.rs::compare_collection_elements`, OR thread the schema `ComparatorType` into the writer sort path so composites compare by their structural comparator instead of falling back to serialized bytes.

## Routing
Oracle-driven (Cassandra `TupleType`/`UserType`/`*Type` comparators are ground truth; local source `~/local_projects/cassandra`, `org.apache.cassandra.db.marshal`).

## Acceptance criteria
- `compare_collection_elements` orders composite element/key types (tuple, UDT, nested frozen collections) recursively by their structural Cassandra comparator — no raw-byte fallback for composites containing signed numerics.
- Pinned tests: `set<tuple<int>>` / `frozen<set<frozen<set<int>>>>` (and a frozen-UDT-keyed map) containing NEGATIVE inner numerics serialize in the recursive comparator order, asserted against the derived byte layout (and a Cassandra golden if #1295 lands one).
- No-heuristics: comparator derived from authoritative type metadata only.

pmcfadin:
Closes #1296 — follow-up to #1275.

## What
Recursive, type-aware comparator for **composite** SET-element / MAP-key types in the SSTable write path, so set/map ordering matches Cassandra for `tuple`, UDT, and nested frozen collections (previously these fell back to raw serialized-byte comparison, which diverges from Cassandra).

- Split `collection_order.rs` (at the 800-line campsite threshold) into a `collection_order/` submodule: `scalar.rs` (the unchanged #1275 scalar leaves) + new `composite.rs` (recursive `tuple`/UDT/`set`/`list`/`map` comparators) + `mod.rs` (dispatcher). Write-path call sites unchanged (route through the same `compare_collection_elements` entry).
- Each composite comparator mirrors the Cassandra 5.0 `compareCustom`: `TupleType`/`UserType` (positional, null-first, prefix-shorter-first), `SetType`/`ListType` (element-wise + length tiebreak; lists keep insertion order, sets canonicalized), `MapType` (key-then-value + length). Leaves recurse into the #1275 signed/float/varint/decimal/uuid logic — no reimplementation, no heuristics (#28).

## Notable fix (found via wiring-evidence review)
The write path did **not** sort nested frozen collections bottom-up: `set<frozen<set<int>>>` with unsorted inner data serialized non-canonically (outer ordered against unsorted inner while inner bytes were written sorted). The comparator now canonicalizes each side before comparing; an end-to-end serialize regression test (`collection_order_serialize.rs`) is proven to fail without the fix.

## Verification
- `scripts/agent-gate.sh`: **PASS**
- roborev: clean (2 Lows addressed — list-insertion-order test + perf note)
- Independent Cassandra `compareCustom` **semantics audit: SOUND** (all 6 rules cited against marshal source behavior)
- TDD-verified throughout (disabling each composite arm / the inner sort / a list-sort makes the corresponding tests fail)
- Oracle-anchored unit + e2e tests. No Cassandra golden fixture exists for these orderings yet — byte-parity goldens are tracked separately by **#1295**.


pmcfadin:
Merging over a non-required, flaky macos-14 failure: `test_parity.py::TestCoverageSummary::test_every_discovered_keyspace_is_classified` (1 failed / 372 passed). This is a **keyspace-classification** test, orthogonal to this PR's collection-ordering change. It passes on ubuntu-latest, windows-latest, an earlier macos-14 attempt, and the local `agent-gate.sh` (python-bindings PASS). macos-14 had dataset-fetch infra flakiness on this run (the first attempt failed at `Fetch datasets if cache miss`), which plausibly left inconsistent dataset state. Required status checks on main are empty; real coverage is green everywhere else.

Issue #944, opened 2026-06-22
Flight/Trino: surface row-count + per-column cardinality so the AUTOMATIC aggregation-pushdown gate can decide


pmcfadin:
Follow-up to #893 (closed by #937). This issue covers the **statistics-surfacing** half of #893's original scope, which #937 deliberately deferred. Until it's done, the new `AUTOMATIC` aggregation-pushdown policy is inert.

## Background

#841 landed Flight/Trino aggregation pushdown (`count`/`sum`/`min`/`max`/`avg` + simple `GROUP BY`) via the single-finalize-split design. The benefit eval (`docs/plans/2026-06-20-issue-841-aggregation-pushdown-benefit-eval.md`) measured that the data-reduction win is enormous for global aggregates and low/mid-cardinality `GROUP BY`, but **degrades to break-even (and slightly negative on bytes) as the number of distinct groups approaches the row count** — when nearly every row is its own group, the single finalize worker does all the merging with no payload reduction and loses intra-aggregation parallelism. Measured crossover: pushdown wins materially once `rows / groups >= ~4-10x`; it is a wash-to-loss above that.

#893 asked for a cardinality-based gate that declines pushdown in the high-cardinality case. #937 shipped the **connector-side operator gate** but not the statistics needed to drive it automatically:

- New catalog properties (`CqliteFlightConfig`):
  - `cqlite.aggregation-pushdown-group-by` = `automatic` | `always` | `never`
  - `cqlite.aggregation-pushdown-max-group-ratio` (default `0.5`, must be in `(0.0, 1.0]`; `DEFAULT_MAX_GROUP_RATIO`)
- `CqliteFlightMetadata.applyAggregation` consults a pure decision:
  `declineGroupByPushdown(GroupByPushdownPolicy policy, OptionalDouble estimatedGroupRatio, double maxGroupRatio)`
  — `NEVER` declines, `ALWAYS` pushes, `AUTOMATIC` declines **only** when an estimate is present and exceeds the ratio.
- Global aggregates (no `GROUP BY`) bypass the gate entirely (unconditional win).

## The gap

```java
// CqliteFlightMetadata.estimateGroupRatio(...) — current body
return java.util.OptionalDouble.empty();
```

`estimateGroupRatio()` returns empty because **no cardinality estimate is surfaced anywhere**:

[...]

PR #1351, opened 2026-07-01
fix(#1325): populate nb RowStatistics counts from authoritative STATS (reuse read_table_counts)

### LINKED ISSUE #1325: enhanced_statistics_parser stubs RowStatistics.{total_rows,live_rows,partition_count} to 0 for real nb SSTables
## Summary
For real Cassandra 5.0 `nb` SSTables, `cqlite-core`'s `enhanced_statistics_parser` hard-codes `RowStatistics.total_rows`, `live_rows`, and `partition_count` to `0` (`cqlite-core/src/parser/enhanced_statistics_parser/mod.rs:~139-146`). Only the synthetic/legacy parser fills them. So any consumer reading `RowStatistics` for these counts on a real file gets 0.

Discovered during #944 (Flight/Trino aggregation-pushdown stats): the gate needed authoritative per-table `partition_count`/`total_rows`, which `RowStatistics` did NOT provide for real files. #944 worked around it by adding `cqlite-core/src/parser/repair_metadata.rs::read_table_counts` (authoritative: `partition_count` = Σ EstimatedHistogram bucket counts; `total_rows` = STATS `totalRows` via the version-gated walk, `None` when not traversable — never guessed).

## Fix (routing: oracle — Statistics.db is ground truth, no OpenSpec)
Populate `RowStatistics.{total_rows,live_rows,partition_count}` for `nb` from the STATS component, ideally by delegating to / reusing the authoritative decode that `read_table_counts` now performs (single source of truth). No heuristics — `None`/absent where the value isn't authoritatively reachable, never a fabricated 0.

## Acceptance
- For real `nb` fixtures, `RowStatistics.partition_count` and `total_rows` match the authoritative STATS values (validate against `sensor_data`=10/2000, `simple_table`=1000/1000, `stock_prices`=3/200, as verified in #944).
- No guessed values; unreachable fields stay `None`/0-with-documented-meaning, not a silent wrong 0.
- No-heuristics compliant.

Discovered in #944 (non-blocking there — #944 surfaced its own authoritative decode).

pmcfadin:
Closes #1325.

## Bug (oracle-driven — Statistics.db is ground truth)
For real Cassandra 5.0 `nb` SSTables, `enhanced_statistics_parser` hard-coded `RowStatistics.{total_rows, live_rows, partition_count}` to `0`, so any consumer reading these counts on a real file got a silently-wrong 0.

## Fix
`parse_nb_format_statistics_data` now delegates to the #944 single source of truth `repair_metadata::read_table_counts` (raw Statistics.db + version gates already threaded through):
- `partition_count` = Σ `estimatedPartitionSize` histogram bucket counts.
- `total_rows` = STATS `totalRows` via the gated walk (`None` → documented 0, never guessed).
- `avg_rows_per_partition` only when both counts are authoritative and `partition_count > 0`.
- `live_rows` deliberately left `0` (documented): STATS carries no per-SSTable live-row count distinct from `total_rows`; fabricating one would violate the no-heuristics mandate (#28). Additive decode — an `Err` logs and leaves 0.

## Test (TDD)
`cqlite-core/tests/issue_1325_nb_rowstatistics_counts.rs` asserts, for real fixtures:
- `sensor_data` → partition_count=10, total_rows=2000
- `simple_table` (test_basic, nb) → 1000, 1000
- `stock_prices` → 3, 200
- `live_rows == 0`

All three fixtures present + asserted (none skipped); fails without the fix (0 vs 10), passes with it. Present-but-zero is a hard failure; skips only when Statistics.db is genuinely absent.

## Gate
`scripts/agent-gate.sh` → **RESULT: PASS** (all 17 components green, incl. python-bindings — no known-flaky failures this run).

### DIFF
--- a/cqlite-core/src/parser/enhanced_statistics_parser/mod.rs
+++ b/cqlite-core/src/parser/enhanced_statistics_parser/mod.rs
@@ -135,13 +135,56 @@ pub fn parse_nb_format_statistics_data(
                 regular_columns,
             ),
         )) => {
-            // Create minimal statistics with only timestamp data populated
+            // Populate the authoritative row/partition counts from the STATS
+            // component (issue #1325). We reuse the single source of truth added
+            // in #944 — `repair_metadata::read_table_counts` — rather than
+            // re-deriving the walk here:
+            //   * `partition_count` = Σ `estimatedPartitionSize` histogram bucket
+            //     counts (self-describing; always decodable).
+            //   * `total_rows`      = STATS `totalRows` via the version-gated walk,
+            //     `None` when not authoritatively traversable (e.g. a clustered
+            //     covered-Slice whose bound values are not modeled).
+            // No-heuristics mandate (#28): when a value is NOT authoritatively
+            // reachable we leave the documented 0 rather than fabricating a count.
+            //   * `total_rows`: `None` from `read_table_counts` → 0 (meaning
+            //     "not authoritatively reachable", never a guessed count).
+            //   * `live_rows`: STATS carries no per-SSTable live-row count that is
+            //     authoritatively distinguishable from `total_rows` here, so it is
+            //     left as 0 (documented "not authoritatively derivable") rather
+            //     than guessed. (Consumers that need an upper bound use
+            //     `total_rows`; see cqlite-flight `read_table_counts` doctrine.)
+            //   * `avg_rows_per_partition`: computed only when BOTH counts are
+            //     authoritative and `partition_count > 0`; otherwise left 0.0.
+            // `full_input` is the raw Statistics.db buffer (with TOC), exactly the
+            // input `read_table_counts` expects. This is strictly additive — an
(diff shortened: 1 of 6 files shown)


pmcfadin:
roborev triage:
- **Finding 2** (None-gates path → 0) — fixing: nb parser now defaults to authoritative nb gates on the `None` path (we're definitively in the nb parser), + test coverage for the None entry point.
- **Finding 3** (test can pass vacuously) — fixing: assert `asserted > 0` when a dataset root is configured.
- **Finding 1** (`live_rows` = 0) — out of scope for this P2 fix. Per #1325's acceptance criteria, unreachable fields stay 0-with-documented-meaning, and STATS has no authoritative per-SSTable live-row count. The broader question of representing 'unknown' vs measured-0 across `RowStatistics` is a public-API/field-contract change → filed as design follow-up #1352. Strengthening the doc comment here as the interim contract.

PR #1735, opened 2026-07-02
fix(#1446): emit Node row objects in SELECT order; intern column-name JsStrings once per result

### LINKED ISSUE #1446: Node row objects in hash order; column-name JsStrings re-interned every row
Parent: #1433

## Why this matters
When a Node user reads a row from `executeNative()`, `Object.keys(row)` comes back in **hash order**, not `SELECT` order — the same bug as Python (#1445), but on the Node side. Any code that assumes column order (CSV headers, `Object.values(row)` positional access) silently breaks, and it disagrees with both the CLI and (once #1445 lands) Python. Separately, every row re-interns each column name into a fresh V8 `JsString`, so a wide-table scan pays O(rows × columns) redundant string creations at the FFI boundary.

`bindings/node/src/value.rs:437-447` — `row_to_object` iterates a `HashMap` (nondeterministic order) and calls `set_named_property` with a freshly created string per cell:
```rust
pub fn row_to_object(
    env: &Env,
    values: &std::collections::HashMap<String, Value>,
) -> Result<JsObject> {
    let mut obj = env.create_object()?;
    for (col_name, value) in values {
        let js_value = value_to_napi(env, value)?;
        obj.set_named_property(col_name, js_value)?;
    }
    Ok(obj)
}
```
`bindings/node/src/database.rs:1271-1277` — the row payload is a `HashMap`, so order is already lost before `resolve()`:
```rust
pub struct QueryResultData {
    rows: Vec<std::collections::HashMap<String, cqlite_core::types::Value>>,
    execution_time_ms: u32,
    columns: Vec<cqlite_core::query::result::ColumnInfo>,
    rows_affected: u32,
}
```
`bindings/node/src/database.rs:1336-1344` — `resolve()` builds each row via `row_to_object` (hash order):
```rust
fn resolve(&mut self, env: napi::Env, output: Self::Output) -> napi::Result<Self::JsValue> {
    let mut result_obj = env.create_object()?;
    let mut rows_arr = env.create_array_with_length(output.rows.len())?;
    for (i, row_values) in output.rows.iter().enumerate() {
        let row_obj = crate::value::row_to_object(&env, row_values)?;
        rows_arr.set_element(i as u32, row_obj)?;
    }
```
`bindings/node/src/database.rs:1355-1371` — the authoritative SELECT column order (`output.columns`, a `Vec<ColumnInfo>`) is already present and iterated separately to build the `columns` array, but is **not** used to order the row properties.
Second caller: `bindings/node/src/streaming.rs:187` also calls `row_to_object(&env, &values)` — its signature change must be handled there too.

1. Thread the ordered column names into row construction. Recommended: change `row_to_object` to take the ordered names and (optionally) precomputed key `JsString`s:
   ```rust
   pub fn row_to_object(
       env: &Env,
       column_names: &[String],       // SELECT order (from ColumnInfo)
       values: &std::collections::HashMap<String, Value>,
   ) -> Result<JsObject> {
       let mut obj = env.create_object()?;
       for name in column_names {
           if let Some(value) = values.get(name) {
               let js_value = value_to_napi(env, value)?;
               obj.set_named_property(name, js_value)?;
           }
       }
       Ok(obj)
   }
   ```
   Invariant to preserve if you refactor differently: property insertion order MUST equal `columns.map(c => c.name)`; V8 preserves string-key insertion order, so ordered `set_named_property` calls are sufficient.
2. Precompute the column-name `JsString`s **once per result**, not per row. In `resolve()` (`database.rs:1336`), before the row loop, build `let col_names: Vec<String> = output.columns.iter().map(|c| c.name.clone()).collect();` (or, to go further, create `napi::JsString` handles once and reuse — `set_named_property` accepts `&str`, so the minimum bar is: derive the ordered `&[String]` once and pass a borrow into every `row_to_object` call). Do not rebuild the name list inside the per-row loop.
3. Update both call sites:
   - `database.rs:1342` → `row_to_object(&env, &col_names, row_values)?`
   - `streaming.rs:187` → pass the streaming result's ordered column names (source them from the same `ColumnInfo` metadata the streaming task already carries; if streaming does not currently retain columns, add the ordered names to its output struct alongside the existing rows).
4. Leave the `columns` array construction (`database.rs:1355-1371`) as-is; it is already correct and is now the single source of order.

## Do NOT
- Do NOT change the `compute()` clone at `database.rs:1329` (`result.rows.iter().map(|r| r.values.clone())`) — that deep-clone→move is **#1447** (W3). Touch only `resolve()` and `row_to_object`.
- Do NOT add Set/Map constructor caching — that is **#1448** (W4). Note: #1448 also modifies `value_to_napi`/`row_to_object`'s call chain to thread a conversion context; if #1448 lands first, thread your ordered-columns parameter through the **same** context struct rather than adding a second parameter. If this issue lands first, keep the signature minimal and expect #1448 to extend it. Coordinate to avoid a merge conflict on `row_to_object`'s signature.
- Do NOT touch Python (`#1445` owns the identical Python bug).

## Tests you must add
`bindings/node/__test__/result.test.js` (Jest; run from `bindings/node`).
Add:
```js
test('Object.keys(row) matches SELECT column order', async () => {
  const db = await Database.open(DATA_DIR, { schema: SCHEMA });
  const result = await db.executeNative('SELECT * FROM test_basic.simple_table LIMIT 5');
  expect(result.rowCount).toBeGreaterThan(0); // fixture present but empty -> FAIL, never skip
  const expected = result.columns.map(c => c.name);
  for (const row of result.rows) {
    expect(Object.keys(row)).toEqual(expected);
  }
  await db.close();
});
```
Add a second case with an explicit reordered projection (`SELECT b, a FROM ...`) asserting `Object.keys(row)` equals `['b','a']`.
Dataset rule: `expect(result.rowCount).toBeGreaterThan(0)` FAILS LOUDLY on present-but-empty datasets — do not guard-skip.
Run: `cd bindings/node && npm run build && npm test -- result.test.js`

## Definition of done
- [ ] `Object.keys(row)` deep-equals `columns.map(c => c.name)` for every row, including a reordered projection
- [ ] Column names derived once per result, not per row; both `database.rs` and `streaming.rs` call sites updated
- [ ] File-size ratchet respected — `bindings/node/src/database.rs` is already 1658L (over threshold per the audit, campsite #1116). Do not grow it; if your change adds net lines, split per the campsite rule (see #1116) or, if a split is out of scope, re-run with `CQLITE_ALLOW_FILE_GROWTH=1` and leave a note linking #1116

pmcfadin:
Closes #1446.

## Problem
`executeNative()`/`executeStreaming()` built Node row objects by iterating a `HashMap`, so `Object.keys(row)` came back in hash order (disagreeing with the `columns` array, the CLI, and Python), and every row re-interned each column-name `JsString` at the FFI boundary (O(rows × columns) redundant string creation).

## Fix (bindings/node only — no cqlite-core public-surface change)
- `value.rs`: new `intern_column_keys()` builds each SELECT-order column-name `JsString` **once per result set**; `row_to_object()` takes those pre-interned key handles and emits properties in authoritative column order (V8 preserves string-key insertion order). Includes a deterministic name-sorted fallback so no cell is ever dropped when column metadata is empty (roborev job 2721).
- `database.rs`: precompute the ordered key handles once before the row loop (not per row).
- `streaming.rs`: thread ordered column names through `StreamingResult`/`NextTask` so streamed rows are emitted in SELECT order too.

## Tests (wiring evidence, real Jest against public API)
`__test__/result.test.js`: `Object.keys(row)` deep-equals `columns.map(c=>c.name)` for `SELECT *`, an explicit reordered projection (`SELECT name, id, age`), and the streaming path. Fail-loud on present-but-empty datasets.

## Notes
- Oracle-driven bug → GitHub issue + pinned parity test; no OpenSpec change.
- `database.rs` is over the campsite file-size threshold (#1116); split out of scope for this focused fix → `CQLITE_ALLOW_FILE_GROWTH=1` ack for +5 lines.

🤖 flow-lead worker

### DIFF
--- a/bindings/node/__test__/result.test.js
+++ b/bindings/node/__test__/result.test.js
@@ -282,11 +282,9 @@ describe('Row property order matches SELECT order (Issue #1446)', () => {
     }
   });
 
-  test('executeNative: every selected column is present on every row (null-filled, never undefined)', async () => {
-    // Locks the null-fill branch: a selected column absent from a row's values
-    // must still appear as a key (emitted as null), so row shape is stable and
-    // Object.keys never diverges from the projection. Missing cells are null,
-    // never undefined/omitted.
+  test('executeNative: every selected column is present on every row (never undefined)', async () => {
+    // For normal projections core populates every selected column, so the row
+    // shape is stable: Object.keys equals the projection and no cell is omitted.
     const cols = ['id', 'name', 'age', 'salary', 'active'];
     const result = await db.executeNative(
       `SELECT ${cols.join(', ')} FROM test_basic.simple_table LIMIT 10`
@@ -295,8 +293,28 @@ describe('Row property order matches SELECT order (Issue #1446)', () => {
     for (const row of result.rows) {
       expect(Object.keys(row)).toEqual(cols);
       for (const k of cols) {
-        expect(row[k]).not.toBeUndefined(); // present as a value or explicit null
+        expect(row[k]).not.toBeUndefined();
       }
+    }
+  });
+
+  test('executeNative: aggregate projection emits no phantom metadata column', async () => {
+    // Regression (#1446 roborev job 2736): aggregate result metadata uses a
+    // fallback name (col_0) while the row value is keyed by the expression name
+    // (Count(*)). A metadata column absent from values must be skipped, not
+    // null-filled — otherwise the row gains a phantom `col_0: null`.
+    const result = await db.executeNative(
+      'SELECT COUNT(*) FROM test_basic.simple_table'
+    );
+    expect(result.rowCount).toBe(1);
+    const row = result.rows[0];
+    const keys = Object.keys(row);
+    // The real aggregate cell is present with a non-null value...
+    expect(keys.length).toBe(1);
+    expect(row[keys[0]]).not.toBeNull();
+    // ...and no phantom null-filled metadata column leaks through.
+    for (const [k, v] of Object.entries(row)) {
+      expect(v === null).toBe(false);
     }
   });
 
--- a/bindings/node/__test__/result.test.js
+++ b/bindings/node/__test__/result.test.js
@@ -241,3 +241,57 @@ describe('QueryResult and ColumnInfo Tests (Issue #303)', () => {
     });
   });
 });
+
+describe('Row property order matches SELECT order (Issue #1446)', () => {
+  let db = null;
+
+  beforeAll(async () => {
+    skipIfNoDatasets();
+    db = await Database.open(global.testPaths.SSTABLES_DIR, {
+      schema: global.testPaths.SCHEMA_BASIC_TYPES,
+    });
+  });
+
+  afterAll(async () => {
+    if (db) {
+      await db.close();
+      db = null;
+    }
+  });
+
+  test('executeNative: Object.keys(row) matches SELECT column order (SELECT *)', async () => {
+    const result = await db.executeNative(
+      'SELECT * FROM test_basic.simple_table LIMIT 5'
+    );
+    // Dataset rule: present-but-empty fixtures FAIL loudly, never skip.
+    expect(result.rowCount).toBeGreaterThan(0);
+    const expected = result.columns.map((c) => c.name);
+    for (const row of result.rows) {
+      expect(Object.keys(row)).toEqual(expected);
+    }
+  });
+
+  test('executeNative: Object.keys(row) honors an explicit reordered projection', async () => {
+    const result = await db.executeNative(
+      'SELECT name, id, age FROM test_basic.simple_table LIMIT 5'
+    );
+    expect(result.rowCount).toBeGreaterThan(0);
+    expect(result.columns.map((c) => c.name)).toEqual(['name', 'id', 'age']);
+    for (const row of result.rows) {
+      expect(Object.keys(row)).toEqual(['name', 'id', 'age']);
+    }
+  });
+
+  test('executeStreaming: Object.keys(row) matches SELECT column order', async () => {
+    const stream = db.executeStreaming(
+      'SELECT name, id, age FROM test_basic.simple_table LIMIT 5'
+    );
+    let seen = 0;
+    for await (const row of stream) {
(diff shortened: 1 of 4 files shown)


pmcfadin:
## roborev convergence note (job history 2721 → 2741)

roborev (codex) ran 5 rounds. Rounds 1-4 findings were fixed in commits:
- **2721**: empty column metadata → `{}` (data loss) → added deterministic extras fallback.
- **2726**: sparse selected column skipped → null-fill (later reconsidered, see below).
- **2728**: length-guard could drop an unmapped extra → detect extras by membership.
- **2731**: per-row `HashSet` rebuild undercut the perf goal → `ColumnKeys` struct built once per result; added null-fill test.
- **2736**: null-fill emitted a **phantom `col_0: null` for aggregates** (verified: `SELECT COUNT(*)` has metadata `[col_0]` but a value keyed `Count(*)`) → removed null-fill, skip absent metadata columns; added aggregate regression test.

**2741 (closed as resolved — refuted):** it reversed 2736, asking to restore null-fill, on the premise that skip-absent 'regresses null shape stability' and 'diverges from execute() which null-fills.' Both are false:
- The **original** `executeNative`/`row_to_object` iterated the values HashMap and emitted **only present cells** (hash order). Skip-absent emits the same key set, now ordered — it regresses nothing; null-fill was never the baseline.
- The JSON `execute()` path (`database.rs:747-759`) also emits `row.values` present-cells-only. Skip-absent is **consistent** with it; null-fill would diverge and reintroduce the aggregate phantom.

Final behavior: SELECT order preserved for normal queries (all cells present → `Object.keys(row)` == `columns.map(c=>c.name)`, verified by the prescribed #1446 tests), correct aggregate output (`{"Count(*)": ...}`, no phantom), no data loss.

**Follow-up for owner (out of scope, cross-cutting):** aggregate result metadata uses a synthetic fallback name (`col_0`) instead of the expression alias, so `metadata.columns` disagrees with row value keys for aggregates in BOTH `execute()` and `executeNative()`. Fixing that is a cqlite-core result-construction change; flagging rather than fixing here.

Issue #1610, opened 2026-07-02
Epic Q — Merge/compaction: allocation discipline, deterministic STCS, budget honesty


pmcfadin:
Epic Q — Merge/compaction: allocation, evenness, budget honesty (P1)

Source: `docs/reports/write-path-performance-audit-2026-07-01.md` (Executive summary points 3/4/6; Epic Q table Q1–Q8; Priority matrix ranks 4/6/7/11).

## Theme

The write-engine merge (the k-way merge used by flush and STCS compaction) is unit-correct and byte-parity-verified, but the audit found four systemic problems in this one hot path:

1. **Allocation tax per merge entry.** Every `MergeEntry` is deep-cloned **twice** on pure-waste paths in the k-way merge core (`merge/mod.rs:2141,2177`) even though `advance()` already returns an owned entry and `run_index` is `Copy`. Reconcile adds its own micro-alloc churn (double hash + double key-clone per cell, an avoidable `.cloned()` of the survivor set).
2. **Data-dependent memory cliff.** `step()` materializes one whole partition **4–5× simultaneously** (Vec → BTreeMap → HashMap → Vec → Vec, `merge/mod.rs:2119-2301`). Per-source streaming (#827) otherwise bounds compaction to 128 MiB, but one fat partition defeats that: peak scales with the widest partition, not with `k`.
3. **Nondeterministic selection.** STCS bucket selection iterates a `HashMap` (`merge_policy.rs:162,171,261`), so **which tier compacts is random per run**. Same inputs, different output file sets across runs — bad for reproducibility and for the byte-parity harness's determinism assumptions.
4. **Dishonest budget.** `--budget-ms` is only checked **between** partitions (`maintenance.rs:307`); a giant partition overshoots unboundedly. Dropped-column tables additionally run a **full unbudgeted merge pre-pass** (`compaction.rs:149`) before the partition-loop budget clock does anything. The documented contract does not match the code.

Two lower-priority items ride along: a linear range-shadow scan over sorted+disjoint ranges (`merge/mod.rs:2914-2934`), and a fresh `Runtime::new()` + scoped thread constructed on **every** `block_on_async` flush/maintenance call (`merge/mod.rs:367,377`).

## Blanket rule — #921 byte-parity gate (applies to EVERY child)

[...]

PR #1756, opened 2026-07-03
fix(#1666): deterministic STCS bucket selection

### LINKED ISSUE #1666: Q3 — Deterministic STCS bucket selection
Part of epic #1610 (Epic Q — merge/compaction allocation + budget honesty). Source: docs/reports/write-path-performance-audit-2026-07-01.md finding Q3.

## Problem

STCS bucket selection is **nondeterministic**: which tier of SSTables gets compacted is random per run because the algorithm iterates and returns a `HashMap` whose iteration order is unspecified. Same inputs → different output file sets across runs. This breaks reproducibility and the byte-parity harness's determinism assumptions.

`cqlite-core/src/storage/write_engine/merge_policy.rs`:

- **`merge_policy.rs:162`** — buckets stored in a `HashMap`:
  ```rust
  let mut buckets: HashMap<u64, Vec<SSTableMetadata>> = HashMap::new();
  ```
- **`merge_policy.rs:171`** — fit-search iterates the HashMap (first-fit depends on iteration order):
  ```rust
  for (&avg_size, _bucket) in buckets.iter() {
      // … if within_ratio || both_small { … break; }
  }
  ```
- **`merge_policy.rs:207`** — buckets returned in HashMap order:
  ```rust
  buckets.into_values().collect()
  ```
- **`merge_policy.rs:261`** — `select_merge_internal` picks the **first** bucket with `>= min_threshold` from that unordered Vec:
  ```rust
  for bucket in buckets {
      if bucket.len() >= self.min_threshold {
          // … selected …
          return Ok(selected);
      }
  }
  ```

So both the bucket assignment (which existing bucket an SSTable joins) and the final tier pick are order-dependent on `HashMap` iteration. Cassandra's STCS deterministically prefers the smallest eligible tier.

## Fix (effort: S)

Make selection deterministic and Cassandra-like: **smallest-eligible-tier-first**, with a deterministic fit-search. VERIFY `SSTableMetadata` has a stable `data_size: u64` and (for a total tie-break) `data_path: PathBuf` — grep the struct.

1. **`group_into_buckets` fit-search (line 171):** iterate candidate buckets in a **deterministic** order (e.g. by ascending `avg_size`, then by a stable secondary key) rather than raw `HashMap::iter()`. The `sorted` input Vec is already size-ascending (`sorted.sort_by_key(|s| s.data_size)` at line 159), so drive the fit decision from a structure with defined order. Options: keep the `HashMap` for O(1) average-key updates but collect its keys into a sorted Vec before the fit loop, OR replace the map with a `Vec<(avg, bucket)>` kept sorted. Pick the minimal change that yields **identical bucket membership** for a given input regardless of run.

2. **Return order (line 207) + tier pick (line 261):** produce buckets as a Vec **sorted by average/representative size ascending**, and pick the smallest bucket whose `len() >= min_threshold`. Sketch for the pick:
   ```rust
   let mut buckets: Vec<Vec<SSTableMetadata>> = self.group_into_buckets(&sstables);
   // deterministic order: smallest tier first (representative = min data_size in bucket)
   buckets.sort_by_key(|b| b.iter().map(|s| s.data_size).min().unwrap_or(0));
   for bucket in buckets {
       if bucket.len() >= self.min_threshold {
           let selected: Vec<PathBuf> = bucket
               .into_iter()
               .take(self.max_threshold)
               .map(|s| s.data_path)
               .collect();
           return Ok(selected);
       }
   }
   Ok(Vec::new())
   ```
   If two buckets have the same min size, add a total secondary tie-break (e.g. by sorted `data_path`) so the pick is fully deterministic. The `.take(self.max_threshold)` should also operate on a size-sorted bucket so the selected subset is stable — sort the bucket before `take`.

Note: `.unwrap_or(0)` in the `sort_by_key` closure is fine (no panic), but prefer `.min().unwrap_or(u64::MAX)` semantics only if it matches "smallest first" intent — an empty bucket never reaches here. Do NOT introduce a real `unwrap()`.

## Do NOT

- Do NOT change the compaction OUTPUT bytes for any single-eligible-tier input — #921 byte-parity harness must stay green. This changes *which* tier is chosen when multiple are eligible; for the harness fixtures the selection must remain byte-identical (the harness inputs typically have one eligible tier — VERIFY that the new deterministic pick equals the old pick for every harness fixture).
- Do NOT change `min_threshold` / `max_threshold` / `bucket_low` / `bucket_high` semantics or the `both_small` grouping rule.
- No `unwrap()`/`expect()` in library code; `-D warnings` clean.

## Acceptance criteria

- [ ] Bucket grouping and tier selection are deterministic (no `HashMap`-iteration-order dependence at lines 171, 207, 261).
- [ ] Smallest eligible tier is chosen first, with a total tie-break.
- [ ] Determinism TDD test passes (red-on-main).
- [ ] #921 harness green; JSONL goldens unchanged.
- [ ] `scripts/agent-gate.sh` PASS.

Construct **two** eligible buckets (two size tiers, each with `>= min_threshold` SSTable metadata entries — use temp files of controlled sizes, or a `select_merge_internal`-level unit test with fabricated `SSTableMetadata`). Call the selection **many times** (e.g. 100 iterations) and assert **the same (smaller) tier is picked every time**:

```rust
let picks: HashSet<Vec<PathBuf>> = (0..100).map(|_| policy.select_merge_internal(&candidates).unwrap()).collect();
assert_eq!(picks.len(), 1, "selection must be deterministic");
assert_eq!(picks.into_iter().next().unwrap(), expected_smaller_tier);
```

On current main the pick varies across runs (HashMap iteration order), so either `picks.len() > 1` over enough iterations OR the picked tier is not the smaller one — the test FAILS on main. (Note: `HashMap` uses a randomized hasher per process, so across a single process runs may or may not vary; make the test robust by also asserting the pick **equals the smaller tier**, which is wrong-or-flaky on main and correct after.)

pmcfadin:
Closes #1666 (Q3, epic #1610).

## What
STCS bucket selection was nondeterministic: buckets lived in a `HashMap<u64, Vec<SSTableMetadata>>` whose iteration order is unspecified, so bucket assignment, returned bucket order, and first-eligible selection all varied per run — same inputs → different compacted file sets, breaking reproducibility and the byte-parity harness's determinism assumptions. All in `merge_policy.rs`:
- **Fit-search:** iterate a snapshot of HashMap keys sorted ascending (`avg_sizes.sort_unstable()`) — smallest matching tier first. HashMap retained for O(1) avg-key updates; only the decision order is now total.
- **Input sort:** `group_into_buckets` sorts input by `(data_size, data_path)` (was `data_size` only) → order independent of caller candidate order for equal sizes.
- **Return order:** buckets sorted smallest-tier-first by representative (min member) size with `data_path` tie-break; members sorted by `(size, path)` so a downstream `take(max_threshold)` selects a stable subset. Empty-bucket handled without panic (`u64::MAX` sentinel, no `unwrap`).
- **Selection:** unchanged logic (first eligible) now deterministically = smallest eligible tier. No policy-parameter / ratio / `both_small` semantics touched.

## Tests (TDD — red on main)
`test_stcs_selection_is_deterministic_smallest_tier` — 100 iterations over two eligible tiers (4×100MB + 4×500MB); asserts one distinct pick (the 4×100MB tier) and ascending-representative return order. Verified RED on main (HashMap order → "got 2 distinct results"). Existing `stcs_bucket_selection_parity` unaffected.

## Gate
```
RESULT: PASS  (compaction-byte-parity PASS — selection-order change does not alter goldens)
```
(HEAD-stamped re-gate against the rebased commit posted as a follow-up comment.)

### DIFF
--- a/cqlite-core/src/storage/write_engine/merge_policy.rs
+++ b/cqlite-core/src/storage/write_engine/merge_policy.rs
@@ -154,21 +154,37 @@ impl STCSPolicy {
             return Vec::new();
         }
 
-        // Sort by size (ascending)
+        // Sort by size (ascending), tie-breaking on path so equal-sized inputs
+        // are processed in a fully deterministic order (independent of the
+        // caller's candidate ordering).
         let mut sorted = sstables.to_vec();
-        sorted.sort_by_key(|s| s.data_size);
+        sorted.sort_by(|a, b| {
+            a.data_size
+                .cmp(&b.data_size)
+                .then_with(|| a.data_path.cmp(&b.data_path))
+        });
 
-        // Map of average size -> bucket
+        // Map of average size -> bucket. The HashMap is retained for O(1)
+        // average-key updates, but the fit-search below never iterates it
+        // directly: iteration order of `HashMap` is unspecified, which would
+        // make bucket membership nondeterministic across runs (issue #1666).
         let mut buckets: HashMap<u64, Vec<SSTableMetadata>> = HashMap::new();
 
         for sstable in sorted {
             let size = sstable.data_size;
 
-            // Look for a bucket containing similar-sized files
+            // Look for a bucket containing similar-sized files. We iterate the
+            // existing bucket average-sizes in ascending order (a deterministic
+            // total order over the unique u64 keys) so that the first-fit
+            // decision is reproducible and prefers the smallest matching tier,
+            // matching Cassandra's smallest-tier-first intent.
+            let mut avg_sizes: Vec<u64> = buckets.keys().copied().collect();
+            avg_sizes.sort_unstable();
+
             let mut found_bucket = false;
             let mut old_average = 0u64;
 
-            for (&avg_size, _bucket) in buckets.iter() {
+            for &avg_size in &avg_sizes {
                 // Check if this SSTable fits in the bucket:
                 // 1. Size is within [bucket_low, bucket_high] ratio of bucket average
                 // 2. OR both this SSTable and bucket average are below min_sstable_size
@@ -203,8 +219,32 @@ impl STCSPolicy {
             }
         }
 
-        // Convert to Vec of buckets
-        buckets.into_values().collect()
+        // Convert to a deterministically ordered Vec of buckets.
+        let mut result: Vec<Vec<SSTableMetadata>> = buckets.into_values().collect();
+
+        // Sort each bucket's members by (size, path) so that any downstream
+        // `take(max_threshold)` selects a stable subset.
+        for bucket in &mut result {
+            bucket.sort_by(|a, b| {
+                a.data_size
+                    .cmp(&b.data_size)
+                    .then_with(|| a.data_path.cmp(&b.data_path))
+            });
+        }
+
+        // Order buckets smallest-tier-first: representative size is the bucket's
+        // minimum member size (its first element after the sort above), with the
+        // corresponding path as a total tie-break. Empty buckets never occur, but
+        // are ordered last defensively without panicking.
+        result.sort_by(|a, b| {
+            let key = |b: &[SSTableMetadata]| match b.first() {
+                Some(s) => (s.data_size, Some(s.data_path.clone())),
+                None => (u64::MAX, None),
+            };
+            key(a).cmp(&key(b))
+        });
+
+        result
     }
 }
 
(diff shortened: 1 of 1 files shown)


pmcfadin:
✅ **Armed merge-on-green.** Quality bar met:
- **agent-gate PASS** stamped against rebased HEAD \`85044a1a\` (clean tree, all 19 components green — **compaction-byte-parity PASS**, so the deterministic selection-order change did NOT alter goldens).
- **C intent audit: PASS** — all three HashMap-order dependencies (assignment/return/selection) totalized; STCS semantics (ratio/thresholds/`both_small`/eligibility) byte-identical to main; determinism test independently reproduced RED on main ("got 2 distinct results"); no `unwrap`/`expect` in lib code.
- **roborev (codex): clean** — no issues found.

Ready for the manager poller to land on green.

Issue #1752, opened 2026-07-03
CI main-red: 'All Compression Build & Test' test build fails to compile (E0433 writer/query feature-gating)


pmcfadin:
## Summary
The **"All Compression Build & Test"** CI job (in the *Minimal Features CI* workflow) fails to COMPILE its test build on `main` — pre-existing main-red as of `a83d1f3d` (2026-07-02). Test modules reference feature-gated modules that are not enabled in this lane's feature combination.

## Failure (verified 2026-07-02/03)
Step "Test with all compression support (lib only)" fails with `E0433`:
```
error[E0433]: cannot find `writer` in `sstable`
   --> cqlite-core/src/storage/sstable/reader/crc.rs:269:34
error[E0433]: cannot find `query` in `crate`
   --> cqlite-core/src/storage/sstable/reader/parsing/block_entries.rs:1115:26
   --> cqlite-core/src/storage/sstable/reader/parsing/block_entries.rs:1177:20
   --> cqlite-core/src/storage/sstable/reader/parsing/block_entries.rs:1229:26
```
- Build step (non-test) PASSES; only the **test** build fails.
- Confirmed on `main` run 28627393743 (job 84896811503, `failure`) AND independently on an orthogonal branch that changes zero Rust — so it is main, not any one PR.

## Root cause (hypothesis)
`crc.rs` and `block_entries.rs` carry `#[cfg(test)]` test modules that reference `sstable::writer` (behind write-support) and `crate::query` (behind `state_machine`). The "All Compression Build & Test (lib only)" lane builds cqlite-core with a feature set (all-compression) that does NOT enable `state_machine`/`write-support`, so those paths are absent and the test module fails to compile.

## Affected files
- `cqlite-core/src/storage/sstable/reader/crc.rs:269`
- `cqlite-core/src/storage/sstable/reader/parsing/block_entries.rs:1115,1177,1229`
- The lane: `.github/workflows/*` "All Compression Build & Test" (Minimal Features CI).

## Routing
Oracle/mechanical (compile error, deterministic fix). No OpenSpec.

## Acceptance criteria
- `cargo test --package cqlite-core --no-default-features --features all-compression --lib` COMPILES and runs clean (reproduce the lane's exact feature set — confirm the real invoked flags in the workflow).
- The offending test references are correctly `#[cfg(feature = "...")]`-gated (or moved) so the test module compiles under the all-compression-only feature set.
- The "All Compression Build & Test" lane goes green on main.
[...]

PR #1758, opened 2026-07-03
fix(#1535): thread real WRITETIME/TTL per-cell metadata under --features tombstones

### LINKED ISSUE #1535: WRITETIME()/TTL() return null under `--features tombstones` (scan_with_cell_metadata returns empty)
## Summary
Under the `tombstones` feature, `SSTableManager::scan_with_cell_metadata` returns EMPTY per-cell metadata by design (`cqlite-core/src/storage/sstable/mod.rs` has an explicit fallback comment: "WRITETIME/TTL will still return null when tombstones are enabled"). So `WRITETIME()`/`TTL()` in SELECT yield null when the `tombstones` feature is active, even for live cells that carry a timestamp.

## Evidence
`issue_885_cross_generation_metadata_merge` passes in the default / `write-support` build (what the agent-gate uses) but fails when run with `--features tombstones`. This predates #1334 — surfaced while implementing PR #1533.

## Why it matters
WRITETIME()/TTL() are advertised v0.12 features; silently returning null under a supported feature flag is a correctness gap and a parity hazard.

## Routing
Oracle-driven (Cassandra semantics: WRITETIME/TTL come from authoritative cell timestamps; no heuristics). Fix should thread real per-cell metadata through the tombstone-enabled scan path.

## Acceptance criteria
- Under `--features tombstones`, a live cell's WRITETIME()/TTL() returns its real timestamp/TTL (not null), matching the non-tombstones build.
- A pinned test asserts WRITETIME()/TTL() parity across default and `tombstones` builds for a live cell.

Found-by: #1334 (PR #1533) implementer.

pmcfadin:
Closes #1535

## Problem
Under `--features tombstones`, `WRITETIME(col)` / `TTL(col)` in SELECT returned **null** even for live cells carrying a real authoritative timestamp/TTL — while the default / `write-support` build returned the real values. `SSTableManager::scan_with_cell_metadata` had a `#[cfg(feature = "tombstones")]` stub that returned an empty per-cell metadata map by design. WRITETIME()/TTL() are advertised v0.12 features, so this was a correctness + parity gap.

## Root cause
- The `tombstones` build of `scan_with_cell_metadata` (`cqlite-core/src/storage/sstable/mod.rs`) delegated to plain `scan` and returned `HashMap::new()` for every row.
- The cross-generation reconciliation helper `merge_generations_for_read_with_metadata` was over-gated `all(not(tombstones), write-support)` even though its body only needs `KWayMerger`.

## Fix
- Delete the empty-metadata `tombstones` stub; the single real implementation now serves both builds. The reader-level `SSTableReader::scan_with_cell_metadata` (not feature-gated) surfaces authoritative per-cell timestamp/TTL, now threaded through the tombstones path.
- Re-gate `merge_generations_for_read_with_metadata` to `#[cfg(feature = "write-support")]` so cross-generation LWW WRITETIME/TTL merge also works under `tombstones + write-support`.
- **No-heuristics honored:** all values come from authoritative per-cell SSTable timestamps/TTL; nothing inferred. Non-tombstones behavior is byte-identical.

Call chain now threading real metadata under tombstones: query executor → `Storage::scan_with_cell_metadata` → `SSTableManager::scan_with_cell_metadata` (unified) → `SSTableReader::scan_with_cell_metadata` / `merge_generations_for_read_with_metadata`. Also fixes the `WHERE pk = ?` path (`scan_partition_with_cell_metadata`).

## Tests (wiring-evidence)
- **New pinned parity test** `cqlite-core/tests/issue_1535_writetime_ttl_tombstones.rs::live_cell_writetime_and_ttl_resolve_under_tombstones` — writes a table with a known write-timestamp + per-cell TTL, reopens via `SSTableManager`, asserts `WRITETIME`/`TTL` equal the written values (not null) and the no-TTL cell's expiration is `None`. FAILED pre-fix, PASSES post-fix under `--features tombstones` **and** in the default build (parity).
- `issue_885_cross_generation_metadata_merge` now PASSES under `--features tombstones` (previously panicked on empty metadata).

## Quality bar
- `scripts/agent-gate.sh` **PASS** — all 20 components (commit 5f0090bd, clean tree); `tombstones-scan` PASS.
- Workspace clippy `--all-targets --all-features -D warnings` clean; no `unwrap()`/`expect()` added.
- roborev **clean** (codex, `--base origin/main`, first pass — no issues).

## Known limitation (out of scope, pre-existing)
Under a `tombstones`-only build **without** `write-support`, multi-generation WRITETIME/TTL falls back to per-reader concatenation (no `KWayMerger`) — identical to plain `scan` in that config, unchanged here. Single-generation metadata works correctly there.

### DIFF
--- a/cqlite-core/src/storage/sstable/mod.rs
+++ b/cqlite-core/src/storage/sstable/mod.rs
@@ -69,7 +69,6 @@ use tokio::sync::RwLock;
 #[cfg(feature = "tombstones")]
 use self::tombstone_merger::{EntryMetadata, GenerationValue, TombstoneMerger};
 use crate::platform::Platform;
-#[cfg(not(feature = "tombstones"))]
 use crate::types::CellWriteMetadata;
 use crate::{types::TableId, Config, Result, RowKey, ScanRow};
 // `RowCells`/`Value` are only referenced by the write-support merge read path
@@ -1557,7 +1556,12 @@ impl SSTableManager {
     /// matching the per-reader scan order (range then limit). With `None`/`None`
     /// bounds the output is byte-for-byte the full reconciled set, unchanged from
     /// before. This stays definitionally in lockstep with the plain helper.
-    #[cfg(all(not(feature = "tombstones"), feature = "write-support"))]
+    ///
+    /// Available whenever `write-support` is on — the `tombstones` build reuses the
+    /// identical `KWayMerger` reconciliation so `WRITETIME(col)` / `TTL(col)` resolve
+    /// from the authoritative winning cell there too (Issue #1535). Nothing in this
+    /// body depends on the tombstones GC path; it only needs the merger.
+    #[cfg(feature = "write-support")]
     async fn merge_generations_for_read_with_metadata(
         &self,
         reader_list: &[Arc<reader::SSTableReader>],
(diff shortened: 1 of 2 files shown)


pmcfadin:
🔧 flow-lead: terminal bar met — `agent-gate.sh` **PASS** (all 20, commit 5f0090bd), roborev **clean** (codex, first pass). Oracle-driven (no OpenSpec/C). Feature-combo compile check across all 5 combos my change touches (incl. the key risk: `tombstones` WITHOUT `write-support`) confirms **no new break**. The one red PR lane — *All Compression Build & Test* — is the **pre-existing orthogonal main-red #1752** (identical E0433 in untouched `crc.rs`/`block_entries.rs` test modules), non-required (`contexts=[]`). Merging on green.

PR #1765, opened 2026-07-03
fix(#1445): Python Row returns columns in SELECT order (no per-row name re-clone)

### LINKED ISSUE #1445: Python Row returns columns in hash order, not SELECT order; re-clones column names per row
Parent: #1433

## Why this matters
When a Python user iterates a result row (`row.keys()`, `row.items()`, `dict(row)`, `row.to_dict()`), the columns come back in **hash order**, not the order they were written in the `SELECT`. That silently breaks any code that assumes positional column order (CSV export, zipping against a header, `assert list(row) == expected`) and diverges from the CLI's ordered JSON. On top of that, every row re-clones every column-name `String`, so a wide-table scan pays O(rows × columns) redundant string allocations.

`bindings/python/src/result.rs:237-240` — `Row` stores an unordered map:
```rust
#[pyclass(module = "cqlite")]
pub struct Row {
    /// Column values as Python objects, keyed by column name
    values: HashMap<String, PyObject>,
}
```
`bindings/python/src/result.rs:260-262` — `keys()` returns hash order:
```rust
fn keys(&self) -> Vec<String> {
    self.values.keys().cloned().collect()
}
```
`bindings/python/src/result.rs:270-284` — `items()`/`to_dict()` iterate the map (hash order), and `to_dict()` clones each key:
```rust
fn items(&self, py: Python<'_>) -> Vec<(String, PyObject)> {
    self.values
        .iter()
        .map(|(k, v)| (k.clone(), v.clone_ref(py)))
        .collect()
}

fn to_dict(&self, py: Python<'_>) -> PyResult<PyObject> {
    let dict = PyDict::new(py);
    for (key, value) in &self.values {
        dict.set_item(key, value.clone_ref(py))?;
    }
    Ok(dict.into_any().unbind())
}
```
`bindings/python/src/result.rs:318-330` — `Row::from_core` builds a fresh `HashMap` per row and `insert`s a cloned key per column:
```rust
impl Row {
    pub(crate) fn from_core(
        py: Python<'_>,
        row: &cqlite_core::query::result::QueryRow,
    ) -> PyResult<Self> {
        let mut values = HashMap::with_capacity(row.values.len());
        for (key, value) in &row.values {
            let py_value = value_to_py(py, value)?;
            values.insert(key.clone(), py_value);
        }
        Ok(Self { values })
    }
}
```
`bindings/python/src/result.rs:162-190` — `QueryResult::from_core` already has the authoritative SELECT column order in `result.metadata.columns` (used to build the `columns` field at :167-172) and drives `Row::from_core` per row at :175-182. The core `QueryRow.values` is an ordered sequence of `(name, Value)` pairs (it is iterated as tuples at :325), so per-row order is available — it is discarded only because `Row` funnels it into a `HashMap`.

1. Change `Row` (`result.rs:237-240`) to ordered, shared storage. Recommended shape:
   ```rust
   pub struct Row {
       /// Interned column names in SELECT order, shared by every row in the result.
       keys: std::sync::Arc<[Py<pyo3::types::PyString>]>,
       /// name -> position, shared by every row (O(1) __getitem__).
       index: std::sync::Arc<std::collections::HashMap<String, usize>>,
       /// Positional values, aligned to `keys`.
       values: Vec<PyObject>,
   }
   ```
   Invariant to preserve if you pick a different concrete type: `keys` must reflect `result.metadata.columns` order exactly; the per-column name String and index map are built **once per result** (shared via `Arc`), never per row; `__getitem__`/`get`/`__contains__` stay O(1) via `index`.
2. In `QueryResult::from_core` (`result.rs:162-190`) build the shared parts once, before the row loop, from `result.metadata.columns` (the same authoritative source the `columns` field uses):
   ```rust
   let keys: std::sync::Arc<[Py<PyString>]> = result
       .metadata
       .columns
       .iter()
       .map(|c| PyString::new(py, &c.name).unbind())
       .collect();
   let index: std::sync::Arc<HashMap<String, usize>> = std::sync::Arc::new(
       result.metadata.columns.iter().enumerate()
           .map(|(i, c)| (c.name.clone(), i))
           .collect(),
   );
   ```
   Pass `keys.clone()` + `index.clone()` into each `Row::from_core` call.
3. Rewrite `Row::from_core` (`result.rs:318-330`) to build a **positional** `Vec<PyObject>` aligned to `index` (look each core `(name, value)` up in `index` and place it; this is robust even if core row order ever drifts from column order):
   ```rust
   pub(crate) fn from_core(
       py: Python<'_>,
       row: &cqlite_core::query::result::QueryRow,
       keys: Arc<[Py<PyString>]>,
       index: Arc<HashMap<String, usize>>,
   ) -> PyResult<Self> {
       let mut values: Vec<Option<PyObject>> = (0..keys.len()).map(|_| None).collect();
       for (name, value) in &row.values {
           if let Some(&pos) = index.get(name) {
               values[pos] = Some(value_to_py(py, value)?);
           }
       }
       let values = values.into_iter()
           .map(|v| v.unwrap_or_else(|| py.None()))
           .collect();
       Ok(Self { keys, index, values })
   }
   ```
4. Rewrite the accessors to iterate positionally:
   - `keys()` (:260-262) → `self.keys.iter().map(|k| k.bind(py).to_string()).collect()` (SELECT order).
   - `values()` (:265-267) → iterate `self.values` in order.
   - `items()` (:270-275) → `zip(self.keys, self.values)` in order.
   - `to_dict()` (:278-284) → iterate `zip(keys, values)`; PyDict preserves insertion order, so the dict is now SELECT-ordered.
   - `__getitem__` (:247-252) / `get` (:290-295) / `__contains__` (:255-257) → look up `self.index.get(key)` then index `self.values` (O(1)).
   - `__len__` (:298-300) → `self.keys.len()`.
   - `__repr__` (:303-315) currently sorts keys for determinism; switch to SELECT order (drop the `sort()`).

## Do NOT
- Do NOT touch `bindings/node/*` — Node's identical ordering bug is **#1446** (W2). This issue is Python only.
- Do NOT change `value_to_py` conversion bodies or add Set/Map/duration behavior changes — those are #1448/#1450.
- Do NOT alter `QueryResult`'s public getters (`rows`, `columns`, `rows_affected`) beyond what the ordered `Row` requires.

## Tests you must add
`bindings/python/tests/test_execute.py` (uses the shared `conftest.py` fixtures; env `CQLITE_DATASETS_ROOT=$PWD/test-data/datasets`).
Add `test_row_keys_match_select_order`:
```python
def test_row_keys_match_select_order(db):
    result = db.execute("SELECT * FROM test_basic.simple_table LIMIT 5")
    assert len(result) > 0, "fixture present but returned 0 rows — datasets unreadable/empty"
    expected = [c.name for c in result.columns]
    for row in result.rows:
        assert list(row.keys()) == expected
        assert list(row.to_dict().keys()) == expected
        assert [k for k, _ in row.items()] == expected
```
Also add an explicit-column-order case (`SELECT b, a, c ...` on a table where that differs from storage order) asserting `list(row.keys()) == ["b","a","c"]`.
Dataset rule: the `len(result) > 0` assertion FAILS LOUDLY when datasets are present-but-empty — do not `pytest.skip` on empty results.
Run: `env CQLITE_DATASETS_ROOT=$PWD/test-data/datasets pytest bindings/python/tests/test_execute.py -v`
Rebuild the extension first: `cd bindings/python && maturin develop`.

## Definition of done
- [ ] `list(row.keys()) == [c.name for c in result.columns]` holds for every row in the new test, including an explicit reordered-column SELECT
- [ ] Column-name Strings are interned once per result (shared `Arc`), not cloned per row
- [ ] File-size ratchet respected (`result.rs` is ~613L; keep it under ~800L — split per campsite rule if the change pushes it over)

pmcfadin:
Closes #1445.

**Change (bindings/python only):** `Row` now stores a shared `RowShape { keys: Arc<[Py<PyString>]>, index: Arc<HashMap<String,usize>> }` + a positional `Vec<PyObject>`, built once per result from `metadata.columns`. Accessors iterate in **SELECT order**; `__getitem__`/`get`/`__contains__` stay O(1). Streaming path caches the same shape once per stream. Kills the per-row column-name re-clone.

**Tests:** 3 new SELECT-order tests in `test_execute.py`; full Python suite green locally (376 passed / 59 skipped / 1 xfailed) via the fast pre-gate. `result.rs` 713 lines (under ratchet).

**Gate note:** local `agent-gate.sh` is currently un-runnable on the dev box (full Rust workspace compile OOMs during clippy). Relying on CI lanes (python-bindings, required, parity) as the authoritative gate for this bindings-only change, per the same standard #1446 merged on. roborev to follow.

Manager-driven (mac-mgr-0702), P1 Ready wave.

### DIFF
--- a/bindings/python/src/result.rs
+++ b/bindings/python/src/result.rs
@@ -176,14 +176,13 @@ impl QueryResult {
         // shares it by reference-count instead of re-cloning column-name
         // Strings per row.
         //
-        // Fallback (issue #1445): the legacy materialized executor wraps rows via
-        // `QueryResult::with_rows(..)` (e.g. point lookups in
-        // `cqlite-core::query::executor`), which leaves `metadata.columns` empty
-        // even though the rows carry values. Without this, every `Row` would be
-        // built with zero columns and drop all its values. When metadata columns
-        // are empty but rows exist, derive the shape from the first row's value
-        // keys (sorted, mirroring the SELECT executor's schema-less path) so the
-        // materialized output matches.
+        // Fallback (issue #1445): the legacy materialized executor wraps rows
+        // via `QueryResult::with_rows(..)` (e.g. point lookups in
+        // `cqlite-core::query::executor`), leaving `metadata.columns` empty even
+        // though rows carry values — without this every `Row` gets zero columns
+        // and drops all values. When empty but rows exist, derive the shape from
+        // the first row's value keys (sorted, mirroring the SELECT executor's
+        // schema-less path) so materialized output matches.
         let shape = if result.metadata.columns.is_empty() {
             match result.rows.first() {
                 Some(first_row) => build_row_shape_from_row_keys(py, first_row),
@@ -429,21 +428,52 @@ impl Row {
     /// lookup so the row is correct even if the core row's `HashMap` iteration
     /// order differs from column order. A column present in `shape` but absent
     /// from the row becomes Python `None`.
+    ///
+    /// Any row value whose key is **not** in `shape` is never dropped (issue
+    /// #1445): such keys are appended after the shaped columns in sorted order.
+    /// This preserves values where `metadata.columns` does not name every row
+    /// value — notably aggregates, keyed by alias (e.g. `Count(*)`) while
+    /// metadata carries a placeholder (`col_0`). When every row key is covered
+    /// (the common scan case) the shared shape is reused unchanged.
     pub(crate) fn from_core(
         py: Python<'_>,
         row: &cqlite_core::query::result::QueryRow,
         shape: RowShape,
     ) -> PyResult<Self> {
         let mut slots: Vec<Option<PyObject>> = (0..shape.keys.len()).map(|_| None).collect();
+        let mut uncovered: Vec<&str> = Vec::new();
         for (name, value) in &row.values {
-            if let Some(&pos) = shape.index.get(name.as_ref()) {
-                slots[pos] = Some(value_to_py(py, value)?);
+            match shape.index.get(name.as_ref()) {
+                Some(&pos) => slots[pos] = Some(value_to_py(py, value)?),
+                None => uncovered.push(name.as_ref()),
             }
         }
-        let values = slots
+
+        let mut values: Vec<PyObject> = slots
             .into_iter()
             .map(|v| v.unwrap_or_else(|| py.None()))
             .collect();
+
+        if uncovered.is_empty() {
+            // Fast path: every row value is covered by the shared shape.
+            return Ok(Self { shape, values });
+        }
+
+        // Slow path: extend the shape for THIS row so no value is dropped.
+        uncovered.sort_unstable();
+        let mut keys: Vec<Py<PyString>> = shape.keys.iter().map(|k| k.clone_ref(py)).collect();
+        let mut index: HashMap<String, usize> = (*shape.index).clone();
+        for name in uncovered {
+            if let Some(value) = row.values.get(name) {
+                index.insert(name.to_string(), keys.len());
+                keys.push(PyString::new(py, name).unbind());
+                values.push(value_to_py(py, value)?);
+            }
+        }
+        let shape = RowShape {
+            keys: keys.into(),
+            index: Arc::new(index),
+        };
         Ok(Self { shape, values })
     }
 }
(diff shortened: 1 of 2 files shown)


pmcfadin:
🧭 **MANAGER** 
Not merge-ready yet. CI is fully green, but roborev flagged a **real HIGH regression**: schema-less **streaming** `SELECT *` builds `RowShape` from an empty `metadata.columns` and drops all row values (`bindings/python/src/result.rs` streaming `row_shape()`). A surgical fix is in progress — build the streaming shape from the first row's value keys when metadata columns are empty, plus a regression test. Will re-roborev + merge on green.

Issue #1773, opened 2026-07-03
Latent: compressed offset-read path (read_value_at_offset/get_cached_data) lacks inline chunk-CRC check (#1411 follow-up)


pmcfadin:
## Summary
Defensive follow-up from #1411. The compressed **offset-read** point-lookup path — `SSTableReader::read_value_at_offset` → `get_cached_data` (`cqlite-core/src/storage/sstable/reader/data_access/mod.rs` ~351-470) — reads `size` raw bytes at an offset and LZ4-decompresses them **without** stripping/validating the 4-byte inline per-chunk CRC32 trailer that `block_io::read_nb_format_chunk_data` enforces.

## Why it's latent (not an active bug today)
#1411 established this path is currently **unreachable** for a compressed `nb` point lookup: `Index.db` is keyed by 16-byte Murmur3 digests, so `find_entry(pk)` misses (issue #517) and `get` falls through to `scan_for_key` (whose CRC bypass #1411 fixed). So no corrupt-input escape exists via `get` today.

## The hazard
If a future change ever makes `find_entry` return a hit for a compressed table (e.g. a real offset index, or BTI point reads routed here), this offset path would **re-introduce the exact #1411 bypass**: a bit-flipped compressed chunk would LZ4-decode to garbage (→ `Ok(None)`/`Ok(Some(garbage))`) instead of surfacing the typed `Error::InvalidFormat("CRC32 mismatch …")`. This is the "refactor reroutes read path around the CRC check" hazard #1397 guards against.

## Routing
Oracle-driven (corruption corpus is the oracle). No OpenSpec.

## Acceptance criteria
- The compressed offset-read path validates the authoritative inline per-chunk CRC32 (reuse `block_io::read_nb_format_chunk_data` — do NOT duplicate the CRC logic) and returns the same typed `Error::InvalidFormat` on mismatch as `scan`/`scan_for_key`.
- A pinned test drives corrupt input through `read_value_at_offset`/`get_cached_data` directly (bypassing the digest-index miss) on the `data_db_bit_flip` LZ4 fixture and asserts the typed CRC error; a clean-fixture analog returns `Ok(Some)`. Fail-closed under `CQLITE_REQUIRE_FIXTURES=1`.
- No-heuristics: authoritative CRC trailer only.

## Notes
The uncompressed analog is already covered by `verify_uncompressed_range` (#1396). Found-by: #1411 implementer.

pmcfadin:
🧭 **MANAGER** 
[...]

Issue #517, opened 2026-05-29
Reader: get() returns None for partition keys that scan() returns


pmcfadin:
## Summary
Revealed by reviving orphan integration tests in #514 (PR #515).

`SSTableReader::get()` returns `None` for partition keys that `scan()` finds and returns from the same SSTable. `get()` and `scan()` disagree about which partitions exist.

## Impact
Point lookups miss partitions that are demonstrably present, producing false-negative query results.

## Affected test (currently `#[ignore]`)
- `test_decompression_integration_with_real_data`

Ignore message cites #514. Remove the ignore once fixed.

## Origin
Found during v0.9.1 work while reviving orphan integration tests (#514). Filed per the stop-condition rather than expanding v0.9.1 scope.

pmcfadin:
Released in **v0.9.2**: https://github.com/pmcfadin/cqlite/releases/tag/v0.9.2

PR #1777, opened 2026-07-03
fix(#1411): surface chunk-CRC corruption on point lookup (get no longer returns Ok(None) on corrupt chunk)

### LINKED ISSUE #1411: Point lookup (get/read_value_at_offset) bypasses inline chunk-CRC check → returns Ok(None) on corrupt COMPRESSED chunk instead of Error
Parent epic: #1380. Found while implementing #1397 (corrupt-query-surface tests).

## Gap / repro
On the corruption corpus `data_db_bit_flip` fixture (a single-bit flip inside the first LZ4 compressed chunk of `test_comp/lz4_table`, single partition `pk=1`, whole table in chunk 0):

- `SSTableReader::scan(...)` → `Err(InvalidFormat("CRC32 mismatch for chunk 0 at offset 0x0 ..."))` — correct fail-fast.
- `SSTableReader::scan_stream(...)` → terminal `Err(... CRC32 mismatch chunk 0 ...)`, 0 rows before error — correct.
- **`SSTableReader::get(table_id, key=pk1)` → `Ok(None)` (deterministic across runs)** — WRONG. The target row lives in the corrupt chunk; the point lookup returns "not found" instead of surfacing the corruption. On the CLEAN fixture the identical `get` returns `Ok(Some)`, so the key/bloom encoding is correct.

## Root-cause hypothesis
The point-lookup path does NOT go through `read_next_block_impl` (block_io.rs:412-440), which is the ONLY place the unconditional per-chunk CRC32 trailer is verified. Instead `get` → bloom(present) → `SSTableReader::read_value_at_offset` → `get_cached_data` (data_access/mod.rs:351-470), which reads `size` raw bytes at an offset and LZ4-decompresses them WITHOUT (a) stripping/validating the 4-byte inline CRC trailer and (b) any CRC check. A bit-flipped chunk therefore either LZ4-decodes to garbage that is then dropped by `filter_tombstone` (observed: `Ok(None)`), or would return `Ok(Some(garbage))` — never the typed corruption error the scan path returns. This is exactly the "refactor that reroutes the read path around `read_next_block_impl` silently returns garbage" hazard #1397 was written to guard.

## Affected files + lines
- `cqlite-core/src/storage/sstable/reader/data_access/mod.rs:351-470` (`read_value_at_offset`, `get_cached_data`) — no inline chunk-CRC validation, no 4-byte trailer handling on the compressed offset path.
- `cqlite-core/src/storage/sstable/reader/data_access/mod.rs:79-166` (`get`/`get_with_resolution` bloom→index path).
- Correct reference: `cqlite-core/src/storage/sstable/reader/block_io.rs:412-440` (the CRC branch the scan/stream paths use).

## Routing
Oracle-driven (parsing/read-path correctness vs Apache Cassandra, which rejects this fixture: `sstableverify -e` → Data.db digest integrity check failed → Invalid SSTable). No OpenSpec.

## Testable acceptance criteria
1. `get`/point-lookup over `data_db_bit_flip` for a partition-key whose row lives in the corrupt chunk returns `Err` (non-recoverable; message names the chunk index + offset), NOT `Ok(None)` and NOT `Ok(Some(garbage))`.
2. The offset-based read path (`read_value_at_offset`/`get_cached_data`) validates the inline chunk CRC32 (or routes through the CRC-checked block reader) for compressed NB SSTables.
3. Parity with the scan path: scan and get agree that the fixture is corrupt.

## Related
- Secondary classification nit surfaced by #1397: the scan-path CRC-mismatch is `Error::InvalidFormat` (not `Error::Corruption`). Both are non-recoverable and the message names chunk+offset, so #1397 asserts the robust invariants; consider unifying the CRC-mismatch to `Error::Corruption` when fixing this.
- Characterized (kept GREEN) in `cqlite-core/tests/issue_1397_corrupt_query_surface.rs` with `// KNOWN LIMITATION: see #<this>`.

Cross-links #1397.

pmcfadin:
Closes #1411

## Problem
`SSTableReader::get(table_id, key)` bypassed the inline per-chunk CRC32 check that `scan`/`scan_stream` enforce, so on a corrupt **compressed** chunk it returned `Ok(None)` ("not found") instead of surfacing the corruption. On the `data_db_bit_flip` LZ4 fixture (`test_comp/lz4_table`, pk=1 in chunk 0), `scan` → `Err(InvalidFormat("CRC32 mismatch for chunk 0 …"))` but `get(pk1)` → `Ok(None)`.

## Root cause (differs from the issue's hypothesis)
The hypothesized `read_value_at_offset`/`get_cached_data` offset path is **never reached** for a compressed `nb` point lookup — `Index.db` is keyed by 16-byte Murmur3 digests so `find_entry(pk)` misses (#517) and `get` falls through to `scan_for_key`. The real bypass: `scan_for_key`'s chunk-stitching branch wrapped `stitch_and_parse_all_chunks` in a blanket `Err(_) => Ok(None)`, swallowing the typed CRC error that `block_io::read_nb_format_chunk_data` (the sole authoritative CRC check) raises.

## Fix (reuses the existing CRC path — no duplication)
Split the combined helper into its two phases in `data_access/sequential.rs`:
1. `stitch_all_chunks(&cursor).await?` — chunk-integrity (CRC32 / decompress / corrupt-offset) now **propagates** via `?`, surfacing the identical `Error::InvalidFormat` the scan path returns.
2. `parse_block(...)` — parse errors classified via `is_parse_soft_miss(schema_present, err)`: soft-miss `Ok(None)` **only** when schema is absent and the error is `Error::Schema` (the legitimate "this reader can't serve this key → try next reader" case); **all** non-recoverable errors (`Corruption`/`InvalidFormat`/`Parse`, and a deep `Error::Schema` when a schema IS present) propagate — at parity with the scan path.

No CRC logic duplicated; the check stays solely in `block_io`. Clean-fixture `get` → `Ok(Some)` unchanged; multi-reader fall-through unchanged.

## Tests (wiring-evidence)
- `issue_1397_corrupt_query_surface.rs`: un-ignored `point_lookup_into_corrupt_chunk_should_error_see_1411` (bit-flip → typed CRC `Err`, same `assert_typed_chunk_corruption` invariant as scan/stream) + added `point_lookup_on_clean_chunk_returns_some_1411` (clean → `Ok(Some)`). Fail-closed under `CQLITE_REQUIRE_FIXTURES=1`; FAILED before / PASS after against real fixture bytes.
- Unit tests for `is_parse_soft_miss`: schema-absent+`Schema` soft-misses; schema-present+`Schema` fatal; `Corruption`/`InvalidFormat`/`Parse` always fatal.

## Quality bar
- `scripts/agent-gate.sh` **PASS** — all 20 components (commit 48a06933; rebased onto current origin/main with zero conflicts and no overlap in `sequential.rs`/`error.rs`).
- Workspace clippy `--all-targets --all-features -D warnings` clean; no `unwrap()`/`expect()` added.
- roborev **clean** (codex, `--base origin/main`; converged after propagating parse-time corruption too).
- Builds verified under default, `all-compression`, and `tombstones`.

## Notes
- `sequential.rs` was already over the file-size threshold (1147L) before this change; the minimal fix pushed it ~+8L → acknowledged with `CQLITE_ALLOW_FILE_GROWTH=1` (a #1116-class split is out of scope for this focused fix).
- Filed **#1773** (P3): a latent identical CRC gap in the currently-unreachable `read_value_at_offset`/`get_cached_data` compressed offset path — defensive follow-up, out of scope here.

### DIFF
--- a/cqlite-core/src/storage/sstable/reader/data_access/sequential.rs
+++ b/cqlite-core/src/storage/sstable/reader/data_access/sequential.rs
@@ -359,23 +359,31 @@ impl SSTableReader {
                 file_guard.seek(SeekFrom::Start(header_size as u64)).await?;
             }
 
+            // Issue #1411: separate the chunk-integrity stitch from the schema-aware
+            // parse so their failure classes are not conflated. `stitch_all_chunks`
+            // drives the authoritative per-chunk CRC32 check
+            // (`block_io::read_nb_format_chunk_data`) + decompression; a CRC mismatch,
+            // decompression failure, or corrupt offset is corruption and MUST surface
+            // (propagate via `?`, matching `scan`), never be masked as a missing key.
+            // Only the `parse_block` step below may soft-miss (schema unavailable /
+            // wrong table type) so the caller can try the next reader. Previously
+            // `stitch_and_parse_all_chunks` combined both under a blanket
+            // `Err(_) => Ok(None)` that swallowed CRC corruption on point lookups.
+            let stitched_buffer = self.stitch_all_chunks(&cursor).await?;
+
             // Pass the reader's own schema so that V5CompressedLegacy rows can be fully
             // parsed and their partition RowKeys emitted.  Without a schema, parse_row_v5
             // fails for all rows in a partition, causing no entries to be pushed and making
             // the key comparison always miss even when the key exists.
             let schema_opt = self.get_table_schema(None);
-            let all_entries = match self
-                .stitch_and_parse_all_chunks(&cursor, schema_opt.as_ref())
-                .await
+            let parser = self.build_v5_parser();
+            let all_entries = match parser.parse_block(&stitched_buffer, schema_opt.as_ref(), self)
             {
                 Ok(entries) => entries,
                 Err(e) => {
                     // Schema may not be available for this reader (e.g., wrong table type).
                     // Return None so the caller can try the next reader.
-                    log::debug!(
-                        "scan_for_key: stitch_and_parse_all_chunks failed (schema missing?): {}",
-                        e
-                    );
+                    log::debug!("scan_for_key: parse_block failed (schema missing?): {}", e);
                     return Ok(None);
                 }
             };
--- a/cqlite-core/src/storage/sstable/reader/data_access/sequential.rs
+++ b/cqlite-core/src/storage/sstable/reader/data_access/sequential.rs
@@ -11,10 +11,36 @@ use super::model::{
     sort_by_token_order, sort_by_token_order_with_meta, table_ids_match, SCAN_FOR_KEY_CALLS,
 };
 use crate::types::{CellWriteMetadata, ScanRow, TableId};
-use crate::{Result, RowKey};
+use crate::{Error, Result, RowKey};
 use std::io::SeekFrom;
 use tokio::io::AsyncSeekExt;
 use tokio::sync::mpsc;
+
+/// Classify a `parse_block` failure on the compressed (`V5CompressedLegacy`)
+/// point-lookup path (`scan_for_key`) as a soft-miss vs. a fatal error.
+///
+/// Issue #1411: the scan path ([`SSTableReader::stitch_and_parse_all_chunks`])
+/// propagates every `parse_block` error via `?`. The point-lookup path mirrors
+/// that so `get()` and `scan()` agree on which failures are fatal — with ONE
+/// deliberate exception, the case the original blanket `Err(_) => Ok(None)`
+/// existed to protect: this reader has no schema for the table, so it cannot
+/// serve the key and the caller must fall through to the next SSTable reader.
+///
+/// Returns `true` (soft-miss → `Ok(None)`) ONLY when both hold:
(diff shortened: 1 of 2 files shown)


pmcfadin:
🔧 flow-lead: gate PASS (commit 48a06933, rebased onto current origin/main, zero conflicts) + roborev clean (codex). Board is greening — the prior #1752 main-red ('All Compression Build & Test') is now resolved upstream and rebased in (that lane PASSES here). Holding merge for the **Compression / corruption parity (regenerate + require-fixtures)** lane to complete — it directly exercises this corruption/CRC read-path change and the local gate does not run it (L2: verify the actual CI lane for corruption/fixture paths). Merge-on-green once it's green.

pmcfadin:
🔧 flow-lead: CI fully green — **Compression / corruption parity (regenerate + require-fixtures)** SUCCESS (the L2-relevant lane exercising this corruption/CRC change), All Compression Build & Test SUCCESS, sstabledump-parity SUCCESS; 20 success / 21 skipped / 0 failures. gate PASS + roborev clean. Merging on green.

Issue #1754, opened 2026-07-03
Node raw-parse path aborts even under a panic=unwind build (corrupt DECIMAL scale)


pmcfadin:
## Summary

While building the #1437 abort-safety regression harness (PR #1753), the Node
binding was found to **abort the host process even under a `panic=unwind`
build** when reading a corrupt SSTable on the uncompressed/raw parse path. The
trigger is a corrupt DECIMAL `scale` field, and the abort happens on the napi
async-worker thread, which cannot unwind across the FFI boundary.

This is a **non-blocking, scope-adjacent finding** — filing only, no fix here.

## Why this matters for the abort-safety guarantee (#1431 / #1440)

#1440 landed a `panic=unwind` build profile for the binding cdylib artifacts as
part of closing Epic U (#1431). This finding shows that **`panic=unwind` alone
does NOT make the Node binding survive corrupt input**: the panic is raised on a
napi async-worker thread and there is no unwind boundary there, so the runtime
still calls `abort()`:

```
fatal runtime error: failed to initiate panic, error 5, aborting
```

So #1440's profile change is necessary but not sufficient — the abort-safety
guarantee for the Node raw-parse path is not fully closed without the two fixes
below.

## Repro

1. Read a corrupt SSTable whose DECIMAL column carries a pathological `scale`
   value, through the Node binding on the raw/uncompressed parse path.
2. Query it via `executeNative()` (or `executeStreaming()`).
3. The Node process aborts (`fatal runtime error: failed to initiate panic,
   error 5, aborting`) — even with a `panic=unwind` build.

## Root-cause hypothesis

`bindings/node/src/value.rs:266` builds:

```rust
format!("0.{digits:0>scale_usize$}")
```

`scale_usize` is derived directly from the DECIMAL `scale` field of the record
with no upper bound. A corrupt/huge `scale` produces a pathological padding
width, and the `format!` panics (allocation/width blowup). Because the decode
runs on a napi async-worker thread, the panic cannot unwind across the FFI
boundary and the process aborts instead of surfacing an error.

Two independent gaps:
- **(a)** `scale` is used without fail-closed validation against authoritative
  metadata (no-heuristics: bound/validate before use).
- **(b)** There is no `catch_unwind` at the napi boundary, so a worker-thread
  panic becomes an abort rather than a typed JS error. (No `catch_unwind`
  currently exists in `bindings/node/src/`.)

## Affected files

[...]

Issue #1742, opened 2026-07-02
Add a query-semantics oracle: corpus goldens are physical sstabledump dumps, so read-reconciliation bugs pass parity green


pmcfadin:
Filed by flow-lead from the #942 design-doc validation (2026-07-02). Companion to the single-gen read-reconciliation P0 (filed same day — see cross-link in comments).

## Gap

The parity suite's JSONL goldens are **physical dumps** (sstabledump-style): they enumerate on-disk cells *including* tombstones, deleted rows, and expired-but-uncompacted TTL cells. Row-count/value parity compares CQLite output to these dumps — so when CQLite fails to apply read-time reconciliation, **both sides include the shadowed/expired rows and the test passes**. Concrete example: `test_basic/ttl_test_table`'s golden contains all 100 expired rows; CQLite returns all 100; parity is green; a real Cassandra `SELECT` returns 0.

Additionally, every tombstone/RT corpus fixture was produced via flush/compaction that reconciled shadowed data away — no fixture presents coexisting tombstone + shadowed-older-data to a single-generation reader. The physical-dump oracle therefore structurally cannot catch read-reconciliation bugs.

## Acceptance criteria

1. A **query-semantics oracle** distinct from the physical goldens: per-fixture expected *post-reconciliation* result sets for canonical SELECTs (what a real Cassandra `SELECT` returns at a pinned evaluation time — TTL evaluation must use a pinned `now`, never wall-clock, to avoid time flakes).
2. New single-generation fixtures carrying coexisting: partition deletion + older live rows; range tombstone + rows inside its bounds; expired TTL cells with values still on disk. Generated via real Cassandra (test-data-management flow) with provenance recorded in `test-data/cassandra-parity-manifest.yml`.
3. Integration test lane comparing CQLite SELECT output to the semantic oracle across the corpus; wired as an agent-gate component (SKIP-aware on missing datasets; **0-rows-on-present-data is a failure**).
4. Docs: validation-playbook page updated to distinguish physical-dump parity vs query-semantics parity.

**Routing: oracle-driven test infrastructure** — no OpenSpec change.

pmcfadin:
Companion P0: #1741 (single-gen read reconciliation) — the bug this oracle exists to catch. — flow-lead

pmcfadin:
🧭 **MANAGER** 
GO
ORDER: 17

[...]

Issue #1437, opened 2026-07-02
Abort-safety regression harness: corrupt SSTable through both bindings must raise, not kill the host


pmcfadin:
Parent: #1431

## Why this matters
The bindings are the surface most users touch. Because the workspace release profile is `panic = "abort"`, a panic raised inside `cqlite-core` while a scan runs through PyO3/napi does not become a catchable exception — it aborts the whole Python interpreter or Node process. The parser audit confirmed core has real corrupt-input panic paths (ZigZag lengths, BTI byte-depth cap), so today a user who points either binding at a corrupt/truncated SSTable can have their host process killed outright. No test anywhere proves the process survives corrupt input. This issue builds the regression harness that makes that guarantee testable; the actual firewall (`panic=unwind` profile) lands in #1440.

## Current behavior (verified 2026-07-01)
Release profile aborts on panic — `Cargo.toml:156-160`:
```toml
[profile.release]
codegen-units = 1
lto = true
panic = "abort"
strip = true
```
Wheels are built release-profile — `.github/workflows/python-release.yml:93-96` (PyO3/maturin-action):
```yaml
        uses: PyO3/maturin-action@v1
        with:
          ...
          args: --release --out dist
```
npm prebuilds are built release-profile — `bindings/node/package.json:29`:
```json
    "build": "napi build --platform --release --features write-support",
```
Neither binding calls `catch_unwind` (rg: 0 hits), and PyO3's/napi-rs's built-in catch_unwind→exception machinery is dead code under abort. There is no test that drives corrupt input through a release build.

## Step-by-step implementation
This issue is the harness only. Do not change the panic profile (that is #1440); build the test that fails today and will pass once #1440 lands.

1. Generate a corrupt fixture without committing large binaries. Add a small helper (Python: new module `bindings/python/tests/corrupt_fixture.py`; Node: new module `bindings/node/__test__/corrupt-fixture.js`) that, given a temp dir, copies a real SSTable directory and mutates the Data.db. Source table: `test-data/datasets/sstables/test_basic/simple_table-6aa08200a25111f0a3fef1a551383fb9/` (files `nb-1-big-*.db`). Two mutation modes, both must be exercised:
   - **truncate**: copy `nb-1-big-Data.db`, then truncate it to ~50% of its byte length.
[...]

PR #1790, opened 2026-07-03
fix(#1741): apply partition/range-tombstone + TTL shadowing on single-gen reads

### LINKED ISSUE #1741: Single-generation read path never applies partition deletions, range tombstones, or TTL expiry — returns deleted/expired data as live
Filed by flow-lead from the #942 design-doc validation (2026-07-02, main @6814f17f). Read-side sibling of CLOSED compaction-side P0 #1382; same defect class (returns deleted/expired data as live). Relates: #942, epic #1378 family, #1600 (TombstoneMerger confinement).

## Bug

Multi-generation reads (`candidates > 1`) reconcile via `KWayMerger`/`ReconcileState` (`storage/sstable/mod.rs:1099, 1446` — note: `#[cfg(feature = "write-support")]`-gated). **Single-generation reads — the common case — bypass reconciliation entirely** via `parse_block_emit_with_metadata` (`reader/parsing/v5_compressed_legacy/block_emit.rs:80-205`):

- **Partition-level deletion dropped:** `block_emit.rs:80` calls the 2-tuple `parse_partition_header`, which discards `_deletion` (`row_framing.rs:557-560`). The emit loop never sees the partition tombstone.
- **Range tombstones skipped, never shadow:** `skip_range_tombstone_marker` (`block_emit.rs:101-109` → `row_framing.rs:120+`) only advances the offset; no RT state retained, zero `apply_range_shadowing` under `reader/`.
- **No read-time TTL expiry:** no `localDeletionTime`-vs-now evaluation exists in the emit path.
- Row-level tombstones ARE handled correctly inline (#932) — verified via `shadow_row_delete`.

## Confirmed reproductions

**(b) Partition deletion — empirical.** Single-gen uncompressed SSTable (schema `CREATE TABLE tomb_ks.t (id int, ck int, v text, PRIMARY KEY (id, ck))`), two rows liveness ts=1000µs, partition header patched to `markedForDeleteAt=2000µs` (bytes 6..10 localDeletionTime → 0x60000000, bytes 10..18 → 2000; stale CRC/Digest removed, reader warn-and-proceeds):
```
cqlite --schema /tmp/tomb_test.cql --data-dir /tmp/patched \
  --query "SELECT id,ck,v FROM tomb_ks.t" --out json
```
Cassandra: **0 rows**. CQLite: **2 rows** (`{id:5,ck:1,v:"aaa"},{id:5,ck:2,v:"bbb"}`). Reproduces on both the default and write-support builds.

**(c) TTL expiry — empirical, real Cassandra fixture.** `test-data/datasets/sstables/test_basic/ttl_test_table-6af66a30a25111f0a3fef1a551383fb9/nb-1-big-Data.db`: all 100 rows `ttl=86400`, expired 2025-10-07 (≈9 months before today).
```
cqlite --schema test-data/schemas/basic-types.cql --data-dir test-data/datasets/sstables \
  --query "SELECT id, expiring_value, session_info FROM test_basic.ttl_test_table LIMIT 3" --out json
```
Cassandra: **0 rows**. CQLite: **all 100 rows returned live**.

**(a) Range tombstones — code-confirmed.** No empirical single-gen repro is currently synthesizable: CQLite's writer reconciles at flush (covered rows purged before hitting disk) and every corpus RT fixture had its shadowed rows reconciled away pre-write. The read path is the identical skip-only code proven wrong for (b).

## Impact

All read surfaces — CLI SELECT, Python, Node, all exports (Parquet/delta/one-shot) — the defect is in the shared v5 emit path below the query engine. Independent of compression and of feature flags. TTL-heavy timeseries fixtures (`log_entries`, `app_metrics`, `tick_data`) serve expired data today.

## Why 33/33 parity is green

JSONL goldens are *physical sstabledump dumps* (they include tombstones and expired-but-uncompacted cells). When CQLite also skips reconciliation, both sides contain the shadowed rows and parity passes while both diverge from a real Cassandra `SELECT`. Companion issue (query-semantics oracle) filed separately.

## Acceptance criteria

1. Single-gen reads apply partition-deletion shadowing, range-tombstone shadowing, and read-time TTL expiry, matching Cassandra `SELECT` semantics — preferably by routing through the same `ReconcileState` core the `KWayMerger` uses (one reconciliation implementation).
2. Read-side reconciliation must not be gated on `write-support` (read correctness cannot depend on a write feature).
3. Pinned parity tests from the repros above: hand-built single-gen fixture(s) with coexisting partition-deletion + older rows, RT + inside rows, and expired TTL cells; tests demonstrably fail on current main (revert-verify) and must not pass on an empty dataset.
4. `scripts/agent-gate.sh` PASS with summary block.

**Routing: oracle-driven** (SSTable parsing/tombstone parity) — issue + pinned parity test, no OpenSpec change, per delivery doctrine.

pmcfadin:
Closes #1741.

## What
**P0 read-path correctness:** single-generation / full-scan reads bypassed reconciliation and returned deleted/expired data as live. This applies **partition-deletion shadowing, range-tombstone shadowing, and read-time TTL expiry** to single-gen reads, matching Cassandra `SELECT` semantics.

Empirically confirmed before/after: `test_basic.ttl_test_table` (100 rows, all TTL-expired ~2025-10-07) → Cassandra `0`, CQLite was `100`, now `0`.

## Approach (design note — deviates from AC1's literal wording; owner please confirm)
AC1 said "preferably route through the write-support-gated `ReconcileState`." Investigation showed that path is a **large/risky refactor** (`ReconcileState` is `write-support`-gated + a different `CellData` data model). Per the escape hatch, I instead implemented AC1's **intent** with a single **un-gated** read-side helper `PartitionShadow` called from all ~5 emit paths — satisfying AC2 (read correctness NOT gated on `write-support`) without pulling write-support into the read path. No-heuristics: authoritative header/RT/`Statistics.db` fields only. Hot path stays zero-per-cell-alloc (per-row temporal aggregate computed in the existing cell loop; `Statistics.db`-gated skip when an SSTable has no tombstones/TTL).

## Emit paths covered (all 5)
streaming `SELECT *`, BTI/big-promoted single-partition + point reads, WRITETIME/TTL projections, `parse_block`, and `bti_scan_with_metadata` + index/sequential/point-get. Physical consumers (integrity verify, delta-scan, compaction) pass `read_shadowing=false` and are byte-unchanged.

## Coupled writer fix (required — surfaced by the read fix)
CQLite's `da`/BTI `DataWriter` wrote the partition-level `DeletionTime` in **nb** layout even for `da` files, whose reader correctly applies **oa** (`hasUIntDeletionTime`) decoding → it misread the live-partition sentinel as a tombstone (harmless before, since deletion was discarded; now shadows every row). Fixed to emit oa layout for `da` (guarded, **nb byte-identical**). Real Cassandra `da` fixtures unaffected; only CQLite-written `da` files (which would otherwise return 0 rows on any SELECT). Validated green vs `issue_909`/`issue_908`/`issue_821`/`issue_1190` byte-parity + compaction-byte-parity.

## Robustness fix (regression I introduced, caught by the #1437 harness)
The new shadowing path changed which corrupt row surfaces first on truncated/bitflipped input, exposing an **unbounded DECIMAL stringify** in the Python binding (`decimal_to_pydecimal`) → python3.14 4300-digit `ValueError` crashing the driver (A/B proven: main 10/10, pre-fix branch 7/10). Guarded fail-closed: bound against `sys.get_int_max_str_digits()`, return a typed `CqliteError` instead of stringifying an unbounded int. abort-safety back to 10/10. (Related pre-existing class: #1754.)

## Tests (AC3) — pinned, revert-verified, not empty-passing
- `cqlite-core/tests/issue_1741_singlegen_tombstone_ttl_shadow.rs`: TTL 100→0 (asserts golden==100 first, anti-empty-pass) + partition-delete on a patched real fixture; both fail on pre-fix code. `partition_shadow.rs` unit tests (partition `<=`/`>`, RT FSM/coverage/boundary, TTL live/expired/live-forever, no-timestamp fail-safe).
- `test_parity.py` reconciled to SELECT semantics via a **wall-clock-aware live-count helper** (excludes tombstoned + TTL-expired golden rows; no hardcoded 0; non-TTL tables assert exact physical parity unchanged). This is the #1742 (query-semantics oracle) territory — done minimally here to keep the gate honest.

## Gate: PASS (all components; `python-bindings` PASS)
`CQLITE_ALLOW_FILE_GROWTH=1` acknowledged — the shadowing implementation grew 6 already-over-threshold parser/writer files; splitting them is tracked under epic #1116, out of scope for this P0.

### DIFF
--- a/bindings/python/python/cqlite/__init__.py
+++ b/bindings/python/python/cqlite/__init__.py
@@ -5,6 +5,8 @@ from cqlite._cqlite import (
     version,
     # Test-support introspection (issue #1437)
     _built_with_panic_abort,
+    # Test-support: direct DECIMAL rendering path (issue #1741)
+    _decimal_from_parts,
     # Exception types
     CqliteError,
     SchemaError,
@@ -41,6 +43,8 @@ __all__ = [
     "version",
     # Test-support introspection (issue #1437)
     "_built_with_panic_abort",
+    # Test-support: direct DECIMAL rendering path (issue #1741)
+    "_decimal_from_parts",
     # Exception types
     "CqliteError",
     "SchemaError",
--- a/bindings/python/src/lib.rs
+++ b/bindings/python/src/lib.rs
@@ -58,6 +58,21 @@ fn _built_with_panic_abort() -> bool {
     cfg!(panic = "abort")
 }
 
+/// Test-support: render a CQL DECIMAL from its raw parts — `scale` and the
+/// big-endian two's-complement `unscaled` magnitude — through the exact
+/// production conversion (`value::decimal_to_pydecimal`).
+///
+/// Lets the pytest suite exercise the fail-closed corrupt-DECIMAL guard (issue
+/// #1741) directly: a large-but-REPRESENTABLE unscaled value must render as a
+/// `decimal.Decimal`, while a genuinely oversized one must raise a typed
+/// `CqliteError` (never abort the interpreter) — without needing a multi-kilobyte
+/// on-disk fixture. Not part of the stable public API; the leading underscore
+/// marks it internal test support.
+#[pyfunction]
+fn _decimal_from_parts(py: Python<'_>, scale: i32, unscaled: Vec<u8>) -> PyResult<PyObject> {
+    value::decimal_to_pydecimal(py, scale, &unscaled)
+}
+
 /// Python module for CQLite.
 #[pymodule]
 fn _cqlite(m: &Bound<'_, PyModule>) -> PyResult<()> {
@@ -67,6 +82,10 @@ fn _cqlite(m: &Bound<'_, PyModule>) -> PyResult<()> {
 
     // Test-support introspection (issue #1437). See the fn doc comment.
     m.add_function(wrap_pyfunction!(_built_with_panic_abort, m)?)?;
+
+    // Test-support: direct DECIMAL rendering path for the corrupt-guard tests
+    // (issue #1741). See the fn doc comment.
+    m.add_function(wrap_pyfunction!(_decimal_from_parts, m)?)?;
 
     // Register exception types
     error::register_exceptions(m)?;
--- a/bindings/python/src/value.rs
+++ b/bindings/python/src/value.rs
@@ -7,6 +7,7 @@ use pyo3::exceptions::PyKeyError;
 use pyo3::prelude::*;
 use pyo3::types::{PyBytes, PyDict, PyFrozenSet, PyList, PyTuple};
 
+use crate::error::to_py_err;
 use cqlite_core::Value;
 
 /// Convert a CQL Value to a Python object.
@@ -283,7 +284,15 @@ fn varint_to_pyint(py: Python<'_>, bytes: &[u8]) -> PyResult<PyObject> {
 }
 
 /// Convert decimal to Python decimal.Decimal.
-fn decimal_to_pydecimal(py: Python<'_>, scale: i32, unscaled: &[u8]) -> PyResult<PyObject> {
+///
+/// `pub(crate)` so the internal `_decimal_from_parts` test helper (lib.rs) can
+/// drive this exact conversion path directly, exercising the fail-closed
+/// corrupt-DECIMAL guard (issue #1741) without a multi-kilobyte on-disk fixture.
+pub(crate) fn decimal_to_pydecimal(
+    py: Python<'_>,
+    scale: i32,
+    unscaled: &[u8],
+) -> PyResult<PyObject> {
     let decimal_mod = py.import("decimal")?;
     let decimal_class = decimal_mod.getattr("Decimal")?;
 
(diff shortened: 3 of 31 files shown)


pmcfadin:
✅ **Merging (single-gen split of #1741).** Multi-gen read visibility split to follow-up **#1849**.
- roborev (codex): **clean, no issues** on the rebased single-gen branch.
- Invariant: `git diff origin/main HEAD -- cqlite-core/src/storage/write_engine/merge` is **EMPTY** — this change touches ZERO merger code; multi-gen reads unchanged from origin/main (→ #1849).
- Gate: 17/19 components PASS incl. compaction-byte-parity. The 2 red components are **proven pre-existing / environment-driven, not this change** (independently verified):
  - `tooling-tests` — the same 2 assertions (`AGENT_GATE_TEST_NO_METADATA_PARSER` no-parser fallback, ~#1821) fail identically on a fresh detached `origin/main` worktree; `git diff origin/main HEAD -- scripts/` is empty (zero scripts changes). A cold main worktree is in fact redder (10 vs 2).
  - `python-bindings` — sole failure `test_streaming.py::test_streaming_next_releases_gil`, a GIL-free-fraction timing assertion (~0.02 vs 0.05 floor) that fails identically on clean `origin/main` (Python 3.14.6/ARM); no streaming/PyO3 src changes; `test_parity.py` (the #1741 requirement) passes 47/1xfail.
- Confirmed single-gen repros: partition-delete → row hidden; `test_basic.ttl_test_table` → 0 rows. Merged over the pre-existing reds with evidence per delivery doctrine (cf #1093/#1623).

Issue #1609, opened 2026-07-02
Epic P — fsync discipline + ingest/flush/replay mechanics


pmcfadin:
## Epic P — fsync discipline + ingest/flush/replay mechanics

Source: `docs/reports/write-path-performance-audit-2026-07-01.md` (Executive summary point 5; Epic P table P1–P5; Test-infrastructure summary). Priority label: `P1`.

### Theme

The write path's durability step is correct but has **no group commit**, and the flush/replay/finish mechanics carry avoidable full copies and full-file re-reads:

- **fsync discipline has no group commit.** A `BEGIN BATCH` of N statements performs **N sequential fsyncs** — the per-statement path appends + `sync()` per mutation (`write_engine/mod.rs:531-534`) and the BATCH loop just calls it N times (`mod.rs:709-719`). At 1–10 ms/fsync a 100-row batch costs 0.1–1 s of serialized fsync latency.
- **Per-append allocation.** Every WAL append allocates a fresh `Vec` via `bincode::serialize` (`wal.rs:642`), and the single-component partition-key path re-wraps an already-serialized `Vec` into a second one (`mutation.rs:434-439`).
- **Replay materializes the whole log.** `WriteAheadLog::replay()` returns `Vec<Mutation>` for the entire WAL (`wal.rs:719-810`), doubling peak memory on recovery.
- **Flush deep-copies the memtable.** Flush clones every key and `to_vec()`s every partition's mutations (`mod.rs:825`) and walks the BTreeMap twice with `.iter().count()` (`mod.rs:788,794`).
- **finish() re-reads Data.db twice.** Digest (`finish.rs:160`) and CRC.db (`finish.rs:185`) each stream the whole finished Data.db from disk again, for checksums that could be computed incrementally while writing.

What is already good and MUST be preserved: the fsync itself is real and instrumented with a latency histogram on the durable step only (`wal.rs:680-698` — keep). WAL replay **correctness** semantics (torn-tail + lossy-replay handling) are owned by the filed parity P0s **#1390** and **#1391** — this epic is their perf sibling and must not alter their error-handling decisions. CRC.db/Digest byte output is golden-gated (#1190 flush / #1017 compaction / #1222 trailer, #1190/#1017/#1222) and must stay byte-identical.

### Children (tracking)

- [ ] P1 — Group commit for BATCH: one WAL fsync per batch, not N (#1659)
- [ ] P2 — WAL append scratch reuse + single-component PK fast path (#1660)
[...]

PR #1806, opened 2026-07-03
perf(#1660): WAL append scratch reuse + single-component PK fast path

### LINKED ISSUE #1660: P2 — WAL append scratch reuse + single-component PK fast path
Part of epic #1609 (Epic P — fsync discipline + flush/replay mechanics). Source: docs/reports/write-path-performance-audit-2026-07-01.md finding P2.

## Problem

Two avoidable allocations on **every** WAL append / write.

**(a) Fresh `Vec` per WAL append.** `WriteAheadLog::append` allocates a brand-new `Vec<u8>` for every mutation via `bincode::serialize` (`cqlite-core/src/storage/write_engine/wal.rs:640-643`):

```rust
pub fn append(&mut self, mutation: &Mutation) -> Result<()> {
    // Serialize mutation using bincode
    let mutation_bytes = bincode::serialize(mutation)          // <-- fresh Vec per call
        .map_err(|e| Error::Storage(format!("Failed to serialize mutation: {}", e)))?;
    let entry_length = mutation_bytes.len() as u32;
    let mut hasher = Hasher::new();
    hasher.update(&mutation_bytes);
    // ... write [length][crc32][mutation_bytes]
```

**(b) Double `Vec` for a single-component partition key.** `PartitionKey::to_bytes` builds an empty `result` Vec, calls `serialize_value` (which itself returns an owned `Vec`), then copies that into `result` and returns it — two allocations + one copy for the common single-PK case (`mutation.rs:432-439`):

```rust
let mut result = Vec::new();
// Single-component key: no length prefix
if self.columns.len() == 1 {
    let value_bytes =
        self.serialize_value(&self.columns[0].1, &schema.partition_keys[0])?;  // Vec #1
    result.extend_from_slice(&value_bytes);   // copy into Vec #2 (result)
    return Ok(result);
}
```

**User impact:** every insert pays 4–6 heap allocations (audit-measured) where a steady-state append needs ~1–2. Under sustained ingest this is pure allocator/GC pressure on the hottest write path.

## Fix (effort: S)

**(a) Persistent scratch buffer + `serialize_into`:**

1. Add a reusable scratch buffer field to `WriteAheadLog`, e.g. `append_scratch: Vec<u8>`.
2. In `append`, clear the scratch and serialize into it instead of allocating:
   ```rust
   self.append_scratch.clear();
   bincode::serialize_into(&mut self.append_scratch, mutation)
       .map_err(|e| Error::Storage(format!("Failed to serialize mutation: {}", e)))?;
   let entry_length = self.append_scratch.len() as u32;
   let mut hasher = Hasher::new();
   hasher.update(&self.append_scratch);
   let crc32 = hasher.finalize();
   self.file.write_all(&entry_length.to_le_bytes())?;   // (keep existing error mapping)
   self.file.write_all(&crc32.to_le_bytes())?;
   self.file.write_all(&self.append_scratch)?;
   self.current_size += 8 + entry_length as u64;
   ```
   `bincode::serialize_into` produces **byte-identical** output to `bincode::serialize` — the on-disk record does not change. Borrow-check note: `self.append_scratch` and `self.file` are different fields, so serialize-into-scratch then write-from-scratch is fine; if the borrow checker complains, `std::mem::take` the scratch into a local, use it, then move it back.

**(b) Single-component PK fast path:**

3. In `PartitionKey::to_bytes` (`mutation.rs:434-439`), return the already-serialized bytes directly for the single-component case, dropping the empty `result` allocation and the copy:
   ```rust
   if self.columns.len() == 1 {
       // Single-component key is written as raw value bytes: return them directly.
       return self.serialize_value(&self.columns[0].1, &schema.partition_keys[0]);
   }
   ```
   Move the `let mut result = Vec::new();` declaration below this early return so it is only allocated for the multi-component path.

## Do NOT

- Do NOT change the WAL record layout or the bincode encoding — `serialize_into` must yield the same bytes as `serialize`. Replay of pre-change WALs must still work.
- Do NOT change the single-component on-disk PK encoding — it is still "raw value bytes, no length prefix". Only the redundant Vec + copy is removed; output bytes are identical.
- Do NOT change the multi-component PK path (2-byte BE length prefix + `0x00` end markers).
- Do NOT share one scratch across threads — `WriteAheadLog` is `&mut self`; keep the scratch a plain owned field.
- No `unwrap()`/`expect()` in library code.

## Acceptance criteria

- [ ] `append` performs no per-call full-size `Vec` allocation for the serialized mutation (reuses scratch).
- [ ] Single-component `PartitionKey::to_bytes` returns the serialized value with no extra Vec/copy.
- [ ] WAL replay round-trip unchanged (existing replay tests pass); PK `to_bytes` byte output unchanged for both single- and multi-component keys.

Extend the Epic O4 alloc-budget lane (dhat-heap-gated, per `cqlite-core/tests/test_issue_827_merge_streaming_memory.rs`) or an allocation counter:

1. Open a `WriteEngine` (single-component TEXT/UUID PK table), warm up one write.
2. Measure allocations across a steady-state `write()` of one mutation.
3. Assert allocs per `write()` ≤ K (a small constant, e.g. ≤ 2).

**On current main this measures ~4–6 allocations per `write()`** — the test FAILS on main and passes after the fix.

pmcfadin:
Closes #1660. Part of epic #1609 (Epic P — write-path mechanics). Oracle-driven perf fix — **on-disk bytes are byte-identical** before/after.

## Changes
- **WAL append scratch buffer** (`wal.rs`): `WriteAheadLog` gains a reusable `append_scratch: Vec<u8>`; `append` now `clear()`s it and uses `bincode::serialize_into` (byte-identical to `bincode::serialize`) instead of allocating a fresh `Vec` per mutation. Record layout unchanged; pre-change WAL replay still works.
- **Single-component PK fast path** (`mutation.rs`): `PartitionKey::to_bytes` returns the serialized value directly for the single-component case, dropping the empty `result` Vec + copy. Multi-component encoding (2-byte BE length prefix + `0x00` markers) and single-component on-disk encoding unchanged.

## Test (TDD — failed on main, passes after)
`write_path_alloc_budget_holds` (`cqlite-core/tests/test_issue_1660_write_path_allocs.rs`): counting-allocator budget over a steady-state `write()` window on a single-PK TEXT table.
- Before: **5.0 allocs/write** → FAILS
- After: **3.0 allocs/write** → PASSES (budget `< 4/write`)

Existing WAL replay round-trip + PK `to_bytes` tests confirm byte output unchanged (single- and multi-component).

## Gate
`scripts/agent-gate.sh` → **RESULT: PASS** (all components; `file-size` acked via `CQLITE_ALLOW_FILE_GROWTH=1` for the +9-line `wal.rs` minimal-fix growth, per epic #1116).

```
file-size:         PASS
fmt:               PASS
clippy:            PASS
core-tests:        PASS
write-tests:       PASS
compaction-byte-parity: PASS
python-bindings:   PASS
node-bindings:     PASS
... (all PASS)
RESULT: PASS
```

### DIFF
--- a/cqlite-core/src/storage/write_engine/mutation.rs
+++ b/cqlite-core/src/storage/write_engine/mutation.rs
@@ -429,15 +429,13 @@ impl PartitionKey {
             )));
         }
 
-        let mut result = Vec::new();
-
-        // Single-component key: no length prefix
+        // Single-component key: no length prefix. Return the serialized value
+        // directly to avoid allocating (and copying into) an intermediate Vec.
         if self.columns.len() == 1 {
-            let value_bytes =
-                self.serialize_value(&self.columns[0].1, &schema.partition_keys[0])?;
-            result.extend_from_slice(&value_bytes);
-            return Ok(result);
+            return self.serialize_value(&self.columns[0].1, &schema.partition_keys[0]);
         }
+
+        let mut result = Vec::new();
 
         // Multi-component partition keys use a `0x00` end-of-component marker
         // after every component, including the last one.
--- a/cqlite-core/src/storage/write_engine/wal.rs
+++ b/cqlite-core/src/storage/write_engine/wal.rs
@@ -667,6 +667,10 @@ pub struct WriteAheadLog {
     buffer_size: usize,
     /// Current size of the WAL file (in bytes)
     current_size: u64,
+    /// Reusable scratch buffer for serializing a mutation in `append`, so the
+    /// steady-state write path does not allocate a fresh `Vec` per call. Cleared
+    /// and refilled on every `append`; never shared across threads (`&mut self`).
+    append_scratch: Vec<u8>,
     /// Byte offset of the last CRC-valid prefix when `open_existing` detected
     /// mid-stream corruption (issue #1391). The corrupt tail is intentionally
     /// left on disk so the caller can preserve it aside as forensic evidence
@@ -754,6 +758,7 @@ impl WriteAheadLog {
             path,
             buffer_size,
             current_size: 0,
+            append_scratch: Vec::new(),
             pending_valid_prefix: None,
             poisoned: None,
             #[cfg(test)]
@@ -860,6 +865,7 @@ impl WriteAheadLog {
             path: path.to_path_buf(),
             buffer_size: Self::DEFAULT_BUFFER_SIZE,
             current_size,
+            append_scratch: Vec::new(),
             pending_valid_prefix,
             poisoned: None,
             #[cfg(test)]
@@ -1118,8 +1124,11 @@ impl WriteAheadLog {
             ));
         }
 
-        // Serialize mutation using bincode
-        let mutation_bytes = bincode::serialize(mutation)
+        // Serialize mutation using bincode into a reusable scratch buffer so the
+        // steady-state write path does not allocate a fresh `Vec` per call.
+        // `serialize_into` yields byte-identical output to `bincode::serialize`.
+        self.append_scratch.clear();
+        bincode::serialize_into(&mut self.append_scratch, mutation)
             .map_err(|e| Error::Storage(format!("Failed to serialize mutation: {}", e)))?;
 
         // Fail-closed size ceiling (issue #1391, roborev r3). `replay()` /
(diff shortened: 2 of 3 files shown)


pmcfadin:
Armed merge-on-green. Gate PASS + roborev clean (codex, no issues). Oracle-driven, no C required. Deferring merge to the manager poller on CI-green.

Issue #1885, opened 2026-07-04
v5 ladder decode has no varint arm → CQL varint read as Value::Blob (mirror of I4 on the v5 side)


pmcfadin:
## Context
Surfaced by the #1617 decoder-lockstep parity net (divergence #2). CQL `varint`:
- **block path**: `Value::Varint` — correct.
- **v5 ladder** (`parse_cell_value_schema_order`): no `varint` arm → falls through to the blob default → `Value::Blob`.

## Problem
A `varint` column read through the live v5 per-cell decoder yields `Value::Blob` instead of `Value::Varint` — the v5-side mirror of the block-path I4 blob-fallback issue. Pinned by `decoder_lockstep_tests` (asserted via `assert_ne` with a link to this issue).

## Routing
Oracle-driven (type-decode correctness; block path + write side are the reference). No OpenSpec.

## Acceptance criteria
- v5 ladder gains a `varint` arm decoding to `Value::Varint`, matching the block path + write side round-trip.
- The `decoder_lockstep_tests` varint case flips from recorded-divergence to converged — update that test as part of the fix.
- Value parity vs sstabledump goldens for a table with a `varint` column read through the v5 path.

Found by the #1617 lockstep net; filed by flow-lead worker.

pmcfadin:
Milestoned 0.15 per owner-approved board grouping (2026-07-09): unscheduled correctness/perf work pulled into the read/write-path release so no P1-P2 bug sits milestone-less.

pmcfadin:
🔒 Claimed by flow-lead worker on machine `pmcfadin-ltmcx74` — branch `issue-1885-v5-varint-arm` pushed (the lock). Oracle-driven P2, fully self-contained (block path is the reference; pinned decoder_lockstep_tests divergence to flip). Implementing.

Issue #1884, opened 2026-07-04
v5 ladder decode widens CQL float to f64 (Value::Float) instead of Value::Float32 — lossy vs block path


pmcfadin:
## Context
Surfaced by the #1617 decoder-lockstep parity net (divergence #1). The two live read decode paths disagree on CQL `float`:
- **block path** (`parse_value_with_schema_type`): `Value::Float32(f32)` — correct.
- **v5 ladder** (`parse_cell_value_schema_order` under `v5_compressed_legacy/cell_value.rs`): `Value::Float(f32 as f64)` — widens single-precision to f64.

## Problem
CQL `float` is IEEE single-precision. The v5 per-cell decoder emits a widened `Value::Float` (f64) rather than `Value::Float32`, a lossy/incorrect representation vs the block path (which was corrected post-#1627). This is a live divergence pinned by `decoder_lockstep_tests` (asserted via `assert_ne` with a link to this issue).

## Routing
Oracle-driven (type-decode correctness; block path + write side are the reference). No OpenSpec.

## Acceptance criteria
- v5 ladder decodes CQL `float` to `Value::Float32(f32)`, matching the block path.
- The `decoder_lockstep_tests` float case flips from recorded-divergence (`assert_ne`) to converged (`assert_eq`) — update that test as part of the fix.
- Value parity vs sstabledump goldens for a table with a `float` column, read through the v5 path.

Found by the #1617 lockstep net; filed by flow-lead worker.

pmcfadin:
Milestoned 0.15 per owner-approved board grouping (2026-07-09): unscheduled correctness/perf work pulled into the read/write-path release so no P1-P2 bug sits milestone-less.

Issue #1856, opened 2026-07-04
Test-hygiene: issue_953_multichunk_cell_seek hard-panics (not skips) when fixture dir present but -Data.db binary absent


pmcfadin:
## Problem
`cqlite-core/tests/issue_953_multichunk_cell_seek.rs` claims (doc comment ~line 45) to be *skipped, not failed* when its fixture is absent. But the skip only fires when the whole `test_da/wide_table-*` directory is missing.

When the directory is **present with only the committed JSONL/CRC/TOC** and the actual `da-2-bti-Data.db` binary absent (the normal state on any checkout that hasn't fetched the local-only large fixture — the binary is NOT in the fetchable `cassandra5-small-full-v3.4` asset), `real_wide_table_dir()` returns `Some`, `build_repacked_sstable` finds no `-Data.db` file, and `data_path.expect("Data.db must be present")` at **line 176 hard-panics**.

## Impact
Any `scripts/agent-gate.sh` run in a worktree pointed at the standard root `test-data/datasets` (the documented worktree pattern) FAILs `core-tests` on this test — a false-red that is unrelated to the change under test. Reproduced byte-identically on clean `origin/main` (features `write-support,cli-helpers,state_machine,lz4`):
```
issue_953_multichunk_cell_seek.rs:176:31: Data.db must be present
```

## Fix (acceptance criteria)
- The test must **skip (eprintln + return)**, not panic, when the fixture directory is present but the `-Data.db` binary is absent — mirror the existing skip-on-absent-dir path. Extend the presence check to require the actual binary, not just the directory.
- Keep the memory rule intact: **0-rows-when-the-binary-IS-present remains a failure** (only genuine absence skips).
- Add/adjust a guard so a worktree gate against root datasets no longer false-reds `core-tests`.

## Routing
Oracle-ish test-hygiene fix; no OpenSpec. Found during #1439 delivery (the refactor's gate false-red on this test). Related to the active CI-failure cleanup.

_Filed by flow-lead worker during #1439._

Issue #1636, opened 2026-07-02
J2 — Collapse the three live ComparatorType value decoders into one (absorbs AJ3)


pmcfadin:
Part of epic #1603 (one decoder). Source: `docs/reports/parser-performance-audit-2026-07-01.md` finding J2 (audit block 2). Capstone §3: **J2 is the single owner of type-string-parser consolidation** — it absorbs platform AJ3 (the nested-type-split helper / 3rd-4th type-string parser in `schema/parser.rs`; scope note only, no separate AJ3 issue) and write-side R6 will cache parse results on top. Precondition: H4 lockstep nets. Sequence after J1.

## Problem
Three near-verbatim live `ComparatorType` **value decoders**:
1. `cqlite-core/src/storage/sstable/reader/parsing/value_parsing.rs` (block path)
2. `cqlite-core/src/storage/sstable/reader/parsing/comparator_value_parsing.rs`
3. `cqlite-core/src/schema/parser.rs` (schema-side copy)
plus scalar decode bodies duplicated against `cqlite-core/src/parser/types/` primitives. Every type fix must land in 3+ places or the paths silently diverge (H4 documents today's divergence set). There are also 3-4 type-STRING grammars (`CqlType::parse`, `ComplexTypeParser`, the ComparatorType from_data_type parser, schema/parser.rs helpers).

## Fix (effort: M)
1. **One value-decode module**: create/choose a single module (recommendation: keep `value_parsing.rs` as the home, or a new `parsing/comparator_decode.rs`) holding THE `decode(comparator: &ComparatorType, bytes: &[u8]) -> Result<Value>` implementation. Scalar arms must route through the canonical `cqlite-core/src/parser/types/` primitive codecs (post-I6, these are clean). Delete the duplicated bodies in the other two files; keep thin re-export shims only if call-site churn is prohibitive.
2. **One type-string grammar**: make `CqlType::parse` (find: `rg -n "fn parse" cqlite-core/src/schema/`) the single grammar for type strings; `ComparatorType::from_data_type` and `ComplexTypeParser` become adapters over it (parse once → convert), and `schema/parser.rs`'s nested-type-split helpers are deleted in favor of it (this is the absorbed AJ3 scope).
3. `partition_key_codec` (find with `rg -n "partition_key_codec" cqlite-core/src`) reuses the consolidated scalar decoders.
[...]

PR #1888, opened 2026-07-04
test(#1617): decoder + codec lockstep parity net (H4 — precondition of J1/J2)

### LINKED ISSUE #1617: H4 — Dispatch + codec lockstep parity tests (precondition of J1/J2)
Part of epic #1601 (parser measurement + safety net). Source: `docs/reports/parser-performance-audit-2026-07-01.md` finding H4 (audit block 2). **Hard precondition of J1 and J2 (epic #1603 children)** — the consolidation refactors must not land before these nets exist.

## Problem
There are (at least) two live decode implementations per CQL type — the v5 string-ladder path (`cell_value.rs` / `raw_value.rs` under `v5_compressed_legacy/`) and the `ComparatorType` enum path (`value_parsing.rs`) — plus the write side's type map (`cqlite-core/src/storage/serialization/types.rs`). Nothing asserts they agree. Divergences today surface later as parity bugs; and J1/J2 will merge these paths, which is only safe if equivalence is pinned FIRST.

## Fix (effort: S/M)
1. **Dispatch lockstep test** (new test file, e.g. `cqlite-core/tests/decoder_lockstep.rs` or a `#[cfg(test)]` module near `value_parsing.rs`): for every CQL type string in a canonical list (all scalars + representative list/set/map/tuple/UDT/frozen/nested), and for a set of byte buffers per type (valid encodings from fixtures + truncated + empty + oversized), decode the same bytes through BOTH paths:
   - v5 ladder entry: the per-cell decode used by `row_data.rs` (find with `rg -n "fn parse_cell_value" cqlite-core/src/storage/sstable/reader/parsing/v5_compressed_legacy/cell_value.rs`).
   - block path entry: `parse_value_with_schema_type` (`value_parsing.rs:440`).
   Assert: identical `Value` on success, and both-`Err` (not one-`Ok`-one-`Err`) on malformed input. Where they legitimately differ today (e.g. I4's Blob fallback), mark the case `#[should_panic]`/known-divergence with a comment naming the I-issue that fixes it — the test documents the divergence set explicitly.
2. **Read↔write codec lockstep**: for each type, encode a `Value` with the write side (`cqlite-core/src/storage/serialization/types.rs`) and decode with both read paths; assert round-trip identity. This pins the read decoders to the write-side truth (the write-side VInt/type modules are the exemplary reference per the audit).
3. Keep the canonical type list + byte-buffer corpus in one shared helper so J1/J2 reuse it.

### Do NOT
- Do not "fix" divergences found here inside this issue (except trivial test-only accommodations) — record each divergence in the test with a link to the owning issue (I4, J2). This issue's deliverable is the NET, not the fixes.

## Acceptance criteria
- [ ] Lockstep test covers every CQL type through both read paths, success + malformed cases.
- [ ] Read↔write round-trip lockstep covers every type.
- [ ] Known divergences are explicitly listed in-test with owning-issue links.
- [ ] `scripts/agent-gate.sh` PASS (paste the AGENT-GATE SUMMARY block verbatim).

## TDD test (must FAIL on current main, then be encoded as known-divergence)
The I4 case: bytes of a `float` column through the block path yield `Value::Blob(..)` while the v5 ladder yields `Value::Float(..)` — write this case first and watch it fail; then encode it as a documented divergence until I4 lands.

pmcfadin:
Closes #1617 (H4 — epic #1601 parser measurement + safety net). **Hard precondition of J1/J2 (#1636 J2 is board-Ready)** — pins decode-path equivalence BEFORE the consolidation refactors land.

## What this adds (test-only net)
`decoder_lockstep_tests.rs` pins agreement across the three live type codecs so J1/J2 can safely merge them:
- **Dimension A (dispatch):** for a canonical type corpus (all scalars + list/set/map/tuple/UDT/frozen/nested) × byte-buffer corpus (valid canonical / empty / truncated / flags-only), decode the SAME bytes through BOTH read paths (v5 string-ladder `parse_cell_value_schema_order` + block `parse_value_with_schema_type`); assert identical `Value` on success and both-`Err` on malformed.
- **Dimension B (read↔write codec):** encode via the write side (`serialization/types.rs`) and decode with both read paths; assert round-trip identity (pins the read decoders to the write-side truth).
- Shared corpus + helpers exposed `pub(crate)` for J1/J2 reuse.

## Documented divergences (recorded, NOT fixed — this issue's deliverable is the net)
Each pinned with `assert_ne` + owning-issue link, so the net fails loudly if a divergence silently changes or re-converges:
- `float`: block→`Float32` vs v5→`Float`(f64-widened) → **#1884**
- `varint`: block→`Varint` vs v5→`Blob` (no v5 arm) → **#1885**
- non-frozen list/set/map: v5 single-cell stub vs block VInt decode → #162 / J1
- frozen collections: write/v5 i32-BE framing vs block VInt framing → J2 (tuple converges — asserted equal)

## Quality bar
- Full `agent-gate.sh` (gate of record, 6cce12db): all change-attributable components PASS (incl. write-tests where dimension B runs). Overall `RESULT: FAIL` is solely the known/unrelated **#1856** (`issue_953` local-only `wide_table` Data.db) — the only failure in the run.
- roborev: **clean** ("No issues found"; 2 rounds — the net was also made fixture-independent so it runs in every CI lane, not just dataset-fetching ones: 7 run / 0 skipped with `CQLITE_DATASETS_ROOT` unset).
- Oracle-driven → no OpenSpec/C-audit.

Follow-ups filed: #1884, #1885 (the two v5-decoder correctness gaps the net surfaced).

### DIFF
--- a/cqlite-core/src/storage/sstable/reader/parsing/v5_compressed_legacy/decoder_lockstep_tests.rs
+++ b/cqlite-core/src/storage/sstable/reader/parsing/v5_compressed_legacy/decoder_lockstep_tests.rs
@@ -1,0 +1,790 @@
+//! Decoder lockstep parity net (issue #1617, Epic H / finding H4).
+//!
+//! CQLite has (at least) two live per-CQL-type value decoders plus the write
+//! side, and nothing asserts they agree. This module PINS that equivalence so
+//! the J1/J2 consolidation refactors (epic #1603) can merge the decoders safely.
+//! It is a **safety net**, not a fix: where the paths legitimately diverge TODAY
+//! the divergence is recorded here with an explicit assertion + owning-issue link
+//! (never silently tolerated), so the net documents the known-divergence set.
+//!
+//! ## The two read paths (and their DIFFERENT framing)
+//!
+//! * **v5 string-ladder** — [`V5CompressedLegacyParser::parse_cell_value_schema_order`]
+//!   (the per-cell decoder `row_data.rs` drives for BIG and BTI). It is a
+//!   *streaming* decoder that consumes a whole on-disk **cell**: `[flags byte]
+//!   [conditional timestamp/ttl VInts][framed value]`. Variable-width values
+//!   carry an unsigned-VInt length prefix; fixed-width values do not. It returns
+//!   the decoded value *and the new offset*.
+//! * **block / `ComparatorType` path** — [`SSTableReader::parse_value_with_schema_type`]
+//!   (`value_parsing.rs`). An *exact-slice* decoder: it receives only the value
+//!   bytes (the outer framing already stripped by the caller) and decodes them
+//!   whole. This is the SAME convention the **write side**
+//!   ([`TypeSerializer::serialize_value`], `serialization/types.rs`) emits — "no
+//!   length prefix for the top-level value" — which is why the write↔read
+//!   round-trip anchors naturally to the block path for scalars.
+//!
+//! Because the two paths use different framing, the lockstep is asserted at the
+//! **logical-value** level: for each type + logical value we produce the exact
+//! value bytes (the write-side / block convention) and *frame* them for the v5
+//! path ([`frame_v5_cell`]), then assert the decoded [`Value`]s agree (or that
+//! the divergence is documented).
+//!
+//! ## Two lockstep dimensions
+//! * **Dimension A — dispatch lockstep** (always compiled): the same logical
+//!   value through BOTH read paths agrees, and malformed input errors on BOTH
+//!   (never one-`Ok`-one-`Err`), for every scalar type + representative
+//!   frozen/collection/tuple.
+//! * **Dimension B — read↔write codec lockstep** (`write-support` only): the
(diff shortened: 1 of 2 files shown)


pmcfadin:
✅ **Merging (squash)** — test-only change, fully validated; every red is pre-existing/unrelated.

- Full `agent-gate.sh` (6cce12db): all change-attributable components PASS (incl. write-tests, where the lockstep net now runs unconditionally via the synthetic reader). Sole gate FAIL = known **#1856**.
- roborev: **clean** ("No issues found").
- CI: **18 lanes pass**, incl. the L2 `Compression / corruption parity (regenerate)`, Code Coverage, fmt/clippy/lib test, Parquet, dependency-isolation, PR smoke ×6.
- The 3 red lanes are pre-existing / CI-team-owned, red on any branch off main: **All/Minimal Compression Build** (#1866 scan_merge dead-code) and **sstabledump-parity** (#1242 dataset guard, red on all PRs).

Merging per owner merge-on-green directive. Follow-ups: #1884, #1885.

Issue #1860, opened 2026-07-04
test: issue_953_multichunk_cell_seek panics instead of skipping when local-only test_da/wide_table has only .jsonl (no Data.db)


pmcfadin:
Found during #1585 (E3) gate runs; pre-existing on main, not caused by that change.

## Problem
`cqlite-core/tests/issue_953_multichunk_cell_seek.rs` `real_wide_table_dir()` keys its skip on the fixture **directory** existing, but `test_da/wide_table-*/` is a **local-only** fixture that is present with only its `.jsonl` reference (no `da-2-bti-Data.db`) in a normal checkout. So instead of skipping, setup panics `Data.db must be present` (issue_953_multichunk_cell_seek.rs:~176) before any read path runs → `core-tests` FAILs the gate in a clean checkout.

This is the known 'local-only fixtures: skip on Data.db presence' hazard (cf. #1843 and the test_da/wide_table note) applied to this specific test.

## Fix direction (oracle-driven; the test's own skip logic is the oracle)
Gate the skip on the **Data.db file presence** (not the directory), matching the repo's established skip-on-presence pattern — but keep '0 rows when the Data.db IS present' a hard failure (do not weaken coverage when the binary exists in CI).

## Acceptance
- [ ] In a clean checkout lacking `test_da/wide_table/*Data.db`, the test SKIPs (does not panic).
- [ ] When the Data.db IS present, the test runs and asserts as today (0 rows = failure).
- [ ] `scripts/agent-gate.sh` core-tests PASS in a clean checkout.

pmcfadin:
🧭 **MANAGER** 
GO
HOLD: merge after #1859
Same defect (issue_953 multichunk cell-seek skip-on-missing-Data.db). #1859 is the primary — land it first, then reconcile/close this as covered if identical.

pmcfadin:
🧭 **MANAGER** 
GO
ORDER: 10

P0 gate-red: multichunk cell-seek panics instead of skipping — ranked ahead of the read-path perf epic. 0.14 policy: **P0 correctness/gate-green first**; the read-path epic runs in parallel on other workers (nothing held).

pmcfadin:
[...]

Issue #1803, opened 2026-07-03
agent-gate python-bindings component intermittently fails in the ephemeral venv (ModuleNotFoundError: cqlite._cqlite) — forces merge-over-red, erodes gate contract


pmcfadin:
Multiple sweep workers hit this: the agent-gate `python-bindings` component intermittently fails at pytest import with `ModuleNotFoundError: cqlite._cqlite` even though the wheel builds and code compiles cleanly. It's a maturin editable-install / venv-resolution issue in the gate's ephemeral venv (`target/agent-gate-venv`), not a source defect (matches the known-gate-failures note; also a PEP604/py3.9 variant seen before).

Impact: when it triggers, the AGGREGATE gate RESULT is FAIL despite the change being green, so workers merge over it under the 'proven pre-existing/unrelated' clause. This NORMALIZES merge-over-red and erodes the gate-is-the-only-run-that-counts contract. Observed on #1562 (and intermittently elsewhere; other runs had python-bindings PASS).

Fix: make the gate's python-bindings step deterministic — pin the venv build (`maturin develop` into a clean venv with the right interpreter, verify `import cqlite._cqlite` before running pytest, fail loudly with a distinct message if the module truly didn't build vs a venv-resolution miss). AC: python-bindings component passes reproducibly on a clean checkout with fixtures present, or fails only on a real binding-code defect. Related: #1737 (gate perf/reliability).

pmcfadin:
🧭 **MANAGER** 
GO
ORDER: 8

P0 gate-flaky: agent-gate python-bindings ModuleNotFoundError — ranked ahead of the read-path perf epic. 0.14 policy: **P0 correctness/gate-green first**; the read-path epic runs in parallel on other workers (nothing held).

pmcfadin:
Claimed by flow-lead (branch `issue-1803-python-bindings-venv-determinism` pushed = lock; board→In Progress). Fixing the gate's python-bindings venv step for determinism (clean venv, verify `import cqlite._cqlite` pre-pytest, distinct fail message on real defect vs venv miss). Oracle-driven tooling fix — no OpenSpec.

pmcfadin:
[...]

Issue #1516, opened 2026-07-02
Epic D — Streaming by default: pushdown + bounded memory


pmcfadin:
**Epic D — Streaming by default: pushdown + bounded memory** (2026-07 read-path audit, block 1)

The common query path materializes when it should stream. `LIMIT 10` on a 1M-row table decodes all 1M rows then truncates (the reader's `scan()` accepts a limit — the executor always passes `None`); `COUNT(*)` buffers the whole table; any table with >1 SSTable generation silently collapses the bounded-streaming guarantee (collect + sort the entire reconciled result before row 1); the CLI `query` command slurps the whole Data.db into RAM; and `MAX_RESULTS = 1M` converts big-but-legal queries into hard errors. The streaming executor already does all of this correctly — this epic makes the default paths behave like it.

Source of truth: `docs/reports/read-path-performance-audit-2026-07-01.md` §Epic D. Capstone §4: **Wave 3**, streaming-by-default story (D1/D2/D3/D6 → D4 → D5). **DECIDED #10 (owner, 2026-07-01): D4 writes the ordering oracle test vs Cassandra FIRST, then fixes to the verdict.**

## Children (tracking)
- [ ] #1577 D1 — push LIMIT/OFFSET into the materializing scan `P1`
- [ ] #1578 D2 — streaming O(1) aggregates + retire the MAX_RESULTS cliff `P1`
- [ ] #1579 D3 — stream the multi-generation merge (no collect+sort) `P1`
- [ ] #1580 D4 — manager scan ordering: oracle vs Cassandra FIRST, then k-way merge `P1`
- [ ] #1581 D5 — CLI query off BulletproofReader → Database/scan_stream `P1`
- [ ] #1582 D6 — byte-bounded result budget (replaces the row-count guard) `P2`

## Rules for implementers
- **Result parity is the law:** the current materializing path is the oracle for every child — identical rows, order, and values before/after (plus JSONL goldens).
- Tombstone/overwrite reconciliation lives in the KWay merge — never reimplement it; only remove staging.
- Semantic safety on pushdown: no early-exit under Sort/Aggregate; count accepted (post-predicate) rows.
- TDD red-then-green (dhat ceilings from A4 prove the memory wins); gate SUMMARY verbatim; `-D warnings`; no `unwrap()`/`expect()`; 33-table parity green.

pmcfadin:
[...]

PR #1890, opened 2026-07-04
feat(#1582): D6 — byte-bounded result budget (typed ResultTooLarge + max_result_bytes knob)

### LINKED ISSUE #1582: D6 — byte-bounded result budget (replace the row-count guard; stream unbounded SELECTs)
Child **D6** of Epic D (#1516) — 2026-07 read-path audit (block 1). Source of truth: `docs/reports/read-path-performance-audit-2026-07-01.md` §Epic D. Anchors as of `main` @ `5c080d2a`. Capstone wave: **Wave 3** (with D2).

## Problem
The memory guard on materializing results is a **row count** (`MAX_RESULTS = 1M`), which is the wrong unit: 1M skinny rows may be fine while 100k wide rows blow the <128MB target. And unbounded SELECTs through CLI/bindings default to the materializing executor instead of the streaming one.

## Fix (step by step)
1. Add a **byte ceiling** to materializing result collection: maintain a running estimate using the existing `estimate_row_size` (`rg -n "estimate_row_size"`), and stop with a clear, typed error when the budget is exceeded (error message must say what to do: add LIMIT / use streaming).
2. Budget default: derive from the <128MB project target (document the chosen number); make it a real config knob wired to behavior (set-knob-assert-behavior-changes test — no decorative knobs).
3. Keep the row-count check only as a secondary safety valve per D2.
4. Default unbounded SELECTs in CLI and bindings to the **streaming executor** where the call shape allows (CLI already streams post-D5; for bindings, route `execute()`-style unbounded SELECTs through the streaming path internally where the public contract permits — do NOT break public API semantics; if a binding's contract requires a materialized list, the byte budget is the protection there).
5. Coordinate with D2 (this replaces the cliff) and A4 (budget test lives in that lane).

## TDD tests (write FIRST, must fail on current main)
- Wide-row fixture (e.g. `test_wide_rows`) trips the BYTE guard before the row-count guard (fails today: only row count exists).
- Skinny-row fixture with > budget rows but < budget bytes under LIMIT-less streaming completes fine.
- Knob test: lowering the budget knob makes a previously-passing query trip the guard.

## Guardrails
- Typed error via `thiserror`, mapped through bindings' error taxonomy (coordinate with the X2 shared error table if it has landed).

## Definition of done
- [ ] TDD tests red-then-green; `scripts/agent-gate.sh` PASS (paste SUMMARY verbatim)
- [ ] `RUSTFLAGS="-D warnings"` clean; no `unwrap()`/`expect()` in library code

pmcfadin:
Part of #1582 (D6 — Epic D #1516). **Does NOT fully close #1582** — the byte guard here is a fail-fast ceiling, not the true peak-memory bound; that remains open on #1582, blocked on #1897 (scan_stream single-gen rearchitecture). Owner-approved partial merge.

Delivers (verified):
- **Byte fail-fast ceiling** on materializing result collection via `enforce_result_budget`, reusing the existing `estimate_row_size`/`estimate_value_size` (no new estimator). Typed `Error::ResultTooLarge` (thiserror; message says add LIMIT / use streaming), classified `Query`, mapped through Python (`QueryError`) + Node (`QUERY`) taxonomy.
- **LIMIT/OFFSET propagation** (`compute_scan_collect_bound`): a LIMITed query stops collecting at the limit and no longer trips the budget on rows beyond it (authoritative plan inspection — only when post-scan steps are exclusively Limit/Project; no heuristic).
- **Config knob** `QueryConfig.max_result_bytes` (default `DEFAULT_MAX_RESULT_BYTES = 64 MiB`, documented derivation), load-bearing (set-knob-assert-behavior test), with `#[serde(default)]` for backward-compat (regression test deserializes a config missing the field).
- **Gate regression net:** new `byte-budget-guard` agent-gate component runs the TDD tests (`write-support,cli-helpers,state_machine`) — they previously never executed in any gate/CI lane.
- Row-count check kept as secondary safety valve. Python/Node `execute()` routing unchanged (materializing + byte-bounded, per conservative guidance).

**Known limitation (tracked, NOT merged-over silently):** does not bound the underlying per-reader scan materialization → #1582 stays open for the true peak bound once #1897 (scan_stream single-gen 0-rows) is fixed.

Gate: all correctness nets PASS incl. the new `byte-budget-guard`; sole FAILs are pre-existing #1860 + #1803 flakes. `file-size`: pre-oversized files grew → documented `CQLITE_ALLOW_FILE_GROWTH=1` (#1116).

### DIFF
--- a/bindings/node/src/error.rs
+++ b/bindings/node/src/error.rs
@@ -449,6 +449,11 @@ mod tests {
                 Error::QueryExecution(_) => {
                     assert_eq!(err.category(), ErrorCategory::Query);
                 }
+                // Byte-bounded result budget (issue #1582): Query category,
+                // maps to the "QUERY" JS code via category().
+                Error::ResultTooLarge { .. } => {
+                    assert_eq!(err.category(), ErrorCategory::Query);
+                }
                 Error::UnsupportedQuery(_) => {
                     assert_eq!(err.category(), ErrorCategory::Query);
                 }
--- a/bindings/python/src/error.rs
+++ b/bindings/python/src/error.rs
@@ -74,10 +74,12 @@ pub fn to_py_err(err: cqlite_core::Error) -> PyErr {
             SchemaError::new_err(message)
         }
 
-        // Query execution errors -> QueryError
-        cqlite_core::Error::QueryExecution(_) | cqlite_core::Error::UnsupportedQuery(_) => {
-            QueryError::new_err(message)
-        }
+        // Query execution errors -> QueryError. `ResultTooLarge` (issue #1582,
+        // byte-bounded result budget) is a query-shaped error whose message
+        // tells the user to add LIMIT or stream, so it maps to QueryError too.
+        cqlite_core::Error::QueryExecution(_)
+        | cqlite_core::Error::ResultTooLarge { .. }
+        | cqlite_core::Error::UnsupportedQuery(_) => QueryError::new_err(message),
 
         // CQL parsing errors -> ParseError
         cqlite_core::Error::CqlParse(_) => ParseError::new_err(message),
@@ -369,6 +371,7 @@ mod tests {
                 Error::Schema(_) => { /* Maps to SchemaError */ }
                 Error::Table(_) => { /* Maps to SchemaError */ }
                 Error::QueryExecution(_) => { /* Maps to QueryError */ }
+                Error::ResultTooLarge { .. } => { /* Maps to QueryError (issue #1582) */ }
                 Error::UnsupportedQuery(_) => { /* Maps to QueryError */ }
                 Error::CqlParse(_) => { /* Maps to ParseError */ }
                 Error::Configuration(_) => { /* Maps to PyValueError */ }
--- a/cqlite-core/src/config.rs
+++ b/cqlite-core/src/config.rs
@@ -375,6 +375,17 @@ fn default_max_result_bytes() -> u64 {
     DEFAULT_MAX_RESULT_BYTES
 }
 
+/// Serde default for [`QueryConfig::max_result_rows`] (issue #1582).
+///
+/// Backward-compat + robustness: a `QueryConfig` serialized without this key
+/// (or a partial JSON/dict config) still deserializes, taking the shipped
+/// 1,000,000-row secondary safety valve rather than failing with a missing
+/// field. Keeps the knob real (not decorative) and consistent with
+/// [`default_max_result_bytes`].
+fn default_max_result_rows() -> u64 {
+    1_000_000
+}
+
 /// Query engine configuration
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct QueryConfig {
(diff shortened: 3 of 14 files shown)


pmcfadin:
🧭 **MANAGER**  — heads-up: this PR is **red on a real build failure** (not a flaky lane), so it should NOT merge as-is — it would break `main`.

**`Minimal Compression Build` fails** (`cqlite-core`, lz4+snappy-only config):
```
error: static `MANAGER_SCAN_FULL_SORTS` is never used
  --> cqlite-core/src/storage/sstable/scan_merge.rs:48:19
error: function `sort_by_token_order` is never used
  --> cqlite-core/src/storage/sstable/scan_merge.rs:204:15
  = note: `-D dead-code` implied by `-D warnings`
error: could not compile `cqlite-core` (lib) due to 2 previous errors
```
Both symbols are only referenced behind a feature/cfg that is **not active in the minimal build**, so under `-D warnings` they are dead code → hard error. `All Compression Build & Test` and `sstabledump-parity` are also failing — worth confirming those arent downstream of the same issue.

**Fix:** gate `MANAGER_SCAN_FULL_SORTS` + `sort_by_token_order` behind the same `cfg`/feature that actually uses them (or `#[cfg(...)]` the whole path), so the minimal build has no dead code; then re-run the FULL `scripts/agent-gate.sh` (its minimal-features component reproduces this locally) and merge on green.

Not merging from here — flagging for the owning team.

pmcfadin:
⏸️ **Holding — not merging.** Owner flagged a cross-session collision on #1582 (another local session used slug \`issue-1582-d6-byte-budget\` with a flow-address artifact; that branch is NOT on origin — this PR is the only #1582 on origin) and asked me to stop after wrapping up. Leaving this for owner adjudication.

State of this PR (narrow subset, owner-approved):
- Reduced to a single **final-result** byte/row budget check (dropped the fragile during-collection LIMIT/OFFSET/predicate machinery that generated 4 roborev rounds); also covers the legacy point-lookup path and the constant (no-FROM) SELECT path (`3255ef16`).
- Config: `max_result_bytes` (64 MiB default) + `max_result_rows`, both `#[serde(default)]` (backward-compat); typed `ResultTooLarge` mapped through Python/Node taxonomy; `byte-budget-guard` gate component added.
- **Known scope boundary (NOT in this PR):** does not bound *peak* scan memory — deferred to #1897 (scan_stream single-gen rearchitecture). #1582 stays OPEN for that.
- **Needs a rebase** onto current origin/main (branch is ~21 commits behind; picks up the #1903 write-throughput flake fix).
- Last full gate: only FAILs were `file-size` (needs the documented `CQLITE_ALLOW_FILE_GROWTH=1` ack) and `core-tests` = the known `test_write_throughput` load flake (fixed on main via #1903) + `block_source_reports_backend` (verified byte-identical to origin/main; environmental sccache/temp-path race, not this diff). `byte-budget-guard`, python, node, tooling all PASS.

Not merging; awaiting owner decision on the collision. — flow-lead worker

Issue #2003, opened 2026-07-05
CLI smoke: remaining simple_table LIMIT tests (csv/table/columns) also need ORDER BY (follow-up to #1997)


pmcfadin:
Follow-up to #1997. After #1998 fixed `test_select_json_simple`, the smoke script (which stops at first failure) exposed 3 SIBLING tests with the identical order-unstable pattern — `SELECT ... FROM test_basic.simple_table LIMIT N` with no ORDER BY on a single-PK table: **test_select_csv_simple** (LIMIT 3), **test_select_table_simple** (LIMIT 2), **test_select_columns** (id,name LIMIT 3). Storage/token order drifts vs the id-sorted goldens. My #1998 fix should have swept all of them.

Fix: add `ORDER BY id` to all 3. Verified each is byte-identical to its existing committed golden (select_simple_csv/select_simple_table/select_columns_json) — no golden regen. The no-LIMIT full-scan tests are run_error_test (exit-code, not order-sensitive). Verified via a full local smoke run (0 failures).

pmcfadin:
Fixed by #2005 (merged): all remaining order-unstable smoke LIMIT queries now use ORDER BY id; full local smoke 11 PASS / 0 FAIL.

Issue #1997, opened 2026-07-05
CLI smoke test_select_json_simple flaps: unordered LIMIT on single-PK table drifts vs golden (add ORDER BY id)


pmcfadin:
## Problem (main CI "CLI tests and smoke" red)
The one-shot smoke test `test_select_json_simple` fails its snapshot comparison on `main`:
```
[FAIL] test_select_json_simple: Output does not match snapshot
--- select_simple_json.golden vs test_select_json_simple.actual
```

## Root cause — order-unstable query, NOT a #1952 regression and NOT a stale golden value
The smoke query is `SELECT * FROM test_basic.simple_table LIMIT 3` (no `ORDER BY`). `simple_table` is single-PK (`id UUID PRIMARY KEY`, no clustering). An unordered `LIMIT` returns rows in **storage/token order**, which the engine legitimately changed at some point (Cassandra-consistent). The golden was captured id-sorted (Dec-2025). Column set/order/formatting are byte-identical — only WHICH 3 rows (and their order) differ. Verified: the golden's ids and the actual ids are BOTH present, unchanged, in the same `simple_table` JSONL reference. Not #1952 (bare `SELECT *` keeps empty projection = scan-all, untouched by that change).

## Fix (one line, deterministic, no golden regen)
Add `ORDER BY id` to the smoke query: `SELECT * FROM test_basic.simple_table ORDER BY id LIMIT 3`. CQLite supports the in-memory sort; the output is then deterministic and — verified — **byte-identical to the existing committed golden** (so no golden regeneration is needed). This future-proofs the smoke against storage-scan-order drift.

## Acceptance criteria
- `ci-one-shot-smoke.sh` `test_select_json_simple` query includes `ORDER BY id`.
- Smoke test passes against the unchanged `select_simple_json.golden`.
- main CI "CLI tests and smoke" job green.

## Routing
CI/test-harness (order-unstable smoke query), no engine change. P2 (part of unblocking main CI for 0.13).

PR #2005, opened 2026-07-05
fix(#2003): deterministic ORDER BY id on all remaining CLI smoke LIMIT queries (follow-up to #1998)

pmcfadin:
Follow-up to #1997/#1998. That PR fixed only `test_select_json_simple`; the smoke script stops at the first failure, masking 4 SIBLING tests with the identical order-unstable pattern (`SELECT ... LIMIT N` with no ORDER BY on single-PK tables): `test_select_csv_simple`, `test_select_table_simple`, `test_select_columns` (simple_table), and `test_select_collections` (collection_table, PK `id UUID`).

Fix: add `ORDER BY id` to all of them. Each verified **byte-identical to its committed golden** (no regen; set/map element order in collections also matches). **Full local smoke run (CI env): 11 PASS, 0 FAIL — ALL TESTS PASSED.** This is the comprehensive sweep #1998 should have been.

### DIFF
--- a/test-data/scripts/ci-one-shot-smoke.sh
+++ b/test-data/scripts/ci-one-shot-smoke.sh
@@ -336,7 +336,7 @@ run_test_suite() {
     # Test 2: Basic SELECT with CSV output (simple_table)
     run_test \
         "test_select_csv_simple" \
-        "SELECT * FROM test_basic.simple_table LIMIT 3" \
+        "SELECT * FROM test_basic.simple_table ORDER BY id LIMIT 3" \
         "csv" \
         0 \
         "${SNAPSHOTS_DIR}/select_simple_csv.golden"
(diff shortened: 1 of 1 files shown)


pmcfadin:
roborev findings reviewed — both are verification requests that a full local smoke run (exact CI env: CQLITE_DATA_DIR/CQLITE_SCHEMA/CQLITE_CLI/GOLDEN_DIR) already answers empirically:
- **Medium (goldens match ORDER BY output):** verified — each ORDER BY id output diffs **byte-identical** to its committed golden (select_simple_csv/select_simple_table/select_columns_json/select_collections_json); no regeneration needed. Full smoke: **11 PASS, 0 FAIL — ALL TESTS PASSED ✓**.
- **Low (collection_table key is id):** confirmed — schema `collection_table (id UUID PRIMARY KEY, ...)`; the ORDER BY id query runs exit 0 and matches its golden.
The goldens were correct all along (captured id-sorted); the queries just lacked deterministic ordering. Merging.

Issue #841, opened 2026-06-18
Flight/Trino: aggregation pushdown (count/sum/min/max/avg) via a single finalize split — evaluate benefit first


rustyrazorblade:
Follow-up to #836 (Arrow Flight server + Trino connector), to be picked up after it merges.

## Idea
Push `count`/`sum`/`min`/`max`/`avg` down to the cqlite-flight server so it computes aggregates **during the compaction merge** (it already scans every reconciled row) and returns ~one row per group instead of the full row set — a large reduction in data shipped to Trino. Aggregates are computed **post-merge** (on LWW/tombstone-reconciled rows) so semantics match `SELECT`.

## Constraint we verified
Trino does **not** re-aggregate across splits. `PushAggregationIntoTableScan` (the rule behind `ConnectorMetadata.applyAggregation`) rewrites the plan to `ProjectNode(TableScanNode)` and **removes the `AggregationNode` entirely** — no final/second-stage aggregation, no multi-split partial handling. So returning per-split partials would surface as separate, un-merged rows (wrong). The connector must return the **fully merged** result.
(Source: https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/PushAggregationIntoTableScan.java)

## Chosen design — "single finalize split" (option 2)
- Implement `ConnectorMetadata.applyAggregation`; translate Trino `AggregateFunction`s + grouping sets into an aggregation spec carried in the Flight ticket.
- For an aggregated table handle, `getSplits` returns **one split**, whose `PageSource` fans out to all replica cqlite-flight endpoints, pulls each node's partial, **merges**, and emits the final row(s).
- Server computes partials during the merge: `count`, `sum`, `min`, `max`, and `sum`+`count` for `avg`.
- Combine rules: count→Σ, sum→Σ, min/max→min/max, avg→Σsum/Σcount.
- Trade-off: the final merge runs on a single Trino worker (lost intra-aggregation parallelism), but per-replica payloads are tiny (≈one row per group per node).

## Scope / non-goals
- Supported: count, sum, min, max, avg; simple `GROUP BY`.
- Not pushable: `count(DISTINCT …)` (not combinable across splits without shuffling), `ROLLUP`/`CUBE`/`GROUPING SETS`, expressions in args (`sum(a*b)`), coercions, ordered/filtered aggregates. Leave these to Trino.

## CAVEAT — evaluate the benefit before building
[...]

Issue #2113, opened 2026-07-07
Follow-up nits from epic #2103 review (PR #2112): table_stats live-dir in snapshot mode, CHM-lock scope, read-mode kit arg


pmcfadin:
Batched NIT-severity findings from the adversarial review of PR #2112 (epic #2103) plus one cross-read finding, per the roborev-severity batching rule. None block correctness.

**N4 — `table_stats` fetched from the live dir even in snapshot mode.** `CqliteFlightMetadata.tableStatsRequest` sends `snapshot=null`, so the optimizer sees live-dir row counts while the scan reads the snapshot. Plan-quality only (row-count estimate / GROUP BY gate), never result correctness — and the snapshot doesn't exist yet at planning time, so fixing this needs a design decision (create-earlier vs accept-estimate). Document or fix.

**N5 — snapshot PUT inside `ConcurrentHashMap.computeIfAbsent`.** Network I/O (up to 30s) under a CHM bin lock in `SnapshotManager`. Intentional per-(query,table) memoization, contention limited to concurrent planners of the same key; documented in-code. Consider a per-key future/lock pattern if it ever shows up in profiles.

**Cross-read (from #2108 authoring) — `cqlite.read-mode` is not an install-time arg of the `trino-cqlite` kit.** The test plan sets it by hand-editing the rendered catalog properties + re-running `update-catalogs.sh` (same workaround the kit uses for `local-datacenter`). Add `--read-mode` (and possibly `--local-datacenter`) to `kit.yaml` args + the catalog template so Phase-3b (live-mode weak-spot hunt) is a one-flag rerun.

Context: review verdict on PR #2112 was 0 blockers / 5 nits; N1–N3 were fixed pre-merge in `3270f403`.

— flow-lead (epic #2103 endgame)

pmcfadin:
Nit from PR #2134 (round-2 harness batch) review: `trino-loadtest` `TRINO_LOADTEST_KEYSPACE`/`TRINO_LOADTEST_TABLE` use `${VAR:-}` inline substitution, so an unset optional arg injected as literal `null` by the kit runner wouldn't crash (the #2132 class) — it would silently target a keyspace/table literally named "null" at query time. Low priority: both args are effectively required for a useful run; worth a treat-null-as-unset normalization pass when this batch is picked up.

pmcfadin:
### Nits batched from PR #2135 (#2129 — LIMIT pushdown)

[...]

PR #2135, opened 2026-07-07
feat(#2129): LIMIT pushdown — bounded per-split scans (connector applyLimit → ticket limit → do_get early stop)

### LINKED ISSUE #2129: cqlite connector: no LIMIT/count pushdown → SELECT ... LIMIT 5 full-scans ~2M rows (235s, then 'Network closed')
Found during the first live run of the #2103 harness. **Performance finding — the read path is correct but does full-table work for trivially-bounded queries.** This is the kind of weak spot the harness was built to surface, so filing with the evidence.

## Setup

- 3× i4i.xlarge Cassandra 5.0 nodes, `cassandra_easy_stress.keyvalue`, ~2M rows, 3 flushed SSTables/node.
- Trino 481, 1 worker (m5.xlarge), cqlite connector 0.13.1, flight v0.13.0, default `snapshot` read mode.

## Symptom

Even the most trivially-bounded query is extremely slow:
- `SELECT * FROM cqlite.<ks>.<tbl> LIMIT 5` — ran **235s** on an idle single worker, then the client returned `Network closed for unknown reason`.
- `SELECT count(*) FROM cqlite.<ks>.<tbl>` — minutes, never observed to complete in the test window.
- Worker stayed healthy throughout (0 restarts, ~788Mi, ~17m CPU) and the flight pods logged no errors — so this is not a crash/OOM; it's the connector doing full-scan work.

## Root cause — no LIMIT pushdown; every query full-scans + merges all SSTables

`EXPLAIN SELECT * FROM cqlite.<ks>.<tbl> LIMIT 5`:

```
Limit[count = 5]
  LimitPartial[count = 5]
    TableScan[table = cqlite:CqliteFlightTableHandle[keyspace=..., table=keyvalue, ...]]
        Estimates: {rows: 1990586 (208.82MB), ...}
```

The `Limit`/`LimitPartial` sit **above** the `TableScan`; nothing is pushed into the connector. The scan is estimated at the full **1,990,586 rows / 208 MB**. So `LIMIT 5` makes each of the (16 vnodes × N-node) token-range splits pull its entire range from `cqlite-flight` — which runs a full k-way compaction-merge of that range's SSTables — and Trino then throws away all but 5 rows.

The connector's own README already notes predicate pushdown is deferred (Trino post-filters). This finding is that **LIMIT (and count(*)) get no connector-side shortcut either**, so *every* interactive query pays full-table compaction-merge cost. On a real dataset that's minutes-to-unusable for the exploratory `LIMIT`/`count` queries a user reaches for first.

## Impact

- Makes the connector feel hung for basic sanity queries (Step 12 of the test plan — the count parity check — could not complete in-window).
- The 235s `LIMIT 5` ending in `Network closed for unknown reason` suggests a client/coordinator idle or response timeout fires before the full merge finishes — worth confirming whether that's a Trino-side timeout vs a flight `do_get` stream timeout (separate potential bug; captured here as the trigger).

## Suggested work

- **LIMIT pushdown**: `applyLimit` in the connector so the flight ticket carries a row cap and `do_get` can stop early (at least per-split), turning `LIMIT k` into bounded work.
- **count(*)**: consider an aggregation/count pushdown path (the connector already has aggregation-pushdown plumbing per #893) so `count(*)` doesn't stream every row to Trino.
- Investigate the `Network closed for unknown reason` at ~235s — is a timeout being hit? If so, which side?

## Repro

1. Load a non-trivial table (≈2M rows), flush.
2. `EXPLAIN SELECT * FROM cqlite.<ks>.<tbl> LIMIT 5` → Limit is not pushed into TableScan; scan estimate = full table.
3. Run the query → minutes; may fail `Network closed for unknown reason`.

Note: this is a *performance/pushdown* issue, not a correctness one — the rows that are read/merged are correct; there's just no bounding. Filed separately from the observability (#2128) and kit bugs.

pmcfadin:
Closes #2129

## What

LIMIT pushdown end to end, so `SELECT ... LIMIT k` becomes bounded per-split work instead of a full range scan (the reported ~235s `LIMIT 5` over ~2M rows).

**Connector (Java):** `ConnectorMetadata.applyLimit` records an `OptionalLong limit` on `CqliteFlightTableHandle`; the non-aggregated split's Flight ticket stamps a `limit` field. Returns `LimitApplicationResult(handle, limitGuaranteed=false, precalculateStatistics=false)` — each split caps independently, so Trino keeps its global `Limit`/`LimitPartial` above the scan (always correct). `applyFilter` preserves the cap when it rebuilds the handle (order-independent). Declines on aggregated handles; idempotent re-application returns empty (a strictly tighter limit still pushes).

**Ticket protocol:** additive-optional `limit: Option<u64>` (`#[serde(default)]`). Connector omits the field when empty.

**Flight server (Rust):** `drive_merge` stops the k-way merge after `limit` rows, counted AFTER token pruning and predicate filtering (filtered-out rows never consume the cap); `limit == Some(0)` emits nothing without stepping the merge. The aggregation path never consults `limit` by construction.

## Ticket-compat posture

The Rust `FlightTicket` is NOT `deny_unknown_fields`, so the field is safe in both skew directions and I chose additive-optional (not fail-closed): an old server silently ignores `limit` from a new connector (correct full scan, just unbounded); an old connector's ticket defaults `limit` to `None` on a new server. Skew only forfeits the speedup, never returns wrong rows. Proven by `unknown_field_is_ignored_not_rejected` in `ticket.rs`.

## count(*) finding — already pushes down; NOT version skew

`SELECT count(*)` already pushes down on current main via `applyAggregation` (issue #841/#893) — test `CqliteFlightMetadataApplyAggregationTest.pushesCountStarGlobal` (`trino-connector/.../CqliteFlightMetadataApplyAggregationTest.java:62`) proves the ticket carries `{"func":"Count","column":null}`. The server side (`cqlite-flight/src/agg.rs`, added by #841 on 2026-06-20) is present at the `v0.13.0` tag (verified: `git cat-file -e v0.13.0:cqlite-flight/src/agg.rs` succeeds; the ticket `aggregation` field is present at that tag). So the live run's slowness was **not** connector/flight version skew — both 0.13.1 connector and v0.13.0 flight image support count pushdown.

The remaining slowness is inherent: counting reconciled live rows requires a full, tombstone/LWW-aware k-way merge-scan server-side (Statistics.db per-SSTable counts are pre-merge estimates, not the reconciled live-row count, so there is no metadata shortcut under snapshot read-mode). Pushdown already avoids streaming 2M rows to Trino; it cannot avoid reading them. **No duplication needed here** — a true fast count would be a separate approximate/stats-based feature.

## "Network closed at ~235s" note (finding, not a fix)

The connector sets **no deadline** on the `do_get` scan stream — `CqliteFlightClient.getStream` (`CqliteFlightClient.java:88`) passes no `CallOptions.timeout`; the only Flight deadline is the 3s planning-time `table_stats` DoAction. So the ~235s cut is **not** from the connector's Flight client. `MergeProducer` materializes all batches before streaming (Phase 1, `producer.rs`), so zero bytes flow to Trino for the entire merge — the most likely trigger is a no-data-flowing idle timeout on the long DoGet stream (gRPC/Flight infra between worker and flight pod, a k8s/LB idle timeout, or Trino's client-facing `query.client.timeout`). Confirming the exact timer needs coordinator/proxy logs. LIMIT pushdown removes the trigger by bounding the merge.

## Tests

- Rust (`cargo test -p cqlite-flight`): **135 pass, 0 fail** (8 new). New: ticket parse with/without limit + limit=0 + unknown-field ignored; `ScanSpec` carries limit; `drive_merge` early stop for limit<rows, limit>rows, limit=0, and limit-with-filter (filtered rows do not consume the cap).
- Java (`./gradlew test`): **187 pass, 0 fail**. New: `applyLimit` handle plumbing + `limitGuaranteed=false`, ticket JSON carries `limit`, idempotent re-application, decline on aggregated handle, limit survives `applyFilter`; `FlightTicketJson` emits/omits `limit`.

## file-size ratchet

`ticket.rs`, `filter.rs`, `producer.rs` were already over the 800-line src threshold before this change (tracked by splitting epics #1116/#1135). Splitting a 2397-line file is out of scope for a LIMIT feature, so the gate was run with `CQLITE_ALLOW_FILE_GROWTH=1` per the campsite-rule escape hatch.

### DIFF
--- a/cqlite-flight/src/filter.rs
+++ b/cqlite-flight/src/filter.rs
@@ -197,6 +197,10 @@ pub struct ScanSpec {
     pub filter: Option<FilterExpr>,
     /// Column projection; `None` emits all columns.
     pub projection: Option<Vec<String>>,
+    /// Row cap (issue #2129): emit at most this many rows for the scan, counted
+    /// AFTER token pruning and predicate filtering. `None` scans the full range.
+    /// Ignored on the aggregation path (partial rows already collapse the set).
+    pub limit: Option<u64>,
 }
 
 impl ScanSpec {
@@ -240,6 +244,7 @@ impl ScanSpec {
             token,
             filter,
             projection: ticket.columns.clone(),
+            limit: ticket.limit,
         })
     }
 }
@@ -581,6 +586,31 @@ mod tests {
     fn no_bounds_means_no_token_filter() {
         let spec = ScanSpec::from_ticket(&ticket_with(vec![]), &simple_schema()).unwrap();
         assert!(spec.token.is_none());
+    }
+
+    #[test]
+    fn limit_is_carried_from_ticket() {
+        // No limit on the ticket → None on the spec.
+        let spec = ScanSpec::from_ticket(&ticket_with(vec![]), &simple_schema()).unwrap();
+        assert_eq!(spec.limit, None);
+
+        // A ticket limit flows straight through (including the 0 edge).
+        let t = FlightTicket {
+            limit: Some(5),
+            ..ticket_with(vec![])
+        };
+        assert_eq!(
+            ScanSpec::from_ticket(&t, &simple_schema()).unwrap().limit,
+            Some(5)
+        );
+        let t = FlightTicket {
+            limit: Some(0),
+            ..ticket_with(vec![])
+        };
+        assert_eq!(
+            ScanSpec::from_ticket(&t, &simple_schema()).unwrap().limit,
+            Some(0)
+        );
     }
 
     #[test]
--- a/cqlite-flight/src/producer.rs
+++ b/cqlite-flight/src/producer.rs
@@ -1404,7 +1404,11 @@ mod tests {
 
         let p = MergeProducer::with_spec(schema, 4, spec_with_limit(Some(100))).unwrap();
         let batches = p.produce(&DirSource::new(&dir)).unwrap();
-        assert_eq!(total_rows(&batches), 10, "LIMIT past the row count keeps all");
+        assert_eq!(
+            total_rows(&batches),
+            10,
+            "LIMIT past the row count keeps all"
+        );
     }
 
     #[test]
--- a/cqlite-flight/src/producer.rs
+++ b/cqlite-flight/src/producer.rs
@@ -480,15 +480,30 @@ impl MergeProducer {
     /// `merger.step()`, so a cancel (e.g. client disconnect) stops the merge
     /// before collecting/reconciling the next partition — never after performing
     /// one more potentially large partition merge.
+    ///
+    /// LIMIT pushdown (issue #2129): when the scan carries a row cap
+    /// ([`ScanSpec::limit`](crate::filter::ScanSpec)), the merge stops as soon as
+    /// `limit` rows have been EMITTED. The cap is counted AFTER the token filter
+    /// and the predicate filter, so partitions outside the range and rows the
+    /// filter rejects never consume it — a filtered scan returns as many matching
+    /// rows as exist up to `limit`, never fewer. `limit == Some(0)` emits nothing
+    /// without stepping the merge at all. The cap applies per split; the connector
+    /// sets `limitGuaranteed = false` so Trino keeps a global `Limit` above.
     fn drive_merge(
         &self,
         merger: &mut dyn PartitionStepper,
         cancel: &CancelFlag,
         batches: &mut Vec<RecordBatch>,
     ) -> Result<(), ProducerError> {
+        let limit = self.spec.limit;
+        // A zero cap produces no rows without touching the merge.
+        if limit == Some(0) {
+            return Ok(());
+        }
         let mut buffer: Vec<QueryRow> = Vec::with_capacity(self.batch_size);
+        let mut emitted: u64 = 0;
 
-        loop {
+        'partitions: loop {
             if cancel.is_cancelled() {
                 return Err(ProducerError::Cancelled);
             }
@@ -518,8 +533,15 @@ impl MergeProducer {
                     }
                 }
                 buffer.push(row);
+                emitted += 1;
                 if buffer.len() >= self.batch_size {
                     batches.push(self.flush_buffer(&mut buffer)?);
+                }
+                // LIMIT reached (counted post-filter): stop the merge early.
+                if let Some(cap) = limit {
+                    if emitted >= cap {
+                        break 'partitions;
+                    }
                 }
             }
         }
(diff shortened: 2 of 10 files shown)


pmcfadin:
## flow-closer endgame — gate of record + roborev

**Gate of record** (full `agent-gate.sh`) ran at `73a0a4ea`, then the branch was rebased onto a moved `origin/main` (node-bindings fix + #2143/#2142/#2138/#2137/#2136/#2134) to final head `df615392`. Evidence for the merge head:

### 1. Full gate @ `73a0a4ea` — 21/22 components PASS
`file-size fmt clippy core-tests tombstones-scan scan-offload-guard work-counters-guard byte-budget-guard memory-budget integration-tests format-compat write-tests cli-tests compaction-byte-parity python-bindings delivery-telemetry parity-report binding-unwind-profile tooling-tests minimal-build smoke` all **PASS**.

The only non-PASS was `node-bindings: FAIL (1s)` — a **worktree/harness env artifact** (a poisoned `NODE_OPTIONS=--require .../cmux-claude-node-options/restore-node-options.cjs` preload whose file was absent, so every `node` invocation died at startup before any test). This is **not caused by the #2129 diff** (which touches zero node code); it was also independently a known pre-existing node-bindings breakage on `main` that another team has since fixed on `origin/main`.

`CQLITE_ALLOW_FILE_GROWTH=1` was set intentionally: `cqlite-flight/src/{ticket,filter,producer}.rs` were already over the 800-line campsite threshold **before** this change (split tracked by #1116/#1135); this PR does not newly cross the threshold.

### 2. node-bindings re-cert @ rebased head `df615392` — PASS
`node-bindings: PASS (142s)` (real `npm ci + build + jest write-readback-content`, 3/3 tests) with a sanitized `NODE_OPTIONS`, against the merge head that includes the other team's node fix.

### 3. Lite re-cert @ `df615392` — PASS
`file-size fmt clippy scoped-tests` all PASS — confirms `cqlite-flight` compiles + tests clean against the **rebased-in** `origin/main` (notably #2134, which touches `cqlite-flight/main.rs` in the same crate, disjoint file — no interaction).

### roborev (codex, thorough) — 1 Medium finding, triaged NIT (not a blocker)
`applyAggregation()` (`CqliteFlightMetadata.java` L491) rebuilds the handle without carrying `table.limit()`. **Assessed structurally unreachable**: (a) `applyLimit` (L249) declines on `isAggregated()`; (b) `applyLimit` returns `limitGuaranteed=false`, so Trino retains its `Limit` node above the scan — `PushAggregationIntoTableScan` matches only a direct `aggregation(tableScan)` adjacency, so the retained `LimitNode` prevents `applyAggregation` from ever firing on a limit-carrying handle. No reachable wrong-result path → batched to the nits follow-up (#2113) as a defensive-guard + regression-test suggestion.

**Verdict: merge on green.** Gate PASS (of-record + rebase re-certs), roborev clean of blockers, CI required lanes green on `df615392`.

PR #2166, opened 2026-07-07
fix(#2164): applyFilter translates TupleDomain summary — domain-shaped predicates now pushed to filterJson

### LINKED ISSUE #2164: trino-connector: applyFilter ignores TupleDomain summary — simple col = literal predicates never pushed (filterJson empty)
## Round-2 harness finding (epic #2103), split out of #2157 after code investigation

`CqliteFlightMetadata.applyFilter` reads ONLY `constraint.getExpression()` (CqliteFlightMetadata.java:169) and returns `constraint.getSummary()` unchanged (:216 — 'domain returned unchanged; we don't consume it'). But Trino represents simple domain-expressible predicates — exactly `key = '001.25.535688'` on a varchar partition key — as a single-value **TupleDomain in the summary**, leaving `getExpression()` as TRUE. `PredicateTreeTranslator` (which only handles ConnectorExpression trees) therefore finds nothing pushable → `applyFilter` returns `Optional.empty()` → `filterJson=Optional.empty` → the flight server full-scans every split for a single-partition point read. This is the lab's EXPLAIN, verbatim.

It is NOT a type/capability gap: varchar maps to capability "full" (producer.rs:744-746) and `translateCompare`/`constantValue` handle varchar Equal (PredicateTreeTranslator.java:341-344). The Rust side would honor the predicate if it arrived (`ScanSpec.filter` from `ticket.effective_filter()`, filter.rs:239; applied per-row in `drive_merge`, producer.rs:530-534).

**Why tests missed it:** every `applyFilter` unit test builds `new Constraint(TupleDomain.all(), <expr>, ...)` — the predicate hand-placed in expression form with an empty domain (CqliteFlightMetadataApplyFilterTest.java:60,83,106,133). The domain-delivery path (how Trino actually sends these) is never tested.

## Fix
1. In `applyFilter`, translate `constraint.getSummary()` (TupleDomain) into the pushed filter tree — merge with whatever the expression path yields; return the remaining/unenforced domain honestly (`filterJson` semantics are best-effort post-filter server-side, connector keeps `ScanFilter` above — same posture as #2135's `limitGuaranteed=false`).
2. Regression tests MUST deliver predicates the way Trino does: single-value domain, multi-value (IN), ranges, on varchar + numeric columns, with `getExpression()=TRUE`. Keep the existing expression-path tests.
3. Follow-on candidate (separate, note only): derive split token bounds from a pushed partition-key equality (`murmur3(pk)`) so a point read prunes to ONE split instead of post-filtering all splits — today split ranges come solely from Sidecar tokenRangeReplicas (CqliteFlightSplitManager.java:44,97-116).

## Acceptance criteria
- [ ] `EXPLAIN SELECT * FROM cqlite.<ks>.<t> WHERE <varchar-pk> = '<v>'` shows a non-empty `filterJson` in the handle (unit-level: applyFilter returns a handle with filterJson for a summary-only constraint).
- [ ] Domain-shaped equality/IN/range predicates round-trip into the ticket JSON and the server's ScanSpec filter (existing Rust filter tests cover application).
- [ ] Tests fail on current main (TDD).
- [ ] Oracle-driven: no OpenSpec change. Gate PASS.

Epic: #2103. Split from #2157. Related: #2135 (limit analog), #834 (server-side predicate semantics).

pmcfadin:
Closes #2164.

## Root cause
`applyFilter` read only `constraint.getExpression()` and returned `constraint.getSummary()` unchanged (":domain returned unchanged; we don't consume it"). But Trino delivers simple domain-expressible predicates — exactly `key = 'v'` on a varchar partition key — as a single-value **TupleDomain in the summary**, leaving `getExpression()` = `Constant.TRUE`. `PredicateTreeTranslator` (expression-tree only) found nothing → `applyFilter` returned `Optional.empty()` → `filterJson` empty → the flight server full-scanned every split for a single-partition point read.

## What changed (Java only, `trino-connector/`)
- **`PredicateTreeTranslator.translateSummary(TupleDomain)`** (new): per-column `Domain` → the SAME `PredicateExpr` node shapes the expression path emits — single value → `Compare Equal`, discrete set → `In`, ranges → `Compare Gt/Gte/Lt/Lte` (bounded range = `And` of two; multiple ranges = `Or`). Capability gating **mirrors `translateCompare` exactly**: `NONE` pushes nothing, `EQUALITY` pushes only exact match / `In` / `IsNull`, ordering (ranges) requires `FULL`. Columns are emitted in stable column-name order.
- **`applyFilter`**: translates the summary and **merges (AND)** it with the expression-path tree via `combinePushed`; the existing accumulation + termination guards, limit threading (`table.limit()`), and aggregated-handle decline are untouched.
- **`PredicateTreeTranslator.and()`**: now **dedups** exact-duplicate conjuncts, so the unchanged summary (re-passed every optimizer iteration) is idempotent instead of growing an ever-larger tree.

Wire format matches the server's internally-tagged `PredicateExpr` (`ticket.rs`) — same shapes the existing expression-path tests already emit; no new JSON invented.

## Enforced vs remaining (honesty contract — mirrors #2135/#2129 `limitGuaranteed=false`)
The summary is returned **UNCHANGED** as the remaining (unenforced) filter. The pushed `filterJson` is a **best-effort server-side optimization**: the server applies it post-decode and Trino keeps its `ScanFilter` above, so results are always correct even if a leaf is dropped server-side. Nothing here claims enforcement, so no over-restrictive push can lose rows. Null-admitting domains (`col = x OR col IS NULL`) are left **entirely unpushed** — the server filter is destructive, so pushing only the non-null values would wrongly drop null rows.

## Test evidence (TDD, fail-first)
New tests in `CqliteFlightMetadataApplyFilterTest` deliver predicates the way Trino does (summary + `Constant.TRUE` expression): single-value varchar, IN, numeric range, bounded range, expression+summary merge, plus negatives (NONE-capability, range-on-EQUALITY, null-allowed) and domain idempotence.

**Fail-first on the unmodified production code** (proves the gap):
```
8 tests failed (pushable-path assertions), e.g.
  singleValueSummaryPushesEquality()   FAILED  NoSuchElementException (filterJson empty)
  multiValueSummaryPushesIn()          FAILED  NoSuchElementException
  numericRangeSummaryPushesComparison()FAILED  NoSuchElementException
  summaryAndExpressionMergeIntoOneFilter() FAILED  AssertionFailedError
16 tests completed, 8 failed
```
The 3 negative tests already passed on unmodified code (they expect empty). After the fix, all pass.

**Gradle** (`./gradlew test`): `TOTAL tests=198 failures=0 errors=0 skipped=0`.

**Lite gate** (Rust untouched — Java-only diff):
```
RESULT: PASS
file-size PASS  fmt PASS  clippy PASS (260s)  scoped-tests PASS (239s)
```
Full `AGENT-GATE SUMMARY` runs once pre-merge in flow-closer.

## Scope notes
- Wire format forced **no compromises** — the summary path reuses the expression path's `constantValue` type mapping, so both emit identical operands.
- Follow-on (NOT in this PR, per the issue): derive split token bounds from a pushed partition-key equality (`murmur3(pk)`) so a point read prunes to ONE split instead of post-filtering all splits.

### DIFF
--- a/trino-connector/src/main/java/in/mcfad/cqlite/flight/CqliteFlightMetadata.java
+++ b/trino-connector/src/main/java/in/mcfad/cqlite/flight/CqliteFlightMetadata.java
@@ -149,9 +149,17 @@ public class CqliteFlightMetadata implements ConnectorMetadata {
      * post-filtering them — results are always correct; pushdown is a pure
      * optimization.
      *
-     * <p>The {@code TupleDomain} summary is returned unchanged (we do not consume
-     * it), and partial-AND pushdown leaves the untranslatable conjuncts in the
-     * residual expression.
+     * <p>Trino delivers simple, domain-expressible predicates (e.g. a single-partition
+     * point read {@code key = 'v'}) not in the expression but as a {@link
+     * io.trino.spi.predicate.TupleDomain} in {@code constraint.getSummary()}, leaving
+     * the expression {@code TRUE} (issue #2164). We therefore ALSO translate the
+     * summary into pushable nodes and merge (AND) them with the expression path's
+     * tree. The summary is still returned <em>unchanged</em> as the remaining filter:
+     * the server applies the pushed tree best-effort post-decode and Trino keeps its
+     * {@code ScanFilter} above, so {@code filterJson} is a pure optimization, never an
+     * enforcement guarantee (same honesty posture as {@code applyLimit}'s
+     * {@code limitGuaranteed = false}, #2129). Untranslatable expression conjuncts are
+     * returned as the residual expression.
      */
     @Override
     public Optional<ConstraintApplicationResult<ConnectorTableHandle>> applyFilter(
@@ -170,8 +178,14 @@ public class CqliteFlightMetadata implements ConnectorMetadata {
 
         PredicateTreeTranslator.Result result =
                 PredicateTreeTranslator.translate(expression, constraint.getAssignments());
-        if (result.pushed().isEmpty()) {
-            return Optional.empty(); // nothing translatable to push
+
+        // Also translate the TupleDomain summary (how Trino actually delivers simple
+        // col = literal / IN / range predicates). Merge (AND) with the expression tree.
+        Optional<JsonNode> domainPushed =
+                PredicateTreeTranslator.translateSummary(constraint.getSummary());
+        Optional<JsonNode> newlyPushedOpt = combinePushed(result.pushed(), domainPushed);
+        if (newlyPushedOpt.isEmpty()) {
+            return Optional.empty(); // nothing translatable to push from either path
         }
 
         // Trino calls applyFilter iteratively, passing only the predicate at the
@@ -180,7 +194,7 @@ public class CqliteFlightMetadata implements ConnectorMetadata {
         // with whatever the handle already carries. Replacing would silently drop
         // an earlier condition whose residual we already reported as satisfied,
         // returning too many rows.
-        JsonNode newlyPushed = result.pushed().get();
+        JsonNode newlyPushed = newlyPushedOpt.get();
         Optional<String> existing = table.filterJson();
         String newlyPushedJson = serialize(newlyPushed);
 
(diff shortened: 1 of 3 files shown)


pmcfadin:
Roborev Medium finding addressed in acc76ca0: the In translation now canonicalizes discrete-set values — each element is converted to its JSON scalar, then sorted by the serialized scalar string (total, type-stable, independent of ValueSet iteration order). The applyFilter idempotence/accumulation guard compares serialized JSON, so the same logical multi-value domain now always serializes byte-identically and cannot be re-applied across planner iterations.

Regression test `multiValueSummaryEmitsCanonicalOrderAndGuardHolds`: builds the identical logical IN domain in opposite insertion orders, asserts canonical emitted order (a,b,c) and that the second applyFilter invocation returns `Optional.empty()` (guard holds).

Evidence: `./gradlew test` — 199 tests, 0 failures. Lite gate at acc76ca0 (dirty: no): `RESULT: PASS` (file-size/fmt/clippy/scoped-tests all PASS).

PR #2185, opened 2026-07-08
fix(#1803): deterministic agent-gate python-bindings — verify import + self-heal stale venv

### LINKED ISSUE #1803: agent-gate python-bindings component intermittently fails in the ephemeral venv (ModuleNotFoundError: cqlite._cqlite) — forces merge-over-red, erodes gate contract

pmcfadin:
Fixes #1803 (P0, milestone 0.14).

## Problem
The agent-gate `python-bindings` component intermittently FAILed at pytest import with `ModuleNotFoundError: cqlite._cqlite` even though the wheel built and code compiled cleanly. Root cause: the step reuses a persistent venv (`target/agent-gate-venv`) and went straight from `maturin develop` to `pytest` with **no verification** that `cqlite._cqlite` actually imported — so a stale/half-built editable install (cleaned `target/`, interrupted prior run, or a concurrent gate in the same checkout) surfaced as an opaque `ModuleNotFoundError`. When it triggered, the aggregate gate RESULT was FAIL despite green code, so workers merged-over-red — eroding the "gate is the only run that counts" contract.

## Fix
New single-source-of-truth helper `_python_build_verify_venv()` in `scripts/agent-gate.sh`, used by BOTH the full `run_python_bindings` and the `--lite` python tier. Exit contract:

| rc | meaning | full path | lite path |
|----|---------|-----------|-----------|
| 0 | healthy — import verified (persistent venv reused when good) | PASS | PASS |
| 1 | venv/pip/maturin **setup** gap (offline/toolchain, not a code defect) | FAIL | **SKIP** (loud) |
| 2 | `maturin develop` non-zero — **real build/compile failure** | FAIL | **FAIL** (distinct) |
| 3 | import failed after a clean-venv rebuild — **real binding defect**, not a venv miss | FAIL (distinct) | FAIL (distinct) |

The load-bearing logic: after `maturin develop` exits 0, verify `import cqlite._cqlite`; on a miss, tear down + rebuild the venv **once** (self-heal); if it still fails, FAIL with a distinct message. The healthy path reuses the persistent venv (no needless rebuild — preserves gate speed).

## Tests
- New hermetic self-test `scripts/tests/test_agent_gate_python_bindings_determinism.sh` (PATH-shadowed `python3`/`pip`/`maturin`/`python`, no real build) — **16/16**: healthy no-rebuild reuse, self-heal-to-PASS, real-defect→exit 3 + distinct marker, build-fail→exit 2, and a venv-reuse fast-path scenario (asserts no rebuild when the venv is already good). Wired into the gate's `tooling-tests` component so it's gated going forward.
- `test_agent_gate_delta.sh` still 66/66 (the rc-2/rc-3 FAIL notes start with `python-tier: FAIL` so the delta python-gap classifier stays correct).
- Real component `--only python-bindings`: PASS; venv-corruption sanity (removed `cqlite*` from site-packages → recovered to PASS, never FAIL).

## Review
roborev round 1 found a Medium (lite path masked rc-2 real build failures as SKIP) + a Low (test never exercised venv-reuse) — both fixed in this PR before first full gate (commits `d50e0044`, `4b1eb118`). Re-review clean.

Shell-only change (0 `.rs`). No new heuristics.

### DIFF
--- a/scripts/agent-gate.sh
+++ b/scripts/agent-gate.sh
@@ -904,49 +904,75 @@ _run_shell_selftest_files() {
   return "$rc"
 }
 
-# _python_build_verify_venv <venv> <maturin-develop-cmd> (issue #1803):
-# build the cqlite python extension into <venv> and GUARANTEE it imports, self-
-# healing a stale/half-built editable install EXACTLY ONCE. The persistent venv
-# under target/ is REUSED for the healthy path (the speed optimization); it is
-# torn down and rebuilt from scratch ONLY when a maturin-exited-0 build still
-# fails `import cqlite._cqlite` — the venv-resolution miss (a cleaned target/, an
-# interrupted prior run, or a concurrent same-checkout gate) that made the
-# python-bindings component flake with ModuleNotFoundError despite green code.
-# Contract (single source of truth for BOTH run_python_bindings and the --lite
-# python tier; exposed to the self-test via the --python-build-verify hook):
-#   exit 0 -> extension built AND `import cqlite._cqlite` verified (possibly after
-#             one clean-venv rebuild — a self-healed transient miss).
+# _python_build_verify_venv <venv> <maturin-develop-cmd> [<active-venv-out-file>]
+# (issue #1803): build the cqlite python extension into <venv> and GUARANTEE it
+# imports, self-healing a stale/half-built editable install EXACTLY ONCE. The
+# persistent venv under target/ is REUSED for the healthy path (the speed
+# optimization). On a miss (maturin exits 0 but the module does not import — a
+# venv-resolution miss from a cleaned target/, an interrupted prior run, or a
+# concurrent same-checkout gate) it self-heals into a PRIVATE per-process venv
+# rather than tearing down the shared one (roborev round-2 Finding B): a
+# concurrent gate in the same checkout (a `--only python-bindings` alongside a
+# full gate, or two lite runs) reuses the SAME $venv, so an `rm -rf "$venv"`
+# here would race that other process's mid-build/pytest use of it. macOS has no
+# flock, so we sidestep the race entirely instead of locking: never destroy
+# shared mutable state. Contract (single source of truth for BOTH
+# run_python_bindings and the --lite python tier; exposed to the self-test via
+# the --python-build-verify hook):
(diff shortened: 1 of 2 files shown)


pmcfadin:
roborev round-4 finding (missing `mktemp` in the hermetic toolchain-absent scenario) is a **false positive** — dismissed with reasoning:

The finding assumes `agent-gate.sh`'s full-gate startup runs (which includes `LOG_DIR=$(mktemp -d)` at line 1176). But the self-test drives the isolated `--python-build-verify` hook, whose case arm (lines 1145-1147) calls `_python_build_verify_venv` then `exit $?` — short-circuiting **before** line 1176. The hook never reaches that `mktemp`, so it isn't needed on the scenario PATH. Empirical confirmation: the scenario passes 27/27 with a PATH built purely from temp stubs (no `mktemp`) and was verified to hit rc 4 even with a real cargo/rustc on the host — if `mktemp` were required the scenario would crash at startup rather than pass.

Review-first complete: roborev rounds 1-3 findings (lite-tier failure masking, shared-venv teardown race, test hermeticity + exit-code matrix) all fixed; round-4 dismissed. Handing to endgame.

Issue #2066, opened 2026-07-06
Database::execute point-lookup (WHERE pk = ?) returns empty metadata.columns — CLI on main prints [{}] for point reads


pmcfadin:
Surfaced out-of-scope during #1581 (flow-lead worker, non-blocking; the #1581 CLI cutover to execute_streaming incidentally FIXES the CLI symptom, but the core API remains broken for other callers).

## Problem
The materializing point-lookup path in `Database::execute`: `SELECT * FROM t WHERE pk = ?` returns an empty `metadata.columns`, so any consumer rendering off metadata prints empty rows. Verified against the real CLI binary on main: `test_basic.simple_table` (drops the `id` pk column) and `test_collections.collection_table` (prints `[{}]`).

## Root-cause area
`cqlite-core/src/query/select_executor/` — the materializing partition-targeted lookup does not populate result metadata the way the scan path does. The streaming path (`execute_streaming`) returns the correct full row for the same query.

## Repro
```
cargo run -p cqlite-cli -- --schema test-data/schemas/basic-types.cql --data-dir test-data/datasets/sstables \
  --query "SELECT * FROM test_basic.simple_table WHERE id = <existing uuid>" --out json   # on main: [{}] / missing pk column
```
Compare with `execute_streaming` output for the same query (full row).

## Routing
Oracle-driven (read-path correctness; streaming path + sstabledump as oracle) — issue + pinned parity test, no OpenSpec.

## Acceptance criteria
- Pinned test: `Database::execute` point-lookup returns metadata.columns identical to the streaming path and values matching the JSONL golden for the row.
- No regression in the 33-table parity matrix.
- NOTE: coordinate with in-flight #1578 (D2 touches the same query engine area) — land after it or rebase onto it.

pmcfadin:
🧭 **MANAGER** 
GO
ORDER: 13

Milestoned **0.14** + **Ready**. P0 — `Database::execute` point-lookup (WHERE pk=?) returns empty `metadata.columns`, so consumers render empty rows. Surfaced out-of-scope during #1581 (its CLI cutover fixes the CLI symptom, but the **core API stays broken**). Point-read routing cluster with #1802/#1750.

pmcfadin:
[...]

Issue #1802, opened 2026-07-03
Point query SELECT * ... WHERE id = ? silently routes to the legacy full-scan executor (access_path=None), bypassing the #949 fast path


pmcfadin:
Surfaced building the read-perf gate (#1562). `QueryEngine::execute` routes a bare `SELECT * FROM t WHERE id = ?` (short query, <=8 whitespace tokens) to the LEGACY full-scan executor — `access_path = None` — so the most natural point query **silently full-scans** instead of using the partition-targeted #949/#956 fast path. A projected `SELECT id, name ... WHERE id = ?` does reach the fast path; the bare `SELECT *` form does not.

Impact: the most common point-read shape a user writes is O(partition-scan) instead of O(targeted lookup) — a latent production perf regression. The new #1562 read-perf gate benches the projected form (targeted), so it won't catch THIS routing quirk until fixed.

Fix: the executor-selection logic must route any equality-on-full-partition-key predicate to the targeted path regardless of projection / token count. AC: `SELECT * FROM t WHERE <full pk> = ?` yields `access_path.is_targeted()` (PartitionLookup); add a wiring test asserting the access path for the bare SELECT * form.

pmcfadin:
🧭 **MANAGER** 
GO
HOLD: merge after #1750
Both are point-query routing. #1750 (in progress) owns the is_simple_id_lookup heuristic fork; build on its resolution, don't fork the router in parallel.

pmcfadin:
🧭 **MANAGER** 
GO
ORDER: 13

P0 point query WHERE id=? routes to legacy full-scan, bypasses fast path — ranked ahead of the read-path perf epic. 0.14 policy: **P0 correctness/gate-green first**; the read-path epic runs in parallel on other workers (nothing held).

pmcfadin:
🧭 Data point from #1578's test work: `M2SelectValidator` rejects ANY WHERE-pk-equality + aggregate combination at the validator layer (`UnsupportedQuery("... M2 supports: SELECT with partition/primary key equality and optional LIMIT ...")`) before the optimizer/executor is reached — so the legacy point-read routing this issue tracks also has a validator-level guard in front of it for aggregate shapes. Worth folding into this issue's scope check when it's picked up.

pmcfadin:
[...]

Issue #175, opened 2025-10-20
Remove WAL and MemTable


pmcfadin:
# Issue #9: Remove WAL and MemTable (Write Infrastructure)

**Priority:** P1  
**Risk Level:** Medium (Core Storage Changes)  
**Estimated Time:** 4 hours  
**Assignee:** Team C (Infrastructure)  
**Branch:** `cleanup/issue-9-remove-wal-memtable`  
**Can Parallelize:** ❌ No (Must Follow #8)

---

## Objective

Remove Write-Ahead Log and MemTable components that are M5 write infrastructure, not needed for M1 reading.

---

## Problem Statement

`StorageEngine` includes write-side components:
- **WAL (Write-Ahead Log):** Durability for writes (377 lines)
- **MemTable:** In-memory write buffer (393 lines)

These are used by write operations (now gated behind `experimental` in Issue #8). Since M1 is read-only, these can be removed.

**Impact:** Reduces memory usage, faster initialization, cleaner architecture.

---

## Files to Delete

```
cqlite-core/src/storage/wal.rs          (377 lines)
cqlite-core/src/storage/memtable.rs     (393 lines)
```

**Total:** 770 lines

---

### 1. `cqlite-core/src/storage/mod.rs`

**Remove module declarations:**
```rust
pub mod wal;
pub mod memtable;
```

**Remove from StorageEngine struct:**
```rust
pub struct StorageEngine {
    // DELETE these fields:
    memtable: Arc<RwLock<memtable::MemTable>>,
    wal: Arc<wal::WriteAheadLog>,
    
    // KEEP these:
    sstables: Arc<sstable::SSTableManager>,
    // ... other read-side fields
}
```

**Remove from constructor:**
```rust
impl StorageEngine {
    pub async fn open(...) -> Result<Self> {
        // DELETE:
        let wal = Arc::new(wal::WriteAheadLog::open(...).await?);
        let memtable = Arc::new(RwLock::new(memtable::MemTable::new(config)?));
        
        // In return struct, DELETE:
        Ok(Self {
            // memtable,
            // wal,
            sstables,
            ...
        })
    }
}
```

### 2. `cqlite-core/src/storage/batch_writer.rs`

**Issue:** BatchWriter imports WAL. This file is M5 and should be deleted too, but that's Issue #10.

**For now:** Keep BatchWriter but make it not import WAL (may break it, but it's not used anyway).

Or just delete BatchWriter in this issue too (simpler).

---

[...]

PR #2191, opened 2026-07-08
fix(#1750): remove is_simple_id_lookup text heuristic — point reads return correct schema-derived columns

### LINKED ISSUE #1750: Legacy 'is_simple_id_lookup' text-heuristic fork returns column-less results — short 'WHERE id =' point reads render empty on every CLI output format
Filed by flow-lead (2026-07-02), confirmed by empirical investigation during the #942/#1749 work. Repro'd on main @6814f17f-era release build.

## Bug

`cqlite-core/src/query/engine.rs:164`:
```rust
let is_simple_id_lookup = cql.contains("WHERE id =") && cql.split_whitespace().count() <= 8;
```
Any SELECT that literally contains `WHERE id =` AND has ≤ 8 whitespace tokens is diverted from the modern `select_executor` to the **legacy `QueryExecutor` point-lookup path** (`query/executor.rs:288-333`), which builds its result via `QueryResult::with_rows` → `QueryMetadata::default()` (`query/result.rs:420-425`) with an **empty `columns` vec**, never applies projection, and never reconstructs the partition-key column.

All CLI output writers key rows off `result.metadata.columns` (`cqlite-cli/src/output/json.rs:64`, `csv.rs:61`, `table.rs:43-44`, parquet Arrow-schema build) — empty columns ⇒ empty output **even though `row.values` is fully populated**.

This is also a **no-heuristics-mandate violation**: query routing decided by substring + token count of the SQL text, with behavior that flips when the PK happens to be named `id` or when the user adds one more token.

## Repro

```bash
export CQLITE_DATASETS_ROOT=$PWD/test-data/datasets
K=0023ece7-7c4e-4705-9068-d1a59ec5fe19   # real key in test_basic/simple_table
./target/release/cqlite --schema test-data/schemas/basic-types.cql --data-dir <dir-with-simple_table> \
  --query "SELECT id,name,age FROM test_basic.simple_table WHERE id = $K" --out json
# => [ {} ]         (expected: the row)
# Adding any token (e.g. LIMIT 1) pushes >8 tokens → routes to select_executor → correct output.
```

## Blast radius (all empirically tested)

| Surface | Result |
|---|---|
| CLI `--out json` | BROKEN — `[{}]` |
| CLI `--out table` | BROKEN — blank row |
| CLI `--out csv` | BROKEN — empty fields |
| CLI `--out parquet` | BROKEN — column-less file (schema-only) |
| Python `.to_dict()` | data present (reads `row.values`) but `columns == []`, **PK column dropped, projection ignored** |
| Node `executeNative` | same degraded shape as Python |
| delta-export | unaffected (CDC path, no `Database::execute`) |

Trigger requires PK literally named `id` + ≤8 tokens; `WHERE pk = 1` fixtures are unaffected — which is why tests miss it.

Note: this fork also produced the false "2-generation blank row" alarm during the latency investigation — the multi-generation merge path itself was verified CORRECT against the `ct_multi_sstable_merge` golden (pk1→85, pk2→60 exact match), so no separate issue is filed for it.

## Fix (recommended)

**Retire the fork** — route all SELECTs through `select_executor` (which populates `metadata.columns`, applies projection, reconstructs the PK via `build_row_from_scan`, and takes the honest `AccessPath` fast path). The alternative (teach the legacy `execute_point_lookup` to populate metadata) keeps a doctrine-violating heuristic alive and a second executor to maintain; only take it if removing the fork exposes a dependency that genuinely needs the legacy path (name it in the PR if so).

## Acceptance criteria

1. Pinned regression test: the exact ≤8-token `WHERE id =` repro above through the public CLI surface, asserting JSON/CSV/table output contains the projected column names + values; fails on current main (revert-verify).
2. Same query through Python and Node returns populated `columns`, honors projection, includes the PK column — e2e binding tests.
3. Removing/bypassing the heuristic: no query text pattern routes to a different executor (grep-level assertion or unit test on the planner), satisfying the no-heuristics mandate.
4. `scripts/agent-gate.sh` PASS.

Routing: **bug fix with pinned regression tests — no OpenSpec** (fix direction is mandated by the no-heuristics doctrine; no design latitude). Straight to flow-implement.

Key files: `query/engine.rs:164-166`, `query/executor.rs:288-333, 743-772`, `query/result.rs:420-425`, `cqlite-cli/src/output/*`.

pmcfadin:
Closes #1750

## Bug (P0, no-heuristics #28 violation)
`query/engine.rs` had a text fork — `cql.contains("WHERE id =") && split_whitespace().count() <= 8` — diverting short point reads to the legacy `QueryExecutor`, which built results via `QueryMetadata::default()` (empty `columns`), so `WHERE id =` reads rendered **column-less** on every CLI/binding output.

## Fix
Removed the heuristic; ALL SELECTs now route through the modern `SelectExecutor` — schema-derived `metadata.columns` (in order, incl. PK), projection applied, PK reconstructed, and fast-path selection made from **parsed structure + schema** (not query text). Legacy `QueryExecutor` now handles non-SELECT only.

## Tests (new: `tests/issue_1750_simple_id_lookup_columns.rs`, via public `Database::execute` on real fixtures)
- `projected_where_id_eq_populates_metadata_columns` — columns in order + PK cell
- `select_star_where_id_eq_populates_metadata_columns` — schema-derived columns incl. PK
- `where_id_eq_still_takes_partition_lookup_fast_path` — `AccessPath::PartitionLookup` retained
All fail pre-fix (empty columns), pass post-fix.

## Review (review-first, before gate)
- **rust-reviewer: PASS** (nits only — see follow-up below)
- **roborev job 3777: 1 Medium finding → verified FALSE POSITIVE.** Claimed the experimental INSERT→SELECT round trip breaks; but `test_database_clone_functionality` fails identically on origin/main (parent commit) — `storage.put()` returns `UnsupportedFormat` since Issue #175, so the experimental INSERT stores nothing and no round trip exists to break. Refutation recorded on the review.

## Routing state for #1802 / #2066
Every SELECT now routes through `SelectExecutor`; a fully-constrained `WHERE pk = ?` takes `AccessPath::PartitionLookup` (fast path preserved). #1802 (bare `SELECT *` full-scan-vs-fast-path) builds directly on this unified routing.

## Follow-up (nits, batched — not fixed here)
rust-reviewer noted: (1) stale test docstrings in `execution_path_parity_tests.rs` still describe the retired heuristic as live design (behaviorally harmless — no routing asserts — but misdescriptive); (2) the SELECT-side legacy `plan_cache` branch in `engine.rs` is now effectively unreachable. Will file one cleanup issue.

### DIFF
--- a/cqlite-core/src/query/access_path.rs
+++ b/cqlite-core/src/query/access_path.rs
@@ -159,8 +159,10 @@ pub enum FallbackReason {
     /// so #962 can flip it to [`AccessPath::MetadataPartitionLookup`].
     MetadataScanPath,
 
-    /// The legacy `QueryExecutor` (simple-id-lookup and prepared SELECTs) issues
-    /// an unconditional `storage.scan`; it does not consult the fast path.
+    /// The legacy `QueryExecutor` issues an unconditional `storage.scan`; it does
+    /// not consult the fast path. Since issue #1750 ad-hoc SELECTs route through
+    /// the modern executor, so this reason now covers the remaining legacy SELECT
+    /// surfaces (a cached legacy SELECT plan reused via the plan-cache HIT branch).
     /// Tracked by #962 (route the legacy/prepared surfaces through the modern
     /// executor) and #961 (param binding).
     LegacyExecutorPath,
--- a/cqlite-core/src/query/engine.rs
+++ b/cqlite-core/src/query/engine.rs
@@ -166,14 +166,12 @@ impl QueryEngine {
     }
 
     /// Enforce the byte-bounded result budget (issue #1582 / D6) on a result
-    /// materialized by the LEGACY [`QueryExecutor`] — the simple `WHERE id =
-    /// <value>` point-lookup path, which is routed to `self.executor` (not the
-    /// budgeted `SelectExecutor`) for key-handling consistency with INSERT. A
-    /// wide-partition point lookup would otherwise materialize unbounded and
-    /// bypass the guard, INCLUDING on a plan-cache HIT (roborev #1582 D6). Reuses
-    /// the SAME estimator + enforcement as the optimizer path so the byte ceiling
-    /// and the row-count safety valve behave identically. Applied exactly once at
-    /// every legacy return site.
+    /// materialized by the LEGACY [`QueryExecutor`] — non-SELECT statements and
+    /// any cached legacy SELECT plan reused via the plan-cache HIT branch. Since
+    /// issue #1750 ad-hoc SELECTs route through the modern `SelectExecutor` (which
+    /// budgets itself); this guard covers the remaining legacy return sites,
+    /// INCLUDING a plan-cache HIT (roborev #1582 D6). Reuses the SAME estimator +
+    /// enforcement as the optimizer path. Applied exactly once at each site.
     fn enforce_legacy_result_budget(&self, result: &QueryResult) -> Result<()> {
         super::result_budget::enforce_materialized_rows(
             &result.rows,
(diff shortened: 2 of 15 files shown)


pmcfadin:
## flow-closer endgame: BLOCKED — full gate FAILED (self-inflicted regression)

The gate of record (full `agent-gate.sh`, run once on branch head `1a05138a`) **FAILED**: components `core-tests` and `python-bindings` red with 3 test failures, all traceable to this change's rerouting of `WHERE id = <uuid>` point reads from the legacy `QueryExecutor` to the schema-aware `select_executor`.

**Failures (all verified NOT pre-existing — the two rust tests PASS on a fresh `origin/main` detached worktree):**
1. `cqlite-core::dead_cache_delete_tests::stats_surface_reports_real_chunk_and_key_cache_observability` — `repeat cached read must yield real, non-zero chunk-cache hits (got 0 hits, rate 0)`. The legacy point-read path consulted the B1 chunk cache + B4 key cache via `get_cached_data`; `select_executor`'s point-read path does not, so cache observability collapses to structural 0.
2. `cqlite-core::dead_cache_delete_tests::stats_block_cache_hit_rate_and_occupancy_are_real` — same root cause.
3. `bindings/python .../test_schemaless_materialized_point_lookup_keeps_values` — a **schemaless** point lookup (`cqlite.open(DATASETS)`, no schema) now returns **0 rows** (`QueryResult(rows=0)`). `select_executor` cannot resolve a point lookup without a schema; the legacy executor could.

**Root cause / design note:** this is exactly the dependency issue #1750's recommended-fix section warned about — *'only take [retiring the fork] if removing the fork exposes a dependency that genuinely needs the legacy path (name it in the PR if so).'* Two real dependencies on the legacy point-read path are now named: (a) B1/B4 cache consultation, (b) schemaless point lookup. The `select_executor` point-read path needs to consult the same caches AND handle the no-schema case (or the fix must preserve the legacy path for the schemaless point read while still populating `metadata.columns`).

roborev job 3778 (codex): **clean, no issues found** — the correctness bug fix itself is sound; only these read-path dependencies regressed. C (intent audit vs #1750 acceptance criteria): the empty-columns bug IS fixed and the heuristic IS gone, but AC is not fully met while the schemaless point read returns 0 rows.

Requires a src-design fix round (fresh implementer, TDD) then re-run the full gate. Not merging.

Issue #2103, opened 2026-07-07
[EPIC] Distributed load-testing harness for cqlite-flight + Trino connector on easy-db-lab


pmcfadin:
## Goal

Stand up a repeatable, **multi-node** load-testing harness for the cqlite-flight data plane and the Trino connector on top of [`easy-db-lab`](https://github.com/pmcfadin/easy-db-lab). The loop:

1. Drive **CQL writes** into a real Cassandra cluster (`cassandra-easy-stress`).
2. `nodetool flush` (the connector reads flushed SSTables only).
3. Drive **heavy concurrent Trino reads** through the cqlite connector.
4. Watch cqlite's own **OTLP metrics + traces** (`cqlite.rpc.*`, W3C-propagated spans) land in the lab's observability stack (VictoriaMetrics / Tempo / Grafana / Pyroscope) to cook the internals and surface flight/gRPC weak spots.

This is the multi-node, real-cluster, real-observability version of the single-node `docker-compose --profile loadtest` E2E that already lives in `trino-connector/docker/`.

## Design decisions (owner-directed)

- **Artifacts are consumed, not built here.** Flight server via the published container `ghcr.io/pmcfadin/cqlite-flight:<tag>`; connector via the published Maven artifact `in.mcfad:cqlite-trino:<version>` assembled into a Trino plugin dir.
- **Kits live in this repo** under `easy-db-lab-kits/` and install out-of-tree via `easy-db-lab kit install <kit> --from easy-db-lab-kits/<kit>` — no fork of easy-db-lab, no fork of cqlite.
- **Read path: snapshot-first, direct-live optional.** Default to Sidecar snapshots for a consistent file set; expose a toggle to read the live data dir (which stresses the compaction-merge path hardest — files changing mid-read).

## What already exists (verified against the repo)

- Flight server: co-located container, mounts C* data dir read-only, serves Arrow Flight gRPC on `:8815`, ticket contract already supports a `snapshot` field (null = live dir).
- Flight observability: per-RPC OTLP metrics (`cqlite.rpc.requests/duration/rows/bytes/in_flight`) + W3C traceparent propagation, gated by `CQLITE_OTEL_ENABLED` + `CQLITE_OTEL_ENDPOINT` (OTLP gRPC :4317 default). easy-db-lab's node OTel collector already listens on OTLP 4317/4318.
[...]

PR #2199, opened 2026-07-08
fix(#2114): cqlite-flight kit — multi-disk detector makes single-data-dir limitation loud, never silent

### LINKED ISSUE #2114: cqlite-flight kit: single-disk data-dir assumption misses SSTables on multi-disk db nodes
Follow-up from #2103 / #2104 (low priority, breadcrumb).

The `cqlite-flight` kit's DaemonSet mounts a **single** host path into the pod (`--data-dir`, default `/mnt/db1/cassandra/data`) and points the flight server at it. This is correct for the standard single-disk db-node layout (e.g. `i4i.xlarge`, one local NVMe), which is what the 3-node test plan uses.

**Latent gap:** on a multi-disk instance type where Cassandra spreads `data_file_directories` across `/mnt/db1`, `/mnt/db2`, … the kit only mounts `db1`. The flight server would then silently miss SSTables living on the other disks — reads return partial/empty results with no error.

## Options (not urgent)

- Detect / accept multiple data dirs and mount each (`--data-dir` accepting a list, multiple hostPath volumes).
- Or, at minimum, have `start.sh` warn if it detects more than one mounted data disk on the node so the silent-partial-result failure mode is visible.

## Why low priority

Every current/planned run uses single-disk db nodes. This only bites if the instance type changes to multi-disk. Filing so the limitation is recorded rather than rediscovered as "queries return too few rows."

pmcfadin:
Closes #2114 (epic #2103 breadcrumb: single-disk data-dir assumption).

## What
The flight server binary accepts exactly ONE `--data-dir` (verified in `main.rs`), so per issue scope this does NOT add a server feature. Instead the kit makes the silent-partial-result failure mode impossible to miss:

- New `--data-root` kit arg (default `/mnt`) + a non-fatal `detect-multidisk` initContainer (reuses the flight image, read-only mount, ALWAYS exit 0 — never blocks rollout).
- Closed decision table — the ONLY quiet outcome is "exactly one existing `*/cassandra/data` candidate AND it equals the served `--data-dir`":
  - >1 candidate (by EXISTENCE, populated or empty) → loud warning naming each unserved dir as silently-missing.
  - Lone candidate ≠ served dir → loud mismatch warning naming both paths.
  - Root absent/auto-created-empty → loud cannot-verify warning (a mistyped `--data-root` can't fake success).
  - Root populated but zero candidates → loud layout-mismatch warning.
- `start.sh` prints the review command; README documents the limitation, the warning taxonomy, and the benign `DirectoryOrCreate` side effect.

## Review-first trail (roborev, 3 fix rounds)
r1 Medium: `DirectoryOrCreate` masked a mistyped root as "0 candidates" OK → absent-root branch added. r2 Medium: empty second data dir counted as absent → existence-based candidacy. r2 Low: stale token list → fixed. r3 Medium: lone-unrelated-candidate false OK → served-dir match required. Final review residual: 1 Low nit (path canonicalization for exotic spellings, e.g. `..` — warn-only worst case), batched to follow-up at merge.

## Evidence
- 6 fixture-based functional cases, all correct outcomes, all exit 0 (incl. trailing-slash normalization check).
- `bash -n`/`sh -n`/YAML parse PASS; AGENT-GATE LITE PASS @ each round (iteration only; full gate of record pre-merge via flow-closer).

## Out-of-scope breadcrumbs (reported, not filed)
- True multi-disk serving needs `--data-dir: Vec<PathBuf>` in cqlite-flight itself.
- Detection assumes the lab's `<root>/*/cassandra/data` layout; `--data-root` is the operator lever.


pmcfadin:
**flow-closer endgame — merge on green** (commit `3962e637`, rebased on `origin/main` @ `3c0ba45a`)

**Full gate of record — RESULT: PASS** (`/tmp/gate-full-2114.txt`, commit `3962e637`, dirty: no)
All 24 components green: file-size, fmt, clippy, core/integration/write/CLI tests, format-compat, compaction-byte-parity, python+node bindings, minimal-build, smoke. accelerators: sccache=on nextest=on lanes=on.

**C intent audit — PASS** (anchored to issue #2114 + PR #2199 closed decision table; oracle/mechanical, no OpenSpec). All 6 acceptance cases satisfied by the wired `detect-multidisk` initContainer: >1 candidate → loud per-disk warning; lone candidate != served → mismatch warning; root absent/empty → cannot-verify; root populated / zero candidates → layout-mismatch; exactly one existing candidate == served → quiet one-line OK; every path exits 0 (never blocks rollout). Wiring evidence: real initContainers block + read-only `cassandra-data-root` (DirectoryOrCreate) mount + `--data-root` kit arg + start.sh review command + README taxonomy.

**Final roborev confirmation — clean modulo 1 known Low nit (0 blockers).** codex job 1546 surfaced only the anticipated path-canonicalization nit (trailing-slash / equivalent-path matching in the `SERVED_IN` prefix compare; warn-only worst case, no current run is multi-disk). Non-blocking → batched to follow-up issue per severity rubric.

Merging squash + delete-branch.

Issue #2226, opened 2026-07-09
Epic AM — trino-flight read-path audit (2026-07-08): correctness + coverage findings


pmcfadin:
Five-lane read-only audit of the trino-flight read codepath at main @ 79a77287 (lanes: do_get/streaming lifecycle, pushdown end-to-end, type mapping, splits/snapshot/stats, test coverage). Full record: `docs/architecture/trino-flight-read-path-audit-2026-07-08.md`.

Verified working (no action): #2157 LIMIT/predicate enforcement genuinely wired with early termination; #2193 error-surfacing wired (observability-only — no framing defect found); #1473 cancellation; bounded do_get channel (cap 4); #1336 RF-correct stats merged + fail-closed; #1485/#1487 fail-closed conversions closed.

Children (waves):
- Wave 0 (before/with round-4): #2227 (snapshot multi-node), #2228 (start==end silent zero rows), #2229 (unsupported column poisons table + TIME)
- Wave 1: #2230 (wide-partition memory), #2231 (expression-path extra rows), #2232 (do_get transport tests), #2233 (e2e LIMIT/pushdown assertions), #2234 (ArrowToTrino golden round-trip), #2235 (scalar i32 offset overflow)
- Wave 2: #2236 (temporal unit-aware mapping), #2237 (ring-coverage validation), #2238 (missing projected column error), #2239 (pushdown capability alignment), #2240 (token prune runtime reuse), #2241 (split replica failover)

Child list is maintained via the `Epic AM` references below (each child links back here). All children Backlog by design — do not promote without owner.

## Implementation guidelines (per wave)

**Routing (all waves):** every child is oracle-routed — GitHub issue + pinned regression test, no OpenSpec — except the per-column-degradation UX sub-call inside #2229 (hide vs error-on-select), which is an owner decision at pickup. Standard loop per CLAUDE.md: TDD → `--lite` per round → review-first → flow-closer endgame. 1:1:1:1 per child.

### Wave 0 — unblock round-4 (P1; before or alongside the round-4 field run)
[...]

PR #2256, opened 2026-07-09
test(#2234): ArrowToTrino golden round-trip against server-emitted Arrow IPC bytes

### LINKED ISSUE #2234: trino-connector: ArrowToTrino golden round-trip test against server-emitted Arrow IPC bytes
Part of Epic AM (#2226). Coverage lane item 5. All 21 connector tests are pure unit tests; `ArrowToTrinoTest` converts hand-built VectorSchemaRoot vectors — schema/type drift between what the Rust server actually emits (field order, types, timestamp unit, nullability) and what the Java side assumes has NO in-process guard; only the ~17-row E2E would catch it, and only for the ~8 types it exercises.

**Acceptance criteria.**
- [ ] A small golden Arrow IPC blob produced by the Rust server for a known fixture is committed (`git add -f` if gitignore-adjacent; tiny).
- [ ] A JUnit test decodes it and feeds the resulting VectorSchemaRoot through `ArrowToTrino.toPage`, asserting values for every type the server emits on the supported set (incl. timestamp unit, uuid extension, date).
- [ ] Regenerating the golden is scripted + documented so drift updates are deliberate.

pmcfadin:
Closes #2234. Part of Epic AM (#2226), coverage lane item 5.

## What
Adds a drift-detection golden round-trip test for the trino-connector. Previously all 21 connector tests were pure unit tests over hand-built `VectorSchemaRoot`s — schema/type drift between what the Rust server actually emits and what the Java side assumes had NO in-process guard (only the ~17-row E2E caught it, for ~8 types).

- **`cqlite-flight/examples/emit_arrow_golden.rs`** — drives a real `MergeProducer` over a `WriteEngine`-flushed 2-row SSTable fixture and serializes with the server's wire schema (`producer.arrow_schema()`) — byte-identical to the `do_get` path (`service.rs:490`). Deterministic (fixed UUID/timestamp, no wall-clock/random).
- **Golden** `trino-connector/src/test/resources/golden/all_scalars.arrows` (~3.8K, `git add -f`) — genuinely server-emitted bytes.
- **`ArrowToTrinoGoldenTest`** — decodes the golden, runs `ArrowTypeMapper.toTrino` + `ArrowToTrino.toPage`, asserts field order, resolved Trino types, values + null encodings for all 12 supported scalars (bool/tinyint/smallint/int/bigint/float/double/text/blob/timestamp+unit/date/uuid-extension), plus (review-driven) exact blob bytes, `cqlite:pushdown` capability metadata (c_text=FULL, c_uuid=EQUALITY, c_blob=NONE via `capabilityOf`), and a schema-level timestamp-unit assertion.
- **Regen** `trino-connector/scripts/regen-arrow-golden.sh` + README doc — deliberate drift updates.

## Review-first
- rust-reviewer: APPROVE — confirmed the golden mirrors the real server wire path (not hand-built), deterministic/reproducible, drift-sensitive assertions.
- roborev (codex): flagged length-only blob check + dropped pushdown metadata; fixed (exact-bytes + capability assertions) + a schema-level timestamp-unit assertion; final pass clean (job 3822, no issues).

No production Java/Rust behavior changed (test/example/fixture only). No drift bug found — server bytes decode cleanly. JUnit + full 22-suite green.


pmcfadin:
🧭 flow-closer endgame — **HOLD (blocked, owner call)**

Gate of record: RESULT PASS (full mode, commit `2b1d6776`, dirty:no; `/tmp/gate-2234.txt`). Gradle golden JUnit `ArrowToTrinoGoldenTest`: BUILD SUCCESSFUL (tests=1, 0 failures) on the committed tree. C (issue #2234 acceptance criteria a/c): satisfied — golden produced by the real server emission path (`emit_arrow_golden.rs` → MergeProducer + wire schema) and committed; regen scripted (`trino-connector/scripts/regen-arrow-golden.sh`) + README-documented.

**Held on one roborev finding (job 3829, Medium) — verified real, owner's scope call:** the golden omits CQL `time`. The server emits `time` as Arrow `Time64(Nanosecond)` (`cqlite-core/src/export/arrow_convert.rs:400`) and the connector fully supports it (`ArrowTypeMapper` → `TimeType`; `ArrowToTrino.toPage` `TimeType`/`TimeNanoVector` path). So `time` is a connector-supported scalar the server emits that the new drift guard does NOT cover — which reads against criterion (b) "assert values for **every type the server emits on the supported set**." Whether (b) requires `time` (vs. the enumerated timestamp-unit/uuid/date cases) is a scope-completeness decision.

Owner decision needed:
- **(A) Accept as-is** — treat criterion (b) as satisfied by the enumerated set; I merge immediately, file the `time` gap as a linked follow-up.
- **(B) Require `time` coverage now** — respawn implementer to add `c_time` to the generator, regenerate the golden, extend the Java test (this is a Rust-example + fixture change → invalidates the gate → I re-run the full gate before merge).

Not merged, not finalized, no telemetry stamped.

pmcfadin:
🧭 Lead decision (owner AFK, 'keep moving' standing directive): taking **Option B — add `time` coverage now**, not merge-with-follow-up. Rationale: CQL `time` is emitted by the server (Arrow Time64/ns) AND fully supported by the connector, so it is squarely inside criterion (b)'s 'every type the server emits on the supported set' — this is completing the issue's written acceptance criterion, not a scope tradeoff. Per doctrine we don't merge over an unmet criterion. Respawning the implementer to add `time` to the golden fixture + regen + Java assertion; the Rust example/fixture change invalidates the gate, so the FULL gate re-runs before merge. Will merge once green.

Issue #2231, opened 2026-07-09
pushdown expression path: server float-NaN ordering and >2^53 bigint f64-equality diverge from Trino — extra rows when a conjunct is removed from the plan


pmcfadin:
Part of Epic AM (#2226). Audit findings N6+N7 (pushdown lane). Silent wrong-results (extra rows), narrow trigger.

**Mechanism.** `applyFilter` returns only UNTRANSLATABLE expression conjuncts as residual (`PredicateTreeTranslator.java:82-99`); a fully-translated expression conjunct (e.g. inside an OR — only representable via `getExpression()`) is REMOVED from the Trino plan, so the server's evaluation must match Trino EXACTLY. Two divergences in `cqlite-core/src/query/select_executor/value_ops.rs`:
1. **NaN Gt/Gte** (:51): `try_compare_values` uses cassandra_double_cmp (NaN sorts greatest) → `d > 1.5` keeps d=NaN; Trino/SQL: UNKNOWN → dropped. Repro: `WHERE d > 1.5 OR name = 'zzz'` with a row (d=NaN) → CQLite returns it, Trino semantics don't. Only Gt/Gte leak; Lt/Lte/Eq consistent.
2. **f64 equality fallback** (:17): `values_equal` falls back to `as_f64() ==` → distinct i64 > 2^53 compare equal. Repro: `WHERE bigcol = 9007199254740993 OR name = 'zzz'` matches bigcol=9007199254740992.

Summary-path predicates are safe (Trino retains + re-applies them).

**Acceptance criteria.**
- [ ] Either the pushed evaluation is SQL-semantics-exact for Gt/Gte-on-NaN and integer equality (no f64 fallback for int-vs-int), OR float-ordering / large-integer-equality conjuncts are kept in the expression residual (never removed from the plan). Pick one, both repro queries pinned as tests.
- [ ] Rust-side truth tests: NaN under Gt/Gte/Lt/Lte/Eq vs the documented Trino semantics; i64 equality at 2^53±1.

pmcfadin:
🔒 Claimed by flow-lead worker (`ip-172-31-18-96`) — branch `issue-2231-pushdown-nan-int-equality` pushed, worktree at `~/projects/cqlite-wt/issue-2231`. Pipelined as the second lane alongside #1668 (independent lane; only full-gate runs serialize on this machine).

pmcfadin:
[...]

Issue #1135, opened 2026-06-26
Reduce large test files for agent-friendly editing — extract shared fixtures + split by scenario


pmcfadin:
Companion to epic #1116 (oversized source-file breakup). Same underlying driver — **context cost during agentic coding** — applied to test files, where the right fix is different.

## Why test files matter for this goal

An agent must **read a file before it can edit it**, so file size is a per-edit token tax that is *format-blind* — a large test file costs exactly as much to load as a large source file. The build-time arguments that make large test files "cheaper" (separate-crate compilation, no private-access tension) **do not apply to context cost.** Two extra effects bite hardest in test files:

- **Edit-anchor ambiguity** — repetitive test bodies make it hard to find a unique string to edit against; the agent reads more just to disambiguate, and mis-targets more often.
- **Working-memory dilution** — loading a 1.5k–4.8k-line test file squeezes out the schema / code-under-test / task context the agent actually needs.

## Why the approach differs from the source epic

Tests are **append-mostly and independent**, so the goal is *"an agent loads only the minimal slice to add/edit one test."* That is achieved less by mechanical splitting and more by:

1. **Extract shared fixtures into `tests/support/`** — the repo already has the pattern (`cqlite-core/tests/support/canonical_jsonl.rs`, `tests/common/`, `tests/helpers/`); it's just not applied consistently. Inline helper functions repeated across files are pure, repeated context tax.
2. **Split catch-all suites by scenario/feature** — one concern per file, greppable names, so an agent opens one focused file instead of a monolith.
3. **Table-drive repetitive cases** (`rstest` is available) — but don't over-rotate: a giant parameter table is itself one big edit target.

**Do not touch healthy large files** — large is fine when it's many small independent tests.

## Threshold

Flag test files **> ~1,500 lines** (8 files today). Within those, prioritize by **lines-per-test** and **inline-helper count**, not raw size — a high lines-per-test value means a single test spans hundreds of lines, which is the worst case for context cost.

## Worst offenders (measured)

High lines-per-test (monolithic test bodies) and/or helper-heavy — these are the real targets:

| File | Lines | Tests | Inline helpers | Lines/test | Problem |
|---|---|---|---|---|---|
| `tests/src/comprehensive_integration_test_suite.rs` | 1,411 | 2 | 0 | **705** | 2 giant test fns — split by scenario |
| `tests/src/cql_integration_tests.rs` | 1,344 | 2 | 2 | **672** | 2 giant test fns — split by scenario |
[...]

PR #2276, opened 2026-07-09
fix(#2257): legacy QueryExecutor drops NaN under Gt/Gte/Lt/Lte WHERE predicates

### LINKED ISSUE #2257: Legacy QueryExecutor: NaN Gt/Gte leak (same mechanism as #2231, different execution path)
## Summary

The identical NaN-under-`Gt`/`Gte` divergence fixed in #2231 for the Trino
pushdown path (`cqlite-core/src/query/select_executor/value_ops.rs`) also
exists, unfixed, in the **legacy `QueryExecutor`**
(`cqlite-core/src/query/executor.rs`, `compare_values`/`evaluate_condition`,
~lines 603-655).

`compare_values` still calls `cassandra_double_cmp` directly for `Float`/
`Float32` operands with no NaN-drop:

```rust
(Value::Float(a), Value::Float(b)) => Ok(dcmp(*a, *b)), // Cassandra order #1870/#2010
(Value::Float32(a), Value::Float32(b)) => Ok(fcmp(*a, *b)), // Cassandra order #1870/#2010
```

`cassandra_double_cmp`/`cassandra_float_cmp` sort NaN as the GREATEST value
(correct for ORDER BY / clustering / MIN/MAX), so `evaluate_condition`'s
`GreaterThan`/`GreaterThanOrEqual` arms (which pattern-match on
`Ordering::Greater`/`Ordering::Greater | Ordering::Equal`) report `true` for a
NaN row against any finite bound — the same SQL-semantics violation #2231
fixed for the Trino pushdown path (`select_executor::value_ops`), just in this
different, still-live execution path.

## Is this reachable / does it matter?

This is a **different execution path** than the one #2231 scopes to. #2231's
mandate ("a fully-translated conjunct is removed from the Trino plan, so
CQLite's own evaluation must match Trino EXACTLY") is specific to the Trino
Flight pushdown/residual-removal contract. This legacy `QueryExecutor` path is
still reachable from `engine.rs` (`:132`/`:230` route some point reads through
it), so it is a live, user-facing divergence for direct CQLite SELECT queries
with a NaN `double`/`float` column and a `>`/`>=` predicate — independent of
Trino. It also inherits the (already-fixed-elsewhere) large-integer f64
concern: `compare_values` has no `Integer`-vs-`BigInt`/wide-integral coercion
path at all today (only `Integer == Integer`), so the 2^53 f64-collapse bug
does not currently reproduce here the same way — but if integer coercion is
ever added to this comparator, the same care (`i128`, never `as_f64`) will be
needed.

## Repro

Same shape as #2231's divergence 1, but routed through the legacy executor
instead of the Trino Flight pushdown path:

```sql
SELECT * FROM ks.t WHERE d > 1.5 OR name = 'zzz';
-- row: d = NaN, name = 'aaa'
```
Executed via a code path that reaches `QueryExecutor::evaluate_condition`
(see `engine.rs:132`/`:230`) incorrectly includes the NaN row; SQL/Trino
semantics would drop it (NaN comparison is UNKNOWN).

## Root cause

`cqlite-core/src/query/executor.rs`'s `compare_values`/`evaluate_condition`
were never migrated to the shared, NaN-predicate-aware comparators added in
`cqlite-core/src/query/select_executor/value_ops.rs`
(`try_compare_values_predicate` / `compare_values_ordering_predicate`,
issue #2231) — they still use the raw ordering comparator
(`cassandra_double_cmp`/`cassandra_float_cmp`) meant for ORDER BY/clustering,
not WHERE-predicate evaluation.

## Suggested acceptance criteria

Oracle-driven bug fix (no OpenSpec needed) — pick one:
- (A) Migrate `evaluate_condition`'s `Gt`/`Gte`/`Lt`/`Lte` arms (and `Equal`,
  for consistency with #2231's large-integer-equality fix, if this path ever
  gains numeric coercion) to call the shared `value_ops` predicate comparators
  instead of `compare_values`/`cassandra_double_cmp` directly, OR
- (B) Determine this path is dead/unreachable from any live query surface and
  document that explicitly (with a code comment + a regression test proving
  it), removing the ambiguity.

Required tests (mirroring #2231's Rust-side truth tests): a `Gt`/`Gte` NaN
truth test against `evaluate_condition`/`compare_values` confirming a NaN
double no longer satisfies `>`/`>=`, an end-to-end regression test reproducing
the `d > 1.5 OR name = 'zzz'` shape through whatever entry point
`engine.rs:132`/`:230` uses, and confirmation `Lt`/`Lte`/`Equal`/ORDER BY are
unaffected.

## Related

- #2231 (Trino pushdown NaN Gt/Gte + large-integer-equality fix — this issue's
  sibling divergence, different execution path)
- #1870, #2010 (origin of the NaN-greatest total-order semantics this
  comparator correctly implements for sort order, but incorrectly reuses for
  predicate evaluation)

pmcfadin:
Closes #2257

## Summary
Issue #2231 fixed a NaN-under-Gt/Gte predicate-evaluation bug in the Trino pushdown path
(`select_executor::value_ops`) — a NaN operand incorrectly satisfied `>`/`>=` because the comparator
used Cassandra's NaN-greatest total order (meant for sorting), not SQL three-valued UNKNOWN semantics.
The identical bug existed, unfixed, in the legacy `QueryExecutor` (`executor.rs`), which is still LIVE
(reachable from `engine.rs` for some point reads) — so it's a real, user-facing divergence for direct
CQLite SELECT queries independent of Trino.

Fix: new `predicate_ordering` helper in `executor.rs`, scoped strictly to the NaN case (checks
`is_nan_value` on both operands, drops the row via `Ok(None)` for either being NaN) and delegates to the
existing `self.compare_values` UNCHANGED for every other case. `Equal`/`NotEqual` were already NaN-safe
(native `==`). `apply_sort_step`/ORDER BY untouched — NaN-greatest total order preserved there.

## Review history (review-first, before any full gate)
- Round 1 (roborev Medium + `rust-reviewer`, converging independently): the initial approach routed
  through the shared `value_ops::try_compare_values_predicate`, which ALSO does cross-numeric coercion
  (`Integer` vs `BigInt`/`Float`) that legacy `compare_values` never had — silently turning previously
  hard-`Err`ing comparisons into coerced successes. Undocumented, untested scope creep beyond the NaN
  fix. Fixed by tightening `predicate_ordering` to intercept ONLY NaN and otherwise delegate to
  `compare_values` completely unchanged.
- Round 2 (re-verify): both roborev and `rust-reviewer` confirm clean — traced the control flow to
  confirm the widening is gone (cross-numeric pairs reach the original `Err` path, not the coercion
  path), confirmed the new regression test would have genuinely failed against the pre-fix code (not a
  tautology), confirmed the visibility revert breaks no other caller.

## Test plan
- [x] `evaluate_condition_drops_nan_for_all_relations` — NaN truth test across Gt/Gte/Lt/Lte/Eq
- [x] `apply_filter_step_drops_nan_row_for_gt` — end-to-end through the real live path
      (`apply_filter_step` → `evaluate_condition`), reproducing the issue's `d > 1.5 OR name = 'zzz'` shape
- [x] `sort_order_nan_greatest_unchanged` — guardrail, ORDER BY/sort semantics untouched
- [x] `predicate_ordering_still_errs_on_cross_numeric_pairs` — pins that cross-numeric-type comparisons
      still hard-`Err` exactly as before (the scope-creep this PR avoids)
- [x] `scripts/agent-gate.sh --lite` PASS at every round
- [ ] Full `scripts/agent-gate.sh` (the gate of record) — runs once via `flow-closer`

### DIFF
--- a/cqlite-core/src/query/executor.rs
+++ b/cqlite-core/src/query/executor.rs
@@ -634,24 +634,32 @@ impl QueryExecutor {
     /// sibling of the Trino-pushdown fix #2231).
     ///
     /// Returns `Ok(None)` — SQL UNKNOWN, so the caller DROPS the row — when
-    /// either operand is a NaN float. It delegates to the SHARED, principled
-    /// predicate comparator `value_ops::try_compare_values_predicate` (the same
-    /// one #2231 introduced), which drops NaN and compares large integers
-    /// exactly (`i128`, no `f64` collapse). On `Err` (genuinely incomparable
-    /// operands, e.g. `Null` vs a value, or two disjoint non-numeric types) it
-    /// falls back to this path's own [`Self::compare_values`], preserving the
-    /// legacy executor's established `Null`-vs-value ordering and its
-    /// same-type/incomparable-type error semantics — the fix is scoped to the
-    /// NaN-predicate leak only, exactly like #2231.
+    /// either operand is a NaN float (issue #2257, the legacy-executor sibling
+    /// of the Trino-pushdown fix #2231). Otherwise delegates to
+    /// [`Self::compare_values`] UNCHANGED.
+    ///
+    /// Deliberately does NOT route through `value_ops::try_compare_values_predicate`
+    /// for the non-NaN case: that shared helper also does cross-numeric
+    /// coercion (`Integer` vs `BigInt`/`Float` via `as_f64`, exact `i128` for
+    /// integrals) that this legacy path has never had — `compare_values` only
+    /// matches identical variants and `Err`s on any cross-numeric pair. Since
+    /// WHERE-clause literals commonly parse as `Value::Integer` regardless of
+    /// the column's actual CQL type, falling through to that coercion would
+    /// silently turn a previously-hard-`Err` query into a successful (and
+    /// unvalidated against real Cassandra semantics) comparison — a functional
+    /// widening this issue does not ask for. The fix is scoped to EXACTLY the
+    /// NaN-predicate leak: same-type/incomparable-type error semantics and
+    /// `Null`-vs-value ordering are untouched.
     ///
     /// This is for filter/`WHERE` evaluation ONLY. ORDER BY / sort
     /// ([`Self::apply_sort_step`]) still uses [`Self::compare_values`] directly,
     /// keeping the Cassandra NaN-greatest total order unchanged.
     fn predicate_ordering(&self, a: &Value, b: &Value) -> Result<Option<Ordering>> {
-        match crate::query::select_executor::value_ops::try_compare_values_predicate(a, b) {
-            Ok(ordering) => Ok(ordering),
-            Err(_) => self.compare_values(a, b).map(Some),
+        use crate::query::select_executor::value_ops::is_nan_value;
+        if is_nan_value(a) || is_nan_value(b) {
+            return Ok(None);
         }
+        self.compare_values(a, b).map(Some)
     }
 
     /// Compare two values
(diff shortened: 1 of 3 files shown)


pmcfadin:
Endgame (flow-closer): full gate of record PASS on rebased head `4335163` (23/23 components: core-tests, compaction-byte-parity, query-semantics-oracle, python/node bindings, etc. all PASS).

`file-size` note: the run used `CQLITE_ALLOW_FILE_GROWTH=1` (documented opt-out). `cqlite-core/src/query/executor.rs` was already over the 800-line src threshold on `origin/main` (1276 lines); this NaN-predicate bug fix grows it to 1475. Splitting the file is epic #1116 / #1135 work and is out of scope for an oracle-driven predicate bug fix — hence the acknowledged growth per CLAUDE.md doctrine.

Intent audit (C): PASS — approach (A) from the issue executed (inequality arms migrated to a NaN-aware `predicate_ordering`, scoped strictly to NaN, delegating unchanged to `compare_values`; ORDER BY untouched). Required tests present on the production path (`evaluate_condition`/`apply_filter_step`): all four inequalities + Equal for Float/Float32, end-to-end filter-step drop, ORDER-BY-unchanged guardrail, and a cross-numeric-error regression test guarding the review-first finding.

Final roborev confirmation pass (claude-code, base origin/main): CLEAN on `4335163` — "No issues found." (review-first converged clean over two prior rounds; 1 Medium blocker + 1 Low nit fixed in round 1, round 2 + this confirmation clean, no open nits).

Issue #2290, opened 2026-07-09
connector deployment: arrow-java requires --add-opens=java.base/java.nio — add to kit jvm.config, document, and FAIL FAST at plugin init when absent


pmcfadin:
Part of Epic #2103 / Epic AM #2226. **This is the actual fix for #2193** (root-caused in round 5, #2286): arrow-java's `MemoryUtil.<clinit>` requires `--add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED` on JDK 17+. Trino 481's stock jvm.config lacks it; the trino-cqlite kit doesn't add it → arrow-java dies on the FIRST frame read → every do_get fails with the cryptic per-query `Failed to read message`. Validated live by the field team: adding the flag + restart → the entire read path works (tiny/mixed/TIME/count all pass).

### 1. Kit: add the flag on install (`easy-db-lab-kits/` trino-cqlite kit)
Wherever the kit provisions Trino's `jvm.config` (coordinator AND worker), append:
```
--add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED
```
Idempotent (don't duplicate if present). This is arrow-java's own documented requirement (https://arrow.apache.org/docs/java/install.html).

### 2. Connector: fail fast at plugin init with an actionable message
In the connector's plugin/connector-factory init path (before any query can run), probe arrow-java memory init — e.g. attempt to initialize `org.apache.arrow.memory.util.MemoryUtil` (or construct a `RootAllocator`) inside a try/catch. On `ExceptionInInitializerError`/`NoClassDefFoundError` mentioning MemoryUtil: throw a configuration error whose message names the EXACT missing flag and the jvm.config remedy. This converts a cryptic per-query wire error into an immediate, self-explanatory startup failure. Pinned test: simulate/verify the error message shape (unit-test the message formatter at minimum; if the init probe is isolatable, test both paths).
NOTE: the probe must be safe when the flag IS present (no side effects, allocator closed if opened).

### 3. Docs: required-JVM-args section
Connector README (+ the docs-site connector page if one exists): a "Required JVM configuration" section with the flag, why (arrow-java JDK17+ module access), and the fail-fast behavior. Keep doctrine current in the same change.

## Also explain the e2e blind spot
[...]

Issue #1935, opened 2026-07-04
Regenerate TTL-carrying corpus tables without TTL + goldens + dataset pin bump (#1896 cluster A, owner-decided)


pmcfadin:
Owner decision on #1896 (2026-07-04): regenerate the TTL-carrying fixture tables WITHOUT TTL (or 100-year TTL), regenerate their JSONL goldens, and bump the dataset pin — the durable fix for the #1853-class fixture time bombs.

**Currently red because of this** (every main CI run): Node CI parity.test.js row-count tests for `ttl_test_table`, `app_metrics`, `log_entries`, `tick_data`, `test_oa.ttl_table`; plus 4 gate-uncovered CLI comprehensive_select_test cases and possibly table_snapshot_tests goldens (#1896 cluster B — verify during regeneration).

**Scope**
1. Identify ALL TTL-carrying tables in the corpus (not just the 5 known) — audit schemas for `default_time_to_live` / USING TTL inserts.
2. Regenerate via the corpus process: real Cassandra 5 generation, `cargo run -p cassandra-parity -- corpus-audit` clean, exhaustive-regeneration workflow contracts (.github/workflows/exhaustive-regeneration.yml), manifest/provenance updated.
3. Regenerate JSONL goldens; new dataset release asset; **SHA-keyed cache update** (DATASET_TAG/ASSET/SHA256 in .github/workflows/ — never tag-only keys, see memory: dataset release gotchas; macOS tar AppleDouble gotcha applies to asset packing).
4. Bump the dataset pin everywhere it's referenced; fetch-datasets.sh round-trips.
5. **Keep #1853's pinned-now seam test as-is** — issue_694's TTL-shadowing coverage intentionally tests expiry semantics and must keep a TTL fixture (its pinned clock makes it time-proof; if its table is regenerated without TTL, coordinate: either keep one dedicated TTL table for it with the seam, or regenerate its fixture WITH TTL and rely on the seam).
6. Verify: Node CI parity green, Python parity green, CLI comprehensive_select green (the #1896 cluster A cases), sstabledump-parity green, full gate PASS.

Depends on: owner decision (recorded), corpus tooling. Ref #1896, #1853, #1742 (semantic-oracle gap), #1928. Oracle-driven, no OpenSpec. — flow-lead

pmcfadin:
[...]

PR #2292, opened 2026-07-09
chore(connector): align arrow-java with Trino 481's line (18.1.0 → 19.0.0) + netty 4.1.130 pin + exhaustive version-drift guards

### LINKED ISSUE #2193: arrow-java cannot decode cqlite-flight do_get on tiny nb-big table ('Failed to read message') — cross-stack Flight interop; Java-side decode oracle missing
Part of Epic AM (#2226). **P1 — round-4 field blocker** (owner's round-4 verdict comment below has the full wire evidence). Deterministic in the field: 3-row `cassandra_easy_stress.tiny` (nb-1-big, LZ4) fails every read through Trino with arrow-java `Failed to read message`, while the server logs success.

## What the code audit PROVES (read before theorizing — these paths are verified, do not re-litigate)

The do_get egress is **100% stock arrow-flight**, and every "malformed framing" hypothesis is disproven in-source:
- The ONLY encoder is `FlightDataEncoderBuilder::new().with_schema(schema_ref).build(batch_stream)` — `cqlite-flight/src/streaming.rs:297-307`. No hand-rolled `FlightData`, no custom `IpcWriteOptions` (defaults: **IPC V5**, align 8, **no compression**), no dictionaries (all primitive/utf8).
- **Schema is always emitted first** — injected synchronously by the encoder; the cap-4 channel (`streaming.rs:202`) carries only `RecordBatch`es. There is no path where a batch precedes/skips the schema; single-batch differs from multi-batch only in timing, never message order. (The field capture's "one h2 DATA frame" is NOT evidence of a missing message — one DATA frame can coalesce both gRPC messages.)
- Aggregate path uses the same encoder (`streaming.rs:247-285`). Empty table = schema-only stream (test `service.rs:998-1030`).
- Why the server logs success (this resolves the round-4 "observability promises not met" finding): if the bytes are well-formed to arrow-rs, the stream completes (`streaming.rs:491-495`), tonic flushes DATA + END_STREAM into local h2 buffers, and the client's GoAway/decode-failure arrives on a stream the server already closed — **tonic surfaces no post-END_STREAM peer reset** to `streaming.rs`/`obs.rs`. Nothing re-enters `record_status_error` (`obs.rs:366-400`). Also: `cqlite.errors.total` is compiled into the field image (Dockerfile builds `--features observability`) but **registers lazily on first increment** — zero server-side errors ⇒ metric name absent. Both round-4 observations are consistent with a server that genuinely believes it succeeded.

## The actual gap: no test anywhere decodes real Rust-emitted FlightData with arrow-JAVA
- The transport test uses an arrow-RS client (`cqlite-flight/tests/do_get_transport_test.rs:169`; the 3-row repro at `:182` passes and asserts row count only).
- The Java golden test (`trino-connector/.../ArrowToTrinoGoldenTest.java:81`) reads **IPC stream-format** bytes via `ArrowStreamReader` — a DIFFERENT framing from Flight's `FlightData` (bare Message flatbuffer in `data_header` + separate `data_body`). It cannot catch a Flight-framing/interop failure.
- Version skew is real and unguarded: **arrow-rs 53.4.1 (flight) ↔ arrow-java 18.1.0 (flight-core)**. Both speak IPC V5, so it is not a metadata-version mismatch, but cross-stack Flight compat is untested.
- The only non-vanilla element on the wire: **field-level `cqlite:pushdown` metadata** (`cqlite-flight/src/producer.rs:441`) — the sole cqlite-specific decode suspect.
- Known NON-repro configuration: the docker e2e (arm64 image, Trino latest, connector 0.13.3) reads small tables through real arrow-java successfully. Field (x86_64 image, Trino 481) fails deterministically. Architecture, Trino/arrow-java runtime version, and exact table shape (header-extracted schema: headerless nb, 2 cols/1 pk/0 ck) are live discriminating axes.

## The work (phased; do them in order; STOP and report between phases)

**Phase 1 — build the missing oracle (the pinned test, required regardless of outcome).**
1. Extend `cqlite-flight/tools-or-tests emit_arrow_golden.rs` (`:212-216` currently emits StreamWriter bytes) to ALSO emit the protobuf-encoded **`FlightData` message sequence** via the same `FlightDataEncoderBuilder` path, from a fixture mirroring the field shape exactly: 3 rows, schema extracted-from-header shape (2 columns `key text, value text`, 1 pk, 0 ck), WITH the `cqlite:pushdown` field metadata. Commit the golden.
2. New Java test `trino-connector/src/test/java/in/mcfad/cqlite/flight/FlightDataGoldenDecodeTest.java`: decode that committed FlightData sequence with arrow-java's **Flight-level** machinery (`ArrowMessage`/`FlightStream` path — NOT `ArrowStreamReader`). Assert: message 0 parses as Schema, message 1 as RecordBatch, 3 rows, values "1","2","3".
3. New Rust low-level assert in `do_get_transport_test.rs`: capture raw `Vec<FlightData>` (before `FlightRecordBatchStream`) and pin the sequence — exactly 2 messages for the 3-row case, `data_header` of [0] parses as Schema, [1] as RecordBatch.

**Phase 2 — bisect ONLY if Phase 1's Java test reproduces `Failed to read message`.** One variable at a time: (a) re-emit the golden WITHOUT `cqlite:pushdown` field metadata → decode again; (b) arrow-flight version matrix (53.4.1 vs the release line matched to arrow-java 18). Report which single variable flips it.

**Phase 3 — fix at the identified boundary.** If (a): move pushdown capability signaling out of per-field IPC metadata (schema-level metadata or FlightData app_metadata) — coordinate the connector-side reader of that metadata in the same PR. If (b): pin/upgrade the version pair + add the interop test to CI. **Do NOT hand-roll framing under any outcome — the stock encoder is correct.**

**Phase 4 — observability hardening (small, do in the same PR as whichever fix lands, or standalone if Phase 1 does NOT reproduce):** register `cqlite.errors.total` **eagerly at 0** on startup when the observability feature is on, so "metric name absent" can never again be ambiguous between "no errors" and "not wired". (Post-END_STREAM peer-reset visibility lives at the h2/tonic layer and is OUT of scope — document it in the metric's doc comment.)

**If Phase 1 does NOT reproduce**: do NOT close. Report, then the remaining axes are environmental (x86_64 image build, Trino-runtime arrow-java version, live gRPC vs replay) — escalate to the owner with the Phase-1 evidence for a field-side live matrix; the issue closes only on the owner's round-4 criteria.

## Acceptance criteria
- [ ] Phase-1 oracle exists and runs in CI: Java Flight-level decode of committed Rust-emitted FlightData (+ Rust sequence pin). This closes the exact blind spot that let this bug survive a green suite.
- [ ] Phase-2/3 executed if reproduced; single-variable cause identified; fix lands with the oracle green.
- [ ] `cqlite_errors_total` eagerly registered (visible at 0 in a metrics scrape of a freshly started server, observability feature on).
- [ ] All existing transport/golden tests green; `scripts/agent-gate.sh` PASS.
- [ ] Final field closure = owner's round-4 criteria on a post-fix build: tiny read returns 3 rows OR ERROR log + `cqlite_errors_total` increment. Issue stays open until that field verdict.

## Routing / scope
Oracle-driven (no OpenSpec). Scope: cqlite-flight (golden emitter, eager metric registration, possibly producer.rs:425-449 metadata placement), trino-connector tests (+ metadata reader if Phase 3a). OUT of scope: #2264 (cancellation stall — separate bug, do not touch streaming backpressure), hand-rolled framing, channel/batch tuning.

pmcfadin:
Part of #2193 / Epic AM #2226 — **hardening, not the #2193 fix**. This bump was started when round-5 evidence pointed at arrow-java 18↔arrow-rs 53 skew; the true root cause landed mid-flight (missing `--add-opens` JVM flag → #2290, whose PR closes #2193). The work stands on its own merits and lands as hardening:

## What
- `arrowVersion` 18.1.0 → **19.0.0** (the line Trino 481's own arrow-consuming plugins use) — one variable, all arrow modules ride it transitively; bundle verified (zero stale 18.1.0 jars in `build/plugin/cqlite_flight/`).
- **netty pinned to 4.1.130.Final** via enforced BOM: flight-core 19's metadata drags netty **4.2.9** by conflict resolution, and netty 4.2's `EmptyByteBuf.memoryAddress()` throws inside `NettyAllocationManager` static init — killing every `RootAllocator` (28 test failures when unpinned). 4.1.130 is arrow-19's tested baseline. Fixed honestly, no suppressions.
- **Version-drift guards** (`ArrowJavaVersionPinTest`, 7 tests): class-anchored checks on the exact classes that broke (`RootAllocator`, `NettyAllocationManager` — loaded reflectively, plus the buffer-patch artifact verified by real class) AND an exhaustive classpath scan asserting EVERY `org.apache.arrow:*` == 19.0.0 (min 6) and every `io.netty:*` == 4.1.130.Final (min 8, tcnative excluded as independently versioned), via each jar's own `pom.properties`. Verified non-vacuous.
- Fail-first golden slot documented in `FlightDataGoldenDecodeTest` for field-captured FlightData bytes (not fabricated; pcaps exist with the field team).

## Verification
- `./gradlew clean test`: **312/312** (was 305), incl. `FlightDataGoldenDecodeTest` + `ArrowToTrinoGoldenTest` decoding cleanly under 19.
- Review-first: roborev 1579 (allocator-module gap → fixed), 1580 (exhaustive-set gap → fixed).
- Note: arrow-java 19 carries the same `--add-opens` requirement as 18 — no deployment delta from this PR; the flag work is #2290.

### DIFF
--- a/trino-connector/build.gradle.kts
+++ b/trino-connector/build.gradle.kts
@@ -31,12 +31,34 @@ repositories {
 // at runtime from the engine classpath; bundling it would clash. It must stay
 // out of the published POM's runtime dependencies for the same reason.
 val trinoVersion = "481"
-val arrowVersion = "18.1.0"
+// arrow-java pinned to 19.0.0 to align with Trino 481's own arrow-consuming
+// plugins (pinot/bigquery) and to fix the round-5 field decode failure
+// ("Failed to read message" on every table read) under 18.1.0 (issue #2193).
+// `flight-core` is the sole arrow artifact declared; every other arrow module
+// (arrow-vector, arrow-memory-core/-netty, arrow-format, flight-grpc) rides in
+// transitively at this SAME version, so bumping this one variable moves the
+// whole arrow-java stack in lockstep. Enforced at runtime by
+// ArrowJavaVersionPinTest against a silent dependency-resolution downgrade.
+val arrowVersion = "19.0.0"
 val jacksonVersion = "2.18.2"
+// Netty is pinned to the 4.1.x line arrow-java 19.0.0 was built and tested
+// against (issue #2193). flight-core:19.0.0's published Gradle metadata drags
+// several netty modules UP to 4.2.9.Final via conflict resolution, but arrow's
+// own `arrow-memory-netty-buffer-patch:19.0.0` and `grpc-netty:1.79.0` both
+// request netty 4.1.130.Final — and arrow's netty allocator is INCOMPATIBLE
+// with netty 4.2.x: its `UnsafeDirectLittleEndian.<init>` calls
+// `EmptyByteBuf.memoryAddress()`, which netty 4.2 changed to throw
+// `UnsupportedOperationException`, failing `NettyAllocationManager`'s static
+// init (`ExceptionInInitializerError`) on the FIRST RootAllocator — i.e. every
+// table read. The enforced BOM below forces the whole netty stack down to the
+// arrow-19-tested 4.1.130.Final so the allocator initializes. This mirrors a
+// real Trino runtime, which supplies netty 4.1.x.
+val nettyVersion = "4.1.130.Final"
 
 dependencies {
     compileOnly("io.trino:trino-spi:$trinoVersion")
 
+    implementation(enforcedPlatform("io.netty:netty-bom:$nettyVersion"))
     implementation("org.apache.arrow:flight-core:$arrowVersion")
     implementation("com.fasterxml.jackson.core:jackson-databind:$jacksonVersion")
 
--- a/trino-connector/src/test/java/in/mcfad/cqlite/flight/ArrowJavaVersionPinTest.java
+++ b/trino-connector/src/test/java/in/mcfad/cqlite/flight/ArrowJavaVersionPinTest.java
@@ -9,6 +9,7 @@ import java.security.CodeSource;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.apache.arrow.flight.FlightClient;
+import org.apache.arrow.memory.RootAllocator;
 import org.apache.arrow.vector.VectorSchemaRoot;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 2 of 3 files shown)


pmcfadin:
flow-closer endgame record (telemetry gap documented honestly)

**Gate of record:** PASS — commit b7c487c0, branch issue-2193-arrow-java-bump, all 24 components PASS. Two prior gate attempts were discarded and NOT counted as the record: (1) first full run FAILed on parity/golden tests (`ttl_table` minLocalDeletionTime, row-count parity) because the shared local fixtures were stale v3.4 while the rebase pulled main's v3.5 TTL-free regenerated goldens (#1935/#2249) — resolved by `fetch-datasets.sh` to v3.5, after which the two components PASS in isolation; (2) a relaunched run passed but ran on the local `main` checkout (a persisted `cd` moved the cwd) — invalid tree, re-run in the worktree.

**roborev:** clean on arrival. Job 1581 (closer pass) = 1 Medium nit (netty enforcedPlatform BOM may not propagate to the published Maven POM for Maven consumers) → follow-up #2300, non-blocking. Prior review-first rounds 1579 (2: Medium+Low) and 1580 (1: Medium) were fixed pre-gate (2 rework rounds).

**Telemetry NOT double-stamped:** #2193 already has a delivery-telemetry record (PR #2284, the original flight-java-decode fix). This hardening PR is a follow-on to the same issue; forcing `--allow-duplicate` would double-count #2193 and skew retro baselines, so the observed counters are recorded here instead: gate_runs_full=3 (1 stale-fixture FAIL, 1 wrong-cwd invalid, 1 PASS of record), rework_rounds=2, rebase_events=1, roborev findings 4 = 3 blockers (1579/1580, fixed) + 1 nit (1581 → #2300), claim_collisions=0.

#2193 remains OPEN; closure rides #2290.

Issue #2308, opened 2026-07-10
Duplicate projection columns collapse in name-keyed QueryRow (SELECT a,a / SELECT a AS x, b AS x)


pmcfadin:
Spun out of #1495 review (duplicate-column reachability analysis).

## Problem
`QueryRow.values` is `HashMap<Arc<str>, Value>` (`cqlite-core/src/query/result.rs:77`) — keyed by column NAME. But `QueryMetadata.columns` is built one `ColumnInfo` per projection expression with NO dedup (`query/select_executor/mod.rs::get_result_columns` ~1283-1397; parser `select_parser.rs:748-773`; no validator guard in `m2_select_validator.rs`).

So a projection with duplicate output names collapses upstream:
- `SELECT a, a FROM t` → metadata has two `ColumnInfo{name:"a"}`, but the row map holds ONE `"a"` slot.
- `SELECT a AS x, b AS x` → row builder inserts by name and the second **overwrites** the first (`select_executor/mod.rs:1059-1065`), so the row ends up `{"x": <b>}` — a's value is LOST.

Neither behavior matches real Cassandra, which carries each projection expression's own value independently. This is a **model-level** gap (name-keyed rows cannot represent two columns that share a name but differ in value), predating #1495.

## Impact
- `SELECT a AS x, b AS x` silently drops a column value (overwrite).
- Any consumer that reads by position / duplicate name (Arrow/Parquet export, CLI) sees collapsed or duplicated data.
- Low real-world frequency (duplicate output names are unusual), hence P3 — but it is silent wrong data when hit.

## Fix direction (design-driven — needs a decision)
Represent result rows POSITIONALLY (parallel to `metadata.columns`) rather than name-keyed, OR reject/disambiguate duplicate output names at projection time. Cross-links the positional-row rework (E2 / #1584 / #1817). Route through OpenSpec.

## Note
#1495 (Arrow accessor refactor) preserves main byte-for-byte for this input (value replicated to all same-named columns) — it does NOT fix or worsen the underlying collapse; that is this issue.

Oracle-vs-design: **design-driven** (touches the row model / projection contract).

Issue #1496, opened 2026-07-02
AE2 — arrow_convert: capacity-hinted builders, drop per-value clones + double-Vec blob


pmcfadin:
Part of epic #1470 (per-cell cost). Source: finding AE2. Pure perf refactor.

## Problem
Per-value clones + throwaway intermediate `Vec`s + capacity-less typed builders:
- Text arms clone every cell: `Value::Text(s) => Some(s.clone())` (`arrow_convert.rs:537, 1345`).
- Blob builds `Vec<Option<Vec<u8>>>` (clone per cell) THEN a second `Vec<Option<&[u8]>>` THEN the array (`:546-558`).
- Inet/Duration clone+format per cell (`:702, 1579`).
- Typed builders created with no capacity: `Decimal128Builder::new()`, `FixedSizeBinaryBuilder::new(16)` (`:605, 679, 1453, 1553`) → they reallocate as they grow.
- Scalar arms collect into `Vec<Option<T>>` then `T::Array::from(vec)` — an extra full-length intermediate per column.

## Fix (effort: M)
- Append `&str`/`&[u8]` directly into `StringBuilder`/`BinaryBuilder` sized with `rows.len()` capacity; drop the double-Vec for Blob.
- Pass `with_capacity(rows.len())` to all typed builders.
- Remove the intermediate `Vec<Option<T>>` where the builder can be appended to directly.
Behavior-preserving; AD parity must stay green.

## Acceptance criteria
- [ ] String/blob columns allocate ≈ rows (not 2–3×); typed builders are capacity-hinted.
- [ ] Output byte-identical; gate PASS.

## TDD test (must FAIL on current main)
Alloc-budget test (epic-H) on a string/blob column asserting allocation count ≈ rows, not 2–3×; criterion throughput bench. On main it is 2–3× with reallocating builders.

pmcfadin:
Promoted 0.16 → 0.14 per owner decision (2026-07-09): arrow_convert is Flight's data plane / bench coverage for it — pulled into the Flight field-readiness release.

pmcfadin:
📋 Routing (lead): **oracle-driven** — no OpenSpec / Seam-1 needed. Internal perf refactor with existing oracles: the gate's `arrow-parity-guard` + compaction-byte-parity guard correctness; pair the win with a pinned bench/alloc observation per the #1494 bench-suite conventions once it lands (#1494 is being pre-approved now so it won't block you). Claim + implement directly.

[...]

Issue #1470, opened 2026-07-02
Epic AE — Per-cell conversion cost (export-side parser J1 instance)


pmcfadin:
**Epic — Per-cell conversion cost (export-side instance of parser epic J1)**

The conversion hot paths repeat per-cell work that should be resolved once per column/schema — the
export-side manifestation of the string-typed per-cell dispatch that parser epic J1 condemned. Shared
fix shape: **resolve a per-column accessor/index once, then index rows positionally**; add
capacity hints; stop cloning per cell.

Source of truth: `docs/reports/export-path-performance-audit-2026-07-01.md` (findings AE1–AE6).
Cross-link parser epic J1 in each PR.

Children:
- AE1 — per-cell `HashMap<String,Value>` string-hash lookup in every Arrow builder
- AE2 — per-value clones, double-Vec blob path, capacity-less typed builders
- AE3 — delta emit rebuilds 3 schema-constant HashSets per record; O(R·V·cells) linear cell dispatch
- AE4 — `ColumnId::new` clones column name per surviving cell
- AE5 — JSON/CSV per-cell churn (key clone per row, fragile `== "null"` sentinel, 16-arg UUID format!)
- AE6 — Node `exportParquet` runs the sync Parquet writer on the async executor

Definition of done for each child: `scripts/agent-gate.sh` PASS; a work-counter/alloc-budget test
(epic-H machinery) proves the per-cell work dropped to O(columns) or O(rows) as appropriate and fails
on `main`; a criterion throughput bench accompanies it. NO behavior/output change — these are pure
performance refactors, so the AD parity tests must still pass byte-identically.

## Children (tracking)
- [ ] #1495 — AE1 per-column accessor
- [ ] #1496 — AE2 capacity-hinted builders
- [ ] #1497 — AE3 delta emit sets + cell dispatch
- [ ] #1498 — AE4 intern column names
- [ ] #1499 — AE5 CLI json/csv churn + `== "null"` bug
- [ ] #1500 — AE6 Node exportParquet spawn_blocking

pmcfadin:
Milestoned 0.16 per owner-approved board grouping (2026-07-09): epic parent aligned with its children's milestone so milestone views show the whole program.

pmcfadin:
[...]

Issue #2264, opened 2026-07-09
cqlite-flight do_get: active snapshot-index re-load/glob loop ignores cancellation — LIMIT queries hang, in_flight gauge stuck (round-5 World-2)


pmcfadin:
Part of Epic AM (#2226). **P1 — the ONE remaining field blocker** after round 5 (#2286): with the #2193 JVM-flag fix in place the entire read path works, EXCEPT `SELECT * FROM keyvalue LIMIT 5` still hangs (>110s, killed manually).

## The mechanism is now field-proven (round-5 thread dumps — do NOT re-litigate the old channel-park theory)

Two eu-stack/gdb dumps 30s apart on the hung flight process, plus metric deltas, establish **World 2 — an ACTIVE internal loop, not a parked thread**:
- Hot frames: **158× glibc `ext_match` + 41× `build_wcs_buffer`** (fnmatch/glob internals), 16× `__res_context_send`; tokio workers idle in epoll.
- `rpc_requests_total{method="do_get"}` UNCHANGED across the ~90s churn → ONE request looping internally, NOT Trino split retries.
- Debug logs show the SAME 414,957-entry snapshot index re-loaded repeatedly.
- The previously-shipped cancel fix (PR #2282: token-racing the channel send) is real but targets the wrong layer for THIS hang: a synchronous filesystem-glob/index-load loop never reaches the channel send, and only polls `cancel.is_cancelled()` between merge steps it never completes.
- Likely-related field observation: the `keyvalue` Sidecar snapshot dir is missing `Index.db`/`Summary.db`/`Filter.db` (only Data.db + header present), and flight WARNs `No partition keys found in SSTable header`. **Prime suspect: missing Index/Summary in the snapshot forces an index rebuild from Data.db on every open, and something opens per-step/per-partition → O(steps × 414k).**

Artifacts (field team has locally, to be attached here): `r5-threaddump-1.txt`, `r5-threaddump-2.txt`, `r5-gdb-bt.txt`; correlated debug log `flight-debug-tiny-r5-full.log` region.

### Phase A — root-cause the loop (read-only investigation, report before fixing)
1. Find the glob: search cqlite-flight + cqlite-core for directory-listing/glob/fnmatch use on the do_get read path (snapshot dir resolution, SSTable component discovery, `Data.db` sibling lookup). The 158× `ext_match` frames mean a glob/fnmatch runs in a hot loop — identify the call site and WHAT ITERATION drives it (per merge step? per partition? per split?).
[...]

PR #2312, opened 2026-07-10
perf(#1495): resolve per-column Arrow accessor once (kill per-cell HashMap lookup)

### LINKED ISSUE #1495: AE1 — arrow_convert: resolve per-column accessor once (kill per-cell HashMap lookup)
Part of epic #1470 (per-cell cost). Source: finding AE1. **Cross-link parser epic J1 in the PR.** Pure perf refactor — output must not change (AD parity guards it).

## Problem
Every Arrow builder loops rows and does `row.values.get(&col.name)` **per cell** — a `HashMap<String, Value>` sip-hash string lookup. For N rows × M columns that is N·M string hashes, even though the column name is known once. This is the export-side instance of the string-typed per-cell dispatch parser epic J1 condemned.

**Files:** `QueryRow.values` is `HashMap<String, Value>` (`cqlite-core/src/query/result.rs:68`). Per-cell `.get()` at `cqlite-core/src/export/arrow_convert.rs:1148, 1241, 1255, 1283, 1298, 1343, 1359`.

## Fix (effort: M) — shared J1 fix shape: resolve per-column accessor once
Resolve each column to a positional index / precomputed accessor **once per schema**, then index rows positionally instead of re-hashing the name per cell. Options:
- Give the builder a resolved accessor closure per column, or
- Represent rows column-major, or
- At minimum hoist a precomputed hash / `RandomState` so the per-cell cost drops.

Keep it behavior-preserving; the AD1/AD2 parity tests must still pass byte-identically.

## Acceptance criteria
- [ ] Per-cell string-hash lookups reduced to O(columns) resolution, not O(rows×columns).
- [ ] Output byte-identical (AD parity green).
- [ ] Gate PASS.

Work-counter test asserting `values.get`/hash invocations scale with columns (once per column), not rows×columns; plus a criterion export-throughput bench. On main it is per-cell. Reuse epic-H work-counter machinery.

pmcfadin:
Closes #1495

## What
AE1 (epic #1470, per-cell cost). Pure perf refactor of Arrow export — **output byte-identical** (parity-guarded). Every Arrow builder previously did `row.values.get(&col.name)` per cell — an N·M `HashMap<Arc<str>,Value>` siphash. Now each schema column is resolved to its output index(es) ONCE via `arrow_columnar::transpose_columns`, then a single pass routes each present cell to its column slot(s). Per-cell name-hashing is gone (`rg 'values\.get\(' arrow_convert.rs` = 0).

## Correctness (byte-identity is the contract)
- New parity guard `cqlite-core/tests/issue_1495_arrow_accessor_parity.rs`: wide mixed-type RecordBatch identity, absent/null collapse, column-reorder (full `ArrayData` compare), type-mismatch fail-closed. Passed pre- AND post-refactor.
- **Duplicate result-column names** (e.g. `SELECT a, a`, `SELECT a AS x, b AS x`): a row value now fans out to ALL same-named columns (restores main's behavior — the last-wins transpose had nulled earlier duplicates). Pinned by `duplicate_named_columns_populate_all_identically` (RED before fix, GREEN after). The deeper name-keyed-row collapse is pre-existing and tracked separately in #2308.
- **Gate wiring:** new `arrow-parity-guard` gate component runs the parity test with `--features arrow` (it's arrow-gated and was otherwise skipped by every gate/CI lane); fails-closed on a 0-run skip.

## Scope
Accessor-resolution only. Does NOT fold in #1496 (AE2 capacity-hinted builders / per-value clones). Transpose peak-memory + reorder-test polish batched to a follow-up (bounded pointer matrix; streaming is #1470/#1496 scope).

Cross-links parser epic J1 (same per-cell dispatch-cost pattern).

### DIFF
--- a/cqlite-core/Cargo.toml
+++ b/cqlite-core/Cargo.toml
@@ -174,6 +174,15 @@ pprof = { version = "0.14", features = ["flamegraph", "criterion"] }
 [target.'cfg(target_os = "macos")'.dev-dependencies]
 mach2 = "0.4"
 
+# Arrow-conversion accessor-hoist parity guard (issue #1495, AE1). Declared with
+# required-features so the agent gate's blast-radius scoping compiles+RUNS it WITH
+# the `arrow` feature it needs — without this, `#![cfg(feature = "arrow")]` would
+# make it a vacuously-passing empty binary under the default `cli-helpers`-only
+# scoped-test invocation.
+[[test]]
+name = "issue_1495_arrow_accessor_parity"
+required-features = ["arrow"]
+
 [[bench]]
 name = "partition_lookup"
 harness = false
--- a/cqlite-core/Cargo.toml
+++ b/cqlite-core/Cargo.toml
@@ -175,10 +175,13 @@ pprof = { version = "0.14", features = ["flamegraph", "criterion"] }
 mach2 = "0.4"
 
 # Arrow-conversion accessor-hoist parity guard (issue #1495, AE1). Declared with
-# required-features so the agent gate's blast-radius scoping compiles+RUNS it WITH
-# the `arrow` feature it needs — without this, `#![cfg(feature = "arrow")]` would
-# make it a vacuously-passing empty binary under the default `cli-helpers`-only
-# scoped-test invocation.
+# required-features = ["arrow"] because it is `#![cfg(feature = "arrow")]`, so it
+# only compiles+runs when the arrow feature is on. The full agent gate runs it via
+# the dedicated `arrow-parity-guard` component (scripts/agent-gate.sh:
+# `cargo test -p cqlite-core --features arrow --test issue_1495_arrow_accessor_parity`),
+# which is its ONLY automated executor and fails closed on a vacuous 0-run — the
+# default `cli-helpers`-only core-tests run and pr-gate's `--lib --all-features`
+# both skip this integration target.
 [[test]]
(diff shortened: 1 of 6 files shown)


pmcfadin:
🧭 **flow-lead: gate status — PR is clean, blocked on a main-red**

Full gate of record ran (`/tmp/gate-1495c.txt`): **RESULT: FAIL, but NOT from this PR.** Every substantive component PASSED — including the new `arrow-parity-guard: PASS` (the parity guard now actually runs with `--features arrow`) and `memory-budget: PASS`. All of core/write/parity/bindings/arrow tests green.

Two FAILs, both diagnosed as NOT caused by this diff:
- `delivery-telemetry: FAIL` — **pre-existing hard main-red**: `docs/reports/delivery-telemetry.jsonl` has two records for issue #2264 (PR #2282 + PR #2301 both closed it), violating the one-record-per-issue lint. Reproduces identically on clean `origin/main`. Tracked in #2317 — blocks EVERY branch until fixed on main.
- `tooling-tests: FAIL` — pre-existing transient race in the parity-report SELF-test on a shared temp-manifest path; passes 7/7 on re-run and on clean main. Not this diff.

This PR (arrow accessor-once, byte-parity preserved, duplicate-name fix, gate-wiring fix) is merge-ready on its own merits. **Holding merge until the #2317 main-red is cleared** so the gate of record can go green. Will re-gate + merge once #2317 lands.

Issue #2319, opened 2026-07-10
Direct I/O windowed scan allocates a bounce buffer per chunk (regressed reused-window from #1940 substrate rework)


pmcfadin:
Follow-up from #1940 (window-as-Bytes substrate) review.

## Problem
After #1940, the windowed scan reads chunks via `point_source.read_exact_at` (`scan_stream_windowed_read.rs:110`). For the **Direct I/O backend**, that path is `DirectReadAt::read_at` (`read_at.rs:341`), which allocates a fresh aligned bounce buffer (`AlignedBuf::new`, ~chunk-sized, rounded to 4096) on EVERY chunk read.

On origin/main, the windowed *scan* for Direct went through a per-scan `DirectCursor` (`source.rs:445,497-518`) that reused ONE window `AlignedBuf` allocated at open. So for the **Direct windowed-scan path specifically**, #1940 traded a reused window buffer for a per-read bounce alloc.

## Why this did NOT block #1940
- The ≤1-alloc/chunk acceptance counter (`CHUNK_PATH_ALLOCS`) does not count the bounce buffer — it tracks the substrate copy chain only (`scan_stream_windowed_read.rs:226`, `block_io.rs:514`, `chunk_source.rs:141,178`). The bounce buffer is invisible to that contract by construction.
- The alloc lane (`issue_1940_scan_window_substrate_allocs.rs`) uses a tiny fixture → `DiskAccessMode::Auto` resolves to Mmap/Buffered, never Direct (Direct is only auto-selected for files > ½ RAM, or explicit opt-in).
- `DirectReadAt::read_at` itself is byte-identical vs main (no new code); the alloc predates the branch.

So it's an uncounted, untested, specialized-backend (non-default) perf detail — real but out of #1940's scope.

## Fix direction
Restore per-cursor bounce-buffer reuse for the Direct windowed-scan path so large-file direct scans regain ≤1-alloc/chunk steady state — e.g. give the feed a reusable aligned scratch for Direct, or route Direct scans through a synchronous reader that reuses `DirectCursor`'s aligned window instead of the generic `ReadAt` direct path. Add a Direct-backend variant of the alloc-invariant lane (force `DiskAccessMode::Direct`) so the invariant is actually guarded for that backend.

Oracle-driven (perf, work-counter measurable). Related: #1940, the alloc-budget epic (#1470/H2 counters).

pmcfadin:
[...]

Issue #1644, opened 2026-07-02
K5 — Zero-copy value extraction (Bytes-backed Value; joint Value-v2 train with E1/E3)


pmcfadin:
Part of epic #1604 (row/cell hot-loop mechanics). Source: `docs/reports/parser-performance-audit-2026-07-01.md` finding K5 (audit block 2). **JOINT: this is the decode side of the "Value representation v2" design train** (capstone §3) — read-path E1 #1583 owns the `Value` enum layout, E3 #1585 (window-as-`Bytes`) is the enabling substrate. **The three must be designed together and shipped as one train; do not start implementation before the joint design exists.** Wave 3.

## Problem
Every text/blob/varint/decimal value is **copied out of the decompressed chunk per cell**: `v5_compressed_legacy/cell_value.rs:303, :375` (e.g. `String::from_utf8(bytes.to_vec())`), plus the same pattern in `raw_value.rs` and `raw_type_value.rs`. A text-heavy scan pays one full payload copy (and for `String::from_utf8(to_vec())`, a throwaway pre-validation `Vec`) per value.

## Fix (effort: L)
1. **Joint design first** (with E1 #1583 + E3 #1585): `Value` variants for text/blob/varint/decimal become `Bytes`-backed slices of the decompressed chunk (refcounted subslice, zero copy), which requires the scan window to hand out `bytes::Bytes` (E3) and the enum layout to stay ≤ the E1 target. Write the design as part of the E1-owned design doc; this issue implements the DECODE side against it.
2. Decode side: replace `to_vec()`/`String::from_utf8(..to_vec())` extraction with `Bytes::slice_ref`/subslice of the chunk buffer; UTF-8 validation via `str::from_utf8` on the borrowed slice (validation without copy), storing the validated `Bytes`.
3. **Interim S-win (can land immediately, independent of the train):** where copies must remain for now, replace `String::from_utf8(bytes.to_vec())` with `str::from_utf8(bytes)?.to_owned()` — validates on the borrowed slice first, avoiding the throwaway `Vec` on the error path and one intermediate. Grep: `rg -n "String::from_utf8" cqlite-core/src/storage/sstable/reader/parsing/`.

### Do NOT
- Do not hold chunk-backed `Bytes` beyond the intended value lifetime without accounting for the retained-chunk memory (a 1-byte value pinning a 64KB chunk) — the design must address retention (e.g. copy-out threshold for long-lived values). This is a named design concern; get it into the E1 design review.
[...]

PR #2320, opened 2026-07-10
perf(#1940): window-as-Bytes substrate — ≤1 alloc/chunk on steady-state windowed scan (D2/Value-v2)

### LINKED ISSUE #1940: E3 follow-up: ≤1 alloc/chunk on the scan path — Arc<[u8]> scan channel / decompress-in-IO-half / mmap Bytes (co-design with E1/K5 Value-v2)
Follow-up to #1585 (E3). Owner decision (2026-07-04): E3 ships the A5 single-read-per-chunk win now (option C); the **≤1 alloc/chunk** target (E3's A4) is deferred here because it requires a hot-scan-seam rearchitecture that the E3 guardrails forbid.

## Why A4 (≤1 alloc/chunk) couldn't land in #1585
Steady-state windowed scan floors at **2 allocs/chunk** given E3's own guardrails:
1. The windowed scan feeds compressed chunks across an `mpsc::channel<Vec<u8>>` — `read_next_block`/`read_nb_format_chunk_data` return an **owned `Vec<u8>` per chunk for every backend** (no borrowed/`Bytes`/mmap-slice path). That owned compressed Vec (alloc #1) is allocated in the I/O half and freed in the parse half.
2. The B1 chunk cache must produce a cached `Arc<[u8]>` (alloc #3), and the guardrail says don't break B1.
3. Eliminating the fresh decompression `Vec` (alloc #2) via `decompress_into` is B1-safe and gets 3→2, but not ≤1.

## Scope (the restructure)
Reach ≤1 alloc/chunk by reworking the windowed-scan hot seam:
- Carry `Arc<[u8]>` (not owned `Vec<u8>`) through the scan channel.
- Decompress in the I/O half; read compressed bytes into a reused per-cursor scratch (buffered backend) or a **zero-copy mmap `Bytes` view** (mmap backend).
- Keep the B1 `Arc` cache; align the window-fill site with the cached `Arc`.

This is the **"window-as-`Bytes` substrate"** #1585 explicitly says to **design-review jointly with the E1 (#1583) / K5 (#1644) "Value v2" train** — do it as part of that co-design, not in isolation.

## Guardrails / oracle
- CRC before decompression trust, unchanged order; window bounded-memory discipline unchanged; `uncompressed_len` from CompressionInfo (authoritative, no guessing).
- Touches the #1156/#1567/#1741 hot scan seam → the 33-table byte-parity harness (all 4 compression algos) is the load-bearing net.

## Acceptance
- [ ] A4 dhat: allocations/chunk on a steady-state scan ≤1 (was ≥3; A5 already cut reads to 1).
- [ ] 33-table byte-parity green across LZ4/Snappy/Deflate/Zstd.
- [ ] No B1 cache regression; `scripts/agent-gate.sh` PASS.
- [ ] Co-designed/sequenced with E1 (#1583) / K5 (#1644).

pmcfadin:
Closes #1940

## What (Stage 2 / D2 of the Value-v2 train — spec: openspec/changes/value-representation-v2)
Reworks the windowed-scan hot seam to carry decompressed chunk bytes as a refcounted `bytes::Bytes` substrate, reaching **≤1 allocation/chunk** on the steady-state scan (was ≥2). Decompression moved into the IO/feed half; B1 chunk cache value `Arc<[u8]>`→`Bytes` (zero-copy insert, refcount-bump on hit).

## Concurrency (owner design decision)
The buffered-backend feed reads **synchronously** via the existing `point_source: Arc<dyn ReadAt>` pread plane — NO `tokio::fs`-under-`block_on` nesting (eliminated the blocking-pool-amplification hazard class rather than bounding it). Decompression stays off the async reactor on the `spawn_blocking` feed thread. Confirmed by rust-reviewer (every await traced) + a no-nesting regression guard that records at the real read site.

## Correctness / invariants (all guarded)
- CRC-before-decompress order preserved; corrupt chunk fails before parse.
- `uncompressed_len` authoritative from CompressionInfo (no-heuristics).
- Bounded-memory / window semantics unchanged; raw/uncompressed zero-copy passthrough.
- Scratch reserve BOUNDED to the largest actual chunk via `try_reserve_exact` (fixes a metadata-driven `max_compressed_length`=i32::MAX OOM — proven: largest single alloc 32 KiB vs ~2 GiB pre-fix; test `issue_1940_scan_window_no_metadata_oom.rs`).
- One-shot transient-I/O retry + `ErrorKind` preservation on the positional read path.
- 33-table byte-parity across LZ4/Snappy/Deflate/Zstd.

## Guards
- `issue_1940_scan_window_substrate_allocs.rs` — ≤1 alloc/chunk + E3 one-read-per-chunk.
- `issue_1940_decode_off_async_worker.rs` — decode+read OFF the async worker set (probe at real read site).
- `issue_1940_scan_window_no_metadata_oom.rs` — bounded alloc under i32::MAX metadata.

## Follow-ups (non-blocking)
- #2319 — Direct I/O windowed scan allocates a bounce buffer per chunk (uncounted/out-of-scope for the ≤1-alloc contract; specialized non-default backend).

Design-driven; spec-auditor C intent audit + roborev + full gate to run in the closer. Stage 3 (K5 #1644) builds on this.


pmcfadin:
### flow-closer endgame record (#1940)

**Gate of record (full):** `RESULT: PASS` at anchor `f74a2f20` — all 24 components PASS
(`/tmp/gate-1940.txt`). Header verified: branch `issue-1940-alloc-per-chunk-scan`, `dirty: no`.

**Post-gate polish re-cert (delta, test-only):** `RESULT: PASS` (MODE: delta) at `34c789c8`,
anchored to `f74a2f20` — `delta-files (2 allowed / 0 offending)`, `scoped-tests: PASS`. The only
change since the full gate is the two regression tests below (test/docs-only → `--delta`, never a
second full gate).

**`CQLITE_ALLOW_FILE_GROWTH=1` justification:** the branch adds ~10 lines to already-over-threshold
reader files (`block_io.rs` 2181L, `data_access/mod.rs` 911L). Splitting those is epic #1116's scope,
not this substrate change — flagged, not silently grown.

**roborev final pass (job 1630, codex):** 2 findings, triaged per `docs/development/roborev-severity.md`:

- **Low → FIXED pre-merge** — the two new regression tests
  (`issue_1940_scan_window_no_metadata_oom.rs`, `issue_1940_decode_off_async_worker.rs`) assumed a
  cold B1 chunk cache without guaranteeing it, so a warm cache could bypass the decompress/reserve
  path and let the guards pass vacuously. Fix (commit `34c789c8`, test-only): disable
  `memory.block_cache` in both setups so the allocation/decompress path is always exercised
  (never-vacuous doctrine).

- **Medium → DECLINED as blocker, batched to hardening follow-up.** roborev flags that the upfront
  `try_reserve_exact(want)` in `scan_stream_windowed_read.rs` still reserves from
  `CompressionInfo`-derived metadata before the per-chunk bounds check. **Evidence it is not
  shipped-broken:** `try_reserve_exact` fails **cleanly** (returns `Err`) on an unsatisfiable
  reserve — it cannot abort or OOM-kill, unlike main's pre-change plain `Vec` allocation which
  aborts the process. The branch is therefore **strictly better than main**; the residual (a
  transient large-but-available reserve before chunk validation) is a hardening polish
  (cap-the-initial-reserve or defer-until-validated), not a correctness/DoS regression. Batched into
  the nit follow-up issue rather than gating merge.

Issue #2286, opened 2026-07-09
Round 4 field run — coordination: pinned assets, run matrix, per-issue verdict checklist


pmcfadin:
Part of Epic #2103 (easy-db-lab harness + live runs). Single source of truth for the round-4 field run — assets, protocol, and what each result adjudicates. Field team: post ALL round-4 results as comments HERE (per-issue detail may additionally go on the issue it adjudicates). Supersedes the scattered round-4 notes across #2103 comments / Epic AM #2226 / #2157 / #2193.

## Pinned assets (authoritative — do not run with anything else)
| Artifact | Value |
|---|---|
| cqlite-flight image | `ghcr.io/pmcfadin/cqlite-flight:dev` @ `sha256:1e313d51c17803917bdc8c0999c2dd6466ca47f41bddccc955684fe86a532254` — the multi-arch INDEX digest (main @ `19b32461`, run 29036815987). ⚠️ Do NOT use `289f97e0…` — that is the arm64 CHILD manifest and breaks amd64 nodes (`exec format error`); see the round-5 report. |
| Trino connector | `in.mcfad:cqlite-trino:0.13.4` (Maven Central — published, run 29036879521 SUCCESS) |
| Kit | easy-db-lab kit with `--read-mode` flag (#2103 round-4 prep) |
| Cluster shape | same as rounds 1–3: 3× i4i.xlarge db nodes, 1× m5.xlarge app, Cassandra 5.0.x, nb-big/LZ4 |

## What this build contains (vs round-4's `50557839…` which predated ALL of it)
Flight/Rust: #2264 (cancellation-aware backpressure — the LIMIT-hang fix), #2228 (full-ring token semantic), #2240, #2193 error surfacing. Connector/Java (0.13.4 vs 0.13.3): #2227 (per-replica-host snapshots — the multi-node snapshot blocker), #2228 (Java side), #2229 (hide+warn + TIME), #2236, #2238, #2239, #2241 (split replica failover).

## Run matrix
1. **Default snapshot mode** (the main event — first-ever multi-node validation of #2227): full query sweep incl. `SELECT *`, LIMIT, filtered, aggregation, on `keyvalue` AND `tiny`.
2. **`--read-mode live`** pass: same sweep — isolates snapshot-placement from read-path behavior (#2157/#2193 closure criteria are defined on this mode).
[...]

Issue #2157, opened 2026-07-07
cqlite-flight do_get eager-merges — LIMIT/predicate not honored at execution (round-2 harness finding, #2135 insufficient)


pmcfadin:
## Round-2 harness finding (epic #2103): `do_get` still eager-merges — interactive queries unusable despite #2135

Second live 3-node AWS run of `test-plans/cqlite-flight-loadtest-3node.md` with the round-1 fixes in place (flight `:dev` = observability + LIMIT early-stop; connector `0.13.2` = `applyLimit`). **#2128 observability is fully fixed** (metrics + traces flowing — see separate note). **But the LIMIT/pushdown work landed in #2135 (closing #2129) did not fix interactive-query latency**, because the underlying materialize-then-emit producer (**#1476** / Epic AB **#1467**) is still in place.

### Data under test
`cassandra_easy_stress.keyvalue`: 3 SSTables/node post-compaction, ~2.1M partitions/node (~6.4M @ RF3), 418MB live/node.

### Symptom — every query shape is an unbounded full merge
| Query | Result |
|---|---|
| `SELECT * ... LIMIT 5` | ran **1445s (24 min)**, killed (never returned) |
| `SELECT count(*)` | ran **358s+**, killed |
| `SELECT * WHERE key = '<pk>'` (point read on partition key) | full-table scan estimate, no pushdown (see EXPLAIN) |

### Evidence

**1. LIMIT reaches the plan but not execution.** `EXPLAIN SELECT * ... LIMIT 5`:
```
TableScan[table = cqlite:CqliteFlightTableHandle[...limit=OptionalLong[5]]]
  Estimates: {rows: 2154312 (226.00MB), cpu: 226.00M ...}
```
The ticket carries `limit=OptionalLong[5]` (the #2135 / #1477 interim fix works), yet the scan still estimates and reads the full 2.15M rows — the early-stop never fires because `do_get` materializes the whole merge before emitting row 1.

**2. No predicate pushdown at all.** `EXPLAIN SELECT * WHERE key = '001.25.535688'` (key IS the partition key):
```
ScanFilter[table = cqlite:CqliteFlightTableHandle[...filterJson=Optional.empty, limit=OptionalLong.empty],
           filterPredicate = (key = varchar '001.25.535688')]
  Estimates: {rows: 2155592 (226.13MB) ...}
```
[...]

PR #2364, opened 2026-07-13
fix(#2361/#2264): stream the non-stitching scan path — bounded memory, effective LIMIT/cancel, joined producers, in-flight phase metric (round-7 field blocker)

### LINKED ISSUE #2361: cqlite-flight do_get: parallel scan/materialize ignores LIMIT + escapes #2301 cancellation — unbounded CPU/memory, full scans never complete (round-7 FAIL)
## Round-7 field FAIL — the sole blocker for #2227/#2228/#2241 closure

`SELECT * FROM keyvalue LIMIT 5` (and `count(*)`) against ~1.13M partitions/node × 3 nodes **never returns** (killed at 8m38s / 5m). Reproduces identically in **both snapshot and live** modes, with **complete snapshot components** (#2295 confirmed fixed in the field — Index/Summary/Filter all present). Build: round6 image `d8c302b9…` (main `6e106d72`, carries PR #2301), connector 0.13.5, Trino 481.

**This is a NEW mechanism, not the R5 glob loop that #2264/PR #2301 (cooperative scan cancellation, merged 0b6fe254) fixed:**

| Signal | R5 (#2264) | R7 (this issue) |
|---|---|---|
| Threads | single synchronous thread in fnmatch/glob (`ext_match`) | **many tokio-rt-worker threads all in R state**, pod pinned ~3.7–3.9 cores |
| Memory | flat | **unbounded growth** (2.6→5.5 GB and rising) |
| perf self-time | glob/fnmatch | cqlite-flight userspace + kernel page alloc (`alloc_anon_folio`, `vma_alloc_folio`, `alloc_pages_mpol`) |
| Phase metrics | — | `cqlite_rpc_phase_duration_seconds_sum` ~0.01–0.17s while wall time is minutes → **hot path is UNINSTRUMENTED** (not resolve/merge_setup/stream) |
| Requests | — | `in_flight_ratio{do_get}=8` steady; `requests_total{do_get}` frozen at 28 → one request looping internally |

**Cancellation is NOT honored end-to-end:** killing the Trino query drains `in_flight_ratio` 8→0 (~40s), but the pod **keeps burning ~3.7 cores and allocating memory for minutes after** — only a DaemonSet restart reclaims it. #2301's cancellation apparently reaches the channel-send layer but not the parallel scan/merge/materialize workers (the gap R5 already flagged).

## Acceptance criteria
1. **Root cause identified in code**: where the do_get parallel scan/materialize path (a) ignores LIMIT, (b) detaches from the #2301 cancellation token, and (c) buffers unboundedly (likely an unbounded channel between scan producers and the stream consumer — cf. the platform-audit "unbounded-channel scan" landmine).
2. **LIMIT-bounded scan or early-abort**: `LIMIT 5` on a multi-million-partition table completes without materializing the full scan (or aborts producers once the limit is satisfied downstream).
3. **Cancellation reaches the workers**: after client cancel, CPU + memory return to baseline without a process restart — covered by a test.
4. **Phase instrumentation added** around the parallel-scan/materialize step (field escalation item 2), so the next field capture is precise.
5. **Bounded memory**: scan path holds bounded buffers regardless of table size (field escalation item 3).
6. Repro/regression coverage: a local test (testbed-scale) proving producers stop on cancel + on limit-satisfied; 0-rows-when-present rules apply.

Oracle-driven (field repro + metrics oracle) — no OpenSpec.

## Links
- Field report: #2286 (round-7, 2026-07-12)
- Field-verdict tracker: #2157 (stays open; closes only on a passing field LIMIT verdict)
- Prior fix (different mechanism, stays closed): #2264 / PR #2301 (`0b6fe254`)
- Blocks field closure of: #2227 (multi-node snapshot full-ring), #2228 (full-range SELECT *), #2241 (killed-pod failover)

pmcfadin:
Fixes #2361. Field-facing anchor: #2264 (reopened — closes on field confirmation, not merge). Verdict tracker: #2157. Field evidence: #2286 round-7 report.

## Root cause
`SSTableReader::stream_all_partitions_for_compaction`'s NON-chunk-stitching branch (uncompressed BIG, `V5_0Uncompressed` — the field snapshot format) materialized the **entire SSTable into one sorted Vec before emitting a single row** (`iterate_all_partitions_cancellable` → `full_index_scan.rs` Vec accumulation). At 1.13M partitions × 8 concurrent do_gets: full scans never complete, memory 2.6→5.5GB unbounded, LIMIT can never fire (consumer-side only, blocked in `recv()`), cancellation reaches the plumbing but producers were detached threads and the multi-GB Vec was never reclaimed. The `stream` phase metric recorded only on completion — a hung merge was invisible (the field's ~0.1s-instrumented-vs-minutes-wall signature).

## Fix (4 items)
1. **Streaming full-index walk** (`full_index_stream.rs`, new): emits each partition as the index walk resolves it. All offset-table-computable gates hoisted up-front (safe `FellBack` → sequential fallback, nothing emitted); mid-walk failures are hard fail-closed `Error::Corruption` — a streamed partition can never be followed by a silent fallback re-scan. Bounded memory, immediate first emit, backpressure restored.
2. **Cancel-aware teardown** (`merge/mod.rs`): producer threads are now JOINED on merger drop (were detached — post-cancel CPU/memory burn in the field), ordering cancel → drop receiver (wakes a send-parked producer) → bounded join; consumer `next()` uses a cancel-aware poll so `drive_merge` observes cancel while blocked in `step()`.
3. **Token-monotonicity guard** (fail-closed): the streaming walk verifies per-partition Murmur3 token order before handing rows to the k-way merger (the materializing sibling had a defensive sort; streaming can't re-sort). Equal tokens pass; strict decrease = `Error::Corruption`.
4. **In-flight phase visibility** (`obs.rs`): `cqlite.rpc.phase.active` up/down counter (concurrency-safe under the field's 8-in-flight shape) — a wedged `stream` phase now shows as a standing level (field escalation item 2).

**LIMIT semantics — tried and removed:** a per-producer partition budget was implemented, then REMOVED after two independent reviews (rust-reviewer + roborev job 1648) converged on the same blocker: LIMIT counts surviving post-reconciliation rows, and partitions ≠ rows (predicate filtering is consumer-side; tombstoned/shadowed partitions emit nothing) — a budget can under-return, and `limitGuaranteed=false` permits over-return, never under-return. LIMIT is enforced by the consumer's post-reconciliation break + item-2 teardown aborting producers promptly — satisfying the "aborts producers once the limit is satisfied downstream" arm of AC2 with zero under-return risk.

## Test evidence
- Red proof: new-API tests compile-red on base (established convention) + behavioral pins only the fix satisfies.
- **Regression pins**: `stream_limit_returns_k_matches_even_when_concentrated_past_limit_index` + `stream_limit_over_shadowed_data_returns_k_surviving_rows` (matches/survivors placed in real token-order TAIL via `decorated_key` — return 0 under any reintroduced producer budget); `first_batch_available_before_merge_completes` + `slow_consumer_bounds_produced_batches` (anti-materialization at the Flight boundary); teardown join tests incl. genuine parked-in-send drop; monotonicity helper unit-tested 4 ways; phase-counter concurrent-overlap tests.
- Review history: 2 rounds — round 1 found 2 blockers (the budget under-return + a false-proof deadlock test), both fixed; delta re-review verdict CLEAN-FOR-PR.

## Scope flags
- **file-size**: grows three PRE-EXISTING over-threshold files (`merge/mod.rs`, `producer.rs`, `data_access/mod.rs`) — full gate runs with `CQLITE_ALLOW_FILE_GROWTH=1`; splitting is epic #1116, out of a P1 hotfix's scope.
- **Follow-ups filed**: nit batch (see linked issue) + the 1.13M-random-index-reads full-scan throughput cliff (streaming fixes the hang/memory; a full unbounded scan is still O(partitions) random reads).
- Compaction/point-read/aggregate call sites verified unchanged (None-routing airtight per review); chunk-stitching path untouched.

## Gate
Lite PASS (`2f45f461`) each round; the ONE full gate of record runs in flow-closer pre-merge.

### DIFF
--- a/cqlite-core/src/observability/catalog.rs
+++ b/cqlite-core/src/observability/catalog.rs
@@ -522,6 +522,18 @@ pub const RPC_BYTES: &str = "cqlite.rpc.bytes";
 /// carries a ticket, key, token range, or query-text attribute.
 pub const RPC_PHASE_DURATION: &str = "cqlite.rpc.phase.duration";
 
+/// `cqlite.rpc.phase.active` — gauge `1` (issue #2361).
+///
+/// In-flight visibility of the phase a `do_get` is CURRENTLY executing, set to 1
+/// on phase entry and back to 0 on exit (via [`super::super`]'s `PhaseTimer`
+/// transition/`Drop`). [`RPC_PHASE_DURATION`] only records a sample once a phase
+/// COMPLETES, so a `do_get` wedged forever in `stream` (the #2361 hang: a merge
+/// that never returns a batch) recorded NOTHING — this gauge shows `stream = 1`
+/// for the entire hang, so a stall is observable BEFORE completion. Bounded
+/// attributes: [`attr::RPC_METHOD`], [`attr::RPC_PHASE`] (the closed three-value
+/// set) — low cardinality (methods × 3 phases). NEVER a ticket/key/query value.
+pub const RPC_PHASE_ACTIVE: &str = "cqlite.rpc.phase.active";
+
 /// `cqlite.warm.cache.hits` — counter `{1}` (issue #2310).
 ///
 /// Flight warm-handle cache hits: a request whose probed SSTable generation set
@@ -604,6 +616,8 @@ pub const ALL_METRICS: &[&str] = &[
     RPC_BYTES,
     // In-progress read/query metrics (#2162)
     RPC_PHASE_DURATION,
+    // In-flight phase gauge (#2361)
+    RPC_PHASE_ACTIVE,
     // Flight warm-handle cache (#2310)
     WARM_CACHE_HITS,
     WARM_CACHE_MISSES,
@@ -674,6 +688,15 @@ mod tests {
         assert!(ALL_METRICS.contains(&RPC_PHASE_DURATION));
         assert!(RPC_PHASE_DURATION.starts_with("cqlite."));
         assert!(attr::RPC_PHASE.starts_with("cqlite."));
+    }
+
+    #[test]
+    fn rpc_phase_active_gauge_is_registered_and_namespaced() {
+        // Issue #2361: the in-flight phase gauge must be catalogued (so the
+        // registration/uniqueness checks cover it) and namespaced like the rest.
+        assert!(ALL_METRICS.contains(&RPC_PHASE_ACTIVE));
+        assert_eq!(RPC_PHASE_ACTIVE, "cqlite.rpc.phase.active");
+        assert!(RPC_PHASE_ACTIVE.starts_with("cqlite."));
     }
 
     #[test]
--- a/cqlite-core/src/observability/otel.rs
+++ b/cqlite-core/src/observability/otel.rs
@@ -312,6 +312,7 @@ struct Instruments {
     memtable_rows: Gauge<i64>,
     compaction_lag: Gauge<i64>,
     rpc_in_flight: Gauge<i64>,
+    rpc_phase_active: Gauge<i64>,
     merge_producer_threads: Gauge<i64>,
 }
 
@@ -571,6 +572,11 @@ fn instruments() -> &'static Instruments {
                 .with_unit(catalog::unit::DIMENSIONLESS)
                 .with_description("Arrow Flight RPCs currently being handled.")
                 .build(),
+            rpc_phase_active: m
+                .i64_gauge(catalog::RPC_PHASE_ACTIVE)
+                .with_unit(catalog::unit::DIMENSIONLESS)
+                .with_description("do_get phase currently executing (#2361).")
+                .build(),
             merge_producer_threads: m
                 .i64_gauge(catalog::MERGE_PRODUCER_THREADS)
                 .with_unit(catalog::unit::THREADS)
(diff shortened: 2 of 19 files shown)


pmcfadin:
**roborev job 1651 (final re-check) — one Medium, ADJUDICATED DECLINED with evidence:** the finding argues the streaming (token,key) order guard is stricter than `SSTableWriter`'s contract (writer enforces token order only, so an equal-token/reverse-key file would be flagged corrupt). Adjudication: (1) real Cassandra SSTables cannot trigger the guard — the partition index is DecoratedKey-ordered (token, key) by construction; (2) removing the validation (the suggested fix) would silently emit out-of-order partitions to the k-way merger — the streaming path cannot re-sort, so fail-closed is the correct behavior per doctrine; (3) the actual gap is writer-side (contract narrower than format) — filed as #2368 (writer canonicalization + regression test + the in-code adjudication comment, added there to avoid re-running the gate of record for a comment-only change). The prior Medium (job 1650, missing tie-break) is FIXED in 99e3d9b5 and did not re-raise. — flow-lead

Issue #2366, opened 2026-07-13
cqlite-flight full scans on non-stitching SSTables: O(partitions) random index reads — replace with sequential streaming pass (throughput cliff behind #2361)


pmcfadin:
Spun out of #2361 (round-7 scan hang). The streaming fix (PR #2364) removes the hang/unbounded-memory failure: full scans now emit per partition with bounded buffers and effective cancel. But the non-chunk-stitching full-scan walk still performs **one random Index.db read + Data.db parse per partition** — ~1.13M random reads per SSTable per full scan at field scale. That's a throughput cliff (minutes-scale full scans), now visible as sustained `stream` phase activity rather than a hang.

## Acceptance criteria
1. Full-partition enumeration on non-stitching SSTables uses a **single sequential streaming pass** over Data.db (or an equivalent O(sequential-IO) strategy), preserving token-order emission for the k-way merger and the fail-closed monotonicity guard from #2361.
2. Parity unchanged: sstabledump goldens + query-semantics oracle stay green; compaction/point-read paths untouched.
3. Field-shape benchmark evidence (testbed): full-scan wall time on a large uncompressed table before/after.

Related: #2361 (root cause + fix), #2362 (testbed uncompressed coverage), #2157/#2264 (field trackers).

pmcfadin:
Adopted into the **0.15 cqlite-trino theme** (epic #2403) — milestoned 0.15. Serves the trino latency/throughput/operations lanes (scoreboard #2399, acceptance = round-over-round field metrics). Board-linked, not re-filed.

pmcfadin:
🔒 Claimed by flow-lead worker (branch `issue-2366-sequential-fullscan-stream` pushed to origin). Routing: oracle-driven performance fix (sequential-streaming-pass acceptance criteria + sstabledump/query-semantics-oracle parity + benchmark evidence), proceeding straight to implement — no OpenSpec.

pmcfadin:
[...]

PR #2466, opened 2026-07-15
fix(#1885): v5 ladder decoder varint arm — decode CQL varint as Value::Varint not Blob

### LINKED ISSUE #1885: v5 ladder decode has no varint arm → CQL varint read as Value::Blob (mirror of I4 on the v5 side)

pmcfadin:
Closes #1885

## Problem (oracle-driven)
The v5 per-cell "ladder" decoder `parse_cell_value_schema_order` had **no `varint` arm**, so a CQL `varint` column read through the live v5 path fell through to the blob default → `Value::Blob` instead of `Value::Varint`. The block path already decodes it correctly (`ComparatorType::Varint => Value::Varint(value_data.to_vec())`) — that block path is the oracle.

## Fix
- Add a `CellKind::Varint` arm to the v5 ladder decoder reading the VInt-length-prefixed raw two's-complement big-endian bytes into `Value::Varint` (empty cell → `Varint([])`), byte-for-byte mirroring the block path and write side. No heuristics (#28).
- Add `CellKind::Varint` + `"varint"` mapping in `cell_kind.rs` (was falling to `Complex` → blob). `frozen<varint>` inner recursion also fixed.
- Consolidated the duplicated vint-length-prefix read into one `read_vint_prefixed_bytes` helper.

## Tests
- Flipped the #1617 decoder-lockstep varint case from recorded-divergence to converged.
- `v5_varint_arm_decodes_edge_cases`: zero, positive, negative, -1, > i64::MAX, < i64::MIN (num_bigint) — block==v5 + round-trip.
- `v5_varint_empty_matches_block`: exercises both zero-length-prefix and flags-only `[0x0C]` (`CELL_USE_ROW_TIMESTAMP | CELL_HAS_EMPTY_VALUE`) → `Value::Varint([])`.

## Review
- rust-reviewer: CLEAN. roborev (codex): 2 LOW, both addressed (DRY helper + empty-varint 0x0C coverage).

## Out-of-scope base-red unblock (flagged)
`full_index_stream_tests.rs` (from #2366) used `#[serial(work_counters)]` without importing `serial_test::serial`, leaving `cqlite-core --lib` compile-red on fresh origin/main. Added the one-line import. Its `windowed_stream_read_pattern_is_sequential` is separately load-flaky (#2366) — follow-up worthy.

### DIFF
--- a/cqlite-core/src/storage/sstable/reader/parsing/row_decoder/cell_kind.rs
+++ b/cqlite-core/src/storage/sstable/reader/parsing/row_decoder/cell_kind.rs
@@ -55,6 +55,11 @@ pub(super) enum CellKind {
     Inet,
     /// The literal CQL `blob` type (decodes to `Blob`, empty-value → `Blob([])`).
     Blob,
+    /// CQL `varint` (arbitrary-precision integer). Decodes the VInt-length-prefixed
+    /// raw two's-complement big-endian bytes to `Value::Varint` (empty-value →
+    /// `Varint([])`), mirroring the block / `ComparatorType::Varint` path exactly
+    /// (issue #1885).
+    Varint,
     /// Frozen / tuple / non-frozen-collection / marshal-UDT / unknown-scalar types:
     /// the already-lowercased declared type string, decoded by the retained string
     /// ladder. Empty-value → `Null` (matching the pre-J1 `_ => Null` empty arm).
@@ -88,7 +93,8 @@ impl CellKind {
             "time" => CellKind::Time,
             "inet" => CellKind::Inet,
             "blob" => CellKind::Blob,
-            // frozen<…>, tuple<…>, list/set/map<…>, marshal forms, varint, and any
+            "varint" => CellKind::Varint,
+            // frozen<…>, tuple<…>, list/set/map<…>, marshal forms, and any
             // unrecognized type: decoded by the retained string ladder. Store the
             // lowercased string so the ladder needs no per-cell `to_lowercase`.
             _ => CellKind::Complex(Arc::from(lowered.as_str())),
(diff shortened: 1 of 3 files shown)


pmcfadin:
**file-size ratchet ack (epic #1116).** The full agent-gate.sh passed every component except `file-size`, which flags that `cell_value.rs` grew. That file was **already 1251 lines on origin/main** (pre-existing over the 800 src threshold, tracked under epic #1116). This change adds a net **+13** lines there for the varint decode arm (already offset by a DRY consolidation, +35/−22) — a minimal in-scope addition, not new bulk. Splitting a 1264-line decoder is out of scope for a one-arm oracle bug fix. Re-certifying the gate with `CQLITE_ALLOW_FILE_GROWTH=1` per the documented campsite-rule ack; file split tracked by #1116.

Issue #2462, opened 2026-07-15
main broken: cqlite-core --lib missing serial_test::serial import after #2437/#2438 merge race


pmcfadin:
## What
`origin/main` at `710dde190` fails `cargo clippy`/`cargo test -p cqlite-core --lib`
(and therefore `scripts/agent-gate.sh` / `--lite`) for every worker on every machine:
```
error[E0433]: failed to resolve: use of undeclared crate or module `serial_test`
  --> cqlite-core/src/storage/sstable/reader/data_access/full_index_stream_tests.rs:326
   |
   #[serial(work_counters)]
```
(4 occurrences: lines 326/373/416/611.)

## Root cause
A merge race between two independently-green PRs:
- `088ad1960` (#2438) removed `use serial_test::serial;` from
  `cqlite-core/src/storage/sstable/reader/data_access/full_index_stream_tests.rs`
  (presumably as an unused-import cleanup, correct AT THAT COMMIT).
- `70bdc178e` (#2437) added 4 new `#[serial(work_counters)]` attributes to the SAME
  file (lines 326/373/416/611), branched BEFORE #2438's import removal, so #2437's
  own gate ran green.

Neither PR's individual diff was wrong in isolation; the merged result on `main` is
orphaned — #2437 rebased/merged after #2438 without picking up the import removal,
or vice versa, and neither round re-ran the full gate against the FINAL merged tree.

## Fix
One-liner: re-add `use serial_test::serial;` near the top of
`cqlite-core/src/storage/sstable/reader/data_access/full_index_stream_tests.rs`
(was at line 20). Confirm `cargo clippy --workspace --all-targets --all-features`
and the `cqlite-core --lib` test target both go green.

## Impact
Every worker's `--lite` and full gate on `cqlite-core` fails until this lands —
found while running the lite gate for an unrelated issue (#1818), whose own change
was independently confirmed clean (isolated `cargo test` of just its new test target
passed). Routing: oracle-driven, trivial, no OpenSpec.

## Acceptance criteria
- [ ] `use serial_test::serial;` restored in
      `full_index_stream_tests.rs`.
- [ ] `cargo clippy --workspace --all-targets --all-features` (matching CI's
      `RUSTFLAGS="-D warnings"`) passes.
- [ ] `cargo test -p cqlite-core --lib` passes.
- [ ] `scripts/agent-gate.sh --lite` PASSes on a clean worktree off `main`.

pmcfadin:
[...]

PR #2070, opened 2026-07-06
perf(#1578): D2 — O(1) streaming global aggregates + MAX_RESULTS demotion + streaming table_id fix

pmcfadin:
Closes #1578.

## What (D2 — Epic D #1516, read-path audit)
1. **O(1) streaming global aggregates**: GROUP-BY-free `COUNT/MIN/MAX/SUM/AVG` fold each row into a running accumulator off the scan stream (new `select_executor/stream_agg.rs`), reusing the buffered path's accumulator logic verbatim — answers byte-identical by construction. Eligibility fails safe: GROUP BY / Sort / Project / Limit / WRITETIME-TTL / targeted lookups all fall back to the buffered path (misrouting risk = slower, never wrong).
2. **`requires_materialization`**: Aggregate materializes only with GROUP BY; the streaming API stops being forced through the buffered path.
3. **MAX_RESULTS demoted to a safety valve**: explicit-LIMIT queries exempt the row-count valve only; the `max_result_bytes` byte budget still applies everywhere (pinned by `explicit_limit_does_not_exempt_byte_budget`).
4. **Cross-cutting fix (flagged in review)**: `scan_stream_windowed.rs` no longer applies `table_ids_match` on the stitched streaming path — the nb parser reports header-default table_ids, so the filter silently dropped ALL rows of CQLite-written SSTables. Mirrors the authoritative materializing `sequential_scan` stitched path (which deliberately skips it); file-level reader scoping is the real enforcement. Reviewer traced both paths and confirmed fail-safe. Likely fixes/overlaps #1897 (coordination note posted there).
5. #1116 splits in passing: `execute_streaming_background` → `streaming.rs`; `execute_aggregation` → `stream_agg.rs`.

## TDD (red-first)
- `count_star_buffers_o1_rows`: buffered-rows probe flat at 0 vs table size (was proportional: 50 vs 2000). Companion test proves the probe isn't dead (GROUP BY ≥300).
- `explicit_limit_exempts_row_count_valve`: was `QueryExecution("Result set too large…")`, now returns rows; `no_limit_still_trips_row_count_valve` stayed red-guarded.
- Aggregate parity matrix: COUNT/MIN/MAX/SUM/AVG × {int,bigint,double,text,predicate,NULLs,empty} — `execute` == `execute_streaming` == oracle. NaN comparator untouched (pre-existing shared comparator, both paths identical).

## Review
[...]

Issue #1614, opened 2026-07-02
H1 — Parser fuzz crate: vint, value decode, block emit, BTI (never panic/hang/OOM)


pmcfadin:
Part of epic #1601 (parser measurement + safety net). Source: `docs/reports/parser-performance-audit-2026-07-01.md` finding H1 (audit block 2). Wave 1 (measurement train, capstone §4).

## Problem
The parser decodes **untrusted bytes** (SSTable files from any source) and has **no fuzz target anywhere** in the repo. Its safety discipline (zero `unsafe`, length-guarded slices, `Err`-not-panic) has never been proven against adversarial input. Any panic reaching a binding aborts the host process today (see #1440).

## Fix (effort: M)
1. Create a cargo-fuzz setup: run `cargo fuzz init` at the repo root (it creates a `fuzz/` crate). The `fuzz/` crate must be **excluded from the main workspace** (add `fuzz` to `[workspace] exclude` in the root `Cargo.toml` if cargo-fuzz doesn't do it) so `scripts/agent-gate.sh` builds are unaffected.
2. Add four fuzz targets under `fuzz/fuzz_targets/`:
   - `fuzz_vint.rs` — feed arbitrary bytes to `parse_vint`, `parse_vuint`, and `parse_vint_length` (`cqlite-core/src/parser/vint.rs`). Assert the result is `Ok` or `Err` — the body must never panic.
   - `fuzz_value_decode.rs` — feed arbitrary bytes to the primitive + collection value decoders: `parse_value_with_schema_type` / `parse_value_with_comparator` (`cqlite-core/src/storage/sstable/reader/parsing/value_parsing.rs`) across a fixed list of type strings covering every scalar, plus `list<int>`, `set<text>`, `map<text,int>`, a tuple, and a nested `frozen<list<list<int>>>`.
   - `fuzz_block_emit.rs` — feed arbitrary bytes as a decompressed block to the partition-loop entry (`parse_block_emit` path in `cqlite-core/src/storage/sstable/reader/parsing/v5_compressed_legacy/`), using one fixed simple schema (e.g. the `test_basic.simple_table` CREATE TABLE).
   - `fuzz_bti.rs` — feed arbitrary bytes to BTI node decode + DFS (`cqlite-core/src/storage/sstable/bti/parser/node_decode.rs`, `traversal.rs` — the footer-based loader entry).
[...]

Issue #2456, opened 2026-07-15
flow-meta: closer merge step must assert PR head SHA == locally-certified tip (stale-merge escape on #2299/PR #2421)


pmcfadin:
## Escape (2026-07-14, #2299 endgame)
The closer certified tip `da9a7cb2` locally (full gate PASS + roborev 1724 blocker-clean) but **never pushed** — then `gh pr merge 2421` squashed the PR's stale head `ca8eb016` (pre-rebase, pre-fix). Main at `b6acc7f5` transiently carried the roborev-1723 Medium data-loss blocker (direct-stream compaction gate fails open on unreadable Statistics.db → tombstone drop; low reachability). Caught by the closer's own post-merge verification; remediation PR #2455 (byte-identical tree to the certified tip) is re-certifying. A round11b field image build from the tainted main was dispatched and had to be CANCELLED mid-build.

## Fix (doctrine + closer brief template)
- [ ] Closer merge precondition (hard, mechanical): `git push` then assert `gh pr view <N> --json headRefOid` == local certified SHA **immediately before** `gh pr merge`. Refuse to merge on mismatch.
- [ ] Add to the flow-closer agent definition + CLAUDE.md implement-loop endgame step.
- [ ] Consider a repo-side backstop: the required CI check already re-runs on push — but it cannot catch "merge of an old green head"; the SHA assertion is the real guard.

Cost: one wasted image build, ~1h delay on a field artifact, transient known-blocker on main.

PR #2467, opened 2026-07-15
fix(#1795/#1884/#1885): overflow-safe v5 cell bounds guards + Float32/varint parity + cell_value split

### LINKED ISSUE #1795: Parser panic: usize add-overflow in v5_compressed_legacy decimal/length bounds guards (uncapped parse_vuint) — found by fuzz #1614
Found by the new parser fuzz crate (#1614, target `fuzz_block_emit`) within seconds of running with `CQLITE_DATASETS_ROOT` set. This is exactly the class of bug the H1 fuzz net exists to catch: **the parser panics (aborts the process) on adversarial input** instead of returning `Err`. Per #1440, a panic reaching a binding aborts the host process.

## Panic
```
thread panicked: attempt to add with overflow
  at cqlite-core/src/storage/sstable/reader/parsing/v5_compressed_legacy/cell_value.rs:351
```

## Root cause
The decimal cell path decodes a length with `parse_vuint` and casts it to `usize` **without the 1 GB cap** that `parse_vint_length` applies, then does an **unchecked add** in the bounds guard:

```rust
let (remaining, total_len) = parse_vuint(&data[offset..])?; // u64, UNCAPPED
let total_len = total_len as usize;
...
if offset + total_len > data.len() {   // <-- offset + total_len overflows usize
    return Err(...);                    //     before the guard can reject
}
```

Contrast `parse_vint_length` (vint.rs), which rejects `value > MAX_VINT_LENGTH` (1 GB) before the `as usize` cast. The decimal path (and siblings) skip that cap and then overflow in the `offset + len` addition.

## Systemic — this is a hardening sweep, not a one-liner
The same `offset + <vint-derived-len> > data.len()` shape (unchecked add + uncapped length) recurs across the v5_compressed_legacy decoders:
- `cell_value.rs` — 28 matches of the `offset + … > …len()` shape
- `frozen.rs` — 8, `complex_column.rs` — 6, `row_data.rs` — 4

Recommended fix pattern (apply to all vint-derived bounds guards):
- bound the length before use (reuse `parse_vint_length`'s 1 GB cap, or reject `len > data.len() - offset`), and/or
- use overflow-safe comparison: `total_len > data.len().saturating_sub(offset)` (never `offset + total_len > data.len()`).

## Reproducer (533 bytes)
- sha256: `c1d613788429984ad1d9e5ce4a02e94c0c7ed62d2ad34301cf00f7f05f675299`
- First triggering type: `decimal`
- Reproduce: `cd fuzz && CQLITE_DATASETS_ROOT=$PWD/../test-data/datasets cargo +nightly fuzz run fuzz_block_emit <this-file>`

Decode the input (base64):
```
gIABXAAQFSkad9c5TnODl7eHRC86H3////+AAAkB9JcCJIJbHsghrwgHAAAAAjA2DwgBCAAAACgIBWFzY2lpCASAAE8hCAAAAZm3E2QSCIHJlN8Hsg5L8K11wpYAJCbVN/ywZNGtR4/tp1B+eDCNz1EMydc9USS/yntKMrp4g7GFd9nlnUjRosSWm++cdty2DQAAAAAAAAA2K7ZYP2Fx6nfBZQ1kCbosYis6yJ////8WY+9gAp/NFz9nY1mWXTMx/QAAAAAAAAAI+KHfBUD/////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////tRV+avmlrG4tMjjGfcHK4Kj1F84tC1EVLvrb1foJs18OrgIX/oqiYfRCg6/Ci1ZVgWh0AAAAAAAAAAit+0Y76ZR4imftzOPeU5BXgIABXAAQFSkad9c5TnODl7eHRC8AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA6//////////8ff////4AACQH0lwIkgixbHsghrwgHAAAAUlyL2oDyBb5jN8I5KeCuCaZZDy115jEt+9M0GInUbQI=
```

## Scope / DoD
- Overflow-safe bounds guards across the v5_compressed_legacy decoders (cell_value/frozen/complex_column/row_data): arbitrary length prefixes return `Err`, never panic.
- Pin the reproducer above as a regression test (bytes → `Err`, no panic).
- After the fix lands, wire `CQLITE_DATASETS_ROOT` dataset-fetch into the #1614 `fuzz.yml` `fuzz_block_emit` lane so it actively guards this path in CI (deferred in #1614 to avoid a red-by-construction lane).

Found by: #1614 H1 parser fuzz crate. Filed by delivery lead per the #1614 spec non-goal (crashes are filed separately, not silently patched).

### LINKED ISSUE #1884: v5 ladder decode widens CQL float to f64 (Value::Float) instead of Value::Float32 — lossy vs block path

### LINKED ISSUE #1885: v5 ladder decode has no varint arm → CQL varint read as Value::Blob (mirror of I4 on the v5 side)

pmcfadin:
Closes #1795
Closes #1884

## Summary
Consolidated lane (per lead's 2026-07-14 audit comment — same file, same fix
shape, one PR):

1. **#1795** — the v5_compressed_legacy decimal/text/blob/counter/duration cell
   decoders decoded a length via `parse_vuint` (u64, uncapped) then did an
   UNCHECKED `offset + total_len > data.len()` bounds check — the add itself
   could overflow `usize` on adversarial input, panicking instead of returning
   `Err` (found by the fuzz crate, #1614). Fix: new shared helper
   `read_vint_length_prefixed_bytes` caps the length against
   `MAX_CELL_VALUE_LENGTH` (64 MiB) BEFORE the `as usize` cast (mirroring the
   existing correct `parse_vint_length` pattern) and uses saturating-subtraction
   instead of an unchecked add. Applied across every vulnerable arm.
2. **#1884** — `CellKind::Float` now yields `Value::Float32(f32)`, not a lossy
   widening to `Value::Float(f32 as f64)`.
3. **#1885** — added the missing `varint` arm (was falling through to
   `Value::Blob`).
4. Flipped the `decoder_lockstep_tests.rs` float/varint cases from
   asserting divergence to asserting v5-ladder == block-path equality.
5. Campsite split: `cell_value.rs` 1251→347 lines; extracted
   `cell_value_scalar.rs` (587) and `cell_value_complex.rs` (333), all under the
   ~800 source target.

Two unrelated pre-existing `main` breakages (byte-identical to what's broken on
`main` today under the pinned 1.88.0 toolchain) fixed as drive-by unblocks since
they block `cqlite-core` clippy/tests: a missing `use serial_test::serial;`
import (same root cause as #2462, will naturally dedupe on rebase) and 3
`.err().expect()` → `.expect_err()` clippy fixups (`issue_1578_limit_exempts_max_results.rs`,
from #2070).

Oracle-driven (fuzz-found parser bug + parity fixes), no OpenSpec.

## Review-first (before any full gate)
`rust-reviewer` + `roborev review --branch` both ran on the lite-green diff.
**No blockers.** rust-reviewer verified the new bounds-check math is exactly
equivalent to the intended `offset + len > data.len()` check for all
non-overflowing inputs (no off-by-one), confirmed every vulnerable arm migrated
to the safe helper, and confirmed the split is a pure move with no visibility
widening. roborev added two Low/informational notes (no fix required):
- The overflow-safe cap now applies more broadly than before (decimal/duration/
  counter arms previously had no length ceiling at all) — intended hardening,
  consistent with the existing frozen-UDT/complex-column caps.
- The Float32/Varint representation change is a real output change for those
  column types via the v5 path — **flagging for the closer**: please confirm
  the full gate's sstabledump-golden parity suite exercises float/varint
  columns through the v5-compressed-legacy path, since goldens are the parity
  oracle for this, not just the in-crate lockstep assertions.

## Gate status
Lite:
```
commit: b9bcd31da branch: issue-1795-decimal-bounds-guard-lane dirty: no
file-size: PASS  fmt: PASS  clippy: PASS (158s)  scoped-tests: PASS (146s)
RESULT: PASS
```
Note: `CQLITE_ALLOW_FILE_GROWTH=1` needed for the full gate too — registering the
3 new sibling modules grew the already-over-threshold aggregator
`row_decoder/mod.rs` (1044→1052); splitting it is #1116 epic work, out of this
lane's scope.

Full `agent-gate.sh` (the gate of record) + final roborev pass to be run by
`flow-closer`. **Dataset root note for the closer**: this sandbox's real fetched
SSTable binaries live at `/data/datasets` (the environment default) — do not
override `CQLITE_DATASETS_ROOT` to a repo-relative path.

Deferred (secondary, not blocking): re-adding `fuzz_block_emit` to `fuzz.yml`'s
CI matrix with `overflow-checks=true` — recommend routing as its own follow-up
issue rather than folding into this PR.

---
**Update (post-rebase, 2026-07-15):** rebased onto `main` after a real conflict — #1885's varint-arm fix landed independently on `main` via PR #2466 (different machine, same bug, claim-protocol gap: my consolidation comment lived on #1795, not #1884/#1885 themselves, so a claim check on `issue-1885-*` found nothing). #1885 is already closed by #2466; dropped the redundant `Closes #1885` line above. Kept this branch's varint arm (in the split `cell_value_complex.rs`, overflow-safe per #1795's own fix — main's version was not) and reverted the auto-merged competing `CellKind::Varint` variant from `cell_kind.rs` back to this branch's Complex-ladder architecture to avoid two competing implementations. Verified equivalent behavior against main's own #2466 edge-case tests (both auto-merged in, both pass: negative/zero/boundary varints, and both empty-value wire framings). Lite gate re-confirmed green post-rebase.

### DIFF
--- a/cqlite-core/src/storage/sstable/reader/parsing/row_decoder/cell_kind.rs
+++ b/cqlite-core/src/storage/sstable/reader/parsing/row_decoder/cell_kind.rs
@@ -55,11 +55,6 @@ pub(super) enum CellKind {
     Inet,
     /// The literal CQL `blob` type (decodes to `Blob`, empty-value → `Blob([])`).
     Blob,
-    /// CQL `varint` (arbitrary-precision integer). Decodes the VInt-length-prefixed
-    /// raw two's-complement big-endian bytes to `Value::Varint` (empty-value →
-    /// `Varint([])`), mirroring the block / `ComparatorType::Varint` path exactly
-    /// (issue #1885).
-    Varint,
     /// Frozen / tuple / non-frozen-collection / marshal-UDT / unknown-scalar types:
     /// the already-lowercased declared type string, decoded by the retained string
     /// ladder. Empty-value → `Null` (matching the pre-J1 `_ => Null` empty arm).
@@ -93,8 +88,7 @@ impl CellKind {
             "time" => CellKind::Time,
             "inet" => CellKind::Inet,
             "blob" => CellKind::Blob,
-            "varint" => CellKind::Varint,
-            // frozen<…>, tuple<…>, list/set/map<…>, marshal forms, and any
+            // frozen<…>, tuple<…>, list/set/map<…>, marshal forms, varint, and any
             // unrecognized type: decoded by the retained string ladder. Store the
             // lowercased string so the ladder needs no per-cell `to_lowercase`.
             _ => CellKind::Complex(Arc::from(lowered.as_str())),
@@ -121,7 +115,6 @@ mod tests {
         assert_eq!(CellKind::from_type("duration"), CellKind::Duration);
         assert_eq!(CellKind::from_type("inet"), CellKind::Inet);
         assert_eq!(CellKind::from_type("blob"), CellKind::Blob);
-        assert_eq!(CellKind::from_type("varint"), CellKind::Varint);
     }
 
     #[test]
@@ -161,6 +154,12 @@ mod tests {
         assert_eq!(
             CellKind::from_type("list<int>"),
             CellKind::Complex(Arc::from("list<int>"))
+        );
+        // `varint` has no dedicated scalar arm — pre-J1 it fell to the default blob
+        // decode (live) / Null (empty); it must land in `Complex`, not `Blob`.
+        assert_eq!(
+            CellKind::from_type("varint"),
+            CellKind::Complex(Arc::from("varint"))
         );
     }
 }
--- a/cqlite-core/src/storage/sstable/reader/parsing/row_decoder/cell_value.rs
+++ b/cqlite-core/src/storage/sstable/reader/parsing/row_decoder/cell_value.rs
@@ -1,7 +1,5 @@
 use super::*;
 
-use super::marshal_element::MarshalCollectionElements;
-
 impl V5CompressedLegacyParser {
     /// Parse a single cell value WITHOUT column name (schema-order format)
     ///
@@ -22,6 +20,11 @@ impl V5CompressedLegacyParser {
     ///   when the cell inherits the row-level timestamp (`USE_ROW_TIMESTAMP` flag).
     /// - `expiration`: TTL / localDeletionTime pair when the cell is expiring, or
     ///   `None` when the cell has no TTL.
+    ///
+    /// The scalar value-decode arms live in [`cell_value_scalar`], the
+    /// frozen/tuple/collection/marshal-UDT ladder in [`cell_value_complex`]
+    /// (campsite split, issue #1795); this function owns the flag/conditional-field
+    /// parsing and the tombstone/empty short-circuits.
     pub(super) fn parse_cell_value_schema_order(
         &self,
         data: &[u8],
@@ -40,7 +43,7 @@ impl V5CompressedLegacyParser {
         // rare recursive frozen-inner / in-crate test callers → resolve it locally
         // from `column.data_type` (bounded, off the per-cell scan hot path).
         kind: Option<&CellKind>,
-        _reader: &crate::storage::sstable::reader::types::SSTableReader,
+        reader: &crate::storage::sstable::reader::types::SSTableReader,
     ) -> Result<(Value, Option<i64>, Option<CellExpiration>, usize)> {
         // Cell flag constants (from Cassandra 5.0 Cell.Serializer)
         const CELL_IS_DELETED: u8 = 0x01;
@@ -175,10 +178,6 @@ impl V5CompressedLegacyParser {
                 _ => None,
             };
 
-        // Step 4: Cell path for complex columns (multi-cell collections/UDTs)
-        // For now, skip this - we'll add in a future iteration when we handle complex columns.
-        // Simple columns (int, text, boolean, uuid, etc.) don't have cell paths.
-
         // === End of Phase 2 conditional field parsing ===
 
         // CRITICAL: Inverted logic for HAS_EMPTY_VALUE_MASK
@@ -255,7 +254,13 @@ impl V5CompressedLegacyParser {
                 CellKind::Blob => Value::Blob(Vec::new()),
                 // Issue #1885: an empty `varint` cell is `Varint([])`, matching the
                 // block / `ComparatorType::Varint` path (empty slice → `Varint([])`).
-                CellKind::Varint => Value::Varint(Vec::new()),
+                // In this split layout `varint` routes through `CellKind::Complex`
+                // (the retained string ladder, decoded in `cell_value_complex`), so
+                // match on the lowered declared-type string rather than a dedicated
+                // scalar tag.
+                CellKind::Complex(lowered) if lowered.as_ref() == "varint" => {
+                    Value::Varint(Vec::new())
+                }
                 _ => {
                     tracing::warn!(
                         "V5CompressedLegacy: EMPTY value for cell '{}' (type {}), treating as NULL",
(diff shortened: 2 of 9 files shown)


pmcfadin:
Nudge from the lead session: `required` has been green here for ~4h. Per the worker-merges-own-PR convention (gate PASS + roborev clean + required green → `gh pr merge --squash --delete-branch` + flow-finalize, with the #2456 headRefOid assertion), please drive this to merge or note what's blocking — don't park green PRs.

pmcfadin:
Lead session adopting this PR's ENDGAME (worker recycled without running it): full gate of record + final roborev + #2456 SHA-assert + merge-on-green + finalize will run from Patricks-MacBook-Pro. If the original worker returns, do NOT start a parallel endgame — this comment is the claim.

pmcfadin:
Lead update: the original worker's machine (ip-172-31-23-32) resumed this lane (commit 7a0cf91c + fresh heartbeat) after my 14:39Z adoption claim. **Ceding the endgame back to the live worker** — my local closer never ran the full gate and its worktree is discarded; there is no parallel endgame from Patricks-MacBook-Pro. Worker checklist, per the PR body's own note that the gate of record has NOT yet run: (1) rebase-check vs main — #1885 already shipped separately via PR #2466/a475f50d, verify no semantic overlap; (2) ONE full `agent-gate.sh` of record on the final tip; (3) final roborev on the tip commit directly (`--branch --base origin/main` false-empty-diff gotcha if main moved); (4) #2456 assert PR headRefOid == certified SHA; (5) merge closes #1795/#1884 (#1885 is already closed). Lead will re-adopt if the heartbeat goes stale (>4h) again.

pmcfadin:
Lead RE-ADOPTING this endgame from Patricks-MacBook-Pro: the worker machine recycled to #2470 (PR #2501) at 15:18Z without running the gate of record — per the supervisor model (one issue per worker process) it will not return to this lane. This supersedes my earlier ceding comment; the >4h staleness clause is moot since the worker has moved on, not gone quiet. Endgame from here: rebase/overlap check (esp. #1885 via PR #2466) → ONE full gate of record → final roborev on tip → #2456 SHA-assert → merge (closes #1795/#1884) → finalize. Second occurrence of #2499 on this same PR.

pmcfadin:
⚠️ FREEZE — endgame is COMPLETE-IN-FLIGHT from Patricks-MacBook-Pro: the full gate of record just PASSED at `6dab2638` (current PR head). Final roborev + #2456 SHA-assert + merge are executing NOW. To the worker heartbeating on #1795 (ip-172-31-23-32): do NOT push to this branch — any push invalidates the certified SHA and the merge will abort on the #2456 assert. If you have residual changes, hold them for a follow-up PR.

pmcfadin:
Lead: WITHDRAWING adoption definitively — the worker machine owns this endgame. It force-pushed `742528ab` (rebase onto main@9f6cfb53) over my certified `6dab2638` while my closer was in final stages; per #2456 the merge was correctly NOT made. Data points for the worker: my full gate of record PASSed at `6dab2638` with all 26 components green, and `git diff 6dab2638..742528ab` is exactly the two new upstream commits (#2074, #2470) — the #1795/#1884 fix files are byte-identical. That gate does NOT certify `742528ab`: run your own full gate of record on your tip, final roborev (tip-commit direct if --branch --base returns empty-diff), #2456 SHA-assert, then merge (Closes #1795/#1884; #1885 shipped via #2466). No further pushes or endgame actions will come from Patricks-MacBook-Pro on this lane. Bonus: main now has #2470, so the #2451 index_probes flake should be gone from your gate.

Issue #2470, opened 2026-07-15
Flaky: windowed_stream_read_pattern_is_sequential asserts seek_calls() < 125 against a shared global counter contaminated by concurrent tests


pmcfadin:
## What
`cqlite-core/src/storage/sstable/reader/data_access/full_index_stream_tests.rs::windowed_stream_read_pattern_is_sequential`
asserts `rwc::seek_calls() < 125`. Observed failing under parallel test execution
with `seek_calls() == 144` — passes reliably in isolation and on a clean re-run.

## Root cause hypothesis
`seek_calls()` reads a process-global counter. Under `cargo test`'s default
parallel test execution, other tests in the same binary that also touch the
seek-call counter run concurrently and inflate the count observed by this
test's assertion — a shared-mutable-global-state flake, not a real regression
in the windowed-stream read pattern.

## Found during
Implementation of #1819 (Docker-probe timeout fix) — unrelated to that change;
this test lives in the same binary and was observed flaking during gate runs.
Not fixed there to keep that PR scoped.

## Suggested fix
Mark the test `#[serial(work_counters)]` (the existing convention this file
already uses for other tests reading the same global counters — see the 4
existing `#[serial(work_counters)]` sites in this file) so it doesn't run
concurrently with other counter-touching tests, OR make the counter
thread-local/test-scoped if that's a smaller change.

## Acceptance criteria
- [ ] `windowed_stream_read_pattern_is_sequential` passes reliably under
      `cargo test`'s default parallel execution (run it in a loop alongside
      the full `cqlite-core --lib` suite a few times to confirm no flake).
- [ ] No regression to the other `#[serial(work_counters)]`-tagged tests' runtime
      (serializing one more test should be a negligible addition).

Routing: oracle-driven (test-infra flake fix, established pattern to mirror),
no OpenSpec needed.

pmcfadin:
🔧 Claimed (flow-lead, machine `ip-172-31-23-32`) — branch `issue-2470-flaky-windowed-stream-serial` pushed. Oracle-driven test-infra fix, no OpenSpec. This flake has now hit multiple full-gate runs this session; fixing it now rather than deferring further.

pmcfadin:
✅ Merged via PR #2501 (squash) — full gate PASS (all 22 components), roborev clean (job 35, no issues found), CI green. Follow-up #2500 (sibling test migration) filed separately, left open.

Issue #1691, opened 2026-07-02
AG2 — Retire execute_parallel_table_scan: unbounded channel, blocking recv, 4 duplicate full scans


pmcfadin:
Part of epic #1684 (platform landmines). Source: `docs/reports/platform-observability-audit-2026-07-01.md` finding AG2 (audit block 6). **Verification-first: the duplicate-scan counter test ships before the fix.**

## Problem
`QueryExecutor::execute_parallel_table_scan` (`cqlite-core/src/query/executor.rs:477-525`) contains, in one function:
1. **The crate's only unbounded data-path channel**: `channel::unbounded()` (crossbeam) at `:489` — a slow consumer buffers the whole table in memory.
2. **A blocking `rx.recv()` on a tokio worker** at `:516` (`while let Ok((row_key, row_data)) = rx.recv()`) — parks an async worker for the duration.
3. **N duplicate full scans**: `thread_count` (default 4) workers each issue the IDENTICAL full `storage.scan` (`:492-509` — the comment admits it; nothing partitions the key space or dedups).
It is **latently dead** on today's routing (SELECTs go via `SelectExecutor`; the cached-plan branch at `:393` only holds point lookups) but it is in the default build and directly callable — a loaded gun for any future routing consolidation.

## Fix (effort: S/M)
**Preferred (delete):**
1. Verify reachability: trace `execute_parallel_table_scan` callers (`rg -n "execute_parallel_table_scan" cqlite-core/src`) and confirm the only path is the `:393` plan branch; write the verification down in the PR.
2. Route that branch's `TableScan` step through the existing bounded streaming scan path (the same machinery `SelectExecutor` uses — bounded mpsc, `spawn_blocking` discipline), and DELETE `execute_parallel_table_scan` plus its now-unused imports (`crossbeam` channel use in this file).
**Minimum bar if deletion is blocked** (justify in PR): bounded channel (`buffer_size`), disjoint key-range assignment per worker, and `recv().await` (async, not blocking). But deletion is the recommendation — do not keep a parallel scan implementation beside the canonical one.

### Do NOT
- Do not keep the unbounded channel "because it's dead" — dead-but-callable is the finding.
- Do not build new parallel-scan machinery — reuse the streaming path.

## Acceptance criteria
- [ ] Work-counter test (H5 #1618 pattern): a `TableScan` through this branch invokes `storage.scan` ONCE (or once per disjoint range), not 4× whole-table (4× today — the red test).
[...]

PR #2471, opened 2026-07-15
fix(#1819): bound the Docker-availability probe with a wall-clock timeout

### LINKED ISSUE #1819: Docker-gated e2e tests hang the gate when the Docker daemon is unresponsive (unbounded `docker info` probe)
## Summary

The `write-support` test binary `issue_911_bti_partition_deletion_stats` HANGS
indefinitely (observed 16+ min at 0% CPU in a #1750 gate run, and reproduced
here with a wall-clock guard) whenever the local Docker **daemon is present but
unresponsive**. The hang is in the Docker-availability gate `cassandra_5_image()`,
which calls `Command::new("docker").arg("info").output()` (and then
`docker images`) with **no timeout**. When `docker info` blocks forever, the test
— and therefore `scripts/agent-gate.sh` — wedges permanently.

This is **not** a code bug in the product; it is a test-infra robustness gap. It is
in the same hazard family as #1691 / #1692 (blocking wait inside a `#[tokio::test]`
current-thread `block_on`), but here the blocking primitive is a subprocess wait,
not an mpmc channel.

## Confirmed unrelated to #1750

During #1750 (retire `is_simple_id_lookup` heuristic) I ran an A/B causation check:

| engine.rs | Result |
|-----------|--------|
| WITH #1750 WIP | `da_stats_marks_partition_level_deletion` PASS, then `partition_deletion_reads_back_under_cassandra5_sstabledump` **HUNG** |
| plain `origin/main` (WIP reverted) | identical: first subtest PASS, second **HUNG** |

The hang is byte-identical with and without #1750. The test is
`#![cfg(feature = "write-support")]` and never touches `QueryEngine` /
`select_executor`, so the #1750 SELECT-routing change cannot reach it.

## Root cause (reproduced directly)

`docker info` on the affected machine blocks > 30s (daemon socket accepts but does
not respond). Guarded repro:

```
(docker info >/tmp/di.log 2>&1 & pid=$!; sleep 30; \
  if kill -0 $pid 2>/dev/null; then echo HUNG; kill -9 $pid; else echo RETURNED; fi)
# => HUNG
```

The reported stack sample (one thread parked on `std::sync::mpmc::list::Channel::recv`
via `Thread::park`/`semaphore_wait`, main thread in `Runtime::block_on` on the
current-thread scheduler) is consistent with `std::process::Command::output()`
draining the child's pipes and waiting on the never-returning `docker info` child,
all inside the `#[tokio::test]` current-thread runtime.

## Blast radius (shared unguarded pattern)

The same timeout-less `docker info` probe is copy-pasted across at least:

- `cqlite-core/tests/issue_911_bti_partition_deletion_stats.rs`
- `cqlite-core/tests/issue_911_bti_sstabledump_parity.rs`
- `cqlite-cli/tests/integration_sstable_tests.rs`

(16 `Command::new("docker")` call sites total.) Any of these can wedge the gate the
same way.

## Proposed fix

Bound every Docker-availability probe with a wall-clock timeout so an unresponsive
daemon degrades to a **clean skip** (the tests already skip cleanly when Docker is
absent). Options: spawn + timed `try_wait` loop, kill on timeout and treat as
"Docker unavailable"; or a shared `docker_available_or_skip()` helper with a bounded
probe reused by all e2e Docker-gated tests. Do NOT silently disable the tests.

## Acceptance

- With the Docker daemon hung/unresponsive, all Docker-gated e2e tests SKIP cleanly
  within a few seconds instead of hanging.
- `scripts/agent-gate.sh` cannot be wedged by a hung Docker daemon.
- No change to behavior when Docker is absent (clean skip) or healthy (test runs).

Related: #1691, #1692. Discovered while validating #1750.

pmcfadin:
Closes #1819

## Problem
The `write-support` e2e tests' Docker-availability gate (`cassandra_5_image()`)
called `docker info`/`docker images` via unbounded `Command::output()`. When the
local Docker daemon is present but unresponsive, this blocks forever, wedging
the test binary — and `scripts/agent-gate.sh` itself — permanently (16+ min
observed). Same blocking-wait-with-no-deadline hazard family as #1691/#1692,
here a subprocess wait rather than an mpmc channel recv.

## Fix
New shared seam `cqlite-core/tests/common/docker_probe.rs`:
`bounded_probe(bin, args, timeout)` spawns the child, polls via
`wait_timeout::ChildExt`, kills+reaps on timeout. Both `cassandra_5_image()`
call sites (`issue_911_bti_partition_deletion_stats.rs`,
`issue_911_bti_sstabledump_parity.rs`) now route through it — an unresponsive
daemon degrades to a clean SKIP (non-strict) or a loud FAIL under
`CQLITE_REQUIRE_FIXTURES=1` (strict), never a hang.

Oracle-driven, no OpenSpec.

## Review-first (two rounds on the lite-green diff, before any full gate)
Round 1 (`rust-reviewer` + `roborev --branch`): 0 blockers on the seam design,
kill/reap/race-safety, and fail-closed conventions — but roborev found a real
**Low-severity correctness regression**: `bounded_probe` only drained stdout
AFTER the child exited and never drained stderr, unlike the `Command::output()`
it replaced (which drains both concurrently). A healthy `docker images`
producing output over the OS pipe buffer (~64KB) would block on its own
write(), get killed at the deadline, and be misreported `TimedOut` — a FALSE
gate failure against a healthy daemon in strict mode. Fixed in `961d7d2bd`:
per-pipe reader threads drain stdout+stderr concurrently with the wait loop,
with a bounded (`recv_timeout`) collection grace so an orphaned grandchild
holding an inherited pipe (as the test harness's `sleep`-based hang simulation
does) can't reintroduce an unbounded wait. New test
`bounded_probe_captures_output_larger_than_pipe_buffer` (200KB on both pipes,
~3x the typical buffer) proves this discriminates old vs new behavior, not
just "eventually returns."

Round 2 (re-review of the fix): clean, no blockers. One more Low/optional
note from roborev: `child.wait()` after `kill()` is technically unbounded in
theory, though `SIGKILL` makes the practical risk negligible (reviewer's own
assessment) — batching as a nit, not fixing (no concrete failure scenario).

## Nits (batched, not fixed here, no re-verify per the nit rule)
- Orphaned-grandchild kill signals only the direct child, not the process
  group — a non-issue for the real `docker` target, only a theoretical test-
  harness edge case.
- Timeout-test ceiling (10s vs 400ms budget) has slack; adequate, not tight.
- `child.wait()` post-kill is technically unbounded (see above).
- Doc comment says "a small drain grace" (singular); actual worst case is up
  to 2x `DRAIN_GRACE` (stdout then stderr sequentially). Wording nit.

## Drive-by unblock (separate commit, not part of #1819's scope)
`cargo test -p cqlite-core --lib` was compile-red on `origin/main` itself
(unrelated merge race between #2437/#2438 — tracked/fixed properly at #2462).
Added the same one-line `use serial_test::serial;` import here too so this
branch's own gate could run; will naturally dedupe on rebase once #2462 merges.

## Also filed (non-blocking, found while implementing, not fixed here to keep
this PR scoped)
#2470 — a pre-existing flaky test in the same file
(`windowed_stream_read_pattern_is_sequential`) contaminated by a shared global
counter under parallel test execution. Unrelated to this change.

## Gate status
Lite:
```
commit: 961d7d2bd branch: issue-1819-docker-probe-timeout dirty: no
file-size: PASS  fmt: PASS  clippy: PASS  scoped-tests: PASS
RESULT: PASS
```
Full `agent-gate.sh` (the gate of record) + final roborev pass to be run by
`flow-closer`. **Dataset root + PATH note for the closer**: this sandbox needs
`export CQLITE_DATASETS_ROOT=/data/datasets` (subagents do not inherit the
parent shell's env) and the rustup 1.88.0 toolchain bin on PATH for fmt/clippy.

### DIFF
--- a/cqlite-core/Cargo.toml
+++ b/cqlite-core/Cargo.toml
@@ -158,6 +158,11 @@ zstd = { version = "0.13", features = ["zdict_builder"] }
 # Inline-CRC32 framing for the hand-built Cassandra chunk image in that same
 # test (integration tests don't inherit the library's normal dependencies).
 crc32fast = { workspace = true }
+# Wall-clock-bounded subprocess wait for the Docker-availability probe used by
+# the write-support Cassandra e2e tests: an unresponsive `docker` daemon must
+# degrade to a clean SKIP instead of wedging the test binary (issue #1819,
+# same hazard family as #1691/#1692). Mirrors cqlite-cli's dev-dep.
+wait-timeout = "0.2"
 
 # In-process sampling CPU profiler for the criterion benches (signal-based, so
 # it works in containers and on CI runners where `perf` is unavailable).
--- a/cqlite-core/tests/common/docker_probe.rs
+++ b/cqlite-core/tests/common/docker_probe.rs
@@ -30,6 +30,8 @@
 
 use std::io::Read;
 use std::process::{Command, Stdio};
+use std::sync::mpsc::{self, Receiver};
+use std::thread::JoinHandle;
 use std::time::Duration;
 
 use wait_timeout::ChildExt;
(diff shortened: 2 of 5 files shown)


pmcfadin:
Nudge from the lead session: `required` has been green here for ~4h. Per the worker-merges-own-PR convention (gate PASS + roborev clean + required green → `gh pr merge --squash --delete-branch` + flow-finalize, with the #2456 headRefOid assertion), please drive this to merge or note what's blocking — don't park green PRs.

pmcfadin:
Lead session adopting this PR's ENDGAME (worker recycled without running it): full gate of record + final roborev + #2456 SHA-assert + merge-on-green + finalize will run from Patricks-MacBook-Pro, queued after #2467's gate (serialized). If the original worker returns, do NOT start a parallel endgame — this comment is the claim.

Issue #1870, opened 2026-07-04
ORDER BY / MIN / MAX float comparator diverges from Cassandra: NaN not sorted last, -0.0==+0.0, non-total order under NaN


pmcfadin:
Surfaced during #1587 (E5) verification; **pre-existing** on main, not caused by E5 (E5's decorate-sort preserves this comparator's semantics unchanged — value_ops.rs untouched).

## Problem
`compare_values_ordering` (`cqlite-core/src/query/select_executor/value_ops.rs`) compares floats via `f64::partial_cmp().unwrap_or(Equal)`:
- (a) NaN compares Equal to everything → **not sorted last** (Cassandra/Java `Double.compare` sorts NaN last).
- (b) `-0.0 == +0.0` → **not `-0.0 < +0.0`** (Java distinguishes signed zero).
- (c) worst: a NaN key makes the comparator a **non-total order**, so `sort_by` leaves even the finite keys in an unspecified order.

Affects ORDER BY and likely MIN/MAX aggregates on float/double columns.

## Oracle / fix direction
Use an explicit Cassandra-matching comparator (NaN last, `-0.0 < +0.0`), mirroring the pre-roborev-findings guidance (float ordering vs Java). Validate with an ORDER BY float/NaN/±0.0 parity test against a Cassandra golden. This is a project-wide ordering behavior change → its own parity validation; route as oracle-driven (Cassandra is the oracle) + a pinned parity test.

## Acceptance
- [ ] ORDER BY over floats incl. NaN and ±0.0 matches Cassandra (NaN last, -0.0<+0.0), total order.
- [ ] MIN/MAX over float columns consistent with the same comparator.
- [ ] Pinned parity test; agent-gate PASS.

pmcfadin:
🧭 **MANAGER** 
GO
ORDER: 14

P0 ORDER BY/MIN/MAX float comparator diverges from Cassandra — ranked ahead of the read-path perf epic. 0.14 policy: **P0 correctness/gate-green first**; the read-path epic runs in parallel on other workers (nothing held).

PR #2488, opened 2026-07-15
fix(#2074): GROUP BY on float/double groups by Cassandra comparator

### LINKED ISSUE #2074: GROUP BY on float/double column uses IEEE equality (merges ±0.0, splits NaN) — diverges from Cassandra comparator grouping
## Problem
Follow-up from #1870 (float total-ordering). #1870 made ORDER BY / MIN / MAX / clustering-key comparison Cassandra-correct (NaN last, all-NaN-equal, `-0.0 < +0.0`) via `cqlite-core/src/float_cmp.rs`. But **GROUP BY grouping** on a float/double column still uses IEEE `PartialEq`:
- `cqlite-core/src/query/select_executor/aggregation.rs` `hash_group_key`/`find_or_init_group` (~lines 64–86, 162) group by IEEE `==` + hash.

This means CQLite GROUP BY on a float column currently:
- MERGES `-0.0` and `+0.0` into one group (IEEE `==` true), and
- puts EACH `NaN` in its own group (IEEE `NaN != NaN`).

Cassandra groups by the **type comparator** (the total order), which does the OPPOSITE: all `NaN` values form ONE group, and `-0.0`/`+0.0` are DISTINCT groups. So after #1870, ORDER BY/MIN/MAX are Cassandra-correct while GROUP BY grouping diverges in the opposite direction, and the two are internally inconsistent.

## Oracle
Apache Cassandra 5.0: GROUP BY grouping uses the clustering/column type comparator (e.g. `DoubleType.compare`), NOT IEEE equality. Confirm exact behavior against a live Cassandra oracle (group all NaN together; keep -0.0/+0.0 distinct) before implementing.

## Acceptance criteria
1. GROUP BY on a `float`/`double` grouping column groups all NaN bit-patterns into a single group and keeps `-0.0`/`+0.0` as distinct groups (matching Cassandra).
2. Group-key equality/hash routed through a canonicalizing key or the `float_cmp` total-order comparator (so eq/hash stay consistent — canonicalize NaN to one bit pattern; treat -0.0 and +0.0 as distinct).
3. A parity test pins GROUP-BY-on-float grouping semantics against a Cassandra/sstabledump-derived oracle (physical golden won't capture this — needs a query-semantics assertion).
4. Correct the `Value` `PartialOrd` doc comment in `cqlite-core/src/types.rs` (~line 1043) so it no longer implies IEEE GROUP BY grouping matches Cassandra — state it as a known divergence tracked by this issue (until fixed).

## Routing
Oracle-driven (Cassandra comparator is the source of truth) — GitHub issue + pinned parity test, no OpenSpec. Surfaced by roborev on PR for #1870 (job 3634). Relates to the #1583/#1644 Value-v2 train only loosely (this is grouping semantics, not layout).

pmcfadin:
## Summary
Closes #2074.

CQLite's GROUP BY on a `float`/`double` column grouped rows by IEEE `==`/hash, which diverges from Cassandra's comparator in BOTH directions:
- IEEE `-0.0 == +0.0` merged them into one group (Cassandra keeps them distinct).
- IEEE `NaN != NaN` split every NaN row into its own group (Cassandra collapses all NaN bit-patterns into one group).

This routes the GROUP BY group-key hashing and equality confirmation through the existing Cassandra/Java total-order comparator (`cqlite-core/src/float_cmp.rs`, added in #1870 for ORDER BY/MIN/MAX), so grouping is now internally consistent with ordering:
- `-0.0`/`+0.0` → DISTINCT groups.
- ALL NaN bit-patterns → ONE group.
- This holds at every group-key nesting depth (`List`/`Set`/`Map`/`Tuple`/`Frozen`), keeping the hash/equality contract consistent throughout (a review-round finding — the first pass only fixed top-level scalar keys, leaving a nested-float hash/eq inconsistency; fixed by making `group_key_eq`/`group_value_eq` recurse identically to the hash function). UDT-nested floats are the sole acknowledged carve-out (fall through to derived `PartialEq`, contract-sound since UDT hashes by discriminant only) — noted as out of scope in the doc comment.
- `cqlite-core/src/types.rs`'s `PartialOrd` doc comment corrected: GROUP BY grouping no longer "stays IEEE" for floats — it shares the same Cassandra total order via the aggregation group-key path.

## Root cause
`cqlite-core/src/query/select_executor/aggregation.rs` (`hash_one_value`/`find_or_init_group`, now split into `aggregation/group_key_cmp.rs` per the campsite file-size rule): the hash normalized signed zeros together and left NaN un-canonicalized, and the group match confirmation used derived `Vec<Value>` `PartialEq` (IEEE) — the exact opposite of what Cassandra's `DoubleType`/`FloatType` comparator does.

## Wiring evidence
`cqlite-core/tests/issue_2074_group_by_float_comparator_parity.rs` drives GROUP BY through the full public path: writes a real uncompressed SSTable via `WriteEngine`, ingests it, and runs `SELECT fv, COUNT(*) FROM ... GROUP BY fv` via `Database::execute`, asserting `+0.0`/`-0.0` form two distinct single-row groups and two different NaN bit patterns collapse into one 2-row group. Unit tests in `aggregation.rs` and the new `group_key_cmp.rs` pin the same semantics (including nested-key hash/eq consistency) at the helper level.

## Query-semantics oracle
Not added to `test-data/query-semantics-oracle.json` — that lane is strictly fixture-driven (requires a committed real-Cassandra SSTable + JSONL golden), which isn't available in this environment. The dedicated end-to-end test above pins the query-semantics assertion instead. Candidate follow-up: add a `test_group_by_float` Cassandra fixture + oracle case when a Cassandra container is available.

## Review
- rust-reviewer + roborev (round 1) both independently flagged the same real issue: a nested-float hash/eq inconsistency in the first-pass fix (top-level-only comparator routing). Fixed by recursing the equality predicate identically to the hash function.
- roborev (round 2, post-fix): no issues found. Two Low items noted (UDT-nested floats stay IEEE; missing Set/Map-key-nested test cases) — assessed as narrow, contract-sound, out of stated scope, no defect.

## Test plan
- [x] `scripts/agent-gate.sh --lite` PASS (fmt/clippy/scoped-tests) post-rebase on `f9945ab`.
- [x] New unit tests (`aggregation.rs`, `group_key_cmp.rs`) pin top-level AND nested hash/eq consistency.
- [x] New end-to-end integration test drives the fix through `Database::execute`.
- [ ] Full `scripts/agent-gate.sh` (the gate of record) — run once by `flow-closer` before merge.

### DIFF
--- a/cqlite-core/src/query/select_executor/aggregation.rs
+++ b/cqlite-core/src/query/select_executor/aggregation.rs
@@ -5,6 +5,8 @@
 //! `finalize_group`) that the executor's `execute_aggregation` drives row by
 //! row.
 
+mod group_key_cmp;
+
 use super::super::select_ast::AggregateType;
 use super::super::select_optimizer::{AggregateComputation, AggregationPlan};
 use super::numeric_acc::NumericAcc;
@@ -14,9 +16,9 @@ use crate::{
     query::result::QueryRow,
     types::{RowKey, Value},
 };
+use group_key_cmp::{group_key_eq, hash_group_key};
 use rustc_hash::FxHashMap;
 use std::collections::HashMap;
-use std::hash::{Hash, Hasher};
 
 // Test-only counter for GROUP-key exact (`==`) comparisons performed while
 // locating a row's group (issue #1587, E5). The hash index makes this
@@ -37,12 +39,13 @@ pub(super) struct AggregationState {
     pub(super) groups: Vec<(Vec<Value>, Vec<AggregateValue>)>,
     /// Issue #1587 (E5): hash index over the group key → candidate `groups`
     /// indices sharing that hash. `Value` does not implement `Hash`/`Eq`
-    /// (floats), so we hash a normalization that is *consistent* with `Vec<Value>`
-    /// equality (equal keys hash equal; `-0.0`/`+0.0` normalized) and confirm the
-    /// match with an exact `==` against each candidate. Distinct-but-colliding
-    /// keys (and every NaN key, which is never `==` itself) simply fall through
-    /// the exact check, so grouping semantics are byte-identical to the prior
-    /// linear scan.
+    /// (floats), so we hash a normalization *consistent* with the group-key
+    /// equality used to confirm a match ([`group_key_eq`]), and confirm the match
+    /// with that predicate against each candidate. Issue #2074: `float`/`double`
+    /// group-key equality follows Cassandra's total-order comparator (`float_cmp`),
+    /// so the hash canonicalizes NaN to one bit pattern (all NaN keys hash-collide
+    /// into ONE group) and keeps signed-zero bits DISTINCT (`-0.0`/`+0.0` hash into
+    /// two groups) — exactly consistent with [`group_key_eq`].
     ///
     /// Issue #1590 (E8): `FxHashMap` rather than the default `HashMap`. The key
     /// is already a well-mixed `u64` group-key hash, so SipHashing it again is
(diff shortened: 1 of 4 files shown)


pmcfadin:
Lead claiming this PR's ENDGAME (queued behind #2467's gate on Patricks-MacBook-Pro): worker ip-172-31-24-182 recycled forward without running the gate of record (PR body checkbox unticked). Review is done (roborev clean round 2). Will run: rebase check → full gate of record → final roborev on tip → #2456 SHA-assert → merge (closes #2074) → finalize. #2499 instance #3.

Issue #2499, opened 2026-07-15
flow-meta: supervisor-recycled workers orphan their open PRs' endgames (gate of record never runs)


pmcfadin:
## Gap

When `worker-supervisor.sh` (#2090) recycles a worker after PR-open (context bound hit), the recycled worker moves to a NEW issue and **nobody owns the previous PR's endgame** — the full gate of record, final roborev, #2456 SHA-assert, and merge never run. The PR sits green-on-`required` looking merge-ready while the PR body itself records only lite-stage PASS.

## Evidence (2026-07-15)
- PR #2467 (fix #1795/#1884/#1885 overflow-safe v5 cell bounds): sat ~9h with only lite PASS recorded; lead nudge went unanswered; lead adopted the endgame, then the original worker resumed and the lanes collided (adoption ceded back).
- PR #2471 (fix #1819 Docker probe timeout): same shape; endgame had to be adopted and driven by the lead.

## Why it's dangerous
- A green `required` check masquerades as merge-ready; merging it would ship code the gate of record never certified (the #2299 stale-merge class, one step earlier).
- Lead adoption is a race against worker resurrection (observed on #2467): two endgames on one lane violates single-writer-per-worktree.

## Proposed fixes (pick at grooming)
1. **Supervisor completes the lane before recycling**: the hard context bound should trigger recycle only at issue DONE (post-finalize), not post-PR-open — or the supervisor re-launches a worker pinned to the SAME issue until the PR merges.
2. **Endgame claim marker**: closer/worker posts a machine-stamped 'endgame owner' marker (PR comment or heartbeat stage) so lead adoption and worker resumption can't race silently.
3. **flow-board reaper extension**: treat 'open PR + stale heartbeat + PR body lacks full-gate SUMMARY' as an orphaned-ENDGAME state distinct from abandoned-claim, with a defined adoption protocol.

Refs: #2084 (flow-closer), #2090 (supervisor), #2456 (SHA-assert), #2089 (heartbeat/reaper).

pmcfadin:
Field data from today's three instances — **the adoption protocol as designed cannot work against a live supervisor fleet**:

- **#2488 (#2074)**: lead claimed the endgame by PR comment at 15:23Z; the recycled worker resurrected anyway, pushed a new head, ran its own endgame and self-merged at 16:12Z. Lead's duplicate closer detected the done state and no-op'd (clean outcome, wasted spawn).
[...]

PR #2523, opened 2026-07-15
fix(#2167): restore compressed CHUNK_READ_CALLS increment + observability nits

### LINKED ISSUE #2167: Follow-ups from #1773 (PR #2154): compressed CHUNK_READ_CALLS + nits
Follow-ups batched from the roborev review of PR #2154 (issue #1773, compressed offset-read CRC fix). None block; all are polish/observability.

## (a) Low — restore compressed CHUNK_READ_CALLS increment
`cqlite-core/src/storage/sstable/reader/data_access/mod.rs` (~516-531): after the CRC fix, the compressed branch no longer increments `CHUNK_READ_CALLS`, so compressed-table read observability underreports chunk reads. Restore the counter increment on the compressed offset-read path so the metric stays accurate for compressed tables.

## (b) NIT — matching warn! on the unreachable compression_reader=None else-branch
`cqlite-core/src/storage/sstable/reader/data_access/compressed_offset.rs` (~59-96): the (unreachable) `compression_reader = None` else-branch returns still-compressed bytes with no `warn!`, unlike the sibling `stitch_all_chunks:374` which does warn. Add a matching `warn!` for parity so an unexpected reach is visible.

## (c) NIT — comment why CHUNK0_SIZE=16 is non-tombstone
In the #1773 test (`cqlite-core/tests/issue_1773_compressed_offset_crc.rs`): add a one-line comment explaining why `CHUNK0_SIZE = 16` bytes is a non-tombstone value, for the next reader.

pmcfadin:
Closes #2167.

Follow-ups batched from the roborev review of PR #2154 (issue #1773, compressed offset-read CRC fix). All polish/observability; none block behavior.

## (a) Restore compressed `CHUNK_READ_CALLS` increment
The #1773 CRC fix moved the compressed point-read backing read into `read_compressed_offset_window` and dropped the `CHUNK_READ_CALLS` bump, so compressed-table read observability underreported. The increment is restored in `get_cached_data`'s compressed branch — the **point-read site only** (`data_access/mod.rs`), symmetric with the uncompressed branch and placed **after** the shared-cache check so a warm/cached repeat read still does zero backing reads.

It is deliberately **not** placed inside `read_compressed_offset_window`: that helper is also called from the scan paths (`summary_scan`, `full_index_scan`, `full_index_stream`), whose uncompressed counterparts do not bump this point-read counter. (Review-first caught this over-reach — see below.)

Pinned by a new in-crate `#[serial]` wiring test `big_point_read_compressed_increments_chunk_read_calls` on `test_comp.lz4_table` — proven TDD-meaningful (fails pre-fix with `got 0`).

## (b) Matching `warn!` on the unreachable `compression_reader = None` branch
`read_compressed_offset_window`'s else-branch returned still-compressed bytes with no log, unlike sibling `stitch_all_chunks`. Added a mirrored `tracing::warn!` (identical wording modulo the function-name prefix) so an unexpected reach is visible.

## (c) Comment: why `CHUNK0_SIZE = 16` is a non-tombstone window
Explained in the #1773 test: the offset-read path emits `ScanRow::RawRow`, which `filter_tombstone` keeps unconditionally (it suppresses only a `ScanRow::Marker` row-tombstone) — so a clean read always surfaces `Ok(Some(_))` regardless of the 16 bytes' content.

## Verification / review
- Lite gate PASS (file-size / fmt / clippy `-D warnings` / scoped-tests).
- Review-first: rust-reviewer APPROVE; roborev initially flagged a Medium (increment in the shared helper over-reaching to scan callers) — fixed by moving it to the point-read site; **roborev re-review: "No issues found."**
- Full `agent-gate.sh` of record runs pre-merge (flow-closer).

Scope: exactly the 3 items above + the verification test. No production behavior change beyond the counter increment.

### DIFF
--- a/cqlite-core/src/storage/sstable/reader/data_access/chunk_cache_wiring_tests.rs
+++ b/cqlite-core/src/storage/sstable/reader/data_access/chunk_cache_wiring_tests.rs
@@ -74,6 +74,35 @@ fn uncompressed_data_db() -> Option<PathBuf> {
     None
 }
 
+/// Locate `test_comp.lz4_table`'s Data.db — a COMPRESSED BIG (`nb`) table, so
+/// `get_cached_data` routes through `read_compressed_offset_window` (CRC-validated,
+/// issue #1773). The wiring evidence is the same `CHUNK_READ_CALLS` counter, which
+/// the compressed branch must increment (issue #2167).
+fn compressed_data_db() -> Option<PathBuf> {
+    let base = datasets_root()?.join("sstables/test_comp");
+    let rd = std::fs::read_dir(&base).ok()?;
+    for entry in rd.flatten() {
+        let name = entry.file_name();
+        let name = name.to_str()?;
+        if name.starts_with("lz4_table-") {
+            let dir = entry.path();
+            if let Ok(files) = std::fs::read_dir(&dir) {
+                for f in files.flatten() {
+                    let p = f.path();
+                    if p.file_name()
+                        .and_then(|n| n.to_str())
+                        .map(|n| n.ends_with("-Data.db"))
+                        .unwrap_or(false)
+                    {
+                        return Some(p);
+                    }
+                }
+            }
+        }
+    }
+    None
+}
+
 async fn open(path: &std::path::Path) -> SSTableReader {
     let config = Config::default();
     let platform = Arc::new(Platform::new(&config).await.expect("platform init"));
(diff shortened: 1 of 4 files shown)


pmcfadin:
Lead adopting this PR's ENDGAME from Patricks-MacBook-Pro (worker gone >6h, no pushes — #2499 policy threshold met): rebase check → full gate of record → final roborev on tip → #2456 SHA-assert → merge (closes #2167) → finalize. If the original worker returns, do NOT start a parallel endgame.

Issue #2172, opened 2026-07-07
Follow-ups from #1703 (PR #2169): faithful subscriber-on bench + wider span/info tests


pmcfadin:
Follow-ups triaged out of #1703 (demote write-side spans to DEBUG + quiet SELECT INFO), PR #2169. None block; batched here per severity-triage policy.

**(a) P2 — subscriber-on overhead bench under-counts spawn_blocking overhead.**
`cqlite-core/benches/observability_overhead.rs` installs the INFO subscriber via thread-local `tracing::subscriber::with_default`, so the `*_subscriber_on` variants only capture spans emitted on the bench's own thread — work that runs on `spawn_blocking` / blocking-pool threads is not observed, so the recorded subscriber-on number under-counts the true default-posture overhead. Fix: install a **global** subscriber with an atomic on/off toggle (measure with it on vs off), or explicitly document the number as a main-thread lower bound. (roborev job 1516, Medium.)

**(b) Low — regression-test breadth.**
`issue_1703_write_span_levels.rs` / `issue_1703_select_info_level.rs` cover full-table `SELECT *` and synchronous `write()` only. Extend to targeted / materialization SELECT branches and the async flush/execute/compaction entry points (`write.cql_execute` / `write_async` / `execute_flushing`) so the demotions are pinned across all span-bearing paths. (roborev job 1516, Low.)

**(c) NIT — optional default-posture SELECT marker.**
After the demotion the SELECT path emits zero INFO lines. Consider keeping ONE top-level "SELECT executed, N rows" INFO marker at the `db.execute()` boundary so default-posture operators retain a minimal per-query signal. Product/observability call — evaluate, don't assume.

pmcfadin:
🔒 Claimed by flow-lead worker (branch `issue-2172-observability-bench-nits` pushed to origin). Items (a)/(b) are clear technical fixes; (c) is flagged in the issue itself as a product/observability call — will surface rather than assume.

pmcfadin:
Shipped in PR #2533 (squash-merge `f36ce88`) via the flow-closer endgame.

**Delivered (acceptance = items a+b):**
- **(a)** `observability_overhead.rs` now installs a single **process-global** INFO subscriber gated by an atomic `SUBSCRIBER_ON` toggle (replacing the thread-local `with_default`), so `spawn_blocking`/blocking-pool spans are counted faithfully — the subscriber-on overhead number is no longer a main-thread-only under-count.
[...]

PR #2537, opened 2026-07-15
docs(process): log lesson 19 — owner decision on default SELECT log posture

pmcfadin:
Docs-only. Records the owner's #2172-item-(c) decision (keep SELECT quiet at INFO) as a standing precedent.


pmcfadin:
Provenance note: lesson 19's "owner standing decision" was not yet verifiable when this PR opened (#2172's thread only shows the worker deferring the call). The owner has NOW confirmed the decision in the lead session (2026-07-15, on the lead's recommendation): default SELECT log posture stays fully quiet at INFO; DEBUG/tracing spans are the per-query surface. Merging with the record corrected — future 'owner decision' lessons must link the comment/thread where the decision was actually made.

Issue #1570, opened 2026-07-02
B4 — key→partition-offset cache (Cassandra key-cache analogue)


pmcfadin:
Child **B4** of Epic B (#1514) — 2026-07 read-path audit (block 1). Source of truth: `docs/reports/read-path-performance-audit-2026-07-01.md` §Epic B. Anchors as of `main` @ `5c080d2a`. Capstone wave: **Wave 3** (after B1/C1).

## Problem
Even with B1's chunk cache, every hot point read still pays the full index/trie descent (bloom → Summary/Index or BTI trie walk) to re-resolve the same partition key to the same offset. Cassandra solves this with its key cache; CQLite has no analogue.

## Fix (step by step)
1. Add a small, bounded LRU cache: key `(sstable identity, partition key bytes or key digest)` → value `(data offset, partition size)` — exactly what the index descent produces.
2. Wire it in front of the index/trie resolution in the point-read path (both BIG and BTI): cache hit → seek directly to the offset, skipping bloom/Summary/Index/trie entirely; miss → resolve as today, insert.
3. Bound it by entry count or bytes (entries are tiny); make the capacity a constructor parameter with a small default (e.g. 64k entries) — coordinate with the B2 config surface only if a knob is wanted; do NOT add a decorative knob.
4. Invalidation: SSTables are immutable — evict by sstable identity on reader remove/reload, same rule as B1.
5. Concurrency: same sharding rule as B1 (no `RwLock<LruCache>` — recency mutates on get).

## TDD tests (write FIRST, must fail on current main)
- Repeated-key lookup: second lookup of the same key has `TRIE_WALKS` == 0 and index-probe counter == 0 (A5 counters). Fails today.
- **Parity:** for N random present keys, cached `(offset,size)` == fresh resolution `(offset,size)` (resolve twice, compare). For absent keys: cache must not fabricate hits.
- Eviction correctness: after eviction, lookup still returns correct rows (falls back to full resolution).

## Guardrails
- The cache stores *locations*, never row data (rows change shape under projection; locations are immutable facts).
- No-heuristics: key→digest hashing uses the existing authoritative `key_digest` machinery.

## Definition of done
- [ ] TDD tests red-then-green; `scripts/agent-gate.sh` PASS (paste SUMMARY verbatim)
- [ ] `RUSTFLAGS="-D warnings"` clean; no `unwrap()`/`expect()` in library code
- [ ] 33-table parity green

Issue #1572, opened 2026-07-02
C1 — BIG get(): hash key to index digest + seek the chunk — stop whole-file decompress per lookup


pmcfadin:
Child **C1** of Epic C (#1515) — 2026-07 read-path audit (block 1). Source of truth: `docs/reports/read-path-performance-audit-2026-07-01.md` §Epic C. Anchors as of `main` @ `5c080d2a` — re-locate with `rg` if drifted. Capstone wave: **Wave 3**, FIRST in the point-read story (C1 → C2 → B1).

## Problem
On BIG (`nb`) SSTables, the `SSTableManager::get()` KV surface reads and decompresses the **entire Data.db on every lookup**. Mechanism:
1. The BIG index HashMap is keyed by **16-byte Murmur3 digests** of partition keys, but `find_entry()` is called with the **raw key bytes** — so it ALWAYS misses. This is already documented in a code comment at `data_access/mod.rs:152-155`.
2. The miss falls back to `scan_for_key` → `stitch_and_parse_all_chunks` (`sequential.rs:330-395`) = read + decompress the whole file per lookup.

Scope note (verified in the audit): the default query-engine single-candidate path seeks correctly — this bug hits the `get()` surface, fallbacks, the multi-candidate path (see C4), and the `tombstones` build.

## Fix (step by step)
1. Before the index lookup, hash the query key to its Murmur3 digest using the existing authoritative machinery (`rg -n "key_digest" cqlite-core/src` — a faithful pinned Murmur3 port exists; do NOT write a new hash).
2. Look up the digest in the index HashMap (`index_reader.rs:203` is the O(1) map). On hit you get the partition's data offset.
3. Seek to the partition's chunk via `CompressionInfo::chunk_for_offset` — the BTI path already does exactly this at `bti.rs:600-640`; mirror that pattern for BIG.
4. Read + CRC-check + decompress ONLY the chunk(s) covering that partition, then decode only that partition.
5. Keep `scan_for_key` as a last-resort fallback ONLY for genuinely index-less situations, and make its invocation observable (the existing `SCAN_FOR_KEY_CALLS` counter).

## TDD tests (write FIRST, must fail on current main)
- On a **multi-chunk** BIG fixture with a present key: `SCAN_FOR_KEY_CALLS` delta == 0 for `get()` (fails today — every get scans).
- Bytes-read counter (A5): a `get()` reads O(one chunk), not O(file size).
[...]

PR #2554, opened 2026-07-16
feat(#2059): global bounded key→partition-offset cache

### LINKED ISSUE #2059: Global bounded key→partition-offset cache (Cassandra key-cache model) — bound aggregate across all open readers
Follow-up from #1570 (B4). #1570 delivered a PER-READER sharded key→partition-offset cache with occupancy-proportional allocation and a budget-conscious per-reader cap (aggregate documented + confirmed within <128MB for a generous open-reader count). 

This tracks the larger architectural improvement roborev flagged: replace the per-reader caches with a SINGLE GLOBAL bounded cache keyed on `(sstable_id, key)` (invalidated on reader drop) — Cassandra's actual key-cache model. A global cache bounds aggregate resident memory regardless of how many SSTable readers are concurrently open, rather than relying on `N_readers × per_reader_cap` staying within budget.

Routing: design (perf/memory architecture). Acceptance: one shared bounded cache across the reader set, keyed by sstable identity + key, correctly invalidated on reader remove/reload; aggregate memory bounded by a single global cap independent of reader count; point-read hit still returns the exact offset a fresh lookup resolves (correctness preserved); wiring-evidence for hit/skip. Part of v0.14+ read-path perf wave (Epic B). Reference: docs/reports/read-path-performance-audit-2026-07-01.md §B4; #1570 delivered the per-reader form.

pmcfadin:
Closes #2059

## Summary
Replaces the per-reader `KeyOffsetCache` (#1570) with a single process-global, byte-bounded, sharded key→partition-offset cache keyed by `GenerationIdentity` (device+inode+size+generation+mtime_ns, fail-closed on identity mismatch) — bounding aggregate resident memory by one fixed budget regardless of open-reader count, matching Cassandra's actual key-cache model.

OpenSpec change: `openspec/changes/bounded-key-cache/` (spec + design approved at Seam 1).

## Design highlights
- Sharded `Mutex<Shard>`, each shard a `HashMap<GenerationIdentity, LruCache<Box<[u8]>, Entry>>` + a monotonic per-shard seq clock for cross-identity global-LRU eviction ordering.
- O(1) whole-identity invalidation on genuine disk removal (compaction/refresh-detected-gone generations); capacity-driven eviction under the warm registry's memory pressure does NOT invalidate — correctness never depended on invalidation completeness, since fail-closed identity matching means a stale entry can never serve a mismatched identity.
- Wiring-evidence: `cqlite-flight/tests/issue_2059_do_get_key_cache_e2e.rs` drives real `FlightService::do_get`, asserts cold vs warm against the real `INDEX_INTERVAL_PARSES_TOTAL` counter.
- `cqlite.cache.key.*` observability counters (hits/misses/invalidations/evictions), docs regenerated.

## Review history (this branch)
- Round 1 (rust-reviewer + roborev): flat cache → O(N) lock-holding scan + per-lookup heap alloc → restructured to nested-map + seq clock.
- Round 2 (rust-reviewer + roborev): warm-registry capacity eviction was invalidating the shared cache exactly under memory pressure (self-defeating) → decoupled capacity-eviction from genuine-removal invalidation; added `capacity_eviction_preserves_global_key_cache` test (verified non-vacuous).
- Round 3 (rust-reviewer clean; roborev): one blocker — missing test for the in-place-replacement byte-accounting path (`saturating_sub` before re-add on same-key reinsert). Added `reinsert_same_key_does_not_grow_resident_bytes`, verified non-vacuous (temporarily broke the subtraction → test failed with bytes doubling 113→226; restored → passed).
- 4 Low/nit findings from round 3 (non-unix identity aliasing — unsupported platform, no action; O(n) eviction-loop scan perf; DDL-rebuild invalidation effectiveness loss; test `read_dir` `.ok()?` robustness) are batched into one follow-up issue at merge time — not blocking.

## Known deviation
`global_key_offset.rs` was already over the campsite ~800-line source threshold before this branch (987 lines) and grew further with the round-3 test addition (→1017 lines). Re-ran `--lite` with `CQLITE_ALLOW_FILE_GROWTH=1` — splitting it by responsibility is out of scope for this change; tracked under epic #1116.

## Gate
`--lite` PASS on every round (latest: commit 320b33b, file-size/fmt/clippy/scoped-tests all PASS). Full `agent-gate.sh` + C intent audit + final roborev to run once via `flow-closer` before merge.


pmcfadin:
Full gate FAILed on first run: \`core-tests\` + \`work-counters-guard\` (all 20 other components PASS; summary at /tmp/gate-2059.txt, RESULT: FAIL). Triage + fix:

- **\`work-counters-guard\`**: test methodology obsolete, not a bug — the 32-reader fan-out test measured per-reader-cache isolation that the new *global* cache correctly eliminates. Fixed in \`23bb30dfd\` by opening each fan-out reader with \`block_cache.enabled=false\` so the hash-hoist measurement stays isolated from the (now-shared) key cache.
- **\`core-tests\`**: genuine correctness bug, but **pre-existing on \`main\`** (reproduced identically on both branches in isolation — not introduced by this PR, just surfaced by this PR's gate run). BTI point-read decode trusted "row-decode closure fired" as "partition fully decoded" for a chunk-straddling partition, causing a present key to fall back to whole-file \`scan_for_key\` (violates #1572). Fixed in \`0ebb43381\`; traceability issue #2561 filed (root cause + BIG-format follow-up to verify). Repro evidence: 0/80 clean post-fix vs deterministic fail pre-fix on the strengthened (every-key) test.

Both fixes are lite-green (\`gate-2059-lite5.txt\`: PASS) and now in review (rust-reviewer + roborev) before a fresh full gate.

PR #2574, opened 2026-07-16
feat(#1918): CQLITE_READ_PATH forcing knob + point-vs-full differential lane

### LINKED ISSUE #1918: CQLITE_READ_PATH forcing knob + point-vs-full differential-equality lane
Child of Epic **#1915** (point-read fast path residue). Anchors as of `main @ b20c5b23`. **Routing: design-driven** (new user-visible env surface + a new CI lane) — goes through `flow-activate` / OpenSpec at pickup.

## Problem

There is no way to force a read down a specific access path, and no automated lane proving the point path and the full-scan path return **equal results** for the same query. `classify_partition_lookup` (`select_executor/lookup.rs:92`) decides the plan at three match sites in `execute.rs` (**:364** streaming, **:643** metadata, **:714** materializing — :714 is the canonical, fully-wired one). #958's guard (`tests/issue_958_partition_lookup_work_bound.rs`) bounds the *work* of the point path, but nothing asserts point-vs-full *result equality* — precisely the class of divergence that let #1741 (single-gen reconciliation skipped on one path) hide. This complements #1742's external semantic oracle: #1742 checks CQLite vs Cassandra; this lane checks CQLite vs CQLite across paths.

`git grep CQLITE_READ_PATH` returns nothing — the knob does not exist.

## Proposed shape (refine in OpenSpec at pickup)

1. **Forcing knob** `CQLITE_READ_PATH=auto|point|full` (naming/values to be settled in the spec; "compact" from the original #942 sketch is stale — the compact-to-memory framing predates #951):
   - `auto` (default): current behavior, `classify_partition_lookup` decides.
   - `point`: if classification yields `Targeted`/`MultiTargeted`, take the fast path; if the query is unclassifiable, **fail loudly** (error, not silent fallback) — the knob exists to make silent fallbacks impossible during testing.
   - `full`: force `Fallback` (full scan + reconciliation) regardless of classification.
   - Implement as a single wrapper around `classify_partition_lookup`'s return (or one gate applied at all three match sites) — do NOT fork per-site logic. Read the env once (`OnceLock`), record the forced choice in `AccessPath` so `--explain`/tests can see it.
2. **Differential-equality CI lane**: for every point-read-eligible query over the corpus fixture tables (single-key, IN, clustering-pushdown, WRITETIME/TTL variants; include multi-generation + tombstone + TTL fixtures — the 17 multi-gen tables like `ct_multi_sstable_merge` and the resurrection pairs), run each query under `point` and `full` and assert **identical rows, values, and order**. Wire as an agent-gate component or an existing-tier integration test — spec decides which (mind the #1825 gate-cost budget; this should be a normal `cargo test` target, not a new heavyweight gate stage).
3. `--explain` already exists (`cqlite-cli/src/commands/query.rs:29–69`) but does not surface the `AccessPath` label — surfacing it (plus the forced-mode marker) is in scope if cheap, else split out.

- [ ] `CQLITE_READ_PATH=point` on an unclassifiable query returns a loud error naming the fallback reason; never a silent full scan.
- [ ] `CQLITE_READ_PATH=full` on a targeted query records `AccessPath::FallbackFullScan` (forced) and returns results identical to `auto`.
- [ ] Differential lane runs point-vs-full over the eligible query matrix on the corpus fixtures and fails on any row/value/order divergence; it demonstrably catches a seeded divergence (verify the lane by temporarily breaking one path locally — regression-test-verification doctrine).
- [ ] Knob is documented (CLI/user docs) and honest: it is a test/debug surface, not a perf recommendation.
- [ ] `scripts/agent-gate.sh` PASS (paste SUMMARY verbatim); `-D warnings` clean; no `unwrap()`/`expect()` in library code.

## Guardrails

- No-heuristics mandate untouched — the knob forces routing, never changes decoding semantics.
- Default behavior (`auto` / unset) must be byte-identical to today; the knob must add zero overhead when unset beyond the one-time env read.

pmcfadin:
Closes #1918

## Summary
Adds a debug/test `CQLITE_READ_PATH` forcing knob (`ReadPathMode::{Auto,Point,Full}`, config-over-env-over-auto) that forces a `SELECT` down a specific access path, wired through a single `apply_forcing` gate at all five `classify_partition_lookup` call sites (`execute.rs` ×3, `streaming.rs`, `stream_agg.rs`). Adds a new differential-equality test lane proving point and full-scan paths return byte-identical results for the same query (pinned `now`, never wall-clock) — the CQLite-vs-CQLite complement to #1742's CQLite-vs-Cassandra query-semantics oracle.

OpenSpec change: `openspec/changes/read-path-forcing-knob/` (spec + design approved at Seam 1, 2026-07-14).

## Design highlights
- Fail-closed by construction: forced `point` on a query the point path can't serve returns `Error::ForcedReadPathUnavailable` rather than silently falling back to full-scan; forced `full` disables the point path entirely at every site.
- Wiring-evidence: `read_path_forcing_e2e.rs` drives the real public `Database::execute` surface; `point_vs_full_differential.rs` proves point/full equality over a multi-generation/tombstone/TTL corpus.
- `--explain` forced-path marker deferred (spec-sanctioned — `database.explain` is plan-only and produces no `AccessPath`; the "recorded identically across surfaces" requirement is already met via `metadata.access_path`).

## Review history (this branch)
- Round 1 (rust-reviewer + roborev): rust-reviewer clean; roborev clean (2 Low nits: untested env-resolution seam, untested aggregate-forcing path, tombstones-build lane exclusion).
- **Fix round: rust-reviewer's IMPORTANT finding** — forced `full` on a schema-less sole-pk point read silently returned 0 rows instead of matching `auto` (the general full-scan predicate path can't reconstruct pk-column identity without a schema), violating the approved spec's SHALL that `full`'s row-set be identical to `auto`'s. Fixed via a new `full_forbids_schemaless_seek` fail-closed guard (mirrors the existing `point_forbids_fallback` pattern) returning `Error::ForcedReadPathUnavailable` instead of a silently-wrong empty result. Verified non-vacuous: reverting the guard reproduced the exact bug and the new regression test caught it.
- Round 2 (rust-reviewer + roborev on the fix): both clean. 7 total Low/nit findings across all rounds (stale doc summary line; WRITETIME/metadata-branch e2e coverage gap; env-var test-hygiene in the differential lane; untested env-resolution seam; untested aggregate-forcing path; tombstones-build lane exclusion; `--explain` deferral needs a tracking issue number) — batched into one follow-up issue at merge time, none blocking.

## Gate
`--lite` PASS on every round (latest: commit cc7491748). Full `agent-gate.sh` + C intent audit + final roborev to run once via `flow-closer` before merge.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -189,6 +189,10 @@ by responsibility (source: epic #1116; tests: #1135). Genuinely out of scope →
   component `query-semantics-oracle`, test `query_semantics_oracle_parity.rs`) records the
   post-reconciliation result set of a canonical `SELECT` at a PINNED `now` (never wall-clock). Add
   the correct oracle for the property under test; correctness of `SELECT` output needs the semantic one.
+  The CQLite-vs-CQLite complement is the *point-vs-full differential lane* (issue #1918,
+  `cqlite-core/tests/point_vs_full_differential.rs`): it runs the same point-eligible query under
+  forced `CQLITE_READ_PATH=point` and `=full` and asserts identical rows/values/order at a PINNED
+  `now` — catching a divergence between the two read paths that a physical dump cannot see.
 
 ### Fuzzing (issue #1614)
 `fuzz/` is a cargo-fuzz/libFuzzer crate in its own workspace, excluded from the main one — the gate
--- a/bindings/node/src/error.rs
+++ b/bindings/node/src/error.rs
@@ -504,6 +504,15 @@ mod tests {
                 Error::UnsupportedQuery(_) => {
                     assert_eq!(err.category(), ErrorCategory::Query);
                 }
+                // Issue #1918: read-path forcing knob errors. `InvalidReadPath`
+                // is Configuration-category; the fail-closed `point` error is
+                // Query-category. Both flow through `category()` like the rest.
+                Error::InvalidReadPath { .. } => {
+                    assert_eq!(err.category(), ErrorCategory::Configuration);
+                }
+                Error::ForcedReadPathUnavailable { .. } => {
+                    assert_eq!(err.category(), ErrorCategory::Query);
+                }
                 Error::CqlParse(_) => {
                     // CqlParse is Query category in cqlite-core
                     assert_eq!(err.category(), ErrorCategory::Query);
--- a/bindings/python/src/error.rs
+++ b/bindings/python/src/error.rs
@@ -403,6 +403,10 @@ mod tests {
                 Error::QueryExecution(_) => { /* Maps to QueryError */ }
                 Error::ResultTooLarge { .. } => { /* Maps to QueryError (issue #1582) */ }
                 Error::UnsupportedQuery(_) => { /* Maps to QueryError */ }
+                // Issue #1918: read-path forcing knob errors fall through to the
+                // base CqliteError like every other query-shaped variant below.
+                Error::InvalidReadPath { .. } => { /* Maps to CqliteError */ }
+                Error::ForcedReadPathUnavailable { .. } => { /* Maps to CqliteError */ }
                 Error::CqlParse(_) => { /* Maps to ParseError */ }
                 Error::Configuration(_) => { /* Maps to PyValueError */ }
                 Error::InvalidInput(_) => { /* Maps to PyValueError */ }
--- a/cqlite-core/src/config.rs
+++ b/cqlite-core/src/config.rs
@@ -355,11 +355,47 @@ fn default_max_result_rows() -> u64 {
     1_000_000
 }
 
+/// Forced SELECT access path (issue #1918).
+///
+/// A **test/debug** control that removes doubt about which access path serves a
+/// `SELECT`. It never changes value decoding, tombstone/timestamp reconciliation,
+/// or WRITETIME/TTL semantics — it governs *routing only* — and is chosen
+/// exclusively from explicit operator config/env, never inferred from data bytes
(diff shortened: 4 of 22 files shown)


pmcfadin:
**flow-closer endgame verdict: green — merging.**

- **Gate of record**: full `agent-gate.sh` on `cc7491748` — RESULT: **PASS** (all 25 components; `/tmp/gate-1918-full-v6.txt`). Postdates the final src change; PR is MERGEABLE/CLEAN so no rebase (gate stays valid).
- **C (intent audit)**: **PASS** — all 7 requirements in `openspec/changes/read-path-forcing-knob/specs/read-path-forcing/spec.md` satisfied with public-surface test evidence, including the "forced `full` rows SHALL equal `auto`" SHALL (proven non-vacuous by `read_path_forcing_schemaless_1918.rs`, which pins the earlier IMPORTANT fail-closed fix). The deferred CLI `--explain` marker is legitimately out of scope (spec mandates no such requirement).
- **roborev (final pass)**: **clean-on-arrival** — job 128 reviewed base..`cc7491748` (my exact tip), returning only 2 Low-severity nits (both already in the batched set), zero blockers.
- **Nits**: 7 batched into follow-up **#2586** (test-hygiene, doc polish, coverage-robustness) — none re-triggers a verify round.

The single earlier blocker (forced `full` silently returning wrong empty results for a schema-less sole-pk query) was fixed pre-close via a fail-closed guard and verified non-vacuous.

Issue #2367, opened 2026-07-13
Round-8 field validation tracker — post-#2361 scan-fix build (LIMIT/full-scan re-verdict + R7-blocked scenarios)


pmcfadin:
## Round-8 field validation — post-#2361 scan-fix build

Successor to the round-6/7 tracker (#2286). Purpose: re-verdict the LIMIT/full-scan scenarios that failed round 7, on a build carrying the #2361 fix (PR #2364), plus the scenarios round 7 left blocked. **Any failure found gets posted HERE.**

### Assets (PINNED — do not substitute)
| Asset | Pin |
|---|---|
| Flight image | `ghcr.io/pmcfadin/cqlite-flight:round8@sha256:cf208c0da8669c95406b5af1cc81021276a32d7708a633468e1e18350d232ae0` (INDEX digest, multi-arch, from main `8542c2d0` — includes the #2361 fix, merge `cf8fbaf2`) |
| Connector | `in.mcfad:cqlite-trino:0.13.5` (unchanged — live on repo1; the fix is entirely flight/server-side) |
| Trino | 481 (unchanged) |

### Verdict checklist (correctness pass/fail only; latency recorded, never blocking)
- [ ] **#2264/#2157 — LIMIT**: `SELECT * FROM keyvalue LIMIT 5` completes, both snapshot AND live modes. Expected: seconds-to-low-minutes (see caveat), NOT a hang
- [ ] **count(*)** completes, both modes
- [ ] **Cancellation reclaims**: kill a running scan → `in_flight_ratio` drains AND pod CPU/memory return to baseline WITHOUT a restart
- [ ] **#2228 — full-range SELECT *** completes (was blocked by the hang)
- [ ] **#2227 — multi-node snapshot full-ring verdict** (infra already proven in R7; needs a completing scan)
- [ ] **#2241 — killed-pod failover** (deferred from R7; needs a completing full-ring baseline)
- [ ] **Regression spot-checks**: #2193 tiny decode, #2229 TIME/mixed, #2207 pushdown (access-path labels `multi_partition_lookup`/`streaming_partition_lookup`, never `full_scan`) — all passed R7, confirm no regression
- [ ] **New observability**: `cqlite.rpc.phase.active` levels visible during scans (this is new in the build — a slow scan shows a standing `stream` level; please capture it, it's our precision instrument for any residual issue)

### Known caveat (NOT a round-8 failure)
[...]

Issue #2086, opened 2026-07-06
doctrine: review BEFORE the first full gate — reorder the implement loop, make review-first default


pmcfadin:
Parent epic: #2083. From audit Speed-R1 (F2/F3).

## Problem
Current loop: lite → (conditional) review-first → FULL gate → C → roborev → fix → RE-GATE. Roborev runs AFTER the full gate, so any finding needing a src change re-buys a 12–25 min full gate. Measured: 47% of all gate runs fail (151/325); issues WITH roborev findings average 2.71 rework rounds vs 0.17 without; 608 findings across 174 issues. Review is discovering fixable problems after we already paid for the gate.

## Routing
Direct implement (skip OpenSpec) — doctrine edit to the loop order.

## Change
Edit flow-implement SKILL.md + CLAUDE.md tiered-gate-loop sections: the loop becomes
`implement → lite (each round) → rust-reviewer + roborev on the LITE-GREEN diff → fix (lite re-cert) → FULL gate ONCE → C → merge`.
Review-first becomes DEFAULT, not conditional; keep the skip only for genuinely mechanical/localized diffs (define: no pub-item change, single call site, no new surface — the existing conditional's inverse). The full gate then runs on already-reviewed code; a post-gate roborev pass remains as confirmation but should converge to clean-on-arrival.

## Expected payoff
Converting even 40% of multi-run issues to single-run ≈ 35–50 fewer full gates per ~174 issues (~2.5–3.5 gate-hours) + fewer rebase-exposure windows.

## Acceptance criteria
- [ ] flow-implement SKILL.md loop order updated; review-first default with the narrow mechanical skip defined
- [ ] CLAUDE.md agent-team conventions + tiered-gate sections match
- [ ] Website delivery-pipeline page updated in the same change
- [ ] Telemetry retro after ~30 issues: gate-runs/issue and rework-rounds compared against the 174-issue baseline (median 1.5 / p90 3; 2.71 rework w/ findings)

Issue #2451, opened 2026-07-14
flaky gate test: full_materializing_scan_does_not_reprobe_index_per_partition — global index_probes counter needs test-local scoping (extend #2428's pattern)


pmcfadin:
Shipped today in #2430 (PR #2446), `full_materializing_scan_does_not_reprobe_index_per_partition` (compaction_cancel_tests.rs:807) asserts a delta on the process-global `index_probes` counter. It flaked #2356's gate of record (`got 1 probes for 100 partitions` — one stray probe from a concurrently-running non-serial test between reset() and read; passes clean in isolation; pre-#2430 red signature was N=100, so the fix itself is fine).

This is the exact class #2428 fixed for `stream_walk_partitions_parsed` with the thread-local `StreamWalkScope`. `index_probes` (and any other delta-asserted work counter) needs the same treatment:
- [ ] Extend the test-local scope pattern to `index_probes` (thread-local scope in work_counters/read_work_counters, matching StreamWalkScope's !Send/Drop/nesting semantics), or tag ALL tests reaching `record_index_probe` into a serial group (structural scoping preferred per #2428's precedent).
- [ ] Migrate `full_materializing_scan_does_not_reprobe_index_per_partition` (and audit the other index_probes delta tests: issue_1570/1576/1599 files) onto it.
- [ ] Sweep: any OTHER work counter with delta assertions and no scoping is the same latent flake — enumerate and fix or file.

Cost evidence: one wasted 25-min gate-of-record run on #2356's endgame (2026-07-14).

pmcfadin:
**P2→P1: three gate-of-record flakes in ONE day (2026-07-14/15), each costing a full ~25-min re-run:**
1. `full_materializing_scan_does_not_reprobe_index_per_partition` (index_probes) — #2356's gate
2. same class hit during #2430's endgame lineage
3. `storage::sstable::reader::block_io::tests::read_compressed_chunk_at_records_one_read_per_chunk` (rwc::read_calls, left:2 right:1) — #2463's gate

[...]

PR #2579, opened 2026-07-16
fix(#2452): connector snapshot retirement hardening — ref-counted rollback retire + background grace-sweep with quiet-table tick

### LINKED ISSUE #2452: Follow-ups: snapshot-lifecycle retirement hardening + metrics note (from PR #2425)
Batched follow-ups from the endgame of PR #2425 (snapshot-lifecycle-closure, #2356/#2306). None block; grouped per the nit-batching rule. The counter-scoping test-isolation flake is tracked separately in #2451 — referenced here, not duplicated.

### Design/hardening (needs a JDK to design + test — the closer environment has none)
1. **Ref-counted retirement of a fresh window on fail-closed rollback** (roborev job 1721, Medium; roborev job 1722 Low #2). With per-`(window,host)` shared create futures, a freshly-created reuse window can be concurrently *reused* by another query. On a partial fan-out failure, `resolveSnapshot`'s rollback currently only drops the window from `windows` (leaving earlier hosts' snapshots to the ~6h Sidecar TTL — a bounded operational leak, no correctness impact, spec-satisfied). Actively retiring the shared window in the rollback (attempted in the endgame, then reverted) removes the leak but introduces a **narrow stranding race**: if a concurrent query reused the same window (different host set after a topology change) and committed it to its tickets, deleting the shared hardlinks strands that reader with a NotFound. Proper fix = reference-count the shared window and retire only when no in-flight query holds it (or retire only this query's own per-host creates). Requires JDK compile + a concurrency test; do NOT blind-patch.

2. **Offload grace-sweep deletes off the split-planning path** (roborev job 1722 Low #1). `sweepRetireDue` runs inline from `resolveSnapshot` on `CqliteFlightSplitManager.getSplits`, so a hot table with a short reuse window pays a synchronous multi-host `clearSnapshot` DELETE fan-out on most planning cycles — new planning latency. Offload to a background best-effort executor (or bound per-resolve retire work); the TTL backstop already covers a missed retire.

### Doc/metrics note
3. **Snapshot-only metrics exposure** — `reader_opens` and `rebind_hits_total` are exposed only via the programmatic `WarmMetricsSnapshot` stats surface (authoritative for field round-over-round verification), NOT mirrored to the OTel counter export, unlike the warm cache hit/miss/evict/refresh counters. This is intentional (the spec states it) and recorded here so a field probe and future audits agree on the surface.

Refs: PR #2425, #2356, #2306, #2451 (counter-scoping flake, separate).

pmcfadin:
Closes #2452

## What (round-13 connector payload — answers #2367's 714-snapshot field flag)
- **Ref-counted rollback retirement** (`SnapshotManager.java`): fail-closed rollback actively retires partial per-host hardlinks (kills the ~6h-TTL leak) via an `AtomicInteger holders` on the shared `Window`, serialized against the `windows.compute` bin-lock; retire fires only at removed-from-map AND holders==0 — the previously-reverted stranding race is pinned by a deterministic latch-driven test (query B reuses+commits while query A rolls back → zero clearSnapshot calls).
- **Background grace-sweep + quiet-table periodic tick** (`SnapshotRetireScheduler.java` new): `sweepRetireDue`'s multi-host DELETE fan-out moves off the `getSplits` planning path to a single-daemon-thread coalescing scheduler with a fixed-delay tick (floored 1s, default = retire-grace cadence) — **quiet tables now prune their backlog with zero intervening queries** (the #2367 714-snapshot mechanism). Explicit `start()` post-construction (no this-escape); `runSafely` catches `Throwable` so an Error cannot silently kill the tick; connector constructs the manager AFTER ArrowMemoryPreflight/RootAllocator so catalog-load failures cannot leak the thread; close is idempotent and post-close submits are safe (wedge-flag test).
- **Metrics-surface note** (`trino-connector/README.md`): reader_opens/rebind_hits_total are programmatic-only (WarmMetricsSnapshot), intentional per spec — now documented where field operators look.

## Review-first (#2086) — done pre-gate
- Java concurrency review (opus): CLEAN, no blockers — ref-count pairing, retire-once, coalescing-loss, lifecycle all verified; race test confirmed a true regression pin. 4 nits folded in.
- roborev job 1753: 1 blocker (constructor-order thread leak) + 2 low — ALL fixed; job 1754 (confirmation) pending.
- Tests: `./gradlew test --rerun-tasks` → 36 classes, 344 tests, 0 failures (6 new hardening tests, all latch/manual-clock deterministic).

## Gate
Connector-only diff (Java + README; zero Rust). Full `agent-gate.sh` (gate of record) runs pre-merge in flow-closer per doctrine. Ships in connector 0.14.4 for round 13.

### DIFF
--- a/trino-connector/README.md
+++ b/trino-connector/README.md
@@ -306,6 +306,15 @@ otherwise-separate problems:
   snapshot dir and forced a full re-open + re-parse. The flight server exposes scale-free work
   probes (`reader_opens`, `index_parses_total`, `rebind_hits_total`) so a field round can verify
   the closure: within a window these stay flat.
+
+  > **Metrics exposure (field-probe note, issue #2452 / #2356).** `reader_opens` and
+  > `rebind_hits_total` are exposed **only** through the flight server's programmatic
+  > `WarmMetricsSnapshot` stats surface (the authoritative surface for round-over-round field
+  > verification) — they are **not** mirrored to the OpenTelemetry counter export, unlike the warm
+  > cache `hit`/`miss`/`evict`/`refresh` counters. This is intentional per the
+  > `flight-warm-snapshot-closure` spec (§D): the snapshot-lifecycle work does not expand the OTel
+  > counter set. A field probe or audit that expects these two on the OTel scrape will not find
+  > them — read them from `WarmMetricsSnapshot`.
 - **Fewer memtable flushes (#2306).** Flush-on-snapshot is **by design** (issue #2305; the
   Sidecar HTTP API has no `skipFlush`, and this connector does not relitigate that). The only
   lever on flush volume is **fewer snapshot creations**: because each snapshot create triggers
--- a/trino-connector/src/main/java/in/mcfad/cqlite/flight/CqliteFlightConnector.java
+++ b/trino-connector/src/main/java/in/mcfad/cqlite/flight/CqliteFlightConnector.java
@@ -45,9 +45,14 @@ public class CqliteFlightConnector implements Connector {
         // The superseded-snapshot grace-sweep runs on a background best-effort scheduler (issue
         // #2452 item 2), NOT synchronously on the split-planning path — a hot table would otherwise
         // pay a multi-host DELETE fan-out in planning latency (roborev job 1722). The periodic tick
-        // (cadence = the retire-grace period) also prunes QUIET tables that receive no further query,
-        // so a short reuse window plus a long TTL never accumulates snapshot dirs (the #2367 field
-        // accumulation). Retired at connector shutdown via SnapshotManager.close().
+        // (cadence = the retire-grace period) is the ACTUAL #2367 fix: it prunes QUIET tables that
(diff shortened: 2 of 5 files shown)


pmcfadin:
## Gate of record (flow-closer) — PASS

Full `agent-gate.sh` run twice at the same certified SHA `cba6ada3` (rebased onto origin/main, connector-Java+README-only diff, zero Rust).

**Run 1 — 27/28, single FAIL:** `work-counters-guard: FAIL (76s)` — the #2451 process-global counter flake class. A connector-Java-only diff cannot change Rust counter semantics; targeted `--only work-counters-guard` re-run PASSed. Flake tallied.

**Run 2 (gate of record) — 28/28 PASS** at `cba6ada3`:
```
work-counters-guard: PASS (32s)
RESULT: PASS
```
All 28 components green (fmt, clippy, core/integration/write/cli tests, python+node bindings, minimal-build, smoke, etc.). datasets: 144 Data.db files. This clean PASS is the gate of record.

PR #2774, opened 2026-07-19
feat(#2679): plan-time split pruning for fully-bound partition keys

### LINKED ISSUE #2679: trino-connector: plan-time split pruning for fully-bound partition keys — point reads currently fan out ~48 DoGets across all pods
## Context (from #2661 v0.15.0 milestone soak + code triage)

The soak capped at **~39 qps @ 32 threads** (p50 798ms = queue wait on the ceiling; server-side ~2ms). Read-only triage of the connector found the largest single lever:

**`CqliteFlightSplitManager.getSplits` never uses the pushed-down constraint to prune ranges** (`CqliteFlightSplitManager.java:87-88`) — the predicate is only carried in the ticket and applied server-side (`CqliteFlightPageSourceProvider.java:63-77`). Every query — including a single-partition point read — emits **one split per read-replica token range**: on the standing rig (3 nodes, num_tokens 16 → ~48 ranges) a point read costs **~48 DoGets across all 3 pods to fetch 1 partition**. Point reads don't concentrate load; they multiply it ~48×.

## Scope: plan-time split pruning for fully-bound partition keys

When the pushed-down constraint fixes the **full partition key** (equality, or IN over full keys), compute the Murmur3 token(s) at plan time and emit split(s) **only for the covering token range(s)** — a single-key point read becomes 1 DoGet to 1 pod. Anything less than a fully-bound partition key (range scan, partial PK, non-PK predicates) keeps the current full fan-out — **fail-safe: any doubt → no pruning**.

**Routing: design-driven** → OpenSpec via flow-activate (token computation source, partitioner metadata plumbing, IN handling, and the correctness harness are real design calls).

## Acceptance criteria

- [ ] Connector test: with a fully-bound PK constraint over a multi-range fixture, `getSplits` emits exactly the covering split(s) (1 for single key); with partial/absent PK constraint, split count unchanged (full fan-out).
- [ ] Correctness identity: pruned vs forced-unpruned execution returns identical rows/values/order for the same query (connector-level differential test, point-vs-full style).
- [ ] IN-list over full PKs prunes to the union of covering ranges (deduped), never fewer.
- [ ] **No-heuristics**: token derived from schema-declared partition key columns + partitioner from Sidecar metadata — never inferred from data. Unknown/unsupported partitioner → no pruning (fail-safe), surfaced in logs.
- [ ] **Wiring evidence**: an end-to-end test through the real Trino query path shows a point read served by a single DoGet (assert via do_get/pruning counter or split count on the public surface).
- [ ] Field verification (report-only, next round): per-pod do_get counts collapse for point workloads; warm point p50 through Trino approaches the server floor; qps @ 32 threads rises off ~39.

Cross-refs: #2661 (soak), #2367 (round history), #2397/PR #2409 (split primary rotation), Lane-B goals `docs/architecture/performance-goals-2026-07.md` (B1/B2).

pmcfadin:
Closes #2679

## What

The Trino/Flight connector's `CqliteFlightSplitManager.getSplits` ignored Trino's pushed-down `Constraint`, so **every** query — including a single-partition point read — emitted one split per read-replica token range (~48 DoGets across all pods to fetch 1 partition). This adds **plan-time split pruning**: when the constraint **fully binds the partition key** (equality on every PK column, or IN over full keys), the connector computes the Murmur3 token(s) at plan time and emits splits **only** for the covering token range(s). A single-key point read becomes **1 DoGet to 1 pod**.

Anything less than a fully-bound PK (range scan, partial PK, non-PK predicates) keeps the current full fan-out. **Split elimination is load-bearing for correctness** (a mis-pruned split silently drops rows), so pruning is **fail-safe on any doubt** → full fan-out.

## How

- Read `constraint.getSummary()` (a `TupleDomain`) in `getSplits` — where Trino delivers full-PK point reads.
- New `Murmur3Token` (byte-exact port of Rust `cassandra_murmur3_token`) + `PartitionKeyBytes` (canonical PK byte layout: single-component = raw bytes; multi = `[len:u16 BE][bytes][0x00]` per component), pinned to the Rust authority via shared vectors.
- New `BoundPartitionKeys` classifies equality / IN (deduped union, `MAX_KEYS` guard) / not-bound from the summary domains vs `PrimaryKeyExtractor` PK columns; capability-gated VARCHAR disambiguation (text→UTF-8, uuid/timeuuid→16 bytes; inet/varint/decimal/duration → no prune).
- `Partitioner.resolve()` — explicit "assume Murmur3Partitioner" (owner-approved; Sidecar exposes no partitioner field); unknown/non-Murmur3 → no prune + log.
- `cqlite.split-pruning-enabled` toggle (default on) so the differential baseline can force pruning off.
- `(start, end]` half-open + wraparound membership reuses the existing `validateRingCoverage` / server `token_in_half_open_range` convention.

## OpenSpec

Change `plantime-split-pruning` (owner-approved at Seam 1). Design decision: explicit Murmur3 assumption (Sidecar has no partitioner field; the ring already hard-assumes Murmur3). Follow-up to surface partitioner name from Sidecar is filed.

## Tests (all 6 spec requirements)

- `Murmur3TokenTest` — Java tokens byte-identical to Rust `cassandra_murmur3_token` (text/int/bigint/uuid/blob/composite + `i64::MIN→MAX` normalize).
- `CqliteFlightSplitPruningTest` — single fully-bound PK → 1 covering split; partial/absent/range → full fan-out; IN deduped union; shared-range collapse; unknown partitioner / un-serializable value / toggle-off → no prune; **uuid/timeuuid prune at the parsed-16-byte token**; non-canonical uuid → full fan-out.
- `CqliteFlightSplitPruningWiringTest` — e2e through the public `getSplits`→`SplitSource`: point read yields exactly 1 split; IN yields union; **differential** pruned vs forced-unpruned cover the key identically and pruned is fewer.

**Gradle connector tests: PASS** (376 tests, 27 new). No Rust changes — the server already prunes by token at DoGet time; this pushes the decision earlier so the fan-out never happens.

Reviews (review-first, pre-gate): roborev — no issues found; rust-reviewer — clean, no blockers (verified Murmur3 byte-exactness, byte layout, fail-safe comprehensiveness, and the capability-gated VARCHAR disambiguation line-by-line).

### DIFF
--- a/openspec/changes/plantime-split-pruning/design.md
+++ b/openspec/changes/plantime-split-pruning/design.md
@@ -1,0 +1,106 @@
+# Design — plantime-split-pruning
+
+## Context
+
+`CqliteFlightSplitManager.getSplits(txn, session, table, dynamicFilter, Constraint)` today ignores its
+`Constraint`. It calls `sidecar.tokenRangeReplicas(keyspace)` → ~48 read-replica ranges, validates ring
+coverage, and `buildSplits` emits **one split per range** unconditionally. The pushed-down predicate is
+carried on each split's ticket (`FlightTicketJson`) and applied server-side (best-effort, post-decode).
+Trino delivers simple full-PK point reads as a `TupleDomain` in `constraint.getSummary()` (expression is
+`TRUE`); `applyFilter` already translates equality/IN/range domains (`PredicateTreeTranslator`), and
+`PrimaryKeyExtractor` already yields partition-key column names + order from the DDL.
+
+The server already makes exactly the "is this a full-PK point read" decision (`point_read::detect_route`
+→ `PartitionPointRead`/`MultiPartitionPointRead`/`Scan`) and prunes by token at DoGet time
+(`token_in_half_open_range`). This change moves that decision **earlier** — to plan time — so the ~48×
+fan-out never happens for a bound-PK query.
+
+## Design calls (the real decisions)
+
+### 1. Token-computation source — port Murmur3 to Java (chosen) vs a server round-trip
+
+There is **no Murmur3 in Java today**. Two options:
+
+- **(A · chosen) Port the token computation to Java.** Add `Murmur3Token` mirroring the Rust authority
+  `cassandra_murmur3_token` (`cqlite-core/src/util/cassandra_murmur3.rs`) plus the canonical
+  partition-key byte layout from `PartitionKey::to_bytes` (`storage/write_engine/mutation.rs`):
+  single-component = the raw serialized value bytes; multi-component = for each component
+  `[len:u16 BE][value bytes][0x00]` concatenated. Compute the token(s) at plan time, no RPC.
+- **(B · rejected) Add a Sidecar/Flight RPC** that returns the covering range for a key. Rejected: adds a
+  network hop on the hot path we're trying to shorten, introduces a new endpoint + failure mode, and still
+  needs the Java-side value→bytes serialization anyway. The token math is small, deterministic, and
+  already has a byte-exact Rust oracle to test against.
+
+**Why A is safe:** the computation is pure and testable against the Rust implementation with shared
+vectors. Correctness risk (a wrong token drops rows — see §4) is contained by a differential test that
+compares pruned vs forced-unpruned execution, and by fail-safe defaults.
+
+### 2. Partitioner metadata under no-heuristics — assume Murmur3 explicitly, fail-safe otherwise
+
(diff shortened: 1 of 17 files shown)


pmcfadin:
Endgame (flow-closer): deferred NITs #2 (`BoundPartitionKeys` `Optional<long[]>` record equals/hashCode) + #3 (timestamp-PK no-prune) and planned tasks 6.1 (partitioner name from Sidecar) + 6.2 (report-only field verification: per-pod DoGet collapse, warm point p50, qps @ 32 threads) are batched into follow-up #2775. None block merge.

Issue #2388, opened 2026-07-13
Add commit log writer


rustyrazorblade:
Feature and test parity with Cassandra

PR #2797, opened 2026-07-20
feat(#2389): Cassandra 5.0 CommitLog reader (descriptor, frame, mutation decode, CLI)

### LINKED ISSUE #2389: Add CommitLog reader — parse Cassandra 5.0 commit log segments into mutations
## Summary
Add a `CommitLogReader` that parses Cassandra 5.0 CommitLog **segment files** (the raw file Cassandra itself writes via `CommitLogSegment`/`CommitLogDescriptor`) into a stream of decoded mutations — a second on-disk format CQLite reads, alongside SSTables. This is the **reader** half of "read and write commit logs." The writer (#2388) is an explicit, separate follow-on issue — not in scope here, per owner decision (reader first).

## Why now / where this fits
Serves the freshness/OLAP-path exploration in #2037 (ArrowMemtable epic) — a "commit-log watermark" is one of the freshness primitives that epic anticipates, and it needs an actual reader to exist. Independent of that epic, a CommitLog reader is also useful standalone (crash-recovery inspection, forensic/debug tooling, future CDC-style export).

## Oracle vs design
**Design-driven.** No sstabledump-style JSONL oracle exists for CommitLog segments (unlike SSTables), so this goes through OpenSpec (`flow-activate`) — the design must propose the test/fixture strategy, the module boundary, and the v1 scope line below.

## Background (existing research)
`docs/storage engine/cassandra-index/commitlog-cdc.md` already indexes the Cassandra-side format: `CommitLogDescriptor` (segment id, version, compression/encryption), `CommitLogSegment` (sync markers, CDC state), `CommitLogReader`/`CommitLogReplayer` (CRC-validated mutation deserialization, `minPosition` seek, `tolerateTruncation`). Use it as a starting map, not gospel — verify field-for-field against actual Cassandra 5.0 source, the way SSTable parsing does.

CQLite already parses `CommitLogPosition` (segmentId + position) as **metadata inside Statistics.db** (`cqlite-core/src/parser/repair_metadata.rs`, `storage/sstable/writer/stats_writer/components.rs`) — that's just an 8+4 byte watermark reference, not the segment format itself. Zero existing code reads an actual `CommitLog-*.log` file.

## Scope (v1)
In scope:
- Parse the `CommitLogDescriptor` header (segment id, version int, compression params) — version-gated the way BIG/BTI SSTable versions are, Cassandra 5.0-era commitlog version only.
- Deserialize the mutation stream: per-record CRC validation, sync-marker handling, clean (non-panic) stop at a torn/truncated tail — mirrors `tolerateTruncation`.
- Uncompressed segments at minimum; the design states explicitly whether LZ4/Snappy-compressed segments are in scope or deferred.
- A named public surface (library API and/or a CLI subcommand alongside existing SSTable dump commands) exercised by an end-to-end test — not just internal unit tests (wiring-evidence rule).
- No-heuristics: segment version/format read from the descriptor header authoritatively, never inferred from file size or byte-pattern sniffing.

Out of scope (explicit follow-ups — do not silently fold in):
- **#2388 (commit log writer)** — separate issue, separate PR.
- CDC-specific behavior: `cdc_raw` hard-link semantics, `.cdc-index` polling/tailing, the `CDCState` machine — that's live-tailing/sidecar work, not segment parsing. Note as a likely v2 in the design doc; don't build it here.
- Encryption support.
- Wiring this reader into the query/Flight surface (e.g. as a freshness source for #2037) — this issue only proves the parser is correct; integration is separate.

1. Given a real Cassandra 5.0 CommitLog segment produced by an actual Cassandra node (generated the same way SSTable test-data is — Docker fixture, not hand-crafted bytes), `CommitLogReader` parses the descriptor header without error and reports the correct segment id/version.
2. For a segment containing N known mutations (inserted via CQL against the same test node before the segment is captured), the reader decodes exactly those N mutations with matching table/partition/cell values — the oracle is "what we inserted" vs "what the reader produced," analogous to CQLite's existing query-semantics-oracle pattern for SSTables.
3. A deliberately truncated/torn final segment is handled per `tolerateTruncation` semantics: mutations that decoded cleanly before the tear are returned and the truncation is reported; never panics/`unwrap()`s.
4. A malformed/fuzzed segment (bad CRC, corrupt header) returns a typed `Error`, never panics — matches the existing fuzz-target safety bar (`fuzz/`); add a fuzz target if the design calls for one.
5. Public-surface end-to-end test (wiring-evidence): a named entry point (library fn and/or CLI command) is exercised by an integration test reading a real segment file end to end.
6. Memory: no unbounded materialization of a large segment (Cassandra caps segments at 32MB by default) — the design states whether the reader streams or whole-loads, matching the `oom-audit` gate component's structural expectations.

## Notes for the design (Seam 1)
- Module naming must avoid confusion with CQLite's own internal WAL (`storage/write_engine/wal.rs` — a different, CQLite-native format used for CQLite's own write-support feature). The design should pick and justify a distinct name (e.g. `storage::commitlog` / `cassandra_commitlog`).
- Decide where fixture/oracle generation lives — likely extends `test-data/scripts/` the way SSTable dataset fetch/generation works, since there is no `sstabledump`-equivalent for CommitLog segments.

Related: #2037 (ArrowMemtable/OLAP freshness epic — a future consumer, not a blocker), #2388 (commit log writer — explicit follow-on, separate issue/PR, not in this pass).

---
Priority: P2 — new capability, not blocking the active 0.15 milestone theme (cqlite-trino latency/throughput). Flag if a different priority is intended.

rustyrazorblade:
## Summary

Adds a `CommitLogReader` that parses Cassandra 5.0 CommitLog segment files (descriptor header, CRC-framed sync sections, schema-aware mutation/cell decode) into decoded mutations — a second on-disk format CQLite reads, alongside SSTables. Ships as a library API (`cqlite_core::storage::commitlog`) plus a `read-commitlog` CLI subcommand.

Design-driven (OpenSpec change `commitlog-reader`, approved at Seam 1). This is the **reader** half of "read and write commit logs" — the writer (#2388) is an explicit, separate follow-on, not in this PR.

Closes #2389

## What's in scope (per the approved spec)
- Descriptor parsing + version gate (Cassandra 5.0-era commitlog version only)
- CRC-framed sync-section walker, torn-tail tolerance
- Schema-aware mutation/cell decode for the common insert path
- Honest bail (never guess) for unmodeled constructs: clustering columns, static rows, complex/collection columns, partition/row deletions, range tombstone markers
- Compression/encryption: fail-closed, not misdecoded
- Fuzz target, real-Cassandra-produced parity fixtures + insert-set-vs-decoded-set oracle
- `read-commitlog` CLI subcommand (public-surface wiring evidence)

## Review-first (before any full gate)
- `rust-reviewer` pass — found and fixed one blocker (clustering-prefix header not decoded, silently misaligning fields for clustered-table schemas)
- **roborev, 6 rounds** (`claude-code`/`opus`) — each round's blockers fixed, re-certified with `--lite`, re-reviewed:
  1. Frame walker: torn-section `truncated_end` not propagated through `CleanEnd`; mid-segment padding treated as whole-segment end instead of section-end
  2. Static-row column read could leave misparsed `column_names` visible even on the honest-bail path; multi-section frame walking had zero test coverage
  3. **Counter columns** treated as fixed-8 bytes (actually a vint-length-prefixed `CounterContext`), and **collection/tuple/UDT columns** claimed-but-not-actually bailed — both silently misaligning data with `rows_decoded == true`. Converged into one systematic `is_simple_scalar_type` allowlist. Also: `parse_table_id` panicking on non-ASCII input, `--limit` reporting a false "clean" verdict when it cut iteration short
  4. `truncated_end` monotonicity bug (assign vs `|=`) at a second site; symmetric present-but-null bug in encryption-key detection; `--limit 0` edge case
  5. `has_partition_deletion` not set before early returns (wrong-looking authoritative field); batch mutations with unreadable updates silently reporting only the first partition update with no signal
  6. Only Medium/Low findings remained (field-semantics ambiguity, defensive-parsing edge cases) — batched to a follow-up issue rather than continuing indefinitely; see linked nit-batch issue

**Note on this repo's roborev availability:** roborev was not installed on the lead's machine at the start of this session; it was installed partway through, so rounds 1+ ran for real (nothing here is simulated).

## Known gaps (explicit follow-ups, not silently deferred)
- Whether the static-columns header read is gated on the right condition needs verification against actual Cassandra 5.0 source (unavailable in this environment) — documented in code + the appendix-h doc as an open risk; the honest-bail fix in round 5 means a wrong assumption here can no longer leak bad data, regardless of how it's eventually resolved.
- A nit-batch follow-up issue (opened alongside this PR) covers everything triaged as non-blocking across all 6 roborev rounds.

## Test plan
- 27 tests: 20 unit (incl. regressions for every fixed bug), 3 integration parity (real Cassandra 5.0.2 fixtures), 3 CLI end-to-end, 1 dhat streaming-memory
- `--lite` gate: PASS (file-size, fmt, clippy, roborev-lints, scoped-tests, python-tier) — `CQLITE_ALLOW_FILE_GROWTH=1` required (this change's necessary additions to `error.rs`/`main.rs` push two already-over-threshold files further over; splitting them is out-of-scope epic #1116 work)
- Full gate + spec-auditor (C) intent audit + final roborev + merge: handled by `flow-closer`


rustyrazorblade:
Nit-batch follow-up filed: #2798 (everything triaged as non-blocking across rust-reviewer + 6 rounds of roborev).

pmcfadin (changes requested):
# Consolidated deep review — 4 parallel audits (code, spec, test-quality, format-vs-Cassandra-source)

This PR was investigated by four independent review lanes, including a field-for-field verification of the parser against the **actual Apache Cassandra 5.0.2 source** (via `git show cassandra-5.0.2:<path>` — the exact version the fixtures were generated from). Overall: this is a strong, honest implementation — real 5.0.2 fixtures with a genuine insert-vs-decoded oracle, honest-bail integrity holding on nearly every path, zero `unwrap`/`expect`, clean no-heuristics posture, disciplined scope. And the good news first:

## ✅ Your #1 flagged open risk is RESOLVED — the static-columns gating is RIGHT

The PR body flags "whether the static-columns header read is gated on the right condition needs verification against actual Cassandra 5.0 source." Verified: it is correct, not right-by-accident. `SerializationHeader.serializeForMessaging` (cassandra-5.0.2, `SerializationHeader.java`):

```java
EncodingStats.serializer.serialize(header.stats, out);
if (selection == null) {
    if (hasStatic) Columns.serializer.serialize(header.columns.statics, out);
    Columns.serializer.serialize(header.columns.regulars, out);
}
```

with `hasStatic = (staticRow != Rows.EMPTY_STATIC_ROW)` ORed into the iterator flags as `HAS_STATIC_ROW` (0x08), and the mirror `deserializeForMessaging`: `statics = hasStatic ? deserialize(in) : Columns.NONE;`. Both the static column subset in the header AND the static row body are present **iff flag 0x08** — exactly what `mutation.rs` gates on. Critically, the Rust bails *before* `read_column_names`, which is the right ordering since statics precede regulars in the header. This can come off the risk register (update the appendix-h open-risk note in this PR).

---

## ⛔ Blocker 1 — `tinyint`/`smallint`/`date`/`time` decoded as fixed-length: silent data corruption

`schema.rs` `cql_fixed_len` returns `Some(1)`/`Some(2)`/`Some(4)`/`Some(8)` for `tinyint`/`smallint`/`date`/`time`, and all four are on the `is_simple_scalar_type` allowlist, so `read_typed_value` reads that many **raw bytes with no length prefix**.

In Cassandra 5.0.2, none of these four types (nor their parents `NumberType`/`TemporalType`) overrides `valueLengthIfFixed()` — all inherit `AbstractType.valueLengthIfFixed() → VARIABLE_LENGTH == -1` (`AbstractType.java:490`), so `writeValue` **vint-length-prefixes** them (`AbstractType.java:539` vs `:590`). For a `date` cell Cassandra writes `[uvint len=4][4 bytes]`; the Rust consumes 4 bytes starting at the *length byte* → wrong value AND every subsequent cell/row in the partition is misaligned, **while `rows_decoded` still reports `true`**. This is the one uncontained escape from the honest-bail design in the whole PR.

Note this repo already encodes the correct answer: `cqlite-core/src/parser/repair_clustering.rs:65-78` carries an AUTHORITY NOTE warning against exactly this mistake ("Classifying them as `Fixed(1/2/4/8)` would read the first VALUE byte as the VInt length and misalign the cursor"). The existing fixtures are int/text-only, so tests stay green.

**Fix:** route all four through the variable-length (length-prefixed) path; add coverage exercising at least one of them (synthetic bytes are fine; extending the fixture insert set is better).

## ⛔ Blocker 2 — compression/encryption detection uses JSON keys Cassandra never writes: fail-closed guard defeated

`descriptor.rs` `parse_params` looks for a `"compression"` object (nested `.class`) and `"encryptionContext"`/`"encryption"` keys. Cassandra 5.0.2 `CommitLogDescriptor.constructParametersString` writes `COMPRESSION_CLASS_KEY = "compressionClass"` (string) + `COMPRESSION_PARAMETERS_KEY = "compressionParameters"` (map); encryption params are `"encCipher"`/`"encKeyAlias"`/`"encIV"` (`EncryptionContext.java:41-43,138-150`). Cassandra never emits a `"compression"` object or an `"encryptionContext"` key.

Result: on a **real** compressed or encrypted 5.0 segment, `compression_class = None` / `encrypted = false`, `is_unsupported_payload()` is false, and the D3 fail-closed guard in `reader.rs:76-89` never fires — the reader walks into an undecodable body and surfaces a misleading `CorruptCommitLogFrame` by CRC luck instead of the spec-required typed `UnsupportedFormat`. The unit test `detects_compression_class_from_params` passes because it uses a hand-invented params shape.

The spec audit independently flagged the same scenario as uncovered: the spec requirement "WHEN `CommitLogReader::open` is given a segment whose descriptor specifies a compressor class THEN it returns a typed Error and does not attempt to decode" has zero coverage through `open` or the CLI.

**Fix:** parse the real keys (`compressionClass`, `encCipher`/…); add a public-surface test that writes a descriptor with **Cassandra-shaped** params to a temp file and asserts `CommitLogReader::open` returns the typed error (same for the encryption branch). That test written against real key shapes would have caught this bug.

---

## Medium (fix in this PR)

- **M1 — unbounded allocation from a hostile CRC-valid record** (`mutation.rs:416-425`). The `num_updates` remaining-bytes guard from an earlier round was never applied to the symmetric column-count loop: `count = c.uvint()?` is unguarded, and CRC32 is forgeable, so a crafted ~128MB body of `0x00` yields ~3GB of empty `String`s through the public `open().mutations()` path — defeating the <128MB target and the module's stated hostile-file claim. Fix mirrors the existing guard: reject `count` > remaining cursor bytes. (Weaker cousin: `num_updates ≈ body.len()/18` empty partitions materializes ~5× body in `PartitionUpdate` structs.)
- **Missing revert-would-fail regression tests for 4 of the ~11 roborev-fixed bugs** — all in the honest-signal fields that are this decoder's value proposition, all byte-craftable with helpers the suite already has (`ALICE_BODY`, `build_segment`):
  1. batch `updates_complete` (`mutation.rs:75`) — zero assertions at library or CLI level; a body of `02` ++ rest-of-`ALICE_BODY` yields `updates.len()==1, updates_complete==false`
  2. static-row bail (`mutation.rs:217`) — craft `iter_flags & 0x08`, assert `rows_decoded==false && column_names.is_empty()`
  3. `has_partition_deletion` (`mutation.rs:199`) — never asserted `true`; craft `iter_flags = 0x05`, assert it propagates
  4. `--limit` CLI path (`read_commitlog.rs:87`) — none of its three fixed behaviors exercised (`--limit 1` → `limited:true` + `truncated:null`; `--limit 0` → `mutation_count:0`)
- **Version-gate caveat (doc):** 5.0 writes commitlog version **8** (VERSION_50) when `storage_compatibility_mode` is `NONE`/`UPGRADING`; version 7 only under the default `CASSANDRA_4` mode. The exact-match-7 gate rejects v8 with a typed error (honest, contained) — but appendix-h's "version 7 for 5.0" claim needs the caveat, and v8 support is a legitimate follow-up.

## Nits (batch to the existing follow-up issue — no re-verify round)

- Torn-vs-corrupt classification: size-field EOF-overrun check fires before the size CRC is validated (`frame.rs:234-291`); Cassandra validates the size checksum first, and `0 < size < 10` is a mutation ERROR (CASSANDRA-2128), negative size likewise — Rust treats `size <= 0` as padding. Signal-accuracy only; both stop reading.
- Valid-CRC marker with backward `nextMarker` → Rust hard `CorruptCommitLogFrame`; Cassandra treats it as tolerable truncation (`CommitLogSegmentReader.java:194-196`). Classification-only.
- `allowSkipSyncMarkerCrc` recovery (zeroed-marker, non-atomic write) not implemented — matches default Cassandra (flag defaults false); diverges only under non-default config. Contained.
- `parse_params` fail-open on unparseable JSON → safer to treat as unsupported payload (consistent with honest-bail).
- `truncated_end = false` hard-assign at the `next_marker == 0` clean-end path (`frame.rs:167`) vs the `|=` invariant everywhere else — provably unreachable-after-true today, but add the safety comment or `|=`.
- TOCTOU between the `fs::metadata` size-cap check and `fs::read` (`reader.rs:64-73`) — negligible for fixed-size segments.
- Frame `CleanEnd` truncated-honoring + second monotonicity site (`frame.rs:124-136,179`) not individually pinned by tests.

## What the audits confirmed solid (no action)

25 format areas verified field-for-field against 5.0.2, including: descriptor layout + header CRC feed (big-endian `updateChecksumInt`), sync-marker layout + CRC over `(id, markerPos)`, record framing + continuous body CRC, `Mutation.serializer` shape (no keyspace/top-level key), PartitionUpdate header order, EncodingStats 3-uvint order, row/cell flags, the **counter CounterContext fix** (`CounterColumnType` inherits VARIABLE_LENGTH — the round-3 fix is right), vint decode vs `VIntCoding.java`, fixtures byte-verified as genuine `cassandra:5.0.2` output, and honest-bail integrity on every traced path except Blocker 1. The dhat memory bound is meaningful for decoded-set materialization (whole-file load is disclosed by design). Scope is clean — no writer/CDC/query-wiring creep.

---
*Review lanes: rust-reviewer (fresh independent pass), spec-auditor (per-requirement C-style audit), coverage-reviewer (revert-would-fail analysis), and a format-adjudication agent reading cassandra-5.0.2 source directly. Posted by the delivery lead on behalf of the owner.*

pmcfadin (approved):
Re-review of the fix round (60d3162e7..e4ac0c0f2) — **APPROVED**. Both blockers verified cleared against cassandra-5.0.2 source by independent audit: D1 (tinyint/smallint/date/time now vint-length-prefixed; all 8 remaining fixed-length claims re-verified one-by-one against valueLengthIfFixed() overrides, zero misclassifications left; alignment regression test fails on revert) and D2 (verbatim Cassandra param keys compressionClass/encCipher/encKeyAlias/encIV; absent-vs-null shape matches constructParametersString; fail-closed proven through public CommitLogReader::open for both compressed and encrypted). M1 guard landed (u64::MAX case killed, with_capacity capped). Residual medium (count≤remaining still admits ~3GB from a crafted 128MB body — comment overstates the bound) + 3 test/doc nits are batched to a linked follow-up issue per severity triage, not merge-blocking for an offline diagnostic surface. Proceeding to the gate-of-record endgame.

pmcfadin:
File-size ratchet growth acknowledged via `CQLITE_ALLOW_FILE_GROWTH=1` on the gate of record: `cqlite-cli/src/main.rs` +10 (1420→1430) and `cqlite-core/src/error.rs` +27 (851→878). Both files were over-threshold **before** this PR; the growth is pre-documented in the PR body and splitting them is out-of-scope campsite work tracked by epic #1116. Full gate PASS at e4ac0c0f2 with this opt-out; all other components clean.

Issue #2805, opened 2026-07-21
field: cqlite 0.16.0-rc1 targeted fix-validation round (AWS easy-db-lab) — probes + short drift soak + HTML report set


pmcfadin:
## Purpose — verify the specific 0.16 fixes in the field

This round is **not** a whole-milestone soak (that was #2661 for 0.15). 0.16 shipped a small set of
targeted Flight + connector fixes; this round **verifies each one with a named probe and a pass/fail
threshold**, then runs a **short drift soak** (~4–6h) to catch regressions the probes can't. Every
probe below returns a verdict: **verified / regressed / could-not-observe** (with evidence).

We are pinning a **pre-release field build (`0.16.0-rc1`)**, not a released version — the 0.16 milestone
is intentionally still open. Pin everything by digest.

## Pins (pin every pod by digest)

| Component | Pin |
|---|---|
| **cqlite-flight image** | `ghcr.io/pmcfadin/cqlite-flight:0.16.0-rc1` @ **`sha256:79c7907dee06b5d9452aedca6dfbec1c21aa8bfd819ebb1f2276a5042d704478`** (multi-arch INDEX: linux/amd64 + linux/arm64; source `f4f13a7cb`) |
| **Trino connector** | `in.mcfad:cqlite-trino:0.16.0-rc1` (Maven Central) — **carries new 0.16 connector code (#2679 split pruning), do NOT reuse the 0.15.0 artifact** |
| Knobs (unchanged) | snapshot-reuse-window 3s / retire-grace-ms 10m; `--max-concurrent-scans 64` |
| **New knob** | `cqlite.split-pruning-enabled` (catalog property, default `true`) — set `false` for the #2679 differential baseline |
| Testbed | standing easy-db-lab rig (#2289); same datasets/tables as rounds 11b–13 / soak-0.15 |

**Verify the pin before attributing any verdict** (D14): every pod must run the digest above.
```
kubectl get pods -l easydblab.com/kit=cqlite-flight -o jsonpath='{range .items[*]}{.metadata.name}{"  "}{.status.containerStatuses[0].imageID}{"\n"}{end}'
```

## Phase −1 — metrics sanity (MANDATORY before any measurement)

Three rounds running, the saturation gauges have been undersampled and `rate()` panels were
structurally wrong. Do these first and record the result in `run-meta.txt`:

1. **Scrape resolution:** set the OTLP export interval for the saturation family (`cqlite.flight.*`,
[...]

Issue #2680, opened 2026-07-18
trino-connector: weight-balanced split→pod assignment — floorMod rotation is count-balanced only (2–4× pod CPU skew, B2 blocker)


pmcfadin:
## Context (from #2661 v0.15.0 milestone soak + code triage)

Soak finding 1: **one flight pod carries ~650m CPU vs 140–300m** on the others — same pattern as rounds 9/10b — capping aggregate at ~39 qps (**Lane-B B2 ≥100 qps @ 32-thr not met**).

History: #2397 (all splits pinned ONE pod) was fixed in PR #2409 with deterministic per-range primary rotation `sorted[floorMod(startToken, size)]`. Triage confirms rotation fixed *participation* but not *balance*:

- The #2409 acceptance test asserts **range-count** spread only (`CqliteFlightReplicaRotationTest.java:64-68`) — nothing about bytes/partitions per pod.
- Splits carry **no weight**: `getSplitWeight()` never overridden (`CqliteFlightSplit.java:72-80`, `CqliteFlightAggregateSplit.java:34-51`); no size estimates enter assignment anywhere.
- Vnode ranges cover very unequal partition/byte slices; scan + count(*) cost ∝ partitions per range. With only ~48 ranges, whichever pod's floorMod-assigned ranges are heavy does 2–4× the CPU. Count-balanced ≠ weight-balanced.

Refuted (with evidence, keep out of scope): count(*) single-host concentration (`CqliteFlightAggregatePageSource.java:83-91` fans per-range), snapshot-host anchoring (reads always follow `split.host()`), Trino address-hint scheduling (affects Trino-worker CPU, not flight pods).

**Routing: design-driven** → OpenSpec via flow-activate. Candidate directions (design decides): (1) weight-aware greedy LPT bin-packing across owners using per-range size estimates; (2) sub-split each range into K equal token slices + rotate, and surface `getSplitWeight()` to Trino (no size-feed dependency). Possibly (2) first, (1) if spread persists.

## Acceptance criteria

- [ ] New connector test asserts **per-owner assigned WEIGHT** (Σ partitions or Σ bytes) within ≤1.25× of the mean under an RF==N fixture with deliberately unequal-weight ranges (augments the count-only test).
- [ ] Assignment stays deterministic and stable across re-planning (preserve the existing `selectionIsDeterministicAcrossInvocations` invariant).
[...]

PR #2810, opened 2026-07-21
fix(#2806): split pruning reads handle-borne bound-key tokens — prunes through the real pushdown path

### LINKED ISSUE #2806: 0.16.0-rc1: plan-time split pruning (#2679) is inert — reads empty constraint.getSummary() instead of the pushed-down filterJson
## Summary
Plan-time split pruning (#2679) is **inert** in the field: a fully-bound partition-key point read still fans out to all token ranges. The connector reads the bound key from `constraint.getSummary()` in `getSplits()`, but by then the connector's own `applyFilter` pushdown has moved the predicate into the table handle's `filterJson`, leaving the summary empty (`TupleDomain.ALL`). Pruning always fail-safes to full fan-out.

## Environment
- cqlite-flight `0.16.0-rc1` @ `sha256:79c7907dee06b5d9452aedca6dfbec1c21aa8bfd819ebb1f2276a5042d704478`
- connector `in.mcfad:cqlite-trino:0.16.0-rc1` (verified loaded)
- Trino 481, C* 5.0.8 RF=3, `cassandra_easy_stress.keyvalue` (PK `key text`), 13 read-replica ranges
- Field round #2805 (easy-db-lab rig)

## Repro
```sql
-- catalog: cqlite.split-pruning-enabled=true (default)
EXPLAIN ANALYZE SELECT key,value FROM cqlite.cassandra_easy_stress.keyvalue WHERE key = '<bound>';
-- ScanFilter ... Splits: 13     <-- expected 1
```
Differential over 20 distinct keys (do_get counter deltas via VictoriaMetrics):
- pruning ON : 114 do_get total, 5.7/query, all 3 pods
- pruning OFF: 114 do_get total, 5.7/query, all 3 pods  (IDENTICAL)

Both `WHERE key = ?` and `WHERE key IN (?,?)` reproduce. Splits = 13 (== the ring's read-replica range count) in every case.

## Root cause (source)
Connector DEBUG (`in.mcfad.cqlite=DEBUG`) on every query:
```
Split pruning skipped for cassandra_easy_stress.keyvalue (full fan-out): constraint summary binds no columns
```
`EXPLAIN` shows the predicate is fully pushed down:
```
filterJson=Optional[{"type":"Compare","column":"key","op":"Equal","value":"<bound>"}]
```
`CqliteFlightSplitManager.getSplits()` → `pruneToBoundPartitionKey(..., constraint, ...)` → `BoundPartitionKeys.compute(constraint.getSummary(), partitionKey)` hits the guard at `BoundPartitionKeys.java` (`summary.isAll()` → `"constraint summary binds no columns"`). Because `applyFilter` fully handled the predicate, Trino does not repeat it in `Constraint.getSummary()` — so pruning can never see the bound key.

## Impact
- **Correctness: intact.** Row sets identical on/off; fail-safe by design; no data loss.
- **Optimization: dead.** The #2679 benefit (1 DoGet to 1 pod for a keyed read) is never realized — every point read still fans out to all ranges/pods. This is the fourth round of the known ~39 qps / fan-out-skew ceiling; #2679 was expected to relieve keyed reads and does not.

## Suggested fix
Derive the bound partition key from the table handle's `filterJson` (where `applyFilter` placed it), not from `constraint.getSummary()` — or additionally consult the handle. A unit test that populates `Constraint.getSummary()` will pass while the real pushdown path empties it, so add a coverage test that exercises the post-`applyFilter` split path end-to-end.

Cross-ref: #2805 (field round), #2679 (the fix under test).

pmcfadin:
Closes #2806. Field-verified on 0.16.0-rc1 (AWS round #2805, Probe A): #2679's plan-time split pruning never fired — `applyFilter` pushdown moves the fully-bound PK predicate into the handle's `filterJson`, so `getSplits` receives `Constraint.getSummary() == TupleDomain.ALL` and the guard fail-safes to full fan-out on every query (13 splits / all 3 pods, identical with pruning on/off).

## What

- **Precompute at `applyFilter` time** — the only point where the typed PK domain still exists: derive the bound key's Murmur3 covering token(s) via the existing `BoundPartitionKeys`/`PartitionKeyBytes` (no new type mapping — rejecting the parse-`filterJson` route which would need a shadow CQL→Trino mapper that could drift, a no-heuristics hazard) and retain `boundKeyTokens: List<Long>` on `CqliteFlightTableHandle`. `pruneToBoundPartitionKey` prefers handle tokens, falls back to the split-time constraint.
- **Fail-safes preserved and extended**: plan-time token computation is exception-contained (a RuntimeException in the optimization path can never fail the query); the handle-token fast path is gated on `partitioner == MURMUR3` (the latent silent-row-loss trap if `Partitioner.resolve()` ever reads Sidecar metadata); partial-PK/arity/serialization guards unchanged; `cqlite.split-pruning-enabled` kill switch gates both token sources; tokens inherit across iterative `applyFilter` calls (superset-safe) and survive `applyLimit` handle rebuilds. No `TokenRangeSlicer`/sub-splits (consistent with the #2680 revert).
- **Tests (393 green, 8 pruning-specific)** drive the REAL pushdown path — `applyFilter` first, then prune with an `ALL` split-time constraint (what `getSplits` actually receives), structurally preventing the #2679 failure mode (its tests fed a populated summary that the real path empties). Positive cases mutation-verified (fail when the fix is neutralized); iterative-inherit and applyLimit-preservation tests fail if their branch is deleted; handle JSON serde round-trip; token-handle + unsupported-partitioner → full fan-out; kill-switch differential.

## Known limitation (documented, observable, intentionally out of scope)
A PK predicate arriving ONLY via the translated-expression channel (summary stays `ALL`) leaves pruning inert — correctly fail-safe, and now visible via a DEBUG log when `filterJson` was pushed but no tokens derived. Follow-up if the field ever observes it.

## Field re-validation
The `EXPLAIN ANALYZE` split-count differential (Probe A, #2805) re-runs against the next 0.16 build: expect 1 source-stage split for `WHERE key = ?`, 2-range union for `IN (k1,k2)`, full fan-out with the kill switch off — with identical row sets throughout.

### DIFF
--- a/trino-connector/src/main/java/in/mcfad/cqlite/flight/CqliteFlightMetadata.java
+++ b/trino-connector/src/main/java/in/mcfad/cqlite/flight/CqliteFlightMetadata.java
@@ -16,6 +16,7 @@ import io.trino.spi.connector.Constraint;
 import io.trino.spi.connector.ConstraintApplicationResult;
 import io.trino.spi.connector.LimitApplicationResult;
 import io.trino.spi.connector.SchemaTableName;
+import io.trino.spi.predicate.TupleDomain;
 import io.trino.spi.statistics.Estimate;
 import io.trino.spi.statistics.TableStatistics;
 import io.trino.spi.expression.Call;
@@ -219,12 +220,22 @@ public class CqliteFlightMetadata implements ConnectorMetadata {
             return Optional.empty();
         }
 
+        // Precompute the fully-bound partition key's covering token(s) NOW (issue #2806),
+        // while the TYPED partition-key domain is still in hand. applyFilter pushes the PK
+        // predicate into filterJson and returns the summary UNENFORCED, so the enforced
+        // Constraint later handed to getSplits binds no columns — the bound key must ride on
+        // the handle for the split manager to prune (issue #2679). When this call does not
+        // fully bind the PK we INHERIT any tokens a prior applyFilter call attached: the PK,
+        // once equality-bound, stays bound across the iterative applyFilter calls, and a
+        // superset of covering ranges is always correct (never drops rows).
+        List<Long> boundKeyTokens = boundKeyTokensFor(table, constraint.getSummary());
+
         // Preserve any pushed-down LIMIT (#2129): this branch only runs for a
         // non-aggregated handle, so aggregation/finalize stay empty, but a prior
         // applyLimit's cap must NOT be dropped when the handle is rebuilt.
         CqliteFlightTableHandle newHandle = new CqliteFlightTableHandle(
                 table.keyspace(), table.table(), table.ddl(),
-                Optional.of(filterJson), Optional.empty(), Optional.empty(), table.limit());
+                Optional.of(filterJson), Optional.empty(), Optional.empty(), table.limit(), boundKeyTokens);
 
         // The residual expression Trino must still evaluate (the untranslatable
         // conjuncts, ANDed). Empty residual => TRUE (fully pushed).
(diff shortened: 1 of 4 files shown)


pmcfadin:
## roborev confirmation pass — nit disposition (job 1985 @ 38659907c)

The final roborev confirmation pass on the terminal commit surfaced **0 blockers, 4 nits**
(roborev summary: "all behaviourally sound"). Per `docs/development/roborev-severity.md`,
none hit a blocker class (no correctness/parity/safety/security/unmet-acceptance; the #2806
optimization's wiring is proven by 8 mutation-verified pruning tests + the Flight↔Trino E2E
lane). Disposed as nits — behavior is unchanged in every case; batched here rather than a
re-verify round. These live alongside the expression-channel-derivation limitation already
documented in the PR body:

1. **[Medium→nit]** `CqliteFlightMetadata.java:~282` — containment handler logs at `DEBUG`
   while the sibling split-time handler logs the same exception class at `WARNING`, and the
   message uses eager string concat. Observability posture + micro-perf; no behavioral change.
   Suggested: raise to `WARNING`, use the `Supplier<String>`+`Throwable` overload.
2. **[Low]** `CqliteFlightMetadata.java:~276` — the exception-containment catch branch is not
   independently delete-verified (the feature itself is mutation-verified). Test-robustness
   suggestion, no concrete failing scenario.
3. **[Low]** `CqliteFlightSplitManager.java:~214` — summary short-circuit placed before the
   partitioner check suppresses the non-Murmur3 DEBUG diagnostic on the common ALL-summary
   call. Behavior unchanged; reorder for the operator-grep line.
4. **[Low]** `CqliteFlightMetadata.java:~232` — inert-case DEBUG message attributes a single
   cause; could also be reached via the empty-inherited catch path. Log-wording polish.

Gate of record: `RESULT: PASS` @ `38659907c` (all 30 components green). Merging on green.

PR #2835, opened 2026-07-23
chore(#2680): archive weight-balanced-splits-v2 + telemetry

pmcfadin:
Finalize stamp for #2680 (P0 #2782 hang fix, merged via PR #2833, commit 121152d03). Archives OpenSpec change `weight-balanced-splits-v2` (4 added + 4 modified requirements synced into `trino-split-weight-balance`) and records the second-cycle delivery telemetry line. Routed via PR — main blocks direct pushes.


pmcfadin:
Lead: arming auto-merge — green finalize chore (archive + telemetry) for merged #2680 (weight-balanced splits), sat unarmed 2 days. Traceable per flow autonomy rules.

pmcfadin:
Superseded by **PR #3118** — closing.

Issue #2680 is closed, its spec sync already landed on main, and its telemetry record is in the ledger — but the OpenSpec change `weight-balanced-splits-v2` was left **unarchived** in `openspec/changes/`. This branch carries that archive, but it reports `CONFLICTING`/`DIRTY` against current main (it predates a large amount of history, and the telemetry ledger it touches is a hot append-only file), so the archive has been redone fresh off current main in #3118 instead of merging a stale branch.

Confirmed there while archiving: `openspec archive` aborted with `trino-split-weight-balance ADDED failed … already exists`, proving the spec half had already been synced — only the archive move was outstanding, so #3118 does it with `--skip-specs`.

Note for the record: my initial read called this branch an *orphan* with no merge base. That was an artifact of a **shallow clone** on the worker box (35 commits, graft boundary at `d0344b76b`); after `git fetch --unshallow` it resolves a normal merge base at `121152d03`. The branch is a normal stale branch — the `CONFLICTING` status is real, the orphan claim was not.

Issue #2857, opened 2026-07-24
Spike: run the gate (tooling-tests first, then full) in a Linux container to kill host/OS crossover bugs — and measure the macOS-Docker perf tax


pmcfadin:
**Motivation.** #2849 (the `tooling-tests` hang that blocked every local gate of record for ~10h) had two root causes, both **host/OS crossover** issues specific to running a Linux-targeted CI gate natively on a dev Mac:
1. **Host-process contamination** — self-tests run real `pgrep -f "$PROC_MATCH_BUILD"` (`cargo|nextest|gate_slot_daemon`) / `PROC_MATCH_WORKER` (`claude … --agent flow-lead`) and can match the **host's** processes (a live Claude Code session + a concurrent gate) rather than only the test's own spawned PIDs.
2. **macOS BSD vs GNU tool divergence** — a supervisor loop hung on an embedded-newline marker reason under BSD `sed`/`awk`/`read`; green in CI (Linux), hangs on macOS.

Both are killed *by construction* in a Linux container: a PID namespace can't see host processes, and GNU userland matches CI exactly. This also retires a standing backlog of macOS-only gate gotchas (`gate_slot_daemon.py` exists only because macOS lacks `flock(1)`; `taskpolicy` vs `nice`; no `setsid`; `stat -f`; `mktemp` Xs-must-be-terminal).

**The catch = speed.** Docker on Apple Silicon runs Linux in a VM; the gate is compile/I/O-heavy (cargo `target/`, sccache, test-data), and bind-mount I/O is the classic tax. Gate *latency* is already our top pain, so "containerize the whole gate" risks trading a correctness pain for a throughput pain. Mitigable (named volume for `target/`, VirtioFS, in-container sccache) but must be **measured**, not assumed.

## Scope — a spike, two deliverables

**(a) Cheap immediate win — containerize the crossover-prone part.** Run the `tooling-tests` component (fast, non-compile, exactly where host-`pgrep` + BSD-tools bite) inside a Linux container. ~zero perf cost, kills the #2849 class mechanically. Decide: does this fully replace the per-test hermeticity hardening, or complement it?

**(b) Measure the macOS-Docker perf tax on a full gate.** Run the full `agent-gate.sh` in a Linux container on this Mac (target/ in a named volume, sccache inside) and compare wall-clock vs the native macOS gate — cold and warm. Report the delta. This is the data needed to decide the **local-vs-CI gate-of-record boundary** (see #2636).

[...]

PR #2860, opened 2026-07-24
fix(#2849): CLAIM_CMD='' truly disables claim stamping — unblocks the local full gate (tooling-tests hang)

### LINKED ISSUE #2849: gate: tooling-tests deterministically HANGS under CQLITE_GATE_MAX_CONCURRENCY=1 (nested self-test gates deadlock on the machine slot) — blocks local full gate of record
**Symptom:** the FULL `scripts/agent-gate.sh` deterministically **hangs** in the final `tooling-tests` component and never completes, on a machine with the pinned `CQLITE_GATE_MAX_CONCURRENCY=1` (#2640). The log freezes on this compound line and produces no further output:

```
>>> [tooling-tests] bash scripts/tests/test_agent_gate_summary.sh; bash scripts/tests/test_agent_gate_notify.sh; \
    bash scripts/tests/test_agent_gate_smoke_target_dir.sh; bash scripts/tests/test_gate_concurrency_cap.sh; \
    bash scripts/tests/test_bootstrap_agent_machine.sh; bash scripts/tests/test_claim_lock.sh; \
    bash scripts/tests/test_premerge_assert.sh; bash scripts/tests/test_worker_supervisor.sh
```

The summary file is left at the `RESULT: INCOMPLETE` start placeholder. Every component BEFORE `tooling-tests` PASSes; only the gate's own self-tests hang.

**This is DISTINCT from #2751** (which was the `AGENT_GATE_SUMMARY_FILE` inheritance *clobber*, now fixed):
- Reproduced **twice, deterministically, at the same line** on this box (2026-07-24), during issue #2165's endgame.
- Reproduced with **`AGENT_GATE_SUMMARY_FILE` unset in the environment** (confirmed `<unset>`) and the gate launched via `env -u AGENT_GATE_SUMMARY_FILE ...` — so the #2751 clobber vector is not in play.
- Symptom is a **hang** (log mtime frozen 280s+ while a gate process is alive), not a clobbered-but-completed summary.

**Hypothesis (needs confirmation):** `test_gate_concurrency_cap.sh` / `test_worker_supervisor.sh` launch nested `agent-gate.sh` invocations that try to acquire the **machine-wide gate slot already held by the parent** gate. Under `CQLITE_GATE_MAX_CONCURRENCY=1` (#2640) there is no second slot, so the nested gate blocks forever → the self-test hangs → the parent gate hangs. If so, #2640's tight cap co-caused this. Fix candidates: the nested self-test gates must run with an isolated lock/temp cap (never contend for the real machine slot), or `tooling-tests`' nested gates must be exempt from the machine-wide cap.

**Impact:** blocks the local FULL gate of record for EVERY code change on a cap=1 machine. First hit cost issue #2165 ~10h of silent stall overnight (undetected until a watchdog was added; the watchdog then caught the recurrence in ~5 min).

**Workaround in use:** certify code changes with `agent-gate.sh --only <all components except tooling-tests>` (or cite a killed run's PASS evidence for the serial lane, which stops-on-first-failure so reaching `tooling-tests` proves all prior components passed) + run `--only minimal-build` for the post-`tooling-tests` gap. Justified when the change touches no `scripts/` files (so `tooling-tests` cannot be affected by it). Used for #2165.

Refs: #2751 (closed, the clobber), #2640 (the cap pin).

pmcfadin:
Closes #2849.

**Root cause:** `scripts/local/worker-supervisor.sh` resolved `CLAIM_CMD="${CLAIM_CMD:-...}"` (colon form). The `:-` operator substitutes the default when the var is unset **OR empty**, so the tooling-tests' documented `export CLAIM_CMD=""` (meant to disable claim stamping) was silently re-defaulted to the real `claim-heartbeat.sh` — whose `git push` / `gh pr list` **network calls wedged non-deterministically**, hanging `test_worker_supervisor.sh` and therefore the whole `tooling-tests` gate component. This is why the full local gate of record hung on this box while CI (which doesn't hit the same path timing) stayed green.

**Fix:** `${CLAIM_CMD:-...}` → `${CLAIM_CMD-...}` (colonless) so an explicit empty string truly disables (no network), while **production (CLAIM_CMD unset) still gets the default** claim-heartbeat command. Verified against the callers: `stamp`/`reap` already guard `[[ -n "$CLAIM_CMD" ]] || return 0`.

**Regression tests added** (`scripts/tests/test_worker_supervisor.sh`):
- `test_claim_cmd_empty_truly_disables_no_network` — pins the fix (empty CLAIM_CMD ⇒ zero claim calls, bounded run).
- `test_real_pgrep_usages_are_pid_scoped` — enforces every real `pgrep -f` in the suite is PID-scoped (hermetic vs a host `claude`/`cargo`/`nextest`/`gate_slot_daemon` — the dev-machine crossover class).

**Verification:** `test_worker_supervisor.sh` 59/59 standalone with a live Claude Code session; `agent-gate.sh --only tooling-tests` COMPLETES + PASS (265s) — the exact component that was hanging; `--lite` PASS @ `51426c83d`.

Strategic follow-up: #2857 (evaluate containerizing the gate to kill this crossover class by construction). Note: touches an over-threshold test file (#1135 split is out of scope for a bug fix) — full gate run with `CQLITE_ALLOW_FILE_GROWTH=1` if the ratchet flags it.

### DIFF
--- a/scripts/local/worker-supervisor.sh
+++ b/scripts/local/worker-supervisor.sh
@@ -1338,6 +1338,11 @@ main() {
   validate_numeric_knobs
   acquire_lock
   log "started: MAX_ISSUES=$MAX_ISSUES MAX_HOURS=$MAX_HOURS LOAD_MAX=$LOAD_MAX DISK_FLOOR_GB=$DISK_FLOOR_GB BREAKER_N=$BREAKER_N"
+  # #2655 machine-claim liveness is disabled when CLAIM_CMD is empty (a machine with
+  # no origin push rights, or the hermetic tooling tests). Announce it ONCE so a
+  # silently-unclaimed run is visible in the log rather than mistaken for a claim
+  # that simply never refreshed (stamp_claim/clear_claim otherwise return with no line).
+  [[ -n "$CLAIM_CMD" ]] || log "claim stamping DISABLED (CLAIM_CMD empty) — no machine-claim liveness ref this run (#2655)"
   while true; do
     [[ -f "$STOP_FILE" ]] && finalize_exit "stop-file" 0
     [[ $(($(date +%s) - START_TS)) -ge "$MAX_HOURS_SECS" ]] && finalize_exit "budget-wallclock" 0
--- a/scripts/local/worker-supervisor.sh
+++ b/scripts/local/worker-supervisor.sh
@@ -232,8 +232,18 @@ LOG_DIR="${LOG_DIR:-$REPO_ROOT/logs/worker-supervisor}"
 #
 # CLAIM_CMD is the claim-heartbeat.sh entrypoint; overridable so the tests can
 # substitute a hermetic stub (no origin/network). Set CLAIM_CMD="" to disable
-# claim stamping entirely (e.g. a machine with no origin push rights).
-CLAIM_CMD="${CLAIM_CMD:-bash $REPO_ROOT/scripts/flow/claim-heartbeat.sh}"
+# claim stamping entirely (e.g. a machine with no origin push rights, or the
+# tooling tests, which must never touch origin/gh).
+#
+# Use `${VAR-default}` (NO colon), NOT `${VAR:-default}`: the colon form
+# substitutes the default for an EMPTY string too, which silently re-enabled the
+# real claim-heartbeat.sh (git push / gh pr list — network ops) whenever a caller
+# set CLAIM_CMD="" to disable it. That defeated the documented "set to empty to
+# disable" contract and let a slow/contended origin push or `gh pr list` WEDGE the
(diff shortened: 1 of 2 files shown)


pmcfadin:
**Review converged (terminal round).** roborev round 2 → 5 findings (2 blockers: hang-regression test had no time bound + a vacuous pin) → fixed at 8611aa802. roborev round 3 → 4 findings, all Low, all test-guard refinements that only matter on a future regression or are cosmetic → **batched to #2861** (convergence-by-scope; not re-verified, per the review-loop-termination rule). Core fix (CLAIM_CMD `:-`→`-`) is correct + production-safe; `--only tooling-tests` PASS proves the hang is gone. Full gate of record running now.

Issue #2897, opened 2026-07-26
Arrow egress: skip per-row byte estimation when --max-batch-bytes is disabled


pmcfadin:
Batched nit from the #2825 (T4 byte-bounded batch sizing) review round. Non-blocking — deferred per `docs/development/roborev-severity.md` (nits do not earn a re-verify round).

## Problem

`estimate_arrow_row_bytes` is called unconditionally per row on the egress hot path at `cqlite-flight/src/producer.rs:997` and `cqlite-flight/src/producer_stream.rs:212` — **including when the byte-cap is disabled** (`max_batch_bytes == usize::MAX`).

An operator or embedder who explicitly disabled the cap still pays a full per-row, per-cell walk of every value, with no possible effect on the batch boundary. On collection-heavy shapes the estimator's per-element walk roughly doubles per-row work relative to the conversion itself.

Both reviewers (roborev + rust-reviewer) flagged this independently.

## Also noted

Throughput evidence is thin: `byte_cap_throughput_is_within_the_expected_band` uses `narrow_fixture` only, so the collection-heavy shape — where the walk costs the most — has **no throughput evidence at all**. The only throughput test is `#[ignore]`d and prints a ratio without asserting.

## Acceptance criteria

- `BatchByteCap::push_row` (or the two call sites) short-circuits when the cap is `usize::MAX`; no estimator walk occurs on the disabled path. Pin with a test that counts estimator invocations, or asserts via a shape whose estimation would be observably expensive.
- Consider caching the estimator's per-column shape resolution instead of re-resolving `column_shape` per row.
- Add a collection-heavy shape to the throughput band test so the estimator's cost is measured where it is largest.
- Any throughput assertion stays `#[ignore]`d and marked `perf-gate-allow` (#2642 — wall-clock asserts in the correctness path FAIL `roborev-lints`).

## Routing

**Oracle-driven** (measurable hot-path cost, no design latitude). Skip OpenSpec.

## Provenance

Review of `issue-2825-batch-byte-cap` @ `2c7e302`, 2026-07-26. Related: #2825, #2821 (epic #2817).

Issue #2896, opened 2026-07-26
Flaky: blocking_tasks_gauge_tracks_real_streaming_do_get races on a process-global counter


pmcfadin:
Discovered while implementing #2825 (T4 byte-bounded batch sizing). **Not caused by that change** — its diff touches neither `saturation` nor `streaming.rs`.

## Symptom

`cqlite-flight/src/streaming_tests.rs:783` — `streaming::tests::blocking_tasks_gauge_tracks_real_streaming_do_get` failed once during a `--lite` round, then passed 11/11 on re-runs including under 12-core load.

## Root cause hypothesis

The test reads a **process-global** blocking-task counter into `base`, then asserts `level() > base` while holding its guard. Other `spawn_streaming` tests in the same test binary mutate that same global concurrently, so between the `base` read and the assertion the counter can be decremented by an unrelated test finishing. `level() > base` is then false while the guard is legitimately held.

This is a race on the *lower bound*, not on the feature under test — the gauge itself is behaving correctly.

## Affected

- `cqlite-flight/src/streaming_tests.rs:783`
- The gauge introduced by #2419.

## Routing

**Oracle-driven** (deterministic race in a test's own bookkeeping — no design latitude). Skip OpenSpec; fix + pinned regression test.

## Acceptance criteria

- The test derives a **race-free lower bound** rather than reading a shared global and assuming exclusivity — e.g. scope the assertion to a delta attributable to its own guard, or serialize the gauge-observing tests.
- The test passes deterministically when the full `cqlite-flight` test binary runs with parallel test threads (the condition under which it flaked).
- No wall-clock threshold assertions introduced (#2642 — `roborev-lints` FAILs on those in the correctness path).

## Evidence

Observed 2026-07-26 on branch `issue-2825-batch-byte-cap` during lite round 4; re-ran 11/11 green. Reported by the implementing agent, which correctly declined to edit another issue's test.

pmcfadin:
⬆️ **Escalating: this flake is now MERGE-BLOCKING on every `cqlite-core/**` PR.**

#2910 merged (PR #3025). Its mandate widened so a `cqlite-core/**` diff now mandates the full Flight tier, and `required` fails closed on a failed tier. So this race is no longer an occasional red in a lane people ignore — **it blocks the merge of any core PR it hits.**

[...]

Issue #2821, opened 2026-07-22
Streaming `do_get` result-budget wiring gap


pmcfadin:
The 64MiB `QueryConfig::n` result budget is enforced **only** on the materializing `CollectSink`
path; `rg result_budget cqlite-flight/src` returns nothing — the streaming `do_get` path is bounded
**only** structurally by the 4-deep batch channel + admission K. On wide rows this leaves per-stream
residency governed by row width × channel depth with no byte ceiling, a B4 hazard the current code
does not name.

## Acceptance criteria

Either:
- A **byte budget is enforced on the streaming egress** (bounded per-stream residency **independent
  of row width**), with a test pinning the per-stream ceiling under a wide-row fixture; **or**
- The gap is **documented** and admission K is made the **sole documented governor**, with a test
  pinning the per-stream ceiling (row-count × batch-channel depth) so the bound is explicit and
  regression-guarded.

The demonstrated ceiling must hold the per-stream working set within the ratified **B4 ≤16Mi
per-query** reading at concurrency 1.

## Evidence

- `docs/research/phase2-verify-parallelism.md` §2 (the streaming `do_get` path is bounded only by the
  4-deep batch channel + admission K; `result_budget` absent from `cqlite-flight/src`).
- `docs/architecture/throughput-program-2026-07.md` §7 manifest item M6.

## Dependencies

None.

**Dedup:** NEW — not covered by #2230/#2423 (which bound *intra-partition* materialization), nor by
#2765/#2600 (outer egress backpressure).

---
Part of epic #2817 (0.17 scan-path throughput program). Full plan + provenance:
`docs/architecture/throughput-program-2026-07.md` (§7 manifest item M6).

pmcfadin:
🔁 **Sequencing change — #2825 lands first (owner decision, 2026-07-26).**

Spec authoring for this issue surfaced that the acceptance clause *"the demonstrated ceiling must hold the per-stream working set within the ratified B4 ≤16Mi per-query at concurrency 1"* **cannot be met standalone**, contradicting this issue's stated `Dependencies: None`:

- **Narrow rows** (~300 B/row): 4 batches × 8192 rows × 300 B ≈ **9.8 MB** resident today. An 8 MiB egress ceiling binds *before* the existing 4-deep batch channel → buffer-depth regression on the narrow path.
[...]

Issue #2817, opened 2026-07-22
[EPIC] 0.17 scan-path throughput program


pmcfadin:
Umbrella for the 0.17 scan-path throughput program — the Phase-3 synthesis of the throughput
research program (Phase 0 ground truth, eight Phase-1 lever memos, seven Phase-2 adversarial
adjudications). Full plan, provenance, and graveyard:
`docs/architecture/throughput-program-2026-07.md` (committed via a docs PR — **PR pending**, see
the throughput-program doc PR).

**Central finding:** the through-Trino field baseline is ~10.6k rows/s/pod (R12/#2367); the gap to
the ~500k rows/s/core local ceiling is **almost entirely server-side and Trino-independent** —
~55–75% of the per-node gap is hardware+C(N), field data-plane heaviness, and cold-IO latency.
Trino is a latency floor, not a throughput tax. **Measurement leads**: the i4i cold-vs-warm profile
(M0) adjudicates the server-side buckets before any IO-heavy lever is committed.

## Verdict table

| Target | Verdict | Limiting factor (named) |
|---|---|---|
| **A4 Stage-1** — 100k rows/s/pod (server-direct) | **REACHABLE** | none binding; L1 batch-channel + fan-out-to-cores clears it comfortably |
| **A4 Stage-2** — 600k rows/s/pod = 150k/core | **NOT reliably reachable** on the pure row engine (central ~400–450k/pod) | **C(N) < 4**: a 4-vCPU pod is 2 physical cores + hyperthreads + reconcile machinery (~2µs/row) |
| **B3 Stage-1** — ≤10s (≈65k rows/s/pod) | **REACHABLE, at the edge** | needs full-core parallelism (~4×) and constants (~1.5×) to both land |
| **B3 Stage-2** — ≤3s (216–323k rows/s/pod) | **REACHABLE-AT-COST on a row feed** | the **Trino coordinator floor** eating the 3s budget + cold-start IO latency |
| **B2** — ≥100 qps @32thr warm | **REACHABLE** (stack: #2680 pod-skew balance + #2600/#2765 egress relief + warm p50 ≤~0.3s) | pod-skew (count-balanced floorMod) — the #2680 lever |
| **A2** — ≥1000 qps/pod warm | **REACHABLE-AT-COST, gated on measurement** | **unknown keyed access skew** — decoded-partition cache rests on an unmeasured hot-set hit rate |
[...]

PR #2906, opened 2026-07-26
feat(#2825): byte-bounded Arrow egress batch sizing

### LINKED ISSUE #2825: T4 byte-bounded batch sizing
Cap **MB/batch** — finish a batch on whichever of the row-cap (8192) or the byte-cap trips first.
Today the Arrow egress buffer scales to ~8MB/batch on wide rows (the 57,344-row buffer), a B4 peak
hazard. Filed as a **robustness/correctness** lever, **not** a throughput lever.

## Acceptance criteria

- Wide-row batches stay **under a configured byte ceiling** (byte-cap trips before the row-cap on wide
  rows).
- The **57,344-row Arrow egress buffer stops scaling to ~8MB/batch** on wide rows (helps B4 peak).
- Throughput impact is **~1.0–1.1×** (neutral-to-slightly-positive) — no throughput regression on
  narrow rows.

## Evidence

- `docs/research/phase2-verify-transport.md` (T4 SURVIVES minor — robustness/B4, not throughput).
- `docs/architecture/throughput-program-2026-07.md` §4 lever table (T4), §7 manifest item M11.

## Dependencies

None.

**Dedup:** NEW connector/server surface; #1476/#2230 are the read-path analogue (intra-partition
materialization), not the egress batch byte-cap.

---
`docs/architecture/throughput-program-2026-07.md` (§7 manifest item M11).

pmcfadin:
Closes #2825. Part of epic #2817 (0.17 scan-path throughput program), manifest item M11.

## What

Arrow egress batches now finish on whichever of the **row-cap** (`batch_size`, 8192) or a new **byte-cap** trips first, so wide rows can no longer scale a batch without limit. New knob `--max-batch-bytes` / `CQLITE_MAX_BATCH_BYTES`, `DEFAULT_MAX_BATCH_BYTES = 4 MiB`.

Wired at **all three** batch-build sites: `producer.rs` drive loop, `producer_stream.rs` drive loop, and `aggregate_paths` (the third was missed by the original terrain mapping and found during implementation).

## Currency — payload, not capacity (owner decision)

The cap is denominated in **Arrow payload bytes**, because the boundary decision must be made while rows accumulate, before the `RecordBatch` exists. `get_array_memory_size()` reports allocated **capacity**, which is a different quantity — measured ratios 1.72-1.80x at typical widths.

`BATCH_BYTES_CAPACITY_FACTOR = 2` is published as the worst-case payload→capacity bound. It is a genuine bound, not a fitted maximum: `MutableBuffer::reserve` gives `capacity <= max(round64(len), 2*capacity_prev)`, so per buffer `capacity < 2*len` for `len >= 64`.

## Bound

`worst_case_batch_capacity_bytes(cap, n_array_nodes, widest_row_payload) = 2 * max(cap, widest_row_payload) + 1024 * n_array_nodes`

The `max(cap, widest_row)` term is load-bearing: a single row wider than the whole cap must still be emitted (splitting one row across Arrow batches is not possible here), so that overshoot is **inherent, not a defect**. `checked_value_bytes` (`arrow_convert.rs:128`) does **not** bound per-cell size — it only rejects cumulative column length above `i32::MAX`, never clamps.

## Scope boundary — what this does NOT yet guarantee

This bounds **per-batch** size only. Per-stream egress residency remains **count**-bounded (`DO_GET_CHANNEL_CAPACITY = 4` plus ~3 in flight), so worst-case resident egress today is **~7 x 8 MiB ~= 56 MiB**. The `6 + 8 = 14 MiB` B4 composition is a **target for #2821**, not in force — #2821 adds the per-stream byte ceiling that makes it true. Documented explicitly in `batch_bytes.rs` so the #2821 implementer cannot read it as already holding.

## Estimator

New `cqlite_core::export::estimate_arrow_row_bytes` (`export/arrow_size.rs` + `arrow_size_shape.rs` + `arrow_size_render.rs`). `estimate_query_row_bytes` was deliberately NOT reused — it is Arrow-blind, counts `row.key` bytes that never enter the batch, and models zero structural overhead, so its error is signed negative exactly where it matters.

Slack constants are derived, not fitted: `ARROW_CELL_OVERHEAD_BYTES = 2*4 + 1 = 9` (an offsets buffer over n slots is `(n+1)*4`; two entries per slot cover the trailing entry for all n>=1, tight at n=1), `ARROW_VALIDITY_BYTES = 1` per slot, `ARROW_COLUMN_SLACK_BYTES = 8` charged only for shapes that can materialize childless array nodes.

## Review history — 3 rounds, 8 blockers

Reviewed by roborev + `rust-reviewer` before the full gate (review-first), then re-reviewed after each fix round.

Round 2 (5 blockers): published bound omitted the crossing-row overshoot (boundary switched push-then-test → **test-then-push**); conservatism contract false for deeply-nested empty collections; slack charged per-slot not per-buffer (**9x** over-cut on `list<int>`); empty-field-list UDT omitted field names; `Value::Text` non-UTF-8 under-counted 3x via `from_utf8_lossy`.

Round 3 (3 blockers): fixed-width columns over-charged → byte-cap bound at **>=40 int columns**, an ordinary schema (now **>=103**, `int` cost 13 → 5 B/row); `MAX_ESTIMATE_NODES` exhaustion degraded a legal schema to **one row per batch indefinitely** (budgets now reset per column and only *branching* slots spend nodes — a 65,535-entry map costs 1 node, not 131,070; no charge changed, so conservatism is preserved); module doc asserted the #2821 composition as in force.

**Extra bug found during fixes, not from either review**: routing `Text`/`Ascii`/`Varchar` to the typed arm was wrong because `convert_column_to_array` dispatches on `data_type` **alone** — a `cql_type=Text` / `data_type=Blob` column is built by `build_binary_array` and was charged **zero content bytes**. Now carried as a `TextFidelity` flag. `rust-reviewer` re-verified the full dispatch mirror and found no second instance.

Guard tests are placed **past** the predicted break point (64 and 100 columns), after two rounds where a guard sat just inside the boundary and gave false assurance.

## Measured

| shape | before | after |
|---|---|---|
| `list<int>` x1000 estimate vs realized | 9x | 1.2x |
| 30-int-column rows before 4 MiB cap | 4,000 | 10,750 |
| byte-cap binds at N int columns | >=40 | >=103 |

## Tests

Synthetic, deterministic, self-contained wide-row fixture (`wide_row_fixture.rs`, `test-util`-gated) — no dependence on the fetched `test_wide_rows` dataset, so it cannot pass vacuously on an empty dataset. Includes an in-tree pre-change control (`without_the_byte_cap_the_wide_scan_is_one_oversized_row_cut_batch`) and a discriminating conservatism property test that a plausible naive estimator actually trips. End-to-end test spawns the real binary and streams a real `do_get`.

Payload assertions were **tightened** across rounds (`cap + one row` → `cap`); no test was weakened.

## Notes for the gate

Run with `CQLITE_ALLOW_FILE_GROWTH=1` — `producer.rs` (3,330→3,396) and `service.rs` (2,036→2,068) were **already far over** the ~800 threshold before this change (epic #1116). All new logic lives in new files, each well under threshold.

## Follow-ups filed

- #2896 — pre-existing `blocking_tasks_gauge_tracks_real_streaming_do_get` flake (races a process-global counter; not caused by this change)
- #2897 — skip per-row estimation when the cap is disabled (`usize::MAX`), plus missing collection-heavy throughput evidence

## Spec

OpenSpec change `egress-batch-byte-cap` — 10 requirements, 33+ scenarios, updated across all three rounds to stay in step with the code. `openspec validate --strict` passes.

### DIFF
--- a/cqlite-core/src/export/arrow_size.rs
+++ b/cqlite-core/src/export/arrow_size.rs
@@ -571,14 +571,21 @@ impl<'a> Estimator<'a> {
     /// unlike the high-fidelity path, the flat builders have a FIXED one-level
     /// child shape, so a constant suffices.
     fn charge_flat(&mut self, dt: &DataType, fidelity: TextFidelity, value: Option<&'a Value>) {
-        // `build_string_array`'s two branches: an authoritative text column
-        // borrows the value's own bytes (`s.len()`, hard error on invalid
-        // UTF-8), everything else renders lossily (up to 3x). Both reach this
-        // function through the SAME flat `data_type` arms, so the branch is
-        // selected here rather than by the shape (review B5).
-        let charge_string = |est: &mut Self, value: Option<&'a Value>| match fidelity {
-            TextFidelity::Strict => est.add(text_content_bytes(value)),
-            TextFidelity::Lossy => est.charge_rendered(value),
+        // `build_string_array`'s two branches, which both reach this function
+        // through the SAME flat `data_type` arms — so the branch is selected
+        // here rather than by the shape (review B5).
+        let charge_string = |est: &mut Self, value: Option<&'a Value>| match (fidelity, value) {
+            // BOTH branches borrow a `Value::Text`'s own bytes after a NON-lossy
+            // `str::from_utf8` and hard-error on invalid UTF-8, so a top-level
+            // text cell is exactly `s.len()` — no U+FFFD expansion. (The lossy
+            // `format_value` path applies only to a `Value::Text` NESTED in a
+            // rendered container; `charge_rendered` charges 3x there.)
+            (_, Some(Value::Text(s))) => est.add(s.len()),
+            // Strict: every other variant makes the builder hard-error, so no
+            // batch exists to charge for.
+            (TextFidelity::Strict, _) => {}
+            // Lossy: `ValueFormatter::format_value` renders it.
+            (TextFidelity::Lossy, v) => est.charge_rendered(v),
         };
         match dt {
             DataType::Boolean | DataType::TinyInt => self.add(1),
--- a/cqlite-core/src/export/arrow_size.rs
+++ b/cqlite-core/src/export/arrow_size.rs
@@ -5,7 +5,7 @@
 //! The `cqlite-flight` egress path finishes a record batch on **row count**
 //! alone, so a batch's byte size is `batch_size × row_width` — an unbounded
 //! function of schema shape. Bounding it needs a byte decision made **while rows
-//! accumulate**, before any [`RecordBatch`] exists: building the batch to
+//! accumulate**, before any `RecordBatch` exists: building the batch to
 //! discover it is too big defeats the purpose, so
 //! `RecordBatch::get_array_memory_size()` cannot be the production trigger (it
 //! is only readable *after* every value has been allocated and copied).
@@ -38,10 +38,10 @@
 //!
 //! | term | charged | covers |
 //! |---|---|---|
-//! | [`ARROW_VALIDITY_BYTES`] | every slot | the slot's validity **bit**; `n` bytes ≥ `ceil(n/8)` |
+//! | `ARROW_VALIDITY_BYTES` | every slot | the slot's validity **bit**; `n` bytes ≥ `ceil(n/8)` |
 //! | content | every slot | the slot's data-buffer bytes |
-//! | [`ARROW_CELL_OVERHEAD_BYTES`] | variable-width slots | its offsets entry + the buffer's trailing `n+1`-th entry |
-//! | [`ARROW_COLUMN_SLACK_BYTES`] | once per projected **column** | array nodes that correspond to no slot at all |
+//! | `ARROW_CELL_OVERHEAD_BYTES` | variable-width slots | its offsets entry + the buffer's trailing `n+1`-th entry |
+//! | `ARROW_COLUMN_SLACK_BYTES` | once per projected **column** | array nodes that correspond to no slot at all |
 //!
 //! The per-cell terms are therefore tight (a 1000-element `list<int>` estimates
 //! ~1.2× its realized payload, not ~9×), and a wide fixed-width schema pays
@@ -567,7 +567,7 @@ impl<'a> Estimator<'a> {
     ///
     /// The empty-collection child arrays these builders always materialize
     /// (`ListArray<Utf8>` / `MapArray<Utf8, Utf8>`) are covered by
-    /// [`ARROW_COLUMN_SLACK_BYTES`], which is derived from exactly that case —
+    /// `ARROW_COLUMN_SLACK_BYTES`, which is derived from exactly that case —
     /// unlike the high-fidelity path, the flat builders have a FIXED one-level
     /// child shape, so a constant suffices.
     fn charge_flat(&mut self, dt: &DataType, fidelity: TextFidelity, value: Option<&'a Value>) {
--- a/cqlite-core/src/export/arrow_size.rs
+++ b/cqlite-core/src/export/arrow_size.rs
@@ -41,12 +41,19 @@
 //! | `ARROW_VALIDITY_BYTES` | every slot | the slot's validity **bit**; `n` bytes ≥ `ceil(n/8)` |
 //! | content | every slot | the slot's data-buffer bytes |
 //! | `ARROW_CELL_OVERHEAD_BYTES` | variable-width slots | its offsets entry + the buffer's trailing `n+1`-th entry |
-//! | `ARROW_COLUMN_SLACK_BYTES` | once per projected **column** | array nodes that correspond to no slot at all |
+//! | `ARROW_COLUMN_SLACK_BYTES` | once per **column whose builder materializes childless array nodes** | those nodes' empty offsets buffers |
 //!
 //! The per-cell terms are therefore tight (a 1000-element `list<int>` estimates
-//! ~1.2× its realized payload, not ~9×), and a wide fixed-width schema pays
-//! ~13 B/column/row rather than ~1 KB/row — so the row-cap, not the byte-cap,
-//! still binds on narrow shapes.
+//! ~1.2× its realized payload, not ~9×), and the residual is charged ONLY where
+//! a childless node can exist — the flat `list`/`set`/`map` builders, whose
+//! `MapArray`/`ListArray` always materialize a rendered `Utf8` child (see
+//! [`column_slack_bytes`]). A fixed-width column materializes exactly one array
+//! node with one data buffer, no offsets and no children, so it pays no residual
+//! at all: an `int` column costs `1 + 4 = 5` B/row against ~4.1 B/row realized.
+//! A 100-column `int` row therefore estimates 500 B and a full 8192-row batch
+//! 4,096,000 B — still inside the 4 MiB default, so the ROW-cap keeps binding on
+//! wide narrow-cell schemas up to 102 `int` columns (issue #2825 review C1;
+//! charging the residual per column put that cliff at 40 columns).
 //!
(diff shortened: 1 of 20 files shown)


pmcfadin:
🔎 **C intent audit (spec-auditor): PASS** — all 11 requirements `satisfied`, no scope drift, every changed file traces to a requirement.

Issue acceptance criterion 3 ("~1.0–1.1× throughput, no regression on narrow rows") is rated **partial, justified**: the *batching* half is proven hard (`batch_bytes_tests.rs:180` shows narrow boundaries byte-identical to a disabled cap; `arrow_size_tests.rs:800` keeps the row-cap binding to 102 int columns), but the 1.0–1.1× figure itself is only *printed*, not asserted, by the `#[ignore]`d perf test. That is exactly what spec R9 mandates (a wall-clock assert in the correctness path FAILs `roborev-lints`, #2642), so it does not block.

## Acted on: the wiring evidence was not executing

C found that `cqlite-flight/tests/issue_2825_max_batch_bytes_e2e.rs` — the end-to-end test that constitutes this knob's **wiring evidence** (spawns the real `cqlite-flight` binary, streams a real `do_get`, exercises both the CLI flag and the env var) — runs in **neither** the full agent-gate's standing components **nor** `flight-ci.yml`'s PR tier (`cargo test -p cqlite-flight --lib`). The full-crate tier is gated on a `ci:flight-full` label (`flight-ci.yml:83`) that this PR did not carry.

Wiring evidence that never executes is not evidence. **Applied the `ci:flight-full` label** (it was referenced by the workflow but did not exist in the repo — created it), so the full-package tier runs before merge.

## Deferred, not blocking

- `arrow_size_tests.rs:726-728` — comment says the two multi-column shapes measure "1.18 and 2.55" while both are pinned at `2`. Stale/imprecise comment; gate is green.
- `docs/flight-trino/JOURNAL.md:659-665` — states the `6 + 8 = 14 MiB` composition prospectively **without** the "~56 MiB count-bounded today" caveat that `batch_bytes.rs` and `design.md` both carry. R10 anchors on the module docs so it passes. Assigned to **#2821**, which makes that composition actually true when it lands and will update the JOURNAL as part of doing so — fixing it here and un-fixing it there would be churn. Recorded so the window is deliberate, not forgotten.

pmcfadin:
⛔ **Gate of record: FAIL — but PRE-EXISTING on `main`, not caused by this change.** Merge held.

`tooling-tests` FAILed; **29 of 30 components PASSed**. Triaged to `scripts/tests/test_agent_gate_summary.sh` block 9c — 3 failed / 90 passed at `:788`, `:804`, `:832`.

**Root cause:** those assertions are `$`-end-anchored against the `accelerators:` line. #2859 added a **Linux-only** ` mold=<state>` suffix (`agent-gate.sh:616-620`), so the emitted line ends `...sccache-health=ok mold=absent`. Content correct; the stale anchor no longer matches. Darwin emits no mold token, so it passes on macOS and fails on every Linux host.

**Decisive evidence it is not this PR:** a fresh detached `origin/main` worktree at `f8068a3` (this branch's own merge-base) reproduces the **identical** 3 failures, 90/93. `git diff origin/main...HEAD -- scripts/` is **empty** — this change touches no `scripts/` files. Determinism 4/4.

Also cleared: an earlier hypothesis that the operator perturbed the run (a GitHub label and issues #2896/#2897/#2908/#2910 were filed inside the 06:48 window). `test_agent_gate_summary.sh` shells out to no `gh`, reads no `refs/claims`/`refs/heartbeats`, touches no telemetry ledger, and has no wall-clock — it drives the gate through `--emit-summary-selftest` into a private tmp. Not reachable.

**Not merging over a red gate.** "gate PASS" is the definition of done, and "the one failure looks unrelated" is exactly the reasoning that should not be trusted on its own. Filed **#2914 (P1)** — it blocks the gate of record for every Linux worker in the fleet, not just this PR — and it is being fixed in its own issue/branch/PR (1:1:1:1 intact).

**Sequence from here:** land #2914 → rebase this PR onto it (`origin/main` has also moved to `0031fdf`, #2900) → re-run the full gate → arm merge.

Still valid and not being redone: C intent audit **PASS** 11/11, final roborev **clean of blockers**, flight-ci `30191277971` **SUCCESS** on `a4055dd` (4/4 e2e wiring tests executed, 0 ignored, 0 filtered). Those will be re-confirmed against the rebased head only where the rebase can affect them.

Gate evidence preserved at `/home/ubuntu/workspace/gate-evidence/` (run `/tmp/agent-gate.VKPkpP`).

pmcfadin:
## Gate of record — PASS (flow-closer endgame, #2825)

Full `scripts/agent-gate.sh` run on the certified head `d2b3096`, **30 components, 0 FAIL**.

```
==== AGENT-GATE SUMMARY ====
commit: d2b3096 branch: issue-2825-batch-byte-cap dirty: no
datasets: 144 Data.db files under test-data/datasets
tooling-tests:     PASS (197s)
RESULT: PASS
```

Summary file: `/tmp/gate-2825-FULL-r2.txt`.

### `CQLITE_ALLOW_FILE_GROWTH=1` justification (campsite rule / epic #1116)

The gate was run with `CQLITE_ALLOW_FILE_GROWTH=1`. This is justified, not a bypass — both files
were **already far over** the ~800-line source threshold *before* this change, and all new logic
in this change lives in **new files, each well under threshold**:

| File | Before | After | Threshold |
|---|---|---|---|
| `cqlite-flight/src/producer.rs` | 3,330 | 3,396 | 800 |
| `cqlite-flight/src/service.rs` | 2,036 | 2,068 | 800 |

Splitting these two pre-existing over-threshold files is out of scope for #2825 and belongs to
epic #1116 (source-file splits).

### An earlier gate FAIL on this branch was pre-existing, not caused by this change

A first full gate on the pre-rebase head `a4055dd` returned `RESULT: FAIL` with **29/30 PASS** and a
single `tooling-tests` FAIL. Triage attributed it to `scripts/tests/test_agent_gate_summary.sh`
block 9c, whose `$`-anchored greps on the `accelerators:` line were broken by #2859's Linux-only
` mold=<state>` suffix (Darwin emits no mold token, so it passed on macOS and failed on every Linux
host). That is **issue #2903, already fixed on `main` by PR #2905** (`83d57c4`).

This change touches no `scripts/` files. The branch was rebased onto current `origin/main`
(`a4055dd` -> `d2b3096`, clean, 12 commits) — the logical diff vs `origin/main` is unchanged at
**20 files, 5600 insertions, 5 deletions** — and `tooling-tests` now PASSes at 197s.

### Other endgame checks

- **C intent audit (spec-auditor)**: PASS — 11/11 requirements `satisfied` with public-surface test
  evidence. Verdict: https://github.com/pmcfadin/cqlite/pull/2906#issuecomment-5082380932
- **Final roborev confirmation pass**: clean of blockers. Three findings, all triaged as batched
  nits per `docs/development/roborev-severity.md` -> **#2917**. (Separately already filed: #2897,
  #2896; the `JOURNAL.md` 14 MiB caveat is assigned to #2821.)
- **flight-ci full-package tier** (run `30206870861`, `ci:flight-full`): SUCCESS on `d2b3096`.
  This is the wiring evidence for the knob — `tests/issue_2825_max_batch_bytes_e2e.rs` ran
  **4 passed, 0 failed, 0 ignored, 0 filtered out**, covering both the `--max-batch-bytes` CLI flag
  and the `CQLITE_MAX_BATCH_BYTES` env var end-to-end. Note this tier is **not** in the branch
  protection `required` set, so it was verified explicitly rather than delegated to `--auto`.
- **Pre-merge SHA assert**: `PREMERGE: OK d2b3096eee088239bb52023a626ca47322e74a95` (exit 0).

Arming `--auto --squash --delete-branch`.

PR #2904, opened 2026-07-26
test(#1883): per-row allocation-count ratchet on the row conversion hot path

pmcfadin:
Closes #1883. Epic #2817 (0.17 throughput program), manifest item M4.

Adds a Rust-side per-row **allocation-count ratchet** over the real public conversion surface
`build_row_from_scan_cached`, using the in-crate `test_alloc_probe` counting allocator. This closes a
gap #1449's binding-layer heap probes (V8 heap-delta / Python `tracemalloc`) physically cannot
observe: transient Rust-heap allocations freed before a row ever crosses the FFI boundary.

## Measured

Over 8 rows in one partition, through the public surface:

| fixture | cols | allocations | per row |
|---|---|--:|--:|
| narrow | 3 | **41** | 5 |
| wide | 32 | **273** | 34 |

≈ **1 allocation per cell** plus a small fixed per-row cost. The dominant per-row cost is the #1644
retention compaction (`Value::into_owned`'s TIER-1 copy of a small payload) — **not** hashing and
**not** key handling.

## Two differential controls, both verified RED-on-revert

Shaped after `lookup.rs`'s `cartesian_product_builds_each_combo_in_one_allocation`: each control keeps
a `reference()` of the pre-fix behaviour and asserts the current impl produces **identical output**
AND allocates **strictly fewer** times, so it cannot pass vacuously.

| property | revert | measured | result |
|---|---|---|---|
| per-cell `Arc<str>` interning (#1334) | fresh key string per cell | narrow 41 → **89**, wide 273 → **785** (exactly +2/cell) | test FAILS ✓ |
| sized value map (#1584) | drop the capacity hint | narrow 41 → **49** (+1/row rehash growth) | test FAILS ✓ |

Both return green on restore. Demonstrated by actually reverting each fix in the real implementation
— not committed here.

## Scope correction — this ratchet cannot gate #1447/#1445/#1446

#1883 was filed on the premise that reverting those three would turn this test red. They are all
**binding-layer** fixes, in a different crate:

| fix | where it actually lives |
|---|---|
| #1447 clone→move | `bindings/node/src/database.rs` — `ExecuteNativeTask::compute` |
| #1446 key interning | Node — JsString interned once per result |
| #1445 name re-clone | Python — `Row` in SELECT order |

Measured directly: reverting the clone→move *inside* `build_row_from_scan_cached` is **exactly
allocation-neutral** (41 vs 41 narrow, 273 vs 273 wide), because `Value::Text` is `Bytes`-backed
[...]

Issue #1883, opened 2026-07-04
Rust-side per-row alloc budget (dhat/allocator hook) to ratchet #1447 clone→move + #1445/#1446 interning — binding-layer heap tests can't observe them


pmcfadin:
## Context / why
#1449 added binding-layer per-row budget tests (Python tracemalloc, Node V8 heap-delta, Rust ctor-lookup counter). During that work we **empirically proved** the binding-layer heap/alloc tests **cannot observe** two of the W-fixes they were meant to pin:
- **#1447 (clone→move)**: the reverted clone is a transient Rust-heap allocation freed inside `executeNative()`/`execute()` *before* any post-execute V8 `heapUsed` / Python `tracemalloc` sample. Reverting #1447 moved the Node median only 1335.0 → 1333.8 (noise).
- **#1445/#1446 (key interning)**: emitting fresh key strings per cell also didn't move the number — V8 internally dedups property-name strings; Python side similarly unaffected.

What #1449's budgets DO pin (kept, honestly documented in-test): the **gross per-row value graph** (a value-graph-doubling regression goes RED: 2086 vs 1800 B/row budget) and the **Set/Map ctor-lookup count** (Rust `#[cfg(test)]` counter, ≤2/result — pins #1448).

## Goal
Add a **Rust-side per-row allocation budget** that CAN observe transient native allocations, so #1447 (move-not-clone) and #1445/#1446 (interning) become ratchets — a regression reverting them fails a test.

## Approach (design latitude — likely oracle-ish once shaped)
- Use **dhat** (heap profiler) or a custom counting global allocator behind a `#[cfg(test)]`/feature gate, measuring allocations across an `executeNative`/`execute` conversion of a wide result, asserting allocs/row under a measured baseline.
- Mirror the parser epic H **H2 dhat alloc-budget lane** pattern (report `docs/reports/parser-performance-audit-2026-07-01.md`) / `perf-gate.json` when/if that lands — coordinate, don't build a second harness.
- Acceptance: reverting #1447 (into_iter→iter().clone()) OR the key interning makes the Rust alloc-budget test go RED; restoring it green. Baseline documented + measured.

## Notes
- Filed from #1449 delivery (PR pinning the observable budgets). Binding-layer tests remain as the gross-value-graph + ctor ratchets; this issue closes the transient-Rust-alloc gap they physically can't cover.
- Not blocking #1449 — that PR honestly documents the coverage boundary.

_Filed by flow-lead worker during #1449._

[...]

PR #2905, opened 2026-07-26
fix(#2903): un-break the gate self-test on Linux — match sccache-health as a field, not end-of-line

pmcfadin:
Closes #2903. **Unblocks the full `agent-gate.sh` of record on every Linux box.**

## The bug
`scripts/agent-gate.sh:632` appends a mold token on Linux, so the accelerators line ends `... sccache-health=ok mold=absent`. Three assertions in `scripts/tests/test_agent_gate_summary.sh` were **end-anchored** on sccache-health (`sccache-health=ok$`), so the trailing token broke the `$` anchor and `tooling-tests` failed 3 of 93 — on Linux only (Darwin emits no mold token, which is why it went unnoticed). Introduced by `50cf721f` (#2859 mold bootstrap) without updating the anchors.

Consequence: every Linux worker's gate of record ended `RESULT: FAIL`, so delivery either stalls or someone starts merging over a red gate and rationalising which components "don't count" — the exact erosion the gate contract exists to prevent. The self-test is also the summary-format **drift detector**, so the drift detector was itself the drift. Found while gating PR #2892 (#2043); confirmed pre-existing on unmodified `origin/main` @ `bf8dfe14` via a detached baseline worktree (`90 passed / 3 failed`, same three names).

## The fix
`scripts/tests/test_agent_gate_summary.sh:788,808,839` — `sccache-health=ok$` → `sccache-health=ok( |$)` (same for `warn`/`na`), reusing the file's existing `assert_mold_token` field idiom. **Matches the field, not end-of-line**, so a future trailing accelerator token cannot break it again. `scripts/agent-gate.sh` is untouched — its emitted line is well-formed; the bug was purely the stale anchors. Diff is 3 regexes + 5 comment lines.

## Evidence the assertions still bite (this is the part that matters)
A loosened anchor is only a fix if it still fails on a real defect. Verified by mutation testing `agent-gate.sh`, each mutation reverted:
- Forcing `_sccache_health` to return `ok` unconditionally → **RED** on `expected sccache-health=warn`, `missing LOUD corruption WARN`, `expected sccache-health=na` (90/93).
- Suffixing the emitted value (`okx`/`warnx`/`nax`) → **all three RED** (83/93). So `ok( |$)` is an exact-value match, not a prefix match.
[...]

Issue #2903, opened 2026-07-26
tooling-tests is structurally RED on Linux: mold token breaks 3 end-anchored sccache-health asserts — the full gate of record cannot PASS on Linux


pmcfadin:
Discovered while gating PR #2892 (#2043). **The full `agent-gate.sh` currently cannot PASS on any Linux box**, because the gate's own self-test contradicts the gate's own output.

## Problem
`scripts/agent-gate.sh:632` prints the accelerators line with a trailing mold token on Linux:

```
printf 'accelerators: sccache=%s nextest=%s lanes=%s sccache-health=%s%s' ...
```

…so the emitted line ends `... sccache-health=ok mold=absent`.

But three assertions in `scripts/tests/test_agent_gate_summary.sh:788-830` are **end-anchored** on `sccache-health`:

```
^accelerators: .* sccache-health=ok$
^accelerators: .* sccache-health=warn$
^accelerators: .* sccache-health=na$
```

The trailing ` mold=absent` breaks the `$` anchor, so `tooling-tests` fails 3 of 93:
- `sccache-health: expected sccache-health=ok for on + 0 errors`
- `sccache-health: expected sccache-health=warn for on + >0 errors`
- `sccache-health: expected sccache-health=na when sccache not in use`

**Linux-only** — Darwin emits no mold token, so the anchors still match there, which is why this went unnoticed.

## Introduced by
Main commit `50cf721f` (#2859, mold bootstrap) added the mold token without updating the three anchors.

## Evidence it is pre-existing on main, not from any feature branch
`--only tooling-tests` on unmodified `origin/main` @ `bf8dfe14` → `passed: 90 failed: 3`, the same three assertion names (`/tmp/agent-gate.G5P5vm/tooling-tests.log`). Established with a detached baseline worktree; the PR under test touches zero files under `scripts/`.

## Why this matters more than 3 assertions
`agent-gate.sh`'s `AGENT-GATE SUMMARY` is the **only run that counts** before merge, and the gate is the definition of done. With `tooling-tests` structurally red on Linux, every Linux worker's gate of record ends `RESULT: FAIL`, so either delivery stalls or someone starts merging over a red gate and rationalising which components "don't count" — which is exactly the erosion the gate contract exists to prevent. The self-test is also the thing that is supposed to catch summary-format drift (`bash scripts/tests/test_agent_gate_summary.sh`), so right now the drift detector is itself the drift.

## Fix
[...]

PR #2915, opened 2026-07-26
fix(#2907): unanchor the sccache-health assertions so the gate can PASS on Linux

### LINKED ISSUE #2907: Gate of record cannot PASS on Linux: three sccache-health self-tests anchor on end-of-line, which the Linux-only mold= token breaks
Found while running the full `agent-gate.sh` for #1883 on a Linux worker. **Reproduced on clean `origin/main`** (detached worktree at `f8068a3f`), so it is not diff-specific.

## Symptom

```
tooling-tests:     FAIL (30s)
RESULT: FAIL
```

with, in `tooling-tests.log`:

```
FAIL - sccache-health: expected sccache-health=ok for on + 0 errors
FAIL - sccache-health: expected sccache-health=warn for on + >0 errors
accelerators: sccache=on nextest=on lanes=on sccache-health=warn mold=absent
FAIL - sccache-health: expected sccache-health=na when sccache not in use
accelerators: sccache=on nextest=on lanes=on sccache-health=na mold=absent
```

Note the tell: each FAIL is immediately followed by a line **containing exactly the token it says is
missing**. The value is right; the *match* is wrong.

## Root cause

`scripts/tests/test_agent_gate_summary.sh` lines **788**, **804**, **832** anchor the assertion at
end-of-line:

```bash
grep -qE '^accelerators: .* sccache-health=ok$'    # :788
grep -qE '^accelerators: .* sccache-health=warn$'  # :804
grep -qE '^accelerators: .* sccache-health=na$'    # :832
```

The `mold=` token was later appended to the `accelerators:` line and is **Linux-only** — the suite's
own `mold-darwin: no mold token (Darwin contract)` case confirms macOS emits no such token. So:

| platform | emitted line | `$`-anchored match |
|---|---|---|
| macOS | `… sccache-health=ok` | matches → PASS |
| Linux | `… sccache-health=ok mold=absent` | **no match → FAIL** |

Verified directly:

```console
$ echo 'accelerators: sccache=on nextest=on lanes=on sccache-health=ok' \
    | grep -qE '^accelerators: .* sccache-health=ok$' && echo MATCH
MATCH
$ echo 'accelerators: sccache=on nextest=on lanes=on sccache-health=ok mold=absent' \
    | grep -qE '^accelerators: .* sccache-health=ok$' && echo MATCH || echo 'NO MATCH'
NO MATCH
```

## Impact

**The full gate of record cannot reach `RESULT: PASS` on any Linux machine**, for any change,
including a no-op. Since `agent-gate.sh` is the only run that counts, this blocks the merge step of
the delivery pipeline for every Linux worker. It has stayed invisible because the fleet is macOS,
where the token is absent by contract.

Every other component passes on Linux — on the #1883 run: core-tests, memory-budget,
python-bindings, node-bindings, write-tests, cli-tests, oom-audit, smoke, minimal-build and the rest
were all PASS. `tooling-tests` was the sole FAIL.

## Fix

Stop anchoring at end-of-line; accept an optional trailing token. One-line change at each of the
three sites, e.g.:

```bash
grep -qE '^accelerators: .* sccache-health=ok( |$)'
```

(or assert against the whole accelerators line the way the `mold-*` cases already do — they pass on
Linux precisely because they do not anchor).

While there, consider a suite-level guard: a case asserting the `accelerators:` line parses
token-wise rather than by position, so the next appended token cannot silently red the gate on one
platform only.

## Notes

- Environment: Linux, `sccache 0.16.0` present and healthy (`sccache=on … sccache-health=ok` in the
  real SUMMARY), `mold=absent`.
- This is a self-test bug, not a product bug: the gate's *emitted* health values are correct in every
  case; only the assertions misread them.

pmcfadin:
Closes #2907. Unblocks PR #2904 (#1883), which is held behind this.

## The bug

`scripts/tests/test_agent_gate_summary.sh` matched the sccache health tri-state with an
**end-of-line anchor**:

```bash
grep -qE '^accelerators: .* sccache-health=(ok|warn|na)$'   # :788, :804, :832
```

But `accelerators:` is an ordered token list that later tokens get **appended** to, and 9d's `mold=`
token is **Linux-only** — Darwin emits none, as the suite's own `mold-darwin: no mold token (Darwin
contract)` case asserts. So:

| platform | emitted line | anchored match |
|---|---|---|
| macOS | `… sccache-health=ok` | matches → PASS |
| Linux | `… sccache-health=ok mold=absent` | **no match → FAIL** |

The gate emitted the **correct** health value in all three cases; only the assertions misread it.
The symptom was self-evidently a matching bug — each `FAIL` was immediately followed by a diagnostic
line containing the exact token it claimed was missing:

```
FAIL - sccache-health: expected sccache-health=ok for on + 0 errors
```

## Why it mattered

`tooling-tests` FAILed, which took `RESULT: FAIL` with it, **on every Linux box for any diff —
including a no-op on `main`** (reproduced on clean `origin/main` in a detached worktree). Since
`agent-gate.sh` is the only run that counts, the merge step of the delivery pipeline was unreachable
from Linux. It stayed invisible because the fleet is macOS, where the token is absent by contract.

Found while running the full gate for #1883; every other component on that run PASSed.

## The fix

Match the **token**, not the end of the line:

```bash
grep -qE '^accelerators: .* sccache-health=(ok|warn|na)( |$)'
```

Plus a note at 9c explaining why the anchor must not come back — this is a footgun that will look
like a harmless tightening to a future reader. The `mold-*` cases already avoid anchoring, which is
exactly why they pass on Linux.

## Verification

`bash scripts/tests/test_agent_gate_summary.sh`: **90 passed / 3 failed → 93 passed / 0 failed.**

Full `agent-gate.sh` running on Linux — expected to be the first `RESULT: PASS` reachable on this
platform.

## Note

Scope kept to the three assertions plus the comment. A broader hardening — asserting the
`accelerators:` line token-wise so the *next* appended token cannot red one platform only — is worth
doing but is deliberately not bundled here, since this PR is on the critical path for #2904.

### DIFF
--- a/scripts/tests/test_agent_gate_summary.sh
+++ b/scripts/tests/test_agent_gate_summary.sh
@@ -80,6 +80,26 @@ assert_accelerators() {
   fi
 }
 
+# assert_health_token <label> <file> <expected>: assert the accelerators line's
+# `sccache-health` token (issue #2641) as a WHITESPACE-DELIMITED TOKEN, independent
+# of where it sits in the line.
+#
+# Issue #2907: these assertions previously anchored `sccache-health=<v>$` at
+# end-of-line. That silently broke the moment #2859 appended the Linux-only `mold=`
+# token AFTER it — green on Darwin (no token by contract), red on every Linux box,
+# for ANY diff including a no-op on `main`, which made the gate of record
+# unreachable on Linux. Match the token, never the end of the line, so the next
+# appended token cannot resurrect this.
+assert_health_token() {
+  local label="$1" file="$2" expected="$3" line
+  line=$(grep -E '^accelerators: ' "$file" 2>/dev/null | head -1)
+  if printf '%s\n' "$line" | grep -qE " sccache-health=$expected(\$| )"; then
+    ok "$label -> sccache-health=$expected"
+  else
+    bad "$label: expected sccache-health=$expected, got: '$line'"
+  fi
+}
+
 # assert_mold_token <label> <file> <expected>: assert the accelerators line's mold
 # token (issue #2859). <expected> is a state (linked|present-unconfigured|absent)
 # or the literal "none" to require NO mold token (the Darwin contract).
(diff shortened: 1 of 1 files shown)


pmcfadin:
Closing unmerged — **superseded by #2905 (issue #2903)**, which landed the same fix on `main` as `83d57c4b` while this was in review.

The two fixes are functionally identical: the same `sccache-health=<v>( |$)` field-match replacing the end-of-line anchor on the same three assertions at `test_agent_gate_summary.sh:788/804/832`. Independent diagnosis, same root cause — the Linux-only `mold=` token (#2859) appended after `sccache-health=`, which made `tooling-tests` (and with it the gate of record) unreachable on Linux while staying green on the macOS fleet. Nothing here to salvage on the core fix.

The only delta this branch carried beyond main is the extra hardening in `40ff80e6`: an `assert_health_token` helper (position-independent token match) plus a per-case `assert_accelerators` whole-line check. I am deliberately **not** rebasing that in, because roborev's review of it flagged the helper as a near-verbatim copy of `assert_mold_token` — landing it would add the duplication that **#2918** then asks to remove. #2918 already specifies the better shape (one shared `assert_accel_token <label> <file> <key> <expected>`, with the `none` case preserved) and now applies cleanly on top of main's fix.

Verification recorded for whoever picks up #2918: with this branch's version, `scripts/tests/test_agent_gate_summary.sh` went 90 passed/3 failed → **96 passed/0 failed**, and the full `agent-gate.sh` returned **RESULT: PASS** on Linux (`tooling-tests: PASS (195s)`) — the first reachable gate of record on this platform.

Issue #2922, opened 2026-07-26
flow-closer merge path: two silent-failure traps (gh pr merge 'fatal:' after success; premerge-assert exit 3 on short SHA)


pmcfadin:
Two independent silent-failure hazards in the `flow-closer` merge path, both hit while merging PR #2906 (#2825) on 2026-07-26. Neither blocked the merge, but each can make a closer report the wrong outcome.

## Trap 1 — `gh pr merge` prints `fatal:` after a SUCCESSFUL merge, and the wrapper still exits 0

`gh pr merge --auto --squash --delete-branch` printed:

```
fatal: 'main' is already used by worktree at ...
```

That is `gh`'s **local post-merge branch switch** failing against the shared root checkout (which is on `main`, as it must be, while worktrees exist). The **API call had already landed** — the PR was merged.

Two things make this dangerous:

1. The output *reads* like a failed merge. A closer that greps stderr for `fatal:` would report failure on a successful merge, then potentially retry or escalate.
2. The wrapper still reported `ARM EXIT: 0`, so exit status and output disagree.

Observed on **both** PRs merged in that session, so it is systematic on any machine using worktrees — i.e. every flow worker.

**Remedy:** never infer merge outcome from `gh pr merge`'s output or exit code. Confirm with `gh pr view <n> --json state -q .state` (`MERGED`) as the authority. Document in `.claude/agents/flow-closer.md`; consider having the closer pass `--repo` / avoid the local branch switch, or explicitly tolerate this specific `fatal:` line.

## Trap 2 — `premerge-assert.sh` rejects a short SHA with exit 3, indistinguishable from a gh outage

`bash scripts/flow/premerge-assert.sh 2906 <short-sha>` exits **3** (`gh-failure`) when handed an abbreviated SHA. Passing the full 40-character SHA succeeds.

Exit 3 is the code for "gh call failed", so a caller cannot distinguish **"you passed a short SHA"** from **"GitHub is down"**. The natural closer reaction to a `gh-failure` is to wait and retry — which will never succeed, because the input is malformed, not transient.

**Remedy:** either normalize/expand an abbreviated SHA inside the script, or return a distinct exit code (e.g. 4 = `bad-argument`) with a message naming the problem. Add a one-line note to the script usage either way.

## Acceptance criteria

[...]

PR #2892, opened 2026-07-26
bench(#2043): measured reconcile overlap k-curve — M9 gen-overlap multiplier + L3 disposition

pmcfadin:
Closes #2043. Epic #2817 manifest item **M9** (the WS7 repoint). OpenSpec change `reconcile-overlap-multiplier` (Seam 1 owner-approved 2026-07-24).

## What this is
Measurement only — a k-parameterized Criterion instrument over the **public `KWayMerger` surface** plus the research record and the tightened program-doc terms. **No production-path change; no new `pub` on the reconcile path.**

- `cqlite-core/benches/reconcile_overlap.rs` + `benches/fixtures/multigen.rs` — k ∈ {1,2,5,10,20} × {disjoint, lww_overwrite, tombstone, ttl_expiring, field_blend}
- `now` pinned via `KWayMerger::with_now_secs` (never `CQLITE_TTL_NOW_OVERRIDE_SECS` — that seam is `#[cfg(debug_assertions)]` and compiles out of the release profile `cargo bench` uses)
- load average recorded at run start; results above the ceiling are **discarded, not merely flagged**
- `docs/research/issue-2043-reconcile-overlap-multiplier.md` (new) + `docs/architecture/throughput-program-2026-07.md` §3/§7

## Measured result (2 quiesced runs, load1m 0.28 / 0.84 vs a 2.00 ceiling, 16 cores, commit `df2dc857`, ±2.5% run-to-run at k≥2)
- **SSTable count is free.** Disjoint control is flat across k: 2876 / 2753 / 2791 ns/row at k=5/10/20.
- **Duplication is the whole cost:** p ≈ **1.53 µs/extra input row** vs q ≈ **1.26 µs/delivered row** ⇒ `D(o) = (q + p·o)/(q + p)`. D(1.25)=**1.14**, D(1.5)=**1.28**, D(3)=**2.12**.
- So §3's **1.1–1.5× gen-overlap term ⇔ o ∈ [1.2, 1.9]**; the floor is exactly **1.00×** for insert-once, and the ceiling is **optimistic** at an STCS p99.
- **TTL expiry is free** — `ttl_expiring` ≡ `lww_overwrite` to within 0.1%.
- Field `o` is labelled **assumption-not-measurement** at every consuming site (#2818/M0 replaces it).

## L3 (#2822) disposition — resolved conditionally
[...]

Issue #2677, opened 2026-07-18
claim.sh nit batch (#2665 roborev 1766): object-fetch infra edge, legacy-guard fail-open on outage, status/smoke coverage


pmcfadin:
Nit batch from #2665's final roborev pass (job 1766 — all Low, adjudicated non-blocking at merge time per the #2088 rubric; core lock "well-designed, heavily-tested"). All three are narrow infra-edge polish in `scripts/flow/claim.sh`:

1. **Object-fetch infra edge**: `fetch_claim` swallows fetch failures (`|| true`); if ls-remote succeeds but the object fetch transiently fails, pre-check/verify emit a conservative false `LOST`/`VERIFY-FAIL` (exit 2) instead of `ERROR infra` (exit 1). Fix: empty `git log -1` on a ref-confirmed SHA → `emit_infra`.
2. **Legacy-guard fail-open on outage**: `legacy_lock_blocks` returns "no block" on an ls-remote failure — during an outage in the transitional mixed-fleet window a claimant could win the ref past a live legacy branch lock. Fix: distinguish failure from empty; fail closed/retry.
3. **Coverage**: no test for `status` under unreachable origin (mirror TEST 9/12 shape) and none for `smoke` (intentionally origin-mutating — document why or add an opt-in smoke-of-smoke).

All conservative (no double-claim possible), retryable, contract-consistency polish. Route: oracle-driven, no OpenSpec.

pmcfadin:
Two additions from #2665's final review passes (both adjudicated non-blocking at merge — evidence-backed):
4. **Identity whitespace validation** (roborev 1767, Medium): space-containing `CLAIM_MACHINE`/`--actor` values corrupt the space-delimited claim-commit message parse → holder locked out of its OWN claim (`not-holder`). Fail-closed + requires a deliberately-malformed operator-set env + recoverable via `--force` → batched rather than a third full gate (shell fix is not delta-eligible). Fix: reject whitespace in machine/actor at entry (exit 3) + runbook note 'must be whitespace-free'.
5. **Legacy-guard misattribution** (roborev 1768, Low): a residual `issue-<N>-*` PR-plumbing branch (failed --delete-branch, reopened issue) blocks a fresh claim with `reason=legacy-branch-lock` — safe but confusing. Fix: scope guard to claim-format tips or emit `reason=residual-pr-branch`.
— flow-lead (merge-time triage per #2088)

pmcfadin:
[...]

Issue #2043, opened 2026-07-06
WS7 bench spike: measure the #2037 flip-risk constants (nb decode, k-way merge ns/row, cache-format candidates on real envelope segments)


pmcfadin:
Child of exploration epic #2037 (ArrowMemtable — coordinator-native OLAP). **WS7: the bench spike that grounds §11/§12 of the research record** (`docs/architecture/issue-2037-arrow-olap-research.md`).

## Why

§12.5 of the record names three constants where a 5× error flips a paper conclusion, and §12.1 lists ten missing measurements — none have committed numbers in the repo. These decide: (a) whether the per-generation cache's advantage holds, (b) S4A (Arrow IPC mmap) vs S4B (Parquet+zone-map) as the cache format, and (c) the owner's standing **CDC re-pose trigger** (2026-07-05 decision: no-CDC holds unless these numbers materially widen S6's edge — §12.7).

## Scope (deliverables)

1. **`nb` decode throughput** — Criterion bench: sequential Data.db scan on real datasets (`CQLITE_DATASETS_ROOT`), reporting MB/s **and** rows/s, buffered + mmap variants. Replaces `[Q-SCAN-ROWS-DERIVED-*]` (derived) with direct measurements.
2. **k-way LWW merge per-row cost** — ns/row at 2, 5, and 20 generations with realistic overlap + tombstone mix. This is the record's declared blind spot (10–500 ns/row unmeasured spread).
3. **Cache-format candidates on the real envelope shape** — convert one real `nb` generation to (a) Arrow IPC and (b) Parquet segments carrying the §6 envelope (per-column writetime/TTL i64 columns; range-tombstone sidecar batch); measure per-segment: conversion (build) rate, full-scan rate, and token-range-slice (projected-column) read latency. Vortex is an optional stretch, not required.
4. **Record update** — fill §12.1's missing-measurements table with measured values (file anchors); recompute the §12.4 totals for any constant >2× off the assumed value; write a one-paragraph verdict on (a) S4A-vs-S4B and (b) whether the CDC re-pose trigger fires.

## Acceptance criteria

- [ ] Benchmarks are committed, runnable via a documented one-line command, and require `CQLITE_DATASETS_ROOT` (fail loudly, never 0-row-pass, when datasets are absent).
- [ ] Results are recorded in the research record (§12.1 table updated) with machine specs and commit SHA; every replaced `[ASSUMED]`/derived entry cites the new measurement.
[...]

PR #2960, opened 2026-07-26
fix(#2945): sanctioned resume past the legacy-branch guard — empty-lease adopt --expect none + audited reason, and close an identity-forgery hole

### LINKED ISSUE #2945: claim.sh: legacy-branch-lock has no working escape hatch — an issue whose branch outlives its claim ref can never be re-claimed
Hit at the start of the #1883 delivery (2026-07-26), which had to proceed on a documented judgement call
because the sanctioned path could not be made to work. Complementary to #2677 item 2 (same function,
opposite direction: that one fails **open** on an outage; this one fails **closed** forever).

## Symptom

```console
$ bash scripts/flow/claim.sh claim 1883
CLAIM: LOST issue=1883 reason=legacy-branch-lock detail=refs/heads/issue-1883-* exists on origin
```

`refs/claims/issue-1883` was genuinely **free** (the prior claim had been released via CAS). The only
thing blocking was the *work branch* `issue-1883-rust-alloc-budget`, which had to be reused — it carried
the owner-approved OpenSpec change, so creating a fresh branch would have orphaned the approved spec.

## Root cause — the documented re-entrancy is unreachable

`legacy_lock_blocks` (`scripts/flow/claim.sh:238`) blocks when any `refs/heads/issue-<N>-*` exists whose
tip is not "ours", and its comment promises an escape hatch:

```sh
# legacy_lock_blocks <N> — 0 (blocks) iff a refs/heads/issue-<N>-* branch exists
# on origin whose tip is NOT our own claim. Re-entrancy: all-ours -> no block.
```

But "ours" is decided by `holder_is_us` (`:202`), which parses `machine=` / `actor=` fields **out of the
tip commit's message**:

```sh
msg="$(git log -1 --format=%B "$sha")"
[ "$(msg_field "$msg" machine)" = "$(this_machine)" ] && [ "$(msg_field "$msg" actor)" = "$actor" ]
```

Work branches are cut from `origin/main` and accumulate ordinary commits. **A work-branch tip never
carries those trailers** — only the synthetic claim commits pushed to `refs/claims/*` do. So
`holder_is_us` is false for every real work branch, from the moment it exists, and the "all-ours" branch
of the guard is effectively **dead code**. The lock is unconditional, not re-entrant.

## `adopt` cannot substitute

`adopt` is the documented adoption/resume path, but it is a compare-and-swap **against an existing ref**:

```sh
[ -n "$expect" ] || die_usage "adopt requires --expect <old-sha>"
git push --force-with-lease="refs/claims/issue-${issue}:${expect}" …
```

In this scenario there is no ref to CAS from. Git *does* express "must not exist" as an **empty** expected
value, but `cmd_adopt` rejects an empty `--expect`, and the all-zeros OID does not satisfy the lease:

```console
$ claim.sh adopt 1883 --expect 0000000000000000000000000000000000000000
CLAIM: ADOPT-LOST issue=1883 expected=000… actual=<gone> holder=unknown
```

So: `claim` refuses, `adopt` cannot express the case, and there is no third option.

## When this bites

Any issue whose branch outlives its claim ref:

- a **resumed** issue (claim released or reaped, branch deliberately kept — the #1883 case);
- a **parked** issue under the #2666 park-and-resume flow;
- an issue whose claim was **reaped** by the #2499/#2655 CI reaper while the branch remained;
- **any merged issue whose remote branch was not deleted** — note `gh pr merge --delete-branch` has been
  observed to report success without deleting (#2922, trap 3). That combination silently makes a *shipped*
  issue unclaimable if it is ever reopened or followed up on the same number.

## Why it was safe to proceed on #1883, and why that is not a fix

The guard is symmetric — it blocks every machine identically — so the branch was itself acting as the
mutex and double-work was still prevented. That reasoning was recorded on the issue, corroborated (no
claim ref, no open PR, no `refs/machine-claims/*`, branch idle 3 days, board still `Ready`, while every
genuinely-live issue did hold a claim ref), and the board markers were set manually.

But it required a human-grade judgement call to bypass a mechanical lock, which is exactly what the claim
protocol exists to remove. The next worker either repeats the analysis or does something less careful.

## Acceptance criteria

1. A worker that legitimately needs to resume an issue with a pre-existing `issue-<N>-*` branch and a
   FREE claim ref can acquire the claim through a **sanctioned, non-`--force` command**, and the command
   records who took it and why.
2. Whatever that path is, it must remain safe against the case the guard exists for: it must NOT let a
   second machine take an issue that another machine is actively working. Suggested shapes (pick one):
   - allow `adopt --expect none` (or `--expect ''`) to mean "the ref must not exist", mapping to git's
     empty-lease semantics — atomic, server-arbitrated, no force; or
   - make the re-entrancy check real by comparing against something a work branch actually carries (the
     branch's presence in this machine's `git worktree list`, or a per-machine marker ref), so
     "my own branch" stops being dead code; or
   - keep the block but emit the exact remediation command instead of a bare `LOST`.
3. Regression coverage in `scripts/flow/tests/` for: free ref + foreign-tip branch present → resume path
   succeeds; **held** ref + branch present → still refused; two machines racing the resume path → exactly
   one wins.
4. The dead "Re-entrancy: all-ours -> no block" comment is corrected or made true — it currently
   documents behaviour that cannot occur.

## Notes

- Route: oracle-driven (script + tests), no OpenSpec — same as #2677, which touches the same function and
  should probably be fixed in the same pass.
- Unrelated to `claim.sh`'s core ref lock, which behaved correctly throughout (`claim`, `release --force`,
  and a later `claim 2907` all worked as designed once a git credential helper was configured — see #2942).

pmcfadin:
Closes #2945. Oracle-routed (script + tests), no OpenSpec.

## The bug
`claim.sh claim <N>` refused with `reason=legacy-branch-lock` whenever any `refs/heads/issue-<N>-*` branch existed on origin — **even when `refs/claims/issue-<N>` was free**. The documented escape hatch ("Re-entrancy: all-ours -> no block") was **dead code**: it decided "ours" via `holder_is_us`, which parses `machine=`/`actor=` trailers out of the *branch tip's* commit message, and a real work branch cut from `origin/main` never carries those trailers. `adopt` couldn't substitute — it is a CAS against an *existing* ref. So `claim` refused, `adopt` had no expression for the case, and there was no third option.

**Reproduced independently twice on 2026-07-26** while resuming #2043 and #1883, both of which carried an owner-approved OpenSpec change on the branch (so a fresh branch would have orphaned the approved spec). Both times the worker bypassed the guard by hand-crafting a claim commit in the script's exact format and pushing straight to `refs/claims/issue-<N>` — correct semantics, but exactly the human-grade bypass the protocol exists to eliminate.

## The fix
`adopt --expect none` → git **empty-lease** semantics ("the ref must not exist"), so **git still arbitrates server-side** and the single-arbiter property the whole design rests on is preserved. A machine-local `git worktree list` check was considered and rejected: it looks equivalent but cannot arbitrate a cross-machine race. `--reason` is **mandatory** on that path, recorded in the claim commit and rendered by `status` (AC1's "records who took it and why"). The refusal message now names the blocking branch, states `claim-ref=free`, and prints the exact remediation command — a bare `LOST` is what sent two workers into manual bypasses (AC option 3, taken as well).

The dead re-entrancy hatch is **removed**, not patched: tip-based `holder_is_us` re-entrancy is gone, replaced by `legacy_branch_scan`, and a test pins that even a claim-shaped branch tip cannot grant a claim — so the hatch can't be silently revived (AC4).

### Two corrections to the issue's own analysis
- The issue reports `adopt --expect 0000…0` fails the lease. Against the live origin it **creates the ref** — a silent, unrecorded create, arguably worse than the bug being fixed. All-zero leases now route through the same audited `--reason` path, while `--expect ''` stays exit 64 so an unset shell variable can never turn a CAS into a create.
- The "merged but branch not deleted" case in *When this bites* is **live, not hypothetical**: `gh pr merge --delete-branch` reported success on PR #2892 the same day without deleting the branch (cf. #2922 trap 3).

## Review findings fixed (6 blockers, two review passes)
- **`cmd_adopt` had no re-entrancy check** → a retry after the transient `ERROR reason=infra` (whose documented remedy *is* retry) reported `ADOPT-LOST` **against ourselves**, exit 2. Workers read exit 2 as "you didn't win, move on", so a worker that hit a network blip mid-resume would abandon an issue it owns **while the claim ref stayed held** — nobody could take it. Now matches `cmd_claim`: `ADOPTED … (re-entrant)` exit 0.
- **`sanitize_reason` killed the script on macOS**: BSD `tr -c` aborts with `Illegal byte sequence` on non-ASCII in a UTF-8 locale, and under `set -euo pipefail` that emitted **no `CLAIM:` line at all** with exit 1 — which the contract defines as "retryable", so a caller would retry forever on input that can never succeed. `--reason "resume — parked claim"` was enough to trigger it. Fixed with `LC_ALL=C` across the pipeline.
- **Identity forgery in the lock (defeats it outright).** `msg_field` was a greedy `.*key=` (last match wins) and `--actor` was unsanitized, so `claim N --actor 'flow machine=machineA'` from machineB made machineA's `claim N` print `HELD … (re-entrant)` **exit 0** on a ref machineB held — two writers on one issue. Root fix: **exact-key, first-match, per-token** `msg_field` (token *equality* also stops `holder-machine=` answering for `machine`), plus `--actor` sanitized at the arg boundary so the record and the comparison use one value. In scope because this change adds a second user-controlled string to the very commit message that parser reads.
- The brand-new `ERROR reason=infra` path (empty-lease push rejected while the ref is absent) had **zero coverage**; now pinned.
- Doctrine: `website/.../delivery-pipeline.md` still said eligibility requires "no legacy `issue-<N>-*` branch" with no escape hatch — the exact prose that caused the bypasses. Updated in this change, per CLAUDE.md's same-change rule.
- Minimal **#2677 item 2** fold, disclosed: the new refusal asserts "claim-ref=free, a branch blocks you", which must not be printed off an unread guard, so an `ls-remote --heads` failure is now `ERROR reason=infra` exit 1 instead of a false all-clear. #2677 items 1 and 3 untouched — that issue's scope is the owner's.

## An honest negative on the race test
The test-quality review **mutation-tested the suite** and found the original race test vacuous: swapping the empty lease for `git push --force` left it green. Investigating that produced a finding worth recording rather than papering over — **no race shape can catch `push --force`**, because when both claimants advertise "absent" they each send an all-zero old value and git's *protocol-level stale-info check*, not the lease, rejects the second push. The witness was therefore moved to the **fresh-advertisement** shape, where the mutant dies deterministically: it now fails **7 assertions**. A `pre-receive` stagger that passed under the mutant was **deleted rather than kept as theatre**, and both roles are labelled in the test header so they can't be collapsed.

## Coverage (AC3)
New hermetic suite `scripts/flow/tests/claim-resume.test.sh` — **50/50**, wired into the gate's `tooling-tests` and `ci.yml`'s `flow-tooling-tests` (macOS included). Against pre-fix `claim.sh` the suite is **12 pass / 38 fail**. Covers: free ref + foreign-tip branch → resume succeeds; **held** ref + branch → still refused with the foreign ref byte-unchanged; two machines racing → exactly one accepted create with `old` all-zeros (post-receive witness), 5 rounds; reaper `release --force` vs resumer `adopt` barrier race; adversarial forged `--reason`/`--actor`; non-ASCII reason; holder release round trip; and arg-surface pins (`--expect <wrong-hex>` on a free ref must not create; non-numeric → 64; empty reason → `reason=unspecified`; 120-char truncation; newline in reason).

`--only tooling-tests` PASS. The ONE full gate of record runs pre-merge in `flow-closer` and its verbatim summary is posted below.

### DIFF
--- a/.claude/skills/flow-activate/SKILL.md
+++ b/.claude/skills/flow-activate/SKILL.md
@@ -37,6 +37,8 @@ committed, owner-approvable OpenSpec change on an isolated worktree. **STOP at a
    if ! bash scripts/flow/claim.sh claim <N>; then
      echo "CLAIM LOST — another session holds refs/claims/issue-<N>. Take the next item."
      # (Adopting a reaped claim instead? Use: bash scripts/flow/claim.sh adopt <N> --expect <old-sha>)
+     # (Refused with reason=legacy-branch-lock — a FREE claim ref but an issue-<N>-* branch
+     #  still on origin? Resume it: claim.sh adopt <N> --expect none --reason <why>, #2945)
    fi
    # CLAIM HELD → set up the worktree + branch. The branch is naming/PR plumbing, NOT the lock:
    wt=".claude/worktrees/issue-<N>-<slug>"
--- a/.claude/skills/flow-activate/SKILL.md
+++ b/.claude/skills/flow-activate/SKILL.md
@@ -38,7 +38,9 @@ committed, owner-approvable OpenSpec change on an isolated worktree. **STOP at a
      echo "CLAIM LOST — another session holds refs/claims/issue-<N>. Take the next item."
      # (Adopting a reaped claim instead? Use: bash scripts/flow/claim.sh adopt <N> --expect <old-sha>)
      # (Refused with reason=legacy-branch-lock — a FREE claim ref but an issue-<N>-* branch
-     #  still on origin? Resume it: claim.sh adopt <N> --expect none --reason <why>, #2945)
+     #  still on origin? Resume it: claim.sh adopt <N> --expect none --reason <why>, #2945 —
+     #  but ONLY when the refusal itself prints that command at open-prs=0; a
+     #  `remediation=withheld open-prs=<n>` line means an endgame may be live: confirm first)
    fi
    # CLAIM HELD → set up the worktree + branch. The branch is naming/PR plumbing, NOT the lock:
    wt=".claude/worktrees/issue-<N>-<slug>"
--- a/.claude/skills/flow-activate/SKILL.md
+++ b/.claude/skills/flow-activate/SKILL.md
@@ -37,14 +37,13 @@ committed, owner-approvable OpenSpec change on an isolated worktree. **STOP at a
    if ! bash scripts/flow/claim.sh claim <N>; then
      echo "CLAIM LOST — another session holds refs/claims/issue-<N>. Take the next item."
      # (Adopting a reaped claim instead? Use: bash scripts/flow/claim.sh adopt <N> --expect <old-sha>)
-     # (Refused with reason=legacy-branch-lock — a FREE claim ref but an issue-<N>-* branch
-     #  still on origin? Resume it: claim.sh adopt <N> --expect none --reason <concrete why>,
-     #  #2945 — but ONLY by running the command the refusal itself PRINTS (reason pre-filled;
-     #  it is printed only at open-prs=0 + all branch tips carrying their OWN commits and
-     #  staler than 4h + no fresh machine-claim/heartbeat ref; the branch this skill pushes
-     #  below has NO commits of its own, so its lane is never advertised as stale).
-     #  A `remediation=withheld <signals>` line means the lane
-     #  may be LIVE or is unproven: confirm ownership first, never resume blind)
+     # (Refused with reason=legacy-branch-lock ... claim-ref=free — a FREE claim ref but an
+     #  issue-<N>-* branch still on origin? The sanctioned resume is
+     #  `claim.sh adopt <N> --expect none --reason <concrete why>` (#2945). The refusal
+     #  deliberately does NOT print it: CONFIRM the lane is abandoned first —
+     #  `claim-heartbeat.sh should-reap <machine>` (age > 4h AND no open PR AND pid-dead if
+     #  local), board Status, branch/PR author. Never resume blind, never hand-craft a claim
+     #  commit)
    fi
    # CLAIM HELD → set up the worktree + branch. The branch is naming/PR plumbing, NOT the lock:
    wt=".claude/worktrees/issue-<N>-<slug>"
--- a/.claude/skills/flow-activate/SKILL.md
+++ b/.claude/skills/flow-activate/SKILL.md
@@ -39,7 +39,9 @@ committed, owner-approvable OpenSpec change on an isolated worktree. **STOP at a
      # (Adopting a reaped claim instead? Use: bash scripts/flow/claim.sh adopt <N> --expect <old-sha>)
      # (Refused with reason=legacy-branch-lock ... claim-ref=free — a FREE claim ref but an
      #  issue-<N>-* branch still on origin? The sanctioned resume is
-     #  `claim.sh adopt <N> --expect none --reason <concrete why>` (#2945). The refusal
+     #  `claim.sh adopt <N> --expect none --reason resume-legacy-branch-lock:branch-outlived-claim`
+     #  (#2945 — a placeholder reason, or one still carrying an unsubstituted <…>, is
+     #  rejected with exit 64, so substitute a concrete why). The refusal
      #  deliberately does NOT print it: CONFIRM the lane is abandoned first —
      #  `claim-heartbeat.sh should-reap <machine>` (age > 4h AND no open PR AND pid-dead if
      #  local), board Status, branch/PR author. Never resume blind, never hand-craft a claim
--- a/.claude/skills/flow-activate/SKILL.md
+++ b/.claude/skills/flow-activate/SKILL.md
@@ -40,8 +40,10 @@ committed, owner-approvable OpenSpec change on an isolated worktree. **STOP at a
      # (Refused with reason=legacy-branch-lock — a FREE claim ref but an issue-<N>-* branch
      #  still on origin? Resume it: claim.sh adopt <N> --expect none --reason <concrete why>,
      #  #2945 — but ONLY by running the command the refusal itself PRINTS (reason pre-filled;
-     #  it is printed only at open-prs=0 + all branch tips staler than 4h + no fresh
-     #  machine-claim/heartbeat ref). A `remediation=withheld <signals>` line means the lane
+     #  it is printed only at open-prs=0 + all branch tips carrying their OWN commits and
+     #  staler than 4h + no fresh machine-claim/heartbeat ref; the branch this skill pushes
+     #  below has NO commits of its own, so its lane is never advertised as stale).
+     #  A `remediation=withheld <signals>` line means the lane
      #  may be LIVE or is unproven: confirm ownership first, never resume blind)
    fi
    # CLAIM HELD → set up the worktree + branch. The branch is naming/PR plumbing, NOT the lock:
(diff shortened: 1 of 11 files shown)


pmcfadin:
## Round 3 — roborev re-review findings fixed (`a3cdce25`)

roborev's re-review returned 1 Medium + 4 Lows. I treated **four** as blockers, because their labels understate them — three break stated acceptance criteria and one is a correctness hole in the lock:

- **F1 (AC1) — the `--reason` audit gate was defeated by whitespace.** It validated the *raw* argument but recorded the *sanitized* one, so `'   '`, `'---'`, `'…'` or an unset `"$VAR "` passed and recorded `reason=unspecified` — indistinguishable from giving no reason, making AC1's "and why" optional in practice. Now sanitizes first, then requires a recordable token (≥3 chars, not the sentinel): those inputs are exit 64; `'wip'` still records.
- **F2 (correctness) — a violated compare-and-swap was reported as success.** The re-entrancy shortcut added in round 2 applied in *both* modes, so a CAS whose lease failed because the ref sits at `Y` ≠ the expected `X` — where `Y` happens to be ours — printed `ADOPTED … (re-entrant)` exit 0 while displaying `from=X`. Empty-lease keeps the plain verdict; CAS now gets its own `(re-entrant, lease-mismatch expected=X actual=Y)` naming both shas. Kept at exit 0 deliberately: exit 2 would make a worker abandon a ref it actually holds, which is the failure mode F-B1 fixed.
- **F3 (AC2) — the refusal advertised a command that can hand away an actively-worked issue.** The remediation was printed unconditionally, and its safety argument doesn't cover the case the guard exists for: an older-fleet worker that locks with the *branch* and holds no claim ref. For that worker `claim-ref=free` is true, so the advertised `adopt --expect none` succeeds and a second machine takes live work — which AC2 explicitly forbids. This one deserves emphasis because the message's primary consumers are agents that follow printed remediations literally; **I am one of those agents, and I followed exactly such a remediation twice today.** The command is now printed only at `open-prs=0`; an open PR or an unreadable list (`-1`) prints `remediation=withheld open-prs=<n>` — fail closed. A deliberate resume still works; only the *advice* changed, never the arbiter.
- **F4 (AC3) — the race test still wasn't racing.** Second time this has been caught. The barrier used two *independent* FIFOs whose writers were launched sequentially, so machine A could finish its entire adopt before B started, and every assertion held under full serialization. Now a single shared start flag both racers block on. Because a vacuous race test slipped through twice, it no longer merely *claims* to race: each racer records ns timestamps around its push and the round **FAILS unless the two windows overlap**. Measured overlap **5/5 rounds, 37–38 ms on a ~40 ms adopt**; no hi-res clock fails loudly.
- **F5 — a caller bug read as a race loss.** `--expect` was hex-checked but not length-checked, so `--expect 0`/`00` silently became a *create* instead of a CAS, and a truncated sha (say from a bad `cut`) produced an unresolvable lease → push fails → `ADOPT-LOST`, i.e. a race-loss verdict for a usage error, in a file otherwise rigorous about separating usage / infra / lost-race. Now requires a full object name (40/64 hex); `0`, `00`, `abc123`, 39-zeros → exit 64, nothing created.

**`push --force` mutation still killed** — 51 pass / 8 fail, led by TEST 2's fresh-advertisement witness (2 server-applied updates, holder evicted) and TEST 10.

**Suite: 59/59** (was 50/50, +9 assertions). Pre-fix baseline against `origin/main`'s `claim.sh`: **16 pass / 43 fail** (the earlier 12/38 was the smaller file). `test_claim_lock.sh` 23/23, `finalize-cleanup.test.sh` 53/53, `--only tooling-tests` PASS.

Doctrine updated in the same change per CLAUDE.md: `CLAUDE.md`, `website/.../delivery-pipeline.md`, `docs/development/agent-machine-setup.md`, and the flow-implement / flow-activate skills.

**Disclosed, not hidden:** `claim.sh` is now 841 lines, over the ~800 source target. The gate's file-size ratchet only covers `.rs`, so nothing fails — but the campsite rule is a real standard here, so a responsibility split is filed as a follow-up rather than quietly ignored.

Handing the endgame to `flow-closer`: the ONE full `agent-gate.sh` of record → final roborev → merge-on-green. Oracle-routed, so no C/spec-auditor step.

pmcfadin:
@/tmp/pr2960-summary-comment.md

Issue #2965, opened 2026-07-26
bootstrap preflight: environment-variant gaps in the git/board checks (GHE multi-host, http:// origins, fine-grained tokens) + share the auth classifier


pmcfadin:
Batched follow-up from #2942's review rounds (roborev jobs 8 and 10, plus an independent shell review), 2026-07-26. All Low, all adjudicated non-blocking at merge time per the #2088 rubric: each is an **environment variant** rather than a defect on the fleet's actual configuration. #2942 fixed everything that could break a real machine or emit a false green on a supported setup.

## 1. `gh_active_block` selects the wrong host's stanza on a multi-host login

`scripts/bootstrap-agent-machine.sh` — `gh auth status` prints a section per **host**, and each host section has its own `Active account: true`. The awk keeps the *last* stanza carrying that marker, so on a machine also logged into a GHE/enterprise host listed after `github.com`, the scopes, missing-scopes line and reported account name come from the wrong host.

This is the same misattribution shape the function was written to fix (#2942 blocker: the verdict must be attributed to the account it actually measured), just across hosts instead of accounts.

**Fix:** track the host per stanza (`/^[^ ]/ { host=$1 }`, or parse the `Logged in to <host>` line) and select the active stanza whose host matches the board host (`github.com` / `$GH_HOST`).

## 2. `http://` origins probe and configure an `https` credential scope

The origin classifier maps `http://*` to `GIT_ORIGIN_KIND=https`, but the probe input (`printf 'protocol=https\nhost=%s\n\n'`) and the written key (`credential.https://$host.helper`) both hardcode `https`. Git's credential matching requires an exact protocol match, so for an `http://` origin the bootstrap probes and configures a scope git will never consult — reporting "credentials resolve" / "configured" while every push still fails with `could not read Username`.

**Fix:** capture the scheme alongside the host and use it in both the `protocol=` line and the config key — or classify `http://` as unsupported and say so plainly.

## 3. Fine-grained PATs and GitHub App tokens can never reach the green verdict

The unqualified `ok` requires `scope_prefilter=1`, i.e. the classic `project` scope string. A fine-grained PAT or App token reports `Token scopes: none`, so a machine whose board reads **and** writes both work will always print `'project' scope MISSING`.

[...]

Issue #2961, opened 2026-07-26
claim.sh: ls-remote read paths still report a permanent credential fault as a retryable transient


pmcfadin:
Found by roborev during #2942 review-first (job on `9cbf41cf`), 2026-07-26. Deliberately out of #2942's scope — that change's approved spec (requirement 3 and both its scenarios) is explicitly **push**-scoped — so filed rather than folded in.

## The gap

#2942 makes an unauthenticated `git push` report `reason=auth … (NOT retryable)` instead of `reason=infra … (transient — retry)`, because telling a worker to retry a permanently broken operation is worse than a bare error.

The **read** paths were not covered. `remote_claim_lookup` (`scripts/flow/claim.sh:216-227`) and every `git ls-remote` call still swallow stderr with `2>/dev/null` and map any failure to `emit_infra … (transient — retry)`. So on a private/GHE remote, an expired token, or an EMU account without repo read:

- `claim.sh verify <N>`, `status <N>`, `list`
- `cmd_claim`'s `push-failed-and-ls-remote-unreachable` branch

all still advise retrying a credential fault that can never self-clear — the exact loop #2942 removed for push.

The code already *knows* this. The comment at `claim.sh:341-344` asserts that on a private repo both the push and the read fail and that "the cause is the same permanent, non-retryable credential fault" — but nothing classifies the read, so the emitted verdict contradicts the comment directly above it.

## Scope

1. Capture `ls-remote` stderr (`out="$(git ls-remote … 2>&1 >/dev/null)"` or a tee'd variant) instead of discarding it.
2. Route it through the existing `git_stderr_is_auth` classifier → `emit_auth` before falling back to `emit_infra`.
3. Never echo the captured stderr — a remote URL can embed a token. #2942's push path classifies without emitting; match that exactly.
4. Coverage in `scripts/tests/test_claim_lock.sh` mirroring TEST 20's shim, for `verify` / `status` / `list` and the `cmd_claim` read-fallback branch: auth signature → `reason=auth`; anything unrecognized → still the retryable infra verdict (the #2665 contract must only ever be narrowed by positively-identified signatures).

## Acceptance

- An unauthenticated `claim.sh verify` reports a non-retryable auth verdict naming the remediation.
- A genuine transient (reachable remote, network blip) still reports the retryable infra verdict.
- No raw git stderr reaches the log on any path.
- The `:341-344` comment and the emitted verdicts agree.

## Related

[...]

Issue #2957, opened 2026-07-26
claim-heartbeat.sh: a failed liveness push reports SUCCESS, and can prompt-block on a credential-less worker


pmcfadin:
Found by both reviewers during #2942 review-first (roborev job 8 + an independent shell review), 2026-07-26. Out of #2942's approved spec scope, so filed rather than folded in.

## Bug 1 — a failed push is reported as success

`scripts/flow/claim-heartbeat.sh` `push_liveness_ref` runs:

```sh
git push "$REMOTE" --force "${commit_sha}:${ref}"     # :197
printf '%s\n' "$commit_sha"
```

The function's return status is **`printf`'s**, not the push's. A push that fails — unauthenticated, offline, rejected — yields a success return and a printed SHA, so the caller records a heartbeat that never landed on origin.

**Why this has teeth.** The deterministic reaper (`claim-heartbeat.sh should-reap`, #2089/#2499) keys on heartbeat *age*: reap when age > 4h AND no open PR. A silently-failing heartbeat means the ref never advances while the worker believes it is alive. The worker keeps working; the reaper concludes it is dead and can free the claim under it — the exact split-brain the liveness mechanism exists to prevent.

This is now **more** likely, not less: #2942 makes `claim.sh` fail loudly on an unauthenticated push (`reason=auth`, non-retryable). So on a credential-less box the claim fails visibly while the heartbeat keeps "succeeding" — the two disagree, and the quiet one is the one the reaper trusts.

## Bug 2 — the pushes can prompt and block

Neither push site (`:197`, `:305`) exports `GIT_TERMINAL_PROMPT=0`. On a **tty-attached** worker with no credential helper, git prompts for a username and blocks indefinitely instead of failing. #2942 adds the one-line `export GIT_TERMINAL_PROMPT=0` to close the hang, but does not address the status propagation or add auth classification.

## Scope

1. `push_liveness_ref` (and the delete path at `:305`) propagate the push's exit status.
2. Reuse `claim.sh`'s `git_stderr_is_auth` classifier so an unauthenticated heartbeat reports `reason=auth` rather than a bare failure — the same misattribution #2942 fixed for `claim.sh`. Factor the classifier into a shared helper rather than copying it.
3. Never echo raw push stderr (a remote URL can embed a token) — `claim.sh` classifies without emitting; match that.
4. Coverage in `scripts/flow/tests/` for: push fails → non-zero return + `reason=auth`; push succeeds → ref actually advanced on origin.

[...]

Issue #2665, opened 2026-07-17
Atomic issue-scoped claim lock: slugless claim ref + scripted claim/adopt (kills different-slug + identical-SHA + adoption races)


pmcfadin:
Epic #2664. Audit FM2. The claim lock arbitrates a ref NAME, not an issue: model-derived slugs mean two sessions claim `issue-N-<different-slug>` and BOTH pushes succeed (field: #1632 slug pair); two sessions branching the same `origin/main` tip push an identical SHA and the loser sees an "up-to-date" success. `worker.md`'s claim recipe (the one the unattended fleet runs) is missing the unique-empty-claim-commit and post-push re-read guards that flow-activate/flow-implement carry, and every re-read verifies only the session's OWN slug, never the `issue-N-*` glob. The only automated >1-lock detector is finalize-cleanup guard 2 — at the END of the lifecycle.
Acceptance:
- Claim = atomic creation of a single fixed-name ref per issue (`refs/claims/issue-<N>` or slugless branch `issue-<N>`), so git server-side arbitration decides every race; slug survives only as worktree/PR naming, never as the lock.
- A `scripts/flow/claim.sh` (create/verify/release/adopt) replaces the prose claim blocks in worker.md, flow-activate, flow-implement; skills call the script.
- Adoption/resume of a reaped claim replaces the ref via compare-and-swap (force-with-lease semantics) carrying adopter id, so a resurrected original worker loses fast-forward and detects loss immediately (fixes the #2467/#2499 two-writer race).
- finalize-cleanup's multi-lock guard also runs at CLAIM time.
- Doctrine (CLAUDE.md + delivery-pipeline page) updated in the same change; fleet-runbook's "collisions impossible by construction" overclaim corrected.
Relates: #2655 (machine liveness refs — complementary; this is issue-claim arbitration), #2499, #2456.

pmcfadin:
🔒 Claimed by flow-lead (Patricks-MacBook-Pro) — wave 1 of epic #2664 Phase-1 hardening. Branch: issue-2665-* pushed to origin.

PR #2967, opened 2026-07-26
fix(#2942): worker environment preflight — verify the operation, not the configuration

### LINKED ISSUE #2942: flow-meta: worker environment deltas silently break the pipeline on non-macOS boxes (push auth, force-with-lease form, board writes)
Groomed from the #1883 delivery (2026-07-26), run on a **Linux** worker. Three distinct pipeline steps
failed in ways the doctrine does not describe, each costing a round-trip to diagnose. None is covered by
an existing issue (checked: #2908, #2922, #2748, #2758, #2677 all cover other things).

The common shape: **CLAUDE.md and the flow-* skills document commands that work on the macOS fleet and
fail on a Linux box**, so every non-macOS worker rediscovers the same three deltas.

## Delta 1 — `git push` has no credentials; the documented command just fails

```console
$ git push -u origin issue-1883-rust-alloc-budget
fatal: could not read Username for 'https://github.com': No such device or address
```

`gh` itself is authenticated (`GH_TOKEN`, scopes `project,repo,workflow`), but git has no credential
helper wired, so **every raw `git push` in the pipeline fails** — pushing a work branch, and the
telemetry-stamp push in `flow-finalize`. Working form:

```bash
git -c credential.helper='!f(){ echo "username=x-access-token"; echo "password=$GH_TOKEN"; };f' push …
```

Worst failure mode observed: a chained `push && gate` where the push failed silently and the gate ran
anyway. Origin sat on a stale SHA while the session believed it was current — a certified gate against
a commit that was never pushed.

## Delta 2 — bare `--force-with-lease` fails "stale info" even when the lease is provably satisfied

```console
$ git push --force-with-lease origin <branch>
 ! [rejected]   (stale info)
```

with `git ls-remote` and the local remote-tracking ref both at the same SHA. The explicit CAS form works:

```bash
git push --force-with-lease=refs/heads/<branch>:<full-40-char-sha> origin <branch>
```

This matters because a worker who hits it may reach for `--force`, discarding the safety the lease exists
to provide. (Related but distinct from `premerge-assert.sh`'s short-SHA rejection in #2922.)

## Delta 3 — `gh project item-edit` fails for missing `read:org`; raw GraphQL works

The board is the **sole dispatch authority** (Path A, #1886), so a worker that cannot write board Status
cannot signal In Progress / In Review / Done. On this box `gh project item-edit` fails for a missing
`read:org` scope, while the equivalent GraphQL mutation succeeds with the *same* token:

```bash
gh api graphql -f query='mutation{ updateProjectV2ItemFieldValue(input:{
  projectId:"…" itemId:"…" fieldId:"…" value:{ singleSelectOptionId:"…" } }){ projectV2Item{ id } } }'
```

Found independently twice in one day: by me setting Status, and by the #1883 `flow-closer` setting Done.

## Why P2

Deltas 1 and 3 are **hard blocks** on a Linux worker — it cannot push a branch or move the board — and
both fail with messages that point away from the real cause (an auth error that looks like a network
problem; a scope error on a token that has the `project` scope). Delta 2 degrades a safety mechanism.
This is the same class as #2907/#2903, where a Linux-only gate defect stayed invisible because the fleet
is macOS.

## Acceptance criteria

1. **Documented**: CLAUDE.md (or `docs/development/fleet-runbook.md`, whichever the flow-* skills read)
   states the credential-helper form, the explicit `--force-with-lease=<ref>:<sha>` form, and the GraphQL
   board-write fallback, each with the failure message that identifies it.
2. **Mechanized where cheap**: `bootstrap-agent-machine.sh` configures a git credential helper from
   `GH_TOKEN` so raw `git push` works, OR a preflight check fails loudly at worker start with the fix,
   rather than at first push.
3. **Verifiable**: on a box with no git credential helper configured, the preflight (or bootstrap) is
   shown to either fix or loudly diagnose all three deltas before any pipeline step depends on them.
4. **No silent-success paths**: a failed push must not be able to leave a gate certifying an unpushed
   SHA — either the pipeline pushes and verifies (`git ls-remote` equals local HEAD) before gating, or the
   two steps are not chained.

## Out of scope (already filed — cross-link, do not duplicate)

- Gate poll predicate `RESULT:` vs `RESULT: (PASS|FAIL)` → **#2908**
- `gh pr merge` silent-failure traps → **#2922** (I observed a THIRD: `--delete-branch` reported success
  but did NOT delete the remote branch; adding as evidence there)
- flow-closer orphaning on lost re-invoke → **#2748** (adding a measured `orphan_minutes=366` data point)
- Gate `CARGO_TARGET_DIR` disk footprint → **#2758**

pmcfadin:
Closes #2942. Design-driven; owner-approved at Seam 1 ("Option A as written"). OpenSpec change: `openspec/changes/worker-env-preflight/`.

## Why

Three pipeline steps failed on a Linux worker during the #1883 delivery. The unifying defect is **not** that the docs were incomplete — it is that each failure was **misattributed**, so the diagnosis pointed away from the cause:

| symptom | actual cause |
|---|---|
| `fatal: could not read Username for 'https://github.com'` | git has no credentials, though `gh` is authenticated — separate paths. `claim.sh`/`claim-heartbeat.sh` push on 10+ sites, so the claim protocol itself doesn't work |
| `CLAIM: ERROR reason=infra … (transient — retry)` | a permanent credential fault, advising a retry that can never succeed |
| `[ok] 'project' scope present — board dispatch works` | the scope *was* present; `gh project item-edit` still failed on `read:org` |

## What changed

- **`scripts/bootstrap-agent-machine.sh`** — a new git-credential preflight, and §3's board check rewritten from a scope-string match into a **read-only functional probe**.
- **`scripts/flow/claim.sh`** — an unauthenticated push reports `reason=auth (NOT retryable)`; unrecognized failures keep the retryable infra verdict, so the #2665 contract is only ever *narrowed* by positively-identified signatures.
- **Docs** — `fleet-runbook.md`, `agent-machine-setup.md`, `CLAIM.md`/`CLAUDE.md`, website `delivery-pipeline.md`: the three deltas indexed **by their identifying failure message**.

## Design decisions worth knowing

- **The `--yes` credential helper dereferences `$GH_TOKEN` at call time and is host-scoped.** No secret is written to disk, and rotation needs no reconfiguration. `gh auth setup-git` is preferred where it works — and is **re-probed** afterwards, because its exit 0 was precisely the non-evidence that hid the original bug.
- **A check must not mutate host state.** The board probe mirrors `flow-board`'s `gh auth switch` so it measures the identity dispatch actually uses, but the restore is `trap`-armed on `EXIT`/`INT`/`TERM` and both probes inside the bracket are bounded — so a hang plus Ctrl-C cannot leave the operator's account switched. On a box where `GH_TOKEN` is in force the switch is skipped entirely (gh ignores the keyring, so it would be theatre).
- **`--force-with-lease` is documented, not mechanized** — the flow scripts already use the explicit `=<ref>:<sha>` form; wrapping git would be disproportionate.

## Review — four rounds, each finding the thesis failing one layer deeper

roborev plus an independent shell review. Every round found the change reproducing the very defect it exists to kill:

| round | the verdict was still derived from… | fixed |
|---|---|---|
| 1 | a scope *string*, not the operation | functional probe |
| 2 | a credential-fill *exit code*, not a usable secret | require non-empty `password=` |
| 3 | an account-ambiguous grep, and a *different identity* than board ops use | active-stanza parse + `CQLITE_PROJECT_ACCOUNT` |
| 4 | a hang bound inert on macOS; a guessed board number | `timeout`→`gtimeout`; resolve-by-title, no guess |

Two blockers were **security**-shaped and are worth calling out: the auto-configured helper was initially **host-unscoped** (demonstrated handing the GitHub PAT to `gitlab.com`, `evil.example`, `internal-registry.corp`), and the probe accepted an **empty password**, which re-created the exact false-OK the change exists to kill.

## Evidence

| suite | before | after |
|---|---|---|
| `test_bootstrap_agent_machine.sh` | 34 / 12 | **71 / 0** |
| `test_claim_lock.sh` | 25 / 2 | **31 / 0** |
| `test_claim_heartbeat.sh` | 23 / 0 | 23 / 0 |

**Non-vacuity is proven per-fix, not in aggregate** — each blocker reverted *in isolation* fails only its own cases (`70/1`, `70/1`, `70/1`, `68/3`). The pre-fix run reproduces the issue's verbatim `reason=infra … (transient — retry)` line, and `7h-ii` simulates a stock-macOS host (`gtimeout` only) that *hangs* against the old resolution — without it a Linux-only CI could never catch that regression.

Two **harness** bugs were found and fixed along the way, both of which had been silently weakening the suite: it inherited `CQLITE_PROJECT_NUMBER` from the invoking shell (masking the entire not-exported path in every prior run), and env overrides passed through `"$@"` were inert (bash treats an expanded `VAR=value` as a command name).

## Test isolation

`GIT_CONFIG_GLOBAL` + `GIT_CONFIG_NOSYSTEM` exported once at the top so every case — including ones added later — inherits it, plus per-case `HOME`/`CARGO_HOME` sandboxes and a throwaway git repo per case. Verified after each round: no `~/.gitconfig`, zero global credential helpers, `~/.cargo/config.toml` absent, **active gh account unchanged**, and no board mutation (8e greps every board log family for `item-edit|item-add|item-delete|--field|mutation`).

## Deliberately out of scope — filed, not folded in

Requirement 3 is push-scoped, so widening it silently would run past Seam 1:

- **#2957** — `claim-heartbeat.sh` reports a failed push as *success* (`push_liveness_ref` returns `printf`'s status), which the 4h reaper keys on. This PR adds only the one-line `GIT_TERMINAL_PROMPT=0` hang fix and narrows the docs so they stop claiming coverage the code lacks.
- **#2961** — `claim.sh`'s `ls-remote` read paths still call a permanent credential fault transient.
- **#2965** — environment variants: GHE multi-host stanza selection, `http://` origins, fine-grained/App tokens reporting `Token scopes: none`. Also notes that if #2957 and #2961 both land, `git_stderr_is_auth` should be factored once rather than copied a third time.

## Known limitation

Requirement 1 scenario 2 ("a subsequent `git push` authenticates") is proven by the credential-fill probe plus a manual `--dry-run` push, **not** by an automated end-to-end push — a hermetic test cannot push to a real remote.

## Gate

`--lite` **PASS** and `--only tooling-tests` **PASS (242s)** at `e405833c`, `dirty: no`, both run independently by the lead. `tooling-tests` is the component that actually executes both shell suites — lite's blast radius defaults to `cqlite-core --lib` and would not have run them. The full gate of record runs pre-merge.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -379,6 +379,13 @@ end-to-end test. Green helper-only unit tests are not sufficient.
   deletes the ref (refuses under an open PR without `--force`). Maintain the liveness heartbeat
   (`scripts/flow/claim-heartbeat.sh beat <N>`, refreshed at claim + every stage transition);
   `flow-board` reaps deterministically (age > 4h AND no open PR) (#2089).
+  **The lock is a plain `git push`, so git — not just `gh` — must be authenticated (#2942).** They
+  are separate credential paths: an authenticated `gh` with an unwired git fails every claim with
+  `fatal: could not read Username`, and `claim.sh` now calls that `ERROR reason=auth (NOT
+  retryable)` instead of the old misleading `reason=infra (transient — retry)` — do not retry it,
+  fix the box (`gh auth setup-git`, or `bash scripts/bootstrap-agent-machine.sh --yes`, which also
+  probes board access functionally rather than trusting the `project` scope string). The three
+  worker-environment deltas and the messages that identify them: `docs/development/fleet-runbook.md`.
 - **Supervisor-authored machine claim + CI reaper (#2655/#2499)**: liveness is now MECHANISM-driven,
   not prose. `worker-supervisor.sh` stamps `refs/machine-claims/<machine>` (issue+supervisor-PID+ts)
   via `claim-heartbeat.sh stamp` at every spawn, refreshes it each iteration, and clears it on a
--- a/docs/development/agent-machine-setup.md
+++ b/docs/development/agent-machine-setup.md
@@ -23,8 +23,9 @@ without it, it prints the exact command for each gap. It verifies, in order:
    a `project` scope-string match (#2942 — see the deltas below); the scope check
    survives only as a cheap pre-filter. Missing scope → `gh auth refresh -s project`.
    Point it elsewhere with `CQLITE_PROJECT_OWNER` / `CQLITE_PROJECT_NUMBER`.
-3b. **git push credentials** (#2942) — a *separate* credential path from `gh`.
-   Under `--yes` it configures one; the token value is never written to disk.
+   It also checks **git push credentials** (#2942) — a *separate* credential path from
+   `gh`. Under `--yes` it configures one, **scoped to the origin host**, and the token
+   value is never written to disk.
 4. **roborev** — installed, and this machine's *configured* agent resolves.
 5. **Datasets** + `CQLITE_DATASETS_ROOT` guidance.
 6. **Health check** — runs the gate's fmt smoke and prints the authoritative
@@ -38,15 +39,22 @@ The bootstrap now checks the first two and fails loudly; the third is a hand-typ
 
 | You see | Real cause | Working form |
 |---|---|---|
-| `fatal: could not read Username for 'https://github.com'` | `gh` is authenticated but **git is not** — they are separate credential paths. `scripts/flow/claim.sh` + `claim-heartbeat.sh` push with plain git on 10+ call sites, so the claim protocol itself does not work. | `gh auth setup-git`, or `bash scripts/bootstrap-agent-machine.sh --yes` (configures a helper that dereferences `$GH_TOKEN` at call time). |
+| `fatal: could not read Username for 'https://github.com'` | `gh` is authenticated but **git is not** — they are separate credential paths. `scripts/flow/claim.sh` + `claim-heartbeat.sh` push with plain git on 10+ call sites, so the claim protocol itself does not work. | `gh auth setup-git`, or `bash scripts/bootstrap-agent-machine.sh --yes` (configures an origin-host-scoped helper that dereferences `$GH_TOKEN` at call time). |
 | `gh project …` fails for a missing **`read:org`** scope on a token whose scopes DO include `project` | A scope match is evidence about a token, not about the operation. `gh project item-edit` needs `read:org`; the equivalent GraphQL mutation does not. | Widen the token (`gh auth refresh -s read:org`), **or** do board writes through the `updateProjectV2ItemFieldValue` GraphQL mutation — it succeeds with the same token. |
 | `stale info` from a **bare** `git push --force-with-lease`, even when local and remote refs demonstrably match | The bare form leases against a remote-tracking ref your checkout may never have fetched. | Always the explicit CAS form: `git push --force-with-lease=<ref>:<sha>` (what the flow scripts already use). |
 
-**Claim verdicts encode this too (#2942).** An unauthenticated claim push now reports
-`CLAIM: ERROR reason=auth … (NOT retryable …)` and names the fix. It used to report
-`reason=infra … (transient — retry)`, which sent workers into a retry loop on a machine
-fault that can never self-clear. `reason=infra (transient — retry)` still means what it
-says: a real blip, retry it.
+**Claim verdicts encode this too (#2942).** An unauthenticated push now reports
+`CLAIM: ERROR reason=auth … (NOT retryable …)` and names the fix, for `claim.sh`'s
+`claim`, `adopt`, `release` and `smoke`. It used to report `reason=infra … (transient —
+retry)`, which sent workers into a retry loop on a machine fault that can never
+self-clear. `reason=infra (transient — retry)` still means what it says: a real blip,
+retry it. **Scope:** this classification covers `claim.sh` only —
+`claim-heartbeat.sh` surfaces git's raw error on its pushes and does not classify.
+
+**The `--yes` credential helper reads `$GH_TOKEN` from the environment.** That is
+deliberate (no secret on disk), but it means the helper only works in shells where
+`GH_TOKEN` is exported — a systemd/cron worker started without it will still fail every
+push. For unattended workers prefer `gh auth setup-git`.
 
 ## Accelerators: what the SUMMARY line means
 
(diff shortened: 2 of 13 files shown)


pmcfadin:
## Endgame certification complete (`flow-closer`, #2084)

**Gate of record: PASS** — `AGENT-GATE SUMMARY` at `commit: 3490a83d dirty: no`, 29/29 components PASS, zero SKIPs. Relevant component for this bash+markdown diff: `tooling-tests: PASS (248s)` (runs `test_bootstrap_agent_machine.sh` + `test_claim_lock.sh`). Datasets present (144 Data.db), so no #2078 fail-closed.

Branch was 3 commits behind `origin/main`; **rebased before gating** (clean, zero file overlap) so the gate of record postdates the final rebase. Certified SHA `3490a83d1d0d7084df5c3c8f9c5d7113d45d1e8c`.

**C intent audit: PASS** — `spec-auditor` anchored to `openspec/changes/worker-env-preflight/specs/**`; 5/5 requirements `satisfied`, 7 scenarios covered, verified by **mutation** (each guard reverted in a scratch copy, suite confirmed to go red at the specific case). `openspec validate worker-env-preflight --strict` valid.

**Final roborev: no blockers** — job 11 (`claude-code`/`opus`): 0 Critical / 0 High / 0 Medium / **5 Low**.

**Host-safety attestation** (this change writes global git config and calls `gh auth switch`): verified untouched after the full gate — no `~/.gitconfig`, zero global credential helpers, repo-local helper still env-deref shape with no `ghp_` literal, active `gh` account still `pmcfadin`. No sandbox escape.

**Nits → #2974** — all 9 Low-severity items (5 roborev + 5 C, one deduped) batched into a single linked follow-up rather than a re-verify round (#2088). Highest-value item there: `bootstrap-agent-machine.sh:663-675` `${hostport#*@}` → `${hostport##*@}`, a one-character fix for a false-green host derivation.

Ratified scope boundaries, already filed and not gaps: #2957 (heartbeat classification), #2961 (`ls-remote` reads — requirement 3 is deliberately push-scoped), #2965 (environment variants).

Pre-merge guards: `premerge-assert.sh 2967 3490a83d…` → `PREMERGE: OK`; no `HOLD:` order on PR or issue. Arming auto-merge.

Issue #2962, opened 2026-07-26
agent-gate.sh runs cargo without --locked, so the gate can mutate the Cargo.lock it is certifying


pmcfadin:
`scripts/agent-gate.sh` invokes cargo **without `--locked` or `--frozen` anywhere**, and `Cargo.lock` is a tracked file. So a gate run can legitimately modify a tracked file in the worktree it is certifying.

Found while specifying #2926 (gate tree-integrity guard), where it forces a carve-out.

## Why it matters

1. **The gate can mutate the tree it is certifying.** A dependency resolution during the run rewrites `Cargo.lock`, so the tree at summary time differs from the tree at start through no fault of any concurrent agent.
2. **It forces a hole in #2926's integrity guard.** That change detects mid-run worktree mutation and fails closed. Because the gate itself can touch `Cargo.lock`, the guard needs a named non-fatal "lockfile-settled" class (fatal only when accompanied by any *other* change). That carve-out exists solely because the gate is not lock-pinned — remove the cause and the carve-out can go.
3. **Reproducibility.** A gate that may silently re-resolve dependencies is not certifying the same dependency graph CI will build. This is the same class as "gate PASS != CI green" (#1310).

## Fix

Add `--locked` (or `--frozen`) to cargo invocations in `scripts/agent-gate.sh`, so a run that would need to change `Cargo.lock` **fails loudly** instead of silently rewriting it.

Note `--frozen` implies `--locked` plus `--offline`; `--locked` alone is probably right, since the gate legitimately fetches.

## Acceptance criteria

- Every cargo invocation in the gate that builds or tests runs with `--locked`.
- A run whose dependency graph would require a `Cargo.lock` change FAILs with a clear message naming the lockfile, rather than mutating it.
- The out-of-date-lockfile failure is distinguishable from a compilation failure.
- Once this lands, #2926's "lockfile-settled" non-fatal class can be **removed** — the guard should treat any `Cargo.lock` change as fatal again. Cross-reference so whoever does this closes that loop.
- CI lanes stay green (they may already pass `--locked` independently; check before assuming).

## Routing

**Oracle-driven** (deterministic flag addition with an exact expected behaviour). No OpenSpec change.

## Provenance

Discovered 2026-07-26 while authoring the OpenSpec change for #2926. Related: #2926 (tree-integrity guard, carries the carve-out), #1310 (gate PASS != CI green).

Issue #2914, opened 2026-07-26
P1: gate of record FAILs on every Linux host — #2859 mold token broke three end-anchored sccache-health assertions


pmcfadin:
The **full `agent-gate.sh` gate of record currently FAILs on any Linux host**, on `main`, for every worker. Not specific to any PR.

Found while gating PR #2906 (#2825) on 2026-07-26; triaged to a pre-existing breakage.

## Failure

Component `tooling-tests` FAILs. Script: `scripts/tests/test_agent_gate_summary.sh` (invoked at `scripts/agent-gate.sh:288`). **3 failed / 90 passed**, all in block **9c "sccache cache-health token (issue #2641)"**:

- `scripts/tests/test_agent_gate_summary.sh:788` — `expected sccache-health=ok for on + 0 errors`
- `scripts/tests/test_agent_gate_summary.sh:804` — `expected sccache-health=warn for on + >0 errors`
- `scripts/tests/test_agent_gate_summary.sh:832` — `expected sccache-health=na when sccache not in use`

## Root cause

The assertions are `$`-end-anchored:

```sh
grep -qE '^accelerators: .* sccache-health=ok$' "$tmp/health-ok.txt"
```

Issue **#2859** added a **Linux-only** ` mold=<state>` suffix to the `accelerators:` line (`scripts/agent-gate.sh:616-620`). The emitted line is now:

```
accelerators: sccache=on nextest=on lanes=on sccache-health=ok mold=absent
```

The **content is correct** — the test's own diagnostic echo confirms the health token's value is right. Only the stale `$` anchor fails to match.

#2859 correctly updated the well-formedness regex at `:76` and the block 9d mold checks, but **missed the three `$`-anchored greps in block 9c**.

**Darwin emits no mold token**, so block 9c passes on macOS and fails on every Linux host. That is why this survived review and landed.

## Evidence

- A fresh `git worktree add --detach` at `origin/main` (`f8068a3`) reproduces the **identical** 3 failures: 90 passed / 3 failed. Pre-existing on `main`, not introduced by any in-flight branch.
- **Determinism: 4/4** (3 re-runs in a feature worktree + 1 on clean `main`). Fully deterministic, host-conditional.
[...]

Issue #2908, opened 2026-07-26
flow-meta: documented gate-poll predicate matches the INCOMPLETE placeholder, yielding a false 'gate done'


pmcfadin:
Found by `flow-closer` while running the gate of record for #2825 (PR #2906), 2026-07-26.

## The trap

`scripts/agent-gate.sh` writes a placeholder into `$AGENT_GATE_SUMMARY_FILE` **at launch**:

```
RESULT: INCOMPLETE (gate did not finish)
```

This is deliberate and good — it makes a killed/orphaned gate detectable rather than leaving an empty file.

But the poll predicate used in flow-closer doctrine is:

```bash
grep -q "RESULT:" "$AGENT_GATE_SUMMARY_FILE"
```

That matches the placeholder **the instant the gate starts**. So the documented "is the gate done?" check returns **true immediately**, before a single component has run.

## Why this is dangerous

A `flow-closer` that trusts the documented predicate will:

1. Poll once, see `RESULT:` present, conclude the gate finished.
2. Read `RESULT: INCOMPLETE (gate did not finish)` as its **gate of record**.
3. Either merge on a non-verdict, or report a spurious FAIL and escalate a healthy run as blocked.

Both outcomes are bad, and (2) is the worse one: the full gate is *the only run that counts*, so accepting a placeholder as its verdict silently voids the entire certification step. Nothing else in the pipeline re-checks it.

This is latent for **every** closer invocation, not specific to #2825.

## Correct predicate

```bash
grep -qE "RESULT: (PASS|FAIL)" "$AGENT_GATE_SUMMARY_FILE"
```

Only `PASS` or `FAIL` is a verdict. `INCOMPLETE` means "still running, or died".

## Acceptance criteria

- The gate-poll predicate is corrected wherever it is documented or scripted: `.claude/agents/flow-closer.md`, `docs/development/gate-ops.md`, `docs/development/pm-operating-loop.md`, the `agents-developing/gate-contract` website page, and any helper script that polls a summary file. Grep for `grep -q "RESULT:` / `RESULT:` poll patterns.
- The distinction is stated explicitly in the gate contract: `INCOMPLETE` is a **liveness placeholder**, not a verdict; only `PASS`/`FAIL` terminate a gate.
[...]

Issue #2874, opened 2026-07-25
gate: harden nested tooling-tests sub-gates against ALL summary/fixture sharing (per-invocation isolation) — residual of #2751/#2849


pmcfadin:
## Context — owner-directed promotion 2026-07-24

The nested-gate state-sharing family keeps killing full gates of record. History:
- **#2751** (CLOSED 07-19, completed): env `AGENT_GATE_SUMMARY_FILE` inheritance → nested tooling-tests sub-gates clobber the parent summary. Mitigation validated (run without the env var) and documented; underlying sharing not eliminated.
- **#2849** (CLOSED today): nested self-test gates deadlocked on the machine slot under `CQLITE_GATE_MAX_CONCURRENCY=1`. Fixed.
- **NEW EVIDENCE 2026-07-24** (#2751 comment, #2856 endgame): a full gate run **without** the env var (the documented workaround) still died INCOMPLETE in `tooling-tests` while a second lane (#2859) ran the same gate self-test files concurrently on the box. Solo re-run passed. So the residual kill surface = same-checkout default summary paths + shared self-test fixtures//tmp namespaces, reachable via cross-lane concurrency even with the env-var vector closed. Cost that day: ~2 full gate runs (~1h wall-clock) + a serialize-everything ops rule that taxes every future endgame.

## Scope

Make the parent gate's summary/state provably immune to nested sub-gate activity, and tooling-tests self-tests hermetic:
1. Nested gate invocations spawned inside `tooling-tests` run with `AGENT_GATE_SUMMARY_FILE` explicitly UNSET **and** an isolated per-invocation summary/default path (never the enclosing checkout's `.agent-gate-summary.txt`).
2. Gate self-tests (`test_agent_gate_summary.sh`, `test_bootstrap_agent_machine.sh`, siblings) use per-run mktemp namespaces for every fixture//tmp path — no fixed shared names — so two lanes running them concurrently cannot collide (mktemp Xs must be terminal on macOS).
3. The parent gate FAILS LOUDLY (named cause) rather than dying INCOMPLETE if its summary file is externally modified mid-run (foreign run-id detection → explicit error line, not a silent kill).

## Acceptance criteria

1. Regression self-test: a nested sub-gate launched with the parent's env + same checkout cannot alter the parent's summary file (asserted byte-identical parent summary).
2. Concurrency self-test: two concurrent invocations of the gate self-test files in one checkout both pass (per-run namespaces proven).
[...]

PR #2976, opened 2026-07-27
feat(#2926): fail closed when the worktree mutates mid-gate

### LINKED ISSUE #2926: Gate integrity: a mid-run worktree mutation can produce a SUMMARY that falsely certifies the new commit
A full `agent-gate.sh` run whose worktree is modified **mid-run** can emit a SUMMARY that attributes mixed-tree results to the final commit — a false certification that is indistinguishable from a legitimate one.

Hit while gating PR #2916 (#2914) on 2026-07-26.

## What happened

1. A full gate of record started at 17:21:32 against worktree HEAD `4686c37`.
2. A review-fix round committed `116d0b9` **into that same worktree** while the gate was running.
3. Early components had already run against the pre-fix file; later components ran against the post-fix file. Which file any given component saw is indeterminate.

The gate was stopped and happened to leave `RESULT: INCOMPLETE (gate did not finish)`, so the fail-safe held this time.

## The hazard

**The SUMMARY stamps `commit:` from HEAD at summary-write time, not from the tree state each component actually observed.** Had that run completed, it would have emitted:

```
commit: 116d0b9 ... RESULT: PASS
```

That is byte-identical in form to a legitimate certification of `116d0b9`, while actually being a mixed-tree run where most components never saw `116d0b9`'s code. Since the full gate is *the only run that counts*, a mixed-tree run wearing the fixed commit's SHA is **worse than no run at all** — it launders an uncertified tree into an apparently-certified one, and nothing downstream re-checks it.

The existing doctrine (one worker per machine; the #1582 retro about live-editing a worktree mid-gate) addresses the *collision*, but not this consequence: there is currently no mechanism by which a mid-run mutation makes the resulting artifact self-evidently invalid.

This is reachable without violating the one-worker rule: a single lead legitimately runs a closer (gating) and a fixer (editing) that can overlap on the same worktree, which is exactly what happened here.

## Suggested remedies (design latitude — pick one or more)

- **Record HEAD + tree hash at gate START**, stamp both start and end values in the SUMMARY, and FAIL the run if they differ (`tree-mutated-midrun`). This makes the corruption self-evident in the artifact itself.
- Snapshot the worktree (or gate a detached copy) so concurrent edits cannot reach a running gate.
- Take an advisory lock on the worktree for the gate's duration so a second writer fails fast with a clear message rather than silently corrupting a certification.

## Acceptance criteria

- A gate whose worktree HEAD or dirty-state changes between start and finish CANNOT emit `RESULT: PASS`. It fails with a distinct, named reason.
- The SUMMARY carries enough provenance to tell, after the fact, whether the tree was stable for the whole run (e.g. start/end commit + dirty flags).
- A test in `scripts/tests/` pins the behaviour: mutate the tree mid-run, assert the gate does not certify.
- Doctrine (`gate-contract` page + `CLAUDE.md`) states that a mid-run tree mutation invalidates the run, and that closers must verify tree stability before trusting a summary.

## Routing

**Design-driven** — touches the gate contract and has real latitude in the remedy (fail-closed stamp vs. snapshot vs. lock). Needs an OpenSpec change.

## Provenance

PR #2916 gate run `/tmp/agent-gate.XpbNTN`, started 17:21:32, killed after the tree moved to `116d0b9`; quarantined summary at `/tmp/gate-2914-FULL.INVALID-tree-mutated-midrun.txt`. Related: #1582 (worktree collision retro), #719 (the gate is the only run that counts), #2908 (INCOMPLETE placeholder vs verdict).

pmcfadin:
Closes #2926.

## What

`scripts/agent-gate.sh` stamped `commit:`/`dirty:` in its SUMMARY **at emit time** and read no tree state at start. A worktree mutated mid-run therefore produced a summary bearing the FINAL sha — formally indistinguishable from a legitimate certification of code most components never compiled. Since the full gate is *the only run that counts*, that launders an uncertified tree into a certified-looking one, and nothing downstream re-checks it.

This adds a **tree-identity guard**: capture identity at gate start, re-verify at every component boundary and before emitting a verdict, **fail closed** on mismatch.

Reachable without breaking the one-worker rule (#1930): a lead legitimately runs a closer (gating) and a fixer (editing) that overlap on one worktree. It happened for real on 2026-07-26 and was caught only by timing luck.

## Mechanism

Identity = HEAD + a NUL-framed **per-path content manifest**, sha256'd:
- `git --no-optional-locks diff --name-only -z HEAD` enumerates changed paths
- working-tree bytes hashed via batched `git hash-object --stdin-paths` **without `-w`** (writing would pollute an ODB shared with the root checkout)
- untracked from `ls-files -o --exclude-standard -z`; `LC_ALL=C sort -z` ordering

It catches the case a naive implementation misses — a **content change whose porcelain line is unchanged**. `git status --porcelain` alone cannot: appending twice to an already-modified file leaves the status listing byte-identical. Measured: porcelain sha `c8094453…` both times, digest `2ca89bd8…` → `47f1c403…`.

Mirrors the existing `summary-integrity` mechanism (#2874) — same `record_result` chokepoint, lane-aware MAIN-exit vs SIDE marker, post-drain application, terminal re-check — rather than inventing a parallel one.

Cost: ~30 ms/capture, ~1 s on a 40–60 min gate.

## Review found the original bug surviving inside the fix

Two reviewers, two rounds. The most serious finding was that `_tree_finalize` took the authoritative terminal capture and **then** the block's `commit:`/`dirty:` was recomputed from a fresh `git rev-parse` — the exact "stamped at emit time" pattern this change exists to eliminate. A correct mechanism had been built *alongside* the defect rather than replacing it. `commit:`/`dirty:` now derive from the verified capture (`TREE_END_HEAD`/`TREE_END_DIRTY`); with no validated capture the block reads `commit: unverified`.

Other blockers fixed:
- **A failed capture compared EQUAL** → `RESULT: PASS` on a mutated tree. `_tree_identity` returned rc 0 even when the hash tool failed, and an empty digest field was swallowed by `IFS=$'\t' read` (tab is IFS-whitespace, so the field collapsed and bound to the *fallbacks* value). Reproduced end-to-end with a stubbed `sha256sum`. Now: header + trailer + count validation, full-length-hex digest, a distinct rc 2 routed fail-closed, and comparison on head+dirty+digest.
- **Truncation after the `H` record** left start/end manifests sharing a byte-identical prefix, so a later-sorted mutation compared equal. Now a trailing `N<TAB><count>` trailer is written and validated. The docs had *claimed* truncation coverage that did not exist; wording corrected.
- **A transient `git rev-parse` blip at the slot grant disarmed the guard** (`TREE_GUARDED=0` → certifies with `tree-integrity: SKIP`). Re-capture now retains the pre-queue capture on rc 1.
- **The guarded window included the gate-slot queue** — with `CQLITE_GATE_MAX_CONCURRENCY=1` a run can sit queued for another gate's full duration, certifying nothing, yet any edit tripped the guard. Start identity is now re-captured after `acquire_gate_slot`.
- **The self-test mutate hook could write and `git commit` into the live `$REPO_ROOT`**; now requires a disposable-checkout marker.
- Process-substitution subshell discarded the fail-closed `OVERALL=FAIL`; tabs unescaped in the report corrupted the `Cargo.lock` classification; the terminal capture path raced across SIDE lanes.

## Guards are proven discriminating — including the harness itself

**8 stubbed wrong implementations**, assertions failed out of 111: inert 30 · hardfail 17 · naive-porcelain 4 · unvalidated-digest 76 · naive-commit-stamp 2 · no-trailer 4 · disarm-on-blip 3 · cap-double-count 1 · **control 0**.

Two harness-level findings worth recording:
- The first `unvalidated-digest` stub **did not discriminate** — neutering `_tree_digest_ok` alone does not reproduce the bug, because `pipefail` on `_tree_digest_file` already fails closed. The true pre-fix shape also needs `|| return 2` removed. A stub that appeared to prove the guard worked proved nothing until corrected.
- Under `pipefail`, `awk … | grep -q` reports **present** call sites as missing (SIGPIPE upstream). All structural checks are now pipe-free.

The "mechanism cannot go inert" check was itself vacuous — `grep -c '_tree_finalize'` counted the definition, doc-comment, internals and selftest hooks (9 occurrences), all of which survive deleting every terminal call site. Replaced with per-call-site `awk`-range predicates plus an ordering check, proved by deleting each site in a scratch copy.

## Scope notes

- **No escape hatch.** No env var turns a mutated run green. The only knob is a perf-only untracked-hash cap, stamped in the block, incapable of suppressing a detection.
- **Exclusions are `.gitignore` itself** via `--exclude-standard`, plus two carve-outs: the run's own summary path when a caller pins a relative one, and `Cargo.lock` as a named non-fatal `lockfile-settled` class — **fatal when accompanied by any other change**. That carve-out exists only because the gate runs cargo without `--locked` (#2962); it should be removed when that lands.
- Guards full/`--lite`/`--delta`/`--only`; exempts `--list`, `--python-build-verify`, stub and selftest emit modes.
- A detected mutation is a **verdict**, never `INCOMPLETE` (#2908), and poll predicates `grep -qE 'RESULT: (PASS|FAIL)'` are unaffected.

## Testing

`scripts/tests/test_agent_gate_tree_integrity.sh` (new, **111 assertions**, ~5 s, wired into `tooling-tests`). `test_agent_gate_summary.sh` 99 → 101 — note this change adds SUMMARY lines, which correctly tripped the accelerators-line grammar canary shipped hours earlier in #2914; the canary was updated, not loosened.

Doctrine updated in the same change: `CLAUDE.md`, `docs/development/gate-ops.md`, and the `agents-developing/gate-contract` page now state that a mid-run tree mutation invalidates the run and that a closer must verify tree stability before trusting a summary.

## Spec

OpenSpec change `gate-tree-integrity`; `openspec validate --strict` passes. One spec scenario was **logically impossible** as first written (a deleted untracked file's digest differing from the clean-tree digest — removing it *restores* that tree) and was rewritten to the properties that are actually true and tested.


pmcfadin:
## Certification record (flow-closer, #2926)

**Gate of record — FULL `agent-gate.sh` PASS at `f4904c97d1d95d3217b4d09dfea6f8125e530932`**
```
commit: f4904c9 branch: issue-2926-gate-tree-integrity dirty: no
tree-start: f4904c97d1d9 dirty: no digest: ac4ebdfc2e2f
tree-end:   f4904c97d1d9 dirty: no digest: ac4ebdfc2e2f
tree-integrity: PASS
file-size PASS · fmt PASS · clippy PASS · roborev-lints PASS · core-tests PASS ·
tombstones-scan PASS · scan-offload-guard PASS · work-counters-guard PASS ·
byte-budget-guard PASS · arrow-parity-guard PASS · memory-budget PASS ·
integration-tests PASS · format-compat PASS · write-tests PASS · cli-tests PASS ·
compaction-byte-parity PASS · query-semantics-oracle PASS ·
flight-query-semantics-oracle PASS · python-bindings PASS · node-bindings PASS ·
delivery-telemetry PASS · oom-audit PASS · parity-report PASS ·
operator-metrics-doc PASS · kit-dashboard-drift PASS · binding-unwind-profile PASS ·
tooling-tests PASS · minimal-build PASS · smoke PASS
RESULT: PASS
```
This run **exercised the new guard end-to-end**: the change's own `tree-start`/`tree-end`/
`tree-integrity` lines agree exactly with an independent manual `git rev-parse` +
`git status --porcelain` taken before and after the ~27-minute run. No `CQLITE_ALLOW_FILE_GROWTH`.

**DELTA re-certification at `409445bbe9368fc1b1362ebfa819d354917c2cad`** (docs-only `design.md` fix
found by the C intent audit; #1892 path, NOT a second gate of record)
```
MODE: delta   delta-anchor: f4904c97d1d95d3217b4d09dfea6f8125e530932 (full-gate PASS commit)
delta-files (1 allowed / 0 offending): [test/docs] openspec/changes/gate-tree-integrity/design.md
tree-integrity: PASS
file-size PASS · fmt PASS · scoped-tests PASS
RESULT: PASS
```

**C (intent audit, spec-auditor):** PASS — 10/10 requirements satisfied, no `partial`, no `unmet`,
no scope drift. Verified by independent mutants re-run against the 111-assertion suite, reproducing
every discrimination count (inert 38 · hardfail 14 · naive-porcelain 5 · naive-commit-stamp 2 ·
no-trailer 3 · disarm-on-blip 3 · cap-double-count 1 · control 0/111).

pmcfadin:
## Certification record (flow-closer endgame, #2926)

### Gate of record — full `agent-gate.sh` PASS at `56165bbc378c4f71b29541843eb0b92e7e106100`

This is the ONE gate of record. It postdates the final src change and the final rebase check.
The earlier full PASS at `f4904c9` and at `f8dbddc`, and the deltas at `409445b` and `a922021`,
are all **superseded and dead** — `agent-gate.sh` itself changed after each.

```
==== AGENT-GATE SUMMARY ====
commit: 56165bb branch: issue-2926-gate-tree-integrity dirty: no
tree-start: 56165bbc378c dirty: no digest: 8b0b3041a62c
tree-end:   56165bbc378c dirty: no digest: 8b0b3041a62c
tree-integrity: PASS
file-size: PASS · fmt: PASS · clippy: PASS · roborev-lints: PASS · core-tests: PASS
tombstones-scan: PASS · scan-offload-guard: PASS · work-counters-guard: PASS
byte-budget-guard: PASS · arrow-parity-guard: PASS · memory-budget: PASS
integration-tests: PASS · format-compat: PASS · write-tests: PASS · cli-tests: PASS
compaction-byte-parity: PASS · query-semantics-oracle: PASS
flight-query-semantics-oracle: PASS · python-bindings: PASS · node-bindings: PASS
delivery-telemetry: PASS · oom-audit: PASS · parity-report: PASS
operator-metrics-doc: PASS · kit-dashboard-drift: PASS · binding-unwind-profile: PASS
tooling-tests: PASS · minimal-build: PASS · smoke: PASS
datasets: 144 Data.db files (datasets-v3)
RESULT: PASS
```

All 29 components PASS. `file-size: PASS` without `CQLITE_ALLOW_FILE_GROWTH`.

**The change certifies itself.** `tree-integrity: PASS` is part of this verdict: `tree-start` and
`tree-end` agree on both sha and content digest (`8b0b3041a62c`). Independently, the closer recorded
HEAD + `git status --porcelain` before and after the run and confirmed `56165bb…6100` / clean at both
ends. The guard's own verdict and the external manual check **agree** — a disagreement would have been
treated as a blocker, not a curiosity.

### C — intent audit (spec-auditor)

- **Main audit: PASS, 11/11** requirements `satisfied`, each with public-surface test evidence.
- **Narrow re-check after the H2 scope change: PASS.** Verified by grep, not by claim: `TREE_MUTATED=1`
  and the verdict occur exactly once (`_tree_fail_closed`); `TREE_COMMIT_SOURCE=start` and the
  POST-MUTATION suffix exactly once (`_tree_label_post_mutation`); and **no fourth detection path exists**.

An earlier C round returned CHANGES NEEDED on stale rationale prose in `design.md`/`tasks.md` (the third
recurrence of that residue class, cf. #2821); fixed and re-certified via a markdown-only `--delta`.

### roborev — 7 passes, converged clean

Rounds 1-5 drove ~25 findings including **three distinct false-PASS routes**, two of which surfaced
*after* a passing full gate and a clean C audit. Round 6 returned 1 Medium + 4 Low; the Medium was a
real H2 asymmetry — the terminal and SIDE-lane detection paths stamped an unlabelled post-mutation sha
while only the boundary path carried the `(VERIFIED START …)` / `(POST-MUTATION …)` labels. Its root
cause was the **spec**, which scoped the rule to boundary detection, so the implementation was
conformant and C's 11/11 was honest. Round 7: **4 Low, zero blockers** — second consecutive
clean-of-blockers pass. All fixed at `56165bb`.

Contract strings are now pinned verbatim in the spec and verified byte-identical to `agent-gate.sh`
via `grep -cF`, so "emitted text matches spec" is checkable by construction rather than by prose
agreement — the property whose absence let the false-PASS routes survive a green gate.

### Deferred to #2981

Extracting `_tree_*` into `scripts/lib/gate_tree_integrity.sh`, and extending the `file-size` ratchet to
shell. The ratchet is `.rs`-only, so `agent-gate.sh` (6,445 lines) and the integrity suite (1,883) drifted
unflagged; a `file-size: PASS` here says nothing about shell growth. Known, accepted, tracked.

Suites at `56165bb`: tree-integrity 160 · portability 29 · provenance 40 · summary 101.

Issue #3040, opened 2026-07-27
0.16 release note: BTI Rows.db written by CQLite <= 0.15 is mis-rooted by the corrected row-index reader (#3002)


pmcfadin:
## Summary

#3002 corrected the `TrieIndexEntry` root-delta base in `resolve_rows_db_entry` from `rows_offset + key_length` to `rows_offset + 2 + key_length` (Cassandra 5.0.8 captures `basePosition` **after** `writeWithShortLength`), and fixed the same expression on the write side in `writer/partitions_writer.rs`.

Consequence: a BTI `Rows.db` **written by CQLite <= 0.15** encodes its root delta against the OLD (2-bytes-low) base. The corrected reader resolves `root + 2` on such a file. That offset may land on a byte that parses as a *valid* trie node, in which case a single-partition clustering `SELECT` returns a **silently wrong row window** — missing or extra rows, no error raised.

## Why this is documentation and not a code fix

Owner decision (2026-07-27, during #3002 delivery): **merge as-is + release note.**

- Those old bytes were **never Cassandra-readable** — the wrong base is precisely why. Any CQLite-written BTI `Rows.db` from <= 0.15 was already non-conformant against the authoritative oracle, so this is not a regression of a working state.
- Fail-closed detection was considered and rejected for now: distinguishing "old-CQLite bytes" from "corrupt bytes" requires inferring provenance from byte patterns, which brushes directly against the **no-heuristics mandate (#28)**, and risks rejecting valid files. There is no provenance marker in the format to key off.

## Acceptance criteria

1. The 0.16 release notes / CHANGELOG carry a **breaking-change / data-compatibility** entry stating that BTI (`da`) `Rows.db` files written by CQLite <= 0.15 have an incorrect row-index root base and MUST be rewritten (re-flush / re-compact, or regenerate from Cassandra) before being read by 0.16+.
2. The entry states the observable symptom explicitly: a clustering-restricted `SELECT` may return wrong rows **without an error**.
3. The entry notes the blast radius is limited to the BTI row-index clustering path — `data_position`/`block_count` consumers (point lookup, successor walk) are unaffected.
4. Docs mentioning CQLite's BTI write support are checked for any claim of forward/backward file compatibility and corrected if present.
5. Cross-reference #3002 (the fix) and #1406 (the uncompressed-write claim boundary) so the write-surface caveats read as one story.

## Related

[...]

Issue #3039, opened 2026-07-27
BTI fixture gap: commission a wide-partition `da` fixture with a MULTI-COMPONENT clustering (unmasks the #3002 root-base failure mode)


pmcfadin:
Follow-up to #3002 (acceptance criterion 9), deliberately deferred out of that branch — it is a fixture-commissioning task, not a code fix.

## Why the corpus cannot currently unmask this class

The shipped corpus has **exactly one non-empty `Rows.db`**: `test_da/wide_table-9099a7c06c1811f19864870fb8444786/da-2-bti-Rows.db` (760 B). Its shape is the precise one that HIDES a row-index root-base error:

* a single `int` clustering (so every separator is `40` + 4 bytes),
* all three partitions structurally identical (`RowsOffset` 242/494/748, `root_delta = -10`, `blockCount = 38`),
* and — the load-bearing coincidence — the trie root's only child is a **2-byte** `SINGLE_NOPAYLOAD_4` node, so a root that is 2 bytes low lands EXACTLY on a node boundary and still parses.

With #3002 fixed this is now correct, but the corpus still cannot *demonstrate* the unmasked failure mode: with any wider child node (`SPARSE`/`DENSE`/`SINGLE_16` — i.e. a multi-component clustering, which is what produces a root child larger than 2 bytes) a 2-low root lands **mid-node**, and the failure is either a loud `Error::corruption` out of the clustering-window path or — worse — a plausible-looking node yielding a silently wrong `body_start_rel`/`body_end_rel` (missing or extra rows, no error). That fail-open class is #28-adjacent and has no regression test today.

## Ask

Commission a wide-partition `da` (BTI) fixture with a **multi-component clustering** (e.g. `PRIMARY KEY (pk, ck1, ck2)`, ck1 `int` + ck2 `text`, ≥ 2 row-index blocks per partition so `Rows.db` is non-empty), generated from real Apache Cassandra 5.0, plus its sstabledump JSONL golden and a `cassandra-parity-manifest.yml` scenario entry.

Then add a pinned test that the fixture's row-index trie root child is **wider than 2 bytes**, and that a deliberately 2-low root (reconstructed in-test) is NOT a valid node — the property that makes the #3002 arithmetic load-bearing rather than cosmetic.

**Shared with #3023 WS4**, which independently needs a genuinely wide corpus (≥ 4 KB rows). One fixture serves both; whoever gets there first commissions it.

[...]

Issue #3032, opened 2026-07-27
Fixture gap: no da (BTI) fixture with a MULTI-COMPONENT clustering key + wide partition, so no Rows.db row-index trie whose root child is LARGER than 2 bytes


pmcfadin:
## Summary

The shipped test corpus cannot express the failure mode that issue #3002 (BTI `Rows.db` row-index root base 2 bytes low) actually fixed. #3002 corrected the arithmetic (`base = rows_offset + 2 + key_length`) and the compensating `0x40 NEXT_COMPONENT` encoder omission, but it could **not prove the unmasked failure** with the corpus we have — because the corpus contains exactly ONE non-empty `Rows.db`, and its shape happens to hide the bug.

This issue commissions the missing fixture: a Cassandra 5.0 **`da` (BTI)** table with a **multi-component clustering key** and a **wide partition**, so the `Rows.db` row-index trie root's child node is **larger than 2 bytes**.

This is AC 9 of #3002, deliberately deferred out of that branch (fixture generation needs a Cassandra container + a dataset-asset republish, which is out of scope for a read-path arithmetic fix).

**Routing: test-data / fixture work — oracle-driven.** The truth is real Cassandra 5.0 container output plus its `sstabledump` goldens. **No OpenSpec change**: GitHub issue + generated fixture + pinned parity tests.

## Why it matters — the size coincidence is what masked #3002

The corpus's only non-empty `Rows.db` is:

```
test-data/datasets/sstables/test_da/wide_table-9099a7c06c1811f19864870fb8444786/da-2-bti-Rows.db   (760 B)
```

Schema: `(pk int, ck int, payload text, PRIMARY KEY (pk, ck))` — a **single `int` clustering component**. All three partitions are structurally identical (`delta = -10`, `block_count = 38`).

Cassandra's incremental trie writer emits children before parents, so the root's last child ends at `root - 1`. On this fixture that child is a **2-byte `SINGLE_NOPAYLOAD_4`** node — so the wrong root, `root - 2`, lands **exactly on the child's first byte**. It parses cleanly, walks the right subtree, and produces correct (merely over-inclusive) output. That is the entire reason a live-`SELECT` off-by-two survived in tree and was even *pinned as the expectation* (`issue_832_bti_traversal.rs:283` asserted `trie_root == 236`; the Cassandra-derived value is `238`).

[...]

PR #3044, opened 2026-07-27
fix(bti): correct Rows.db row-index root base + OSS50 leading NEXT_COMPONENT (two compensating defects, #3002)

### LINKED ISSUE #3002: BTI Rows.db row-index root base is 2 bytes low vs Cassandra (missing short-length prefix) on the live SELECT clustering path — masked by a compensating encoder defect
## Summary

`resolve_rows_db_entry` computes the `TrieIndexEntry` root-delta base **2 bytes low** vs Cassandra. This is on a **LIVE `SELECT` read path**. It is currently masked by a **second, compensating defect** in the OSS50 clustering encoder (it omits the leading `0x40 NEXT_COMPONENT`), plus a fixture-specific node-size coincidence. **The two bugs cancel today — so fixing either one ALONE will break BTI clustering reads.** They must be fixed together.

Found during the #2953 BTI guide audit.

**Routing: oracle-driven** (Cassandra source + real fixture bytes are the truth) → no OpenSpec change; GitHub issue + pinned parity test.

## 1. The CQLite expression

`cqlite-core/src/storage/sstable/bti/parser/rows.rs:391-392`:

```rust
// base for the SIGNED root delta = RowsOffset + key_length (see module note).
let base = rows_offset + key_length;
```

Note the same file already computes the correct value 3 lines earlier for the *entry* start (`rows.rs:383`):

```rust
let entry_start = rows_offset + 2 + key_length;
```

`rows_offset` is the `RowsOffset` from a Partitions.db lookup — the offset of the **start of the key record** (the `u16` length prefix). `key_length` is the `u16` value itself, i.e. the key bytes only, **excluding** its own 2-byte prefix. So `base` is short by exactly the 2-byte short-length prefix.

The module-level layout comment `rows.rs:307` encodes the same wrong claim:
```
//   [trie_root - base    : SIGNED vint]         ← base = RowsOffset + key_length
```

## 2. Cassandra 5.0.8 authority

**Writer** — `src/java/org/apache/cassandra/io/sstable/format/bti/BtiTableWriter.java:184-187` (`IndexWriter.append`):

```java
long indexStart = rowIndexWriter.position();
ByteBufferUtil.writeWithShortLength(key.getKey(), rowIndexWriter);
((TrieIndexEntry) indexEntry).serialize(rowIndexWriter, rowIndexWriter.position(), descriptor.version);
```

`basePosition` is `rowIndexWriter.position()` **after** `writeWithShortLength` — i.e. after both the 2-byte length and the key bytes. `indexStart` (= `RowsOffset`) is captured *before*.

**Reader** — `BtiTableReader.java:191` (`retrieveEntryIfAcceptable`):

```java
ByteBuffer indexKey = ByteBufferUtil.readWithShortLength(in);   // consumes 2 + keyLen
...
return TrieIndexEntry.deserialize(in, in.getFilePointer(), descriptor.version);
```

`in.getFilePointer()` is again *after* the short length + key.

**Delta application** — `TrieIndexEntry.java:deserialize`:
```java
long indexTrieRoot = in.readVInt() + basePosition;
```

**Correct formula:** `base = RowsOffset + 2 + key_length` (= `entry_start`, the position where the `data_position` vint begins). CQLite's `base` is `entry_start - 2`.

## 3. LIVE, not latent — concrete call chain

`resolve_rows_db_entry`'s `trie_root` is consumed on the user-facing single-partition clustering read path:

```
Database SELECT (pk = ? AND ck <bounds>)
  -> storage::sstable::mod.rs:1683  SSTableReader::scan_single_partition_clustering
  -> data_access/big_promoted.rs:238  resolve_clustering_seek_window
  -> data_access/bti.rs:301  bti_clustering_row_window
  -> bti.rs:343  resolve_rows_db_entry(rows_db, rows_offset)     // WRONG BASE
  -> bti.rs:348  let root = header.trie_root;
  -> bti.rs:390  rows_floor_block(rows_db, root, &start_bytes)
  -> bti.rs:396  rows_strict_ceiling_block(rows_db, root, &end_bytes)
  -> ClusteringRowWindow { body_start_rel, body_end_rel }   // the decoded byte window
```

Not test-only, not behind a non-default feature. It is `#[cfg(not(feature = "tombstones"))]` — i.e. **compiled in the DEFAULT build**. `header.data_position` from the same function is also used by the point-lookup path (`reader/partition_lookup.rs:393`) and the successor walk (`reader/partition_successor.rs:142`) — those two consume only `data_position`/`block_count`, which are unaffected, so the blast radius is the `trie_root` consumers above.

## 4. Empirical byte demonstration (real `da` fixture)

`test-data/datasets/sstables/test_da/wide_table-9099a7c06c1811f19864870fb8444786/da-2-bti-Rows.db` (760 B, the only non-empty `Rows.db` in the corpus). Decoded entry for pk=1: `RowsOffset=242`, `key_length=4`, key `00000001`, `entry_start=248`, `data_position=0`, `root_delta=-10`, `block_count=38`.

| | base | trie_root | byte | node type | payloadFlags |
|---|---|---|---|---|---|
| CQLite (`+key_length`) | 246 | **236** | `0x12` | `SINGLE_NOPAYLOAD_4` | 0 (type cannot carry payload) |
| Cassandra (`+2+key_length`) | 248 | **238** | `0x21` | `SINGLE_8` | **1** |

Identical divergence at all three partitions (`RowsOffset` 242/494/748 → CQLite 236/488/742 vs Cassandra 238/490/744; every entry has `delta=-10`, `block_count=38`).

Full faithful traversal (`TrieNode` semantics transcribed from 5.0.8) from each root:

```
root=236 (CQLite):    38 entries; first = ('80000008', 16512)      <- no empty-key entry
root=238 (Cassandra): 39 entries; first = ('', 7)                  <- the ByteComparable.EMPTY root payload, block 0
                                  then ('4080000008', 16512), ...
```

So the fixture **does** distinguish the two, and it confirms the reporter's headline: CQLite's root misses the empty-key block-0 payload (38 vs 39 entries).

**But the crux — why reads still work today.** The true root (238) is `SINGLE_8` with transition byte `0x40` and pointer delta 2, so its only child is at **236 — exactly CQLite's wrong root**. Two coincidences make this benign on this fixture:

1. Cassandra's incremental trie writer emits children before parents, so the root's last child ends at `root-1`. Here that child is a 2-byte `SINGLE_NOPAYLOAD_4`, so `root-2` lands precisely on its first byte. **If that child were any other node type (SPARSE/DENSE/SINGLE_16 — i.e. a wider clustering trie), `root-2` would land *mid-node*.**
2. CQLite's production encoder omits the leading `0x40`, so its bounds are keyed for exactly the subtree rooted at the `0x40` child — which is what the wrong root points at.

Net effect on this fixture: the traversal walks the right subtree with the right (un-prefixed) keys, and the lost block-0 empty separator is silently absorbed by the #1968 "implicit first block" fallback (`rows_floor_block` returns `None` → `includes_implicit_first_block` → decode from rel 0). Over-inclusive, therefore not wrong *output*.

**Unmasked failure modes** (not demonstrable with the shipped corpus — only one non-empty `Rows.db` exists, single `int` clustering, all three partitions structurally identical):
- child node size != 2 → `root-2` lands mid-node → arbitrary type nibble → either `Error::Parse` surfaced as `Error::corruption` from `bti.rs:343`/`390`/`396` (**loud failure of a valid `SELECT`**), or a plausible-looking node → **silently wrong `body_start_rel`/`body_end_rel` → missing or extra rows with no error**. The window is a byte range handed to the decoder; a wrong-but-parseable window is not validated against anything.

The silently-wrong-rows branch is the **#28-adjacent fail-open class** and is the reason to fix this now rather than treat it as cosmetic.

## 5. Why current tests miss it

`cqlite-core/tests/issue_832_bti_traversal.rs:283` **pins the buggy value as the expectation**:

```rust
assert_eq!(header.trie_root, 236, "pk=1 trie root must resolve to 236");
```

The correct value is 238. This test actively locks in the defect.

And `issue_832_bti_traversal.rs:32-34`:

```rust
fn encode_ck_int(ck: i32) -> Vec<i32-bytes> {
    ((ck as u32) ^ 0x8000_0000).to_be_bytes().to_vec()   // -> 80000008
}
```

**The `0x40` half of the reporter's claim HOLDS.** The real on-disk separator is `40 80 00 00 08` (proved by the traversal above); the helper synthesizes `80 00 00 08`. Per `ClusteringComparator.java:260-275` (5.0.8), `ByteComparableClustering` "Adds a NEXT_COMPONENT byte **before each** component" — including the first (worked examples: `"A", 0005, Clustering -> 40 4100 40 0005 40`).

Critically this is **not only a test-helper defect** — the *production* encoder has the same omission: `cqlite-core/src/storage/sstable/bti/parser/encoding.rs:121-133` and `:164-186` emit `OSS50_NEXT_COMPONENT` only *between* components (`if i > 0`), never a leading one. The doc comment at `encoding.rs:114-117` states the wrong invariant as intentional ("A single-component clustering encodes to the bare component bytes ... NO framing"), calibrated against the same mis-rooted traversal.

**The `0x38` terminator half of the claim is REFUTED for row-index separators.** The real separators carry no `0x38`: `RowIndexWriter.add` (`RowIndexWriter.java:66-75`) stores `ByteComparable.separatorGt(prevMax, firstName)` and `complete()` stores `nudge(prevMax, i)` — nudged *prefixes*, not complete clusterings, so no `ByteSource.TERMINATOR` (`0x38`, `ByteSource.java:72`) is present. Adding `0x38` would be wrong.

So the two production defects are mutually compensating, and the test suite validates the pair against fabricated bytes rather than against Cassandra.

## Acceptance criteria

1. `rows.rs:392` becomes `let base = rows_offset + 2 + key_length;` (i.e. reuse `entry_start`); update the layout comment at `rows.rs:307` and the two inline unit tests at `rows.rs:833` / `rows.rs:871`.
2. `encoding.rs` (`encode_clustering_bound_oss50` and `encode_clustering_bound_oss50_with_order`) emits a leading `0x40 NEXT_COMPONENT` before the **first** component, matching `ClusteringComparator.ByteComparableClustering`. Do **not** add a `0x38` terminator. Correct the `encoding.rs:114-117` doc invariant.
3. **Both together in one change.** A pinned test must demonstrate that each fix alone regresses the read path (the compensation is load-bearing).
4. **New pinned parity test against the real `da` fixture that FAILS on today's tree**: `resolve_rows_db_entry` on `test_da/wide_table` `RowsOffset=242` yields `trie_root == 238`, the node there is `SINGLE_8` with `payloadFlags == 1`, and a full traversal from it yields **39** entries whose first is the empty-key separator with `data_offset == 7`. Assert the same for `RowsOffset` 494 and 748 (roots 490/744).
5. Update `issue_832_bti_traversal.rs:283` (`236` → `238`) and fix `encode_ck_int` at `:32` to emit genuine OSS50 bytes (`40 80 00 00 08`).
6. `rows_floor_block` must now return the block-0 empty separator as a real floor (rather than relying on `None`/implicit-first), and the #1968 implicit-first path must keep its behavior for genuinely-below-first-separator bounds. Add a case proving block 0 is reachable as a stored entry.
7. **Query-semantics parity** (`test-data/query-semantics-oracle.json`): a `SELECT ... WHERE pk=? AND ck` slice over `test_da.wide_table` returns identical rows before/after, at a pinned `now` — proving the compensation removal is behavior-preserving where it was previously benign.
8. Add a `point_vs_full_differential` case for a BTI clustering slice on `test_da.wide_table` (forced `CQLITE_READ_PATH=point` vs `=full`).
9. **Fixture gap:** the corpus has exactly one non-empty `Rows.db`, with a single `int` clustering, whose root child happens to be 2 bytes — the precise shape that hides this. Commission (or file a follow-up for) a wide-partition `da` fixture with a **multi-component** clustering, which produces a root child larger than 2 bytes and would expose the unmasked failure.

## Related

#832 (BTI trie traversal iterators — introduced `resolve_rows_db_entry` and the `236` pin), #1647 (`separatorFloor` walk — consumes `trie_root`), #1968 (implicit first block — the fallback currently absorbing the lost block-0 payload), #2058 (uncounted resolve variant — shares the same arithmetic).

pmcfadin:
Closes #3002

## What this fixes

Two **mutually-compensating** production defects on the BTI (`da`) live `SELECT` clustering path. Each fix alone regresses reads, so they ship together.

1. **Row-index root base was 2 bytes low.** `resolve_rows_db_entry` computed the `TrieIndexEntry` root-delta base as `rows_offset + key_length`; Cassandra 5.0.8 captures `basePosition` *after* `writeWithShortLength` (`BtiTableWriter.IndexWriter.append`, `BtiTableReader.retrieveEntryIfAcceptable`), making the correct base `rows_offset + 2 + key_length` — i.e. the already-computed `entry_start`. On the shipped `test_da/wide_table` fixture the root moves `236 → 238` (also `488 → 490`, `742 → 744`).
2. **OSS50 clustering encoders omitted the leading `0x40 NEXT_COMPONENT`.** Per `ClusteringComparator.ByteComparableClustering`, a NEXT_COMPONENT byte precedes *each* component including the first. No `0x38` TERMINATOR was added — the issue refutes that for row-index separators (`RowIndexWriter` stores `separatorGt`/`nudge` prefixes, not complete clusterings).
3. **Write side, same arithmetic.** `writer/partitions_writer.rs` carried the identical base expression. Left unfixed, the corrected reader would mis-root **CQLite-written** `Rows.db`, so it moves in lockstep.

## Evidence

- Traversal from the corrected root yields **39** entries (was 38), first = the empty-key `ByteComparable.EMPTY` separator at `data_offset == 7` — the block-0 payload the old root missed.
- Node at each corrected root is `SINGLE_8` with `payloadFlags == 1`; asserted for `RowsOffset` 242 / 494 / 748.
- **Each half alone regresses, proven executably** (not asserted in a comment): the resolved `(start, end)` window is pinned per half and asserted disjoint from the measured ck=100..109 byte range `[198144, 231168)` — half A collapses to `(last_block, MAX)`, half B to `(0, 16512)`. Both make the public `SELECT ... WHERE pk=1 AND ck >= 100 AND ck < 110` fail loudly.
- Wiring evidence through the real public path: `SSTableReader::scan_single_partition_clustering` → `resolve_clustering_seek_window` → `bti_clustering_row_window`.

## Tests

New `cqlite-core/tests/issue_3002_bti_rows_root_base.rs`; both parity oracles added (`test-data/query-semantics-oracle.json` at a pinned `now`, plus a `point_vs_full_differential` BTI clustering-slice case); `issue_832_bti_traversal.rs` de-pins the defect (`236 → 238`) and `encode_ck_int` now emits genuine OSS50 bytes.

Hardened against vacuous passes: `CQLITE_REQUIRE_FIXTURES` fail-closed hooks + explicit SKIP messages on the new lane; the core oracle now fails closed on *any* skipped case; and the four pre-existing `"returned 0 rows (Data.db not fetched?)"` guards in `issue_954`/`issue_1647` are now **hard failures** — those are the lanes that should have screamed during this change (verified: they now scream under each half-revert). Window size itself is pinned via `wc::rows_decoded` (17 for `ck < 8`, 9 for `ck = 150`, vs 301 full), because an over-inclusive window returning correct rows is exactly how this bug hid for two releases.

`rows.rs` was 1007 lines and could not grow, so its inline tests moved to `rows_tests.rs` (campsite rule; precedent `partitions_writer_tests.rs`) — verified 14/14, no `#[ignore]`, no weakened assertions.

## Review-first (before any full gate, per #2086)

- **rust-reviewer**: 0 blockers. Independently confirmed writer/reader symmetry against Cassandra's `basePosition`, and that the test extraction is 1:1.
- **coverage-reviewer**: 1 **blocker** — AC 3's half-B arm only asserted `rows_floor_block` returning `None`, which is the *documented-benign* #1968 implicit-first signal, so it proved "narrowing lost", not a wrong answer; the real half-B failure is the end bound via `rows_strict_ceiling_block`, which had no test hook at all. **Fixed** (item 1 above). It also independently decoded the fixture, confirming the pins are Cassandra-derived rather than calibrated to CQLite's new behavior.
- **roborev** (`claude-opus-5`): 1 Medium + 4 Low, no high/critical. The Medium — the writer's empty-key branch silently filed an empty separator under transition byte `0x00` instead of failing — is **fixed**: `insert_row` now returns `Error::InvalidInput`. Three of the four Lows independently duplicated findings already queued from the other reviewers.

## Owner decisions recorded

- **Old-bytes compatibility break → release note only** (#3040). A BTI `Rows.db` written by CQLite ≤ 0.15 is now mis-rooted and may yield a silently wrong window. roborev suggested a root sanity-check; **deliberately not added** — distinguishing old-CQLite bytes from corrupt bytes requires inferring provenance from byte patterns, violating the no-heuristics mandate (#28). Those bytes were never Cassandra-readable to begin with.

## Follow-ups filed

- **#3032** — fixture gap: no `da` fixture with a multi-component clustering (the 2-byte root child is what masked this). #3039 closed as a duplicate.
- **#3040** — 0.16 release note for the compatibility break above.
- Writer block-0 canonicality gap — pinned as a KNOWN GAP comment in this PR, follow-up issue filed separately.

## Gate

Lite PASS at `276d77e` (`tree-integrity: PASS`). The **full gate of record has not run yet** — `flow-closer` runs it once, then C (intent audit) + a final roborev pass, before merge.


pmcfadin:
## Independent read-path verification of the #3002 defects — three items that may not be covered by the current diff

Drive-by from a **parallel lead session** running an independent read-path audit of the #3002 defect pair. **Nothing here blocks your merge decision** — it is input for your consideration before re-cert, and the call is yours.

Up front: **the core of this PR looks right.** The writer-side fix (`base = entry_start + 2 + key_length`, `partitions_writer.rs:1167`) and the arity-irrelevance framing (`0x40 NEXT_COMPONENT` before **every** component, including the first) both match exactly what independent verification found. The `validate_rows_trie_root` extent-ends-at-`RowsOffset` invariant is a genuinely strong structural check, and the "necessary but not sufficient" caveat in the module docs is the honest framing.

Three items below, with confidence labels I've tried not to inflate.

---

### FINDING 1 — the break condition is node WIDTH, not clustering arity

Cancellation of the two framing defects requires the root's child to be **exactly a 2-byte `SINGLE_NOPAYLOAD_4` node**, so that `root - 2` pre-consumes exactly the single `0x40` hop. Byte-decoding the fixture confirms this is precisely what happens, identically in all three partitions:

```
root=238  byte=0x21  ord=2 (SINGLE_8)          pflags=1  transition=0x40  delta=2  child=236
child=236 byte=0x12  ord=1 (SINGLE_NOPAYLOAD_4)          transition=0x80  width=2
root=490 / child=488  — identical shape
root=744 / child=742  — identical shape
```

That child's transitions are the distinct **first bytes** of the first clustering column's byte-comparable values. It is 2 bytes wide here only because all ~38 separators are small positive `int`s whose sign-flipped image begins `0x80` — i.e. **one** transition. So the fixture is safe *incidentally*, not because it is arity-1.

The defect therefore surfaces at **arity 1** whenever those first bytes vary:

- **`text` / `ascii`** first clustering column → first bytes spread across the alphabet → `SPARSE_8` (`2 + 2c` bytes: header, count, `c` transition bytes, `c` 1-byte pointers — per `parse_sparse_node` + `pointer_bytes_for_ordinal(5) == 1`)
- **`uuid`** (random v4) → effectively arbitrary first byte (CQLite encodes the raw 16 bytes; `timeuuid`'s leading byte varies more slowly, so it is a weaker trigger)
- **`bigint` / `timestamp`** with varied high bytes
- **`int`** spanning negatives (first byte `< 0x80`) or values `>= 2^24` (`16_777_216`), since the first byte is `0x80 | (v >> 24)`
- ANY row-index trie exceeding one 4096-byte `PageAware` page, where `IncrementalTrieWriterPageAware.layoutChildren` places children non-adjacently and the root widens to `SINGLE_16`+

Framing the trigger as multi-component/arity **understates it**: realistic wide partitions with a `text` or `uuid` clustering key land in the unsafe regime at arity 1.

**Confidence:** the `0x40`-before-every-component rule and the arity-irrelevance of the defect are **VERIFIED BY BYTE DECODE** of the real `test_da/wide_table` `da-2-bti-Rows.db` (root transition `0x40`; the second separator's on-disk bytes are literally `40 80 00 00 08`) — this matches your own pin. *Correcting my own provenance claim:* I could **not** verify against a Cassandra source checkout — `~/local_projects/cassandra` is absent on this machine — so for `ClusteringComparator.ByteComparableClustering` I am relying on your citation (`ClusteringComparator.java:260-275`) plus the fixture bytes, not on an independent source read. The node-width break mechanism is **HIGH** confidence; the exact size threshold is **MEDIUM**; the `PageAware` page-crossing bullet is an **inference from `TrieNode`/`PageAware` layout that I could not verify locally**. The precise row-index size at which the root's child stops being 2 bytes is **UNVERIFIABLE** without a Cassandra-written multi-block BTI fixture with a `text`/`uuid` first ck — none exists in the corpus.

---

### FINDING 2 — `bti_point.rs` is not in this PR's changed files, and it has an unguarded window inversion

Verified against `gh pr view 3044 --json files`: the diff touches `reader/data_access/bti.rs` but **not** `reader/data_access/bti_point.rs`.

In `cqlite-core/src/storage/sstable/reader/data_access/bti_point.rs`, `bti_collect_partition_rows` (lines **783-786** on `origin/main`) clamps both ends to the available bytes but never checks `start <= end`:

```rust
let clamped_window = row_body_window.map(|(start, end)| {
    let avail = window.len().saturating_sub(within);
    (start.min(avail), end.min(avail))
});
```

Traced through `parse_block_emit_windowed`: an inverted `(start, end)` fast-forwards `offset = body_start` (line ~213), then the row loop's `if offset >= body_end { break }` (line ~316) fires immediately → **zero rows, no error**. Both call sites (`bti_point.rs:624` and `:646`) map straight to `Some(rows)`, so an empty vec is returned as an authoritative "partition has no matching rows" — a silent wrong result rather than a fall-back to a full-partition decode. `reader/data_access/big_promoted.rs:454` has the identical clamp with the identical gap, and is likewise not in this diff.

To be precise about reachability: I do **not** have a case where the *current* walks produce an inverted pair on their own. `bti.rs` builds `body_start_rel` from the floor block and `body_end_rel` from the ceiling block, so inversion needs a ceiling resolving *below* a floor — i.e. it needs a mis-resolved trie, which is exactly the failure class this PR is fixing. So this is **defense-in-depth for the residual**, not an independently reachable input.

The related hardening in the same family, which is what makes the residual *silent*: `classify_node_nibble` (`bti/parser/node_decode.rs:24`) accepts **all 16** ordinals (`0`, `1..=4`, `5..=9`, `10..=15` — the error arm is unreachable for a nibble), and every child offset uses `saturating_sub` (lines 109/120/136/188/207/287). So a wrong offset parses as *some* valid node and never panics.

Worth deciding **explicitly** whether a `start > end → treat as un-narrowable, return `None`` guard belongs in this PR or in a separate hardening issue. It is arguably a distinct defect from the framing fix, and I would not push it into scope on your behalf.

---

### FINDING 3 — mostly already addressed by your diff; keeping the note only for the corrected detail

My original concern was that `issue_908_bti_canonical_write.rs` cannot serve as authoritative evidence for the framing, because it is CQLite-**written** and CQLite-**read**: reader and writer shared the same wrong `base` expression, so they compensated for each other and the test was **invariant** to a uniform framing error.

That concern is real for the *pre-PR* test, but **your diff already closes it** — `assert_trie_index_entry_base_is_canonical` decodes the emitted `TrieIndexEntry` by hand against the canonical base and asserts the resolved root's serialized extent ends exactly at `rows_offset`, and your own comment states the lockstep hazard verbatim ("a round-trip through the reader would pass just as well with the OLD (2-low) base"). The `regressed_root = root + 2` self-check is exactly the discriminator that makes it non-vacuous. So: no action requested here, and please read this as agreement rather than a finding.

One factual correction to carry, in case it matters for how you word the evidence: **the test the canonical-base assertion attaches to is arity-1** — `wide_schema()` / `ClusteringKey::single` / `PRIMARY KEY (pk, ck)` (file lines 280-290, 332), not arity-2. (The same file *does* contain an arity-2 mixed ASC/DESC case, `wide_mixed_schema` / `bti_wide_mixed_order_partition_resolves_through_rows_db` at line ~644, which is presumably where an "arity-2" impression came from — mine included.) The structural check remains good evidence for ordering/writer-base; only a **Cassandra-written** fixture can falsify the read-side framing end to end.

---

### Closing — on the fixture gap (#3032)

#3032 (*fixture gap: no `da`/BTI fixture with a multi-component clustering key + wide partition*) reads **narrower than the defect as verified**. Its AC 1 requires a compound clustering key of *">= 2 components of differing types"*, which misses:

- (a) a wide partition with a **single** `text`/`uuid` clustering column, and
- (b) a row index exceeding one 4096-byte page,

both of which break at **arity 1** per Finding 1. Note its AC 2 (*"root child node type is NOT a 2-byte node"*) is already the *right* property — it is just tied to the too-narrow schema requirement in AC 1.

Suggestion: **widen #3032's acceptance** (make the discriminating property "root child is not a 2-byte node", and let a single `text`/`uuid` ck or a page-crossing row index satisfy it) rather than filing a fourth issue. To be explicit: **widening #3032's scope is the OWNER's call, not ours** — neither of us should re-scope it unilaterally.

---

*Posted by a parallel lead session doing an independent read-path audit. I am **not** taking over #3002 or #3044 — no claim ref, label, or issue state touched, and no files edited. Merge timing and disposition of all three items are entirely yours.*

pmcfadin:
## Certification record — #3002 (BTI `Rows.db` row-index root base)

Merged head: **`6b810e521ed60f83fdffe47becf67f905a6077d7`**. Pre-merge SHA assert (#2456) returned
`PREMERGE: OK 6b810e521ed60f83fdffe47becf67f905a6077d7`; no `HOLD:` order was open on #3002 or #3044.

### 1. Gate of record — FULL `agent-gate.sh` PASS at `d77bf09`

Verified against the reader contract (#2874/#2926) before being trusted: process exit code 0, `run-id`
equal to the run launched by the closer, `tree-integrity: PASS`, `tree-start == tree-end`, and no
`.integrity-fail.*` sibling.

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.ZUqcfk
commit: d77bf09 branch: issue-3002-bti-rows-root-base dirty: no
datasets: 144 Data.db files under /home/ubuntu/workspace/repo/test-data/datasets
tree-start: d77bf09acd9d dirty: no digest: 8c2790956950
tree-end: d77bf09acd9d dirty: no digest: 8c2790956950
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (7s)
clippy:            PASS (157s)
roborev-lints:     PASS (1s)
core-tests:        PASS (546s)
tombstones-scan:   PASS (44s)
scan-offload-guard: PASS (151s)
work-counters-guard: PASS (92s)
byte-budget-guard: PASS (25s)
memory-budget:     PASS (150s)
integration-tests: PASS (21s)
format-compat:     PASS (4s)
write-tests:       PASS (144s)
cli-tests:         PASS (74s)
flight-query-semantics-oracle: PASS (9s)
python-bindings:   PASS (325s)
node-bindings:     PASS (250s)
delivery-telemetry: PASS (0s)
oom-audit:         PASS (2s)
parity-report:     PASS (3s)
operator-metrics-doc: PASS (29s)
tooling-tests:     PASS (288s)
minimal-build:     PASS (58s)
smoke:             PASS (69s)
logs: /tmp/agent-gate.ZUqcfk
summary-file: /tmp/gate-full-3002-d77bf09.txt
RESULT: PASS
```

### 2. `--delta` re-certification at the merged head `6b810e5`

The merged head is `d77bf09` plus a **tests-only** diff. `git diff --name-only d77bf09..6b810e5` contains
no `src/` path — all five files are under `cqlite-core/tests/` — which the delta gate independently
confirms as `5 allowed / 0 offending`. Per the `--delta` contract this block is **NOT** the gate of
record; it is recorded alongside its anchor's full SUMMARY above.

```

run-id: /tmp/agent-gate.wyY234
MODE: delta (TEST/DOCS-ONLY RE-CERTIFICATION — NOT the gate of record; gate of record = the full agent-gate.sh PASS at anchor d77bf09)
commit: 6b810e5 branch: issue-3002-bti-rows-root-base dirty: no
delta-anchor: d77bf09acd9df011aa6c52c7b7391ec10b7ae83b (full-gate PASS commit)
delta-anchor-run-id: /tmp/agent-gate.ZUqcfk
gate-of-record: full agent-gate.sh run at d77bf09acd9df011aa6c52c7b7391ec10b7ae83b (this DELTA re-certifies a test/docs-only diff; it is NOT a substitute for the full gate)
delta-executors: scoped-tests(rust/python) (executors that RAN this re-cert)
tree-start: 6b810e521ed6 dirty: no digest: 10bdb51f3e33
tree-end: 6b810e521ed6 dirty: no digest: 10bdb51f3e33
tree-integrity: PASS
delta-files (5 allowed / 0 offending):
      [test/docs] cqlite-core/tests/issue_1647_rows_floor_walk.rs
      [test/docs] cqlite-core/tests/issue_1968_bti_open_lower_bound.rs
      [test/docs] cqlite-core/tests/issue_3002_bti_rows_root_base.rs
      [test/docs] cqlite-core/tests/issue_3002_multipage_rows_root.rs
      [test/docs] cqlite-core/tests/issue_954_clustering_slice_seek.rs
file-size:         PASS (0s)
fmt:               PASS (5s)
scoped-tests:      PASS (96s)
logs: /tmp/agent-gate.wyY234
summary-file: /tmp/gate-delta-3002b.txt
RESULT: PASS
```

### 3. Intent audit (C) — PASS

Routing was **oracle-driven** (Cassandra 5.0.8 source + real `da` fixture bytes are the truth), so there
is no OpenSpec change and no archive step; `spec-auditor` was anchored to the numbered acceptance criteria
on #3002 instead. Across five audits: **ACs 1-8 `satisfied`** with public-surface test evidence,
**AC 9 out of scope** (deferred to #3032 — the multi-component `da` fixture). Final audit reported
`AC-REGRESSION: none`, and confirmed the `trie_root` type change *strengthened* AC 4's evidence
(`require_trie_root()` is a hard `expect`, no assertion degraded to a bare `is_ok()`).

### 4. roborev — 6 passes, converged

`--model claude-opus-5` throughout. Severity decreased monotonically: Medium+4 Low → Medium+3 Low →
2 Low → **final round: 2 Low, zero Medium/High**. Every blocker was fixed pre-merge; the final two Lows
were **batched** under the agreed rule (a fresh Low that is not in the blocking exception set —
correctness, data-parity, no-heuristics, unwrap-panic, wiring-evidence, stated-AC, or a fix introducing a
new defect — is batched rather than triggering another full gate). Both are recorded individually with
roborev's suggested fixes in **#3076**, not summarised away:

- `verify.rs:1092-1172` — the "intact != corrupt" wording covers only 2 of 7 rejection reasons.
- `rows.rs:~660-680` — a stale doc block asserting the implicit first block "CANNOT" be returned, which
  the corrected root makes false.

### 5. Disclosure — the gate ran with `CQLITE_ALLOW_FILE_GROWTH=1`

The new `cqlite.read.bti.rows_root_rejected` counter grew two files that were **already over the
file-size threshold**:

| file | before | after |
|---|---|---|
| `catalog.rs` | 1316 | 1344 |
| `otel.rs` | 937 | 947 |

Stated plainly: **`file-size` PASSed only because of that flag.** Splitting the observability registry is
epic **#1116** and was explicitly out of scope here; CLAUDE.md's campsite rule sanctions this path
provided the bypass is disclosed and linked, which is the purpose of this section.

### 6. Caveat — the counter's wiring assertions are feature-gated

The `rows_root_rejected` wiring tests assert the counter and its `rows_root_reject_reason` attribute and
are mutation-verified (commenting out the `add_counter` block fails both). But they require
`--features observability-testing,cli-helpers`, because `add_counter` is a compile-time no-op without
`observability` — so **the default gate build does not run them**, and within the gate's own feature
matrix the "delete the emission and the suite stays green" hazard technically persists. Wiring it into a
CI lane needs a workflow change, which `--delta` forbids, so it was correctly not attempted here. This is
a pre-existing structural gap rather than a defect of this PR (metric-wiring tests generally fall outside
the gate's 29 components). Tracked in **#3076**, cross-referenced to **#3070**.

### 7. Provenance

**4 full gates, 2 `--delta` re-certs, 2 `rust-reviewer` passes, 2 `coverage-reviewer` passes, 5 C audits,
6 roborev passes; 3 blockers and 4 self-inflicted regressions found and closed before merge.**

The self-inflicted regressions are worth naming, because they are the reason the round count is high:
a `verify` diagnostic that perturbed the `ROWS_DB_ENTRY_RESOLVES` work-counter invariant; hard asserts
added without a fixture probe; a fixture probe whose `read_dir(...).ok()?` early-returned *before* the
`CQLITE_REQUIRE_FIXTURES` assert (16/16 lanes passing **vacuously** on an absent `test_da`); and missing
end-to-end coverage of the #1968 implicit-first branch. Each was caught by a later round and pinned with a
mutation-checked test, so all four now have standing protection.

Follow-ups: **#3076** (batched nits from this PR), **#3032** (multi-component `da` fixture), **#3040**
(release note for the mis-rooted-writes compatibility break), **#3045** (writer block-0 canonicality),
**#3070** (gate/disk observability).

Issue #3048, opened 2026-07-27
row_build bench: run-to-run variance 7-26% at N=32/64 makes it unable to resolve a <10% effect (blocks #2901-class A/B)


pmcfadin:
Tooling debt created by #3027, which introduced `cqlite-core/benches/row_build_bench.rs` — the
row-conversion instrument #2901 named as its blocker.

## Problem

The bench is stable at N=8/16 but **not** at N=32/64. Two runs of the *identical* binary, clean idle box,
`taskset -c 2`, `--warm-up-time 2 --measurement-time 6`:

| bench id | run 1 | run 2 | spread |
|---|--:|--:|--:|
| `row_build/columns_8` | 2.084 M rows/s | 2.119 M | 1.6% |
| `row_build/columns_16` | 1.203 M | 1.177 M | 2.2% |
| `row_build/columns_32` | 771.4 k | 722.0 k | **6.8%** |
| `row_build/columns_64` | 300.1 k | 379.2 k | **26%** |
| `row_build/hashonly_8` | 3.442 M | 3.369 M | 2.1% |
| `row_build/hashonly_16` | 1.750 M | 1.803 M | 3.0% |
| `row_build/hashonly_32` | 1.021 M | 804.2 k | **21%** |
| `row_build/hashonly_64` | 560.4 k | 419.5 k | **25%** |

`columns_64` also showed a 34% spread *within* a single run: `[5.8253 ms 6.6655 ms 7.7911 ms]`.

## Why it matters

This is exactly the resolution needed for the #2901 question (is FxHashMap worth ~5%?). At N=32/64 the
noise is 4-5x the effect, so the instrument **cannot answer the question it was built for** at realistic
column counts. #3027 had to fall back to in-situ `perf` attribution instead and reported the end-to-end
A/B as **inconclusive**.

## Suspected cause

The noise tracks memory footprint, so it is contention/allocator-state sensitive rather than random:

- `iter_batched(..., BatchSize::LargeInput)` rebuilds 2,000 rows in the untimed setup closure each
  iteration. The setup is not *timed*, but its side effects — allocator state, heap fragmentation, cache
  pollution — land inside the following timed region.
- At N=64 the working set is large enough that this dominates.

The bench author measured <1.6% spread on a quiet detached worktree with `--measurement-time 3`, so the
configuration matters a lot — which is itself the problem: an instrument whose verdict depends on ambient
conditions is not yet trustworthy for a <10% decision.

## Suggested directions

[...]

Issue #3047, opened 2026-07-27
Hoist RowColumnResolution from per-partition to per-scan — 17.4% of on-CPU rebuilding scan-invariant state (residue of #1635 J1)


pmcfadin:
Found while profiling for #3027 (WS1 of #3023). Filed rather than fixed inline — see "Why not landed in #3027" below.

## The finding

`RowColumnResolution::build` is called **once per partition** on the streaming read path
(`cqlite-core/src/storage/sstable/reader/parsing/row_decoder/partition_driver.rs:179`). The code comment
says so explicitly:

```rust
// Issue #1046: per-PARTITION resolution build (this driver is re-entered
// once per partition by the sliding-window caller; allocations scale with
// partition count, not row count). Borrows header strings + schema.
let resolution = RowColumnResolution::build(schema, reader);
```

But the resolution depends only on `(schema, reader.header)` — **both invariant for the entire scan**.
Per column it allocates two `String`s via `to_lowercase()` on two *different* strings:

- `CellKind::from_type` lowercases the **schema** type (`row_decoder/cell_kind.rs:72`)
- `is_complex_column` lowercases the **header marshal** type (`row_decoder/udt.rs:1216`)

...plus a `HashMap<&str, &Column>` build, a `Vec<ColumnToParse>` allocation per row-kind, and an
`Arc::from` per Complex column.

## Measurement

`perf record -F 1999` over `read/scan_partition_dense_stream`, pinned, warm, Xeon 8488C:

| symbol | % on-CPU |
|---|--:|
| `RowColumnResolution::build` | 2.86 |
| its `Map<Filter<Iter<ColumnInfo>>>::next` | 2.60 |
| `<str>::to_lowercase` | 2.99 |
| `CellKind::from_type` | 1.80 |
| `is_complex_column` | 1.41 |
| `Vec<ColumnToParse>::from_iter` | 1.40 |
| `drop_glue::<Vec<ColumnToParse>>` | 1.32 |
| `hash_one::<&String>` | 1.50 |
| `make_hash::<str>` | 1.49 |
| **total** | **17.4%** |

## The caveat that must travel with that number

The fixture is `test_basic.simple_table` — **999 one-row partitions**. That is the *pathological* shape:
per-partition work amortizes over exactly one row. The cost falls linearly as rows/partition rises, so
on a table with 10-100 rows/partition this is ~1.7%-0.17%, not 17.4%.

[...]

Issue #2818, opened 2026-07-22
i4i cold-vs-warm server-direct `samply` profile


pmcfadin:
> **📋 EXECUTION PLAN 2026-07-26 — two arms, run the first one NOW (owner-directed).**
> Both IO fixes are on `main` but **unreleased**: #2876 (`bc14ff0ed`, scan/point read-plane split) and
> #2877 (`c02414152`, per-chunk read coalescing) are 27+ commits past the `v0.16.0` tag. The field team
> holds only the pre-fix image, so this issue is now a **two-arm A/B**, not a single profile.
>
> | Arm | Image | When | Expected |
> |---|---|---|---|
> | **1 — pre-fix (RUN NOW)** | existing `v0.16.0` / `0.16.1` | immediately, no blockers | reproduces #2866: `rareq-sz` ~4.4 KB, ~97% of reads in [4K,8K), ~27.7 MB/s single-stream |
> | **2 — post-fix** | `0.17-dev` (to be published from `main`) | after #2819 lands | read-size mass moves to **≥64 KB**; single-stream MB/s up from 27.7 |
>
> **Run arm 1 before the image exists.** It is not redundant with #2866: it establishes the baseline on
> **this** corpus with **this** method, so the arm-2 delta is apples-to-apples (#2866 measured a
> different corpus — ~13.5M partitions), and it shakes out the `bpftrace`/kernel tooling while nothing
> is blocked on the result.
>
> **`0.17-dev` publication:** `gh workflow run flight-image.yml --ref main -f image_tag=0.17-dev`
> → `ghcr.io/pmcfadin/cqlite-flight:0.17-dev`. Uses the free-form `image_tag` dispatch input, so it does
> NOT move `latest` and cannot disturb anything the field is running. Gated on **#2819** (data-plane
> sub-phase timers) landing first — without it the `stream` phase still folds the entire data plane into
> one histogram, which is exactly what made #2866's first analysis conclude "serialization-bound" and be
> wrong. #1883 / #2043 / #2874 / #2870 are NOT blockers (test-only, bench-only, CI-only, build-only).
>
> **Do NOT report an arm-2 result from the 0.16 image.** It predates both fixes, so "still ~4.4 KB"
> would read as "the fix failed" when the fix simply is not in that binary.
>
> **What the result gates:** whether #2820 / #2822 / #2823 / #2826 (parked in Backlog) get promoted, and
[...]

Issue #1930, opened 2026-07-04
flow-meta: one worker per machine — worker governs load, serial full gate, any-slug pre-claim check


pmcfadin:
**Owner decision 2026-07-04** (v0.13 Ready-queue push; two failures on one machine).

## Model
- **Exactly one flow-lead worker session per machine.** That worker owns the machine's Ready-queue throughput and is the SOLE machine-load authority: it fans out *implementation* to subagents (edits + `--lite` gates), but **serializes the full `agent-gate.sh` (concurrency = 1)** and caps heavy fan-out. Read-only reviews (rust-reviewer/spec-auditor) may overlap.
- **Cross-machine** coordination still uses the origin `issue-<N>-<slug>` branch lock; one-per-machine composes with multiple machines. The branch lock is no longer load-bearing *within* a machine.

## Evidence
1. **Gate-contention flake:** 2 concurrent full gates made `mixed_p99_bounded_by_k_times_baseline` (tail-latency, `cqlite-core/tests/tail_latency_harness.rs`) fail under CPU oversubscription. The #1825 cap prevents SIGKILL but NOT timing flakes (#1625 core-tests PASSED solo at 693s, FAILED at 87s alongside another gate). → full gates MUST be serial even though the cap allows N=2.
2. **Cross-session dup (#1632):** two same-machine sessions claimed #1632 with *different slugs* (`issue-1632-parser-hardening` vs `issue-1632-parser-hardening-bundle`) so the exact-slug push didn't collide. Fix: pre-claim check ANY `issue-<N>-*` branch (any slug).

## Acceptance
- [ ] `docs/development/pm-operating-loop.md`: single-worker-per-machine + serial full gate + any-slug pre-claim documented.
- [ ] `.claude/commands/worker.md`: same-machine single-worker preflight guard + any-slug pre-claim (`git ls-remote --heads origin 'issue-<N>-*'`) + full-gate concurrency=1 note.
- [ ] CLAUDE.md concurrency bullet updated for consistency.

Docs-only; no full gate (no source change). Routing: design/process → doctrine (no OpenSpec).

PR #3049, opened 2026-07-27
perf(#3027): row-decode CPU attribution + the row-conversion instrument (#2901 blocker)

### LINKED ISSUE #2901: L5: revisit FxHashMap for the per-row QueryRow.values map — needs a benchmark and a HashDoS answer first
Split out of #1883 on the owner's call. L5 was implemented there and then **deliberately reverted** before merge.

## What L5 is

Swap the per-row `QueryRow.values` map (`build_row_from_scan_cached`,
`cqlite-core/src/query/select_executor/row_build.rs`) from the default SipHash
`HashMap<Arc<str>, Value>` to `rustc_hash::FxHashMap`, to take SipHash off the row hot path.
Projected in the 0.17 throughput program (epic #2817, M4) at ~1.04x.

## Why it was reverted rather than shipped

Two things surfaced during implementation that the original design did not anticipate:

**1. It is a PUBLIC breaking API change, not a local one.** `row_values` is moved directly into
`QueryRow.values`, so the hasher cannot be changed without changing that public field's type. The
working implementation introduced `pub type RowValues = FxHashMap<Arc<str>, Value>` and rippled
through construction sites in **`cqlite-core`, `cqlite-flight` and `cqlite-cli`** (read paths —
`get(&str)`, `keys()`, iteration, serde — were all unaffected). Any downstream code naming
`HashMap<Arc<str>, Value>` breaks.

**2. It contradicts a written invariant.** `cqlite-core/Cargo.toml`:

```toml
# Fast non-cryptographic hasher (FxHashMap) for integer/digest-keyed hot-path
# maps only — NOT for maps exposed to untrusted string keys (issue #1590, E8).
rustc-hash = "1.1"
```

`QueryRow.values` is precisely a string-keyed map. Column names are interned from `TableSchema`,
but on the default read path that schema is reconstructed from the file's `Statistics.db`
serialization header — so for a **hostile** SSTable (a real threat model here; see the fuzzing
epic #1614) an attacker controls the key strings. FxHash is trivially collidable, giving O(N**2)
per-row insert behaviour with N attacker-chosen colliding column names.

**And the benefit was never measured.** #1883 ran no row-conversion benchmark, so the ~1.04x
remained a projection. Paying a public breaking change plus a HashDoS exposure for an unquantified
win was judged a bad trade.

## What must be true to land L5

1. **A measured benchmark** of the row conversion showing the actual hashing win (not the projection).
   Credit 1.0x if that is what the profile says.
2. **A HashDoS answer**, one of:
   - a bound on attacker-controlled column count that makes O(N**2) tolerable, written down; or
   - a guard (e.g. fall back to SipHash above a column-count threshold); or
   - a deliberate, documented amendment of the #1590 E8 invariant in `cqlite-core/Cargo.toml`.
3. **An API plan** for the breaking `QueryRow.values` type change (the `RowValues` alias approach
   worked cleanly and is the suggested shape), including a CHANGELOG entry.

## Useful prior art from #1883

- The alias approach (`pub type RowValues`) compiled cleanly across all three crates; the ripple is
  entirely mechanical (`HashMap::new()` -> `RowValues::default()`, plus explicit annotations).
- **Iteration-order audit came back clean**: every order-sensitive consumer already sorts or drives
  off `metadata.columns` (`result.rs` sorted fallback, `select_executor/execute.rs:305` sorted,
  arrow_columnar's name->indices routing, all CLI writers). Three sites emit raw map order into a
  `preserve_order` `serde_json::Map` and would go from per-process-random to deterministic-arbitrary
  — an improvement, but worth restating.
- The hasher is **alloc-neutral** and therefore invisible to #1883's allocation ratchet (measured:
  identical 41/273 counts either way). Any L5 attempt needs its own wiring evidence — #1883 used a
  `type_name_of_val(row.values.hasher())` assertion on the public conversion output, which correctly
  failed when the alias was reverted.

### LINKED ISSUE #3027: WS1: row-decode CPU (26.7% of on-CPU) — attribute and cut; includes re-scoping #2901 FxHashMap as non-breaking
**#3023 WS1** — the #1 measured warm lever: row decode at **26.7% of on-CPU** (up from 5.0% in Arm 1 as other costs were removed).

## Scope

1. **Attribute the 26.7% to named functions.** `perf` on Arm 2 names `parse_row_data`, `entry_to_row`, utf8 validation, and **sip-hash** inside this bucket. Run `perf record -F 999 --call-graph=dwarf` locally over `benches/read.rs` + `benches/decode_bench.rs` and produce a function-level breakdown with cycles/row per function.
2. **The one concrete sub-lever already isolated — #2901 (L5 FxHashMap).** `cqlite-core/src/query/select_executor/row_build.rs:246` still builds the per-row map with SipHash:
   ```rust
   let mut row_values: HashMap<Arc<str>, Value> = HashMap::with_capacity(cells.len() + pk_hint);
   ```
   `FxHashMap` is **already the house pattern three places over in the same module** — `aggregation.rs:54`, `select_executor/mod.rs:173`, `stream_agg.rs:96` (all from #1590/#1817), and `rustc-hash = "1.1"` is already a `cqlite-core` dependency.

   **Read #2901 and #1883's close-out before touching this.** L5 was implemented and then **reverted before merge on the owner's decision**: it was a public breaking API change and it contradicted the `rustc-hash` invariant re: untrusted string keys (hash-DoS), for an *unmeasured* benefit. Two of those three objections are now answerable:
   - **Measured benefit:** Arm 2 puts sip-hash inside the 26.7% bucket. Quantify it specifically.
   - **Non-breaking surface:** scope the swap so the map type does not appear in any `pub` signature (internal-only, or a newtype). If it cannot be made non-breaking, say so and stop.
   - **Untrusted keys:** the keys here are **column names from the schema**, not user data — state that explicitly with evidence, or the objection stands.
3. **Redundant utf8 validation** — look for values validated more than once across `parse_row_data` → `entry_to_row` → Arrow encode, and for `str` validation on bytes that are immediately re-encoded.
4. **Per-cell re-parsing** — look for the same cell bytes being walked twice (bounds check then decode, or decode then re-decode for type dispatch).

## Acceptance criteria

- Function-level attribution of the 26.7% bucket, each with **cycles/row** and **bytes-of-memory-traffic/row**.
- Each candidate fix lands with a measured before/after in those same units — **not CPU% alone** (see the reporting contract: memory bandwidth binds first, so a CPU%-only win can be illusory).
- **Stopping condition: row decode < 10% of on-CPU AND ≥ 400k rows/s/phys-core warm.**
- Jointly with WS2 (#3028): **T1 = 600k rows/s/phys-core warm.**
- #2901 either lands as an internal non-breaking swap with the three objections answered, or is closed with the reason recorded. Do not merge a breaking public-surface change.
- No regression on the #1883 alloc ratchet (`cqlite-core/src/query/select_executor/row_build_alloc_budget_test.rs`).

## Notes

- **Do not re-litigate columnar/DataFusion** — #2605 owns that question.
- The `Value` materialize step is ~1,400 B of the ~4,408 B/row memory traffic — the single largest term. If a change can avoid materializing `Value` at all (borrow from the decompressed chunk), that is T2 territory and worth a note even if out of scope here.

---

### The goal
**T1 = 600k rows/s per PHYSICAL core, warm** (2.5× today's measured 238k). Unit is per-*physical*-core because #2818 Arm 2 §7 proved the C(N) plateau tracks physical cores (xlarge N=2, 2xlarge N=4) at a stable ~216–238k each — the only figure that composes to a pod.

| Tier | rows/s/phys-core | vs today | Basis |
|---|--:|--:|---|
| T0 today | 238k | 1.0× | measured |
| **T1 TARGET** | **600k** | **2.5×** | kill decode (26.7%) + allocator (19.6%) |
| T2 stretch | 1.2M | 5.0× | + zero-copy decode, fused decode→Arrow |
| T3 roofline | ~1.8M | 7.6× | **memory-bandwidth** bound |

Today = **12,597 cycles/row** vs a first-principles budget of **~929** → we are **13.6× off**. The binding roofline is **memory bandwidth (~1.8M rows/s/core at ~4.4 KB traffic/row, 8 GB/s/core)**, not cycles (~3.2M).

### Ground truth from #2818 (do NOT re-measure)
i4i, compressed corpus, LZ4 `chunk_length=16384`, server-direct, `v0.16.0` → `0.17-dev ed3ac7ef`:

| Fact | Value |
|---|---|
| single-stream warm | 199.3 MB/s / 291.7k rows/s |
| **row decode** | **26.7%** on-CPU (was 5.0%) |
| **allocator** | **19.6%** on-CPU |
| k-way merge | 6.4% — *small; #2822/#2823 closed on this evidence* |
| LZ4 decompress | 0.5% on merge thread (#2876 moved it to the producer; `stream_decompress` flat ~150 ms) |
| block reads | **99.3% in [4K,8K)**, `rareq-sz` 4.4 KB — unmoved by #2877 |
| `stream_cold_fault` | 4437.9 ms cold → 105.8 ms warm (**42× collapse**) |
| `stream_merge` / `stream_encode` | 7351 / 1868 ms, **identical cold vs warm** |
| disk knobs | none help; forcing mmap is 40% *slower* |

**Corollary: disk is NOT the warm bound.** Warm scan 11.9 s vs cold 15.9 s; if cold IO were free, cold improves ~1.39× and **warm improves 1.00×**. The tiny reads are a cold-start tax on this shape.

**Row width correction:** this corpus is **~690 B/row uncompressed** (197 B compressed, ~3.5× LZ4; Arrow wire 716 B). Earlier "narrow rows ~196 B" statements quoted the *compressed* size. These are medium rows.

### Reporting contract (non-negotiable)
1. **Report every win as cycles/row AND bytes-of-memory-traffic/row**, not just CPU%. Memory bandwidth binds before cycles, so a change that cuts CPU% without cutting traffic will stall — the percentage looks like progress while rows/s does not move.
2. **Report rows/s per physical core, warm**, as the headline. Include cold separately.
3. **A null result is a result.** #2818 Arm 2's flat read histogram is the standard: it was predicted, reported honestly, and redirected the whole program. Do not manufacture a win.
4. **Post findings as a comment on the umbrella #3023**, not only here — that issue is the single reporting surface for this exploration.
5. Every claimed multiplier must be **measured, not modeled**.

### Tooling gotchas (already paid for — do not rediscover)
- **samply records 0 samples** on HWE `7.0.0-*-aws` kernels (hit in BOTH M0 arms). Use `perf record -F 999 --call-graph=dwarf`.
- **bcc `xfsslower` will not compile** there (bcc 0.35, `struct filename` static-assert). Use `bpftrace` on `block:block_rq_complete`.
- **Never divide** `/proc/<pid>/io` `read_bytes` by `syscr` — different populations (fault-driven vs syscall). Report separately. This produced a bogus "~59 KB/syscall" in #2866.
- Local harness: self-built unstripped `cqlite-flight --features observability` + in-repo `tools/flight-loadgen` (its `Shape::Point` correctly sets token bounds). Benches: `benches/read.rs`, `decode_bench.rs`, `concurrent_scan.rs`, `partition_lookup.rs`, `open.rs`.
- Corpora: generate locally via `test-data/scripts/` so shape is a controlled variable. Set `CQLITE_DATASETS_ROOT`.

### Method doctrine
**A test written from the implementation cannot falsify the implementation.** Two bugs this cycle (#2866 IO trace, #3002 BTI root base) were invisible to green tests because the tests encoded CQLite's own assumption rather than Cassandra's behavior. Where Cassandra 5.0.8 source is the authority, derive the expectation from Cassandra.

### Process
One issue ↔ one claim ref ↔ one worktree/branch ↔ one PR (1:1:1:1). `bash scripts/flow/claim.sh claim <N>` FIRST. Run the full gate with `CQLITE_GATE_MAX_CONCURRENCY=1`. Parent: #3023. Program epic: #2817.

pmcfadin:
Closes #3027. WS1 of #3023. Full report: `docs/reports/issue-3027-row-decode-attribution.md`.

Measurement-driven work — the deliverable is a defensible number, not a diff. **There is no
production-code change in this PR**: both candidate optimizations were null results and were reverted
rather than shipped as churn.

## Headline: the WS1 premise is partially refuted

Row decode is the largest **CPU** consumer but it is **traffic-light**. Sampling on the memory event
directly (`perf record -e l2_lines_in.all`) instead of inferring traffic from CPU share:

| Function | cycles % | cycles/row | **traffic %** | **B/row** | ratio |
|---|--:|--:|--:|--:|--:|
| `parse_row_data_with_offset_impl` | **9.64** (#1) | 1,088 | 4.20 (#6) | 523 | **0.44** |
| `memory::estimate_value_size` | 6.35 | 716 | **11.22** (#1) | **1,398** | **1.77** |
| `drop_glue::<QueryRow>` | 4.63 | 522 | **7.62** | 949 | **1.65** |
| `Value::into_owned` | 6.22 | 702 | **7.59** | 945 | 1.22 |
| `sip::Hasher::write` | 2.72 | 307 | 2.45 | 305 | 0.90 |

The decoder is **compute-bound**; the `QueryRow` materialize -> walk -> free lifecycle is
**traffic-bound** (26.4% of L2-fill traffic vs 17.2% of cycles = 3,292 B/row of 12,456 B/row). Since
memory bandwidth binds before cycles, **cutting decode CPU cannot reach T1** — the lever is T2's "avoid
materializing `Value` at all." `estimate_value_size` alone is 1,398 B/row, independently reproducing
#3023's "~1,400 B is the `Value` materialize step."

## Baseline and corroboration

`read/full_scan`, Xeon 8488C, 1 physical core, warm, +-0.2%: **314.77 k rows/s**, **11,283 cycles/row**
(vs #2818's 12,597), **12,456 B/row** L2-fill traffic. Buckets corroborate #2818: decode **22.1%** (vs
26.7%), allocator **20.0%** (vs 19.6%), LZ4 **0.9%** (vs 0.5%).

## What ships

- **`cqlite-core/benches/row_build_bench.rs`** — the row-conversion instrument #2901 named as its
  blocker (*"#1883 ran no row-conversion benchmark, so the ~1.04x remained a projection"*).
  `row_build/columns_{8,16,32,64}` drives the real public `build_row_from_scan_cached`;
  `row_build/hashonly_{N}` isolates map construction + SipHash. Carries a **compile-time pin** so that
  changing `QueryRow.values`' hasher **breaks the bench build** rather than silently making the two
  groups measure different hashers.
- **`docs/reports/issue-3027-row-decode-attribution.md`** — full attribution + all null results and confounds.

## #2901 disposition — 1 answered, 1 refuted, 1 stands

- **Measured benefit — ANSWERED.** SipHash = **4.94% of on-CPU / 557 cycles/row** in situ (8.12% on the
  streaming profile); map+SipHash = **57-80% of the entire row conversion**. Materially more than 1.04x.
- **Non-breaking surface — REFUTED BY CONSTRUCTION.** `QueryRow.values` is a **`pub` field** whose type
  *is* the map (`result.rs:77`); `QueryRow::with_interned_values` is a **`pub fn`** taking it
  (`result.rs:595`). A newtype cannot bridge two `HashMap`s with different `S` via `Deref`. Compiling the
  swap surfaced **6 construction sites in `cqlite-core` alone**. Per #3027's own instruction — stop.
- **Untrusted keys — OBJECTION STANDS.** No bound on serialization-header column count exists (the only
  nearby limit, `row_cell_state_machine.rs:710`, is a per-row cell count on a different path). #1590 E8 applies.

**Owner decision requested** (not taken unilaterally): close #2901, or re-scope it to a positional
`Vec` + shared name->index table — which removes the hashing *and* the per-row map allocation *and*
sidesteps hash-DoS, and is T2-aligned.

## Null results (both reported, neither shipped)

- **Lazy `row_has_non_key_cell`**: no measurable change (317.7k/313.5k vs 314.8k). `.any()`
  short-circuits on the first non-key cell, so the real cost is ~1 string compare/row, not ~90. Reverted.
- **End-to-end FxHashMap A/B: INCONCLUSIVE.** Two runs of the *identical* binary differ 26% at
  `columns_64`; the `hashonly_*` control — which the swap cannot affect — moved 8-25%. Filed #3048.

## Stopping condition

**NOT met**: decode 22.1% (target <10%), 314.8 k rows/s/phys-core (target >=400 k). The attribution
explains why — remaining levers are either traffic-bound/T2-shaped or gated on a corpus shape this box
cannot produce.

## Confounds recorded so they are not rediscovered

- **DRAM-level PMU counters are not virtualized on EC2**: `longest_lat_cache.miss`, `offcore_requests.*`,
  `mem_load_retired.l3_miss`, `cache-references` all read exactly 0. Only `l2_lines_in.all` works — all
  traffic figures are L2-fill, not DRAM.
- **Do not pin the streaming bench to one core** for throughput: 5.27 s of 12.43 s goes to `sys`.
- **I contaminated part of my own A/B** by measuring in the same worktree as a working subagent. #1930
  "one worker per machine" is not only about gate concurrency.

## Follow-ups filed

- **#3047** — `RowColumnResolution` rebuilt per-partition though it is scan-invariant: **17.4% of on-CPU**
  on the pathological 1-row/partition shape. Deliberately not rushed — with no Docker there is no
  realistic-shape corpus, so its true value could only be *modeled*, which the #3023 contract forbids.
- **#3048** — `row_build` bench variance.

## Verification

- `--lite`: **PASS** (`file-size`, `fmt`, `clippy`, `roborev-lints`, `scoped-tests`), `tree-integrity: PASS`.
- #1883 alloc ratchet: unaffected — no production code changed.

### DIFF
--- a/cqlite-core/Cargo.toml
+++ b/cqlite-core/Cargo.toml
@@ -309,6 +309,18 @@ name = "decode"
 path = "benches/decode_bench.rs"
 harness = false
 
+# Row-conversion (row-build) CPU instrument (issue #3027; unblocks #2901). Two
+# groups over N ∈ {8,16,32,64} columns: `row_build/columns_<N>` drives the REAL
+# public `build_row_from_scan_cached` over 2,000 synthetic partition-grouped rows
+# with one shared PartitionKeyCache, and `row_build/hashonly_<N>` isolates the
+# HashMap-construction + SipHash term. Fully synthetic (no dataset fixture), needs
+# only the default `state_machine` feature. NOT a perf-gate entry — an instrument
+# whose absolute rows/sec is machine-dependent.
+[[bench]]
+name = "row_build"
+path = "benches/row_build_bench.rs"
+harness = false
+
 [[bench]]
 name = "write"
 harness = false
--- a/cqlite-core/benches/row_build_bench.rs
+++ b/cqlite-core/benches/row_build_bench.rs
@@ -1,0 +1,632 @@
+//! Row-conversion (row-build) CPU micro-benchmarks (issue #3027; the missing
+//! instrument that blocked issue #2901).
+//!
+//! # Why this target exists
+//!
+//! Issue #2901 was blocked because **#1883 ran no row-conversion benchmark**, so
+//! the projected ~1.04x speedup of a `HashMap`-construction change stayed a
+//! projection: nothing in the tree measured
+//! [`build_row_from_scan_cached`][bs] — the per-row `(RowKey, ScanRow) →
+//! `QueryRow`` conversion every scan pays once per emitted row. The existing
+//! `read/*` and `decode/*` benches measure the layers *below* it (open → seek →
+//! chunk decompress → cell decode); a row-conversion regression hides inside
+//! their I/O. This target isolates the conversion itself.
+//!
+//! [bs]: cqlite_core::query::build_row_from_scan_cached
+//!
+//! # What is measured
+//!
+//! Two criterion groups, both parameterized over the column count `N ∈ {8, 16,
+//! 32, 64}` and both reporting **rows/sec** via `Throughput::Elements(2_000)`:
+//!
+//! - **`row_build/columns_<N>`** — the REAL public function
+//!   `build_row_from_scan_cached(key, row, &[] /* SELECT * */, Some(&schema),
+//!   &mut pk_cache)`, driven over 2,000 prepared rows through ONE shared
+//!   [`PartitionKeyCache`], exactly as the scan loop calls it. This is the whole
+//!   per-row cost: map pre-size, `Arc<str>` name moves, `Value::into_owned()`
+//!   retention copies, and the memoized partition-key reconstruct.
+//! - **`row_build/hashonly_<N>`** — the SAME prepared names/values, but the timed
+//!   routine does ONLY `HashMap::with_capacity(n)` + `n` inserts. No partition-key
+//!   reconstruct, no `into_owned()`, no projection closure. This isolates the
+//!   *map construction + SipHash* term, which is the precise number #2901 needs:
+//!   `columns_<N> − hashonly_<N>` is everything else in the conversion.
+//!
+//! The bench is **NOT** a `benches/perf-gate.json` entry: it is an instrument, and
+//! its absolute rows/sec is machine-dependent. Its value is the ratio between the
+//! two groups (and between column counts) on ONE machine, in ONE run.
+//!
+//! # Fidelity of the synthetic input
+//!
+//! No dataset fixture is used — the input is fully synthetic so the measurement is
+//! independent of any corpus (and so it runs without `CQLITE_DATASETS_ROOT`). The
+//! synthesis deliberately reproduces the shape the V5 decoder hands to the
+//! conversion:
+//!
+//! - **Column names** are realistic Cassandra-style identifiers 16–19 chars long
+//!   (`col_metric_value_07`, `col_session_uuid_12`, …), interned ONCE into shared
+//!   `Arc<str>` handles — name LENGTH is what SipHash consumes, so `a`/`b`/`c`
+//!   would understate the map term.
+//! - **Values** cycle through a realistic mix — `text` (24 B), `int`, `bigint`,
+//!   `uuid`, `timestamp`, `double`, `boolean` (~9.9 B of payload per column, i.e.
+//!   ~79 B/row at N=8 up to ~631 B/row at N=64; the exact per-row average is
+//!   printed at setup).
+//! - **Text payloads are slices of ONE shared backing `Bytes`**, mimicking a
+//!   decompressed chunk. That matters: `Value::into_owned()` (the #1644 D2
+//!   retention boundary inside the conversion) copies a shared payload, so a
+//!   privately-allocated text value would measure a cheaper path than production.
+//! - **Rows are partition-grouped, 10 rows per partition** (200 partitions over
+//!   the 2,000 rows), so the `PartitionKeyCache` sees the production-realistic
+//!   9-hits-per-miss ratio rather than an all-hit or all-miss extreme.
+//!
+//! # What is and is not inside the timed region
+//!
+//! **Input construction is OUTSIDE.** `build_row_from_scan_cached` CONSUMES its
+//! `ScanRow`, so each measured iteration needs fresh inputs; they are rebuilt by
+//! `iter_batched`'s **untimed setup** closure (`BatchSize::LargeInput`). No input
+//! clone is ever measured.
+//!
+//! **Each produced row is consumed and dropped INSIDE, one at a time.** The
+//! routine accumulates a cheap `values.len()` checksum and lets each `QueryRow`
+//! drop before building the next, so the live working set is one row — a streamed
+//! scan's lifetime. The obvious alternative (push all 2,000 outputs into a `Vec`
+//! so their drops fall outside the timer) was measured and REJECTED: retaining the
+//! batch makes the timed region touch ~14 MB of fresh map memory at N=64, and that
+//! memory-bandwidth cliff swamps the CPU term this instrument exists to isolate.
+//! Measured on the same machine, retaining the batch cost 2 790 ns/row for
+//! `hashonly_64` vs 1 869 ns/row when dropping per row (+49%), and broke the
+//! linear scaling — it read out as "map construction = 98% of the conversion at
+//! N=64", an artifact, where the per-row-drop design reads a stable ~72–80% across
+//! all four column counts. The trade is that map DEALLOCATION is now inside the
+//! measurement; that is real per-row work the conversion's caller pays either way,
+//! and both groups pay it identically, so the `columns_<N> − hashonly_<N>` delta
+//! stays clean.
+//!
+//! **Wiring guard over the WHOLE pass.** The checksum both routines return must
+//! equal `2_000 × (N + 1)`, asserted once per bench id, so a pass that dropped or
+//! short-built rows fails loudly instead of measuring less work than advertised.
+//!
+//! # Determinism
+//!
+//! Every name, value, and key is a pure function of `(row_index, column_index)` —
+//! no wall-clock input, no RNG (seeded or otherwise), so the workload is
+//! byte-identical on every run and machine. There are no wall-clock threshold
+//! asserts anywhere; the only `Instant` use is the best-effort ledger pass, which
+//! RECORDS a number and never asserts on one.
+//!
+//! # Gating
+//!
+//! Needs only `state_machine` (the feature that gates `cqlite_core::query`, where
+//! `build_row_from_scan_cached` + `PartitionKeyCache` are re-exported) — it is a
+//! default feature. `fixtures/mod.rs` is deliberately NOT included: this target
+//! opens no SSTable and needs no RNG, so including it would only add compile
+//! cost. Without `state_machine` the target compiles to an empty-but-valid
+//! criterion main (mirrors `read.rs` / `decode_bench.rs`).
+//!
+//! Reproduce:
+//! ```text
+//! cargo bench -p cqlite-core --bench row_build
+//! ```
+
+use criterion::{criterion_group, criterion_main};
+
+#[path = "bench_ledger/mod.rs"]
+mod bench_ledger;
+
+#[path = "profiling/mod.rs"]
+mod profiling;
+
+// ---------------------------------------------------------------------------
+// Real benches (need `state_machine` for cqlite_core::query)
+// ---------------------------------------------------------------------------
+
+#[cfg(feature = "state_machine")]
+mod row_build_impl {
+    use super::bench_ledger;
+    use bytes::Bytes;
+    use cqlite_core::query::{build_row_from_scan_cached, PartitionKeyCache, QueryRow};
+    use cqlite_core::schema::{Column, KeyColumn, TableSchema};
+    use cqlite_core::types::{RowKey, ScanRow, Value};
+    use criterion::{black_box, BatchSize, Criterion, Throughput};
+    use std::collections::HashMap;
+    use std::sync::Arc;
+
+    /// The map type the `hashonly_<N>` group constructs. It MUST be the same type
+    /// `QueryRow::values` is — see [`hashonly_map_type_is_pinned_to_query_row`].
+    type ValuesMap = HashMap<Arc<str>, Value>;
+
+    /// **Compile-time pin (issue #3027).** `hashonly_<N>` is only interpretable if
+    /// it builds the SAME map type — and therefore runs the SAME hasher — as the
+    /// real conversion: if `QueryRow::values` switched hasher (e.g. SipHash →
+    /// FxHash) while [`ValuesMap`] did not, the two groups would hash differently
+    /// and `columns_<N> − hashonly_<N>` would be meaningless while still producing
+    /// confident-looking numbers.
+    ///
+    /// This function does nothing at runtime; it exists so that such a change
+    /// BREAKS THE BUILD here (a loud, actionable compile error) instead of
+    /// silently invalidating the instrument. If it fails to compile, update
+    /// [`ValuesMap`] to `QueryRow::values`' new type — do not delete this pin.
+    #[allow(dead_code)]
+    fn hashonly_map_type_is_pinned_to_query_row(row: QueryRow) -> ValuesMap {
+        row.values
+    }
+
+    /// Rows converted per measured iteration (the `Throughput::Elements` unit, so
+    /// criterion reports rows/sec directly).
+    const ROWS: usize = 2_000;
+
+    /// Rows per partition — 10 gives the shared [`PartitionKeyCache`] the
+    /// realistic 9 hits per miss (200 distinct partitions across [`ROWS`]).
+    const ROWS_PER_PARTITION: usize = 10;
+
+    /// Regular-column counts benched (the `_<N>` suffix of each bench id).
+    const COLUMN_COUNTS: [usize; 4] = [8, 16, 32, 64];
+
+    /// Byte length of every synthesized `text` payload.
+    const TEXT_LEN: usize = 24;
+
+    /// The single TEXT partition-key column. Single-component ⇒ the raw row key
+    /// bytes ARE the value (no length prefix), which is the common Cassandra
+    /// shape and the one `decode_partition_key_columns` handles on the hot path.
+    const PK_COLUMN: &str = "pk_device_id";
+
+    /// The repeating value mix applied across a row's columns (`col_index % 7`).
+    #[derive(Clone, Copy)]
+    enum Kind {
+        Text,
+        Int,
+        BigInt,
+        Uuid,
+        Timestamp,
+        Double,
+        Boolean,
+    }
+
+    const KINDS: [Kind; 7] = [
+        Kind::Text,
+        Kind::Int,
+        Kind::BigInt,
+        Kind::Uuid,
+        Kind::Timestamp,
+        Kind::Double,
+        Kind::Boolean,
+    ];
+
+    impl Kind {
+        /// The CQL type string recorded in the schema for this kind.
+        fn cql_type(self) -> &'static str {
+            match self {
+                Kind::Text => "text",
+                Kind::Int => "int",
+                Kind::BigInt => "bigint",
+                Kind::Uuid => "uuid",
+                Kind::Timestamp => "timestamp",
+                Kind::Double => "double",
+                Kind::Boolean => "boolean",
+            }
+        }
+
+        /// On-wire payload width, used only to report the per-row average.
+        fn payload_bytes(self) -> usize {
+            match self {
+                Kind::Text => TEXT_LEN,
+                Kind::Int => 4,
+                Kind::BigInt => 8,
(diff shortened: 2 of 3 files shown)


pmcfadin:
## flow-closer endgame status — MERGE HELD (not armed)

**Gate of record: PASS.** Full `scripts/agent-gate.sh` at `1ee870c` (rebased onto `origin/main` `5e1a76a` first, so the gate postdates the final rebase). All 29 components PASS; `tree-integrity: PASS` (`tree-start == tree-end == 1ee870cfd6b9`, `dirty: no`); `run-id: /tmp/agent-gate.ISWyaL` verified against the run launched; real fixtures (`144 Data.db files under /data/datasets`), so no #2078 fail-closed. `PREMERGE: OK 1ee870c`; no `HOLD:` order; `required` check green.

**Auto-merge deliberately NOT armed.** Two owner calls, neither of which the closer may decide:

1. **roborev — AUTH OUTAGE, not a clean review.** Both agents fail on credentials, not findings:
   - repo-pinned `--agent claude-code` → `Failed to authenticate: OAuth session expired and could not be refreshed`
   - `--agent codex --model gpt-5.6-sol` → `401 Unauthorized: Missing bearer or basic authentication` (no `~/.codex/auth.json` on this box)

   Definition of done requires roborev clean; that condition cannot be evaluated here. An outage is not "clean", and waiving the stage is not the closer's call.

2. **Two acceptance criteria are not satisfied** (verified directly against the issue body — this is oracle/measurement-routed, no OpenSpec change exists, so the C intent audit was correctly skipped):
   - *Stopping condition* (`row decode < 10%` on-CPU **and** `>= 400k rows/s/phys-core`) — **NOT met** (22.1%, 314.8k). Both candidate optimizations were null results and were reverted; the diff carries **no production-code change**.
   - *#2901 disposition* — the reason is **recorded** (report §7: benefit answered, non-breaking **refuted by construction** via the `pub` `QueryRow.values` field + `pub fn with_interned_values`, hash-DoS objection **stands**), but close-vs-re-scope is explicitly left as an owner decision.

   Whether an attribution report + instrument is mergeable with its perf stopping condition unmet is a scope/product judgement for the owner.

**Ready to merge the moment the owner clears both.** rust-reviewer found no blockers; its three doc-accuracy should-fixes are fixed in the tip commit.

pmcfadin:
## Endgame complete — merge HELD, needs owner

Everything except the merge is done. Holding on **two independent triggers**, neither of which I will waive autonomously.

### Verified green

| Stage | Result |
|---|--:|
| **Full `agent-gate.sh` (the gate of record)** | **PASS** — all 29 components |
| `tree-integrity` | **PASS** (start == end `1ee870cfd6b9`, `dirty: no`) |
| gate provenance (#2874/#2926) | exit 0, `run-id` matches the launch, no `.integrity-fail.*` sibling, 144 Data.db under `/data/datasets` |
| `--lite` | PASS (x2) |
| `rust-reviewer` | **no blockers**; 3 should-fix doc-accuracy items all fixed in the second commit |
| CI `required` | SUCCESS, mergeState CLEAN |
| C (intent audit) | **n/a** — oracle/measurement-routed, no `openspec/changes/*` exists |
| #1883 alloc ratchet | unaffected — no production code in the diff |

Certified SHA is **`1ee870c`** (rebased onto `origin/main` `5e1a76a` *before* gating, so the gate certifies the post-rebase tree). `premerge-assert.sh 3049 1ee870c` → OK. No `HOLD:` order present. **Auto-merge deliberately NOT armed.**

### Hold 1 — roborev could not run (auth outage, not a clean result)

Both agents fail on credentials, on this box:
- repo-pinned `claude-code` → `Failed to authenticate: OAuth session expired and could not be refreshed`
- `codex --model gpt-5.6-sol` → `401 Unauthorized: Missing bearer` (no `~/.codex/auth.json`)

That is **0 findings obtained**, which is not the same as "clean". The definition of done requires roborev clean, and that condition cannot currently be evaluated. Fixing this needs re-auth on the worker or a different box — it is an environment problem, not a code problem.

(For scope: `.roborev.toml` sets `exclude_patterns = ['docs/**', '*.md']`, so roborev would have covered `row_build_bench.rs` + `Cargo.toml` only — the report doc is outside its scope regardless.)

### Hold 2 — 2 of 5 acceptance criteria unmet (independent of Hold 1)

This one stands even if credentials are restored.

| Criterion | Status |
|---|---|
| Function-level attribution, cycles/row **and** bytes-traffic/row | **Met** |
| Each candidate fix measured before/after in those units | **Met** — both null |
| Stopping condition: decode <10% on-CPU **and** >=400k rows/s/phys-core | **NOT met** (22.1%, 314.8k) |
| #2901 lands non-breaking, or closed with reason recorded | **Recorded** — close-vs-re-scope is an owner call |
| No regression on #1883 alloc ratchet | **Met** |

The question for the owner: **is an attribution report + a reusable measurement instrument a legitimate deliverable when the perf target the issue was chartered to hit was not reached and both candidate fixes were null?**

The case for shipping: the null results and the #2901 refutation are real, reusable findings, and the headline — that decode is compute-bound while the traffic sits in the `QueryRow` materialize→walk→free lifecycle — **redirects WS1 away from a lever that cannot reach T1**. Suppressing that costs more than shipping it. But that is a product judgement, not mine to make.

### #2901 disposition (also an owner call)

- Measured benefit — **answered** (SipHash 4.94% on-CPU / 557 cycles/row in situ).
- Non-breaking — **refuted by construction**: `QueryRow.values` is a `pub` field of the map type and `QueryRow::with_interned_values` a `pub fn` taking it. Per #3027's own "if it cannot be made non-breaking, say so and stop" — the swap was **not** made.
- Untrusted keys — **objection stands**; no bound on serialization-header column count exists.

Recommendation: close #2901, or re-scope to a positional `Vec` + shared name->index table (removes hashing *and* the per-row map alloc *and* sidesteps hash-DoS; T2-aligned).

Follow-ups filed: #3047 (per-partition `RowColumnResolution` rebuild), #3048 (bench variance).

pmcfadin:
## Hold 1 CLEARED — roborev ran. One Medium finding, refuted with evidence.

### The auth outage is resolved (it was three distinct failures, not one)

| Attempt | Failure | Actual cause |
|---|---|---|
| `--agent codex --model gpt-5.6-sol` (earlier) | `401 Unauthorized: Missing bearer` | no `~/.codex/auth.json` — real auth outage |
| `--agent claude-code` (repo default) | `OAuth session expired and could not be refreshed` | **still broken** on this worker |
| `--agent codex --model gpt-5.6-sol` (after auth restored) | `400: The 'gpt-5.6-sol' model requires a newer version of Codex` | auth fine; **`codex-cli 0.142.5` predates the pinned model** |
| **`--agent codex --model gpt-5.5`** | — | **ran to completion** |

Two doctrine items fall out of this, both worth a fix:
1. **CLAUDE.md's pinned `--model gpt-5.6-sol` is stale for this box's CLI.** 0.142.5 advertises `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, `codex-auto-review`. CLAUDE.md already warns that `--agent codex` alone inherits `opus` and 400s "in a way that looks like an outage" — this is a *second* 400-that-looks-like-an-outage from the same flag. Either bump the CLI or re-pin the model.
2. **The repo-default reviewer is the broken one.** `.roborev.toml` pins `agent = 'claude-code'`, so any roborev invocation without explicit overrides fails on this worker.

### The finding — and why I am refuting rather than fixing it

> **Medium** · `cqlite-core/benches/row_build_bench.rs:582` — "reuses one `PartitionKeyCache` across all
> Criterion iterations, so after the first pass the cache is fully warm and the benchmark no longer has
> the documented 9-hits-per-miss partition grouping."

**`PartitionKeyCache` is a SINGLE-ENTRY cache.** `row_build.rs:124-132`:

```rust
#[derive(Default)]
pub struct PartitionKeyCache {
    /// Raw partition-key bytes + partition-key schema fingerprint of the LAST
    /// decoded partition and its decoded `(interned name, value)` columns
    decoded: Option<DecodedPartitionKey>,
}
```

One `Option`, holding only the **last** partition. It cannot become "fully warm" across the bench's 200
distinct partitions (`row_build_bench.rs:361`, `partition = row / ROWS_PER_PARTITION`, 2,000 rows at 10
rows/partition).

Per pass, in order, the hit/miss sequence is therefore invariant:
- row 1 of partition *k* → **MISS** (cache holds partition *k-1*)
- rows 2-10 of partition *k* → **9 HITS**

= exactly **200 misses + 1,800 hits on every iteration**, first or thousandth — precisely the documented
9-hits-per-miss ratio. Carrying the cache across iterations can affect at most the very first row of a
pass, and even there it still misses (the cache holds partition 199 from the prior pass; the new pass
starts at partition 0). The measurable difference is **zero**.

Separately, one shared cache is **what the real scan loop does**, and is what the function's own doc
prescribes — `build_row_from_scan_cached`'s rustdoc: *"Loops over many rows should call
`build_row_from_scan_cached` with a shared cache to hoist the per-partition decode (Issue #1817)."*
Adopting the suggested fresh-cache-per-batch would be churn **and** slightly less faithful to the
calling convention the bench exists to model.

The finding would be correct against a memoize-every-partition cache. It is not correct against this
one. Recording the refutation rather than applying it; happy to be overruled with a counter-argument.

### Where that leaves the merge

**Hold 1 (roborev) is cleared.** Gate of record PASS + roborev ran with its sole finding refuted.

**Hold 2 still stands and is unaffected by any of the above** — 2 of 5 acceptance criteria unmet: the
stopping condition (decode <10% on-CPU **and** >=400k rows/s/phys-core) was **NOT met** (22.1%, 314.8k),
and #2901's close-vs-re-scope is explicitly an owner call. An unmet requirement is an escalate-and-hold
trigger, so auto-merge remains **unarmed** pending your decision.

Issue #3056, opened 2026-07-27
specialist agent definitions carry fabricated + stale facts — test-validator's exit-code map, retired 33-table universe, dead module paths, unrunnable external-clone instructions


pmcfadin:
## Problem

`.claude/agents/**` and `.claude/skills/**` are the **instruction surface the lead spawns specialists
against**. A fabricated fact here does not merely fail — it **MISDIRECTS triage**, sending a subagent
down a wrong diagnosis *with confidence*. Three items below are outright fabrications; the rest are
doctrine drift, dead paths, or literally unrunnable commands.

**All items VERIFIED against `origin/main`** (via `git show origin/main:<path>` / `git cat-file -e` /
`git ls-remote`) unless marked otherwise. Root tree is ~30 commits behind origin/main — verify there,
not in the working tree.

---

## A. FABRICATED / MISDIRECTING (fix first)

1. **`.claude/agents/test-validator.md:77-79` — the CLI exit-code map is FABRICATED.**
   It claims `Exit 3: Parsing error (static columns, frozen types)` and
   `Exit 5: Partition key extraction failure`. TRUTH (`cqlite-cli/src/error.rs:18-30`,
   `CliExitCode`): `2=InvalidCliArgs, 3=SchemaError, 4=DataDirError, 5=QueryExecutionError,
   6=WriteError`. **Both stated meanings are false** — a triaging agent chases a parsing bug on a
   schema error, and `4=DataDirError` (the *most likely* real-world failure: missing datasets) is
   absent entirely.

2. **`.claude/agents/test-validator.md:17,99` — "all 33 test tables" / `Pass: X/33`** is the
   **RETIRED hand-typed allowlist** (issue #1229 (retire the hand-typed 33-table allowlist), CLOSED).
   TRUTH: the corpus is enumerated **FROM DISK**; `test-data/validation-matrix.md:16` states the
   33-table list is "no longer the coverage source of truth", and `:52` records
   **50 enforced tables across 8 keyspaces**. A hardcoded denominator **silently under-reports new
   tables** — a validator can report "33/33 PASS" while 17 tables were never touched.

3. **`.claude/agents/coverage-reviewer.md:9` — claims it complements "the numeric coverage gate
   enforced by the TaskCompleted hook". VERIFIED FALSE:** `.claude/settings.json:5` sets
   `ISSUE_GATE_COVERAGE_CMD: ""` and **no** gate component measures coverage. The false premise leads
[...]

Issue #1082, opened 2026-06-25
Deflate decompression uses raw DEFLATE, not zlib-wrapped (Cassandra DeflateCompressor unreadable)


pmcfadin:
Found during Epic #970 / issue #998.

## Problem
`cqlite-core/src/storage/sstable/chunk_decompressor.rs::decompress_deflate_chunk` decoded with `flate2::read::DeflateDecoder` (raw DEFLATE), but Cassandra's `DeflateCompressor` (java.util.zip.Deflater/Inflater) emits **zlib-wrapped** streams (chunk payloads begin with the zlib header `78 9c`, + Adler-32 trailer). The full-scan path (`stitch_all_chunks` → `Compression::decompress` in `compression.rs`) additionally read the first 4 bytes as a bogus uncompressed-size prefix. **Zstd had the identical full-scan bug** (frame magic `28 B5 2F FD` read as a ~2GB size → decompression-bomb guard tripped).

Result: every Deflate-compressed SSTable failed with `corrupt deflate stream`, and `SELECT *` over a Zstd nb table failed with `size … exceeds limit`.

## Reproduce (pre-fix)
~~~bash
# 1. Generate the compression fixtures (Docker cassandra:5.0.2)
bash test-data/scripts/generate-compression-parity.sh
export CQLITE_DATASETS_ROOT=$PWD/test-data/datasets

# 2. Confirm the deflate payload is zlib-wrapped (starts with 78 9c after the 0-offset)
xxd test-data/datasets/sstables/test_comp/deflate_table-*/nb-1-big-Data.db | head -1

# 3. Full-scan the deflate (and zstd) table — FAILS before the fix
cargo run -p cqlite-cli --features write-support -- \
  --schema test-data/schemas/compression-parity.cql \
  --data-dir test-data/datasets/sstables \
  --query "SELECT * FROM test_comp.deflate_table" --out json
# pre-fix: Error: ... corrupt deflate stream
# zstd_table pre-fix: Error: ... Decompression bomb protection: size 2023550423 exceeds limit
~~~

## Fix
Decode Cassandra Deflate with `flate2::read::ZlibDecoder` and Zstd as a bare frame, in BOTH the chunk path (`chunk_decompressor.rs`) and the stitch path (`compression.rs`); never strip a 4-byte size prefix for Deflate/Zstd (that is an LZ4-only convention). Bound output by `chunk_length` to keep the decompression-bomb guard.

## Verify (post-fix)
~~~bash
[...]

PR #3075, opened 2026-07-28
docs(#3054): correct 13 wrong format tables in the auto-loaded agent skills + mechanize the drift guard

### LINKED ISSUE #3054: skills: agent-facing format tables are factually WRONG — compression-formats.md, sstable-parsing row flags, and cql-type-system encodings teach known-P0 bugs back to agents
## Why this is P1 and not a docs nit

`.claude/skills/sstable-parsing/` and `.claude/skills/cql-type-system/` are **auto-loaded skills** — they enter agent context whenever anyone does binary-format or type-decode work in this repo. An agent trusts a skill table over reading the code, so **a wrong table is strictly worse than no documentation**: it produces a confidently-wrong implementation and a confidently-wrong review.

Two of the tables below encode bugs **we already shipped, diagnosed as P0, and fixed** — the skill teaches the bug back to the next agent:
- the Deflate sample uses raw-deflate decoding, which is exactly P0 #1082 (deflate-as-zlib / zstd bare-frame decode bug);
- the row-flag table mis-assigns `0x01`, which is the partition-boundary bit.

Related but distinct: #2989 (Appendix G calls Cassandra Deflate a "raw Deflate stream") is the same error *class* in the published guide; this issue is the **agent-facing skill surface**. Fix both; neither supersedes the other.

## Routing: SKIPS OpenSpec — go straight to `flow-implement`

This is **doc/instruction-surface work with NO design latitude and NO Cassandra oracle dispute**. Every correction below is a lookup against an existing code constant on `origin/main` or an existing line of `docs/sstables-definitive-guide/chapters/appendix-b-encodings-cheat-sheet.md`. There is nothing to design and nothing to litigate with Cassandra. **This issue skips OpenSpec / `flow-activate` entirely — dispatch it directly to `flow-implement`.**

## IMPORTANT: read the sources via `origin/main`

The root working tree is roughly 30 commits behind `origin/main`. Verify and edit against `origin/main` — read every cited source with `git show origin/main:<path>` (or work in a fresh worktree branched from `origin/main`), never from a stale local checkout. Several cited files were moved/deleted by epic #1116 (source splits).

---

## Scope A — `.claude/skills/sstable-parsing/compression-formats.md`

This is the single worst file in the repo for factual accuracy. All six items below are **verified** against `origin/main`.

**A1 — `CompressionInfo.db` layout is fabricated.** Lines 12–20 show the file as a flat array of `[offset u64][length u32]` pairs. TRUTH (`cqlite-core/src/storage/sstable/compression_info.rs:6-20,83`): `writeUTF(compressor_name)`, `i32 option_count`, options, `i32 chunk_length`, `i32 max_compressed_length`, `i64 data_length`, `i32 chunk_count`, then `chunk_count` × `i64` offsets **only**. There is **no per-chunk length field** — a chunk's physical length is the delta between consecutive offsets, and that delta **includes the trailing 4-byte CRC**.

**A2 — Deflate sample decodes raw deflate (this is P0 #1082 taught back).** Lines 80–89 use `flate2::read::DeflateDecoder`. TRUTH: Cassandra's `DeflateCompressor` emits a **zlib-wrapped** stream, so it must be `flate2::read::ZlibDecoder` — verified at `cqlite-core/src/storage/sstable/compression.rs:306,313,324`, which carries an explicit `#1082` comment marking this exact trap.

**A3 — LZ4 sample uses a crate we do not depend on.** Lines 45–50 call `lz4::block::decompress(compressed, Some(size))`. TRUTH: the repo uses **`lz4_flex`** over a **4-byte little-endian raw length prefix + raw block** (`cqlite-core/src/storage/sstable/chunk_decompressor.rs:361,391`). The `lz4` crate is not a dependency at all, so the sample cannot even compile here.

**A4 — "three compression algorithms" (lines 5–8, and `.claude/skills/README.md:20`).** TRUTH: **four plus Noop** — LZ4, Snappy, Deflate, Zstd, Noop (`compression_info.rs:43-48`). Zstd is silently missing from the agent's mental model.

**A5 — chunk params attributed to the wrong file, and a wrong default (lines 31–34, 157).** Chunk params are said to come from `Statistics.db` with `chunkLength` "typically 64KB". TRUTH: they live in **`CompressionInfo.db`**, and the Cassandra 5.0 default `chunk_length` is **16 KiB**.

**A6 — CRC32 is described as optional and over the wrong bytes (lines 26–29, 96–103).** The file says `[crc32: 4 bytes] (if CRC enabled)` and computes it over the *decompressed* data. TRUTH: for `na`/`nb` the 4-byte **big-endian** inline CRC32 is **unconditional**, and it is computed over the **compressed** bytes (`chunk_decompressor.rs:275-292`).

## Scope B — `.claude/skills/sstable-parsing/SKILL.md` + `cassandra5-format-reference.md`

**B1 — the row-flag table is inverted/shifted, and self-contradictory.** `SKILL.md:112-118` and `cassandra5-format-reference.md:17-24` claim `0x01 = HAS_IS_MARKER`, `0x02 = HAS_ALL_COLUMNS (inverted)`, `0x20 = HAS_ALL_COLUMNS`, `0x40 = IS_STATIC`. Note it **self-contradicts**: it labels *both* `0x02` and `0x20` as `HAS_ALL_COLUMNS`.

TRUTH, verified in `cqlite-core/src/storage/sstable/reader/parsing/row_decoder/mod.rs:714,820,821,825` and `docs/sstables-definitive-guide/chapters/appendix-b-encodings-cheat-sheet.md:206-212`:
- `END_OF_PARTITION = 0x01`
- `IS_MARKER = 0x02`
- `ROW_HAS_COMPLEX_DELETION = 0x40`
- `IS_STATIC = 0x01` **of the EXTENDED flag byte** — not the main flag byte.

Mis-decoding `0x01` means **mis-detecting partition boundaries** — the highest-consequence single bit in the row format.

**B2 — `SKILL.md:97` names a module that no longer exists.** It points the agent at `cqlite-core/src/storage/sstable/reader/parsing/v5_compressed_legacy.rs` ("1997 lines"). **Verified deleted** by epic #1116 (source splits), commit `cb049f7a8`. Successor: the `cqlite-core/src/storage/sstable/reader/parsing/row_decoder/` directory (30 files). Point the agent at the successor.

## Scope C — `.claude/skills/cql-type-system/`

**C1 — `duration` components shown as plain VInts (`cql-types-reference.md:196-203`).** TRUTH: `duration`'s three components (months, days, nanos) are the **only ZigZag-signed VInts in Data.db** (`appendix-b-encodings-cheat-sheet.md:63-65,92-97`). Following the skill, a **negative duration decodes wrong**.

**C2 — frozen-only collection layout presented as THE collection layout (`SKILL.md:68-74`, `collections-and-udts.md:10-20`).** The skill presents `[i32 count][i32 elem_len]…` as the collection format. TRUTH: that is the **frozen** layout only. A **non-frozen** collection cell uses a **VInt `cell_count`** and **unsigned-VInt value-length prefixes** (`appendix-b-encodings-cheat-sheet.md:530-536`, `cqlite-core/src/storage/sstable/reader/parsing/row_decoder/complex_column.rs:331-340`). Document both, clearly labeled.

**C3 — frozen inner-list blob sizes are arithmetically wrong (`collections-and-udts.md:96,104`).** Given as `0x18` (24) and `0x24` (36). TRUTH: the arithmetic yields **`0x14` (20)** for `[1,2]` and **`0x1C` (28)** for `[3,4,5]`. Show the derivation next to the corrected number.

**C4 — invented CQL alias (`cql-types-reference.md:284`).** Claims `blob` = `bytea` "(PostgreSQL compatibility)". TRUTH: **no such CQL alias exists.** Delete it.

## Scope D — the ONE UNVERIFIED item: DO NOT FIX BLIND

**D1 — `date` decode bias. UNCONFIRMED; the implementer must settle it before touching it.**

An audit claimed `cql-types-reference.md:177-183` and `SKILL.md:48` wrongly show `date` as a plain `u32::from_be_bytes` days-since-epoch, alleging the on-disk value carries an `Integer.MIN_VALUE` / `0x80000000` bias (epoch day 0 encoded as `0x80000000`), citing `cell_value_scalar.rs:330-333`.

**This claim did not reproduce.** A direct grep for `0x8000_0000|0x80000000|2147483648` in that file returned **nothing**. Do **not** apply a bias "fix" on the strength of the audit note.

The implementer must settle it independently against both:
1. Cassandra 5.0.8: `git show cassandra-5.0.8:src/java/org/apache/cassandra/db/marshal/SimpleDateType.java` (in `~/local_projects/cassandra`), and
2. the actual CQLite decode path on `origin/main`,

then either fix the skill **or** record it as dismissed with the Cassandra 5.0.8 citation. Either outcome closes this item; an unexplained silence does not.

---

## Acceptance criteria

Each criterion is checkable by `rg` against the named file on the PR branch, or by a test.

- [ ] **AC1 — A1** `compression-formats.md` documents the real `CompressionInfo.db` header (`writeUTF` name, `i32 option_count` + options, `i32 chunk_length`, `i32 max_compressed_length`, `i64 data_length`, `i32 chunk_count`, then `chunk_count` × `i64` offsets), states explicitly that **no per-chunk length field exists**, and states that the offset delta includes the trailing 4-byte CRC. The `[chunk_length_1: u32]` pair-array no longer appears anywhere in the file.
- [ ] **AC2 — A2** No `DeflateDecoder` remains in `compression-formats.md`; the sample uses `ZlibDecoder` and the text says Cassandra's Deflate output is zlib-wrapped, citing #1082 (deflate-as-zlib / zstd bare-frame P0) so the trap is named.
- [ ] **AC3 — A3** No reference to the `lz4` crate remains; the sample uses `lz4_flex` with the 4-byte little-endian raw length prefix + raw block, citing `chunk_decompressor.rs`.
- [ ] **AC4 — A4** Both `compression-formats.md` and `.claude/skills/README.md` say **four algorithms plus Noop** and name Zstd; no "three compression algorithms" string survives in either file.
- [ ] **AC5 — A5** Chunk params are attributed to `CompressionInfo.db` (not `Statistics.db`) and the stated Cassandra 5.0 default `chunk_length` is 16 KiB; no "64KB" default claim remains.
- [ ] **AC6 — A6** The inline CRC32 is documented as **unconditional** in `na`/`nb`, 4 bytes **big-endian**, over the **compressed** bytes; the `(if CRC enabled)` qualifier and any "over decompressed data" claim are gone.
- [ ] **AC7 — B1** The row-flag tables in **both** `SKILL.md` and `cassandra5-format-reference.md` read `END_OF_PARTITION=0x01`, `IS_MARKER=0x02`, `ROW_HAS_COMPLEX_DELETION=0x40`, and state that `IS_STATIC` is `0x01` of the **extended** flag byte. `HAS_IS_MARKER` is gone, and no value is labeled `HAS_ALL_COLUMNS` twice (the self-contradiction is resolved).
- [ ] **AC8 — B2** `rg 'v5_compressed_legacy' .claude/skills/` returns **zero** hits; the module pointer names the `row_decoder/` directory instead.
- [ ] **AC9 — C1** `duration`'s three components are documented as ZigZag-signed VInts (the only signed VInts in Data.db), with the appendix-b citation.
- [ ] **AC10 — C2** Frozen vs non-frozen collection layouts are documented as **two distinct encodings**, explicitly labeled, with the non-frozen one showing VInt `cell_count` + unsigned-VInt value lengths.
- [ ] **AC11 — C3** The frozen inner-list sizes read `0x14` (20) and `0x1C` (28), with the byte-count derivation shown.
- [ ] **AC12 — C4** `rg 'bytea' .claude/skills/` returns zero hits.
- [ ] **AC13 — traceability (applies to all of the above)** Every flag byte and every type encoding in these skill files carries an **inline citation** next to its table — either a code constant path+symbol on `origin/main` or an `appendix-b-encodings-cheat-sheet.md` line reference. A table with no citation is an incomplete AC.
- [ ] **AC14 — D1 resolved explicitly** The `date` bias question is either (a) fixed in the skill with the Cassandra 5.0.8 `SimpleDateType.java` citation, or (b) documented in the PR body as **dismissed**, with that same citation and the decode-path evidence. Silence fails this AC.

### Strongly recommended (in scope if cheap, otherwise file a linked follow-up)

- [ ] **AC15 — drift test** Add a test that pins the documented row-flag table against the **actual constants** in `row_decoder/mod.rs`, so this class of rot cannot recur silently. The existing `roborev-lints` and `tooling-tests` gate components are both plausible homes (a small script that parses the markdown table and asserts each name→value pair matches the source constant). If this turns out to be more than a cheap addition, file a linked follow-up issue and reference it here rather than expanding this PR.

## Notes

- **No-heuristics**: no impact — no decode behavior changes. But note that several of these wrong tables would *lead* an agent into a heuristic (e.g. guessing collection framing), so fixing them is protective of the #28 mandate.
- **Memory budget**: no impact.
- **Wiring evidence**: the "public surface" here is the auto-loaded skill content itself; AC13's inline citations plus AC15's drift test are the evidence that the surface stays true.

pmcfadin:
Closes #3054

Six AUTO-LOADED agent skill files carried 13 factually wrong on-disk-format tables. Two taught already-shipped-and-fixed **P0** bugs back to every agent that touches binary format work. A wrong table is strictly worse than no table: an agent trusts a skill table over reading the code.

Every table/constant in this diff now carries an **inline citation** — either a code constant on `origin/main` or a line in `docs/sstables-definitive-guide/chapters/appendix-b-encodings-cheat-sheet.md`. A skill fact without a citation is how this rotted.

## Defects fixed, per file

**`.claude/skills/sstable-parsing/compression-formats.md`** (6) — invented `chunk_length_1` per-chunk length field replaced with the real `CompressionInfo.db` header + offsets-only chunk array, plus an explicit *"There is NO per-chunk length field"* and the delta-includes-the-trailing-4-byte-CRC rule (`compression_info.rs:6-20,76-83`); **P0 #1082 regression** — the Deflate sample used flate2's raw-DEFLATE reader; now `ZlibDecoder` with a `⚠️ TRAP` callout (`compression.rs:306,313,324`); the lz4 sample used the bare `lz4` crate's `block::decompress`; now `lz4_flex` with Cassandra's 4-byte **little**-endian raw-length prefix (`chunk_decompressor.rs:391,426`); "three compression algorithms" → four + Noop (`compression_info.rs:43-48`); 64KB default → **16 KiB** attributed to `CompressionInfo.db`, not the schema (`CompressionParams.java:47`); *"if CRC enabled"* → the per-chunk 4-byte **big-endian** CRC32 over the **compressed** bytes is **unconditional** (`chunk_decompressor.rs:275-292`). Also corrected the offset-calculation snippet and a dead `reader/compression/` path.

**`.claude/skills/sstable-parsing/SKILL.md`** (1) — the row-flag table labelled `0x01` as `IS_MARKER`; `0x01` is `END_OF_PARTITION`, so the table taught **partition-boundary mis-detection**. Replaced with two tables (main byte + EXTENDED byte) citing `row_decoder/mod.rs:709-715,820,821,825` and `appendix-b:206-212`. Retargeted the stale single-file module pointer at the `row_decoder/` directory.

**`.claude/skills/sstable-parsing/cassandra5-format-reference.md`** (3) — same row-flag correction + common combos; the cell-flag table listed `0x20 HAS_NULL_VALUE` and `0x40 EXTENDED_FLAG`, **both fabrications** — Cassandra 5.0.8 `db/rows/Cell.java:262-266` defines exactly five masks and a cell has no extended flag byte; and the complex-cell section conflated frozen with non-frozen, now split explicitly.

**`.claude/skills/cql-type-system/cql-types-reference.md`** (2) + **`SKILL.md`** (1) — `date` documented without its `Integer.MIN_VALUE` bias (see below); `duration`'s three components documented as unsigned when they are the **only ZigZag-signed VInts in `Data.db`** (`appendix-b:92-97`); `bytea` — a **PostgreSQL** type, not CQL — deleted (`CQL3Type.java:88-111`).

**`.claude/skills/cql-type-system/collections-and-udts.md`** (2) — the frozen-vs-non-frozen encodings were conflated (fixed 4-byte BE `i32` vs unsigned VInt + per-cell flags/path); nested frozen inner-list sizes were wrong, corrected to `0x14`/`0x1C` with a `4 + 8n` derivation table. Also killed a *"Wire format: Same as non-frozen"* claim that directly contradicted the file's own corrected section.

**`.claude/skills/README.md`** (1) — line 20's "three compression algorithms".

## Item 13 (`date` bias): SETTLED — **CONFIRMED REAL, FIXED** (not dismissed)

The audit's grep for a literal `0x80000000` / `2147483648` found nothing, which is why this was flagged unverified. The bias **is** in the decode path — it is just spelled `i32::MIN as u32`:

`cqlite-core/src/storage/sstable/reader/parsing/row_decoder/cell_value_scalar.rs:329-333`
```rust
let stored = u32::from_be_bytes([...]);
let days_since_epoch = stored.wrapping_add(i32::MIN as u32) as i32;
```

Independently confirmed as the **format** truth against Apache Cassandra 5.0.8 `serializers/SimpleDateSerializer.java:36-37` (*"we shift by Integer.MIN_VALUE … w/epoch sitting in the center @ 2^31"*), `:110`, `:113-115`. Both files now document the bias with an un-bias snippet and both citations, including the note that a grep for the hex literal will not find it.

## AC15: drift test — **landed**

- **`scripts/ci/check-skill-flag-tables.sh`** (new) — harvests `const NAME: u8 = 0xNN;` from `row_decoder/{mod.rs,row_data.rs}` and asserts every documented skill-table row's name is real, its value matches, and no source constant is undocumented. Fails **closed** on a missing source, a zero-constant harvest, absent #3054 anchors, or a file with no parseable table row. `HAS_EMPTY_VALUE 0x04` is pinned as a documented-by-Cassandra-only exception with its authority.
- **`scripts/tests/test_check_skill_flag_tables.sh`** (new) — hermetic 6-assertion self-test proving non-vacuity by mutating a temp sandbox: a shifted value (the exact pre-#3054 partition-boundary bug), an invented name, a dropped row, plus both fail-closed paths.
- Wired into the **`tooling-tests`** gate component (pure/offline; no cargo, datasets, or network).

So this class of rot now FAILs the fast loop rather than costing a review round.

## Validation

- **`AGENT-GATE LITE SUMMARY`: RESULT: PASS** (`/tmp/lite-3054.txt`, commit `0e11d36`, `tree-integrity: PASS`) — file-size, fmt, clippy, roborev-lints, scoped-tests all PASS.
- The new guard PASSes on the real repo: `OK: 23 documented flag rows across 2 skill file(s) match the decoder constants; all 14 source flags are documented.`
- Self-test: `PASS: test_check_skill_flag_tables.sh — all 6 assertions hold`.
- All AC literal `rg` checks read clean in scope: `chunk_length_1`, `DeflateDecoder`, `lz4::block`, `if CRC enabled`, `three compression`, `64KB`, `bytea`, `v5_compressed_legacy`.

## Out of scope (reported, not fixed)

4 remaining `v5_compressed_legacy` references live in #3056's lane: `rust-patterns/SKILL.md:337`, `rust-patterns/zero-copy-patterns.md:7,295`, `test-data-management/validation-workflow.md:355`.

## Routing

Doc/instruction-surface + a mechanized drift guard; no design latitude → **no OpenSpec change** (per issue routing).

### DIFF
--- a/.claude/skills/README.md
+++ b/.claude/skills/README.md
@@ -17,7 +17,7 @@ plus general-purpose Rust guidance.
 **Files:**
 - `SKILL.md` - Main skill guidance
 - `cassandra5-format-reference.md` - Complete row format specification
-- `compression-formats.md` - LZ4, Snappy, Deflate formats
+- `compression-formats.md` - the four compression algorithms plus Noop: LZ4, Snappy, Deflate, Zstd, Noop (`cqlite-core/src/storage/sstable/compression_info.rs:43-48`)
 
 ---
 
--- a/.claude/skills/cql-type-system/SKILL.md
+++ b/.claude/skills/cql-type-system/SKILL.md
@@ -45,8 +45,12 @@ Never try to infer type from data alone - always use schema.
 - `bigint` - 8 bytes signed, big-endian
 - `float` - 4 bytes IEEE 754
 - `double` - 8 bytes IEEE 754
-- `date` - 4 bytes (days since epoch)
-- `time` - 8 bytes (nanoseconds since midnight)
+- `date` - 4 bytes BE unsigned, days since epoch **biased by `Integer.MIN_VALUE`**
+  (epoch day 0 is stored as `0x80000000`) — un-bias with
+  `stored.wrapping_add(i32::MIN as u32) as i32`; a plain `u32` read is wrong by 2^31 days.
+  Citations: `row_decoder/cell_value_scalar.rs:329-333`; cassandra-5.0.8
+  `serializers/SimpleDateSerializer.java:36-37,110,113-115`.
+- `time` - 8 bytes signed BE (nanoseconds since midnight; no bias)
 
 #### Variable-Size Primitives
 - `text`/`varchar` - UTF-8 encoded string
@@ -58,20 +62,52 @@ Never try to infer type from data alone - always use schema.
 - `inet` - 4 bytes (IPv4) or 16 bytes (IPv6)
 - `varint` - variable-length big integer
 - `decimal` - scale (4 bytes) + unscaled varint
-- `duration` - months, days, nanoseconds (3 VInts)
+- `duration` - months, days, nanoseconds as 3 **ZigZag-SIGNED** VInts — the ONLY signed
+  VInts in `Data.db`, wherever a duration payload occurs (including nested in a
+  collection/tuple/UDT). Decoding them unsigned makes every negative duration wrong.
+  Citations: `appendix-b-encodings-cheat-sheet.md:63-65,92-97`;
+  `DurationSerializer.java:49-51`.
 - `timestamp` - 8 bytes (milliseconds since Unix epoch)
 
 ### 2. Collection Types
 
 See [collections-and-udts.md](collections-and-udts.md) for detailed format.
 
-**Collection Format:**
+> ### ⚠️ There are TWO collection encodings — pick by frozen-ness, never by byte pattern
+> Frozen and non-frozen collections do NOT share a wire format. Assuming the frozen layout
+> for a non-frozen column (or vice versa) mis-frames every element.
+
+**FROZEN collection** (`frozen<list<int>>`, and any collection nested inside another
+collection/tuple/UDT) — one opaque cell, **fixed 4-byte big-endian `i32`** count and
+element lengths (NOT VInts):
 ```
-[4 bytes: element_count (big-endian)]
+[i32 BE: element_count]
 [for each element:]
-    [4 bytes: element_size (big-endian)]
+    [i32 BE: element_len]   // -1 = null
     [bytes: element_data]
 ```
+> **Citation**: `cqlite-core/src/storage/sstable/reader/parsing/row_decoder/frozen.rs:21`
+> (count), `:98,279,370` (element/key lengths). Guide:
+> `appendix-b-encodings-cheat-sheet.md:537-539`. Cassandra:
+> `CollectionSerializer.java:67-92`, `TupleType.java:341-364`.
+
+**NON-FROZEN collection** (the default `list<int>`, `set<T>`, `map<K,V>`) — a *set of
+cells*, framed with **unsigned VInts**:
+```
+[if the row sets ROW_HAS_COMPLEX_DELETION (0x40): complex deletion, 2 unsigned VInt deltas]
+[unsigned VInt: cell_count]
+[for each cell:]
+    [1 byte: cell flags]
+    [conditional timestamp / local_deletion_time / ttl unsigned-VInt deltas]
+    [unsigned VInt: path_len][path bytes]     // element key / path
+    [unsigned VInt: value_len][value bytes]   // omitted when IS_DELETED or HAS_EMPTY_VALUE
+```
+> **Citation**: `cqlite-core/src/storage/sstable/reader/parsing/row_decoder/complex_column.rs:332`
+> (`cell_count` via `parse_vuint`), `:1089` (path length), `:1136` (value length),
+> `:279-300` (complex deletion). Guide: `appendix-b-encodings-cheat-sheet.md:530-536` — a
+> non-frozen collection cell length-prefixes BOTH path and value with an unsigned VInt,
+> **fixed-width element types included** (the `valueLengthIfFixed()` raw-bytes shortcut is
+> a simple-cell rule only, `AbstractType.java:538-543`).
 
 **Types:**
 - `list<T>` - Ordered, allows duplicates
--- a/.claude/skills/cql-type-system/collections-and-udts.md
+++ b/.claude/skills/cql-type-system/collections-and-udts.md
@@ -135,7 +135,7 @@ CREATE TABLE t (
 );
 ```
 
-### Nested Collection Wire Format
+### Nested Collection Wire Format (FROZEN inner lists)
 ```
 CQL: list<frozen<list<int>>>
 Value: [[1, 2], [3, 4, 5]]
@@ -144,7 +144,7 @@ Wire format:
 [0x00, 0x00, 0x00, 0x02]  // outer count = 2
 
 // First nested list [1, 2]
-[0x00, 0x00, 0x00, 0x18]  // size of entire frozen list
+[0x00, 0x00, 0x00, 0x14]  // size of entire frozen list = 20 (see derivation below)
 [0x00, 0x00, 0x00, 0x02]  // inner count = 2
 [0x00, 0x00, 0x00, 0x04]  // size = 4
 [0x00, 0x00, 0x00, 0x01]  // value = 1
@@ -152,7 +152,7 @@ Wire format:
 [0x00, 0x00, 0x00, 0x02]  // value = 2
 
 // Second nested list [3, 4, 5]
-[0x00, 0x00, 0x00, 0x24]  // size of entire frozen list
+[0x00, 0x00, 0x00, 0x1C]  // size of entire frozen list = 28 (see derivation below)
 [0x00, 0x00, 0x00, 0x03]  // inner count = 3
 [0x00, 0x00, 0x00, 0x04]  // size = 4
 [0x00, 0x00, 0x00, 0x03]  // value = 3
@@ -161,6 +161,18 @@ Wire format:
 [0x00, 0x00, 0x00, 0x04]  // size = 4
 [0x00, 0x00, 0x00, 0x05]  // value = 5
 ```
+
+**Derivation of the frozen inner-list sizes.** A frozen `list<int>` body is
+`4 (i32 count) + n * (4 (i32 element_size) + 4 (int value))` = `4 + 8n`:
+
+| Inner list | n | Bytes | Hex |
+|---|---|---|---|
+| `[1, 2]` | 2 | `4 + 2*8 = 20` | `0x14` |
+| `[3, 4, 5]` | 3 | `4 + 3*8 = 28` | `0x1C` |
+
+Count the bytes listed above each inner list to confirm: `[1,2]` shows 5 four-byte words
+after its size word (count, size, value, size, value) = 20; `[3,4,5]` shows 7 = 28. The
+size word itself is NOT included in the size it declares.
 
 ## Deserialization Code
 
--- a/.claude/skills/cql-type-system/collections-and-udts.md
+++ b/.claude/skills/cql-type-system/collections-and-udts.md
@@ -359,10 +359,17 @@ UPDATE t SET data = [1, 2, 3] WHERE id = 1;
 -- Cannot: UPDATE t SET data[0] = 99 WHERE id = 1;
 ```
 
-**Wire format:** Same as non-frozen, but:
-- Serialized as single blob in parent structure
-- No tombstones for individual elements
-- Entire collection replaced on update
+**Wire format: NOT the same as non-frozen** — see the two-encoding table at the top of this
+file. A frozen collection is:
+- serialized as a **single opaque cell value**, with fixed 4-byte BE `i32` count and element
+  lengths (`row_decoder/frozen.rs:21,98`) — not a set of cells with unsigned-VInt framing;
+- unable to represent per-element tombstones (there is no per-element cell to flag);
+- replaced wholesale on update.
+
+A **non-frozen** collection is a set of cells with `cell_count` and value lengths as
+**unsigned VInts**, per-element cell flags/paths, and per-element tombstones
+(`row_decoder/complex_column.rs:332,1089,1136`;
+`appendix-b-encodings-cheat-sheet.md:530-536`).
 
 ### Frozen UDTs
 ```cql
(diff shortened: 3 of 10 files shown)


pmcfadin:
## Review round 1 — `rust-reviewer` (opus): 3 BLOCKERs fixed in `d59b523f7`

The markdown is out of a code reviewer's scope; the three shell artifacts (`scripts/ci/check-skill-flag-tables.sh`, its self-test, the `agent-gate.sh` wiring) were the review target. All three blockers were real and are fixed. Summarized because each is a reusable lesson:

**BLOCKER 1 — `set -e` + `pipefail` made the guard's most actionable diagnostic UNREACHABLE.** `lookup()` ended in `grep | awk`; on a miss (exactly the invented-flag-name case) grep exits 1, `pipefail` propagates it out of the function, and under `set -e` *an assignment whose command substitution fails terminates the script*. So the script died at the assignment having printed **nothing** — the "documents flag X which is NOT a constant in the decoder source" message, the pre-#3054 examples, and the authority citation were all dead code. An operator would have seen "something drifted" with no flag name and no path. Second-order: it made the self-test's invented-name assertion **vacuous**, since it only asserted non-zero exit and non-zero arrived via an unintended silent abort. Fixed with `|| true` plus an inline comment naming the trap, and **every** negative assertion now greps the intended diagnostic rather than merely checking the exit code.

**BLOCKER 2 — the cell-flag authority was a `#[cfg(test)]` mirror, and the exception was justified by a code comment.** The guard harvested `CELL_*` from `row_data.rs:860`, which sits inside `parse_cell_header_end_offset` — test-only, and carrying just 4 of the 5 flags. `HAS_EMPTY_VALUE 0x04` was therefore hardcoded, with a comment claiming CQLite has no named const for it, citing `row_data.rs:282` — *a comment, not a constant*. The claim was false: `cell_value.rs:51` is `const CELL_HAS_EMPTY_VALUE: u8 = 0x04;`, in `parse_cell_value_schema_order`, the **production** decoder, which has all five. That made the guard documentation asserting itself — precisely the rot class this issue exists to kill — and left the four real flags pinned to a test helper that a refactor could delete out from under it. Now harvests `cell_value.rs:49-53`; the carve-out is **deleted entirely**, restoring "every documented name comes from a source constant" with no exception. Skill citations retargeted with a note on why `row_data.rs` is the wrong authority.

**BLOCKER 3 — a hand-maintained allow-list silently exempted any newly added flag.** The classifier matched `ROW_HAS_*|END_OF_PARTITION|IS_MARKER|EXTENDED_IS_STATIC`; anything else entered no namespace and so was never required to be documented. Cassandra 5.0's extended byte really does define `HAS_SHADOWABLE_DELETION = 0x02`, so the day someone adds it the guard stays green on an incomplete table — the #28 hazard the guard's own error message names. This is the same deferred can't-fail shape #2926's portability suite already had to remove twice. An unclassified `u8` const now **FAILs closed**; a genuine non-flag constant opts out **visibly** via a `flag-table-lint-ignore` marker. The harvest regex also now accepts any visibility qualifier, so a `pub(super) const` flag (cf. `row_framing.rs:8`) can no longer slip past the parse.

**NITs, all taken:** namespace-aware checking (`0x01` is `END_OF_PARTITION` / `EXTENDED_IS_STATIC` / `IS_DELETED` depending on the byte — a correct name+value under the wrong heading now FAILs, and a table that drifts out from under a recognizable heading FAILs closed); a harvest-size floor so a half-broken parse cannot pass (set below today's 14 so *adding* a flag never trips it); dropped `mapfile` for bash-3.2 portability, verified against `/bin/bash 3.2.57`; `trap` extended to `INT TERM`.

Self-test grew **6 → 10 assertions**, adding: a shifted **cell** value in `cassandra5-format-reference.md` (the cell table's only home — previously the entire `CELL_*` path had no mutant), the namespace swap, the unclassified-new-flag fail-closed, the escape hatch, and the drifted-heading fail-closed. All 10 green under both bash 5 and stock macOS bash 3.2.

The reviewer explicitly cleared the `agent-gate.sh` wiring ("matches the surrounding pattern exactly … nothing to fix here") and found no injection surface.

Issue #3061, opened 2026-07-27
Streaming scan shows 1,017 MiB peak RSS with Data.db mapped TWICE — determine whether the <128MB target is violated or measured wrongly


pmcfadin:
## Summary

During the #3026 (WS0) head-to-head, a streaming CQLite scan showed **peak RSS of 1,017 MiB**, and `Data.db`
appeared as **two separate `r--s` mappings of 783,799,203 B each** (1,567.6 MB of virtual address space) — i.e.
the same file mapped twice by one process.

Two separate questions, and I want to be explicit that I have **not** established either is a real defect:

1. **Is the same `Data.db` genuinely mapped twice, and if so why?** This is the concrete, checkable part.
2. **Does this violate the project's `<128MB for large files` memory target** — or is the target simply not
   measured by RSS?

**⚠️ Observed once, not reproduced.** Context: #3023, epic #2817.

## The important caveat — RSS may be the wrong instrument

`mmap`-backed file pages count toward RSS but are **clean, page-cache-backed and evictable under pressure**.
So a 1,017 MiB RSS on a process that memory-maps a 783 MB file does **not** by itself mean the process is
holding 1 GB of unreclaimable memory, and it may not violate a `<128MB` target that is really about
anonymous/heap footprint.

**So the first task is to decide what the target means and measure the right thing**, not to "fix" RSS:
- anonymous/heap resident (`RssAnon` in `/proc/<pid>/status`, or a heap profiler such as the existing
  `dhat-heap` feature) — this is probably what `<128MB` intends;
- vs file-backed resident (`RssFile`) — expected to scale with the mapped file and largely harmless.

If `RssAnon` stays well under 128 MB, the honest outcome is **"no violation; the target should be restated in
terms of anonymous memory"** — a documentation fix, and a perfectly good result.

## The part that looks like a genuine defect regardless

**Why is `Data.db` mapped twice?** One process holding two independent 783,799,203 B mappings of the same file
suggests two readers each mapping it rather than sharing one mapping (or a mapping not being unmapped when a
reader is recreated). Even though the pages are shared in the page cache and the real memory cost is near zero,
this is worth understanding because:
- it doubles VA consumption per open SSTable, which matters at scale (many SSTables × many tables);
- if it indicates a reader/handle being constructed twice per scan, there may be duplicated setup work behind it;
- it interacts with the mmap-vs-pread question raised on #3031 — CQLite mmapping `Data.db` here (115 `read()`
[...]

Issue #3060, opened 2026-07-27
cqlite-flight may never exit when signalled mid-stream — spins at ~176% CPU in 'draining in-flight RPCs' with the client blocked (observed once, needs reproduction)


pmcfadin:
## Summary

A `cqlite-flight` server that is signalled **mid-stream** (while a `do_get` scan is actively streaming to a
client) did not exit. It stayed spinning at **~176% CPU** reporting **"draining in-flight RPCs"** indefinitely,
with its client blocked. It required a `SIGKILL`.

**⚠️ Observed ONCE and NOT reproduced.** Filing it because a shutdown path that hangs and wedges its client is
too serious to leave as a note in a perf report — but the first job here is reproduction, and it is entirely
possible this is environment-specific or an artefact of how the signal was delivered.

Found incidentally while running the #3026 (WS0) stock-Cassandra head-to-head. Context: #3023.

## What was observed

- Server: release `cqlite-flight` (0.16.1, built with `--features observability`, debug symbols retained) serving
  a full-table scan of `ws0.events` — 3,999,890 rows, 783,799,203 B `nb` SSTable, LZ4 `chunk_length=16384`.
- A signal was delivered to the server process while the stream was in flight.
- Server logged **"draining in-flight RPCs"** and then **never exited**.
- It continued burning **~176% CPU** (i.e. >1 core — consistent with a spin/busy-wait rather than a blocked
  wait) for as long as it was observed.
- The client remained **blocked**, never receiving an error or a stream termination.
- Recovery required `SIGKILL`.

## Why it is worth chasing

- **~176% CPU is the diagnostic detail.** A shutdown that merely *waits forever* on a channel or join handle
  would idle near 0%. Burning more than a core suggests a busy-loop in the drain path — e.g. polling for
  in-flight RPCs to reach zero while a counter never decrements, or a producer task that is never cancelled.
- A blocked client with no error is worse than a crash: it presents as a hung query, not a failure.
- Plausibly related to the producer task structure noted in #3058: the `do_get` data plane holds long-lived
  `spawn_blocking` FEED/PARSE tasks (`scan_admission.rs`), and a long-lived blocking task cannot be cancelled by
  dropping its handle — if drain waits on those, it may wait forever. **That is a hypothesis, not a diagnosis.**

## Acceptance criteria

1. **Reproduce it deterministically** — or establish that it does not reproduce and close with the negative
   result recorded. A null result is a valid outcome here.
[...]

Issue #3058, opened 2026-07-27
cqlite-flight do_get runs k-way merge/compaction reconciliation unconditionally on a SINGLE SSTable — 6.0x slower than our own bare scan, 3.48x slower than stock Cassandra


pmcfadin:
## Summary

Measured in the #3026 (WS0) stock-Cassandra head-to-head: **`cqlite-flight`'s `do_get` data plane runs the
k-way merge / compaction-reconciliation path unconditionally — even when the table has exactly ONE SSTable and
there is nothing to merge.** It is the dominant cost on the only CQLite surface comparable to a real Cassandra
client, and it is why we lose that comparison by 3.48×.

Found by profiling, not by inspection. Parent context: #3023 (WS0 result), epic #2817.

## Evidence

Corpus: `ws0.events`, **3,999,890 rows**, LZ4 `chunk_length=16384`, 692.70 B/row uncompressed,
**single `nb` SSTable** (flushed + `nodetool compact`). Same box (c7i.4xlarge), byte-identical files
(`Data.db sha256 22d9ae224b439b2176c287a59eee6a7d1f08b4f1fafc4d2198b3da50cdce922c`).

Warm, per **physical** core (CPUs 2+10), median of 3, spread <±2%:

| Surface | rows/s | cycles/row | mem B/row |
|---|--:|--:|--:|
| CQLite **bare scan** (`execute_streaming`) | **367,760** | 17,648 | 1,933 |
| CQLite **Flight** (`do_get`) | **61,151** | 96,198 | 32,188 |
| Cassandra `SELECT *` (native protocol) | 212,981 | 30,061 | 18,008 |

- **Flight is 6.0× slower than our own bare scan** on identical bytes.
- **Flight is 3.48× slower than stock Cassandra**, while the bare scan is 1.18× *faster* than Cassandra.

### The call chain (from `perf record --call-graph=dwarf`, symbols resolved)
```
MergeProducer
  → KWayMerger::reconcile_cluster_with_overlap_counted
    → build_compaction_row_data
      → CompactionPolicy::on_data_row
```
with a **`HashMap<String, CellWriteMetadata>` constructed per CELL**.

Server-side CPU split:
- **~13% SipHash** (hashing keys into those per-cell maps)
- **~16% allocator**
- **only ~5% actual cell decode**

**It is not gRPC/Arrow-IPC framing.** Arrow encode is a separate, additive cost (measured at 59% of cycles /
[...]

Issue #3031, opened 2026-07-27
WS5: request SIZE + QUEUE DEPTH (QD is structurally 1; zero vectored reads) — latency-parallelism, not bandwidth


pmcfadin:
**#3023 WS5** — the IO lever, correctly framed as **latency-parallelism, not bandwidth**. Two multiplicative axes; we have built neither.

## The model (peer-review corrected)

```
throughput ≈ (queue_depth × request_size) / latency
```

Little's Law on the field's own `iostat` (`r_await` = **0.11 ms**): a strictly serial (QD=1) reader is capped at **1 / 0.11 ms ≈ 9,091 IOPS**. Against that ceiling every measurement we have is at or near saturation:

| Run | measured | % of QD=1 ceiling |
|---|--:|--:|
| #2866 | 7,000 r/s @ 4.45 KB → 30 MB/s | 77% |
| #2818 Arm 1 cold | 9,600 r/s @ 4.2 KB → 39 MB/s | **~100%** |
| #2818 Arm 2 cold | 4,700 r/s @ 4.4 KB → 20 MB/s | 52% |

**At QD=1 with 4.4 KB requests the theoretical max is ~39 MB/s. Arm 1 hit it.** The device was never the constraint, and `%util` 52–79% is not spare bandwidth — it is what a serial request stream looks like (device idle between completions, nothing queued).

**The fio comparison used in #2866/#2818 was not apples-to-apples:** 711.8 MB/s was `bs=1M iodepth=32` (1 MB requests, 32 in flight) vs the scan's 4.4 KB one-at-a-time — different on *both* axes. The fair denominator is `bs=4k iodepth=1` ≈ 36 MB/s, against which the read path is already ~100% efficient *for the pattern it issues*. **The entire opportunity is changing the pattern.** (WS0 #3026 records both fio numbers.)

At 0.11 ms, **either axis alone** clears the old fio figure:
- **256 KB requests at QD=1** → ~2,270 MB/s ceiling
- **4.4 KB requests at QD=32** → ~1,250 MB/s ceiling

## Confirmed in code: QD is structurally 1

- `read_compressed_offset_window_impl` (`cqlite-core/src/storage/sstable/reader/data_access/compressed_offset.rs:104-122`) loops **`for chunk_idx in first_chunk..=last_chunk`**, issuing one blocking positional read per chunk via `read_compressed_chunk_at`.
[...]

PR #3081, opened 2026-07-28
docs(#3026): WS0 stock-Cassandra read-throughput baseline — CQLite 1.13-1.24x on read, 0.29x shipping; T3 refuted

### LINKED ISSUE #3026: WS0: stock-Cassandra read throughput baseline + honest fio denominator (grounds the 0.17 roofline)
**#3023 WS0** — the measurement that grounds (or invalidates) the whole roofline model.

## Problem

There is **no cqlite-vs-Cassandra throughput comparison anywhere in this repo.** Verified: no such number in `docs/reports/`, `docs/research/`, or `docs/architecture/throughput-program-2026-07.md`. Every "we are slower than stock Cassandra" statement in the program is an intuition, not a measurement — so we cannot size the opportunity, prioritize levers against it, or ever declare done.

## Scope

1. **Head-to-head on the SAME SSTables.** Run Cassandra 5.0's own local range-scan read path over the identical `da`/`nb` files cqlite reads — same box, same page-cache state (drop caches identically for cold arms), same corpus. Report **rows/s and MB/s per physical core**, warm and cold.
2. **Fix the efficiency denominator.** Also record `fio` at the request size and queue depth cqlite *actually issues* — **`bs=4k iodepth=1`** (expected ~36 MB/s) — not only `bs=1M iodepth=32` (711.8 MB/s). Every past "% of NVMe ceiling" claim in #2866/#2818 used the `bs=1M iodepth=32` number as the denominator, which conflates request size with queue depth and reads a latency problem as a bandwidth problem. Record both, state which is which.
3. **Shape map, not a single number.** If cqlite is faster on some shapes and slower on others, that map is the useful output. Cover at minimum: narrow-ish (the M0 corpus, ~690 B/row uncompressed), and note where Cassandra's advantage concentrates.
4. **Validate or correct the roofline.** The T1/T2/T3 tiers rest on two engineering estimates: a ~929 cycles/row budget and 8 GB/s/core achievable memory bandwidth. Measure what you can of both (e.g. a memory-bandwidth microbench on the box). **If stock Cassandra is already near T1/T2 on the same files, our budget assumptions are wrong and the roofline must be re-derived from its numbers.** Say so loudly if that happens.

## Acceptance criteria

- A defensible **"cqlite is X× stock Cassandra (rows/s per physical core, warm)"** number, with the method written down in `docs/reports/` so it is re-runnable by a future agent.
- `fio` recorded at **both** `bs=4k iodepth=1` and `bs=1M iodepth=32`, with an explicit note that the former is the fair denominator for cqlite's current request pattern.
- A measured (not assumed) achievable **memory bandwidth per core** figure for the test box.
- An explicit verdict on whether the #3023 roofline tiers survive contact with the Cassandra numbers.
- If Cassandra's read path is doing something structurally different (larger requests, queue depth > 1, different buffering), **name it with a file:line citation** into `~/projects/cassandra` — that becomes a directly actionable lever.

## Notes

- Cassandra 5.0.8 source is at `/Users/pmcfadin/projects/cassandra` for reading the read path. Relevant prior art already established: CASSANDRA-15452 shipped a 256 KiB scan-only `ThreadLocalReadAheadBuffer` in 5.0.4 (`compressed_read_ahead_buffer_size`, gated `> chunkLength`), which is a **real structural advantage Cassandra has and we do not** — expect it to show up in your numbers, and quantify it.
- Cassandra's default table `chunk_length_in_kb` is **16** (`CompressionParams.DEFAULT_CHUNK_LENGTH`), matching the M0 corpus — so the comparison is like-for-like on chunking.
- This is the one workstream needing Cassandra installed alongside cqlite. Still one box.

---

### The goal
**T1 = 600k rows/s per PHYSICAL core, warm** (2.5× today's measured 238k). Unit is per-*physical*-core because #2818 Arm 2 §7 proved the C(N) plateau tracks physical cores (xlarge N=2, 2xlarge N=4) at a stable ~216–238k each — the only figure that composes to a pod.

| Tier | rows/s/phys-core | vs today | Basis |
|---|--:|--:|---|
| T0 today | 238k | 1.0× | measured |
| **T1 TARGET** | **600k** | **2.5×** | kill decode (26.7%) + allocator (19.6%) |
| T2 stretch | 1.2M | 5.0× | + zero-copy decode, fused decode→Arrow |
| T3 roofline | ~1.8M | 7.6× | **memory-bandwidth** bound |

Today = **12,597 cycles/row** vs a first-principles budget of **~929** → we are **13.6× off**. The binding roofline is **memory bandwidth (~1.8M rows/s/core at ~4.4 KB traffic/row, 8 GB/s/core)**, not cycles (~3.2M).

### Ground truth from #2818 (do NOT re-measure)
i4i, compressed corpus, LZ4 `chunk_length=16384`, server-direct, `v0.16.0` → `0.17-dev ed3ac7ef`:

| Fact | Value |
|---|---|
| single-stream warm | 199.3 MB/s / 291.7k rows/s |
| **row decode** | **26.7%** on-CPU (was 5.0%) |
| **allocator** | **19.6%** on-CPU |
| k-way merge | 6.4% — *small; #2822/#2823 closed on this evidence* |
| LZ4 decompress | 0.5% on merge thread (#2876 moved it to the producer; `stream_decompress` flat ~150 ms) |
| block reads | **99.3% in [4K,8K)**, `rareq-sz` 4.4 KB — unmoved by #2877 |
| `stream_cold_fault` | 4437.9 ms cold → 105.8 ms warm (**42× collapse**) |
| `stream_merge` / `stream_encode` | 7351 / 1868 ms, **identical cold vs warm** |
| disk knobs | none help; forcing mmap is 40% *slower* |

**Corollary: disk is NOT the warm bound.** Warm scan 11.9 s vs cold 15.9 s; if cold IO were free, cold improves ~1.39× and **warm improves 1.00×**. The tiny reads are a cold-start tax on this shape.

**Row width correction:** this corpus is **~690 B/row uncompressed** (197 B compressed, ~3.5× LZ4; Arrow wire 716 B). Earlier "narrow rows ~196 B" statements quoted the *compressed* size. These are medium rows.

### Reporting contract (non-negotiable)
1. **Report every win as cycles/row AND bytes-of-memory-traffic/row**, not just CPU%. Memory bandwidth binds before cycles, so a change that cuts CPU% without cutting traffic will stall — the percentage looks like progress while rows/s does not move.
2. **Report rows/s per physical core, warm**, as the headline. Include cold separately.
3. **A null result is a result.** #2818 Arm 2's flat read histogram is the standard: it was predicted, reported honestly, and redirected the whole program. Do not manufacture a win.
4. **Post findings as a comment on the umbrella #3023**, not only here — that issue is the single reporting surface for this exploration.
5. Every claimed multiplier must be **measured, not modeled**.

### Tooling gotchas (already paid for — do not rediscover)
- **samply records 0 samples** on HWE `7.0.0-*-aws` kernels (hit in BOTH M0 arms). Use `perf record -F 999 --call-graph=dwarf`.
- **bcc `xfsslower` will not compile** there (bcc 0.35, `struct filename` static-assert). Use `bpftrace` on `block:block_rq_complete`.
- **Never divide** `/proc/<pid>/io` `read_bytes` by `syscr` — different populations (fault-driven vs syscall). Report separately. This produced a bogus "~59 KB/syscall" in #2866.
- Local harness: self-built unstripped `cqlite-flight --features observability` + in-repo `tools/flight-loadgen` (its `Shape::Point` correctly sets token bounds). Benches: `benches/read.rs`, `decode_bench.rs`, `concurrent_scan.rs`, `partition_lookup.rs`, `open.rs`.
- Corpora: generate locally via `test-data/scripts/` so shape is a controlled variable. Set `CQLITE_DATASETS_ROOT`.

### Method doctrine
**A test written from the implementation cannot falsify the implementation.** Two bugs this cycle (#2866 IO trace, #3002 BTI root base) were invisible to green tests because the tests encoded CQLite's own assumption rather than Cassandra's behavior. Where Cassandra 5.0.8 source is the authority, derive the expectation from Cassandra.

### Process
One issue ↔ one claim ref ↔ one worktree/branch ↔ one PR (1:1:1:1). `bash scripts/flow/claim.sh claim <N>` FIRST. Run the full gate with `CQLITE_GATE_MAX_CONCURRENCY=1`. Parent: #3023. Program epic: #2817.

pmcfadin:
Closes #3026. Part of umbrella #3023, program epic #2817.

**Docs-only. No production code changed.**

## What this delivers

The **first cqlite-vs-Cassandra throughput comparison in this repo**. Before this, every "we are slower than
stock Cassandra" statement in the 0.17 program was an intuition, not a measurement.

Both engines read **byte-identical SSTables** (`sha256 22d9ae22…ce922c`, re-verified before/between/after the
arms and across a Cassandra drain+restart), same box, warm and cold, per physical core.

### The headline is two numbers, and they point in opposite directions

| Comparison | Result |
|---|---|
| **Read + materialise** | **CQLite is 1.18× stock Cassandra** (367,760 vs 311,196 rows/s — mixed pinnings, disclosed) |
| **Ship every row to a client** | **CQLite Flight is 0.29× — Cassandra is 3.50× FASTER** (60,892 vs 212,981 rows/s/phys-core) |

**This is not a CQLite win.** On the read path the advantage is a margin, not a multiple; on the honest
end-to-end surface we lose by 3.5×.

**Strategic consequence:** the 0.17 program is optimising the read path where CQLite already leads, while the
measured deficit sits in serialization/egress — which no #3023 workstream owned. Filed **#3058** (P1,
confirmed: Flight `do_get` runs k-way merge/compaction reconciliation **unconditionally on a single
SSTable**; ~13% SipHash + ~16% allocator vs ~5% actual decode), #3060, #3061.

## Roofline verdict: T3 is refuted, by measurement

| | Modeled | **Measured** |
|---|--:|--:|
| Memory bandwidth per core | 8 GB/s | **10.84** single / **10.15** all-8-cores |
| Memory traffic per row | 4.4 KB | **1,240 B** (on a >L3 corpus) |

Both inputs were wrong **in the same direction**. The scan pulls **0.435 GB/s ≈ 4.0% of one core** at IPC
2.67–3.73 — it is **instruction-throughput-bound, not memory-bound**, so the umbrella's "memory bandwidth binds
before cycles" framing does not hold. The tiers themselves survive contact with Cassandra (it sits at ~0.50× T1),
but T1 would *extend a lead we already hold*.

Suspected origin of the modeling error: 4.4 KB/row is numerically identical to the measured **disk** `rareq-sz`
— bytes per disk *request* is not bytes of DRAM traffic per *row*.

## Cassandra findings, cited and confirmed on a live node

- **256 KiB scan preads are ACTIVE on a 16 KiB-chunk table** (`CompressedChunkReader.java:236-238`; the gate is
  `262144 > 16384`). Empirically: **4,041 of 4,067 preads were exactly 262,144 B; the `[16K,32K)` bucket is
  empty.** CQLite's ~58× smaller requests are the real I/O gap — **not** queue depth, since Cassandra is QD-1 too
  (`ChannelProxy.java:141-152`, zero vectored/async I/O).
- **The ChunkCache is DISABLED by default** (`Config.java:499`), so stock Cassandra re-decompresses every chunk
  exactly like CQLite — and **enabling it disables the readahead** (`ChunkCache.java:261-265` drops `isScan`).
- **Cassandra CRC32s every chunk** (`crc_check_chance` 1.0), so it does *more* per-chunk CPU work than we do —
  per-chunk CPU cannot explain CQLite being slower.

## Honesty about what is NOT closed

**AC#2 (the honest `fio` denominator) stays OPEN.** Measured on **EBS** (5.6 MB/s @ 4k/QD1, 131.3 MB/s @
1M/QD32) because this box is `c7i.4xlarge` with **no instance store**. The i4i denominator it was meant to
correct is 5.4× / 6.4× these figures, so these **cannot** replace it. All absolutes here are c7i-measured — the
ratios transfer, the absolute rows/s do not.

## Eight contradicted premises recorded for downstream workstreams

Including: the **288,725 rows/s Cassandra reference was CLIENT-bound** (7.1× higher with concurrency);
**CQLite mmaps `Data.db`** (115 `read()` syscalls for a whole 4M-row scan), so the "99.3% of reads in [4K,8K),
4.4 KB" ground truth does not reproduce and #3031 needs that reconciled first; **`perf stat -p` costs >2%** on
the CQLite scan and would publish a 2.1×-wrong ratio; and **rows/s is not portable across row shapes** (7.8×
spread), so both arms use one fixed table.

## Review status — stated plainly

- **`--lite` PASS** on `bbb922c` (`tree-integrity: PASS`).
- **roborev: no reviewable content.** `.roborev.toml` sets `exclude_patterns = ['docs/**', '*.md']`, so a
  docs-only diff has nothing to review. Its "No issues found" is **not** a code review passing, and I am not
  presenting it as one. (Separately: CLAUDE.md's prescribed `--model gpt-5.6-sol` was rejected by codex-cli
  0.142.5; the runtime is now 0.145.0, which accepts it. Recorded in the report's tooling traps.)
- **Substituted a stronger review for a numbers-dense doc:** every figure was adversarially fact-checked against
  the raw artifacts, and every `file:line` citation re-opened. It found **six wrong figures, five unsupported
  claims, one off citation and one internal inconsistency** — all fixed in the second commit, with public
  corrections posted to #3023 and #3026. Notably the T3 proxy calibration had **no retained artifact**, so it was
  **re-measured** (it agrees: +1.08% vs +1.2%) and the artifact is now retained. **No correction reversed a
  conclusion; the largest strengthens the headline.**

## Also in this PR

`process_improvements.md` lesson 28: an agent session can **OOM-kill itself** by oversubscribing the box — and
needing to warn subagents about each other *is* the evidence of oversubscription. Plus the durable fact-checking
rule above.


pmcfadin:
## ⛔ HOLD the close — independent measurement-validity review, 7 blockers

Reviewed because this report is the evidentiary base for all of 0.17. **Verdict: the two directional conclusions are trustworthy; the specific absolute numbers are not yet safe to anchor the program on.**

Credit first, because it is the reason the gaps are findable: the self-published corrections comment caught **six** wrong figures including a "median of 3" that was a mean of 2; `perf stat -p` was found to cost >2× and every figure switched to CPU-wide `-C` (that artefact alone would have published a 2.1×-wrong ratio); the flattering surfaces were identified and *avoided* (`benches/read.rs::read/full_scan`, `db.execute()` 2.05× inflating, value-blind `flight-loadgen`); corpus identity is sha-pinned across a drain+restart; materialization proved by digest in 21/21 artifacts; and AC#2 was left **OPEN** rather than closed on wrong-hardware numbers. That is real discipline.

### 🔴 B1 — TIME-CRITICAL: nothing in this PR is reproducible, and the evidence is on an ephemeral box
`git ls-files | grep -i ws0` → **empty**. §10 (lines 407-411) points at `/home/ubuntu/ws0/...` (236 raw artefacts) and `head-to-head-method.md`, which **never landed** despite AC#1 being marked ✅. Five figures are explicitly labelled **"no raw run artefact retained"** — the 292,849/13.66s and 2,086,231 client-bound numbers (364-366), the two `r--s` mmap observation (370), 360K rows/s + ~540K context switches (384-385), 18.74s vs 11.16s unpinned (387-388), and the four-table 81.2k-629.4k spread (392-394). **#3058's P0 promotion and #3031's blocking rest on these.** When that instance terminates they become unverifiable assertions. The report's own lesson 28 says to prevent exactly this. **Commit the artifacts + method doc into `docs/reports/` before the box dies — this is the one item that cannot be fixed later.**

### 🔴 B2 — the Cassandra arms are never shown to be server-bound, and §9.1 undercuts the headline
§9.1 establishes 288,725 rows/s was **client-bound**, reproducing at 292,849 at `--inflight 1`. The headline Cassandra read figure is **311,196** — 6% *above* a demonstrated client ceiling on the same rig. No Cassandra client config is stated (driver path, inflight/concurrency, pinning, whether `count(*)` used the 512-token-range splitter), and the pinned server core is never shown saturated. **If those are client-bound floors, the 1.18× "CQLite leads the read path" — the finding that re-pointed the program — is an artifact.** §10 says "Pin, always" but never says *what* was pinned.

### 🔴 B3 — the headline 1.18× is the one mismatched-pinning ratio in the dataset
1.18× = CQLite **1-core** ÷ Cassandra **1-hw-thread**. The report discloses this and gives both matched pairs: 1-core↔1-core = **1.24×**, 1-hw-thread↔1-hw-thread = **1.13×**. The defensible claim is **"1.13-1.24×"**. Please make that the headline — the cross-pinning quotient is what propagated to the PR title, the umbrella, and #3058.

### 🔴 B4 — the memory-traffic proxy's inbound-only caveat was dropped from the report
The `perf` line counts `l2_lines_out.non_silent` (293) but it is never reported or explained. **Your own issue comment said why** — "identical for read-only and read-write kernels, so dirty writebacks are not separable on this PMU; inbound only." That sentence is absent here. So `l2_lines_in.all × 64` is an **inbound-only** proxy refuting a *total*-traffic model, unqualified on the write-heavy arms (Arrow 14,147 B/row, Flight 32,438 B/row, 2.70 GB payload). Restore the caveat. Other limitations (>L3-only, L3-refill upper bound, the quantified 56% SMT gap) are stated well.

### 🔴 B5 — the CQLite anchor's read-path config is unstated, and #3068 is never cited
§2 (44) says only "0.16.1 release + DWARF". §9.3 notes CQLite **mmapped** `Data.db` (115 `read()` syscalls total) — consistent with **#3068**'s finding that `reader/mod.rs:608-613` backs `scan_positional_source` with an mmap — but the report never states whether the build contains **#2877** (coalesced 4 MiB window) or **#2210** (MADV_RANDOM), and #3068 is absent from §12. **My explicit ask — "if any fio/iostat numbers were taken assuming the window was issuing large reads, flag it" — is unanswered.** Every 0.17 delta measures against this anchor.

### 🔴 B6 — "the tiers survive" is inconsistent with refuting T3
The only grounds given are that Cassandra sits at ~0.50× T1 — that is not evidence T1 is *achievable*. T1=600k was derived from the roofline whose T3 this report refutes at **both** inputs (bandwidth 34-35% high, traffic ~3.5× low), and no replacement ceiling is offered. Please state plainly: **T1=600k has lost its derivation.**

### 🔴 B7 — the headline comparator has no profile behind it
"Cassandra's `count(*)` does not skip cell deserialisation (verified from the profile)" (168) — but §5's profile is `SELECT *` only. No `count(*)` profile appears anywhere. The non-headline comparator gets 13 verified percentages; the headline gets an assertion. Also note the asymmetry: CQLite's bare scan builds a per-row `HashMap<Arc<str>, Value>` and digests every cell (6.6% SipHash) — work `count(*)` doesn't do. Probably conservative, but say so.

### Notable nits
- **N1** — Q1 is otherwise **answered**: every WS0 fio number carries `bs`/`iodepth`, which closes the original mispricing. But **`711.8 MB/s` (353) is quoted with no job shape** — the exact number that caused the mispricing is the one without its shape. Also missing `numjobs`/`ioengine` (at QD32 these decide whether the depth was real or impossible) and the job file. Little's-law and cap arithmetic both check out.
- **N4** — cold is **n=1** with no I/O counter proving the cache was evicted, and "cold ≈ warm within 6.3%" is in unreconciled tension with our own `stream_cold_fault` 4437.9→105.8 ms. Also page-cache-cold only: no JVM restart, so JIT/key cache/BufferPool stay warm. **Cold/warm separation itself is sound** (never blended).
- **N6** — off-CPU is **unaddressed and not declined**. `cpudist`/`off-CPU`/`#3068` appear zero times. My ask permitted declining; silence carries the program's on-CPU-only blind spot forward.
- **N2** — Cassandra `count(*)` 1-core is **4.3% slower** than 1-hw-thread while both other arms go the other way (+4.9%, +7.6%). Exceeds the stated <±2% spread, undiscussed, and it is the denominator of the matched 1.24×.
- **N3** — one aggregate "spread <±2%" for 10 warm rows, no per-arm dispersion, right after a correction round caught a mean-of-2 captioned median-of-3.
- **N5** — 0.435 GB/s ÷ 10.84 divides a **pinned** measurement by the **unpinned** ceiling (matched is 10.73); "IPC 2.67-3.73" spans both engines.
- N7 (110-row generator gap), N8 (16 GiB swapfile added **mid-run**, arms not attributed), N9 (SMT topology asserted, no `thread_siblings_list`), N10 (no literal harness/query/fio commands).

### What is safe to build on now
The **direction** — CQLite leads the bare read path ~1.1-1.2×, loses the shipping path ~3.5× — and the **T3 refutation**, the best-supported result in the document (scan pulls ~4% of one core's bandwidth at high IPC → instruction-bound; the 4.4 KB/row → disk `rareq-sz` provenance error is convincingly diagnosed). **#3058 as P0 stands regardless of B2**, because it is a CQLite-vs-CQLite gap (60,892 Flight vs 350,580 bare scan) that does not depend on the Cassandra number. The 256 KiB pread evidence (4,041/4,067 = 99.4% at exactly 262,144 B) is the strongest measurement here. All ratio arithmetic I recomputed checks out.

### To close this out
**B1 first and immediately** (artifacts + method doc committed — irreversible otherwise), then B2, B3, B5. B4/B6/B7 are text fixes. Everything else can be follow-ups.

pmcfadin:
## flow-closer: merge HELD on the owner's HOLD order — endgame stopped, certification banked

Ran the terminal stages for #3026 (WS0 stock-Cassandra read-throughput baseline). **Not merged.** The pre-merge `HOLD:` re-read found @pmcfadin's `⛔ HOLD the close` (7 blockers) posted at `2026-07-28T01:30:58Z` — *after* the certified commit and after my gate started. It is the newest comment on both this PR and #3026, so nothing lifts it. Per doctrine a HOLD is an owner-level call the closer surfaces and never overrides, so I stopped short of arming `--auto`.

Banking the work that is done so the next session does not repeat a ~25-minute gate:

- **Full gate of record: `RESULT: PASS`** at certified SHA `bbb922c4f4820b43899bc61c185653fc4505a405` — every component PASS, `tree-integrity: PASS` (`tree-start` == `tree-end` == `80ce5687c1f4`, `dirty: no`), run-id `/tmp/agent-gate.Pew8IU` verified as mine per the #2874 reader contract, `datasets: 144 Data.db` so the #2078 fail-closed did not trigger.
- **Pre-merge SHA assert: `PREMERGE: OK bbb922c4...`** — PR head still equals the certified SHA.
- **C (intent audit): skipped, not-applicable** — #3026 is an oracle/measurement issue with no OpenSpec change, so there is nothing under `openspec/changes/` to anchor a `spec-auditor` audit to.
- **roborev: `No issues found — the combined diff contains no code changes to review`.** Reporting this honestly as **no reviewable content**, *not* as a clean review: `.roborev.toml` sets `exclude_patterns = ['docs/**', '*.md']` and this diff is docs-only, so the reviewer was handed an empty diff. It carries no evidence about the report's contents.

**Note the gate PASS certifies only that the tree builds and tests green — it says nothing about measurement validity, which is exactly what the 7 blockers are about.** A docs-only diff can be fully gate-green and still not safe to anchon the program on.

### B1 is time-critical and I confirmed it is real
```
git ls-files | grep -i ws0   ->  docs/reports/ws0-cassandra-baseline-2026-07-27.md   (the report only)
/home/ubuntu/ws0/ws0-results ->  273 files, 318 MB, incl. head-to-head-method.md     (UNCOMMITTED)
```
The method doc **does** exist on the box — it was simply never committed. The artefacts are on ephemeral EC2 storage and become unverifiable assertions when the instance terminates. Sizing is a scoping call for the owner/implementer: 318 MB is too large to commit wholesale, so someone has to choose the subset (the JSON run summaries + `head-to-head-method.md` look like the load-bearing part; the 3.4 GB parent tree is corpus, not evidence).

Claim `refs/claims/issue-3026` stays **held** and the worktree stays in place — an unfinished endgame keeps its lane rather than being orphaned. No telemetry stamped (no delivery occurred), board left at `In Progress` since the issue needs further work rather than review, and #3026 remains open. Umbrella #3023 and the backlog items #3058/#3060/#3061 untouched.

pmcfadin:
## HOLD accepted — B1 landed immediately (`196fef80`); B2–B7 in progress

The review is correct and the closer was right to refuse the merge. Taking the blockers in your stated order.

### ✅ B1 — CLOSED, pushed as `196fef80`. This was my error, and the worst kind.

I marked **AC#1 ("re-runnable by a future agent") satisfied while citing `/home/ubuntu/ws0/...` paths on
ephemeral EC2 storage.** `head-to-head-method.md` existed on the box and was simply never committed. The
report's own lesson 28 warns about exactly this class of mistake, which makes it worse, not better.

Committed **307 files / 1.5 MB** at `docs/reports/ws0-3026-artifacts/`:
- `ws0-results/head-to-head-method.md` — the method doc, including the F1–F7 notes the report cites
- every per-run `summary-*.json` / `scan-*.json` / `*.jsonl` behind §4's rows/s, cycles/row, IPC and
  `l2_lines_in.all` figures — so the `rep-*` triples can now be re-medianed independently
- all four `fio` JSONs + per-second bandwidth/latency logs
- `membw-calibration.txt` (the re-measured §7 proxy validation)
- `stream-*/readbw-*` outputs + their C sources
- **`ws0-corpus/` in full** — `rerun.sh`, `gen-corpus.sh`, profile, `schema-as-created.cql`,
  `cassandra.yaml.diff`, `trace-scan.bt`, `verify-claims.sh`, `corpus-geometry.txt`, `claims-evidence.txt`

**Scoping, stated so it can be challenged:** three perf/jstack reports were ~300 MB of the 318 MB tree
(`prof-rows` 161 MB, `prof-count` 121 MB, `jstack-rows` 17.7 MB). I committed **top-400/top-300 truncations**
that carry the ranked frames §5 cites. **The truncation is a real limitation** — a reader cannot re-aggregate
a different view (per-thread, per-callee) from them — and the README says so rather than implying the profiles
are preserved. Also excluded: perf `.data` binaries, the 1.3 GB harness build tree, and the 754 MB corpus
(reproducible via `rerun.sh`, identity pinned by the sha256 in `corpus-geometry.txt`).

The README **flags rather than hides** the five figures labelled "no raw run artefact retained" — the
client-bound 292,849 / 2,086,231, the mmap observation, the `perf stat -p` artefact, unpinned-vs-pinned, and
the four-table spread — naming them the weakest-evidenced claims and the first things to regenerate. You are
right that #3058's priority and #3031's blocking status lean on some of these.

### In progress, your order: B2, B3, B5, then B4/B6/B7 + nits

- **B3** — agreed and it is the cleanest fix: `1.18×` is the single mismatched-pinning quotient in the dataset
  (CQLite 1-core ÷ Cassandra 1-hw-thread). Headline becomes **`1.13–1.24×`**, propagated to §1/§7/§12, the PR
  title and the umbrella.
- **B2** — this is the one that could invalidate the read-path lead, and I am treating it that way. The
  instruction is explicit: if the artifacts cannot establish the Cassandra arms were **server**-bound, the
  report will say so and carry a caveat that the figures may be client-bound floors making the lead an
  artifact. **No number gets strengthened to satisfy a blocker.** Your point that 311,196 sits only 6% above a
  demonstrated 292,849 client ceiling on the same rig is the sharpest observation in the review.
- **B5** — will state whether #2877 / #2210 are in the 0.16.1 anchor or mark it undeterminable, add **#3068**
  to §12, and answer your unanswered ask directly: given §9.3 shows mmap with ~no data-path read syscalls,
  spell out what the fio numbers do and do not bear on.
- **B4** — the inbound-only caveat was in my issue comment and dropped from the report. Restoring it, with the
  point that an **inbound-only** proxy is being used against a **total**-traffic model and is unqualified on
  the write-heavy arms.
- **B6** — accepted without reservation. "The tiers survive" was incoherent next to refuting T3 at both
  inputs. It will say plainly: **T1 = 600k has lost its derivation.**
- **B7** — `prof-count.top400.txt` is now committed; if it substantiates the `count(*)` claim I will cite it
  properly, and if not the claim gets downgraded to a labelled assertion. Either way the asymmetry gets stated
  explicitly (CQLite's bare scan builds a per-row `HashMap<Arc<str>, Value>` and digests every cell — 6.6%
  SipHash — work `count(*)` does not do, so probably conservative).
- **Nits**: all in scope, including **N6** — off-CPU will be addressed or **explicitly declined** with the
  on-CPU-only blind spot named, since silence was the objection.

### One process note

The gate PASS at `bbb922c4` no longer covers the new tip, so the **full gate will be re-run from Stage 1**
before any merge. And to be explicit about something the closer flagged: a gate PASS certifies that the tree
builds and tests green — it says **nothing** about measurement validity, which is what all seven blockers are
about. A docs-only PR can be fully gate-green and still unsafe to anchor a program on.

Nothing here is merged, the claim stays held, and the board stays `In Progress` rather than `In Review`.

pmcfadin:
## All 7 blockers + 10 nits resolved — `442ec3e0`. B2 came back YES.

### 🟢 B2 — VERDICT: the Cassandra arms WERE server-bound. The read-path lead is not a client artifact.

This was the one that could have invalidated the headline, so it got the hardest look. Every reported
Cassandra arm carries a measured **`daemon_core_utilization` of 0.9973–1.0182** and **`"client_bound": false`**.

Config, now stated in the report (it was missing entirely — your point stood):
- DataStax **Python** driver, protocol **v5**, `fetch_size=5000`
- **`--inflight 8` × 6 shard processes = 48 queries in flight**
- **512 token ranges in BOTH modes** — so yes, `count(*)` used the splitter
- daemon pinned `taskset -acp` on CPU `2` or `2,10`; clients on CPUs 4,5,6,7,12,13

And the **negative control was already in the artifacts**: the unsharded single client reached only
**0.35 / 0.63** utilisation at 94k rows/s. So §9.1's 288,725 does **not** bound 311,196 — different client,
different pinning. **The 6% proximity you spotted is a coincidence**, and it was the right thing to challenge;
it just resolves in the measurement's favour.

### 🟡 B5 — both levers are IN, and this explains #3061

**"0.16.1" was misleading and I should have caught it.** That is the *version string*; the measured build is
`main`@**`d0344b76`**, **78 commits ahead of the `v0.16.1` tag**.
- **#2210** (`f5e18390`) — in the tag.
- **#2877** (`c0241415`) and **#2876** (`bc14ff0e`) — landed **2026-07-25, after the tag**, so they are in the
  measured build but *not* in `v0.16.1`.

**The 8 MiB `MADV_RANDOM` gate is exceeded, so a second mapping is created by design — which explains
#3061's "Data.db mapped twice."** That is not a leak; I'll re-scope #3061 to the RSS-measurement question only.
**#3068 added to §12.**

Your fio/mmap ask, answered directly: `rerun.sh` literally labels its 4k/QD1 arm *"CQLite's real pattern
(~4.4 KB reads at QD1)"* — which §9.3 **contradicts** (mmap, 115 `read()` syscalls total). Flagged in the
report. **No report figure divided a CQLite rate by an fio number**, so nothing needed retracting — but #3031
is explicitly warned off that assumption.

### ✅ B3 · B4 · B6 · B7
- **B3** — headline is now **`1.13–1.24×`**. 1.18× is named as the one non-like-for-like combination and is no
  longer quoted as a figure (§1, §4, §7, §11). PR title updated.
- **B4** — inbound-only caveat restored: `l2_lines_out.non_silent` is identical for read-only and read-write
  kernels, so **dirty writebacks are not separable on this PMU**. An inbound-only proxy is therefore being used
  against a *total*-traffic model, unqualified on the write-heavy arms.
- **B6** — accepted in full. "The tiers survive" is gone; §7 now answers **NO: T1 = 600k has lost its
  derivation.** AC#4 restated.
- **B7** — **claim downgraded.** `prof-count.top400.txt` does *not* substantiate it. Now a labelled assertion,
  with the asymmetry stated: CQLite's bare scan builds a per-row `HashMap<Arc<str>, Value>` and digests every
  cell (6.6% SipHash), work `count(*)` does not do — so probably conservative, and it says so.

**N1–N10 all fixed**, including **N6 by explicit declination** of off-CPU with three named blind spots, since
silence was the objection.

---

## Six things this round found that neither of us had

1. **§5 has no committed backing at all — not just `count(*)`.** All three `h2h-perf-summaries/*` files contain
   **zero `%` characters**; they are head-truncations of raw SJK/`jstack` thread dumps, not perf reports. **My
   artifacts README claim that they "back the §5 percentages" was false**, and is corrected in both places.
   Every §5 percentage and method-doc F2 are now labelled **unretained**. **#3058's rationale is untouched** —
   it rests on the artifact-backed 60,892-vs-350,580 rows/s and 96,260-vs-17,648 cyc/row gaps, not on any share.
2. **Your N4 "no I/O counter proving eviction" was wrong** — `read_bytes` is **0 on every warm arm** and
   **783.9–793.2 MB on every cold arm**. The proof existed; the report just never cited it. The n=1 and
   no-JVM-restart legs stand and are now stated.
3. **The box rebooted mid-session** (prior boot ended 16:00:05Z, current began 16:58:36Z). **All four fio arms
   ran on the earlier boot; every head-to-head arm on the current one.** So §8 and §4 differ by *kernel
   session*, not merely by the swapfile — a stronger caveat than N8 asked for.
4. **The Flight arm is the ONLY arm whose pinned set was not saturated** (`0.844–0.866` vs ~1.00 for Cassandra).
   Added as a precision caveat on **3.50×** — explicitly **not** on #3058.
5. **Two Flight 1-hw-thread replicates failed to CONNECT and were silently dropped**, so that row is **n=1**
   from a single 360 s step. The method doc's "median of 3, spread <±2%" was wrong on two counts; erratum added.
   The empty `.jsonl` files are retained deliberately as evidence of the failure.
6. **Client shards were not on 6 distinct physical cores** (4,5,6,7,12,13 → cores 4 and 5 ran both siblings).
   Immaterial given server utilisation ~1.00, but now stated accurately.

New evidence file `ws0-results/session-timeline.txt` captures the kernel-log OOM/swapon/reboot lines and
`thread_siblings_list` — otherwise N8 and N9 would have become unretained the moment this instance dies.

**No claim was strengthened to clear a blocker.** The T3 refutation, the 256 KiB pread evidence (4,041/4,067)
and #3058's rationale are untouched or explicitly reinforced as independent of the weakened material.

`--lite` is re-running now, then the **full gate from Stage 1** (the `bbb922c4` PASS no longer covers this
tip), then back to `flow-closer`. Claim held, board still `In Progress`.

pmcfadin:
## ✅ HOLD LIFTED by the owner — merge authorized on gate-green

Owner instruction, this session: **"merge it when the gate passes."** Recording it here so the lift is
traceable and so the closer's mandatory `HOLD:` re-read sees an explicit, dated release rather than inferring
one from silence.

**What this authorizes:** arm the merge once the full gate of record PASSes on `442ec3e0` — the tip that
contains the B1 evidence commit and the B2–B7 + N1–N10 resolutions. It does **not** authorize merging the
earlier certified tip `bbb922c4`, whose PASS no longer covers this tree.

State at the time of the lift:
- **All 7 blockers resolved**, all 10 nits fixed — see the resolution comment above.
- **B2 resolved affirmatively**: the Cassandra arms were server-bound (`daemon_core_utilization` 0.9973–1.0182,
  `"client_bound": false`, with a 0.35/0.63 negative control in the artifacts). The read-path lead is not a
  client-bound artifact.
- **Headline is now the defensible range `1.13–1.24×`**; the mismatched-pinning 1.18× is no longer quoted as a
  figure anywhere, including the PR title.
- **`T1 = 600k has lost its derivation`** is now stated plainly (B6), replacing the incoherent "the tiers
  survive".
- **B1 evidence is committed** — 307 files at `docs/reports/ws0-3026-artifacts/`, so nothing depends on
  ephemeral storage any more.
- **`--lite` PASS** at `442ec3e0` (`tree-integrity: PASS`).

Carried forward unchanged, because merging does not settle them:
- **AC#2 stays OPEN.** The honest `fio` denominator cannot be measured on an EBS-only `c7i`; it needs
  `c7id`/`i4i`. Both 711.8 MB/s and 36 MB/s are labelled shape-unknown and explicitly not usable as denominators.
- **§5's percentages and method-doc F2 remain unretained** — the substitute files do not contain them. #3058's
  priority does not depend on them.
- **#3058 / #3060 / #3061 stay at board `Backlog`** for owner prioritisation. #3061 has been re-scoped by the
  B5 finding (the double mapping is designed `MADV_RANDOM` behaviour) and now reduces to the
  `RssAnon`-vs-`RssFile` question.
- **roborev carries no evidence about this diff** — `.roborev.toml` excludes `docs/**` and `*.md`, so the
  reviewer was handed an empty diff. Not being presented as a clean review.

Next: full gate on `442ec3e0` → `premerge-assert.sh` against that SHA → `gh pr merge --auto --squash
--delete-branch` → finalize (no `openspec archive`; this is a measurement issue with no OpenSpec change).

Issue #3068, opened 2026-07-27
Scan window's 4 MiB coalesced refill never becomes a large I/O — COLD-scan lever (~1.39x cold, ~1.00x warm); scan_positional_source is mmap so QD stays 1 and rareq-sz stays 4.4 KB


pmcfadin:
## Summary

The userspace read buffer and the read-intent split **both already shipped** (#2877's `CompressedScanWindow`, #2876's `scan_positional_source`), but the scan window is bolted to a backend that **cannot issue a large read**. That is why #2818 Arm 2 measured `rareq-sz` **unmoved at 4.4 KB after #2877 landed** — an unexplained result until now.

## The defect

`cqlite-core/src/storage/sstable/reader/mod.rs:608-613`:

```rust
let scan_positional_source: Arc<dyn read_at::ReadAt> = match &scan_source {
    ScanSource::Mapped(mmap) => Arc::new(read_at::MmapReadAt::new(mmap.clone())),
    #[cfg(unix)]
    ScanSource::Direct { .. } => point_source.clone(),
    ScanSource::Buffered { .. } => point_source.clone(),
};
```

- Under **`Mapped`**, the coalesced window's "4 MiB refill" is **slice indexing into an mmap** — no syscall, no `pread`. The kernel services it as page faults at whatever readahead that mapping gets. The window coalesces *logically*; the substrate cannot express a large request. **QD stays 1 and request size stays one page.**
- Under **`Direct`/`Buffered`**, `scan_positional_source` is `point_source` — the `MADV_RANDOM` plane. Benign today (no per-mapping advice for those backends) but it means the read-intent split is only genuinely honored on the mmap branch.

So `COMPRESSED_SCAN_WINDOW_TARGET_BYTES = 4 MiB` is currently a **logical** window only. Nothing downstream turns it into one large I/O.

## Scope

1. **Establish the shipped-binary backend first.** `SSTableReaderConfig::default()` has `use_mmap: false` ("buffered I/O is the portable, safe default"), yet #2818 Arm 2 reports "forcing mmap is 40% slower" — you cannot force what is already on. One of those is wrong about the field binary. Resolve with one `strace -c` / debug line before changing anything.
[...]

Issue #2876, opened 2026-07-25
Scan path reads Data.db through the MADV_RANDOM point mmap → ~4KB page-at-a-time IO (root cause of #2866 tiny reads; #2210 × #1940 cross-path regression)


pmcfadin:
## Summary

The field IO trace in #2866 (97% of a full-table Flight scan's block reads land in the 4–8 KB bucket; `iostat rareq-sz` ~4.4 KB at ~7000 reads/s → ~31 MB/s with the NVMe only ~79% utilized, vs 711.8 MB/s fio sequential) has a **located mechanism in code**, and it is not Arrow serialization, not disk, and not "missing readahead" in the sense first assumed.

**The Summary-guided compressed scan walk reads Data.db through the dedicated `MADV_RANDOM` point-read mapping, one positional read per partition. `MADV_RANDOM` exists precisely to suppress kernel readahead, so each small partition body faults ~one 4 KiB page.**

This is a silent cross-path regression: #2210 optimized the *point-read* plane, but #1940 had already repointed the *scan* walk at that same plane.

## Mechanism (verified call chain)

1. Flight `do_get` row path → `KWayMerger::new_from_readers` → `SSTableReader::stream_all_partitions_for_query` (`reader/data_access/summary_scan.rs:432`).
2. Compressed `nb` → `stream_partitions_summary_guided_compaction` → **`walk_in_range_partition_slices`** (`summary_scan.rs:123`).
3. Per partition, it reads **only that partition's span** — `size = successor_offset - offset`, tens-to-hundreds of bytes for narrow `keyvalue` rows:
   ```rust
   // summary_scan.rs:264-269
   let raw = if let Some(ci) = self.compression_info.as_deref() {
       self.read_compressed_offset_window(ci, start, size).await?   // ONE call PER PARTITION
   ```
4. `read_compressed_offset_window` passes **`self.point_source`** (`reader/data_access/compressed_offset.rs:44`) — note the module's own header calls itself "the compressed offset-read **point-lookup** path" (`compressed_offset.rs:3-9`).
5. `point_source` for an mmap-backed reader is a **second mapping advised `MADV_RANDOM`** whenever `file_size >= POINT_MMAP_MADV_RANDOM_MIN_BYTES` = 8 MiB (`reader/mod.rs:571-579`, `mod.rs:343`, `point_read_mmap` at `mod.rs:1157-1185`). Field Data.db files are 0.3–1.4 GB → **every one is advised**.
[...]

Issue #2084, opened 2026-07-06
flow-closer: per-issue subagent owning full-gate → C → roborev → merge (kills lead context accretion + #1855 tension)


pmcfadin:
Parent epic: #2083. From audit R1 (context-economy synthesis F2/F4, docs/reports/agentic-workflow-audit-2026-07-06.md).

## Problem
flow-implement steps 6+8 run the full gate and roborev IN THE PERSISTENT LEAD CONTEXT — the two largest text streams in the pipeline (gate stdout thousands of lines × median 1.5/p90 3/max 9 runs; roborev up to 40 findings across up to 9 rounds). The doctrine's own step 4 forbids exactly this. Cause: a subagent idle-waiting on a 12–25 min gate is killed by the 600s stall watchdog and orphans the gate process (#1855) — so the lead does it itself.

## Change (design-driven — run flow-activate / OpenSpec at pickup)
New `.claude/agents/flow-closer.md` (short-lived, spawned per issue by flow-implement) owning the terminal stages:
1. Runs the FULL gate via `run_in_background` (turn ends; harness re-invokes on gate exit — no idle-wait, so the #1855 watchdog kill cannot occur) with the #2079 summary-file pattern
2. Spawns spec-auditor (C) for design-routed issues
3. Drives roborev to clean (fix rounds per the scoped-re-verify + severity doctrine children of #2083)
4. Merges on green per the worker-merges-own-PR model, then flow-finalizes
5. Returns ONLY a terminal packet: {verdict, PR URL, summary-file path, ≤10 lines residual findings}

All gate stdout, diffs, and review churn die with the closer's context.

## Design questions for the OpenSpec pass
- Closer vs implementer boundary: who fixes roborev findings (closer respawns implementer vs fixes inline)?
- Idle-wait prohibition: run_in_background mandatory vs grep-poll fallback — spell out the failure mode if violated (#1855 recurrence)
- Interaction with the machine-wide full-gate serialization rule (#1930) when closer + next lane overlap

## Acceptance criteria
- [ ] Agent definition exists; flow-implement SKILL.md delegates steps 6–8 (+merge/finalize) to it
- [ ] Closer brief FORBIDS idle-waiting on the gate; background invocation is the required path
- [ ] Terminal packet contract (≤10 lines + paths) stated in both closer brief and flow-implement
- [ ] A full issue driven end-to-end via the closer: lead context receives only the terminal packet (verify by transcript inspection)
[...]

PR #3090, opened 2026-07-28
docs(#3068): correct stale disk-access-mode + scan-advice comments

pmcfadin:
Refs #3068 (track "C" — doc-only; #3068 stays open for the perf lever).

**Comments/docs only. Zero behavior change, zero signature change, zero logic change.**

## The defect class: comments that document a path nobody takes

`use_mmap` does **not** select the read backend. `config.storage.disk_access_mode` defaults to
`DiskAccessMode::Auto` (`cqlite-core/src/config.rs:126`), and `resolve_disk_access_mode`
(`reader/mod.rs:245-282`) maps `Auto` + `size >= mmap_min_size_bytes` (4 KiB) + `size <= 0.5 * RAM`
to `Mmap`. `use_mmap` is folded in at `reader/mod.rs:532-537` only as a **promotion** of an
*explicit* `Buffered` request to `Mmap` — it can never produce the buffered path. Every field reader
construction (cqlite-cli, cqlite-flight, python/node bindings) uses `Config::default()` => `Auto`
=> `Mmap`. The only forced-`Buffered` sites are the write/compaction merge path
(`write_engine/merge/mod.rs:758-770`, `merge/from_readers.rs:39`, `merge/point_read.rs:467`).

Evidence (strace, release `cqlite read-sstable`, 647 KB Data.db, 1000-row scan):
`openat -> mmap(647164, MAP_SHARED) -> close`, and **zero** `read`/`pread64` syscalls against
Data.db. The buffered path documented in these comments is not executed.

Comments that describe a configuration nobody runs are worse than no comment: per the owner, the
"SCAN mapping is NEVER advised" line below is why #2876 stayed hidden for five days.

## The three corrections

1. **`cqlite-core/src/storage/sstable/reader/types.rs`** (`SSTableReaderConfig::use_mmap`) — said
   "Opt-in; buffered I/O is the portable, safe default", describing a field configuration that does
   not exist. Now states that the flag is promote-only (explicit `Buffered` -> `Mmap`), that `false`
   does not mean buffered, and that the backend comes from `storage.disk_access_mode` (`Auto`
   default: buffered below 4 KiB, mmap up to `direct_io_memory_fraction` of RAM, direct above),
   pointing at `resolve_disk_access_mode`. The `mmap_min_size_bytes` doc and the `Default` inline
   comment were corrected to match.

2. **`cqlite-core/src/storage/sstable/reader/mod.rs`** (`POINT_MMAP_MADV_RANDOM_MIN_BYTES`) — said
   "The SCAN mapping is NEVER advised", technically true only of `MADV_RANDOM`. Now says what is
   true: the scan mapping never gets `MADV_RANDOM`; the default `PrefetchMode::Auto` leaves it
   unadvised entirely (#1143) while an explicit `Sequential`/`WillNeed` does advise it
   (`build_block_sources`, `reader/mod.rs:1093`); and that the mapping backs BOTH
   `BlockSource::Mapped` and `scan_positional_source`, with the windowed / Summary-guided scan feed
   reading through the latter — **not** through `BlockSource` — so a `BlockSource`-scoped claim
   describes nothing a real scan does (#2876). The duplicate `#1143: the scan mapping is never
   advised` parenthetical at the `scan_positional_source` construction site was tightened the same
   way.

3. **`cqlite-core/src/storage/sstable/reader/source.rs`** (module doc, where a reader of the two
   `BufReader::new` constructors will see it) — records the dropped observation: both `buffered`
   and `buffered_sized` use `BufReader::new`, i.e. tokio's 8 KiB `DEFAULT_BUF_SIZE`;
   `BufReader::with_capacity` appears nowhere in `cqlite-core/src`; and
   `SSTableReaderConfig::read_buffer_size` (64 KiB) is a `block_io.rs` scratch /
   chunk-materialization cap (lines 281, 825, 833, 942), **not** a reader capacity. Scoped as an
   observation, not a defect: buffered is not the resolved default read backend, so the 8 KiB refill
   reaches only the explicitly-buffered merge readers. **Buffer sizes are unchanged** — explicitly
   out of scope per the owner.

Also corrected in passing (same defect class, same doc, in-place, no line growth):
`cqlite-core/src/config.rs` said the legacy `use_mmap` flag "still forces mmap when `Auto` would
otherwise pick buffered" — it does not; `Auto` is left untouched by the promotion. Two adjacent
"defaults to buffered I/O" phrasings were corrected to "no promotion".

## Verification

- All claims re-verified against the files (not from a prior read) before editing; all three held.
- `cargo doc --no-deps -p cqlite-core` with `-D rustdoc::broken_intra_doc_links`: 66 errors before
  and 66 after (all pre-existing, none in the touched files) — the new intra-doc links resolve.
- Campsite rule: `mod.rs` (1721), `source.rs` (835) and `config.rs` (1133) are over the ~800
  threshold, so their comment rewrites were made **line-count-neutral** (each file is at exactly its
  `origin/main` line count). `CQLITE_ALLOW_FILE_GROWTH` was not needed.

## Gate

`AGENT-GATE LITE SUMMARY` — RESULT: PASS at 983bc82 (file-size PASS, fmt PASS, clippy PASS,
roborev-lints PASS, scoped-tests PASS; tree-integrity PASS). The full gate of record runs once
pre-merge in `flow-closer`.

### DIFF
--- a/cqlite-core/src/config.rs
+++ b/cqlite-core/src/config.rs
@@ -53,14 +53,14 @@ pub struct StorageConfig {
     /// Sync mode for durability
     pub sync_mode: SyncMode,
 
-    /// Memory-map SSTable Data.db files instead of using buffered file I/O.
+    /// Legacy promote-only flag: it upgrades an **explicit**
+    /// [`DiskAccessMode::Buffered`] request to [`DiskAccessMode::Mmap`].
     ///
-    /// **Opt-in.** Defaults to `false` (buffered I/O), which is portable and
-    /// safe on every filesystem. When enabled, the reader maps Data.db files at
-    /// or above [`Self::mmap_min_size_bytes`] into the process address space and
-    /// serves reads from the OS page cache with no per-block `read` syscall,
-    /// mirroring Cassandra's `disk_access_mode: mmap`. This speeds up repeated
-    /// local scans of the same files.
+    /// It does **not** select the backend — [`Self::disk_access_mode`] does, and its
+    /// `Auto` default already memory-maps most Data.db files (see that field). So
+    /// `false` does not mean "buffered I/O", and `true` changes nothing unless
+    /// something explicitly requested `Buffered`. A mapped file is served from the
+    /// page cache with no per-block `read` syscall, as Cassandra's mmap mode does.
     ///
     /// # Safety / platform constraints
     ///
@@ -76,10 +76,10 @@ pub struct StorageConfig {
     ///
     /// # Interaction with the write engine (Issue #591)
     ///
-    /// This setting only affects the read path. Compaction always reads its
-    /// input SSTables through buffered I/O regardless of `use_mmap`, and deletes
-    /// each input by removing its `TOC.txt` first (unpublishing it) before the
-    /// data components, best-effort. So enabling mmap for queries is safe
+    /// This setting only affects the read path. Compaction's input readers force
+    /// `use_mmap = false` + explicit `Buffered` (only `CQLITE_USE_MMAP=1` promotes even
+    /// those); each input is unpublished by removing its `TOC.txt` before the data
+    /// components, best-effort. So enabling mmap for queries is safe
     /// alongside background compaction: a compaction never holds a mapping over a
     /// file it then deletes, and on Windows a data file still pinned by a mapped
     /// reader becomes an invisible orphan (reclaimed on the next startup) rather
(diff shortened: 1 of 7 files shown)


pmcfadin:
Endgame (flow-closer, #2084):
- **Full gate of record: PASS** — run-id `/tmp/agent-gate.t9Mvwe`, commit `c4e62a2`, `tree-integrity: PASS` (tree-start == tree-end == `c4e62a261f3b`, dirty: no), 144 Data.db files under the main checkout's datasets root (no 0-row false pass). All 29 components PASS.
- **C intent audit: skipped** — no OpenSpec change for this PR (doc-only correction, not design-driven work with a spec). Not silently omitted.
- **Final roborev: 1 finding, triaged NIT** (0 blockers). Comment/doc polish at mod.rs:602-607 re: the sub-8MiB shared-mapping case; the comment is not factually wrong ("deliberately `MADV_RANDOM`" is load-bearing) and the fallback is documented accurately at `point_read_mmap`. Batched to #3105 per `docs/development/roborev-severity.md`.
- **Pre-merge assert**: `PREMERGE: OK c4e62a261f3b999d6536b29b2a7321133581d975`. No `HOLD:` order on PR or issue. No rebase needed — `MERGEABLE` (not `BEHIND`) and zero file overlap with the 13 commits main gained since the merge-base, so the gate of record still covers the head.

Auto-merge armed. Note this PR uses `Refs #3068`, not `Closes` — **#3068 stays open** for the perf lever.

Issue #3098, opened 2026-07-28
flow-nits(#3058): read_path_probe cacheline/compaction reach + on_merger_built phase skew between the two read arms


pmcfadin:
Batched nits from the #3058 roborev round (single-SSTable Flight merge bypass). Neither is a
correctness defect; both are deferred deliberately and recorded here so they are not lost.

## 1. `read_path_probe` counters are reachable from the compaction WRITE path and share a cacheline

`KWayMerger::reconcile_cluster_with_overlap_counted` (`cqlite-core/src/storage/write_engine/merge/mod.rs`,
top of the fn) records `read_path_probe::record_reconcile_entry()` once per clustering group. That
function is also the compaction reconciler, so a compaction pays one relaxed `fetch_add` per group,
and the module's three `AtomicU64` statics are adjacent (same cacheline) — N concurrent
merges/compactions contend on one line.

Options: `#[repr(align(64))]` padding per counter; or split the always-on read-arm counters from the
compaction-reachable one; or gate the reconcile counter behind the existing `work-counters` feature
and have the Flight tests enable it.

Acceptance: a compaction-throughput microbenchmark shows no measurable regression from the probe, or
the probe is moved off the compaction-reachable path. The #3058 path-taken assertions must keep
working (they need `reconcile_entries` to be observable from a separate integration-test crate).

## 2. `on_merger_built()` fires at a different point on the two arms

`cqlite-flight/src/producer_warm.rs`: the merge arm fires the `merge_setup -> stream` phase boundary
right after `KWayMerger::new_from_readers` returns, i.e. before any row is decoded. The fast arm
fires it after `ScanRowSource::open`, which PRIMES the walk (it must, to see a pre-emit
`Unsupported` before falling back), so the first batch's decode time is attributed to `merge_setup`
on the fast arm and to `stream` on the merge arm.

Consequence: the `cqlite.rpc.phase.duration` split is not comparable across arms; no rows or
statuses are affected.

Acceptance: the two arms attribute the first batch to the same phase (either by firing the fast
arm's boundary before priming and tolerating a fallback that has already fired it, or by teaching
the phase timer about a "primed" sub-phase), pinned by a metrics-capture assertion.

pmcfadin:
Two more nits from the #3058 reviewing pass (deferred deliberately; do not block that PR):

## 3. `ScanRowSource::decorate()` clones a `DecoratedKey` per row

[...]

Issue #3097, opened 2026-07-28
stream_all_partitions_for_query ignores the caller's authoritative schema, so nb-header-less clustering columns decode as NULL on the merge arm


pmcfadin:
## Summary

`SSTableReader::stream_all_partitions_for_query` — the **warm k-way merge arm's** per-SSTable
enumeration for Flight `do_get` (`from_readers::drive_query_stream`) — resolves its decode schema
from the reader's own four-tier lookup (`get_table_schema(None)`) and IGNORES the caller's
authoritative schema. An `nb` SSTable header carries no embedded schema, so a table whose ticket DDL
declares clustering keys decodes those columns as **NULL** on the merge arm.

Found while building the #3058 forced-path differential: the single-source fast path threads the
caller's schema and returns `ck` correctly; the merge arm, over the SAME bytes, returned
`{"ck": null, ...}`. The fast path is fixed under #3058; the merge arm is deliberately left as-is
there (spec R4 requires the multi-source path to be unchanged), with a comment at the call site
pointing here.

## Repro

Run any `test_compaction_tombstone_ttl` fixture (real Cassandra `nb`) through Flight `do_get` with
`CQLITE_FLIGHT_MERGE_PATH=merge` and compare against `test-data/query-semantics-oracle.json`: the
clustering column comes back null. With `=bypass` (single source) it is correct.

## Root-cause hypothesis

`stream_partitions_summary_guided` took no caller schema at all before #3058; #3058 added a
`caller_schema` parameter but the merge-arm call site passes `None` on purpose. The compaction
sibling `stream_partitions_summary_guided_compaction` already takes and honours `schema`, so the
asymmetry is in the ScanRow-emitting walk plus its `stream_all_partitions_for_query` caller.

## Affected

- `cqlite-core/src/storage/sstable/reader/data_access/summary_scan/mod.rs`
  (`stream_partitions_summary_guided`, `stream_all_partitions_for_query` — the `None` call site
  carries the comment referencing this issue)
- `cqlite-core/src/storage/sstable/reader/data_access/full_index_stream.rs`
  (`stream_all_partitions_via_full_index`, same reader-derived resolution)
- consumer: `cqlite-core/src/storage/write_engine/merge/from_readers.rs::drive_query_stream`

## Routing

Oracle-driven bug fix: GitHub issue + pinned parity test. No OpenSpec change.

## Acceptance

[...]

Issue #3095, opened 2026-07-28
Flight do_get returns wrong rows for STATIC columns on the merge arm: phantom ck=null row + statics null on real rows (verified vs cassandra-5.0.8)


pmcfadin:
## Summary

On a table with **static columns**, `cqlite-flight`'s `do_get` **merge arm — the production path today** — returns a result set that does not match Cassandra:

- it emits a **phantom `ck = null` row** for every static-bearing partition (N+1 rows where Cassandra returns N), **and**
- it returns the static column as **null on the actual clustering rows** (no static injection).

Found while delivering #3058 (single-SSTable merge bypass), which put two independent implementations side by side and made the divergence observable. **This is PRE-EXISTING on `main` and is NOT introduced by #3058** — it reproduces under `CQLITE_FLIGHT_MERGE_PATH=merge`, which is exactly pre-#3058 behavior.

## Authority — verified against the pinned `cassandra-5.0.8` tag

Per CLAUDE.md's authority order (pinned Cassandra source first; a CQLite `file:line` is never format authority). Blobs read at the pinned tag:

`src/java/org/apache/cassandra/cql3/statements/SelectStatement.java`, `processPartition()`, L1089-1152:

| Case | Cassandra 5.0.8 | Evidence |
|---|---|---|
| static + N clustering rows | **N rows**, static merged into **each** | per-row loop `while (partition.hasNext())` (L1123) → `result.newRow(..., row.clustering(), ...)` (L1126); `case STATIC: result.add(staticRow.getColumnData(def), nowInSec)` (L1148-1149) from the partition-level `staticRow` captured at L1097 |
| static-only partition | **1 row** (null clustering + null regular), for bare `SELECT *` | L1099-1120: `if (!partition.hasNext())` emits one row iff `!staticRow.isEmpty() && restrictions.returnStaticContentOnPartitionWithNoRows()`; `default: result.add((ByteBuffer)null)` |

The static row is delivered **out-of-band** via `partition.staticRow()` (`db/rows/BaseRowIterator.java` L54-58) and is never an element of the iteration, so `hasNext()` counts clustering rows only. The zero-row branch `return`s at L1119, making the two cases **mutually exclusive**.

[...]

Issue #3094, opened 2026-07-28
Flight do_get fails the stream when a CQLite-written cell tombstone reaches the Arrow encoder — "expected Text value, got Tombstone"


pmcfadin:
## Summary

A CQLite-written **simple cell tombstone** (a `CellOperation::Delete` on one text column of an
otherwise-live clustering row) survives the read path as a raw `Value::Tombstone` in the row
carrier. When that carrier reaches the Arrow encoder, `do_get` **errors the whole Flight stream**:

```
Internal: column 'w': expected Text value, got Tombstone(TombstoneInfo { CellTombstone .. })
```

A deleted cell must read as **NULL**, not fail the request.

## PRE-EXISTING ON `main` — NOT introduced by #3058

This reproduces on **both** arms, including under `CQLITE_FLIGHT_MERGE_PATH=merge`, which is
exactly pre-#3058 behavior. #3058 (Flight single-SSTable bypass) only made it *visible* while
building the forced-path differential test — it did not create it.

## Repro

1. Using the write engine, write a clustering row with two text columns (`v`, `w`).
2. Write a later-timestamp `CellOperation::Delete` on `w`.
3. Flush to a **single** SSTable.
4. `do_get` a `SELECT *` ticket over it.

Fails identically on the merge arm (`CQLITE_FLIGHT_MERGE_PATH=merge`) and the bypass arm.

## Impact

User-visible on the shipping Flight surface: the entire `do_get` stream errors out. Any table
containing a CQLite-written cell tombstone is unreadable over Flight — not "one NULL column",
the whole request.

## Root-cause hypothesis

The cell tombstone is never dropped to null on the read path, so it reaches Arrow as
`Value::Tombstone`. Two candidate sites:

- `PartitionShadow::cell_shadowed_or_expired`
  (`cqlite-core/src/storage/sstable/reader/parsing/row_decoder/partition_shadow.rs`) does not
  suppress a CQLite-**written** simple cell tombstone (Cassandra-written ones ARE handled — see
  the covered fixtures in the #3058 differential), **or**
- the merged-read assembler's "simple cell tombstones are dropped" rule
  (`assemble_read_cells` in `cqlite-core/src/storage/write_engine/merge/read_assembly.rs`,
  related handling in `merge/reconcile.rs`) misses this shape.

The Arrow converter then correctly **fails closed** — that guard is working as designed (#1485);
it is the messenger, not the defect.

## Affected

[...]

PR #3103, opened 2026-07-28
perf(#3058): single-SSTable merge bypass on the Flight do_get data plane — 3.32x, gap 5.73x→1.47x

### LINKED ISSUE #3058: cqlite-flight do_get runs k-way merge/compaction reconciliation unconditionally on a SINGLE SSTable — 6.0x slower than our own bare scan, 3.48x slower than stock Cassandra

pmcfadin:
Closes #3058

## What this does

`cqlite-flight`'s `do_get` built a `KWayMerger` **unconditionally** — even when a table has exactly ONE SSTable and there is nothing to reconcile. Entering the merger forced every row through `CompactionPolicy::on_data_row`, which requests per-cell write metadata, allocating a `HashMap<String, CellWriteMetadata>` and doing a `column.name.clone()` + SipHash insert **per cell** — metadata the Flight read path never reads (every Flight `QueryRow` sets `cell_metadata: None`).

This adds a **fail-closed single-source bypass**: when the post-prune source count is exactly 1, the row source is swapped from `KWayMerger` to the single-generation scan, reusing core's existing `build_row_from_scan_cached` adapter. The merge arm keeps its exact code path.

Approved OpenSpec change: `openspec/changes/flight-single-sstable-bypass/` — **8 requirements, 33 scenarios**, `openspec validate --strict` clean. Design **Option A** (source swap inside `MergeProducer`), chosen over a shortcut inside `KWayMerger` (leaves the thread + `sync_channel` + `MergeEntry` machinery, ceiling unknown) and over routing to `Database::execute_streaming` (discards token pushdown #2412, warm registry #2356, admission #2420, `max_batch_bytes` #2825).

## Measured result (warm, CPUs 2+10 verified siblings, CPU-wide `perf stat -C`, `taskset`-pinned, median of 3)

| shape | arm | rows/s | cycles/row | cell_meta_maps |
|---|---|--:|--:|--:|
| — | bare scan | 310,743 | 22,051 | — |
| full-ring | merge | 63,600 | 94,770 | 4,000,000 |
| full-ring | **bypass** | **211,105** | **27,954** | **0** |
| token-bound | merge | 62,783 | 94,964 | 2,012,000 |
| token-bound | **bypass** | **206,068** | **26,688** | **0** |

**3.32x** full-ring / **3.28x** token-bound; bare-scan gap **5.73x → 1.47x / 1.51x**. Reconciliation work on a single source is **zero** (`reconcile_entries` 4,000,000 → 0).

**Both access shapes measured** — the token-bounded shape (Summary-guided walk) is what the Trino connector actually issues, so it is measured on its own path rather than assumed from the full-ring figure.

### Two disclosures

1. **The ~1.3x acceptance target was NOT reached** (1.47x / 1.51x). Not rounded toward the target. The residual is Arrow encode + IPC framing + loopback gRPC, with every merge-path probe counter at zero → owned by **#3096**. Owner accepted shipping at this ratio.
2. **The merge arm got 18.1% faster as a side effect** (53,873 → 63,600 rows/s; its own metadata maps halved 8M → 4M) because the `structure_only` coverage check stops driving the compaction parser twice per row. Emitted rows are byte-identical. **This supersedes an earlier 3.90x figure** measured before that fix; 3.32x is the honest number.

### Acceptance substitution (owner decision at Seam 1)
The WS0 rig (`/home/ubuntu/ws0/**`) is **absent from the delivery machine**, never committed, and there is no Docker/Cassandra to regenerate it; CQLite's write surface is also uncompressed-only (#1406) vs WS0's LZ4. So acceptance became **ratio closure on a locally generated 4M-row corpus**, both surfaces over identical bytes, same box/cores/session. **Owed, explicitly not reproduced:** the WS0 absolute (>= ~280,000 rows/s), the AC #5 digest `0x4903ffa446163c4b`, and the stock-Cassandra leg.

## Correctness

The bypass predicate is **conjunctive, total and fail-closed**, from authoritative state only (#28) — never inferred from bytes, file sizes or `Statistics` estimates:
post-prune count `== 1` · `dropped_columns.is_empty()` · no STATIC column · not aggregating · not the point-read route · `CQLITE_FLIGHT_MERGE_PATH` not forcing merge.

`CQLITE_FLIGHT_MERGE_PATH=bypass|merge` (unset = auto) is a **permanent documented seam** mirroring `CQLITE_READ_PATH` — it is how correctness is proven, and the field's kill switch.

- **Path-taken assertions, both directions**, on a real observed `read_path_probe` marker — never timing: 1 source → merger not constructed; ≥2 → constructed.
- **Forced-path differential** over the SAME bytes at a PINNED `now` — 14 cases (range tombstone, row delete, partition delete, TTL at two pinned `now`s, predicate, projection, token range, byte cap, UDT). `assert_arms_agree` fails if the merge run didn't merge **or** the bypass run didn't bypass, so it cannot pass having run one arm twice.
- **NEW ≥2-overlapping-SSTable pin** (AC #4) — closes a real coverage hole: every committed `test_compaction_tombstone_ttl` fixture holds exactly ONE `Data.db`, so both semantics oracles previously exercised only a single-SSTable table.
- Merge arm **byte-identical**: `caller_schema = None` restored at its entry; a diff-wide re-check confirmed every merge-reachable path passes `None`/`None` or is observability-only.
- `now_secs` threaded to the fast arm's shadow clock (never ambient wall-clock); `read_shadowing = true` at all construction sites.
- #2988 not regressed — the multi-generation core pins stay green.

## Review

`rust-reviewer` → 2 blockers, both fixed: a fallback that cancelled the **caller's** `ScanCancel` (poisoning the merge arm it fell back to), and the merge arm not being byte-identical. roborev (verified **non-vacuous** — 107,332 input tokens, findings citing branch-only files, per the #2964 worktree trap) → **0 real blockers**; its 4 High/Medium findings were each disproven with evidence. Two further items were promoted to blocker and fixed: a silent-duplicate-rows path, and a differential case that could pass vacuously. A `ScanAdmission` item was checked rather than deferred — the core semaphore caps at `available_parallelism()` (16) while `--max-concurrent-scans` defaults to 64, so it **would** have throttled; fixed with `Exempt`.

## Bugs discovered (all PRE-EXISTING, none introduced here)

Putting two implementations of the same read path side by side exposed three latent defects:

- **#3095** (P1) — static columns: the merge arm emits a **phantom `ck=null` row** per static-bearing partition AND leaves statics null on real rows. Verified against the pinned `cassandra-5.0.8` `SelectStatement.processPartition` (L1089-1152); confirmed **not** a #1074 write-side artifact (observed on the Cassandra-written `test_writeparity.static_clustering_shape`). Static-bearing schemas are excluded from the bypass fail-closed here — a **deferral owned by #3095**, not a design position.
- **#3094** (P1) — a CQLite-written cell tombstone reaches the Arrow encoder as `Value::Tombstone` and errors the whole stream, on **both** arms.
- **#2339** (commented, not double-filed) — `set<frozen<UDT>>` fails closed on the merge arm while the scan arm decodes it; now arm-dependent.

Follow-ups: **#3096** (Arrow encode — closes 1.47x → ~1.3x), **#3097** (merge arm ignores the ticket's authoritative schema), **#3098** (batched nits). **#3028 and #3047 should be re-priced** against the post-#3058 profile — much of their measured allocator churn lived inside a merge that no longer executes.

## Residuals

- `CQLITE_ALLOW_FILE_GROWTH=1` for 7 already-over-threshold core files (+1-28 lines each: `now_secs`/`caller_schema` parameter plumbing, two probe calls, two re-exports — no new logic block). Splitting them is a separate campsite job → **epic #1116**.
- Rig preserved at `/home/ubuntu/ws0-local/` (`method.md` §4b, `run-measure-final.sh`, `digest-final.sh`, `results-final/`) so the measurement is re-runnable — the original WS0 rig was lost because it was never committed.


pmcfadin:
## Certification of record (flow-closer)

Two distinct certifications. **The delta is NOT the gate of record.**

### 1. Gate of record — full `AGENT-GATE SUMMARY` @ `e1c84f9`
```
run-id: /tmp/agent-gate.3QMiuG
commit: e1c84f9  branch: issue-3058-flight-single-sstable-bypass  dirty: no
tree-start: e1c84f944661 dirty: no digest: 63696e6de4c1
tree-end:   e1c84f944661 dirty: no digest: 63696e6de4c1
tree-integrity: PASS
RESULT: PASS
```
30 components PASS, **0 SKIP, 0 FAIL**; `datasets:` fixture-backed under `/data/datasets`;
`accelerators: sccache=on nextest=on lanes=on sccache-health=ok`. Summary file
`/tmp/gate-3058-full6.txt`. Process exit 0; `run-id` verified against the launch log (#2874);
no `.integrity-fail.*` sibling.

### 2. Delta re-cert — `AGENT-GATE DELTA SUMMARY` @ `7254b77` (this PR head)
```
run-id: /tmp/agent-gate.W0tro5
MODE: delta (TEST/DOCS-ONLY RE-CERTIFICATION — NOT the gate of record)
commit: 7254b77  dirty: no
delta-anchor: e1c84f944661e7cae71d3b88f93ec0902ba5dce0 (full-gate PASS commit)
delta-anchor-run-id: /tmp/agent-gate.3QMiuG
delta-files (2 allowed / 0 offending)
tree-integrity: PASS
RESULT: PASS
```
Envelope independently verified: `e1c84f9..7254b77` = 2 files, both under `cqlite-flight/tests/`
(+83/-18) — no `src/`, `Cargo.*`, `scripts/`, workflow or `test-data/` byte, so `--delta` (#1892)
is eligible and did not pass vacuously. `delta-anchor-run-id` matches the gate of record above.

**Superseded / void — do not read as current:** full gates at `32a6ad0`(preflight fail-close),
`5f0a512`, `8ce5bb5`, `2942afd`, and deltas at `37caf22`, `e87ec6d`.

### Intent audit (C)
`spec-auditor`, anchored to `openspec/changes/flight-single-sstable-bypass/specs/**`: **PASS** across
three passes — every requirement `satisfied`, **R2** a justified `partial` (the `structure_only`
coverage-check exception, named observables zero and tested), **R7** evidence restored, and the
8-arm `assemble_complex` divergence enumeration independently verified **complete**. Spec unchanged
since that PASS (`e1c84f9..7254b77` touches 0 files under `openspec/`), so it carries to this head.

### roborev — NOT clean; triaged, not silent
Four passes (jobs 9/10/11/12), each `--repo` + explicit `--since c1f436f6` per the #2964 guard,
non-vacuity confirmed by commit count (16 → 18 → 20 → 21) and branch-only findings — never the
vacuous empty-diff signature. Final pass returned **1 Medium + 3 Lows (no High, no regression)**.
The Medium plus two arm-parity Lows are **deferred, not fixed**, to **#3112** (one root cause: the
predicate trusting the caller's ticket DDL), which carries the ordering hazard that its fail-closed
guard must land **with or after #3109**. The fourth finding was fixed here (it was R7's sole
`max_batch_bytes` evidence). Nits batched to **#3098**.

pmcfadin:
## 🔎 Provenance of the merge decision (recording explicitly)

The final `flow-closer` packet correctly flagged that it **could not verify** whether deferring roborev job 12's Medium finding was an owner decision or an agent one — every write from this fleet uses the same GitHub account, so authorship cannot distinguish them. That was the right thing to surface rather than assume, so recording the answer here where it is durable:

**The deferral was an explicit owner decision.** After roborev job 12 returned **1 Medium + 3 Lows (no High, no regression)**, the agreed stopping rule named *any* Medium as a stop condition, so the finding was escalated to the owner rather than fixed. The owner was presented four options — merge now and file the class; fix the frozen-unwrap Low too; fix everything including the Medium; or stop and re-scope the whole change — and chose **"Merge now, file the DDL-trust class."**

Consequently:
- The Medium (unenforced fail-closed precondition on the public `open_query_row_stream` full-ring arm) and two arm-parity Lows (missing `unwrap_frozen` in the multicell guard; empty-key token decoration) are deferred to **#3112**, filed with the finding verbatim and with the ordering hazard recorded — its fail-closed guard must land **with or after #3109**, because this delivery's round-5 version of that same shape broke pre-existing BTI batched scans and had to be reverted.
- The fourth finding **was** fixed here (`fb78654`), because it was test-only and it was spec R7's *sole* `max_batch_bytes` evidence: the assertion compared batch-size **sums**, which the row-count equality above it already implied, so a fast arm ignoring the byte cap entirely would have passed. It now compares **boundaries**, and the cap moved 512 → 16 bytes because 512 let the whole 4-row result fit in one batch.
- **`roborev: clean` was deliberately NOT claimed.** The closer declined to write it even though its packet template offered the field. The honest state is *not clean — 1 Medium deferred, tracked in #3112.*

### Why the deferral is defensible on the shipping path
- The Medium is unreachable in production: `cqlite-flight`'s `ReaderUnsupported` precondition rejects BTI readers before `open_query_row_stream` is reached. It is a defence-in-depth and doc-accuracy defect, dangerous to a future or third-party caller.
- Both deferred Lows require a **stale or hand-built ticket DDL**. The Trino connector derives its DDL from live schema, and Cassandra rejects empty partition keys, so neither is reachable with a correctly-generated ticket over Cassandra-written data.
- Adding a fail-closed guard for an unreachable-today seam is *structurally identical* to this delivery's round-5 mistake, which cost a HIGH regression and two gate cycles. Deferring it to land alongside #3109 is the safer sequencing, not the lazier one.

If that call should be revisited, the squash-merge is cheap to revert and **#3112** is the handle.

Issue #3120, opened 2026-07-30
k-way merge adapter treats a dead producer thread as end-of-input (silent truncation, incl. compaction output)


pmcfadin:
Follow-up split out of #3106 (query row stream fail-closed on a dead producer), item 3.

## Defect (same class as #3106, different row source)

`SSTableRowIteratorAdapter::next` (`cqlite-core/src/storage/write_engine/merge/mod.rs`, the `RecvTimeoutError::Disconnected` arm) returns `None` — "producer finished normally" — whenever the per-input channel closes. Both producer-thread shapes (`producer_thread`, path-based; `producer_thread_from_reader`, shared-reader, `merge/from_readers.rs`) drop their `SyncSender` on exit with **no explicit terminator**, so "the walk finished" and "the producer thread UNWOUND" are indistinguishable exactly as they were in the query row stream.

Consequence: a panic anywhere in a producer's walk/decode makes that run look EXHAUSTED. The k-way merge then completes successfully having merged only the rows that made it into the channel:

* on the READ path (Flight warm/cold merge arm) → a silently short result set, no error, no log;
* on the WRITE path (compaction/flush merge) → the rewritten SSTable is missing rows, i.e. **silent data loss on compaction**, and `Drop` explicitly ignores the producer's failed join ("the panic was already surfaced through the error channel" — which it was not).

## Why it was NOT folded into #3106

#3106 fixed the query row stream (explicit `Done` sentinel + `catch_unwind` forwarding the panic as an error) and deliberately left this one alone: it is a materially larger, higher-risk change in a different subsystem.

1. **Wider churn / different channel shape.** The item is `Result<MergeEntry, MergeProducerError>` across 17 sites in 3 files, and the egress-depth gauge accounting (`channel_depth::sent`/`received`, the post-join reconcile residual) is keyed on `msg.is_ok()`, so a third message kind has to be threaded through that balance too.
2. **File-size ratchet.** `merge/mod.rs` is ~7.5k lines (far over the ~800 campsite target), so ANY growth there FAILs the gate's `file-size` component. Doing this properly first requires splitting the adapter's `impl SSTableRowIterator` / producer thread out of the compaction merge core (epic #1116) — a refactor of its own.
3. **Different risk class + evidence bar.** This changes COMPACTION semantics (a panicking producer would now fail the compaction instead of writing a short SSTable), which is the byte-parity write path: it needs the full write/compaction suite and a compaction-parity review, not the read-path lite loop.
[...]

Issue #2964, opened 2026-07-26
roborev in a worktree silently reviews the BASE commit — vacuous 'No issues found' is indistinguishable from a clean pass


pmcfadin:
## Problem

`roborev review --branch --base origin/main` run **from inside a git worktree** enqueues the review for
`origin/main` (the BASE) instead of the branch HEAD, so the reviewer receives an **empty diff** and
replies:

```
Enqueued job 4651 for 39900e4db454724c2 (agent: codex)
Review (by codex) — No issues found.
Summary: The provided combined diff contains no code changes to review.
```

`39900e4db` is `origin/main`; the branch HEAD was `4e7ab591e`. `roborev log 4651` confirms the reviewer
genuinely got no content (17,333 input tokens → 21-token reply).

## Root cause

`roborev repo list` tracks only the root checkout `/Users/pmcfadin/projects/cqlite`. Worktrees under
`~/projects/cqlite-wt/issue-<N>` are **not registered**, and `roborev repo` has no `add` subcommand
(repos self-register on first use). `--branch` therefore resolves against the *root* checkout — which
normally sits on `main` — yielding a base-vs-base no-op regardless of the cwd the command runs in.

## Why this is P1

**A vacuous pass is textually identical to a genuine clean pass.** Every worktree-based roborev run in
the flow-* pipeline (i.e. all of them — 1:1:1:1 puts each issue in a worktree) can silently record
"roborev clean" for code nobody reviewed, then proceed to the one gate of record and merge.

Observed on #2950: two vacuous runs "passed". Re-run correctly against the real SHA, the SAME diff
produced **two real blockers** (a false claim that every non-frozen collection cell value carries an
unsigned-VInt length — sets store the element in the cell path with an EMPTY value per
`SetType.valueComparator() → EmptyType.instance`, and fixed-width values are written raw; plus a false
"signed VInt is used only for duration" absolute — `writer/index_writer.rs:667` also uses
`encode_signed` for promoted-index width deltas). Accepting the empty-diff verdict would have shipped
both.

## Acceptance criteria

1. A roborev invocation whose resolved diff is EMPTY while the requested branch/SHA range is NON-empty
   must **fail loudly** (non-zero exit + explicit message), never return "No issues found."
2. The flow-* skills' roborev step invokes with an explicit SHA + explicit `--repo <worktree path>`
   (`roborev review <sha> --repo <abs-worktree> --agent codex --model gpt-5.6-sol --wait`) rather than
[...]

Issue #1592, opened 2026-07-02
F2 — batch the public streaming channel (stop one async wake per row)


pmcfadin:
Child **F2** of Epic F (#1518) — 2026-07 read-path audit (block 1). Source of truth: `docs/reports/read-path-performance-audit-2026-07-01.md` §Epic F. Anchors as of `main` @ `5c080d2a`. Capstone wave: **Wave 3**.

## Problem
The internal windowed scan batches rows deliberately (its own docs measured per-row seam costs at 31–42%), but the PUBLIC streaming channel forwarder **re-flattens to one row per channel send** (`mod.rs:1982`, `:1887`, `:2013`) — one async wake per row survives the pipeline that was specifically designed to avoid that.

## Fix (step by step)
1. Change the public streaming channel element type from `Row` to `Vec<Row>` batches (**API-additive**: add a batched stream surface, e.g. `scan_stream_batched()`/`next_batch()`, and reimplement the existing per-row surface as a thin flattening adapter over it so no public contract breaks).
2. Forward the internal batches straight through — do not flatten and re-batch.
3. Batch size: reuse the existing internal batch constant (`BATCH_EMIT_ROWS` pattern); do not add a config knob unless one already exists.
4. Bindings note: Python/Node streaming iterators sit on this channel — they may keep per-row semantics via the adapter; wiring them to the batched surface is bindings work (V/W epics), not this issue.

## TDD tests (write FIRST, must fail on current main)
- Wake/send counter: streaming a 10k-row result performs O(rows/batch) channel sends, not O(rows) (fails today).
- Streaming scan throughput bench before/after; add to gate.
- Stream-content parity: batched-then-flattened output identical to current per-row output (order + values), including mid-stream drop/cancel behavior (drop-safe cancellation preserved).

## Guardrails
- Public API additive only; existing signatures unchanged.
- Backpressure semantics preserved (bounded channel, blocking_send discipline).

## Definition of done
- [ ] TDD tests red-then-green; `scripts/agent-gate.sh` PASS (paste SUMMARY verbatim)
- [ ] `RUSTFLAGS="-D warnings"` clean; no `unwrap()`/`expect()` in library code

pmcfadin:
🧭 **MANAGER** 
[...]

PR #3123, opened 2026-07-30
fix(flight): fail closed when the query-row producer thread dies instead of silently truncating the result set (#3106)

### LINKED ISSUE #3106: a producer-thread panic in the query-row stream completes the request successfully with a silently truncated result set
## Summary

`QueryRowStream::next_batch` (`cqlite-core/src/storage/sstable/reader/data_access/summary_scan/query_rows.rs`)
is `self.rx.recv().ok()`, which collapses a **disconnected** channel into `None`. The Flight
single-source row source maps `None` to `SourceStep::Complete`, i.e. "the scan finished".

If the producer thread UNWINDS — a panic anywhere in the walk/decode rather than an `Err` return —
the `SyncSender` drops without sending a terminal message, so the consumer sees a disconnect and
reports a clean end of stream. The `do_get` then completes **successfully with a silently truncated
result set**: no error, no log, fewer rows than the table holds.

Found in the #3058 review. It matches the existing k-way merge adapter's stance
(`write_engine/merge/from_readers.rs`, whose channel close is likewise treated as end-of-input), so
it is pre-existing IN KIND rather than introduced by #3058 — but silent truncation is a correctness
failure mode that deserves its own issue rather than a nit bucket.

## Repro sketch

Inject a panic into the walk (e.g. a poisoned fixture that makes the decoder panic rather than
return `Err`), run a Flight `do_get` over a single-SSTable table, and observe: the RPC returns `Ok`
with fewer rows than the fixture contains, and nothing surfaces the panic.

## Root cause

The channel protocol has no explicit terminator. "Sender dropped" is ambiguous between
"walk finished" and "producer died".

## Fix shape (acceptance)

1. Make completion EXPLICIT: the producer sends a `Done` sentinel (or the thread body is wrapped in
   `catch_unwind` and forwards `Err`), and the consumer treats a disconnect WITHOUT a sentinel as a
   corruption error.
2. A test that panics the producer thread mid-stream and asserts the request FAILS (an error to the
   client) rather than returning short rows.
3. Consider applying the same treatment to the merge adapter, so both row sources fail closed on a
   dead producer.

Routing: oracle-driven bug fix — issue + pinned test, no OpenSpec change.

pmcfadin:
Closes #3106

## Problem

`QueryRowStream::next_batch` was `self.rx.recv().ok()`, which collapses a **disconnected** channel into `None`. The Flight single-source row source mapped `None` → `SourceStep::Complete` — "the scan finished".

So if the producer thread **unwound** (a panic anywhere in the walk/decode rather than an `Err` return), the `SyncSender` dropped with no terminal message, the consumer read the disconnect as a clean end of stream, and `do_get` completed **successfully with a silently truncated result set**: no error, no log, fewer rows than the table holds.

Root cause: the channel protocol had no explicit terminator. "Sender dropped" was ambiguous between *walk finished* and *producer died*.

## Fix

Completion is now **explicit**, belt-and-braces:

- `QueryRowMsg::{Item, Done}` — the producer sends a `Done` sentinel; the consumer treats a disconnect **without** a sentinel as an error, never clean EOS.
- The producer thread body is wrapped in `catch_unwind`, which forwards the panic message as a terminal `Err` so the client gets an *informative* error rather than a generic "producer died".

Both are carried because `[profile.release]` is `panic = "abort"`, where `catch_unwind` never fires — so the **sentinel** is what makes completion an observed fact in every profile, and the unwind catch is the diagnosability layer on top.

## Wiring evidence

`SourceStep::Complete` is now reachable **only** after an observed `Done`. Call chain, exercised end-to-end by the Flight test (not asserted by inspection):

`FlightService::do_get` → `MergeProducer::produce_streaming_from_readers` → `bypass::ScanRowSource::next_step` → `QueryRowStream::next_batch` → `producer_stream::drive_row_source` (maps `Err` → `ProducerError::Merge`) → `streaming::spawn_streaming` → `tonic::Status`.

## Tests — both watched RED before the fix

Verified by temporarily restoring `recv().ok()` + no-forward semantics and confirming each test fails for the right reason, so neither can pass vacuously:

- `query_rows::panic_tests::a_producer_panic_mid_stream_fails_the_stream_instead_of_truncating_it_silently` — real producer thread, real walk, 384-row control drain.
- `cqlite-flight/tests/issue_3106_producer_panic_fails_closed.rs::a_dead_query_row_producer_fails_the_do_get_instead_of_truncating_it_silently` — asserts at the **RPC surface**.
- Plus 4 protocol unit tests in `query_rows_tests.rs`.

Panic injection is a test-only seam (`storage/producer_fault.rs`, non-default feature `producer-fault-injection`; ZST + empty fn in production — no production knob can flip it). Panic-hook noise is filtered **by message** and the hook is restored before every assertion, so a real unexpected panic is never swallowed.

## Scope: acceptance item 3 deferred, not dropped

Item 3 (same treatment for the k-way merge adapter, `write_engine/merge/from_readers.rs`) is **deferred to #3120** with a KNOWN GAP note in that file. It is disproportionate here: 17 call sites over a differently-shaped channel whose `channel_depth` sent/received balance keys on `msg.is_ok()`; `merge/mod.rs` is ~3.5k lines so any growth trips the file-size ratchet (needs an #1116 split first); and it changes **compaction** semantics, requiring the byte-parity write suite plus a second fault seam. The Flight read path under this PR does **not** route through the still-broken adapter for the case #3106 covers.

## Review status

- **roborev: NITS ONLY (2), no blockers.** Batched to a single linked follow-up per severity doctrine — nits never trigger a re-verify round. Neither touches correctness, wiring evidence, or no-heuristics.
- `rust-reviewer` pass reported separately in a comment.
- Lite gate PASS (`tree-integrity: PASS`). The **full gate of record** runs once, pre-merge, in `flow-closer`.

Note for the reviewer-tooling folks: this run reproduced the **#2964** display symptom — roborev printed the *base* SHA (`f603c7f`) in `Enqueued job`/`list` while the review content was unambiguously the branch diff. Non-vacuity was confirmed by the review citing symbols that exist only on branch HEAD.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A dead producer no longer completes a query SUCCESSFULLY with a silently
+  truncated result set (#3106).** Both channel boundaries behind the
+  single-generation read path (used by the Flight `do_get` row route) treated
+  "sender dropped" as "the scan finished", so a producer thread/task that UNWOUND —
+  a panic anywhere in the walk or decode, rather than an `Err` return — ended the
+  request with **fewer rows than the table holds, no error and no log**.
+  - The query row stream's producer→consumer protocol now has an EXPLICIT terminator
+    (`Done`) and a distinct `Failed` message; a disconnect WITHOUT a terminator is an
+    `Error::Internal` naming the truncation, never a clean end of stream. The
+    producer thread additionally runs under `catch_unwind`, so a panic reaches the
+    client as an informative error carrying the panic message.
+  - The batched streaming scan's driver task is now JOINED when its channel closes
+    (`BatchedScanStream`), so a task that panicked/was cancelled surfaces as an
+    `Error::Internal` instead of an empty-but-successful end of stream. This closes
+    the same hole for the streaming `SELECT` and aggregate-fold paths, which
+    consume the same surface.
+  - Both errors are `is_recoverable() == false`: the failure is deterministic, so a
+    retry would reproduce it.
+
 - **BTI `Rows.db` row-index base + leading `NEXT_COMPONENT` byte (#3002, #3040).** Two
   defects that cancelled each other on the BTI clustering read path are corrected:
   (a) the per-partition `TrieIndexEntry`'s SIGNED root delta is now measured from
@@ -54,6 +73,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `Result<ValidatedRowsTrieRoot, RowsTrieRootRejection>` instead of a bare `usize`, so
   an unvalidated row-index root cannot be traversed by accident; use
   `trie_root_offset()` / `require_trie_root()` (#3002).
+- **Breaking (API):** the batched streaming scan now returns
+  `sstable::reader::BatchedScanStream` instead of
+  `tokio::sync::mpsc::Receiver<Result<Vec<(RowKey, ScanRow)>>>` — affecting
+  `StorageEngine::scan_stream_batched`, `SSTableManager::scan_stream_batched` (both
+  feature variants) and `SSTableReader::scan_stream_batched`/`_admitted`. The new type
+  owns the scan task's `JoinHandle` so a dead task cannot masquerade as end-of-stream
+  (#3106). Its `recv()` keeps the same `async fn(&mut self) -> Option<Result<..>>`
+  shape, so the usual `while let Some(batch) = stream.recv().await` consumer compiles
+  unchanged; code that named the `Receiver` type, stored it in a struct field, or
+  called `Receiver`-specific methods (`try_recv`, `blocking_recv`, `close`) must be
+  updated.
 
 ## [v0.16.1] - 2026-07-23
 
--- a/cqlite-core/Cargo.toml
+++ b/cqlite-core/Cargo.toml
@@ -461,6 +461,18 @@ bench-internals = []
 # links it.
 arrow-shape-corpus = ["arrow"]
 
+# TEST-ONLY producer-fault injection (issue #3106) — gates the arming API of
+# `storage::producer_fault`, the deterministic seam a test uses to kill the query
+# row stream's producer thread mid-walk so the consumer's fail-closed handling of
+# a dead producer (no terminal `Done` sentinel) can be PROVEN instead of asserted
+# by inspection. Adds NO dependencies and NO production behaviour: without this
+# feature the arming symbols do not exist, the armed-state statics do not exist,
+# `ProducerFault` is a zero-sized no-op struct and every default build / gate /
+# CI run is byte-identical. NOT in defaults; `cqlite-flight` enables it from its
+# `[dev-dependencies]` (the `arrow-shape-corpus` convention), so no shipped
+# binary links it.
+producer-fault-injection = []
+
 # Legacy heuristics support (backward compatibility for pre-5.0 formats)
 # NOT enabled in default CI configuration - opt-in only
 legacy-heuristics = []
--- a/cqlite-core/src/storage/mod.rs
+++ b/cqlite-core/src/storage/mod.rs
@@ -2,11 +2,19 @@
 
 /// Shared, bytes-bounded, sharded decompressed-chunk cache (issue #1567).
 pub mod cache;
-/// TEST-ONLY producer-fault injection for the query row stream (issue #3106):
-/// the deterministic seam a test uses to kill the producer thread mid-stream and
-/// prove the consumer fails closed instead of reporting a clean, silently
-/// truncated end of stream. Inert (and, for the arming API, non-existent) in a
-/// production build.
+/// TEST-ONLY fault injection for the query row stream's producer boundaries
+/// (issue #3106): the deterministic seam a test uses to kill a producer
+/// thread/task mid-stream and prove the consumer fails closed instead of
+/// reporting a clean, silently truncated end of stream.
+///
+/// `pub(crate)` in a production build — the module then has NO public items at
+/// all (every arming symbol is cfg'd out), so publishing it would advertise an
+/// empty module. It becomes `pub` only where the arming surface exists, i.e. for
+/// in-crate tests and for the `producer-fault-injection` feature that
+/// `cqlite-flight` enables from its `[dev-dependencies]`.
+#[cfg(not(any(test, feature = "producer-fault-injection")))]
+pub(crate) mod producer_fault;
+#[cfg(any(test, feature = "producer-fault-injection"))]
 pub mod producer_fault;
 /// Always-on read-path ARM probes (issue #3058): explicit markers for
 /// "the k-way merge ran" vs "the single-generation query scan ran".
@@ -537,7 +545,7 @@ impl StorageEngine {
         end_key: Option<&RowKey>,
         schema: Option<&crate::schema::TableSchema>,
         buffer_size: usize,
-    ) -> Result<tokio::sync::mpsc::Receiver<Result<Vec<(RowKey, ScanRow)>>>> {
+    ) -> Result<sstable::reader::BatchedScanStream> {
         record_table_scan_call();
         self.sstables
             .scan_stream_batched(table_id, start_key, end_key, schema, buffer_size)
--- a/cqlite-core/src/storage/mod.rs
+++ b/cqlite-core/src/storage/mod.rs
@@ -2,6 +2,12 @@
 
 /// Shared, bytes-bounded, sharded decompressed-chunk cache (issue #1567).
 pub mod cache;
+/// TEST-ONLY producer-fault injection for the query row stream (issue #3106):
+/// the deterministic seam a test uses to kill the producer thread mid-stream and
+/// prove the consumer fails closed instead of reporting a clean, silently
+/// truncated end of stream. Inert (and, for the arming API, non-existent) in a
+/// production build.
+pub mod producer_fault;
 /// Always-on read-path ARM probes (issue #3058): explicit markers for
 /// "the k-way merge ran" vs "the single-generation query scan ran".
 pub mod read_path_probe;
--- a/cqlite-core/src/storage/producer_fault.rs
+++ b/cqlite-core/src/storage/producer_fault.rs
@@ -1,0 +1,220 @@
+//! TEST-ONLY producer-fault injection for the query row stream (issue #3106).
+//!
+//! # Why this exists
+//!
+//! The query row stream ([`crate::storage::sstable::reader::QueryRowStream`]) is
+//! fed by a detached producer thread over a bounded channel. Before issue #3106
+//! the consumer collapsed a channel DISCONNECT into a clean end of stream, so a
+//! producer thread that UNWOUND (a panic anywhere in the walk/decode, rather than
+//! an `Err` return) dropped its `SyncSender` without a terminal message and the
+//! request completed SUCCESSFULLY with a silently truncated result set. The fix
+//! makes completion explicit (a `Done` sentinel) and forwards a caught panic as a
+//! real error; this module is how a test PROVES that, deterministically, without
+//! waiting for a real decode bug to unwind.
+//!
+//! # Why this is not a production knob (no-heuristics safe)
+//!
+//! * Everything that can arm a fault is `#[cfg(any(test, feature =
+//!   "producer-fault-injection"))]`. In a default build the arming API does not
+//!   exist, the armed-state statics do not exist, [`ProducerFault`] is a
+//!   zero-sized struct with no fields, and [`ProducerFault::before_batch_handoff`]
+//!   compiles to an empty function — the production build is byte-identical to
+//!   one without this module's body.
+//! * No environment variable, config field or on-disk byte pattern can arm it:
+//!   the only way in is a Rust call to [`arm_query_row_producer_panic`], which is
+//!   compiled out of production builds. So it cannot influence a decoding
+//!   decision (issue #28) even accidentally.
+//! * `cqlite-flight` enables the feature from its `[dev-dependencies]` (the same
+//!   convention `arrow-shape-corpus` / `test-util` already use), so the shipped
+//!   Flight binary never links it.
+//!
+//! # Arming semantics
+//!
+//! [`arm_query_row_producer_panic`] arms the NEXT query row stream opened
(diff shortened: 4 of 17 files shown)


pmcfadin:
## 🔍 rust-reviewer pass — `BLOCKERS: 1`, 5 nits

**BLOCKER: the fail-closed property does NOT hold on the full-ring arm — the default `do_get` path, i.e. this issue's own repro.**

The outer channel is airtight, but on the `token_bound == None` arm rows come from a **second, still-unterminated channel**, and the query-row thread converts that inner disconnect into a *normal* return:

- `sequential.rs:561-577` — `scan_stream_batched_admitted` creates `mpsc::channel(cap)`, spawns `run_scan_stream_batched` in a bare `tokio::spawn` with no terminator and no `catch_unwind`, and **discards the `JoinHandle`**.
- `sequential.rs:643-667` — in the **non-stitching (uncompressed)** branch, `parse_block_entries_at_now` runs inside that task; a decode panic unwinds it, tokio's default `unhandled_panic = Ignore` swallows it, the sender drops.
- `query_rows.rs:604` ends on `None` → `Ok(())` → `Done` is sent → `SourceStep::Complete` → `do_get` **succeeds truncated**. Byte-for-byte #3106, one layer down.

`token_bound` is `None` for any `do_get` without a token filter (`producer_warm.rs:112`) — the full-table scan the issue describes. **Neither headline test can see it**: the core test forces `Some(full_ring())`, the Flight test injects on the outer channel. The compressed/stitching sub-path is already covered (`scan_stream_windowed.rs:558-563`, `:717-725` map `JoinError` → `Error::corruption`), which is what makes the uncompressed branch the live hole.

Being fixed now (not deferred — deferring would leave #3106's acceptance unmet): the inner boundary will distinguish *inner scan finished* from *inner task died*, mirroring the windowed path, plus a test on this arm watched RED pre-fix. The universal doc claim at `query_rows.rs:77-80` that rests on it is being corrected in the same round.

### Nits also folded in (all on this fix's own correctness surface)
1. Unify to `Error::Internal` for both the disconnect backstop and the caught panic — `Error::Storage` maps to `is_recoverable() == true`, so "the producer panicked" is currently advertised as **retryable** for a deterministic failure (and `Corruption` misdirects operators toward a bad `Data.db`).
2. Replace `self.terminated |= item.is_err()` with a distinct `QueryRowMsg::Failed` so `Item` can only carry `Ok` — otherwise a future mid-stream `Item(Err(..))` silently reverts to clean EOS, reintroducing #3106 with no test failing.
3. `pub mod producer_fault` → `pub(crate)` + cfg-gated `pub use` (currently ships a public module with zero public items).
4. A unit test's blanket `set_hook` → use the file's own `silence_injected_panics`.
5. `ArmedProducerPanic` holds a `parking_lot::MutexGuard` across `.await`.

### Batched to a follow-up, not fixed here
roborev's client-visible panic-message exposure: it is a **codebase-wide error-redaction policy** question for the owner (every pre-existing `Error::Storage` on this path already propagates verbatim), and client diagnosability is an explicit #3106 acceptance criterion — "fixing" it inline would revert required behavior.

### Verified clean by the reviewer
Outer-channel exhaustion (terminal state sticky; exactly one terminal message on every exit path incl. `Unsupported` and consumer-dropped; `tx` outside the `catch_unwind` closure; spawn failure returns `Err` before a stream exists); `panic = "abort"` reasoning; the fault seam has **no** production reachability (not in any default set, dev-dep only, resolver 2, genuine ZST + empty fn, no env/config path, uncalled from production even if forced on); per-**batch** not per-row enum tag, no added allocation/copy; both tests non-vacuous by construction via a control arm asserting the full row count; file-size ratchet respected; no `unwrap()`/`expect()` in library code; no-heuristics clean; and the #3120 deferral confirmed valid (the query-row path is a distinct arm from `KWayMerger`).

pmcfadin:
## 🔍 Round-2 re-review: `AUDIT: PASS` + `BLOCKERS: 1` (doc-only)

Two independent passes on the round-2 delta — `spec-auditor` against #3106's acceptance criteria and `rust-reviewer` on the diff — **converged on the same single blocker**, which is documentation, not behavior.

### The inner-boundary fix itself: verified airtight
- `BatchedScanStream::recv()` has four exits; the join is on the **only** closed-channel path. No ordering hole (every sender clone lives inside the task's own graph, and `run_scan_stream_windowed` awaits the forwarder before returning), no leak or join-on-drop stall (drop = detach, task ends cooperatively at its next failed send).
- Fail-closed in **both** panic strategies: unwind → `JoinError` → `Error::Internal` (`is_recoverable() == false`); `panic = "abort"` → the process dies loudly. `Ok(())` is reachable only from an actual normal task return, so an unavailable unwind cannot degrade the join into a clean EOS.
- **Drop-in confirmed for all four consumers** (`drive_full_scan_rows`, streaming SELECT, aggregate fold, `rechunk_into_batches`): every one is `while let Some(b) = recv().await { b? }`. No `while let Some(Ok(..))` exists on this type anywhere, so nothing re-collapses `Err` → end-of-stream.
- Acceptance: criterion 1 **satisfied**, criterion 2 **satisfied for the arm the repro sketch names** (no token filter → `token_bound = None` → `drive_full_scan_rows`, both boundaries pinned), criterion 3 **partial-justified** (the issue says "*Consider*"; `SSTableRowIteratorAdapter` grep-verified confined to `write_engine/merge/**` behind `KWayMerger`, so deferring it does **not** leave this path half-fixed).
- Non-vacuity verified structurally: every test drains a control run asserting `error == None` **and** the full row count before arming, then asserts the faulted run errors **and** is short. Fixture sizes make the handoff certain (800 rows vs a 256-row batch).

### The blocker (being fixed now)
This test file's header claimed the fault fires in the block decode "on this uncompressed/non-stitching fixture", while the code arms checkpoint `0` — the cursor-open prelude — and the fixture is a CQLite-written `nb` file, which resolves to `V5_0NewBig` → `requires_chunk_stitching() == true`, i.e. the **stitching** branch. Both halves false, and the file self-contradicted its own inline comment 25 lines later.

Doc-only, but a blocker rather than a nit: as written, the next reader concludes the non-stitching branch — the exact branch round 1 named as the hole — is covered end-to-end, so the missing fixture never gets filed. **This was the third green-but-blind near-miss on this issue** (round 1's two tests were watched-RED yet blind to the default arm; a later draft passed vacuously outright), which is why it is not being waved through.

### Also in the final round
sticky terminal state on `BatchedScanStream` (public API — an external `recv()` retry must not get a clean EOS from a dead task); resolving the never-armed `parse_batched_block` checkpoint (cover it or delete it); a misplaced `mod` line that detached the #1576 comment; an honest re-confirmation that `file-size` passed with **no** `CQLITE_ALLOW_FILE_GROWTH` override; a release note for the `scan_stream_batched*` signature break (now returns `BatchedScanStream`, not a raw `Receiver`); and softening one over-strong doc claim.

### Follow-ups filed (nothing dropped)
- **#3124** — the remaining discarded-`JoinHandle` boundaries on the **multi-generation** path (`sstable/mod.rs:2194`/`2215`, `rechunk_into_batches:2361`, and the `scan_stream_windowed.rs:569` forwarder hop that discards its `JoinError`). Same defect class, off this repro, still able to truncate a query-engine full scan silently.
- **#3125** — batched nits: the client-visible-panic-text **policy** question for the owner, and the absent uncompressed/non-stitching fixture (which also gates #1592).
- **#3120** — the k-way merge adapter (also affects compaction output).

#3106 + #3120 + #3124 together are the complete audit of "channel close treated as end-of-input" in this codebase; worth landing as a set so the property is uniform rather than arm-by-arm.

pmcfadin:
## 🔴 Final roborev pass: 2 blockers — merge HELD, gate invalidated

The full gate of record **PASSED** at `48d73a6` (all 32 components, 4964/4964 tests, 144 `Data.db`, no SKIP-mask, `tree-integrity: PASS`). But the final roborev confirmation pass found two genuine correctness blockers, both verified against the code rather than relayed. **Not merging.** The src fix invalidates that gate; a full re-run is required.

### Blocker 1 — `BatchedScanStream::recv` cancellation hole reintroduces silent truncation
```rust
let task = self.task.take()?;   // handle moved OUT of self
match task.await {              // ← cancelled HERE → handle dropped → task DETACHED
```
Cancel `recv()` while it awaits the join (the canonical `tokio::select!`-with-timeout pattern — and this type is `pub`, returned from `storage/mod.rs:548`) and the handle is dropped with **`task_died` never set**. A later `recv()` finds `task_died == false`, gets `None` from the closed channel, and `self.task.take()?` **short-circuits on `None`** — reporting a **clean end-of-stream for a task that may have panicked**.

That is the #3106 defect reintroduced through the door built to close it. The in-code comment ("the `task_died` flag, not the absent handle, is what a later call reads") asserts an invariant that cancellation breaks: on that path *neither* is set.

Fix is belt-and-braces: (a) poll the `JoinHandle` **in place** (it is `Unpin`) so a cancellation leaves it owned by `self` and the next `recv()` can still join it; (b) make the ambiguous state — handle absent, `task_died == false`, channel closed — return an **error**, never `None`. `None` must be reachable only from a *proven* clean completion. The thesis of this whole issue is that ambiguity between "finished" and "died" resolves to an error; a `?` that turns "I don't know" into "clean EOS" is that same bug in miniature. The file is being swept for any other such short-circuit.

### Blocker 2 — the process-global fault arm is unserialized, so the panic tests can pass for the wrong reason
`producer_fault.rs` documents the arm as "PROCESS-GLOBAL, caller serializes — one file = one test". That assumes **separate test binaries**. `query_rows_panic_tests.rs` is a `src/` module, so it compiles into the **same `cqlite-core` lib test binary** and libtest runs it in parallel; `serial` appears **0 times** in it. A concurrent test can consume the arm, so the panic-injection test can pass **for the wrong reason** — undermining the wiring evidence for this fix.

Being made **structurally** exclusive rather than by convention (a doc comment saying "caller serializes" is precisely what failed here), without reintroducing the `MutexGuard`-across-`.await` shape flagged earlier. The cross-binary case (`cqlite-flight/tests/*`) is fine; the issue is in-`src` module tests sharing the lib binary.

### Why both are blockers rather than nits
Per `docs/development/roborev-severity.md`: #1 is wrong control flow that fails a stated acceptance criterion of this very issue; #2 is a wiring-evidence gap. And the standing count for this PR is now **four** green-but-blind test situations — two watched-RED-yet-blind tests, one outright vacuous draft, and now a global arm that can be consumed by a stranger. On this issue in particular, "the test is green" has repeatedly not meant what it appears to mean, so the bar stays high.

Finding 3 (panic-hook restore ordering, Low) is a nit → batched to #3125, not fixed here.

### Certification bookkeeping
Gate of record is run #3 only (`/tmp/gate-full-3106-b.txt`, `run-id /tmp/agent-gate.g2xG5n`). Runs #1 and #r2 were invalidated by datasets-root problems (#1 preflight fail-closed; #r2 the missing `../schemas` sibling, now filed as #3131) — **do not cite them**. roborev's pass was non-vacuous: it cites `BatchedScanStream::recv`, `query_rows_panic_tests.rs` and `producer_fault.rs` despite printing the base SHA, which is the #2964 display symptom again.

pmcfadin:
## ✅ Both blockers cleared — roborev `NITS ONLY: 3`, rust-reviewer `NITS ONLY: 5`, zero blockers

### Blocker 1 fixed by making the bad state unrepresentable
`Option<JoinHandle>` + `task_died` became `TaskState { Running(handle) | Finished | Died }`, so "handle gone, verdict unknown" cannot be constructed. The handle is polled **in place** (`&mut JoinHandle: Future + Unpin`), so a cancelled `recv` leaves the stream `Running` and the next call re-joins.

Independently verified by the reviewer rather than accepted on the implementer's word — it swept the file itself:
- **Zero** `?`, zero `.ok()`, zero `unwrap_or`, zero `let else` remain in `batched_scan_stream.rs`.
- Exactly **two** routes to `None`: the sticky `Finished` state, and the observed `Ok(())` join. Both require a *proven* clean completion.
- `self.task` is written in exactly two places, both inside the post-join `match`, with **no suspension point between observing the outcome and recording it** — so no `select!`/`timeout` can interleave and discard the verdict.
- Transition table closed: cancellation writes nothing (state preserved, next `recv` re-joins); `Died` is sticky and re-reports forever; runtime shutdown yields `JoinError` → `Died`; a never-spawned task is unrepresentable (the constructor demands a handle).
- Tokio's "JoinHandle polled after completion" panic is unreachable: the only poll site is guarded by `Running`, and that state is left in the same poll the handle resolves in.

### Blocker 2 fixed structurally, and proved
Every arm now registers a **scope** and fires only when the scanning reader's `Data.db` path matches, held in a registry (`Vec` + per-guard id) rather than one slot — so a foreign scan cannot consume an arm and a second armer cannot clobber the first. Proof is an end-to-end test that arms **both** seams *foreign* and asserts `do_get` returns all 800 rows with no error, plus three registry unit tests. Guards deregister on unwind, ids are monotonic and never reused, and the scope closure is lazy + `cfg`-erased so production pays **nothing** (no allocation, no path formatting, no branch on the hot scan path).

### Why the cancellation tests are non-vacuous
The dead-task test is paired with `a_recv_cancelled_mid_join_still_reports_a_clean_finish_as_clean`, closing the symmetric failure — a cancelled join must not manufacture a spurious error either. Determinism comes from an explicit close-signal + parked task with a `biased` `select!`, not timing, and a precondition `panic!` makes a mis-landed cancellation **loud** rather than silent. No wall-clock asserts anywhere in the delta.

### Final round in flight: two one-line hardening fixes
Folded in only because the full gate must re-run at the new HEAD anyway, so the marginal cost is zero (nits normally never trigger a re-verify round):
1. `Drop` on the panic-hook guard gets `if std::thread::panicking() { return; }` — `set_hook` aborts when called from a panicking thread, and fallible calls still run inside silenced blocks. This is the **unreadable SIGABRT that already bit this issue once**; in the flight binary it would destroy every sibling test's result. Fixed at the source rather than by hoisting calls out, since the latter is discipline that erodes.
2. A `debug_assert!` on scope specificity — an empty/short scope would silently match every reader, reverting to exactly the process-global behaviour blocker 2 eliminated. "The property holds because callers are careful" is what failed the first time.

Everything else batched to **#3125** (hook stack-discipline across parallel guards — verified it can never swallow a real panic; the E2E positive re-take assertion; the non-canonicalized `TempDir` scope path symlink trap; client-visible panic text).

Both reviewers independently identified **#3124** as the sole remaining residual (`rechunk_into_batches` reporting a proven-clean finish while its upstream producer's handle is still discarded on the multi-generation arm) — the same scope boundary we drew, reached independently.

Next: full gate of record re-run at the new HEAD, then premerge assert → arm merge-on-green → finalize.

Issue #2988, opened 2026-07-27
read path (core): multi-generation SELECT still drives the buffered KWayMerger::step — O(partition) memory + partition-granular cancellation (core-library residual of #2230/#2423)


pmcfadin:
Found during the #2956 guide audit (merge/lifecycle chapter).

**Route: oracle-driven // mechanical follow-through.** No OpenSpec change — this mirrors an
already-decided fix: #2230 (full-scan `do_get`) and #2423 (point-read + cache-warm) established
`StreamingMerger::step_streaming` as the sanctioned shape for exactly this defect. Go straight to
implement, reusing the existing #1668 merger — do NOT duplicate reconciliation.

## Defect

#2230/#2423 converted the **cqlite-flight** read paths to row-granular
`StreamingMerger::step_streaming`. The **core-library** multi-generation read path was never
converted: it still drives the buffered `KWayMerger::step()`, which collects EVERY reconciled row of
one partition into a `Vec` before returning `MergeStep::Partition { key, rows }`. Peak memory is
therefore O(widest partition) and cancellation is partition-granular, independent of `limit`.

Unconverted call sites, all in `cqlite-core/src/storage/sstable/generation_merge.rs`:

| line | function | reached by |
|---|---|---|
| **268** | `merge_generations_for_read` | full/range scan + point-target scan |
| **358** | `seek_merge_generations_for_read` | multi-candidate point read (`WHERE pk = ?`) |
| **464** | `merge_generations_for_read_with_metadata` | `WRITETIME`/`TTL` projection + `scan_partition` |

Site 268 additionally accumulates the whole reconciled table into an `out: Vec` across ALL partitions
before returning, so the materialization there is O(result set) on top of O(partition).

A fourth, lower-severity instance: `stream_generations_for_read` (the *lazy* channel-fed sibling used
by `scan_stream`) is bounded across partitions but still calls the buffered `merger.step()` at
**line 627**, so it holds one full partition resident — its own probe records exactly that
(`stream_merge_probe::record_resident(live.len())`, line 655). Converting it to `step_streaming`
would tighten the resident window from O(partition) to O(cluster group).

## Wiring evidence — this is on the common public `SELECT` path (not internal-only)

```
Database::execute(sql)                       cqlite-core/src/lib.rs:455
  -> QueryEngine::execute / execute_select_query   cqlite-core/src/query/engine.rs:239,331,401
    -> SelectExecutor (full-scan + forced-plan branches)
[...]

PR #3128, opened 2026-07-30
fix(#3097): merge-arm query enumeration honours the caller's authoritative schema

### LINKED ISSUE #3097: stream_all_partitions_for_query ignores the caller's authoritative schema, so nb-header-less clustering columns decode as NULL on the merge arm

pmcfadin:
Closes #3097.

## Problem
The Flight warm k-way merge arm — `stream_all_partitions_for_query` → `stream_partitions_summary_guided` — resolved its decode schema from the reader's own four-tier lookup (`get_table_schema(None)`) and hard-coded `caller_schema = None`, ignoring the caller's authoritative schema. An `nb` SSTable header carries no embedded schema, so a table whose ticket DDL declares clustering keys decoded those columns as **NULL** on the merge arm (the single-source fast path, fixed under #3058, decoded them correctly over the same bytes).

## Fix
Thread the caller's authoritative schema through the merge-arm enumeration:
- `summary_scan/mod.rs`: `stream_all_partitions_for_query`'s non-stitching branch now forwards its received `schema` as `caller_schema` to `stream_partitions_summary_guided` (was `None`).
- `full_index_stream.rs`: `stream_all_partitions_cancellable` gains a `caller_schema: Option<&TableSchema>` param, threaded into `stream_all_partitions_via_full_index` and the sequential fallback.
- `compaction.rs`: forwards `Some(schema)` (net-zero behavior — compaction never routed the query method, so byte-parity is unchanged).
- None-fallback to the reader's four-tier lookup is preserved everywhere.

No-heuristics: authoritative metadata only (caller schema, else reader lookup). No byte-pattern inference.

## Tests (RED before fix, GREEN after)
- `merge_arm_query_stream_honours_caller_schema_clustering` (cqlite-core) — pins that the merge-arm surface surfaces the clustering column under the caller's name with `Some(schema)`, and falls back to the reader-derived placeholder with `None`.
- `forced_path_differential_agrees_on_every_shape` → new `clustered_uncomp` case (cqlite-flight) — drives `do_get` under `CQLITE_FLIGHT_MERGE_PATH=merge` vs `=bypass` at a pinned `now`, asserts arm equality AND non-null `ck`.

## Scope note (acceptance criterion 1)
The defect is only observable for `V5_0Uncompressed`-classified readers (`!requires_chunk_stitching()`). Real fetched Cassandra `nb` fixtures classify as `V5_0NewBig` (chunk-stitching, which already honored the caller schema), so a literal real-Cassandra `nb` fixture cannot reproduce it — the pinning test uses a CQLite-written `V5_0Uncompressed` fixture. The fix is path-agnostic and correct regardless of classification.

## Acceptance
1. ✅ Merge-arm `do_get` returns the clustering column populated, matching arm-equality vs bypass (via the forced-path differential; see scope note re: fixture classification).
2. ✅ Test pins merge-arm-vs-bypass equality under `SELECT *`.
3. ✅ compaction-byte-parity + #2988 multi-generation pins unaffected (compaction schema resolution unchanged).

Oracle-driven bug fix — no OpenSpec change.

### DIFF
--- a/cqlite-core/src/storage/sstable/reader/data_access/compaction.rs
+++ b/cqlite-core/src/storage/sstable/reader/data_access/compaction.rs
@@ -595,7 +595,7 @@ impl SSTableReader {
         // and would hit the no-schema `parse_block_entries` error; the stitch+parse
         // drain below decodes BTI correctly (as `bti_scan_with_metadata` does).
         if !self.requires_chunk_stitching() && self.bti_partitions_db.is_none() {
-            self.stream_all_partitions_cancellable(scan_cancel, schema, |(key, value)| {
+            self.stream_all_partitions_cancellable(scan_cancel, |(key, value)| {
                 let row =
                     super::super::compaction_row::CompactionRow::from_legacy_value(key, value, 0);
                 emit(row)
--- a/cqlite-core/src/storage/sstable/reader/data_access/compaction.rs
+++ b/cqlite-core/src/storage/sstable/reader/data_access/compaction.rs
@@ -595,7 +595,7 @@ impl SSTableReader {
         // and would hit the no-schema `parse_block_entries` error; the stitch+parse
         // drain below decodes BTI correctly (as `bti_scan_with_metadata` does).
         if !self.requires_chunk_stitching() && self.bti_partitions_db.is_none() {
-            self.stream_all_partitions_cancellable(scan_cancel, |(key, value)| {
+            self.stream_all_partitions_cancellable(scan_cancel, schema, |(key, value)| {
                 let row =
                     super::super::compaction_row::CompactionRow::from_legacy_value(key, value, 0);
                 emit(row)
--- a/cqlite-core/src/storage/sstable/reader/data_access/compaction.rs
+++ b/cqlite-core/src/storage/sstable/reader/data_access/compaction.rs
@@ -595,7 +595,7 @@ impl SSTableReader {
         // and would hit the no-schema `parse_block_entries` error; the stitch+parse
         // drain below decodes BTI correctly (as `bti_scan_with_metadata` does).
         if !self.requires_chunk_stitching() && self.bti_partitions_db.is_none() {
-            self.stream_all_partitions_cancellable(scan_cancel, |(key, value)| {
+            self.stream_all_partitions_cancellable(scan_cancel, None, |(key, value)| {
                 let row =
                     super::super::compaction_row::CompactionRow::from_legacy_value(key, value, 0);
                 emit(row)
--- a/cqlite-core/src/storage/sstable/reader/data_access/full_index_stream.rs
+++ b/cqlite-core/src/storage/sstable/reader/data_access/full_index_stream.rs
@@ -404,19 +404,9 @@ impl SSTableReader {
     /// post-reconciliation early break (`drive_merge`) stops pulling, and the
     /// cancel-aware Drop teardown (cancel → drop receiver → join) then stops the
     /// producer promptly — bounding work WITHOUT any risk of under-return.
-    ///
-    /// `caller_schema` (issue #3097): the caller's AUTHORITATIVE schema, threaded
-    /// on to [`stream_all_partitions_via_full_index`](Self::stream_all_partitions_via_full_index)
-    /// so a `V5_0Uncompressed` reader (whose `nb` header carries no embedded
-    /// schema, hence a header-derived schema whose clustering columns bear the
-    /// placeholder `clustering_key` name) decodes clustering columns under the
-    /// caller's real names. `None` keeps the reader's own four-tier lookup — which
-    /// is exactly what the compaction caller passes today, so its byte-parity
-    /// resolution is unchanged.
     pub(in crate::storage::sstable::reader) async fn stream_all_partitions_cancellable<F>(
         &self,
         scan_cancel: &ScanCancel,
-        caller_schema: Option<&TableSchema>,
         mut emit: F,
     ) -> Result<()>
     where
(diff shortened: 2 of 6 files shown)


pmcfadin:
🧭 Review-first triage (Yoshi, flow-lead worker):

**rust-reviewer:** APPROVE (1 NIT).
**roborev (codex):** 1 Medium finding → triaged as **BLOCKER**.

Both flagged `compaction.rs:598` but disagree: roborev says forwarding `Some(schema)` diverges the streaming compaction path from the materializing `iterate_all_partitions_for_compaction` path (reader-derived) when schemas differ; rust-reviewer calls it a net improvement. To eliminate the ambiguity and keep acceptance criterion 3 (compaction byte-parity) provably safe, scoping the caller-schema threading to the query/merge arm ONLY — compaction's schema resolution restored to exactly its origin/main behavior. Fix round in progress: lite re-cert + compaction-parity target + re-roborev before the full gate.

pmcfadin:
🧭 Re-review round 2 (Yoshi): roborev flagged a **real coverage gap** (BLOCKER) at `summary_scan/mod.rs:591` — the query arm's full-index/sequential fallback (when Summary.db is absent/unusable) still discards the caller schema, so clustering cols can surface NULL there; both current tests force a usable summary and miss it. Fixing per roborev's own recommendation: thread caller schema through the query-arm fallback while keeping compaction's decode provably unchanged (compaction call sites pass None), plus a new test that forces the fallback path (absent/unusable summary). compaction-byte-parity must stay 12/12.

pmcfadin:
🧭 Re-review round 3 (Yoshi): roborev clean of correctness/parity issues — only 1 **Low** left (test-coverage): the fallback test removes only Summary.db (covers full-index route) but not the sequential-scan fallback (needs Index.db removed too). Folding in a both-removed case to fully pin the sequential path, then handing to the closer for the full gate + final roborev + merge-on-green. compaction-byte-parity stays 12/12; net src diff unchanged.

pmcfadin:
🧭 Gate-of-record verdict + override decision (Yoshi, flow-lead):

Full `agent-gate.sh` (macOS, cap=1) returned **FAIL** — but the SOLE failing component is `tooling-tests → scripts/tests/test_gen_perf_corpus_3068.sh`. Independently verified this is a **pre-existing, diff-unrelated, macOS-local** failure, NOT a defect in this PR:
- This PR's diff touches **zero** scripts/tooling (only `cqlite-core` reader files + one `cqlite-flight` test).
- `test_gen_perf_corpus_3068.sh` is **byte-identical to origin/main** (empty diff) → origin/main fails it identically on macOS.
- Root cause: macOS `$TMPDIR` (`/var/folders/...`) symlinks to `/private/var/folders/...`; the #3068 prune generator emits realpath'd `/private/var/...` paths vs the test's `/var/...` expected set. Linux CI (`$TMPDIR=/tmp`) is unaffected.
- **All diff-relevant components PASS**: core-tests, compaction-byte-parity (12/12), query-semantics-oracle, flight-query-semantics-oracle, arrow-parity-guard, memory-budget, write/integration/cli tests.
- The GitHub **`required` check (Linux) = SUCCESS, mergeState = CLEAN** — the enforced merge gate is green.

Decision: this is the known `macOS gate ≠ Linux CI` divergence on a diff-unrelated tooling test — **not a real merge blocker for #3097**. Proceeding to final roborev + auto-merge on the green Linux required check. Filed follow-up **#3135** to canonicalize the expected paths so the local macOS gate goes green.

Issue #3140, opened 2026-07-30
Flight fast arm aborts do_get on a simple cell tombstone: bypass surfaces Value::Tombstone where the merge arm drops it, so Arrow rejects it ("expected Text value, got Tombstone")


pmcfadin:
## Summary

On a **single-generation** table, the #3058 fast (bypass) arm surfaces a **simple CELL tombstone** as a raw `Value::Tombstone` in the assembled row, where the k-way MERGE arm's `assemble_read_cells` **drops** it (so the column reads null). The Arrow encoder then hard-errors on the fast arm and `do_get` aborts with **zero rows**:

```
column 'row_col': expected Text value, got Tombstone(TombstoneInfo {
  deletion_time: 1609545600000000, tombstone_type: CellTombstone,
  local_deletion_time: 1782341946, ttl: None, range_start: None, range_end: None })
```

This is a **read-path ARM DIVERGENCE on real Cassandra bytes**: the merge arm is correct (it returns Cassandra's rows), the fast arm aborts.

## Repro — Cassandra-written, one generation

Fixture: `test-data/datasets/sstables/test_tomb/static_with_tombstones-*` (real Apache Cassandra 5.0 flush, binaries COMMITTED; `ck = 3` carries a cell tombstone on `row_col`, written by `DELETE row_col FROM ... USING TIMESTAMP T_GEN2 WHERE pk=1 AND ck=3`).

Ticket: `SELECT *` over
`CREATE TABLE test_tomb.static_with_tombstones (pk int, ck int, stat_col text static, row_col text, PRIMARY KEY (pk, ck))`,
at a pinned `CQLITE_TTL_NOW_OVERRIDE_SECS=1800000000`:

| arm | result |
|---|---|
| `CQLITE_FLIGHT_MERGE_PATH=merge` | **3 rows** — `ck=1 row_col=row_1`, `ck=3 row_col=<null>`, `ck=6 row_col=row_6` (Cassandra's answer; the row-tombstoned `ck=2` and range-deleted `ck=4..5` are correctly hidden) |
| `CQLITE_FLIGHT_MERGE_PATH=bypass` | **0 rows** + the hard error above |

## This is DISTINCT from #3094

[...]

Issue #3130, opened 2026-07-30
roborev-lints wall-clock guard: hardcoded scan roots omit cqlite-flight/tests (and every non-core/cli test tree), so "MECHANIZED — a reintroduction FAILs the fast loop" is false there; TIME_TOKEN also cannot see a yield_now poll-count bound


pmcfadin:
## Summary

CLAUDE.md tells every agent that three roborev-finding classes are **MECHANIZED** as `--lite` gate
lints (`roborev-lints`) "so a reintroduction FAILs the fast loop, not a review round". For the
wall-clock-assert class (#2642, `scripts/tests/check-no-wallclock-asserts.sh`) that claim holds for
**two** test directories and **one** non-recursive file glob, and is **false everywhere else** — most
consequentially for **all of `cqlite-flight/tests/`**, which has never been in the scan roots.

The scan roots are a **hardcoded allowlist** of two paths. The guard has no positive-control test
against the real roots, so silently dropping (or never adding) a root is undetectable by its own
self-test. Separately, the `TIME_TOKEN` regex only recognizes `Duration`/`Instant`-shaped
measurements, so the semantically identical **poll-count / busy-spin stabilization** hazard
("assume the other task gets scheduled within N `yield_now()` polls") is not matched at all — and it
is not documented as a non-goal either, so today it is simply ambiguous.

Found during #3095 delivery while diagnosing a load-sensitive Flight flake (now #3127). Not fixed
inline, to keep that diff scoped.

## The gap

**Blind spot 1 — hardcoded scan roots; `cqlite-flight/tests/` is not among them.**

`scripts/tests/check-no-wallclock-asserts.sh:62-66` — the entire default surface:

```bash
  for p in \
    "$REPO_ROOT/cqlite-core/tests" \
    "$REPO_ROOT/cqlite-cli/tests"; do
    [ -e "$p" ] && ROOTS+=("$p")
  done
```

plus `:83-87`, a deliberately **non-recursive** top-level glob added by #2902:

```bash
  shopt -s nullglob
  for f in "$REPO_ROOT"/tests/*.rs; do ROOTS+=("$f"); done
```

There is no `cargo metadata`, no workspace-member enumeration, no `find` — just those two dirs and
those top-level files. Enumerated coverage of every Rust test location in the workspace
(`Cargo.toml:2-13` members: `cqlite-core`, `cqlite-cli`, `cqlite-flight`, `bindings/python`,
[...]

Issue #3127, opened 2026-07-30
Flaky under load: fast_arm_stream_stops_when_the_client_drops_it — the fast arm’s ~2177-row buffered read-ahead exceeds the 800-partition fixture, so `settled < PARTITIONS` is a CPU-scheduling race, and its yield_now stabilization loop cannot tell "walk stopped" from "walk starved"


pmcfadin:
## Summary

`cqlite-flight/tests/issue_3058_bypass_path_taken.rs::fast_arm_stream_stops_when_the_client_drops_it`
(the spec-R7 mid-stream-cancel pin for the #3058 single-source `do_get` merge bypass) is
**load/CPU-scheduling sensitive**. Its stop assertion rests on a bound the test's own doc comment
claims holds "by construction" (`CANCEL_EGRESS_CEILING`,
`cqlite-flight/tests/issue_3058_bypass_path_taken.rs:55-66`) — but that ceiling caps only the
**Flight-level credited egress**, downstream of two intermediate buffered layers whose combined
read-ahead is roughly **2.7x the fixture's size**. The bound is therefore a probabilistic margin,
not an invariant, and the convergence ("stabilization") loop that reads it cannot distinguish
"the walk stopped" from "the walk was descheduled".

Found during delivery of #3095 (Flight static-column `do_get` semantics); reported by the
implementing agent and **not fixed inline to keep that diff scoped**.

## Observed symptom

- Failed **once** while running under **maximal parallel load** (the fleet-normal condition: a full
  gate plus concurrent nextest lanes on the same box).
- Passed on **3 subsequent isolated re-runs**, unchanged.

This is a **load/scheduling flake**. The production behaviour the test covers — the fast/bypass arm
stopping its scan when the client drops the response stream — was **NOT shown to be broken**. No
product defect is being reported here; the defect is in the test's bound and its observation
mechanism.

## Root-cause hypothesis

**Labelled a hypothesis, not a verified cause: this was investigated by source reading only and was
deliberately NOT reproduced** (reproducing a load-sensitive flake means adding load to a box already
running a gate).

### The two mechanisms, with `file:line`

**(1) The "stabilization loop"** — `cqlite-flight/tests/issue_3058_bypass_path_taken.rs:456-478`.
After `drop(stream)` (line 442) it samples the process-global monotone counter
`work_counters::stream_walk_partitions_parsed()` and declares the walk stopped on **two consecutive
[...]

PR #3141, opened 2026-07-30
fix(flight): Cassandra static-column SELECT semantics on both do_get arms (#3095)

### LINKED ISSUE #3095: Flight do_get returns wrong rows for STATIC columns on the merge arm: phantom ck=null row + statics null on real rows (verified vs cassandra-5.0.8)

pmcfadin:
## Summary

Fixes Flight `do_get` STATIC-column semantics to match Cassandra 5.0.8 `SelectStatement.processPartition()`. Closes #3095.

On a static-bearing table the **merge arm** (the production path before this change) returned a result set that did not match Cassandra: a **phantom `ck = null` row** for every static-bearing partition (N+1 rows where Cassandra returns N), **and** the static column as **null on the real clustering rows**. The single-generation scan arm had the mirror-image defect: it injected statics correctly but **dropped static-only partitions entirely**.

## Root cause (deeper than the issue's original diagnosis)

Two distinct defects, both on the read path:

1. CQLite's k-way merger is a **compaction** merger — it streams the reconciled static row as an ordinary `clustering_key: None` entry. The Flight drive loop faithfully emitted that as a row and never merged its values into the real rows.
2. `merge_static_cells` ran **before** `build_display_row`, so a static cell made a pure row tombstone look live. Statics were contaminating the row-liveness decision.

Fix (2) introduces `build_display_row_read_path`, which takes the tombstone decision over the row's **own** cells, riding the pre-existing `read_shadowing` seam (#1741) so physical-dump / sstabledump-parity / compaction consumers are **byte-unchanged**.

### Cassandra authority (verified at the pinned `cassandra-5.0.8` tag)

`db/transform/Filter.java` → `applyToRow(Row row) { return row.purge(DeletionPurger.PURGE_ALL, nowInSec, enforceStrictLiveness); }`, with a **separate** `applyToStatic(Row)`. `BTreeRow.purge` → `update(...)` returns `null` when `info.isEmpty() && deletion.isLive() && BTree.isEmpty(newTree)` — the row's OWN liveness/deletion/btree. `FilteredRows.isEmpty() { return staticRow().isEmpty() && !hasNext(); }`. So the liveness decision is per-row over its own cells, statics arrive out-of-band via `staticRow()`, and **a static cell cannot revive a purged clustering row**.

Also verified: `hasRegularColumnsRestrictions = nonPrimaryKeyRestrictions.hasRestrictionFor(Kind.REGULAR)` — a **STATIC** restriction is *not* counted, so `returnStaticContentOnPartitionWithNoRows()` stays TRUE and the static-only row is produced and *then* filtered by the restriction. Pinned by `static_with_rows/static-column-restriction`.

## Acceptance criteria

- **AC1** — N clustering rows ⇒ exactly N rows, each carrying the static value, no `ck = null` row. `issue_3095_flight_static_columns.rs`: `static_clustering_shape/select-star` (1 row, was 2), `static_with_rows/select-star` (13, was 16), `static_with_tombstones/*` (ck=2/4/5 hidden ⇒ 3 rows).
- **AC2** — static-only partition: bare `SELECT *` ⇒ exactly 1 row (null clustering + null regular); with a clustering or regular restriction ⇒ 0 rows. Same file, `pk=99` cases.
- **AC3** — both arms agree at a PINNED `now` over the same bytes; every case runs `CQLITE_FLIGHT_MERGE_PATH=merge` and `=bypass` and asserts identical rows **and order**, with probe-proven distinct arms.
- **AC4** — static case added to the query-semantics oracle (`test-data/query-semantics-oracle.json`, `static_with_rows__select_star`), green fail-closed on both `query_semantics_oracle_parity.rs` and `query_semantics_flight_parity.rs`. Physical-dump parity cannot catch this class (#1742).
- **AC5 — delivered NARROWED.** A static column no longer forces the merge arm (deletion-free static-bearing SSTables now take the single-generation bypass, proven on Cassandra-written `test_deltas.static_with_rows` and `test_writeparity.static_clustering_shape`), but a static-bearing SSTable that declares any deletion still fails closed to the merge arm pending **#3140** (cell-tombstone arm divergence), as does a static column present on disk but absent from the ticket DDL. New `BypassReason::StaticColumnsWithDeletions` keys on `(schema-declared OR on-disk statics) AND SSTableReader::may_contain_deletions()`, the latter being Cassandra's own no-deletion test (`EncodingStats.minLocalDeletionTime != DELETION_TIME_EPOCH`). **Owner-approved narrowing** rather than expanding this diff into the Arrow/tombstone fix.

## Wiring evidence

`StaticMergeSource` was initially installed on one route only. The per-partition decisions now live once in `statics::StaticPartitionState`, behind the single choke point `statics::drive_partition_rows` (streaming via an `emit` callback + `ControlFlow`, so no per-partition `Vec<QueryRow>` and `LIMIT` still breaks early), called by **both** `drive_merge` and `drive_aggregate`. Without this, `SELECT count(*)` would have kept counting phantom rows and **disagreed with `SELECT *`** over the same bytes. Per-route tests `static_with_rows/count-star` (13, not 16) and `static_with_rows/produce_from_paths` were each **verified to FAIL when that route's wiring was temporarily reverted** — a positive control, not merely a green test.

## Review history

Two full review-first rounds before any full gate (#2086), each `rust-reviewer` + roborev:

- **Round 1** — 2 blockers + 3 roborev findings ⇒ **6 blockers** dispatched. Both reviewers independently converged on the central defect (`emitted_clustering_row` set before suppression ⇒ tombstone-only partitions lost the static row, and a live **arm divergence** re-opening the very hole the AC5 exclusion covered).
- **Round 2** — all 6 verdict **CLOSED**; 2 NEW blockers found: the `count(*)`/`produce_from_paths` wiring gap above, and the AC5-exposed fast-arm Arrow hard-error on real Cassandra bytes (originally mis-attributed to #3094 — corrected; #3094 is a **CQLite-written, both-arms** defect, #3140 is **Cassandra-written, fast-arm-only**).
- Nits fixed: per-fixture fail-closed booleans (the old `ran >= 2` floor could pass with a committed fixture missing), the gate lane's inherited unrelated SKIP predicate, dead motion in `display_row.rs`, and the per-block `partition_complete` residual now recorded with its direction.

New direct tests where there were none: `cqlite-flight/src/statics_tests.rs` (`SourceStep` transition table — its `observe` distinguishes phantom `ck=null`, missing static, lost static-only row, and double emit, and panics on any raw `PendingRow`) and `cqlite-core/.../display_row_tests.rs` (12 tests: `row_is_visible` totality, statics-don't-revive-a-tombstone, the physical inject-first order pinned side-by-side as a byte-parity guard, `on_partition_close` × {complete, incomplete, physical, after-visible-row}).

## Gate

`scripts/agent-gate.sh` change is **strictly additive**: the static lane is appended to `run_flight_query_semantics_oracle` under the same `CQLITE_REQUIRE_FIXTURES=1`; `COMPONENTS` and the dispatch are untouched, no component's condition loosened, `--lite` unaffected. The full `AGENT-GATE SUMMARY` of record is posted separately by the closer.

`on_partition_close` now carries `complete`, true only at `END_OF_PARTITION`/next-header, so all three decode sites agree and a truncated body no longer synthesizes a row. `CompactionPolicy` does not override the hook — compaction untouched.

## Notes for reviewers / CI

- The full gate **requires** `bash test-data/scripts/fetch-datasets.sh` first; without the corpus 20 core test binaries fail with explicit fixture-absence messages and the gate fails closed per #2078.
- Campsite rule honored throughout (no `CQLITE_ALLOW_FILE_GROWTH`): `row_decoder/mod.rs` shrunk **below its pre-change size**, and the buffered drive loops split out of `producer.rs` (3437 → 3243).

## Known residuals (filed, not fixed here)

- **#3140** — fast-arm cell-tombstone / Arrow divergence; AC5 fails closed on it.
- **#3127** — `fast_arm_stream_stops_when_the_client_drops_it` load flake (its `CANCEL_EGRESS_CEILING` "holds by construction" claim is arithmetically wrong: ~2177 buffered rows vs an 800-partition fixture).
- **#3130** — `roborev-lints` wall-clock guard scan roots omit `cqlite-flight/tests` entirely.
- An all-clustering-rows-deleted-with-live-static partition has **no Cassandra-written corpus fixture**; that half is covered by unit tests plus an absolute-shape CQLite-written differential. The owed fixture is noted for follow-up.

Refs #3058, #1741, #1742, #3042, #2086, #2087.


pmcfadin:
## Gate of record + C intent audit + final roborev — cleared for merge

Endgame certification for #3095, run in a disposable `flow-closer` context.

### Gate of record (full `agent-gate`, 29 components) — `RESULT: PASS`

Certified SHA: `807f295703f8fda07767b2ba68a3147143d572a1`. Verbatim block:

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.dcSS8q
commit: 807f295 branch: issue-3095-flight-static-columns dirty: no
tree-start: 807f295703f8 dirty: no digest: 3c7116486ed0
tree-end: 807f295703f8 dirty: no digest: 3c7116486ed0
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (5s)
clippy:            PASS (2s)
roborev-lints:     PASS (22s)
core-tests:        PASS (81s)
tombstones-scan:   PASS (1s)
scan-offload-guard: PASS (1s)
work-counters-guard: PASS (17s)
byte-budget-guard: PASS (15s)
arrow-parity-guard: PASS (1s)
memory-budget:     PASS (10s)
integration-tests: PASS (0s)
format-compat:     PASS (1s)
write-tests:       PASS (9s)
cli-tests:         PASS (5s)
python-bindings:   PASS (70s)
node-bindings:     PASS (235s)
delivery-telemetry: PASS (0s)
oom-audit:         PASS (1s)
parity-report:     PASS (1s)
tooling-tests:     PASS (330s)
minimal-build:     PASS (0s)
smoke:             PASS (3s)
logs: /tmp/agent-gate.dcSS8q
summary-file: /tmp/gate-3095-full-2.txt
RESULT: PASS
```

`tree-integrity: PASS` with `tree-start` == `tree-end` digest (`3c7116486ed0`), so the worktree did not mutate mid-run (#2926).

### Why there are TWO gate runs on this PR — the first FAILed *environmentally*, not on this change

Being explicit so this does not read as retry-until-green.

An earlier run (`run-id: /tmp/agent-gate.P72GLh`, same `commit: 807f295`, `tree-integrity: PASS`) returned `RESULT: FAIL` on exactly two components — `core-tests` (490s) and `memory-budget` (491s) — with the other 27 green. Cause: the test helpers derive `schemas_root()` as `datasets_root().join("../schemas")`, so `CQLITE_DATASETS_ROOT=/data/datasets` resolved to `/data/schemas`, which did not exist. Every failing test panicked in schema *discovery* on a missing `basic-types.cql` / `collections.cql` / `wide-rows.cql`, before any code from this change executes.

Failing tests were, in `dead_cache_delete_tests`: `memory_stats_semver_shape_preserved`, `small_config_budget_forces_eviction_and_bounds_residency`, `config_block_cache_max_size_is_the_b1_budget`, `direct_reader_open_honors_block_cache_disabled` — plus the three `memory_budget::*` cases, `issue_1494_converter_alloc_budget` and `issue_2075_row_assembly_alloc_budget`.

Attribution, established before re-running:
1. All five failing test *files* are byte-identical between `807f295` and `origin/main`.
2. The panic fires in schema discovery, ahead of the read path under test.
3. Re-running the **already-built** binaries against a root with a real sibling `schemas/` gave a full pass on the **unchanged tree** — no rebuild, no source edit. `dead_cache_delete_tests` went 4/4.

Remedy was environmental only: a `/data/schemas` symlink to the checkout's `test-data/schemas` (17 `.cql` reachable through the exact `datasets/../schemas` path). It lives outside the git tree, so it cannot dirty the worktree or affect `tree-integrity`, and this branch touches zero schema files. **No source was changed to turn the gate green.** The tree at `807f295` is byte-identical across both runs; only missing fixtures were supplied, the same class of fix as fetching an absent corpus.

The preflight gap that let `--preflight-fixtures` report `STATUS: OK` while the schemas root was absent is filed as **#3148**.

### C — intent audit: OVERALL PASS

Oracle-driven issue, so `spec-auditor` was anchored to #3095's acceptance criteria (no OpenSpec change exists).

- **AC1 satisfied**
- **AC2 satisfied**
- **AC3 satisfied**
- **AC4 satisfied**
- **AC5 satisfied (narrowed, owner-approved; residual #3140)**

AC5 as delivered, narrowed with explicit owner approval: a static column no longer forces the merge arm — deletion-free static-bearing SSTables now take the single-generation bypass, proven on the Cassandra-written `test_deltas.static_with_rows` and `test_writeparity.static_clustering_shape`. A static-bearing SSTable that **declares any deletion** still fails closed to the merge arm pending **#3140** (cell-tombstone arm divergence), as does a static column present on disk but absent from the ticket DDL. #3122 (#3094, cell-tombstone to Arrow encoder) has not landed, so that fail-closed exclusion remains necessary.

No scope drift: every non-test file in the diff is the fix, a campsite split, or the `map_scan_error` -> `map_core_error` dedupe the choke point required.

**Wiring evidence confirmed on three routes** — `drive_aggregate` (`producer_drive.rs:208`), `drive_merge` (`:111`, exercised from the public `produce_from_paths`), and the streaming `do_get` row route via `StaticMergeSource`; all reaching the static semantics through the single `statics::drive_partition_rows` choke point. The `count-star` and `produce_from_paths` tests were each verified to FAIL when their route's wiring was reverted.

Two points that strengthen the record:

- **The AC1 evidence hole is narrow.** A corpus scan (162 goldens, 118 static-bearing partitions) found no fixture with a static partition where *all* clustering rows are purged. However `test_tomb.static_with_tombstones` **is** Cassandra-written with a live static row over 6 clustering positions, 3 purged (row tombstone `ck=2`, range tombstone `ck=4-5`), and the lane asserts an **absolute 3-row shape** — a static cell reviving a purged row yields 5-6 rows and fails it. So the revival property does have a Cassandra-written oracle; only the strictly narrower all-rows-purged conjunction does not. This is a reconciliation property, not a framing one, so the #3042 symmetric-round-trip blind spot does not apply (that needs a writer error and reader error that cancel; there is no encode/decode pair here).
- **The fail-closed fixture floor was verified empirically**, not by reading: with `test_tomb` removed from a staged dataset root, `CQLITE_REQUIRE_FIXTURES=1` failed with a named fixture-absent error, and *without* the env var it still failed on the unconditional per-fixture assert (`issue_3095_flight_static_columns.rs:824`). The old loose `ran >= 2` counter is gone.

One non-blocking residual is being filed separately and deliberately **not** fixed here (scope): the tombstone branch of `build_display_row_read_path` has no end-to-end assertion on the core `Database::execute`/CLI surface — unreachable on the Flight fast arm by construction, but reachable and untested there, and the existing `point_vs_full_differential` case is CQLite-vs-CQLite over the same decoder, so it cannot catch a revival bug.

### Final roborev — clean

`roborev review --since 9b7551a --agent codex --model gpt-5.6-sol` over 20 commits: **No issues found.** Summary returned: *"Implements Cassandra-compatible static-column query and merge semantics with expanded core, Flight, oracle, route, and fixture coverage"* — it names the actual change, so this is a genuine pass and not a vacuous no-diff review. Review-first (#2086) had already run two full rounds pre-gate (6 blockers then 2, all closed); nits were fixed inline, so there is no deferred nit follow-up.

### Pre-merge

- `scripts/flow/premerge-assert.sh 3141 807f295703f8fda07767b2ba68a3147143d572a1` -> `PREMERGE: OK` (#2456).
- Comments on #3095 and #3141 re-read immediately before arming: no `HOLD:` order.
- `main` moved twice during the endgame (`9b7551a` -> `55dd6ba`, #3139 lifting `SSTableRowIteratorAdapter` out of `merge/mod.rs`) with 0 conflicts against `807f295`; that is the compaction/write-engine path, not the Flight read path. GitHub reports `MERGEABLE`, the PR head still equals the certified SHA, so the certification stands and the branch was deliberately **not** rebased or re-gated.

Arming `gh pr merge --auto --squash --delete-branch`. The Flight e2e tier is mandated by this diff (`cqlite-core/**` + `test-data/**`) and runs as expected; no `ci:waive:*` label applied.

Issue #2419, opened 2026-07-14
WS2 (epic #2313): saturation instrumentation — fd/thread/RSS/blocking-pool/channel-depth gauges in obs.rs


pmcfadin:
**Workstream WS2 of epic #2313** (flight read-throughput saturation program). Filed on activation 2026-07-13.

## Problem

The saturation ramp (#2313 §C3/C4) can only attribute a throughput plateau to a resource if that resource is *observed*. The `cqlite-flight` server today exposes a good RPC-level surface in `cqlite-flight/src/obs.rs` — `cqlite.rpc.*` counters, `cqlite.rpc.in_flight`, `cqlite.rpc.phase.duration` / `cqlite.rpc.phase.active` (the #2361 per-phase gauge), plus `query.rows_scanned` — but it has **no gauges for the OS-level resources that saturate first** under concurrent readers. The research ranks the expected order-of-failure as thread/scheduler collapse → no-admission-control queueing → fd exhaustion → memory, and every one of those is currently invisible:

- **Thread count** — a wide `do_get` merge historically over-subscribed the box; #2316 added `cqlite.merge.producer_threads`, but there is no process-wide `/proc/<pid>/task` gauge to see the aggregate across N concurrent queries.
- **fd count** — the read path does a fresh `File::open` per SSTable per scan (`cqlite-core/src/storage/sstable/reader/source.rs`, no reader pool by #815 design); N×M fds vs a container ulimit ~1024 → `EMFILE`. No fd gauge exists.
- **RSS** — the flight path bypasses the query engine's result-byte budget, so RSS ≈ N × per-scan peak; no process-memory gauge.
- **Tokio blocking-pool depth / egress channel depth** — streaming runs on `spawn_blocking` threads (`cqlite-flight/src/streaming.rs:317`), and the producer→consumer `sync_channel` is bounded at `STREAMING_CHANNEL_CAPACITY = 256` (`cqlite-core/src/storage/write_engine/merge/mod.rs:515`). Neither the blocking-pool queue depth nor the channel occupancy is sampled, so "stuck in do_get" (in_flight>0, rows_scanned flat) can't be distinguished from disk-bound.

Without these gauges the ramp can measure that throughput plateaus but not *why*, defeating the epic's purpose (prove the only ceiling is disk bandwidth, or file the resource that binds first).

## Scope — add saturation gauges to `obs.rs`

[...]

PR #3142, opened 2026-07-30
refactor(merge): lift SSTableRowIteratorAdapter + producer-thread machinery out of the 13.5k-line merge/mod.rs (#3139)

### LINKED ISSUE #3139: extract SSTableRowIteratorAdapter + merge producer-thread machinery out of the 13.5k-line merge/mod.rs (pure code motion, prerequisite for #3120)
Prerequisite lane for **#3120** (P0: k-way merge adapter treats a dead producer as end-of-input → silent truncation **including compaction output**). Owner-directed sequencing: minimal extraction first as a pure-code-motion PR, then the semantic fix on a clean file.

## Why

`cqlite-core/src/storage/write_engine/merge/mod.rs` is **13,558 lines** — note #3120's body claims ~7.5k, which is stale; measured on `origin/main` @ `9b7551a`. The gate's `file-size` component FAILs on any change that grows an already-over-threshold `.rs` file (campsite target ~800 for source), so #3120's fix cannot land without either this extraction or the `CQLITE_ALLOW_FILE_GROWTH=1` escape hatch.

The owner chose extraction over the override, for a good reason: #3120 changes **compaction semantics** (a panicking producer must fail the compaction instead of writing a short SSTable). That diff has to be genuinely reviewable, and it will not be if it is buried inside a 13.5k-line file. This lane makes the #3120 diff small.

## Scope — deliberately MINIMAL, and strictly behavior-preserving

Lift **only** `SSTableRowIteratorAdapter` and the producer-thread machinery it owns out of `merge/mod.rs` into their own module (suggested: `merge/producer_iter.rs`; final name at the implementer's discretion). Explicitly **not** in scope: splitting the rest of the file, reorganising the compaction merge core, or any behavior change whatsoever.

**This must be pure code motion.** No logic edits, no signature changes beyond the visibility strictly required to move the code, no "while I'm here" cleanups, no renames. If something cannot move without a judgement call, **stop and report it** rather than making the call — that is a signal the boundary is wrong, and it is exactly the information needed to decide whether to fall back to the override.

## Acceptance criteria

1. `SSTableRowIteratorAdapter` + its producer-thread machinery live in their own module; `merge/mod.rs` shrinks by roughly the moved amount.
2. **Zero behavior change**, and the diff shows it: moved lines are identical modulo `use`/visibility. Anything that is not pure motion is called out explicitly in the PR body.
3. Full `agent-gate.sh` PASS. Because this touches the **write/compaction** path, the byte-parity compaction suite and write suite are the evidence bar — not the read-path loop.
4. `file-size` PASSES **without** `CQLITE_ALLOW_FILE_GROWTH=1`. If an override turns out to be unavoidable, that is a finding to report, not something to paper over.
5. The new module is under the ~800-line source target, or a documented reason why not.

## Notes for the implementer

- The `#2419` egress-depth gauge (`merge/channel_depth.rs`, 225 lines) is coupled to this machinery: the `channel_depth::sent`/`received` balance is keyed on `msg.is_ok()`, with a post-join reconcile residual. **Do not alter that accounting in this lane** — #3120 will have to thread a third, untracked message kind through it, and keeping this PR inert makes that change legible.
- Two producer shapes exist and both matter downstream: `producer_thread` (path-based, in `mod.rs`) and `producer_thread_from_reader` (shared-reader, in `merge/from_readers.rs`, which carries the #3120 KNOWN GAP note). Decide deliberately whether the `from_readers` shape moves too or stays put, and say why in the PR body.
- Related epic: **#1116** (source file splits). This is a targeted slice of it, not the whole thing.

Routing: mechanical refactor — no OpenSpec change. Sibling: **#3124** (remaining discarded `JoinHandle`s on the multi-generation path, unclaimed). #3106 (merged, PR #3123) established the fail-closed pattern this class of fix follows.

pmcfadin:
Closes #3139

Prerequisite lane for **#3120** (P0: k-way merge adapter treats a dead producer as end-of-input → silent truncation **including compaction output**). Owner-directed sequencing: minimal pure-code-motion extraction first, then the semantic fix on a clean file — chosen over the sanctioned `CQLITE_ALLOW_FILE_GROWTH=1` override so the #3120 diff is genuinely reviewable rather than buried in a 13.5k-line file.

## What moved

`cqlite-core/src/storage/write_engine/merge/mod.rs`: **13,558 → 12,668** (−890).

- **`merge/producer_iter.rs` (480)** — the adapter struct, `RECV_CANCEL_POLL`, `open`, `producer_thread`, the `SSTableRowIterator` impl, and the `Drop` teardown. Channel/thread plumbing.
- **`merge/producer_iter_convert.rs` (489)** — `build_merge_entry`, `compaction_bound_to_mutation`, `extract_clustering_key_from_compaction`, `compaction_row_data_to_row_data`, `value_to_row_data`. Pure row conversion.

Two files rather than one because a single module landed at ~960 lines, and the `file-size` ratchet fails a **new** file at that size (base = 0). The split point is the natural seam: plumbing vs conversion.

## Pure motion — verified mechanically, not asserted

The base `origin/main` line ranges were diffed against the new files (blank-insensitive). The **only** differences are:

- `pub(super)` on the struct, its 6 fields, `open`, and 4 convert methods;
- three intra-doc links requalified with `super::` (`STREAMING_CHANNEL_CAPACITY`, `egress_budget`, `block_on_async` — targets stayed in `mod.rs`);
- one added `}` closing the split `impl`;
- one rustfmt signature re-wrap on `value_to_row_data`.

`mod.rs`'s diff is exactly two hunks: module decls in, moved regions out.

## Deliberate decisions (stated so a reviewer can disagree)

- **Depth-gauge accounting untouched.** `merge/channel_depth.rs` is not in the diff at all. `channel_depth::sent`/`received`/`reconcile_residual` call sites moved verbatim; the `msg.is_ok()` keying and the post-join residual are unchanged. Keeping this inert is deliberate: #3120 must thread a third, *untracked* message kind through that invariant (#2419), and that change is only legible if this PR doesn't touch it.
- **`MergeProducerError` + `From<Error>` deliberately left in `mod.rs`.** It is the wire protocol shared by both producer shapes, so #3120's third message kind becomes one enum edit visible to both, not a cross-module dance.
- **`from_readers.rs` stays put** (740 lines, under threshold). Moving it would have needed a third file or pushed `producer_iter` over. Effect on #3120: the two producer shapes are now peer files constructing the same `pub(super)` adapter and sharing one payload enum in the parent, and the terminator's *observation* point (`next`/`Drop`) is now in exactly one small file. Mild cost: new adapter state for a terminator must be declared in `producer_iter.rs` and set in both constructors — the same two call sites as today.
- **Three doc comments repointed** (`9d6bf96`): the `mod.rs` from_readers module doc, and `from_readers.rs`'s delegation note + #3120 KNOWN GAP note, all of which said `mod.rs` is the home of `producer_thread`/`next` and were made factually false by the move. Comment text only. This is the sole reason `from_readers.rs` appears in the diff; revert that commit if a zero-sibling-touch diff is preferred.

## Evidence

- Lite gate **PASS** at `9d6bf96` (HEAD), `tree-integrity: PASS`, **`file-size: PASS` with no `CQLITE_ALLOW_FILE_GROWTH` override** — the file *shrank*, which is what the ratchet wants (acceptance criterion 4).
- Beyond lite, because this touches the write/compaction path: `cargo build -p cqlite-core --no-default-features --features all-compression` (the gate's `minimal-build`) clean, and `issue_1017_live_cell_compaction_byte_parity` + `issue_1020_udt_frozen_compaction_byte_parity` pass (5 tests).
- The full `agent-gate.sh` of record — with the byte-parity compaction and write suites as the real evidence bar — runs once pre-merge in `flow-closer`.

## Scope note

`merge/mod.rs` is still **12,668 lines**, far over the ~800 campsite target. This lane only unblocks #3120; the rest remains **#1116**'s scope by design.

## Unrelated pre-existing finding

`cargo check -p cqlite-core --no-default-features` (i.e. *without* `all-compression`) fails on a dead `MAX_DECOMPRESSED_SIZE` in `sstable/compression.rs`. Not a configuration the gate builds, not caused by this diff, filed separately.


pmcfadin:
## ✅ roborev: CLEAN — and the pure-motion claim independently corroborated

codex (repo-pinned, no flags) reported **`No issues found.`** across all 3 commits.

**Non-vacuity substantiated** (#2964 guard, which matters here — a vacuous "no issues found" on a motion refactor is indistinguishable from a real pass): the review's own summary is branch-specific ("Refactors the SSTable producer iterator and conversion helpers into dedicated modules without changing behavior"), which cannot describe `origin/main`. Corroborated by token accounting — base `f603c7f` is an openspec/docs-only archive commit and `.roborev.toml` excludes `docs/**`/`*.md`, so a base review would have had a near-empty diff rather than 167k of input.

### Independent mechanical verification of "pure motion"

Not taken on trust, since any behavior delta here is disqualifying. A **whitespace-normalized multiset diff** of the 908 lines removed from `merge/mod.rs` against the 969 lines of the two new files accounts for **every** delta as an inert category, with **no unaccounted code line**:

- `pub(super)` on the struct, its 6 fields, `open`, and the 4 convert methods — required because `from_readers.rs` holds a sibling inherent `impl` on the same adapter.
- `use` re-imports in the new files, plus the 18 added lines in `mod.rs` (module docs, two `mod` decls, one `use`).
- 3 intra-doc links requalified (`super::block_on_async`, `super::STREAMING_CHANNEL_CAPACITY`, `super::egress_budget`), module `//!` headers, and the 3 repointed doc comments.
- Structural only: one `impl` opener + closing braces, one rustfmt re-wrap of the `value_to_row_data` signature.
- 3 new `#[cfg(feature = "write-support")]` attributes gating the new `mod`/`use` decls, consistent with the existing `mod from_readers;` — **no feature gate widened or narrowed**.

### #2419 egress-depth gauge: verified untouched

All 4 `channel_depth::` sites moved **byte-identically** (`mod.rs` now has 0). The decrement is still keyed on the data-only arm — `Ok(Ok(entry)) => { channel_depth::received(); self.received_count += 1; }` (`producer_iter.rs:380-389`) — and the post-join `channel_depth::reconcile_residual(residual)` (`:478`) is unchanged. This was the constraint most likely to break subtly, and #3120 depends on it being inert: that fix must thread a **third, untracked** message kind through this balance.

### #3120's defect correctly left unfixed

The KNOWN GAP note survives at `from_readers.rs:53-60`, correctly repointed from `mod.rs` to `producer_iter`. Nothing was raised about the dead-producer-as-EOF gap, which is right — that is #3120, deliberately out of scope here.

`rust-reviewer`'s independent pass is still running; the full gate of record (with the byte-parity compaction + write suites as the real evidence bar) runs once after it, in `flow-closer`.

pmcfadin:
## ✅ rust-reviewer: `BLOCKERS: 0`, `NITS ONLY: 5` — motion verified independently

Verdict on the central question: **yes, this is pure code motion.** Verified by a different method than roborev's, and the two agree.

### How it was verified (worth recording, because "trust me, it's just a move" is unfalsifiable)
- **Function-by-function comparison** against `main` for all nine moved items (`open`, `producer_thread`, the 5 convert fns, the `SSTableRowIterator` impl, `Drop`): every statement, `?`, match arm, early return, `Ordering::SeqCst`, `unwrap_or(0) as i64`, `RECV_CANCEL_POLL = 50ms`, and both `compaction_bound_to_mutation(start)/(end)` argument positions identical.
- **The arithmetic closes exactly**: 908 lines removed from `mod.rs` (101 + 807), 18 added → net −890 = the observed shrink. New files hold 903 moved content lines + 5 dropped structural braces = 908.
- **~20 grep anchors from `mod.rs:2649` to `:10809` all map to *n*−890 with ZERO drift** — which proves there is no insertion or deletion anywhere in the remaining 11k lines. That is the check that makes "nothing else changed" a fact rather than a hope.
- `channel_depth.rs` and `point_read.rs` anchors at **identical** line numbers in both trees; `from_readers.rs` shifts by a constant **+13**, fully accounted for.
- **Nothing dropped or orphaned**: every `#[cfg(test)]` block, `const`, and attribute (`#[allow(clippy::type_complexity)]`, the `#[cfg(test)]` on `value_to_row_data`/`egress_channel_capacity`) survived; all five sibling test modules remain declared.

### Two findings worth surfacing rather than burying
1. **Undisclosed deviation (my error in the PR body, now corrected below):** the `KNOWN GAP` section at `from_readers.rs:53-64` is a **13-line doc ADDITION**, not one of the "three doc comments repointed" — `#3120` appears nowhere in `origin/main`'s `cqlite-core/src`. Content verified accurate (neither producer shape sends a terminator: `producer_iter.rs:353-358`, `from_readers.rs:299-304` both just drop the sender; `next` maps `Disconnected → None` at `:414`). But in a motion-only PR an unlisted hunk makes an auditor stop, so it belongs in the exception list. **Additions and repoints need separate lines.**
2. **The #3120 false-premise comment and the #3120 gap note now live in different files.** `producer_iter.rs:458-464` carries the defect verbatim — *"A failed join (producer panicked) is ignored; the panic was already surfaced through the error channel"* — which is byte-identical to `main` (correctly **not** accidentally softened), but `producer_iter.rs` has no #3120 reference at all. The #3120 implementer's first read is `Drop` there, where the false claim sits unqualified. Being cross-referenced now, with the premise explicitly marked known-false.

Also fixing: the stale `mod.rs:6-11` header (still claims splitting is out of scope), and two now-cross-privacy intra-doc links in `from_readers.rs`. All four are **doc-only** and cost no gate cycle, since the full gate has not run yet.

### Confirmed good
- **Net visibility change is exactly zero** — a bare `struct` in `merge/mod.rs` and `pub(super) struct` in `merge::producer_iter` have identical effective visibility. Every `pub(super)` is load-bearing, and two items (`producer_thread`, `compaction_bound_to_mutation`) were **narrowed**, not widened.
- **#2419 gauge provably inert** — send side unchanged in place, receive side moved verbatim, post-join residual character-for-character identical including `SeqCst` and the `join()`-before-`load()` ordering.
- **Seam is sound** — nothing in `producer_iter_convert.rs` touches a channel/gauge/thread/`self`; nothing in `producer_iter.rs` decodes a cell.
- No `unwrap`/`expect`/`panic!` in either new file; both under the ~800 target.

### For the #3120 implementer (me, next lane)
Both terminator **observation** points are now in one 480-line file: `next` (`:364-417`) and `Drop` (`:433-480`), and `MergeProducerError` stayed in `mod.rs:456-471` so a third `Done` variant is one edit both shapes see. Residual friction: new adapter state (e.g. `saw_terminator`) must be declared in `producer_iter.rs:81-123` **and** initialized in both struct literals (`producer_iter.rs:229-237` and `from_readers.rs:246-254`, different files), and the **emit** side lives in `from_readers.rs`. So #3120 is three small legible edits rather than one.

Deferred as a follow-up, not done here: the five conversion helpers remain inherent methods on the adapter (`producer_iter_convert.rs:28` reopens `impl SSTableRowIteratorAdapter` for functions that never touch `self`). The reviewer agreed that is the correct call for a motion-only PR because it keeps every call site unchanged.

Issue #3144, opened 2026-07-30
merge conversion helpers are inherent methods on SSTableRowIteratorAdapter but never touch self (producer_iter_convert)


pmcfadin:
Deferred follow-up from #3139 (PR #3142), agreed by the reviewer as correctly out of scope there.

## Observation

`cqlite-core/src/storage/write_engine/merge/producer_iter_convert.rs:28` reopens `impl SSTableRowIteratorAdapter` for five functions that **never touch `self`**:

- `build_merge_entry`
- `compaction_bound_to_mutation`
- `extract_clustering_key_from_compaction`
- `compaction_row_data_to_row_data`
- `value_to_row_data`

They are pure `CompactionRow` → `MergeEntry` translation. The result is a module whose `impl` block is coupled to a type defined in a sibling module for no semantic reason.

## Why it was NOT fixed in #3139

#3139 was a **pure code motion** PR (a prerequisite for #3120, a P0 silent-data-loss fix), and its whole value was being provably inert — verified by function-by-function comparison plus line arithmetic closing to zero drift. Converting these to free functions would have changed every call site (`from_readers.rs:157`, `point_read.rs:250`/`:359`, and `mod.rs`'s tests), which is exactly the kind of churn that makes a "is this really just a move?" audit impossible. Keeping them as inherent methods was the right call there.

## Fix shape (acceptance)

1. Convert the five helpers to free functions (or associated functions on a purpose-built type if some grouping is wanted) in `producer_iter_convert`, and update the call sites.
2. Pure refactor — **zero behavior change**, same test coverage passing.
3. Full `agent-gate.sh` PASS. Note this is the **write/compaction** path, so the byte-parity compaction suite and write suite are the evidence bar, not the read-path loop.

## Sequencing

**Do this AFTER #3120 lands**, not before. #3120 adds a terminator protocol to this same machinery and will touch `producer_iter.rs` plus both producer constructors; landing a call-site-churning refactor first would make that P0's diff harder to review, which is the opposite of what #3139 was for.

Related: #3139 (the extraction), #3120 (P0 fail-closed fix), #1116 (source file splits — `merge/mod.rs` is still 12,668 lines).

Issue #3124, opened 2026-07-30
discarded JoinHandles on the multi-generation scan path: a dead task still reads as end-of-scan (silent truncation, query-engine full scan)


pmcfadin:
Split out of #3106 review (PR #3123). #3106 closed the **single-source** query-row path at its two boundaries; this issue covers the boundaries it did NOT close.

## The defect class

Same shape as #3106: a `tokio::spawn` whose `JoinHandle` is **discarded**, plus a consumer that reads channel-close as "the scan finished". A task that dies (panic → `unhandled_panic = Ignore` swallows it) drops its sender, and the consumer reports a clean end of stream — so the caller gets **fewer rows with no error**.

## Sites (all verified during the #3106 review)

1. `cqlite-core/src/storage/sstable/mod.rs:2194` — the k-way merge is spawned with a discarded `JoinHandle`.
2. `cqlite-core/src/storage/sstable/mod.rs:2215` — each per-reader `scan_stream_admitted` likewise.
3. `cqlite-core/src/storage/sstable/mod.rs:2361` — `rechunk_into_batches` reads `per_row.recv() == None` as "the scan finished".
4. `cqlite-core/src/storage/sstable/reader/data_access/scan_stream_windowed.rs:569` — `let _ = forwarder.await;` **deliberately discards the forwarder task's `JoinError`**, so on the chunk-stitching (compressed) branch a forwarder death still reads as a clean end of scan. Low likelihood (the forwarder only flattens/forwards already-decoded rows — no decode work), but it is the third hop on a path whose other two are now closed.

## Why it matters

This is the **≠1-generation** path, i.e. the query-engine full scan. It is off the #3106 Flight single-source repro, so #3106's fix does not reach it, but the user-visible failure is identical: a successful query returning silently short results.

## Relationship to the sibling issues

- **#3106** (merged via #3123): closed the single-source query-row path — outer producer thread (`Done` sentinel + `catch_unwind`) and inner batched-scan task (`BatchedScanStream` joins the handle, `JoinError` → `Error::Internal`).
- **#3120**: the k-way merge *adapter* (`write_engine/merge/from_readers.rs`), which also affects **compaction output**.
- **This issue**: the remaining discarded-`JoinHandle` spawn sites on the multi-generation read path.

Together these three are the full audit of "channel close treated as end-of-input" in the codebase. Worth landing them as a set so the property is uniform rather than arm-by-arm.

## Fix shape (acceptance)

[...]

PR #3152, opened 2026-07-30
fix(merge): fail closed when a merge producer thread dies instead of writing a short SSTable (#3120)

### LINKED ISSUE #3120: k-way merge adapter treats a dead producer thread as end-of-input (silent truncation, incl. compaction output)

pmcfadin:
Closes #3120

**P0.** The k-way merge adapter treated a dead producer thread as end-of-input. On the read path that meant a silently short result set; on the **write path it meant a rewritten SSTable missing rows — silent data loss at rest.**

Unblocked by #3139 (PR #3142), which extracted this machinery out of the 13.5k-line `merge/mod.rs` so this diff is reviewable.

## The defect

`SSTableRowIteratorAdapter::next` mapped `RecvTimeoutError::Disconnected → None` — "producer finished normally" — whenever the per-input channel closed. Both producer shapes dropped their `SyncSender` on exit with **no explicit terminator**, so "the walk finished" and "the producer thread UNWOUND" were indistinguishable. `Drop` additionally discarded the producer's failed join on the false premise that "the panic was already surfaced through the error channel" (it was not).

## The fix

**1. An explicit channel protocol.** New `merge/producer_msg.rs`: `MergeMsg { Item(MergeEntry), Failed(MergeProducerError), Done }`, plus `MergeProducerError::Panicked`. Both producer shapes wrap their body in `catch_unwind` and emit **exactly one** terminal message on every exit path. `MergeProducerError` moved out of `mod.rs` in the process, so **`mod.rs` net shrank 12,671 → 12,655**.

`Panicked` and the dead-disconnect error map to **`Error::internal`, not `Error::Storage`** — `Storage.is_recoverable() == true`, which would have advertised a deterministic dead-producer failure as *retryable* to the Node/Python error mapping.

**2. A state machine that makes the bad state unrepresentable.** `RunState { Streaming, Finished, Failed(MergeProducerError), Died }` in `producer_iter.rs`:
- `None` is reachable **only** from a `Done`-proven `Finished`.
- `Failed` **carries its payload**, so a repeat poll re-reports the same error rather than decaying.
- `Died` is **sticky** — permanently.
- `receiver.as_ref()?` (a second silent `None`) is folded into `Died`.

Note #3106's `TaskState` pattern does **not** transfer wholesale: `tokio::JoinHandle` is `Future + Unpin` and can be polled in place, whereas `std::thread::JoinHandle::join(self)` **consumes self and blocks**. So the terminator — not the join — is the primary mechanism, and `next()` stays non-blocking. Sticky `Died` is load-bearing here specifically because `RunReader::refill_buffer` propagates the `Err` but keeps **no** sticky error of its own; #3106's non-sticky `terminated` flag would have downgraded a dead producer to clean end-of-input on the *second* poll.

**3. The `#2419` egress-depth gauge stays balanced.** The terminator is untracked on **both** sides: one *exhaustive-match* `MergeMsg::is_tracked_data` at the send site, and the `Item` arm only at the receive site — replacing two divergent expressions of one rule (`msg.is_ok()` vs an `Ok(Ok(_))` pattern). This was the highest-risk part: a terminator counted on one side only makes the residual negative, which `reconcile_residual`'s `> 0` guard skips and `record`'s `max(0)` floor then **hides permanently**. So `debug_assert!(residual >= 0)` is now in `reconcile_residual`, pinned by `a_terminator_is_untracked_on_both_sides_so_the_residual_is_exactly_zero` against a **private** atomic (never the global `DEPTH` — #2451 flake class).

**4. `forward_row` no longer sends a mid-walk `Err`.** It previously emitted an `Err` into the data slot **while returning `Continue`**, so the walk kept going after sending what the consumer treats as terminal — exactly the #3106 regression shape, sitting in the code already. It now RETURNS the conversion error out of the emit callback and the thread body emits the single terminal `Failed`. **Behaviour delta:** the walk stops at the first row-conversion failure instead of continuing. Observable outcome to the merge is unchanged today (`RunReader::refill_buffer` returned that `Err` immediately anyway) and no test asserted the message text. Noted in the module header and the commit body.

## Tests — each watched FAILING pre-fix

Verified in a throwaway worktree with the terminator, `catch_unwind` and state reverted:

- `a_dead_producer_run_reports_an_error_that_never_decays_to_end_of_input`
- `the_path_based_merger_fails_when_a_producer_dies_instead_of_returning_short`
- `the_shared_reader_merger_fails_when_a_producer_dies`
- **`a_compaction_whose_producer_dies_fails_and_publishes_no_output`** — pre-fix this failed at *"must return `Err`"*, i.e. it **did rewrite a short SSTable and report `Ok`**. That is the data-loss claim demonstrated rather than argued.

Plus 4 `producer_msg` and 4 `producer_fault` unit tests. All use **control-arm-first**: the identical call runs unfaulted, the full count is recorded, and only then is the fault armed — so "the faulted run is short" is measured against a known-good baseline. Arms are scoped to **one input `nb-*-big-Data` path**, not the `TempDir`, because a TempDir-wide scope would kill whichever of K producers reached its first row first.

Two deliberate non-tests, stated so absence isn't read as an oversight:
- **There is no flush merge.** `WriteEngine::flush` → `flush_internal_async` writes the memtable straight through `SSTableWriter`, constructing no `KWayMerger` and spawning no producer thread. Criterion 3's write arm is **compaction only**; a flush-arm test would be vacuous by construction. Suggest the criterion be reworded "compaction".
- The pin lives **in-src** under `cfg(test)`. A pin in `cqlite-core/tests/*.rs` could not fire: the gate's `write-tests` runs `-p cqlite-core --features write-support` and never enables `producer-fault-injection`.

## File-size

`from_readers.rs` **741 → 457** by lifting its inline #2346 parity tests verbatim into `merge/from_readers_tests.rs` (364). New modules: `producer_msg.rs` 296, `producer_panic_tests.rs` 497. **No `CQLITE_ALLOW_FILE_GROWTH` override.**

## Evidence

Lite PASS at `28c47b7` (the final HEAD), `tree-integrity: PASS`. The full `agent-gate.sh` of record — with the **byte-parity compaction and write suites** as the real evidence bar for a write-path change — runs once pre-merge in `flow-closer`.

Follow-up #3144 (conversion helpers as inherent methods) is deliberately sequenced *after* this PR so it does not churn call sites ahead of a P0. Sibling #3124 (remaining discarded `JoinHandle`s on the multi-generation path) is still open and unclaimed; #3106 + #3120 + #3124 together are the complete audit of "channel close treated as end-of-input".


pmcfadin:
## 🔴 roborev: `BLOCKERS: 1` — the defect line itself is unpinned

Non-vacuity substantiated (#2964): job 7, branch `issue-3120-merge-adapter-fail-closed`, review cites `producer_panic_tests.rs:217`, `MergeMsg::Failed` and `RecvTimeoutError::Disconnected` — all branch-only symbols.

### The finding

`producer_iter.rs:549` — the `RecvTimeoutError::Disconnected → RunState::Died` arm is **the literal line that held the P0 defect** (`Disconnected → None`), and it has **zero behavioral coverage**. Verified independently:

- All four panic tests (`producer_panic_tests.rs:241/333/379/423`) arm the injected fault, which `catch_unwind` converts into a `MergeMsg::Failed` **terminator** — so they exercise the `Failed` arm and **never reach `Disconnected`**.
- The only other reference, `producer_msg.rs:281`, calls `dead_producer_error()` **directly** and asserts on its string / `is_recoverable` classification. It never routes through `next()`.

**Consequence: revert that arm to `return None` and the whole suite stays green.** The regression guard for the exact silent-data-loss defect this P0 fixes does not exist.

roborev rated it Medium; I am treating it as a **blocker**. An unpinned defect line on a P0 is the "test that proves nothing" class, and this is the fourth time in this defect family that green tests have not meant what they appeared to.

### Why it is subtle, and worth understanding rather than just patching

The `catch_unwind` is *so* effective that it converts every **injected** panic into a proper terminator — which means the injected-fault tests can no longer reach the bare-disconnect path at all. **The fix closed the front door so well that the back door became untestable by the front-door tests.** A bare disconnect remains reachable in production by unreportable deaths: `abort`, a double-panic, or a panic inside the terminal `send` itself.

### Fix in flight
One adapter-level test that drops every sender **without** sending `Done` or `Failed`, asserting that the **first** `next()` returns the dead-producer error *and* that a **repeat** `next()` returns the same **sticky** error rather than end-of-input — the latter pinning `Died`'s stickiness, which is separately load-bearing because `RunReader::refill_buffer` keeps no error of its own. It must be watched red against the reverted arm; if it cannot be made to fail there, it is not pinning what we think it is.

No production code changes — the state machine, gauge and `forward_row` were all reviewed clean.

### Everything else clean (independently corroborated)
- Terminator uses **blocking `send`**, never `try_send`.
- **Gauge symmetric**: `is_tracked_data` returns `false` for *both* `Done` and `Failed` on the send side, and `channel_depth::received()` lives only in the `Item` arm — so no negative residual can be laundered by the `> 0` guard / `max(0)` floor. That was the highest-risk axis and it holds.
- **Cancellation stays `Error::Cancelled`**, pinned by `cancellation_survives_the_terminator_protocol_as_cancelled` — a cancelled scan is not misreported as a dead producer.
- **`Drop` logs an unwind rather than re-panicking**, preserving the #2361 teardown pins.

None of the pre-declared INVALID classes were raised (`merge/mod.rs` size / #1116, the deferred #3124 / #3144, the absent flush-arm test, pre-`na`). `rust-reviewer`'s independent pass is still running.

pmcfadin:
## 🔴 rust-reviewer: `BLOCKERS: 1` (a **different** one) + 4 nits

### BLOCKER — the `debug_assert!` makes `Drop` panicking. This one is my fault: I mandated it without checking where it is reachable from.

`channel_depth.rs:109-114`. `reconcile_residual` has exactly **one** production call site — `SSTableRowIteratorAdapter::drop` step 4 (`producer_iter.rs:648`) — so **100% of the assert's reachable paths are inside a `Drop`**.

Failure mode: if a future change reintroduces the asymmetry the assert exists to catch, the manifestation is **not** a named test failure — it is a panic inside `Drop`. `teardown_tests.rs` drops mergers deliberately, and any test failing an assertion with a live `KWayMerger` in scope drops the adapter **while already unwinding** → double panic → **process abort**, which under libtest destroys the original assertion message *and every sibling test's result in the binary*.

The irony is instructive: that is exactly the failure this same PR engineers around 100 lines away, in `producer_fault.rs:554` — `if std::thread::panicking() { return; }`. The gauge assert needs the same treatment. The invariant stays checked; only its reporting changes.

For the record, the assert is **latent-only today**: `received_count` can transiently exceed `local_sent`, but `Drop` reads `sent_count` strictly *after* `join()`, which retires every pending `fetch_add`. The objection is the failure mode, not a current break.

### NIT — the "one shared predicate on both sides" claim is not what the code does
`channel_depth.rs:43-47`, `producer_msg.rs:157-161`, and the test docstring at `channel_depth.rs:255-257` ("an asymmetry between the two sites cannot hide from this test") all assert both sites express the rule via `MergeMsg::is_tracked_data`. They don't: the receive side is a hand-written `Ok(MergeMsg::Item(entry))` arm that never calls it, and on the send side `MergeMsg::Item` is built unconditionally so the call is a **tautology**. The test therefore drives `is_tracked_data` on both loops and **cannot** detect divergence at the real receive site.

The invariant nonetheless holds, by two mechanisms the docs fail to name: `next()`'s `match` is **exhaustive with no wildcard** (a 4th variant is a compile error), and `channel_depth::received()` / `received_count` / `add_merge_run_entry_decoded()` each appear at exactly one site, all inside the `Item` arm. Being rewritten to name the real mechanism — this is the **third** time in this defect family a doc has claimed coverage it doesn't have, and an overclaiming doc is what stops the next reader from noticing the gap.

Plus three cheap nits: an unused `#[derive(Clone)]` whose doc-comment justification is factually wrong; one bare `assert!` missing the "otherwise the fault never fired" rationale its siblings all carry; and a note that `Drop`'s new join-outcome warning is effectively unreachable under `panic="unwind"` (a legitimate backstop, but a reader will otherwise expect the panic tests to hit it).

### Verified independently — the reviewer's one caveat is resolved
It had no git access and diffed against a stale `f603c7f`. Against the **true** base `7247756`: `from_readers_tests.rs` is **364 added / 0 deleted** (zero `-` lines — a pure addition, extraction verbatim as claimed); `from_readers.rs` is 106/390.

### The three flagged risks all came back clean
- **RISK 1 gauge — clean.** Both sides exclude the terminator *independently*; both terminal-send sites are bare `send` with no gauge call or `local_sent` touch; the post-join residual formula is byte-identical to pre-change; the new pin uses a **private** atomic and only calls the global with `residual == 0` (a proven no-op), so the #2451 flake class is avoided.
- **RISK 2 verdict decay — clean, exhaustively.** Exactly two `None` paths, both `Done`-proven. No `?`, `.ok()`, `let else`, `unwrap_or`, or `_ => None` anywhere in `next()`. `Died` sticky forever; `Failed` re-reports its payload (load-bearing, because `RunReader::refill_buffer` propagates without setting `exhausted`); a producer that ends without closing the channel **hangs rather than lying** — the fail-closed direction. The #3106 take-handle-before-committing bug is structurally inapplicable since `next()` performs no join at all.
- **RISK 3 tests — non-vacuous.** Reachability is *proved by the control arm* (`healthy_rows == 24`, `complete.rows == 48`), not asserted by inspection. Arms scoped to one input's full `Data.db` path. The write-arm test asserts the real data-loss claim — it enumerates the output dir for `*TOC.txt` and requires it **empty** — which holds because `merger.merge()?` propagates *before* `writer.finish()`. Specifically credited: `assert_names_the_dead_producer` requiring **both** `"PANICKED"` and `INJECTED_PANIC_MESSAGE` is a materially better oracle than "an `Err` came out", since the bare-disconnect backstop contains `"TRUNCATED"` and neither of those.

Both blockers are being fixed in one round (this one plus roborev's `Disconnected`-coverage gap). No full gate spent yet.

pmcfadin:
## 🔴 Gate of record FAIL — one component: `minimal-build`

`run-id /tmp/agent-gate.Bs988d`, `commit c0f73d5`, `tree-integrity: PASS`, `datasets: 144 Data.db`, **30 components, 0 SKIP, exactly 1 FAIL**:

```
minimal-build:     FAIL (15s)
```

**Everything else passed — including the full write-path evidence bar:** `core-tests: PASS (624s)`, `write-tests: PASS (139s)`, `compaction-byte-parity: PASS (6s)`, plus `memory-budget`, `python-bindings`, `node-bindings`, `tooling-tests`, `smoke`, `oom-audit`, `parity-report`, both query-semantics oracles and every guard.

So the **behaviour** of this fix is certified; what broke is a **feature-combination compile error**. `minimal-build` is `cargo build -p cqlite-core --no-default-features --features all-compression`, and a 15-second failure is a compile error, not a test. It passes on `main` and on the #3139 base, so it is **ours** — the docs-only CITE-AND-WAIVE path does not apply, since the diff touches compiled inputs.

Likely cause: new code, a new `mod` include, or the `MERGE_ARMS` seam depending on something only present under a **default** feature (`write-support` being the obvious suspect — default-on, which is exactly why lite and all 29 other components were happy). That is the gap the doctrine warns about: **`agent-gate.sh` lite ≠ the full gate, and the full gate's feature-matrix components catch what the blast-radius loop cannot.**

**Explicitly distinct from #3143**, which I filed earlier for *bare* `--no-default-features` (without `all-compression`) failing on a dead `MAX_DECOMPRESSED_SIZE`. That is a different, pre-existing configuration the gate never builds; it must not be conflated with this regression or used to excuse it.

### Handling
- **`c0f73d5` is NOT certified.** Merge is not armed and will not be until a full gate PASSes.
- Fix dispatched to the implementer, with instructions to iterate on the **15-second** `minimal-build` command directly rather than rediscovering this 25 minutes later, and to fix it by **correct `cfg` gating — not an `#[allow(dead_code)]` blanket**, since an unused-item warning here may be telling us the wiring is wrong.
- A src change **invalidates this gate**, so the full gate of record re-runs at the new head, to a **fresh summary path**, and the FAIL block above is retired rather than cited again.
- roborev stays sequenced **after** a PASS gate — roborev can drop untracked files into the working tree and trip the #2926 `tree-integrity` check, which would void a 25-minute run.

No scope creep: the state machine, gauge accounting, `forward_row` and the tests are all reviewed-clean and stay untouched unless the compile error forces otherwise — and if it does, that is a scope signal I want reported, not absorbed.

Issue #883, opened 2026-06-20
Read path: SELECT * over multiple SSTable generations in one dir does not do cross-generation LWW merge + tombstone suppression


pmcfadin:
Discovered while building the DS11 reconciliation round-trip test (#707, epic #696).

**Observation:** When a table directory contains multiple SSTable generations (e.g. `nb-1-big-Data.db`, `nb-2`, `nb-3` from successive flushes), CQLite's `SELECT *` does not reproduce the fully-merged authoritative table state: it does not apply cross-generation last-write-wins or suppress a row/cell deleted in a later generation whose original lives in an earlier one. The read path appears to pick the latest generation (`DirectoryScan::latest_generation()`) and `select_executor.rs` suppresses tombstones only within the SSTable(s) it reads, not across generations.

**Why it matters:** The design doc (`docs/plans/2026-06-10-delta-scan-envelope-design.md`) describes the query engine's contract as "merge generations, suppress tombstones" — the opposite of delta-scan. That contract is only partially realized for multi-generation directories.

**Impact on #707:** DS11 was therefore validated against **Cassandra's authoritative ground truth** (captured from the live container before teardown) instead of CQLite `SELECT *`. The DuckDB consumer merge over delta-exported per-generation Parquet files reproduces Cassandra's merged state exactly (23 rows / 92 cells), which is the stronger validation.

**Scope:** Investigate whether the query engine should merge all generations in a table dir (LWW + tombstone suppression across generations), matching Cassandra read semantics. Not required for epic #696 (delta-scan is reader-layer, consumer-side merge is downstream), but a real read-path fidelity gap.

pmcfadin:
Resolved in #884 (merged to `main` as `7a85df4d`).

**Fix:** `SSTableManager::scan` now reconciles across SSTable generations instead of concatenating their live rows. When more than one generation serves a table, it drives the same authoritative `KWayMerger` the compaction path uses — per-cell last-write-wins by write timestamp, row + cell tombstone shadowing across generations, fully-deleted rows dropped. So `SELECT *` over a multi-generation directory now reproduces the fully-merged authoritative table state, matching Cassandra read semantics (the contract the delta-scan design doc described as the query engine's job).

**Scope / safety:**
- Single-generation reads are untouched — no merger invoked, zero behavior/perf change on the common path.
- The merge path requires a schema (cells carry no column names on disk) and the `write-support` feature. If either is absent it falls back to the prior concatenation, so reads never fail on an unsupported format.

[...]

PR #3169, opened 2026-07-30
fix(3154): only a merger-ineligible input earns the non-reconciling concat fallback

### LINKED ISSUE #3154: a transient KWayMerger::new error degrades to the non-reconciling concat: query returns full-length UNRECONCILED rows under Ok
Found during the #3124 review (PR #3153). **Pre-existing on `main` — not a regression from #3124**, and deliberately not fixed there to preserve one-issue-one-change.

## The defect

`cqlite-core/src/storage/sstable/generation_merge.rs:685` builds `MergeStreamSetupError::Construction(e)` from **any** error returned by `KWayMerger::new`. The consumer (`storage/sstable/mod.rs:2176`/`:2185`) treats `Construction` as fallback-eligible and degrades to `spawn_fanout_merge` — the **non-reconciling token-order concat**.

So a transient I/O error, or a corrupt input encountered while constructing the merger, causes the query to return a **full-length UNRECONCILED result set** under `Ok`: duplicate overwritten rows and resurrected deleted rows, behind a `tracing::warn!`.

## Why this matters

This is the same silent-wrong-data class as the blocker fixed in #3124 (a dead producer falling back to the concat), but reachable by a **more likely trigger**: #3124's route needed a panic in `KWayMerger::new`, whereas this one needs only an I/O hiccup or a bad file. Silent truncation returns too few rows; this returns the wrong rows and reports success.

#3124 made fallback eligibility a type-level decision (`fallback_eligible()`, true for `Construction` alone), which closed the producer-death route and made `Construction` the **single load-bearing fallback label**. That is what turns this from a diffuse concern into a precisely-scoped one.

## Fix shape

Narrow what earns the concat fallback. The fallback exists so that inputs the reconciling merger cannot handle still return *something* — that justification covers an **unsupported-format / merger-ineligible** condition, and nothing else. Split `Construction` accordingly:

- a genuinely merger-ineligible input ⇒ fallback-eligible (current behaviour, kept);
- an I/O error, a corruption/parse error, or any other runtime failure ⇒ **NOT** fallback-eligible; propagate as `Err`.

Reuse the existing `MergeStreamSetupError` predicate rather than adding a second mechanism, and keep the decision type-level (no string matching on error messages).

## Acceptance criteria

1. `KWayMerger::new` returning an I/O error, and returning a corruption/parse error, each make `SSTableManager::scan_stream` return `Err` — never a concat fallback, never `Ok` with unreconciled rows.
2. A genuinely merger-ineligible input still degrades to the concat exactly as today (assert this explicitly — over-restricting is its own regression, since it turns previously-working queries into failures).
3. A test per arm using an **overlapping** multi-generation fixture, so reconciled and unreconciled output are distinguishable **by row count and by value** — a disjoint fixture cannot tell the two apart, and an `is_err()`-only assertion cannot either. **Control-arm-first**: assert the reconciled expected rows on a good run before injecting the failure.
4. The fallback decision remains a single type-level predicate; no error-message matching.

## Notes

- Sibling issues in this audit set: #3106 (single-source path, merged via #3123), #3124 (multi-generation scan path, PR #3153), #3120 (k-way merge adapter / compaction output), #3149 (delta-scan path).
- Also noted during the same review, out of scope here: two bare-receiver surfaces remain one layer up on the query-result layer — `query/select_executor/mod.rs:611` and `:737`, both via `query/result.rs:317` `QueryResultIterator`.

Routing: oracle-driven bug fix — issue + pinned tests, no OpenSpec change.

pmcfadin:
Closes #3154.

> **Note:** an earlier revision of this body claimed the fallback-eligible set is "`UnsupportedFormat | UnsupportedVersion` — the only errors that are evidence about the input". That was **wrong, and the inverse of the truth**; `rust-reviewer` caught it. The body below is corrected, and [the correction is recorded in a comment](https://github.com/pmcfadin/cqlite/pull/3169#issuecomment-5136883657) rather than silently edited away.

## The defect

`generation_merge.rs` built `MergeStreamSetupError::Construction(e)` from **any** error out of
`KWayMerger::new`, and the consumer in `storage/sstable/mod.rs` treats `Construction` as
fallback-eligible — degrading to `spawn_fanout_merge`, the **non-reconciling token-order concat**.

So an I/O error, a corrupt input, or a schema-validation failure made the query return a full-length
**UNRECONCILED** result set under `Ok` — duplicate overwritten rows and resurrected deleted rows —
behind only a `tracing::warn!`. Pre-existing on `main`; found during the #3124 review (PR #3153).

## The fix — a type-level split, one predicate

`Construction` splits into:

- **`MergerIneligible`** — fallback-**eligible**, for errors that are *evidence about the input*
  (`Error::UnsupportedFormat` | `Error::UnsupportedVersion`).
- **`ConstructionFailed`** — anything else: **propagates verbatim**, so `is_recoverable()` stays honest
  for the Python/Node bindings (`Io`/`Storage` → true, `Corruption`/`Schema` → false).

`from_construction_failure` matches on the `Error` variant with a **fail-closed catch-all** (a future
variant cannot silently become fallback-eligible; the arm points at the enumerated test table for a
future author). `fallback_eligible()` remains the **single** predicate. No error-message string
matching anywhere (AC4).

## What actually changes behaviour — read this before the AC table

`KWayMerger::new` **cannot return `UnsupportedFormat` or `UnsupportedVersion`.** Its only fallible
steps are `Error::InvalidInput` on empty inputs (`merge/mod.rs:1718`, unreachable under
`readers.len() > 1`), `Error::Schema` from `validate_dropped_columns` (`:1728` → `schema/mod.rs:611`),
and `Error::Storage("streaming producer: failed to spawn thread")` from
`SSTableRowIteratorAdapter::open` (`:1746` → `producer_iter.rs:275`). That `open` **opens nothing** —
`SSTableReader::open` and every format/version gate run inside the spawned producer thread
(`producer_iter.rs:385-388`) and surface mid-stream at `merger.step()`.

Therefore:

1. **`MergerIneligible` is dead in production** — only the test-only injection seam constructs it. An
   unsupported-format generation never degraded to the concat *at this site*, before or after this
   change. The arm is retained **deliberately defensively**, so a future eager-validating constructor
   keeps the documented #883 degradation.
2. **The real behaviour change is that `Error::Schema` and `Error::Storage` now propagate** — those are
   the two errors that genuinely were reaching the concat and serving unreconciled rows.
3. **AC2 therefore holds for a narrower reason than it looks:** the concat path is preserved, and the
   AC2 test proves the *classifier*, not an end-to-end degradation.

All of that is now stated in the code (`merge_stream_setup.rs` module doc + variant docs + the
consumer comment), which is the substance of the review blocker this PR fixed.

## Evidence the tests can fail

The AC1 arms were **watched RED before the fix**: each observed **36 UNRECONCILED rows under `Ok`**
where the reconciled answer is **12**. The fixture rewrites the *same* 12 partition ids across 3
generations at ts 1000/1001/1002 with values `gen{N}-row-{id}`, so reconciled (12, all `gen2-`) and
concat (36, including older prefixes) differ by row count **and** by value. Every arm calls
`assert_reconciled_control` **first**, on a good run, before injecting the failure (AC3).

| AC | Covered by |
|----|-----------|
| 1 — I/O and corruption each propagate as `Err` | `setup_fail_closed_tests.rs` Io + Corruption arms (RED pre-fix) |
| 2 — a merger-ineligible input still degrades to the concat | `a_merger_ineligible_input_still_degrades_to_the_documented_concat` (36 rows **and** older-gen values present; green before *and* after) |
| 3 — overlapping fixture, control-arm-first | `multi_gen_fixture.rs`; `assert_reconciled_control` in every arm |
| 4 — single type-level predicate, no string matching | `merge_stream_setup.rs` unit table (9 variants, incl. the fail-closed default) |

## Two claims this change invalidated, also fixed here

- **`scan_stream` and `scan` are no longer "definitionally in lockstep"** — for a schema whose
  `dropped_columns` fails validation, `scan_stream` now returns `Err` while `scan` still returns
  concatenated rows. The comment asserting lockstep is corrected and points at #3170.
- **The fallback `tracing::warn!` never told the operator the rows were unreconciled.** It now names
  the non-reconciling concatenation and its consequence (duplicated overwritten rows, resurrected
  deleted rows, #883).

## Review

- **roborev: PASS, findings NONE** — non-vacuous, re-run after the blocker fix
  (`reviewed-sha: f8f7adf..396a649`, 9/9 census paths in the prompt, both vacuity tiers PASS).
- **rust-reviewer: 1 blocker, now fixed** (the claim layer above). It independently verified as clean:
  the fail-closed direction, the honest `cfg` gating (no `#[allow(dead_code)]`; a minimal build compiles
  none of the seam), the `is_recoverable` honesty per variant, and the non-vacuity of every test arm.

## Files

| File | Change |
|------|--------|
| `generation_merge/merge_stream_setup.rs` | 135 → 356 (the split, the predicate, the unit table, the corrected enumeration) |
| `generation_merge.rs` | 707 → 732 |
| `generation_merge/multi_gen_fixture.rs` | new, +349 (overlapping fixture, extracted from the #3124 suite, which shrank 560 → 284) |
| `generation_merge/setup_fail_closed_tests.rs` | new, +177 |
| `producer_fault/construction.rs` | new, +225 (scoped `Err`-injection registry, `cfg`-gated to `write-support ∧ ¬tombstones`) |
| `producer_fault.rs` | 758 → 749 (silencing split to `producer_fault/silence.rs`) |
| `sstable/mod.rs` | **line-neutral** (3138 → 3138, comments only — the file is over the size threshold) |

## Out of scope — follow-up #3170

The materializing `scan` path swallows *any* error from `merge_generations_for_read` — including
mid-collection `step()` failures — on all four sites (`sstable/mod.rs:1134`, `:1482`, `:1650`,
`:1847`), and #3154's predicate is **bypassed** there entirely. A `tombstones` build routes
`scan_stream` through `scan`, so that configuration retains this defect until #3170 lands.

### DIFF
--- a/cqlite-core/src/storage/producer_fault.rs
+++ b/cqlite-core/src/storage/producer_fault.rs
@@ -97,6 +97,40 @@
 //! * `cqlite-flight` enables the feature from its `[dev-dependencies]` (the same
 //!   convention `arrow-shape-corpus` / `test-util` already use), so the shipped
 //!   Flight binary never links it.
+//!
+//! # One seam injects an `Err`, not a panic (issue #3154)
+//!
+//! [`arm_merge_construction_error`] (child module [`construction`]) makes
+//! `KWayMerger::new` REPORT a chosen error variant on the cross-generation merge
+//! path, which is what proves the narrowed fallback classification: an I/O or
+//! corruption failure must propagate, while a merger-ineligible unsupported-format
+//! failure must still degrade to the documented concat. See that module's doc.
+
+/// The `Err`-reporting construction seam (issue #3154), in a child module so this
+/// file stays under the ~800-line campsite target (epic #1116).
+///
+/// Gated to exactly where the seam it injures EXISTS —
+/// `generation_merge::stream_generations_for_read` is `write-support` AND
+/// `not(tombstones)` (a `tombstones` build routes `scan_stream` through the
+/// materializing `scan`, which has no `MergeStreamSetupError` at all) — for the same
+/// reason [`ScanTaskSite::CrossGenerationMerge`] is: a symbol that nothing in a
+/// configuration can ever reach would be a lie about that configuration's coverage,
+/// and `#[allow(dead_code)]` would hide it instead of stating it.
+#[cfg(all(
+    any(test, feature = "producer-fault-injection"),
+    feature = "write-support",
+    not(feature = "tombstones")
+))]
+mod construction;
+#[cfg(all(
+    any(test, feature = "producer-fault-injection"),
+    feature = "write-support",
+    not(feature = "tombstones")
+))]
+pub use construction::{
+    arm_merge_construction_error, ArmedMergeConstructionError, MergeConstructionFault,
+    INJECTED_CONSTRUCTION_MESSAGE,
+};
 
 /// Producer-fault state captured ONCE when a query row stream is opened, then
 /// owned by that stream's producer thread.
@@ -353,6 +387,39 @@ impl FaultScope {
     }
 }
 
+/// The `Err`-reporting construction seam's take side (issue #3154).
+///
+/// Gated to the one configuration whose call site exists — the cross-generation
+/// reconciling merge's construction window in
+/// `generation_merge::stream_generations_for_read` — so no build carries a method
+/// nothing can reach (see [`construction`]'s `mod` declaration above).
+#[cfg(all(feature = "write-support", not(feature = "tombstones")))]
+impl FaultScope {
+    /// The construction error a test armed for this scope, if any — i.e. the error
+    /// `KWayMerger::new` is made to report instead of building the merger.
+    ///
+    /// ALWAYS `None` in a production build, where [`Self::armed_construction_error`]
+    /// is a function that returns `None` without touching (or even compiling) any
+    /// registry.
+    #[inline]
+    pub(crate) fn injected_construction_error(&self) -> Option<crate::Error> {
+        self.armed_construction_error()
+    }
+
+    #[cfg(any(test, feature = "producer-fault-injection"))]
+    #[inline]
+    fn armed_construction_error(&self) -> Option<crate::Error> {
+        construction::take(&self.path.to_string_lossy())
+    }
+
+    /// Production build: nothing can be armed, so there is nothing to take.
+    #[cfg(not(any(test, feature = "producer-fault-injection")))]
+    #[inline]
+    fn armed_construction_error(&self) -> Option<crate::Error> {
+        None
+    }
+}
+
 /// The panic message both checkpoints raise. Exported so a test can (a) assert
 /// the forwarded error carries it and (b) suppress exactly this panic in its
 /// panic hook without silencing a real one.
@@ -411,7 +478,10 @@ mod armed {
     /// own registration.
     static NEXT_ARM_ID: AtomicU64 = AtomicU64::new(0);
 
-    fn next_id() -> u64 {
+    /// `pub(super)` so the sibling [`construction`](super::construction) registry
+    /// shares ONE id space with these three, rather than minting a second counter
+    /// whose ids could collide with theirs in a debugger or a log.
+    pub(super) fn next_id() -> u64 {
         NEXT_ARM_ID.fetch_add(1, Ordering::SeqCst)
     }
 
@@ -426,8 +496,10 @@ mod armed {
     /// order of magnitude.
     const MIN_SCOPE_LEN: usize = 8;
 
-    /// Fail LOUDLY on a scope that would match everything.
-    fn check_scope(scope: &str) {
+    /// Fail LOUDLY on a scope that would match everything. `pub(super)` so every
+    /// registry — including the sibling [`construction`](super::construction) one —
+    /// enforces the SAME minimum rather than re-deriving it.
+    pub(super) fn check_scope(scope: &str) {
         debug_assert!(
             scope.len() >= MIN_SCOPE_LEN,
             "producer-fault scope {scope:?} is shorter than {MIN_SCOPE_LEN} chars: a \
--- a/cqlite-core/src/storage/producer_fault.rs
+++ b/cqlite-core/src/storage/producer_fault.rs
@@ -100,22 +100,20 @@
 //!
 //! # One seam injects an `Err`, not a panic (issue #3154)
 //!
-//! [`arm_merge_construction_error`] (child module [`construction`]) makes
+//! [`arm_merge_construction_error`] (child module `construction`) makes
 //! `KWayMerger::new` REPORT a chosen error variant on the cross-generation merge
-//! path, which is what proves the narrowed fallback classification: an I/O or
-//! corruption failure must propagate, while a merger-ineligible unsupported-format
-//! failure must still degrade to the documented concat. See that module's doc.
+//! path, proving the narrowed fallback classification: an I/O or corruption failure
+//! must propagate, while a merger-ineligible unsupported-format failure must still
+//! degrade to the documented concat. See that module's doc.
 
 /// The `Err`-reporting construction seam (issue #3154), in a child module so this
 /// file stays under the ~800-line campsite target (epic #1116).
 ///
 /// Gated to exactly where the seam it injures EXISTS —
 /// `generation_merge::stream_generations_for_read` is `write-support` AND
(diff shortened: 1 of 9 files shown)


pmcfadin:
Out-of-scope follow-up filed as **#3170** (P1) — the materializing \`scan\` path.

Worth recording what that audit established, because it sharpens what *this* PR does and does not close:

- **#3154's predicate does not reach the materializing path — it is bypassed.** \`MergeStreamSetupError\` is returned by exactly one function (\`stream_generations_for_read\`) and consulted at exactly one call site (\`scan_stream\`). The materializing \`merge_generations_for_read\` returns a flat \`Result<Vec<_>>\`, so \`fallback_eligible()\` is never reached from \`scan\`.
- That flat \`Result\` collapses **three** classes into one \`Error\`: merger construction, **mid-collection \`merger.step()\`**, and the \`spawn_blocking\` \`JoinError\`. The \`step()\` case is why #3170's trigger is broader than this one — it can fire *after* partitions are already reconciled.
- Three same-shape materializing siblings were also recorded there: \`sstable/mod.rs:1482\`, \`:1650\`, \`:1847\`.
- The two query-layer bare-receiver surfaces (\`select_executor/mod.rs:611\`, \`:737\` via \`query/result.rs:317\`) remain unfixed and are noted as related-but-separate — that is the *truncation* variant, not this *wrong-rows* variant.

So this PR closes the streaming setup route specifically, and #3170 carries the materializing route.

pmcfadin:
## Correction to this PR's own claim (rust-reviewer BLOCKER)

I wrote in the PR body that the fallback-eligible set is `UnsupportedFormat | UnsupportedVersion` — "the only errors that are evidence about the input". **That claim is wrong, and it is the inverse of the truth.** Recording it here rather than quietly editing the body.

`KWayMerger::new` **cannot return either of those.** Its only fallible steps are `Error::InvalidInput` (empty `input_paths`), `Error::Schema` (`validate_dropped_columns`), and `Error::Storage` (producer thread-spawn failure). `SSTableRowIteratorAdapter::open` opens nothing — it spawns a thread, and `SSTableReader::open` (with every format/version gate) runs *inside* that thread, surfacing as a `MergeMsg::Failed` at `merger.step()`. `readers.len() > 1` at the consumer also implies every generation already opened successfully, making an `UnsupportedFormat` at construction doubly impossible.

So the accurate description of this PR is:

- **`MergerIneligible` is dead in production.** Only the test-only injection seam constructs it. An unsupported-format generation never degraded to the concat *at this site* — before or after this change.
- **The real behavior change** is that `Error::Schema` and `Error::Storage` — the two errors that genuinely were reaching the concat — now propagate as `Err` instead of serving unreconciled rows. That is the fix, and it was undocumented.
- **AC2 holds for a narrower reason than it appears to:** the concat path is preserved, but the AC2 test proves the *classifier*, not an end-to-end degradation, because the eligible arm was already unreachable. The arm is retained deliberately, so a future eager-validating constructor keeps the #883 degradation.

A docs-truth commit is in flight to say all of this in `merge_stream_setup.rs` and at the consumer site, plus two claims my change invalidated: the "definitionally in lockstep" comment (`scan_stream` vs `scan`) is now false for the dropped-column case, and the fallback `tracing::warn!` never told the operator the rows are unreconciled.

The logic, the fail-closed direction, the cfg gating, the error-variant honesty, and the non-vacuity of the tests were all verified clean — the blocker was entirely in the claim layer.

pmcfadin:
Batched nit follow-up filed: **#3178** (P2, `testing`/`storage`) — the injection-free e2e pin the C audit recommended.

That audit verified something worth stating here: **no injection-free dropped-column test exists anywhere in the repo.** The only test that drives a `validate_dropped_columns` failure through a public surface is `issue_847_dropped_column_filter.rs::dropped_column_absent_from_columns_is_rejected`, and it goes through **`compact_sstables`**, not `scan_stream` — different surface, fixture and oracle. All five tests that mention both `scan_stream` and `dropped_columns` set `dropped_columns: HashMap::new()`, so none can trigger it. So the gap is real, and #3178 is correctly a new test rather than an extension.

Also recorded on #3178, not fixed here: this PR's enumeration cites `validate_dropped_columns` at `schema/mod.rs:611`; the real definition is **`:608`** (`:611` is inside the error body). Left alone deliberately — the gate of record is running against `396a649` and a source edit would invalidate it for a one-line citation.

Issue #3168, opened 2026-07-30
Multi-generation read path never injects static cells: the N-gen arm emits an extra static-only row AND strips the static column from every clustering row (generation_merge has zero static handling)


pmcfadin:
## Summary

**The multi-generation read path performs no static-column injection at all.** `generation_merge.rs`
— the cross-generation emission kernel for `scan`, `scan_stream`, the point read and the
`WRITETIME`/`TTL` projection — contains **zero** static handling (literally no `static` token in the
file). The consequences, both visible in one `SELECT`, are:

1. the partition's static row is emitted as **its own standalone result row** (an extra all-null-except-statics row), and
2. every real clustering row comes back **without the static column**.

The single-generation path does the right thing (`build_display_row_read_path` → `merge_static_cells`),
so the same bytes give **two different result sets depending on the table's generation count** — the
answer to a `SELECT` changes when compaction runs.

**Found by**: the 1-gen-vs-N-gen differential axis added for **#3129**
(`cqlite-core/tests/point_vs_full_differential/one_vs_n_generation.rs`, branch
`issue-3129-one-vs-n-gen-differential`, head `1b49705`), run against unmodified `origin/main`. It is
quarantined there as a `known_divergent` case so it does not permanently red the enforcing lane; this
issue releases that quarantine.

Adjacent but **distinct** from #3121 (row-DELETED row resurfacing because `merge_static_cells` ran
*before* `build_display_row`'s tombstone-MARKER rule). #3121 was about the *ordering* of static
injection on the single-gen path; this is about static injection being **entirely absent** on the
multi-gen path.

## Repro

```bash
bash test-data/scripts/fetch-datasets.sh
export CQLITE_DATASETS_ROOT=$PWD/test-data/datasets
cargo test -p cqlite-core --features cli-helpers --test point_vs_full_differential \
  -- --ignored one_vs_n_generation_known_divergences --nocapture
```

* **Case**: `test_tomb.static_with_tombstones` (schema `test-data/schemas/tombstone-parity.cql:194`),
  divergence classes `["static", "tombstone"]`.
* **Fixture** (single Cassandra-written generation, single partition `pk=1`): a live static cell
  `stat_col` written at `T_GEN1`, clustering rows `ck=1..6`, then a **row tombstone** at `ck=2`, a
  **cell tombstone** on `row_col` at `ck=3`, and a **range tombstone** over `ck>=4 AND ck<=5` — all at
[...]

Issue #3164, opened 2026-07-30
Widen the static_with_tombstones__deleted_cell_null oracle case to the full pk=1 slice (retire the deliberate ck>=3 bound) — blocked on PR #3122


pmcfadin:
## Summary

Retire the deliberate slice bound on the **`static_with_tombstones__deleted_cell_null`** query-semantics
oracle case: widen its query from `WHERE pk = 1 AND ck >= 3` to the full `WHERE pk = 1` slice.

Split from **#3121 AC2** (closed as substantially delivered by PR #3159). This is the last unretired
piece of that acceptance criterion, and it is **currently unactionable** — see Blocked-by.

## Background

#3094 added the oracle case and **deliberately bounded its slice to `ck >= 3`** so the case would own
the cell-tombstone→NULL property only, and would **never encode the #3121 phantom-row divergence as
expected behaviour**. That was the correct call at the time: `ck=2` (a pure row deletion carrying zero
cells) was wrongly being returned as an all-null row, and a full-slice case would have baked the bug
into the oracle.

The phantom row is now fixed (PR #3141, #3095) and pinned at all three gated decode sites (PR #3159,
#3121). So the bound is now pure under-coverage: the case verifies less than it could, and the reason it
was narrowed no longer exists.

## Blocked by

**PR #3122 (#3094) must merge first.** The `static_with_tombstones__deleted_cell_null` case **does not
exist on `main`** — it lands with that PR. As of 2026-07-30 PR #3122 is green (`required` SUCCESS) but
`CONFLICTING`/`DIRTY` and idle; it needs a rebase onto current `main`. There is nothing to edit until it
lands, so do not pick this up before then.

Verify with: `git grep -c 'static_with_tombstones__deleted_cell_null' origin/main -- test-data/query-semantics-oracle.json`
(expect `0` today).

## Expected result set after widening

Fixture `test_tomb.static_with_tombstones` (real Apache Cassandra 5.0 `nb`, single flush), `pk = 1`, at a
**PINNED** `now`:

| `ck` | expected |
|---|---|
| 1 | live, `row_col = row_1`, `stat_col = surviving_static` |
| 2 | **absent** — pure row deletion, zero cells; Cassandra 5.0.8 never yields it |
| 3 | present, `row_col` **NULL** (its only regular cell is a cell tombstone, per #3094) |
| 4, 5 | absent — covered by a range-tombstone bound pair |
[...]

Issue #3129, opened 2026-07-30
Multi-generation merge resurrects a partition-deleted row as a phantom all-null row: filter_live drops cell tombstones before the shadow-evidence fold (single-gen vs multi-gen now disagree on the same bytes)


pmcfadin:
## Summary

**A row inside a DELETED partition reappears as an all-null phantom row as soon as the table has
more than one SSTable generation.** The same bytes read at 1 generation correctly return 0 rows;
read through the multi-generation merge they return 1 row (clustering key populated, every regular
column null). Cassandra 5.0 returns 0 rows.

The affected shape is a clustering row whose only non-primary-key content is a **cell tombstone**
(the `UPDATE t SET w = null …` / marker-less shape), sitting in a partition covered by a partition
deletion. #3094 (PR #3122) fixed exactly this shape on the **single-generation** path by giving the
row-level shadow decision a timestamp-less *presence* bit. The **multi-generation** path never got
the equivalent, so the two read paths now **disagree on identical bytes** — the answer to a `SELECT`
depends on the table's compaction state.

Two halves, both first-class here:

1. the multi-gen read-visibility defect (below), and
2. a **structural gap in the point-vs-full differential lane**: it has no 1-gen-vs-N-gen axis, so it
   is *incapable* of seeing a single-gen/multi-gen disagreement at all.

**Found during**: **#3094 (Flight cell-tombstone → NULL on the read path), PR #3122**.

## Repro

Table with at least one clustering column (see "Scope of the trigger" — a partition-key-only table
fails differently; that is #2799). Two generations are required: with a single generation the
`KWayMerger` is not engaged and #3094's fix returns the correct 0 rows.

1. `bash test-data/scripts/fetch-datasets.sh`; `export CQLITE_DATASETS_ROOT=$PWD/test-data/datasets`.
2. Generation 1 — a **marker-less** row whose only cell is a tombstone, written *after* the deletion
   it must still be hidden by:
   `DELETE w FROM t WHERE pk = 1 AND ck = 2 USING TIMESTAMP 6000` → flush.
3. Generation 2 — the partition deletion:
   `DELETE FROM t WHERE pk = 1 USING TIMESTAMP 5000` → flush. (Two separate flushes are load-bearing:
   flush-time reconciliation inside one memtable purges the covered row and makes the assertion
   vacuous — the same fixture constraint #3094's pin records.)
4. `SELECT * FROM t WHERE pk = 1` through the CQLite read path with the authoritative schema, at a
[...]

PR #3171, opened 2026-07-30
test(#3129): add the 1-gen-vs-N-gen differential axis (AC5/AC6)

pmcfadin:
**Refs #3129** (deliberately NOT `Closes` — this delivers only AC5/AC6, so the issue must stay OPEN for AC1/AC2, which land with #3122, and AC9 which is #3164). See **Scope** below: this PR delivers acceptance criteria **5 and 6** only.

## Merge: on green (HOLD released by the owner, 2026-07-30)

An earlier revision of this PR carried a merge-hold directive naming #3122. **The owner released it on 2026-07-30.** The HOLD's only
real purpose was preventing this partial delivery from closing #3129, and that is handled properly by using
`Refs` instead of `Closes` — so blocking a green, self-contained, test-only PR behind a separate lane bought nothing and cost a rebase + a 15-25 min re-gate every time `main` moved. **#3129 stays OPEN** after this
merges, for AC1/AC2 (PR #3122) and AC9 (#3164).

## Scope — why this is only half of #3129

#3129 has two halves, and the **fix half was already implemented in the in-flight PR #3122** before this
work started. Verified per-criterion against branch `issue-3094-cell-tombstone-null`:

| AC | Owner | Status |
|----|-------|--------|
| 1 (0 rows multi-gen at pinned `now`) | PR #3122 | covered there |
| 2 (all four multi-gen drivers) | PR #3122 | partial there (kernel fix covers all four; only one is pinned) |
| 3 (presence bit, not the timestamp) | PR #3122 | covered there |
| 4 (corrected scope note) | PR #3122 | covered there |
| **5 (1-gen-vs-N-gen differential axis)** | **THIS PR** | **delivered** |
| **6 (axis verified by seeded divergence)** | **THIS PR** | **delivered** |
| 7 (Cassandra oracle) | PR #3122 | covered there |
| 8 (anti-vacuity) | PR #3122 | covered there |
| 9 (query-semantics oracle case ≥2 gen) | **#3164** | separate issue, also blocked on #3122 |

`has_deleted_data_cell` / `has_shadow_evidence` **do not exist on `main`** — #3129's root-cause section
describes the #3122 branch, not `main`. Deliberately **no `src` change here**: duplicating #3122's fix
would collide with it. `git diff --stat origin/main...HEAD -- cqlite-core/src` is **EMPTY**.

## What this adds

A **1-gen-vs-N-gen differential axis** for the #1918 point-vs-full differential lane
(`cqlite-core/tests/point_vs_full_differential/one_vs_n_generation.rs`, wired in as a `#[path]`
submodule). It materializes two temp trees from the SAME Cassandra-written single-generation fixture —
one at 1 generation, one with that generation copied N≥2 times under distinct generation numbers — and
requires **identical rows, values, and order** for the full scan, every per-partition read under BOTH
forced `CQLITE_READ_PATH` modes, and the multi-key `IN`, at a **pinned `now`**
(`CQLITE_TTL_NOW_OVERRIDE_SECS`, never wall-clock).

This closes the structural gap #3129 names as first-class: the existing lane's only axis is point-vs-full
**at a fixed generation count**, so both arms route through the same `KWayMerger` and a
single-gen/multi-gen disagreement is invisible to it *by construction*.

**It is a real multi-generation read** (verified against `main`): every multi-gen dispatch is
`readers.len() > 1` + schema present (`storage/sstable/mod.rs:1115/1461/1629/1834/2136`), with no
content/digest dedup and no single-generation short-circuit, so N byte-identical copies genuinely route
through `merge_generations_for_read` / `..._with_metadata` / `stream_generations_for_read` /
`seek_merge_generations_for_read`. **Documented limitation**: because the copies are byte-identical every
cross-generation comparison is a *tie*, so the axis exercises structural merge (interleaving, static
injection, tombstone application, order, dedup) but **not** last-write-wins precedence.

## AC6 — divergence detection verified two independent ways

1. **Code-seeded**: temporarily made `partition_live_rows`' `RowData::Tombstone` arm emit a phantom row
   (multi-gen kernel only). The enforcing lane FAILED, naming the case and query and printing per-row
   `1-gen: <absent>` / `2-gen: {...}`. Reverted; `src` verified byte-identical to `origin/main`.
2. **Unseeded, on unmodified `main`** — stronger than the criterion asks: the axis found **three genuine
   divergences** with nothing seeded at all. Now tracked and quarantined (below).

## Defects this lane found on `main` (not fixed here — out of scope)

* **#2189** (raised to P1) — multi-gen kernel resurrects **TTL-expired** rows as all-null phantom rows:
  `test_tomb.gc_before_boundary` 1 row → 3, `test_compaction_tombstone_ttl.ttl_expired_live` 1 → 2. Root
  cause is the missing row-liveness rule in that kernel, **not** #3129's cover-based defect. Note PR
  #3122 will **not** fix these: `merged_row_shadowed_by_partition` short-circuits on `cover = None`
  (`partition_shadow.rs:270-272`) and neither fixture has a partition deletion.
* **#3168** (P2) — multi-gen read path never injects static cells: the N-gen arm emits an extra
  static-only row AND strips the static column from every clustering row.

Both are quarantined as `known_divergent` and each cites its issue (enforced — see below).

## The quarantine is self-releasing

`one_vs_n_generation_quarantine_still_diverges` is **NOT `#[ignore]`d** — it asserts each quarantined case
**still diverges**, and FAILS with a flip-to-`None` instruction when one stops. A fixed divergence
therefore trips a test rather than sitting unnoticed, and `assert_corpus_invariants()` requires every
`known_divergent` reason to cite an issue (`#\d+`). Both the release-signal and harness-error branches were
seed-verified.

## Review

`rust-reviewer` (3 blockers) + roborev (2 findings; genuine review, `reviewed-sha`
`f8f7adf..1b49705`, census `4 files, +931/-10`) — consolidated to 4 blockers, all fixed:

1. **Wall-clock leak into the correctness path** (found independently by BOTH reviewers): the
   `CQLITE_TTL_NOW_OVERRIDE_SECS` set/remove raced the sibling `#[tokio::test]` in the same binary, which
   could unpin a run mid-flight → spurious divergence, or **masking a real one**. Fixed with a
   `#[must_use] pin_read_clock()` RAII guard that restores the *previous* value (including "was unset",
   and on panic) plus `#[serial]` on all three env-writing tests; removed both unconditional
   `remove_var`s and the now-false "single-threaded here" comment.
2. **Quarantine had no release signal** (`#[ignore]`d ⇒ green whether broken or fixed) → the
   non-ignored self-releasing pin above.
3. **A quarantined case cited no issue** (a waiver with no cited issue is not a waiver) → reasons now cite
   #2189/#3168, enforced by `cites_an_issue`, itself unit-tested.
4. **Point mode never probed a deleted partition** (roborev, new): probe keys came only from the 1-gen
   scan, so a deleted partition yielded no key and `seek_merge_generations_for_read` was untested for
   exactly this issue's shape — an acceptance gap, since #3129's repro *is* a point query against a
   deleted partition and AC2 names that driver. Fixed with `empty_probe_keys` merged into the discovered
   keys, asserted undiscoverable AND zero-rows on both arms in both modes, plus a new enforcing case
   `test_deltas.partition_tombstones` (N=3).

## Anti-vacuity

Exact row/partition counts per case; both arms' generation counts re-scanned after materialization;
fixture-drift FAILs; verified fail-closed under `CQLITE_REQUIRE_FIXTURES=1` with an empty datasets root.
Enforcing lane is GREEN on `origin/main`. Direct target run: **6 passed, 0 failed, 0 ignored**.

## Sequencing

The axis is green on `main` over the enforcing corpus, so this PR stands alone. #3129's remaining criteria
(AC1/AC2, and #3164's AC9) land with #3122, so **#3129 stays open** after this merges — enforced by `Refs`
rather than by holding this PR.

Docs updated in the same change (website `validation-playbook`, `test-data-management` skill) for the
self-releasing-pin rule, empty-partition probes, and the tie-only limitation.

### DIFF
--- a/.claude/skills/test-data-management/validation-workflow.md
+++ b/.claude/skills/test-data-management/validation-workflow.md
@@ -133,7 +133,7 @@ Real entrypoints (these exist; there is no `sstable_validation` target):
 |------------|--------|
 | `cqlite-core/tests/query_semantics_oracle_parity.rs` | query-semantics parity — the post-reconciliation `SELECT` result set at a PINNED `now` |
 | `cqlite-core/tests/point_vs_full_differential.rs` | point-vs-full differential lane — same query under `CQLITE_READ_PATH=point` vs `=full` |
-| `cqlite-core/tests/point_vs_full_differential/one_vs_n_generation.rs` | 1-gen-vs-N-gen differential axis (#3129) — same bytes read at 1 vs N ≥ 2 generations must agree |
+| `cqlite-core/tests/point_vs_full_differential/one_vs_n_generation.rs` | 1-gen-vs-N-gen differential axis (#3129) — same bytes read at 1 vs N ≥ 2 generations must agree, incl. point reads of deleted partitions; quarantined shapes are pinned as EXPECTED divergences (self-releasing), never `#[ignore]`d |
 | `bash test-data/scripts/smoke-test-all-tables.sh` | all-tables smoke across the enumerated corpus |
 
 The shape of a per-table parity assertion (illustrative, not a copy of a real file):
--- a/.claude/skills/test-data-management/validation-workflow.md
+++ b/.claude/skills/test-data-management/validation-workflow.md
@@ -133,6 +133,7 @@ Real entrypoints (these exist; there is no `sstable_validation` target):
 |------------|--------|
 | `cqlite-core/tests/query_semantics_oracle_parity.rs` | query-semantics parity — the post-reconciliation `SELECT` result set at a PINNED `now` |
 | `cqlite-core/tests/point_vs_full_differential.rs` | point-vs-full differential lane — same query under `CQLITE_READ_PATH=point` vs `=full` |
+| `cqlite-core/tests/point_vs_full_differential/one_vs_n_generation.rs` | 1-gen-vs-N-gen differential axis (#3129) — same bytes read at 1 vs N ≥ 2 generations must agree |
 | `bash test-data/scripts/smoke-test-all-tables.sh` | all-tables smoke across the enumerated corpus |
 
 The shape of a per-table parity assertion (illustrative, not a copy of a real file):
--- a/cqlite-core/tests/point_vs_full_differential.rs
+++ b/cqlite-core/tests/point_vs_full_differential.rs
@@ -434,8 +434,7 @@ async fn run_case(case: &TableCase) -> Result<bool, String> {
     let full_db = open_db(&root, &schema, case.keyspace, ReadPathMode::Full).await?;
     let point_db = open_db(&root, &schema, case.keyspace, ReadPathMode::Point).await?;
 
-    let discovered =
-        discover_pk_ints(&full_db, case.keyspace, case.table, case.pk_column).await?;
+    let discovered = discover_pk_ints(&full_db, case.keyspace, case.table, case.pk_column).await?;
     // Merge discovered (live) keys with the always-probe keys, deduplicated and
     // sorted so the probe set is deterministic.
     let mut key_set: BTreeMap<i64, ()> = BTreeMap::new();
--- a/cqlite-core/tests/point_vs_full_differential.rs
+++ b/cqlite-core/tests/point_vs_full_differential.rs
@@ -36,7 +36,27 @@
 //! `comparison_detects_a_seeded_divergence` below (feeding the compare helper two
 //! different row sets must report a mismatch), complementing the manual
 //! seed-a-real-divergence verification recorded in the PR.
+//!
+//! ## Second axis: 1 generation vs N generations (issue #3129)
+//!
+//! `one_vs_n_generation` (submodule, same target) adds the orthogonal axis this
+//! file's point-vs-full comparison structurally CANNOT see: both of the point/full
+//! arms read the same fixture at the same generation count, so a divergence
+//! between single-generation reconciliation and the cross-generation merge kernel
+//! reproduces identically on both arms and stays green. That submodule reads the
+//! SAME bytes at 1 generation and at N ≥ 2 generations and requires identical
+//! result sets, reusing this file's corpus conventions, pinned `now`, SKIP
+//! contract and `normalize`.
 #![cfg(all(feature = "state_machine", feature = "cli-helpers"))]
+
+// `#[path]` because this file IS the integration target's crate root: a bare
+// `mod` would resolve to `tests/one_vs_n_generation.rs`, which cargo would then
+// ALSO auto-discover as its own (helper-less, non-compiling) test target. Keeping
+// the submodule under `tests/point_vs_full_differential/` — a directory without a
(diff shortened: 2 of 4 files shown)


pmcfadin:
Batched review nits from this PR are tracked in **#3172** (P3) — quarantine release precision (the pin matches a bare `DIVERGENCE_MARKER` rather than a per-case signature), the 1-gen anti-vacuity baselines being unreachable when a quarantined case diverges early, a corpus-shape guard gap, an `sstables_scanned` positive assertion, sidecar-allowlist brittleness, and a proactive file split under #1135.

Per doctrine nits are batched into one linked follow-up and do not trigger a re-verify round. None of them affect the enforcing lane's soundness: it filters to `known_divergent.is_none()` and fails on ANY error, so it cannot pass while a real 1-gen-vs-N-gen divergence exists in an enforced shape, nor fail spuriously from these.

pmcfadin:
## C intent audit (spec-auditor) — **PASS**

Scoped to **AC5 + AC6 only** (AC1/2/3/4/7/8 belong to PR #3122; AC9 to #3164; AC10 is the gate).

**AC5 `satisfied`** — wiring verified rather than assumed: no `[[test]]` stanza and no `required-features`, so `point_vs_full_differential` is auto-discovered and its crate-root `cfg(state_machine + cli-helpers)` is satisfied by the gate's `core-tests` invocation (`scripts/agent-gate.sh:6296-6303`); `core-tests` is in `DATASET_COMPONENTS` so the datasets root is exported; the `#[path]` submodule has **no `#[ignore]`** and neither test name is in the gate's skip list. Genuine multi-gen read confirmed at all four dispatches (`storage/sstable/mod.rs:1115/1461/1629/1834/2136`, no content dedup, no single-gen short-circuit), and the `not(feature="tombstones")` `seek_merge_generations_for_read` arm IS compiled under `core-tests`. The auditor independently confirmed each of the 7 corpus fixtures holds exactly ONE Cassandra-written generation, so the drift guard is live rather than dormant.

**AC6 `satisfied`** — and on stronger grounds than the reverted code-seed, which is not re-observable from the tree: `one_vs_n_generation_quarantine_still_diverges` is a **permanently-executing positive control**, requiring the axis to actually emit a `DIVERGENCE`-marked error for three real unseeded defects (#2189 ×2, #3168) on every gate run, with harness errors explicitly excluded from counting as detection. A lane that could not detect divergence would fail that pin rather than pass it.

**Blocking findings: none.** Zero `src` change confirmed.

Documented residual limitations, none defeating the criteria: (a) byte-identical copies make every cross-generation comparison a *tie*, so last-write-wins precedence is out of this axis's reach by construction; (b) a further regression *inside* an already-quarantined case would not fail, since the pin matches the marker and not the divergence shape — this is #3172; (c) the ratchet is class-level rather than case-level.

Issue #3192, opened 2026-07-31
migrate ~15 remaining two-candidate `parent()?.join("schemas")` fixture sites to the fixture_roots resolver, so the reintroduction guard can cover the whole rule


pmcfadin:
Follow-up from **#3148** / **#3131** (PR on branch `issue-3131-schemas-root-contract`). Filed at review time so the parent PR does not widen; this is the deliberate out-of-scope residue recorded in that change's `proposal.md` non-goals.

## What remains

~15 `cqlite-core/tests/**` files resolve schema fixtures with a **two-candidate idiom**: `datasets_root.parent()?.join("schemas")` first, then a `CARGO_MANIFEST_DIR`-anchored `../test-data/schemas` fallback. Examples:

- `cqlite-core/tests/issue_1143_windowed_scan_straddle_parity.rs:71,77`
- `cqlite-core/tests/issue_693_writetime_threading.rs:42,48`
- `cqlite-core/tests/issue_1562_perf_gate_access_path.rs:50,56`
- `cqlite-core/tests/issue_1650_bti_child_lookup.rs:295,301`

## Why it was left out of #3148

Unlike the four sites #3148 fixed, these do **not** hard-fail when the corpus root has no `schemas` sibling — the `CARGO_MANIFEST_DIR` fallback saves them. So this is **duplication plus a latent wrong-root hazard**, not an outage: the first candidate still wins when a `schemas` sibling *happens* to exist next to a relocated corpus, and that sibling may not be the checkout's. Migrating 15 more files would have widened a two-issue PR past its scope.

## Why it is still worth doing

1. It is the same defect class #3148 exists to eliminate — locating **committed source** by climbing `..` from an env var that points at **fetched, relocatable** data. Fifteen copies of the retired idiom keep the class alive.
2. The reintroduction guard in `scripts/tests/test_agent_gate_schemas_preflight.sh` currently greps only the literal `join("../schemas")`. It **cannot** be widened to cover the `parent()`-based form while ~15 legitimate uses remain, so the guard stays narrower than the rule it enforces until these are migrated.
3. The wrong-root path is silent. A test reading the wrong `schemas` tree either passes on the wrong fixture or fails with a confusing message — the attribution problem #3148 was filed for.

## Acceptance criteria

[...]

Issue #3041, opened 2026-07-27
doctrine: two agent instructions are actively WRONG — the closer's gate-completion probe matches the INCOMPLETE placeholder, and the Cassandra format-authority path does not exist


pmcfadin:
Two agent-facing instructions are not *missing* — they are **actively WRONG**. Both can silently
produce a wrong verdict from an agent that follows them correctly, which is why this is P1: these are
**CORRECTIONS to incorrect instructions**, not additions of new guidance.

## Routing — process/doc work, SKIPS OpenSpec

This is an **instruction-surface correction**, not an SSTable-format oracle bug and not a
design-exploration change. There is no Cassandra oracle and there is nothing to design: the correct
strings are already known and stated below. **Do NOT run `flow-activate` / open an OpenSpec change** —
go straight to `flow-implement`.

---

## Defect (a) — the closer's gate-completion probe matches the START-OF-RUN placeholder

`.claude/agents/flow-closer.md` around line 54 documents the full-gate completion probe as:

```bash
grep -q 'RESULT:' /tmp/gate-<N>.txt && echo done   # summary present ⇒ gate finished
```

`scripts/agent-gate.sh` writes

```
RESULT: INCOMPLETE (gate did not finish)
```

into `$AGENT_GATE_SUMMARY_FILE` **at launch**, via the EXIT trap, and only *overwrites* it with
`RESULT: PASS` / `RESULT: FAIL` on completion. The placeholder is deliberate and good (it makes a
killed/orphaned gate detectable). But it means `grep -q 'RESULT:'` is satisfied **the instant the gate
starts**, before a single component has run.

Consequence: a closer that follows its own documented predicate can treat a **just-launched** gate as
a finished one, read `INCOMPLETE` as its gate of record, and proceed toward merge on a verdict that
does not exist. The full gate is *the only run that counts*, so accepting a placeholder as its verdict
silently voids the entire certification step — and nothing downstream re-checks it.

**FIX**: the completion probe must be

```bash
grep -qE 'RESULT: (PASS|FAIL)' "$AGENT_GATE_SUMMARY_FILE"
```

Only `PASS`/`FAIL` is a verdict; `INCOMPLETE` means "still running, or died".

**Sweep ALL agent/skill/doc surfaces for the same weak pattern**, not just this one line:
`.claude/agents/**`, `.claude/skills/**`, `docs/development/gate-ops.md`,
[...]

PR #3194, opened 2026-07-31
fix(3131,3148): resolve committed schema fixtures checkout-relative, and fail the gate closed when they are unreachable

### LINKED ISSUE #3131: no single CQLITE_DATASETS_ROOT works on the fleet: /data/datasets has the corpus but no ../schemas sibling; repo test-data/datasets has the sibling but no corpus
Found while running the #3106 full gate of record (PR #3123). Cost one full-gate cycle, and a second partial one.

## The defect

Test fixtures resolve CQL schemas at **`$CQLITE_DATASETS_ROOT/../schemas/*.cql`**. On this fleet neither candidate root satisfies both requirements:

| Candidate root | Corpus (`Data.db`) | `../schemas` sibling |
|---|---|---|
| `/data/datasets` (the machine's configured root, where `fetch-datasets.sh` caches + extracts) | ✅ ~144-155 | ❌ `/data` has no `schemas/` |
| `<repo>/test-data/datasets` (what CLAUDE.md documents) | ❌ only 30 committed byte-parity refs; `test_basic` has **0** | ✅ |

Result: pointing at `/data/datasets` clears `preflight` and runs all 32 components, then fails 7 tests across **`core-tests`** and **`memory-budget`**, every one panicking with the same shape:

```
Path does not exist: /data/datasets/../schemas/basic-types.cql
```

Pointing at the repo path instead fails earlier and more honestly — `preflight: FAIL (canonical corpus test_basic absent …)` + `missing-fixtures: FAIL-CLOSED (#2078)`.

## Why this is worse than it looks

1. **The failure presents as 7 test failures, not as a config error.** An agent that trusts the component names concludes it broke `core-tests` and `memory-budget` and starts editing source. The #3106 run came within one instruction of discarding a correct triage because of this.
2. **`fetch-datasets.sh` exits 0 having done nothing** when the cache is populated — sole output `Dataset cassandra5-small-full-v3.5.tar.gz (tag datasets-v3) already present in /data/datasets; skipping download`. A green fetch is therefore **not** evidence that the repo tree gained fixtures, so the documented remedy silently fails to remedy.
3. **The documented setup in CLAUDE.md does not work as written** on this fleet: `bash test-data/scripts/fetch-datasets.sh` + `CQLITE_DATASETS_ROOT=$PWD/test-data/datasets` yields a corpus-less root, because the fetch extracts to `/data/datasets` and never populates the repo tree.
4. The tempting shortcut is `AGENT_GATE_ALLOW_MISSING_FIXTURES=1`, which buys a green by letting dataset-dependent components **SKIP** — a vacuous PASS that certifies nothing. That is exactly what #2078 exists to prevent, so the wrong fix is also the easy one.

## Current workaround (used to certify #3106)

A **non-destructive composite root** assembled by symlink — `sstables` from `/data/datasets`, git-tracked `commitlog` from the repo, and the `test-data/` siblings so `../schemas` resolves — touching neither the shared root checkout nor any worktree. Verified: full `Data.db` count, `../schemas/basic-types.cql` resolves, 4 commitlog fixtures present.

Note two of the 7 failing tests are `cli-helpers`-gated and so **do not appear in a default run** — their absence must not be read as coverage.

## Fix shape (acceptance)

1. **One documented root that works**, verified end to end on a fresh box. Either (a) `fetch-datasets.sh` extracts/symlinks into `<repo>/test-data/datasets` so the documented invocation is true, or (b) the `schemas/` tree ships alongside the extracted corpus so `/data/datasets` is self-sufficient. (a) is preferable — it keeps CLAUDE.md honest and keeps the corpus adjacent to its `.jsonl` oracles.
2. **`fetch-datasets.sh` must not exit 0 without leaving a usable root.** If it skips the download because the cache is warm, it must still guarantee (and verify) the extraction target, and fail loudly otherwise.
3. **`preflight` should detect the `../schemas` sibling**, not just the corpus, and name the missing path — so this class of misconfiguration fails at preflight with an actionable message instead of masquerading as 7 test failures deep in `core-tests`.
4. Update CLAUDE.md / the [test-data doctrine page](https://pmcfadin.github.io/cqlite/agents-developing/test-data/) with whatever the real answer is, in the same change.

Related: **#2878** (`fetch-datasets.sh` `rm -rf` deletes git-tracked commitlog fixtures; `restore_ci_tracked_dataset_files` silently no-ops) is the sibling defect in the same script — worth fixing together, since both are "the fetch script leaves the tree in a state the gate can't use." In the #3106 run the `rm -rf` hazard did **not** fire (zero deleted tracked files, all 4 commitlog fixtures intact) because the fetch short-circuited on the warm cache.

Routing: oracle-driven infra fix — issue + a verification that a fresh box reaches a green gate from the documented commands alone. No OpenSpec change.

### LINKED ISSUE #3148: agent-gate: fixture preflight validates SSTable data but not the sibling schemas root — a mis-laid-out corpus passes preflight STATUS: OK and fails ~8 min later as opaque missing-.cql panics
## Summary

The #2078 fail-closed fixture preflight in `scripts/agent-gate.sh` validates **only** the SSTable data corpus — `$CQLITE_DATASETS_ROOT/sstables/test_basic/*-Data.db`. It never validates the **sibling schemas root** that the fixture helpers reach via `datasets_root().join("../schemas")`. `scripts/agent-gate.sh` contains **zero** occurrences of the string `schemas` (verified: `grep -c schemas scripts/agent-gate.sh` → `0`).

Consequence: a corpus whose `sstables/` is complete but whose sibling `../schemas/` is unreachable **passes the preflight with `STATUS: OK`**, builds for ~8 minutes, and then fails two components with opaque "missing `.cql` path" panics — instead of failing closed in one second with a one-line remedy.

## Observed (today, a real certification run)

Full gate with `CQLITE_DATASETS_ROOT=/data/datasets` (the box's full 155-dataset corpus, a real directory) and **no `/data/schemas`**:

- Preflight: `STATUS: OK` — no warning, no marker, nothing in the SUMMARY suggesting an incomplete layout.
- ~8 minutes of build, then 14 components PASS and **2 FAIL**: `core-tests` and `memory-budget`.
- **Five failing tests**, all panicking on an absent `.cql` path:
  - `dead_cache_delete_tests::stats_*` (4 tests) — need `basic-types.cql`
  - `memory_budget::*`, `issue_1494_converter_alloc_budget`, `issue_2075_row_assembly_alloc_budget` — need `basic-types.cql`, `collections.cql`, `wide-rows.cql`
- Total cost: a ~20-minute certification run, discarded.

Worse than the wasted time: the failures were **nearly misattributed to the read-path change under test** (#3095 / PR #3141). The panics name a missing file, but the gate had already told the agent its fixtures were fine, so the first hypothesis was "my change broke fixture loading."

## Root cause

Two halves.

**1. The preflight's validation scope is narrower than the fixture contract.**

`_fixture_status()` — `scripts/agent-gate.sh:1245-1255` — is the entire fixture check. Its one substantive line:

```
scripts/agent-gate.sh:1250
  n=$(find "$CQLITE_DATASETS_ROOT/sstables/$CANONICAL_FIXTURE_KEYSPACE" -name "*-Data.db" 2>/dev/null | wc -l | tr -d ' ')
```

`CANONICAL_FIXTURE_KEYSPACE="test_basic"` (`:1229`). Non-zero count → `OK`. That is the whole predicate. `apply_fixture_preflight()` (`:1262-1287`) consumes it; the call site is `scripts/agent-gate.sh:6194`, inside `if selected_needs_datasets`. The hidden self-test hook that reports `STATUS: <token>` is `--preflight-fixtures` at `scripts/agent-gate.sh:1591-1593`.

So the preflight asserts "SSTable bytes exist" and reports that as fixture readiness. It says nothing about schemas.

**2. The schemas root is derived from the *datasets* root by climbing `..`, in four independently written places.**

```
cqlite-core/benches/fixtures/mod.rs:73-75
  pub fn schemas_root() -> PathBuf { datasets_root().join("../schemas") }
```

```
cqlite-core/tests/dead_cache_delete_tests.rs:604
  let schema_path = root.join("../schemas").join(schema_file);
```

```
cqlite-core/tests/observability_correctness.rs:548
  let schema_path = datasets_root().join("../schemas").join(fx.schema_file);
```

```
cqlite-cli/benches/export_csv.rs:124
  let schema_path = datasets_root().join("../schemas").join(SCHEMA_FILE);
```

The named helper `schemas_root()` exists exactly **once** (the bench module). Three of the four sites open-code the same `join("../schemas")` idiom instead of calling it. The bench module is pulled into four test binaries by `#[path = "../benches/fixtures/mod.rs"]`: `cqlite-core/tests/memory_budget.rs`, `issue_1494_converter_alloc_budget.rs`, `issue_2075_row_assembly_alloc_budget.rs`, `tail_latency_harness.rs`. Its fixture catalog references 6 distinct `.cql` files (`basic-types.cql`, `da-test.cql`, `time-series.cql`, `wide-table-bti.cql`, `collections.cql`, `wide-rows.cql` — `cqlite-core/benches/fixtures/mod.rs:144-224`), resolved at `:282`.

`test-data/schemas/` holds **23 committed files** (18 `.cql`, 4 `.json`, plus `core.list`), including the `legacy/` and `udts/` subdirectories. These are **committed source**, not part of the fetched corpus.

### The two `datasets_root()` implementations do NOT agree

This is a second, independent defect. The bench helper and the `dead_cache_delete_tests` helper resolve the datasets root differently:

```
cqlite-core/benches/fixtures/mod.rs:60-66     -> PathBuf  (infallible)
  Ok(root)  => PathBuf::from(root)
  Err(_)    => CARGO_MANIFEST_DIR/../test-data/datasets     // checkout fallback
```

```
cqlite-core/tests/dead_cache_delete_tests.rs:36-44   -> Option<PathBuf>  (fallible)
  env set AND p.is_dir()  => Some(p)
  otherwise               => None                            // no checkout fallback; test SKIPs
```

`cqlite-core/tests/observability_correctness.rs:500-506` is a third copy, matching the bench semantics. So the same logical root has infallible-with-checkout-fallback semantics in one place and fallible-skip-if-unset semantics in another. Neither checks that the resolved `../schemas` exists; both defer the failure to a `panic!`/`expect` deep inside ingest. A gate whose `CQLITE_DATASETS_ROOT` is set but incomplete therefore gets **different behavior per test target** for the same environment — the exact silent divergence that makes this class of failure hard to attribute.

## The symlink trap (the part a future operator will re-hit)

`join("../schemas")` is not a lexical parent operation at the syscall level. The kernel resolves `datasets/..` against the **symlink target's** parent, not the symlink's own parent. So for a corpus assembled with symlinks:

- `ln -s <checkout>/test-data/datasets /data/datasets` → `/data/datasets/../schemas` resolves to `<checkout>/test-data/schemas`, **not** `/data/schemas`. Whether this works is then an accident of what the target's sibling happens to be.
- `/data/datasets` a **real** directory + `ln -s <checkout>/test-data/schemas /data/schemas` → works.

Two visually equivalent layouts, opposite outcomes, and **no error message explains why**. An operator who symlinks `datasets` (the natural choice — it's the big directory) gets a preflight `STATUS: OK` and a panic 8 minutes later naming a path that, from their shell, appears to exist.

## Why it matters

**#2078 exists precisely so that a fixture-absent gate fails closed instead of passing vacuously.** Its own comment block (`scripts/agent-gate.sh:1216-1227`) states the reasoning: dataset components SKIP internally when the corpus is missing, the gate SKIP-then-PASSes, and the result is "a green SUMMARY validating ZERO dataset-backed correctness."

A preflight that green-lights an **incomplete** fixture layout is the same class of defect, one directory over. And it is **worse than having no preflight at all**, because `STATUS: OK` is *positively misleading*. With no preflight, an agent hitting five `.cql` panics investigates its environment. With `STATUS: OK` already recorded, the agent reads "fixtures verified" and looks at its own diff instead. That is exactly what happened on this run: an environmental failure was nearly attributed to a read-path change it had nothing to do with. A verification step that produces false confidence actively degrades attribution.

This is the third recent instance of the same shape — **the verification was the broken part**: #3130 (lint scan roots), #3127, and this.

## Proposed fix

1. **Extend the preflight to validate the schemas root**, with a **named fail-closed error plus a remedy line** naming the expected absolute path — mirroring the existing `missing-fixtures: FAIL-CLOSED (#2078)` treatment (a distinct marker, e.g. `missing-schemas:`, so the two causes are separable in a pasted SUMMARY). Validate readability of the specific `.cql` files the gate's components actually consume, not just directory existence.
2. **De-duplicate the resolution into one helper** both the bench and test paths use, and **reconcile the two divergent `datasets_root()` semantics** into one documented contract (fallible-skip vs infallible-fallback is a real behavioral choice — pick one and state it).
3. **Make resolution symlink-robust**: either resolve the schemas root without climbing `..` through a possibly-symlinked `datasets`, or **detect the symlinked-`datasets` layout and reject it with an explanatory error** — never silently mis-resolve.
4. **Reconsider the design premise.** The schema fixtures are **committed source** (23 files in `test-data/schemas/`, including `legacy/` and `udts/`), not part of the fetched corpus. Deriving their location from `CQLITE_DATASETS_ROOT` — an env var whose entire purpose is to point at *fetched, relocatable* data — is arguably the real design error. Resolving them relative to the **checkout** (`CARGO_MANIFEST_DIR`, as `cqlite-core/tests/schema_aggregator_integration_test.rs:54` already does: `PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("../test-data/schemas")`) would make the whole failure mode structurally impossible, since a checkout always has them. If that is adopted, the preflight check becomes a cheap belt-and-braces assert rather than the primary defense. This issue should decide between the two, not just paper over the current coupling.

## Routing

**Design-driven** — this is a gate/tooling **contract** change (what the preflight promises, where committed schema fixtures are discovered from, which env var owns which root). There is no on-disk oracle to pin. Route through OpenSpec via `flow-activate`.

Precedent: **#2964** and **#3130** were routed the same way (mechanism/verification-contract changes, design-driven, no parity oracle).

## Acceptance criteria

- [ ] **(a)** The FULL-gate preflight **FAILS CLOSED** with a named error and a remedy line when the schemas root is missing or unreadable — never `STATUS: OK`.
- [ ] **(b)** The failure text names the **exact expected absolute path** and the **fix command**, and the SUMMARY carries a marker distinguishable from `missing-fixtures:` (#2078).
- [ ] **(c)** A **positive-control self-test** proves the preflight FAILS on a schemas-less root — not merely that it passes on a good one. (A preflight tested only on the happy path is untested; that is how this gap survived. Extend the existing `--preflight-fixtures` hook + the `scripts/tests/test_agent_gate_summary.sh` style self-test.)
- [ ] **(d)** The schemas-root resolution exists **once** and is used by every caller: `cqlite-core/benches/fixtures/mod.rs:74`, `cqlite-core/tests/dead_cache_delete_tests.rs:604`, `cqlite-core/tests/observability_correctness.rs:548`, `cqlite-cli/benches/export_csv.rs:124`. No open-coded `join("../schemas")` remains.
- [ ] **(e)** The two divergent `datasets_root()` implementations (`benches/fixtures/mod.rs:60` infallible-with-fallback vs `dead_cache_delete_tests.rs:36` fallible-no-fallback) are reconciled, or their difference is documented as deliberate with a stated reason.
- [ ] **(f)** The **symlinked-`datasets` layout** either works correctly or is rejected with an explicit error naming the symlink as the cause — never silently mis-resolved via `..`.
- [ ] **(g)** `--lite` and `--only` remain **lenient** (unchanged from #2078's contract); only the FULL gate is strict.
- [ ] **(h)** Scope decision recorded: whether schema fixtures move to checkout-relative discovery (proposed fix 4), and whether any other env-var-derived sibling is in or out of scope.

## Scope note — other unvalidated roots (survey result)

Surveyed every `join("..")` and `CQLITE_DATASETS_ROOT` consumer. The **only** env-var-derived sibling escaping the preflight is `../schemas` (4 sites, above). Everything else divides into:

- **Checkout-relative** (`CARGO_MANIFEST_DIR`-anchored), hence always present in any worktree and not a preflight concern: `test-data/parity/*.jsonl` (`stcs_bucket_selection_parity.rs:32`), `test-data/schemas` direct (`schema_aggregator_integration_test.rs:54,693`), `trino-connector/src/test/resources/golden/` (`cqlite-flight/tests/do_get_transport_test.rs:262`), and the `join("..")` uses in `commitlog_reader_parity.rs`, `issue_911_bti_sstabledump_parity.rs`, `issue_1007_complex_type_parity.rs`, `issue_1295_signed_collection_order_parity.rs`, `statistics_corrupt_hardfail_test.rs`.
- **Under `$CQLITE_DATASETS_ROOT/sstables/`** and thus already inside the preflight's reach, though only the `test_basic` keyspace is actually asserted — other keyspaces (`test_compactionparity`, `test_compaction_tombstone_ttl`) are checked ad hoc by individual components (`scripts/agent-gate.sh:4238`, `:4295`) rather than by the preflight. Not a new fail-open (those components SKIP loudly), but worth noting as adjacent inconsistency; out of scope unless the design pass chooses to unify.

So the scope stays as titled: one sibling root, one idiom, four call sites.

## References

- **#2078** — the fail-closed fixture preflight this extends (`scripts/agent-gate.sh:1216-1287`, call site `:6194`).
- **#3095** / **PR #3141** (Flight static-column `do_get` semantics) — the delivery during which this was found; that run's `core-tests` + `memory-budget` failures were **environmental**, not caused by the change under test.
- **#3130** (lint scan roots) and **#3127** — sibling mechanism gaps; all three are "the verification was the broken part" cases, and the routing precedent for this one.
- **#2964** — design-driven routing precedent for a gate/tooling contract change.
- CLAUDE.md, "The Agent Gate": the FULL gate FAILs CLOSED when the fetched validation corpus is absent (#2078), stamping `missing-fixtures: FAIL-CLOSED (#2078)`.

pmcfadin:
# Resolve committed CQL schema fixtures checkout-relative; make the gate preflight validate them

Closes #3131
Closes #3148

The CQL schema fixtures under `test-data/schemas/` are **committed source** (23 files including
`legacy/` and `udts/`), but four independently-written call sites located them by climbing `..` from
`$CQLITE_DATASETS_ROOT` — an env var whose entire purpose is to point at *fetched, relocatable* data.
Two failures followed from that one coupling:

* **No single root worked on the fleet (#3131).** `/data/datasets` (where `fetch-datasets.sh` caches
  and extracts) holds ~155 `-Data.db` but `/data` has no `schemas/`; the CLAUDE.md-documented
  `<repo>/test-data/datasets` has the sibling but **zero** `test_basic` fixtures. Pointing at the
  former cleared the preflight and then failed 5–8 tests on
  `Path does not exist: /data/datasets/../schemas/basic-types.cql`.
* **The gate's preflight validated only half the fixture contract (#3148).**
  `grep -c schemas scripts/agent-gate.sh` was **0**: it counted `sstables/test_basic/*-Data.db` and
  reported that as *fixture readiness* — `STATUS: OK` — then failed `core-tests` and `memory-budget`
  ~8 minutes later on opaque missing-`.cql` panics. That is **worse than no preflight**: `STATUS: OK`
  is positively misleading, so the agent looks at its own diff. On #3095 / PR #3141 an environmental
  failure was nearly attributed to an unrelated read-path change.

`join("..")` is also not a lexical parent at the syscall level — the kernel resolves `datasets/..`
against the **symlink target's** parent — so two visually identical layouts gave opposite outcomes
with no error explaining why (#3148's "symlink trap").

## The design decision (#3148 AC (h))

**Resolve the schemas root CHECKOUT-RELATIVE, never from `CQLITE_DATASETS_ROOT`** (#3148's proposed
fix 4). A checkout always holds these files, so the failure mode becomes *structurally impossible* and
the symlink trap **disappears** rather than being papered over: zero `..` climbing from the datasets
root remains, so there is nothing left to mis-resolve. `CQLITE_SCHEMAS_ROOT` is honored as an
absolute-only override for out-of-tree runs.

Two alternatives were considered and rejected:

| Alternative | Why rejected |
|---|---|
| Ship `schemas/` inside the dataset archive (#3131 item 1b) | Duplicates **committed source** into a versioned binary asset — a schema edit would then need a dataset re-release, and the two copies can disagree with no signal. |
| Make the fetch populate `<repo>/test-data/datasets` (#3131 item 1a) | Does not fix the general case (any relocated root re-breaks it) and fights the purpose of `CQLITE_DATASETS_ROOT` as a large, relocatable, machine-local cache. |

This **supersedes #3131 item 1's either/or**: with the schemas decoupled, *any* corpus root works, so
`/data/datasets` is self-sufficient unchanged and no data is duplicated.

## Acceptance criteria → evidence

| AC | Evidence |
|----|----------|
| **#3148 (a)** FULL gate fails closed, named error + remedy | `apply_schemas_preflight` (`scripts/agent-gate.sh`); self-test `3148-full-fail` (real full gate, exits at the preflight, non-zero + `RESULT: FAIL`, never `PASS`); negative-control run below |
| **#3148 (b)** exact absolute path + fix command; marker distinguishable from `missing-fixtures:` | self-tests `3148-remedy`, `3148-marker-distinct`, `3148-relative-absolute-label` (the last asserts no relative path ever reaches an "expected absolute path" line, against the **real emit**) |
| **#3148 (c)** positive-control self-test proves the preflight FAILS | `scripts/tests/test_agent_gate_schemas_preflight.sh` — **32 cases**, wired into `tooling-tests`. Rejection cases: schemas-less root, present-but-**incomplete** root, directory-named-like-a-`.cql`, relative override, control-character override. Every fix in this PR was proven to **fail on a revert** (see review history) |
| **#3148 (d)** one definition, used by every caller; no open-coded `join("../schemas")` | `test-data/support/fixture_roots.rs`; self-tests `3148-shared-file`, `3148-all-sites`, `3148-no-dupe-root`, `3148-no-dotdot-climb` (structural reintroduction guard; zero matches) |
| **#3148 (e)** the divergent `datasets_root()` impls reconciled | one impl, two documented shapes (`datasets_root` infallible-with-fallback / `datasets_root_if_present` env-only). Behaviour preserved per test: `dead_cache_delete_tests` still SKIPs; benches still run from a plain checkout. `cqlite-core/tests/issue_3148_fixture_roots_contract.rs` |
| **#3148 (f)** symlinked-`datasets` layout works or is rejected | asserted **behaviourally**: `3148-symlink-independence` (identical resolved root across real / symlinked / nonexistent datasets roots) + `3148-no-dotdot-climb`. The trap is removed, not detected |
| **#3148 (g)** `--lite` / `--only` stay lenient | self-tests `3148-lite` (no schemas line at all), `3148-only`, `3148-only-reject-lenient`, `3148-only-no-false-positive`, `3148-full-positive-line` |
| **#3148 (h)** scope decision recorded | `openspec/changes/schemas-root-contract/{proposal,design}.md` incl. the rejection table above |
| **#3131 (1)** one documented root that works | superseded by (h) — any corpus root works. `3148-symlink-independence`; hostile-layout run below |
| **#3131 (2)** fetch never exits 0 leaving an unusable root | `guarantee_usable_root` on both exit paths + `--verify-only`; self-tests `3131-hollow-root`, `3131-export-line`, `3131-export-quoting`, `3131-verify-no-mutation`, `3131-arg-safety`, `3131-symlinked-root` |
| **#3131 (3)** preflight detects the schemas half | same as #3148 (a); per-**file** readability, not directory existence (`3148-hook-partial`) |
| **#3131 (4)** docs updated in the same change | commits `3077893` (CLAUDE.md), `2daec4a` (website test-data + gate-contract), `0c44bc0` (fleet-runbook + the #3027 report; retires the `/data/schemas` symlink instruction) |

## Review history — three rounds, and three vacuous verifications found in this PR's own work

Recorded in full because it is the strongest evidence the fix is real. This change is *about* a
verification that was silently not verifying, and the same mistake was made three times while fixing it.

* **Round 1 — 3 blockers, all closed pre-gate.** (B3) `fetch-datasets.sh` matched `--verify-only` only
  as `$1`, so `--quiet --verify-only` or any typo silently selected the **destructive** path and
  reached `rm -rf "${DATASET_ROOT}"` against the operator's corpus — a data-loss hazard introduced by
  adding a flag. (B2) `--verify-only` promised to mutate nothing but created the root's parent
  directory. (B1) a **relative** `CQLITE_SCHEMAS_ROOT` was resolved rather than rejected, and the gate
  (CWD = repo root) and cargo (CWD = package dir) resolve it differently — so the SUMMARY stamped
  `schemas: 6/6 … readable under packaged/schemas` while the tests read the checkout. Plus N4 (the
  checkout was identified by the *fixtures*, so a worktree nested inside another checkout silently
  borrowed the outer tree), N6 (dead helper deleted), N7 (`-r` accepted a *directory* named
  `basic-types.cql`).
* **Round 2 — no blockers; 6 nits; roborev round VOIDED** because `origin/main` advanced 4 commits
  mid-review, so `sha-assert` FAILed against a stale base. Rebased onto `8e85b9e` and re-reviewed.
* **Round 3 — mergeable as-is; roborev job 10 valid, 3 findings**, all triaged and fixed.

**The three vacuous verifications this PR found in itself:**

1. **A positive assertion for a check that never ran.** `_schemas_status` returns `OK` unconditionally
   under `--only` (correct leniency), and the OK branch then stamped
   `schemas: 6/6 canonical .cql readable under <root>` **without checking**. Demonstrated by removing
   the fix: it stamped `6/6 … readable` for a root with **zero** readable fixtures. Now stamps an
   explicit `schemas: not checked (<mode> is lenient) — this block asserts NOTHING about the schemas
   root`; silent omission was rejected because it lets a reader assume the full contract held.
2. **An assert that passed after a full revert of the code it guarded.** The AC-(b) "no relative path
   labelled absolute" case matched hook **stdout**, but the string is `apply_schemas_preflight`
   **stderr** — so it could never appear and the case passed unconditionally. Re-pointed at the real
   full-gate emit and proven to FAIL on a revert.
3. **A control that could not discriminate.** The N4 marker-rule test ran only from
   `CARGO_MANIFEST_DIR`, where the retired fixtures-keyed walk returns the *same* path — so it passed
   either way. Now driven over a synthetic `outer/{Cargo.toml,test-data/schemas}` +
   `outer/inner/{Cargo.toml,cqlite-core}` layout where the two rules disagree; proven to FAIL on a revert.

Two roborev findings had a **real defect behind a wrong stated mechanism**, so each premise was
verified before acting: 9-2 was *understated* (a symlinked datasets root did not "verify with count 0",
it failed verification outright, because `find` never descends an un-`-H`'d symlink base), and 9-1's
exploit path was *mis-sited* (unreachable from a caller's CWD, because the gate `cd`s to its own repo
root — the code inconsistency was still real and fixed). One finding, 10-2, was a **regression
introduced by round 2's own fix**: consuming the override through `$( )` strips trailing newlines, so
the gate certified `/abs/dir` while Rust kept the newline and degraded to the checkout.

Neither the shell mirror nor the Rust resolver claims equivalence "by construction" — the docs now say
what is true: **two hand-written mirrors, equivalent today, pinned by the self-test**, with an explicit
instruction to re-walk the input table when either side changes.

## Scope boundaries

* **#2878 untouched.** `rm -rf "${DATASET_ROOT}"` and the `[ -n "${CI:-}" ] || return 0` short-circuit
  in `restore_ci_tracked_dataset_files` are deliberately unchanged; self-test `3131-2878-boundary`
  asserts both remain verbatim so the boundary cannot be silently crossed later.
* **~15 `parent()?.join("schemas")` sites deferred to #3192.** Those files try the datasets sibling
  **and then** a `CARGO_MANIFEST_DIR` checkout fallback, so they *degrade* correctly on the layout that
  hard-failed the four sites fixed here — duplication and a latent wrong-root hazard, not an outage.
  The reintroduction guard therefore greps only the literal `join("../schemas")`; widening it to the
  `parent()`-based form must land **with** that migration, or it would red the gate on ~15 untouched
  files. The spec's single-definition requirement is explicitly scoped to the four sites.
* **Canonical fixture list scoped to 6 of 23 by consumption** — the `.cql` the gate's dataset-backed
  components actually read. Growing it is guard-*completeness* work needing per-component evidence of
  which file each consumes; deliberately not attempted here.
* **No library surface, no format work.** The shared module is test/bench support and never compiles
  into the library.

## Verification

* **Hostile layout** (the layout that caused this): `CQLITE_DATASETS_ROOT` at a scratch root holding
  the corpus with **no** `../schemas` sibling, `CQLITE_SCHEMAS_ROOT` unset — `dead_cache_delete_tests`
  8/8 (incl. the 4 `stats_*`), `memory_budget` 3/3, `issue_1494_converter_alloc_budget` 1/1,
  `issue_2075_row_assembly_alloc_budget` 1/1. Run with `--features cli-helpers` (+ `dhat-heap,arrow`)
  so the feature-gated targets are actually compiled rather than silently empty.
* **Non-vacuity control 1:** with `CQLITE_SCHEMAS_ROOT` at an empty directory, exactly the 4
  schema-consuming `stats_*` tests FAIL with the new actionable message — proving the passes above are
  real schema loads, not skips.
* **Non-vacuity control 2:** with a *relative* `CQLITE_SCHEMAS_ROOT`, all 4 fail closed on
  `must be an ABSOLUTE path` — no silent fallback.
* **Negative-control FULL gate** (complete `/data/datasets` corpus + unreachable schemas root):
  exits 1 **at the preflight, before any cargo component**, stamping
  `missing-schemas: FAIL-CLOSED (#3148)` + `RESULT: FAIL`.
* Self-test 41/41 (grew across six review rounds; certified by `tooling-tests: PASS (334s)` in the gate of record), contract tests 9/9, `test_agent_gate_summary.sh` 101/101 (unperturbed),
  `check-dataset-pin.sh` green, clippy clean on both touched packages.
* **The gate of record runs with this box's `/data/schemas` symlink REMOVED**, so it cannot borrow the
  local workaround this PR exists to make unnecessary.

## Delivery history (not a clean single-gate history)

**Five full-gate runs, six review rounds, three intent audits.** The superseded runs are recorded
because the review history is the strongest evidence the fix is real — not because any of them
certifies anything. Two of the five were genuine `RESULT: PASS` runs that a later review finding
invalidated; presenting only the last one would hide that this change needed four rounds of
adversarial review to become correct.

### Full-gate runs

1. **Run 1 — `919b921` — FAIL on `tooling-tests`.** *Ours, and not waivable*: the diff touches
   `scripts/agent-gate.sh`, so CLAUDE.md's cite-and-waive path (which requires a diff touching no
   compiled input) did not apply. Root cause: round 3's NIT-D fix stubbed `emit_summary`/
   `_tree_meta_array` and thereby created a **duplicate `_tree*` definition**, caught by the
   *pre-existing* `test_agent_gate_tree_portability.sh` uniqueness assert (`n=45 uniq=44`). Fixed in
   round 4; portability 29/29.
2. **Run 2 — `4c3a4c4` — KILLED MID-FLIGHT, no verdict.** Its summary correctly still holds the
   launch-time `RESULT: INCOMPLETE` liveness placeholder (#3041). The intent audit had just found the
   env opt-out defect below; the fix touches `agent-gate.sh`, so finishing the run would have spent
   ~40 min certifying a dead SHA.
3. **Run 3 — `e40cb68` — `RESULT: PASS`, but SUPERSEDED.** A genuine full PASS that is *not* the gate
   of record: the roborev blocker below required a `src` fix, and any src change invalidates the gate
   that preceded it.
4. **Run 4 — `b64bdab` — `RESULT: PASS`, but SUPERSEDED too.** Certified the non-UTF-8 blocker fix,
   then review round 5 raised two further findings against `scripts/agent-gate.sh` and its self-test,
   which again invalidated it.
5. **Run 5 — `0494f8c` — the gate of record.** Block below. Run under `env -u CQLITE_SCHEMAS_ROOT`
   with `/data/schemas` absent (retired to `/data/schemas.RETIRED-3131`) and
   `CQLITE_DATASETS_ROOT=/data/datasets` — **no fixture env var was set to obtain green** in any of
   the five runs.

### The roborev blocker (job 11, finding 1) — non-UTF-8 override drift

The fourth review round found that the mechanism this change installs had a **third** drift instance
of a class it already pinned twice. `std::env::var` maps a non-UTF-8 `CQLITE_SCHEMAS_ROOT` to
`Err(NotUnicode)`, which a catch-all `_ => None` arm collapsed to "unset", so Rust silently fell back
to the checkout **while Bash validated the same byte-path as an override** — the gate certifying a
root the run never used, which is precisely this change's stated forbidden condition. The premise was
verified empirically before fixing: for a real `bad\xff\xfedir` directory, Bash reported `STATUS: OK`
+ `SOURCE: CQLITE_SCHEMAS_ROOT override` while Rust returned `Err(NotUnicode)` and fell back.

Fixed on **both** sides. Rust takes `resolve_schemas_root(Option<&OsStr>)` via `var_os`, so non-UTF-8
is *representable* and gets an explicit `Err` arm rather than inheriting a lossy conversion — and the
catch-all over an error variant, which is what created the defect, is gone. Bash gained
`_gate_schemas_override_is_utf8` plus a `non-utf8` reject kind with its own prose and marker; pure
ASCII is accepted with no external tool, anything else is validated with `iconv`, and **an absent
`iconv` REJECTS** — "could not check" must not mean "accept". Both sides now agree across the full
input table: unset/empty/whitespace → checkout; non-UTF-8, control-char, relative → REJECT;
absolute-non-dir → checkout; absolute-dir → override. A legitimate **multibyte UTF-8** root is still
accepted and asserted, so this is not a ban on non-ASCII paths.

**The datasets root was hardened in the same round, and its hazard was worse than the reported one.**
A non-UTF-8 `CQLITE_DATASETS_ROOT` would have had the gate certify a corpus while every test silently
read the checkout's `datasets`, which holds only committed byte-parity refs and **no `test_basic`** —
a #2078-class vacuous pass with the preflight vouching for a corpus the run never used. (The schemas
fallback was at least the *complete* committed tree; this one is not.) Since `datasets_root()` is
infallible the mechanism is **honor the OS value** (`PathBuf::from(&OsStr)`), not a panic: three
lines, no new failure mode, one fewer silent fallback, and Rust now agrees with what Bash already did.

**Three drift instances are now pinned, up from two of three**: `3148-non-utf8-override` at the real
emit (revert the Bash guard → FAILS with rc=124, the run sails past the preflight),
`non_utf8_override_is_rejected_fail_closed` (revert the Rust arm → FAILS), and
`3148-utf8-override-ok` guarding against over-rejection. The control-char and relative pins were
re-confirmed green.

Two nits from the same round were fixed in that round rather than deferred, since a full re-gate was
already mandated and they therefore triggered no extra verification round: the `resolve_schema_path`
remedy is now `git -C <workspace_root> restore …` (asserted to carry that exact absolute `-C` and to
be CWD-independent — the previous relative form failed when pasted from cargo's package-dir CWD, the
same CWD asymmetry this module rejects *relative overrides* for), and `fetch-datasets.sh` applies
`%q` to `PIN_FILE`/`DATASET_ROOT`, asserted by an `eval` round-trip on a path containing a space
and `&`.

### Review round 5 (roborev job 12) — two Medium findings, both fixed

Round 5 returned `RESULT: FAIL` with two findings, both against the guard this change installs.
Neither could produce a false green; both were fixed anyway rather than deferred, because each
*removes* a dependency and a failure class rather than papering over one.

**Finding 1 — `_gate_schemas_override_is_utf8` piped ASCII probe. Defensive hardening, with a
measured non-reproduction — NOT a fixed live bug.** The probe was `printf | grep -q`, and under the
script-wide `set -o pipefail` (`agent-gate.sh:369`) a **negated** pipeline whose *left* side can take
SIGPIPE is a latent branch-inversion hazard: had it fired, a malformed override would have taken the
"pure ASCII" branch and skipped validation, so Bash would accept a value Rust rejects — this change's
own forbidden condition.

It **did not reproduce**, and that measurement is recorded rather than quietly replaced by a bug-fix
narrative: bash's **builtin** `printf` (what the code used) yields `PIPESTATUS=[0 0]` at 5 B, 100 KB,
1 MB and **5 MB** on bash 5.2.21, so the inversion never fires; and independently, a value large
enough to fill a 64 KB pipe is ~16× this platform's `PATH_MAX` of **4096** and could therefore never
name a real directory. Hardened regardless, because the hazard is platform-scoped rather than absent:
the probe is now a subprocess-free `case` under a function-local `LC_ALL=C`, which removes the
pipeline, the SIGPIPE class **and** the `grep` dependency at once, and was verified to classify
identically under `C`, `C.UTF-8` and `en_US.UTF-8`.

**Finding 2 — `3148-utf8-override-ok` required `iconv`.** The pin demanded a multibyte override be
*accepted*, while the implementation deliberately **rejects** non-ASCII when `iconv` is absent. On an
`iconv`-less host that pin redded `tooling-tests`, and with it the whole gate, on a perfectly valid
root — a **false RED**, not a false green, but a fleet foot-gun now that the self-test runs inside the
gate. It now asserts whichever outcome is *documented* for the host it runs on.

That conditional branch is unreachable wherever `iconv` exists, which would have left the
fail-closed-when-unverifiable promise untested exactly where it matters — the "coverage that never
runs" pattern this change has had to dig out of itself repeatedly. So a new **`3148-utf8-no-iconv`**
case builds a PATH farm symlinking every executable *except* `iconv` and asserts the **same valid
multibyte root** is rejected fail-closed (`STATUS: FAIL`, `SOURCE: … REJECTED`, `must be valid
UTF-8`). It is not a silent skip: a farm that cannot be built is a loud, named failure. Self-test
41/41 (40 → 41).

The **full input table was re-walked through the real hook** after the rewrite, and all eight rows
classify exactly as the intent audit's table walk established: unset / whitespace → checkout;
relative, non-UTF-8 and control-char → REJECT each naming its own reason; absolute-non-dir →
checkout; absolute-dir and multibyte-UTF-8 → override accepted. No row changed.

### Load-bearing check ORDER in the two classifiers (do not reorder)

The two sides deliberately check in different orders — Bash `cntrl → utf8 → relative`
(`_gate_schemas_override_reject_kind`), Rust `utf8 → cntrl → relative`. Consequences, established by
the intent audit's side-by-side table walk:

* A value that is **both** non-UTF-8 and control-bearing gets a different `kind` **label** on each
  side. Both still **REJECT**, so there is no mis-certification — but the labels are *not*
  interchangeable, and a test asserting a specific `kind` must target the right side.
* **The ordering is load-bearing, and its REASON changed with the round-5 rewrite.** Originally it was
  load-bearing because `grep` treats a newline as a line terminator, so a newline-bearing value would
  slip the ASCII fast path. The `grep` is now gone — but the ordering requirement is not, for a
  different and simpler reason: **a newline is valid UTF-8**, so `_gate_schemas_override_is_utf8`
  accepts one. Were control-char screening not to run first, a newline-bearing **absolute** path would
  be accepted outright instead of classified `control-chars`. The in-line comment in
  `scripts/agent-gate.sh` was corrected to state this rather than left asserting a rationale that no
  longer holds — the stale-justification failure mode is the same one this PR is about.

### Intent audits (C) — three verdicts

First **PARTIAL** on 4 items (all later cleared, 5 of 5 on re-audit). The re-audit then found a **new
blocker**: `_SCHEMAS_PREFLIGHT_REPORT_ONLY` was read as `${…:-}` and never initialized, so an
**inherited/exported value flipped the FULL gate's fail-closed `exit 1` into `return 1`** — the run
could proceed to `RESULT: PASS` with the schemas root unreachable. That is an environment opt-out,
violating this change's own requirement 8. Fixed by replacing the variable with a **positional
argument** (`apply_schemas_preflight [report-only]`): the hook passes `report-only`, the real FULL
gate passes nothing, so strict is the default and *no state has to be correct*. Pinned by
`3148-no-env-report-only` (four plausible env spellings; asserts non-zero + `missing-schemas:
FAIL-CLOSED` + `RESULT: FAIL` present + no `RESULT: PASS`); the real discriminator is `RESULT: FAIL`
**present** — impossible for a run that proceeds, since launch stamps `INCOMPLETE` — so subtler
reintroductions are caught too. A class audit of every variable this change introduces found no other
instance. Final verdict **PASS**: requirement 8 holds *by construction* — mode is a `local` set only
from `$1`, exactly two call sites, and `$1` is unreachable from `export`/`env`/`env -i`/a same-named
variable/an inherited exported function.

### Honest boundaries (informational, verdict-neutral — from the intent audit)

* **`BASH_ENV` is arbitrary pre-parse code execution** and defeats any mechanism, including a
  positional argument. Requirement 8 is therefore read as *"no supported/readable env flag"*, **not**
  *"immune to shell code injection"*.
* **The preflight validates the 6 canonical `.cql` while `test-data/schemas` commits 23.** A root
  holding only those 6 passes the preflight, after which components panic on a missing
  `legacy/`/`udts/` file → `RESULT: FAIL`. That is fail-**late**, not fail-open — it cannot certify a
  bad root — and it is the honest boundary of the 6-of-23 consumption scoping (widening it is
  guard-completeness work; see Scope boundaries).

## Agent gate (gate of record)

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.wJwk9k
commit: 0494f8c branch: issue-3131-schemas-root-contract dirty: no
schemas: 6/6 canonical .cql readable under /home/ubuntu/workspace/cqlite-wt/issue-3131/test-data/schemas (checkout-relative)
tree-start: 0494f8cfbadb dirty: no digest: fd0ea68ab67c
tree-end: 0494f8cfbadb dirty: no digest: fd0ea68ab67c
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (5s)
clippy:            PASS (1s)
roborev-lints:     PASS (22s)
core-tests:        PASS (74s)
tombstones-scan:   PASS (0s)
scan-offload-guard: PASS (2s)
byte-budget-guard: PASS (13s)
arrow-parity-guard: PASS (1s)
memory-budget:     PASS (10s)
integration-tests: PASS (0s)
format-compat:     PASS (0s)
write-tests:       PASS (10s)
cli-tests:         PASS (4s)
python-bindings:   PASS (79s)
node-bindings:     PASS (120s)
delivery-telemetry: PASS (1s)
oom-audit:         PASS (1s)
parity-report:     PASS (0s)
operator-metrics-doc: PASS (0s)
tooling-tests:     PASS (334s)
minimal-build:     PASS (0s)
smoke:             PASS (2s)
logs: /tmp/agent-gate.wJwk9k
summary-file: /tmp/gate-3131-record5.txt
RESULT: PASS
```

Run under `env -u CQLITE_SCHEMAS_ROOT` with `CQLITE_DATASETS_ROOT=/data/datasets` and this box's
`/data/schemas` symlink **retired**, so the run could not borrow the local workaround this PR
exists to make unnecessary. No fixture env var was set to obtain green in any of the five runs.
Reader contract (#2874) discharged by the closer: process **exit code 0**, and
`run-id: /tmp/agent-gate.wJwk9k` **matches the sentinel captured at launch**, so this block
belongs to this run and not to a concurrent peer.

## roborev (final pass)

> [!IMPORTANT]
> **This PR does NOT claim "roborev clean."** The final review round reported
> `RESULT: FAIL`, and the block below is reproduced **verbatim, including that line**.
> It is merged under an explicit, narrowly-scoped owner authorization recorded at
> https://github.com/pmcfadin/cqlite/issues/3131#issuecomment-5145226722
>
> The override covers **round 6's three findings only** — one Medium locale-dependent
> classifier divergence (Rust *panics*, so it is loud and fail-closed, it cannot produce a
> false certification, it is unreachable in the gate of record, and two independent intent
> audits rated it verdict-neutral) and two Low findings (a test-only panic-hook swap, and
> an unquoted interpolation in a *printed* remedy string). It does **NOT** waive the gate
> of record — run 5 is `RESULT: PASS`, 0 FAIL components, `tree-integrity: PASS` on the
> certified SHA — and it does **NOT** waive the intent audit, which returned PASS across
> three audits. All three findings are tracked in the linked follow-up issue, each with
> its reproduction.

```
repo: /home/ubuntu/workspace/cqlite-wt/issue-3131
branch: issue-3131-schemas-root-contract
base: origin/main
head-sha: 0494f8cfbadb01db81fdc5a2e5cf0e33ad753ad6
reviewed-sha: 221d16dc5ea3927cd15befaf78bf4a5a416d7e62..0494f8cfbadb01db81fdc5a2e5cf0e33ad753ad6
job: 13
model: gpt-5.6-sol
census: 18 files, +3759/-84
tokens: input=180883 cached=119296 output=8073
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (9/9 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: PRESENT (3)
log: /tmp/roborev-3131-r6.log
RESULT: FAIL
```

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -141,10 +141,18 @@ cat /tmp/gate-summary.txt   # the SUMMARY block is the ONLY gate text an agent r
   DuckDB amalgamation (cqlite-cli `duckdb-tests`) + OTel stack (`observability`/
   `observability-testing`); parquet/arrow stay linted. `CQLITE_CLIPPY_FULL=1` (nightly `gate.yml`)
   runs the full matrix.
-- The FULL gate FAILs CLOSED when the fetched validation corpus is absent (#2078), stamping
-  `missing-fixtures: FAIL-CLOSED (#2078)`; `AGENT_GATE_ALLOW_MISSING_FIXTURES=1` opts out visibly
-  (`missing-fixtures: OPT-OUT (...)`). Remedy: `bash test-data/scripts/fetch-datasets.sh`.
-  `--lite`/`--only` stay lenient.
+- The FULL gate FAILs CLOSED on **either half** of the fixture contract; `--lite`/`--only` stay
+  lenient for both.
+  - Fetched corpus absent (#2078): `missing-fixtures: FAIL-CLOSED (#2078)`, remedy
+    `bash test-data/scripts/fetch-datasets.sh`; `AGENT_GATE_ALLOW_MISSING_FIXTURES=1` opts out
+    visibly (`missing-fixtures: OPT-OUT (...)`).
+  - Committed CQL schemas unreachable (#3148): `missing-schemas: FAIL-CLOSED (#3148)` — textually
+    distinct from #2078's marker, with two causes, an unreadable `test-data/schemas/*.cql` or a
+    **rejected relative `CQLITE_SCHEMAS_ROOT`**, each carrying its own remedy line. Success stamps a
+    positive `schemas: N/N canonical .cql readable under <root> (<source>)` line, so a pasted SUMMARY
+    shows the check RAN. **There is deliberately NO opt-out env var, and none may be added**:
+    committed source in a checkout is never legitimately absent, so an escape hatch could only buy a
+    vacuous green.
 - **A run whose worktree mutates MID-RUN cannot certify (#2926).** Every mode captures a tree
   identity at start, re-verifies it at each component boundary + the terminal emit, and FAILs closed
   with `tree-integrity: FAIL (tree-mutated-midrun; head <a>→<b>; changed: …)`. Every SUMMARY carries
@@ -160,11 +168,12 @@ cat /tmp/gate-summary.txt   # the SUMMARY block is the ONLY gate text an agent r
 
 ```bash
 cargo build
-env CQLITE_DATASETS_ROOT=$PWD/test-data/datasets cargo test --package cqlite-core
+cargo test --package cqlite-core            # needs CQLITE_DATASETS_ROOT exported — see "Test Data"
 env RUSTFLAGS="-D warnings" cargo clippy --workspace --all-targets --all-features   # CI mode
 cargo fmt
 bash test-data/scripts/smoke-test-all-tables.sh
-bash test-data/scripts/fetch-datasets.sh    # fetch real SSTable binaries (required for integration tests)
+bash test-data/scripts/fetch-datasets.sh    # fetch real SSTable binaries; USE the export line it prints
+bash test-data/scripts/fetch-datasets.sh --verify-only   # is my root usable? mutates nothing
 ```
 
 Everything else (CLI usage/modes/output precedence, Python/Node build + test + examples, write
@@ -270,7 +279,21 @@ Location: `test-data/datasets/sstables/` — keyspaces `test_basic` (8), `test_c
 `test_timeseries` (9), `test_wide_rows` (8). **Pass rate: 100% (33/33, Dec 2025).**
 
 The repo ships only JSONL reference files; fetch real binaries with
-`bash test-data/scripts/fetch-datasets.sh` and set `CQLITE_DATASETS_ROOT=$PWD/test-data/datasets`.
+`bash test-data/scripts/fetch-datasets.sh`, then export **the exact
+`export CQLITE_DATASETS_ROOT=<abs>` line that script prints** — it names the only root that run
+guarantees, and on a fleet box it is often a machine-local root (e.g. `/data/datasets`), NOT
+`$PWD/test-data/datasets`. The printed line beats any root remembered from this file. The script
+rejects every unrecognized argument (exit 2) because its default path is destructive
+(`rm -rf` on the dataset root); `--verify-only` probes a root without mutating anything, `--help`
+lists the flags.
+
+**`CQLITE_DATASETS_ROOT` alone is sufficient on every layout (#3131/#3148)** — the corpus root needs
+no `schemas` sibling. The CQL schema fixtures (`test-data/schemas`, 23 committed files incl.
+`legacy/` + `udts/`) are **committed source resolved checkout-relative** (anchored on the
+workspace-root `Cargo.toml`), never derived from `CQLITE_DATASETS_ROOT`. `CQLITE_SCHEMAS_ROOT` is an
+optional out-of-tree override and **MUST be absolute**: a relative value is rejected fail-closed by
+both the gate and the tests, because the gate resolves it against the repo root while cargo resolves
+it against each package dir — so it would certify one schemas root while the tests read another.
 Without Data.db files, query tests pass but return 0 rows. Dataset pins:
 [test data](https://pmcfadin.github.io/cqlite/agents-developing/test-data/).
 
@@ -285,8 +308,10 @@ recipes: `docs/development/dev-cookbook.md`.
 
 ## Troubleshooting
 
-- **Missing test data / 0 rows**: `export CQLITE_DATASETS_ROOT=$PWD/test-data/datasets` +
-  `bash test-data/scripts/fetch-datasets.sh`
+- **Missing test data / 0 rows**: `bash test-data/scripts/fetch-datasets.sh`, then export the
+  `CQLITE_DATASETS_ROOT=` line it prints — NOT `$PWD/test-data/datasets`, which on a fleet box is a
+  corpus-less root the fetch never populates. `--verify-only` re-checks an existing root
+  non-destructively. No `schemas` sibling is needed (#3131).
 - **Clippy failures**: run with `RUSTFLAGS="-D warnings"` to match CI
 - **Parsing issues**: `docs/sstables-definitive-guide/chapters/appendix-f-known-limitations.md`
 - **Python bindings**: Rust 1.85+, Python 3.9+, `pip install maturin`, then
@@ -462,7 +487,9 @@ end-to-end test. Green helper-only unit tests are not sufficient.
   thing. Doctrine: [delivery pipeline](https://pmcfadin.github.io/cqlite/agents-developing/delivery-pipeline/).
 - **1:1:1:1**: one issue ↔ one worktree/branch `issue-<N>-<slug>` (branched from `origin/main`) ↔
   one OpenSpec change `<slug>` ↔ one PR. Worktrees lack gitignored Data.db binaries — point
-  `CQLITE_DATASETS_ROOT` at the main repo's `test-data/datasets`.
+  `CQLITE_DATASETS_ROOT` at the root the fetch's printed export line names (often machine-local, e.g.
+  `/data/datasets`), which is not necessarily the main repo's `test-data/datasets`. The committed CQL
+  schemas need no env var: they resolve from the worktree's own checkout (#3148).
 - **Board = sole dispatch authority (Path A, #1886)**: the GitHub Project `Status` field
   (`Backlog/Ready/In Progress/In Review/Done`); exactly one `P0`–`P3` per issue. New issues auto-land
   at `Backlog`. Empty Ready column = no work ready → STOP. Board unreachable (auth/scope) → STOP and
--- a/cqlite-cli/benches/export_csv.rs
+++ b/cqlite-cli/benches/export_csv.rs
@@ -21,7 +21,10 @@
 //! type-heavy `test_collections.collection_table` (500 rows × 7 cols). The
 //! fixture-open logic is duplicated here (~30 lines) rather than sharing the
 //! `cqlite-core/benches/fixtures` module across crates — a deliberate,
-//! spec-blessed duplication (no new shared crate).
+//! spec-blessed duplication (no new shared crate). The *root resolution* is NOT
+//! duplicated: since #3148 it comes from the single shared
+//! `test-data/support/fixture_roots.rs`, so the datasets/schemas contract cannot
+//! drift between this bench and cqlite-core.
 //!
 //! Wiring evidence / non-vacuity: the bench runs the real `SELECT *`, **panics**
 //! at setup if it returns zero rows, and asserts the CSV writer produced a real
@@ -42,18 +45,19 @@ use criterion::{criterion_group, criterion_main, Criterion};
 // CSV export writer over a single loaded type-heavy fixture (cli-helpers).
 // ---------------------------------------------------------------------------
 
-/// Locate the `test-data/datasets` root (mirrors the core bench fixture loader).
-/// Prefers `CQLITE_DATASETS_ROOT`; else the workspace-relative fallback (the
-/// crate dir is `<workspace>/cqlite-cli`, so datasets live at
-/// `<workspace>/test-data/datasets`).
+/// The ONE definition of every `test-data/` root (issues #3131 / #3148), shared
+/// verbatim with `cqlite-core`'s benches/tests. Hosted under `test-data/support/`
+/// rather than in either crate: it encodes the layout of `test-data` itself, so the
+/// include path is symmetric (`../../` from any `<crate>/benches/`) and this bench
+/// still takes no dependency on `cqlite-core`'s bench internals.
+#[path = "../../test-data/support/fixture_roots.rs"]
+mod fixture_roots;
+
+/// Locate the `test-data/datasets` root — infallible shape (checkout-relative
+/// fallback when `CQLITE_DATASETS_ROOT` is unset).
 #[cfg(feature = "cli-helpers")]
 fn datasets_root() -> std::path::PathBuf {
-    match std::env::var("CQLITE_DATASETS_ROOT") {
-        Ok(root) => std::path::PathBuf::from(root),
-        Err(_) => {
-            std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("../test-data/datasets")
-        }
-    }
+    fixture_roots::datasets_root()
 }
 
 /// Resolve the CFID-suffixed `<keyspace>/<table>-<hash>` SSTable directory, or
@@ -121,7 +125,8 @@ fn bench_csv_export(c: &mut Criterion) {
         .join(src.file_name().expect("fixture dir has a final component"));
     copy_dir_recursive(&src, &dst);
 
-    let schema_path = datasets_root().join("../schemas").join(SCHEMA_FILE);
+    // Committed source, checkout-relative — never `datasets_root/../schemas` (#3148).
+    let schema_path = fixture_roots::schema_path(SCHEMA_FILE);
     let cfg = IngestionConfig {
         schema_paths: vec![schema_path],
         data_dir: tmp.path().to_path_buf(),
--- a/cqlite-core/benches/fixtures/mod.rs
+++ b/cqlite-core/benches/fixtures/mod.rs
@@ -50,28 +50,36 @@ pub fn seeded_rng() -> rand::rngs::StdRng {
     rand::rngs::StdRng::seed_from_u64(BENCH_SEED)
 }
 
-/// Locate the `test-data/datasets` root.
-///
-/// Prefers `CQLITE_DATASETS_ROOT`; otherwise falls back to the workspace-relative
-/// path derived from `CARGO_MANIFEST_DIR` (the crate dir is
-/// `<workspace>/cqlite-core`, so the datasets live at
-/// `<workspace>/test-data/datasets`). The fallback lets the benches run from a
-/// plain checkout with fetched datasets and no environment setup.
+/// The ONE definition of every `test-data/` root (issues #3131 / #3148). Hosted
+/// under `test-data/support/` because it encodes the layout of `test-data` itself
+/// and is shared with `cqlite-cli`'s benches and the integration tests; the
+/// helpers below are thin delegations so there is exactly one implementation.
+/// `#[path]` here is relative to THIS file's directory, so it resolves identically
+/// no matter which bench/test target `#[path]`-includes `fixtures/mod.rs`.
+#[path = "../../../test-data/support/fixture_roots.rs"]
+pub mod roots;
+
+/// Locate the `test-data/datasets` root — infallible shape (checkout-relative
+/// fallback when `CQLITE_DATASETS_ROOT` is unset), so benches run from a plain
+/// checkout with no environment setup. See [`roots`] for the full contract and the
+/// fallible [`roots::datasets_root_if_present`] shape used by SKIP-gated tests.
 pub fn datasets_root() -> PathBuf {
-    match std::env::var("CQLITE_DATASETS_ROOT") {
-        Ok(root) => PathBuf::from(root),
-        Err(_) => PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("../test-data/datasets"),
-    }
+    roots::datasets_root()
 }
 
 /// The `sstables/` subtree holding per-keyspace fixture data.
 pub fn sstables_root() -> PathBuf {
-    datasets_root().join("sstables")
+    roots::sstables_root()
 }
 
-/// The `test-data/schemas` directory (sibling of `datasets`).
+/// The committed `test-data/schemas` directory.
+///
+/// Resolved CHECKOUT-RELATIVE, never as `datasets_root().join("../schemas")` — these
+/// are committed source, not fetched data, and the old `..` climb mis-resolved through
+/// a symlinked `datasets` (#3148 AC (h)). Prefer [`roots::schema_path`], which also
+/// verifies the file is readable and names the absolute path when it is not.
 pub fn schemas_root() -> PathBuf {
-    datasets_root().join("../schemas")
+    roots::schemas_root()
 }
 
 /// Resolve the on-disk SSTable directory for `<keyspace>/<table>-<hash>`.
@@ -279,7 +287,7 @@ pub fn open_read_db_with_config(fx: &ReadFixture, core_config: cqlite_core::Conf
         .join(src.file_name().expect("fixture dir has a final component"));
     copy_dir_recursive(&src, &dst);
 
-    let schema_path = schemas_root().join(fx.schema_file);
+    let schema_path = roots::schema_path(fx.schema_file);
     let cfg = IngestionConfig {
         schema_paths: vec![schema_path],
         data_dir: tmp.path().to_path_buf(),
--- a/cqlite-core/tests/dead_cache_delete_tests.rs
+++ b/cqlite-core/tests/dead_cache_delete_tests.rs
@@ -33,15 +33,17 @@ fn require_fixtures() -> bool {
     )
 }
 
-fn datasets_root() -> Option<PathBuf> {
-    if let Ok(root) = std::env::var("CQLITE_DATASETS_ROOT") {
-        let p = PathBuf::from(root);
-        if p.is_dir() {
-            return Some(p);
-        }
-    }
-    None
-}
+/// The ONE definition of every `test-data/` root (issues #3131 / #3148): the datasets
+/// corpus (`CQLITE_DATASETS_ROOT`, relocatable/fetched) and the committed schema
+/// fixtures (checkout-relative, NEVER `datasets_root().join("../schemas")`).
+#[path = "../../test-data/support/fixture_roots.rs"]
+mod fixture_roots;
+
+/// Fallible shape of the datasets root: `Some` only when `CQLITE_DATASETS_ROOT` is set
+/// and names a directory — no checkout fallback, so these tests SKIP (rather than run
+/// against the committed byte-parity references and report a 0-row pass) when the
+/// fetched corpus is unavailable. See `fixture_roots` for the two-shape contract.
+use fixture_roots::datasets_root_if_present as datasets_root;
 
 fn data_db(ks: &str, tbl: &str) -> Option<PathBuf> {
     let base = datasets_root()?.join("sstables").join(ks);
@@ -588,7 +590,6 @@ async fn open_fixture_db(
 ) -> Option<cqlite_core::Database> {
     use cqlite_core::ingestion::{ingest, IngestionConfig};
 
-    let root = datasets_root()?;
     let src = data_db(ks, tbl)?.parent()?.to_path_buf();
     let tmp = tempfile::TempDir::new().expect("temp dir");
     let dst = tmp
@@ -601,7 +602,10 @@ async fn open_fixture_db(
     // reaping the dir mid-test would break reads.
     let _persisted = tmp.keep();
 
-    let schema_path = root.join("../schemas").join(schema_file);
+    // Committed source, resolved checkout-relative — NOT `datasets_root/../schemas`
+    // (#3148). Verified readable here, so an unreachable fixture names its absolute
+    // path instead of panicking anonymously deep inside ingestion.
+    let schema_path = fixture_roots::schema_path(schema_file);
     let cfg = IngestionConfig {
         schema_paths: vec![schema_path],
         data_dir: dst
--- a/cqlite-core/tests/issue_3148_fixture_roots_contract.rs
+++ b/cqlite-core/tests/issue_3148_fixture_roots_contract.rs
@@ -76,10 +76,59 @@ fn absent_or_blank_override_resolves_to_the_checkout() {
 /// must not break every fixture load. (A relative one is rejected instead; see above.)
 #[test]
 fn absolute_but_unusable_override_degrades_to_the_checkout() {
-    let (root, source) = resolve_schemas_root(Some("/nonexistent-cqlite-schemas-3148"))
+    // A GUARANTEED-absent absolute path, built under a fresh TempDir with native path
+    // handling rather than hard-coded as `/nonexistent-…` (roborev job 10, finding 3): a
+    // hard-coded root-level name is not absolute under Windows path semantics and is not
+    // guaranteed absent on Unix — someone can create it, and then this test would silently
+    // assert the OPPOSITE branch. The TempDir is unique per run and the child is never
+    // created, so absence is a property of the construction.
+    let tmp = tempfile::TempDir::new().expect("temp dir");
+    let absent = tmp.path().join("no-such-schemas-dir");
+    assert!(absent.is_absolute(), "TempDir children are absolute");
+    assert!(!absent.exists(), "the child is deliberately never created");
+    let raw = absent.to_str().expect("utf-8 temp path");
+
+    let (root, source) = resolve_schemas_root(Some(raw))
         .expect("an absolute-but-absent override degrades, it is not an error");
     assert_eq!(source, SchemasRootSource::Checkout);
     assert_eq!(root, checkout_test_data_dir().join("schemas"));
+}
+
+/// A control-character-bearing override is REJECTED on both sides (roborev job 10, finding
+/// 2). The gate's shell mirror can only obtain the value through command substitution
+/// somewhere in its history, and `$( )` strips trailing newlines — so admitting such a value
+/// let the gate certify `/abs/dir` while this resolver kept the newline, failed `is_dir()`,
+/// and degraded to the checkout. Rejecting closes the divergence for good.
+#[test]
+fn control_character_override_is_rejected_fail_closed() {
+    let tmp = tempfile::TempDir::new().expect("temp dir");
+    let dir = tmp.path().to_str().expect("utf-8 temp path");
+
+    for (label, raw) in [
+        ("trailing newline", format!("{dir}\n")),
+        ("leading newline", format!("\n{dir}")),
+        ("carriage return", format!("{dir}\r")),
+        ("embedded tab", format!("{dir}\tsub")),
+    ] {
+        match resolve_schemas_root(Some(&raw)) {
+            Err(e) => {
+                assert!(
+                    e.contains("control characters"),
+                    "{label}: rejection must name the rule; got: {e}"
+                );
+                assert!(
+                    e.contains("remedy"),
+                    "{label}: rejection must be actionable; got: {e}"
+                );
+            }
+            Ok((root, source)) => panic!(
+                "{label}: a control-character override was ACCEPTED ({} via {source:?}) — the \
+                 gate's shell mirror strips trailing newlines, so this is a root-certification \
+                 divergence, not a cosmetic issue",
+                root.display()
+            ),
+        }
+    }
 }
 
 /// An ABSOLUTE override naming a real directory wins, and is reported as an override so
--- a/cqlite-core/tests/issue_3148_fixture_roots_contract.rs
+++ b/cqlite-core/tests/issue_3148_fixture_roots_contract.rs
@@ -12,6 +12,7 @@
 //! mutating `CQLITE_SCHEMAS_ROOT`. Env vars are process-global, so an env-mutating test
 //! races every other test in the binary — and a flaky guard gets deleted, not fixed.
 
+use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
 
 #[path = "../../test-data/support/fixture_roots.rs"]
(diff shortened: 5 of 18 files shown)


pmcfadin:
## Follow-ups filed

Merged at `0494f8cfbadb01db81fdc5a2e5cf0e33ad753ad6` (gate of record run 5: `RESULT: PASS`, 0 FAIL components, `tree-integrity: PASS`; C PASS across three audits).

- **#3195** — batched follow-ups from this delivery: the locale-dependent Bash/Rust classifier divergence (reproduced: under `LC_ALL=C` an existing U+0085 directory is accepted as an override while Rust panics — loud/fail-closed, unreachable in the gate of record), the panic-hook test-hygiene item, the unquoted `git -C` remedy interpolation (a regression introduced by this PR's own round-5 fix), plus the verdict-neutral intent-audit items and the twice-observed `roborev list` reporting artifact (correlated to #3157).
- **#3196** — `flow-meta`: a nits-only roborev `FAIL` has no terminating path, since the wrapper emits `FAIL` for *any* finding while doctrine says nits must not trigger a re-verify round. This delivery is the worked example — two genuine `RESULT: PASS` gates were invalidated for findings that could not produce a false certification, and the sixth round still required an owner decision.

Final review round 6 reported `RESULT: FAIL`; that block is reproduced verbatim in the PR body and this PR does **not** claim "roborev clean". Merged under the narrowly-scoped owner authorization at https://github.com/pmcfadin/cqlite/issues/3131#issuecomment-5145226722

Issue #3104, opened 2026-07-28
CLAUDE.md's CQLITE_DATASETS_ROOT guidance makes agents OVERRIDE the fleet machine's correct corpus with the 31-file parity-ref checkout — burns a full-gate run per issue


pmcfadin:
## Problem

CLAUDE.md and the flow-* skills instruct agents to run the gate with the datasets root pointed at the repo checkout:

> Worktrees lack the gitignored `Data.db` binaries — run the gate with `CQLITE_DATASETS_ROOT` pointed at the main repo's `test-data/datasets`.

On this fleet machine that is **wrong, and actively worse than doing nothing**:

| path | `*-Data.db` count | contents |
|---|--:|---|
| `<repo>/test-data/datasets` (what the doc says) | **31** | committed byte-parity references only |
| `/data/datasets` (machine's real corpus) | **155** | 21 keyspaces, superset, tag `datasets-v3` |

`/data/datasets` is **already exported as `CQLITE_DATASETS_ROOT` in the machine profile.** So an agent that follows the doc verbatim *overrides a correct environment variable with an inferior path*, and the FULL gate then fail-closes at preflight (#2078, `missing-fixtures: FAIL-CLOSED`) because `test_basic` binaries are absent — **no component runs, no verdict on the code, one full-gate cycle burned.**

## Observed

Hit while delivering #3058 (PR #3103). Gate run #1 fail-closed at preflight with `tree-integrity: PASS`, matching run-id, exit 1, zero components executed. The closer diagnosed it, re-pointed at `/data/datasets`, and re-launched. Cost: one gate cycle (~15-25 min) plus the diagnosis.

## The second-order bug (worse than the wasted cycle)

`--lite` is deliberately **lenient** about missing fixtures, so it does **not** fail-close on the wrong root. It silently skips dataset-dependent tests. On #3058 all six lite iteration rounds ran against the 31-file root, meaning dataset-dependent coverage was quietly absent for the entire fix loop and only became real at the gate of record. That is precisely the "never let a dataset-dependent test pass on an empty dataset" hazard CLAUDE.md warns about — reintroduced by CLAUDE.md's own instruction.

## Suggested fix

[...]

Issue #2078, opened 2026-07-06
agent-gate: fail closed on absent datasets — full gate must FAIL, not SKIP-then-PASS, when CQLITE_DATASETS_ROOT is empty


pmcfadin:
## Context
From the agentic-workflow audit (docs/reports/agentic-workflow-audit-2026-07-06.md, finding Q1 / quality-synthesis F6): in a worktree without the gitignored `*-Data.db` binaries and no `CQLITE_DATASETS_ROOT` override, the parity/smoke/compaction components SKIP (loud per-component) but the overall gate still returns **PASS**. A worker can paste a green `==== AGENT-GATE SUMMARY ====` block in which zero correctness was validated — the single most dangerous silent pass in the quality chain.

## Routing
**Direct implement (skip OpenSpec).** Behavior fully specified below; no design latitude. The gate self-test suite (`scripts/tests/`) is the oracle.

## Change
In `scripts/agent-gate.sh`: for the FULL gate (not `--lite`, not `--only`), when the resolved datasets root is unset/empty/missing Data.db fixtures, dataset-dependent components must record **FAIL** (with a one-line remedy: `bash test-data/scripts/fetch-datasets.sh` or point `CQLITE_DATASETS_ROOT` at the main checkout) instead of SKIP, making the overall gate FAIL. Provide an explicit opt-out env (`AGENT_GATE_ALLOW_MISSING_FIXTURES=1`) that restores SKIP and stamps the SUMMARY so the opt-out is visible in the pasted block (~15-line change).

## Acceptance criteria
- [ ] Full gate run with empty/unset datasets root → overall verdict FAIL; dataset components show FAIL with remedy line
- [ ] `--lite` and `--only` behavior unchanged (stay lenient)
- [ ] `AGENT_GATE_ALLOW_MISSING_FIXTURES=1` restores SKIP and the SUMMARY carries a visible marker (machine-checkable line)
- [ ] Self-test case added to `scripts/tests/test_agent_gate_summary.sh` (or sibling) covering FAIL-closed + opt-out marker
- [ ] Gate with datasets present is byte-identical in behavior (no regression)

Groomed by flow-lead from the 2026-07-06 audit (Tier 1, item 1).

PR #3236, opened 2026-08-03
fix(#3220): table-granular fixture resolution + per-case fail-closed guards so the #3032 differential case cannot skip silently

### LINKED ISSUE #1: Integrate swarm coordination system into main codebase
## Problem
The repository has extensive swarm coordination files (.swarm/, .claude/) and test frameworks that are not integrated into the main codebase structure.

## Current State
- .swarm/memory.db-* files are modified but not properly managed
- Multiple test frameworks exist in root directory (test-env/, testing-framework/)
- Docker integration files need proper organization
- Admin command demos and scripts are scattered

## Tasks
- [ ] Integrate .swarm/ coordination system properly
- [ ] Organize test frameworks under proper directory structure  
- [ ] Move demo scripts to examples/ or docs/
- [ ] Clean up root directory organization
- [ ] Document swarm coordination integration

## Priority
High - affects project organization and maintainability

## Component
Build System, Project Structure

### LINKED ISSUE #3220: point_vs_full_differential: the #3032 multi-component clustering case can skip silently — not covered by any fail-closed gate component
Split out of #3032 as the one `partial` in that issue's C intent audit. The auditor judged it **non-blocking**; filing it rather than re-running an 85-minute cold gate to close a narrow hole.

## The gap

#3032 added a fail-closed gate component `bti-multiclustering` (`scripts/agent-gate.sh`) that runs both new fixture lanes with `CQLITE_REQUIRE_FIXTURES=1` **unconditionally, with no SKIP branch**. That covers:

- `cqlite-core/tests/issue_3032_multiclustering_rows_trie_shape.rs`
- `cqlite-core/tests/issue_3032_multiclustering_clustering_slice_select.rs`

**It does not cover the AC6 differential case.** `cqlite-core/tests/point_vs_full_differential.rs` runs only under `core-tests` without `CQLITE_REQUIRE_FIXTURES`, and even with that variable set the harness asserts only `ran > 0` **across the whole corpus** — not that any specific case ran.

## Why that is a real (if narrow) hole

`sstables_root` (`point_vs_full_differential.rs:327-342`) selects a datasets root by **keyspace** presence. A root that contains `test_da` but *not* `multiclustering_table` therefore satisfies the harness, and the new multi-component clustering case **skips silently** while the suite reports green.

That is precisely the class of failure #3032 exists to eliminate — a dataset-dependent assertion that can pass without ever running. Every other lane added by #3032 is fail-closed; this one is not.

**Exposure is narrow**, which is why it did not block:
- the fixture is git-tracked, so it is present in any checkout;
- CI's `restore_ci_tracked_dataset_files` re-adds it under an in-repo root;
- the gate of record's staged corpus did contain it (the case genuinely ran).

So this is a **guard gap, not a known-failing test**.

## Related observation from the same audit

Fixture-root resolution is inconsistent between the two consumers: `query_semantics_oracle_parity` and `point_vs_full_differential` both select by keyspace, and a root holding `test_da` without `multiclustering_table` produces a **hard FAIL** in the former (correct, but with a confusing message) and a **silent skip** in the latter. Worth reconciling while here.

## Acceptance criteria

1. The AC6 multi-component clustering differential case cannot pass without running. Either add `point_vs_full_differential` to the `bti-multiclustering` gate component, or assert that the specific `test_da.multiclustering_table` case executed (not merely that some case did).
2. Demonstrated fail-closed: with the fixture absent or present-but-empty, the lane **FAILs** rather than skipping green. Show both directions.
3. `sstables_root` resolution reconciled between `point_vs_full_differential` and `query_semantics_oracle_parity`, or the divergence documented with a reason — selecting a root by keyspace when the requirement is a specific *table* is the underlying defect.
4. Full agent gate PASS.

## Note for the implementer

Touching `scripts/agent-gate.sh` means `--delta` re-certification **fails closed** (scripts are outside its allowed set), so this needs a full gate run. Batch it with other gate-touching work if any is pending.

## Related
- #3032 (added the fixture, the lanes, and the `bti-multiclustering` component)
- #1918 (the point-vs-full differential lane)
- #2078 (fail-closed missing-fixtures guard)

pmcfadin:
Closes #3220

## The defect, reproduced on a fleet box

`point_vs_full_differential` selected its datasets root by **keyspace** presence (`root.join(keyspace).is_dir()`), committed to that candidate with no fallback, then checked the table only against the already-chosen root. Its terminal assertion was suite-wide `assert!(ran > 0)`.

On `ip-172-31-5-109`, `CQLITE_DATASETS_ROOT=/data/datasets` holds `test_da/` but **not** the git-committed `test_da/multiclustering_table-*`. So the #3032 AC6 multi-component clustering case **skipped silently behind a green suite** — the exact class of failure #3032 exists to eliminate.

Measured asymmetry (this is why no preference ordering can fix it):

| root | `*-Data.db` | tables |
|---|--:|--:|
| `/data/datasets` (fleet, exported) | 144 | 122 |
| `<repo>/test-data/datasets` (checkout) | 31 | committed parity refs |

Overlap 30; only-in-fleet 92; **only-in-checkout 1 — `test_da/multiclustering_table`**. Neither root is a superset. "Prefer the exported root" silently skips the #3032 case; "prefer the checkout" fail-closes the corpus lanes (#3104). **Resolution must be by evidence, not preference.**

## The fix

- **Shared table-granular resolver** `cqlite-core/tests/support/datasets_root.rs::sstables_root_for_table` — walks *every* candidate root and selects the first that actually holds that table's `*-Data.db`. Consumed by `point_vs_full_differential.rs`, `one_vs_n_generation.rs`, `query_semantics_oracle_parity.rs`, `read_path_forcing_e2e.rs`. (AC3)
- **Per-case must-run guards, unconditional** — declarative `must_run` on `TableCase` *and* `GenerationCase`, replacing suite-wide `ran > 0` in all three lanes. All nine values audited against `git ls-files` as authority. (AC1b)
- **Gate coverage** — `bti-multiclustering` now also runs `point_vs_full_differential` and the AC2 self-test. (AC1a, applied *after* the resolver fix so it cannot red for the wrong reason)
- **`scripts/tests/test_point_vs_full_failclosed.sh`** — committed automated positive control, 11 assertions. (AC2)

## AC2 — fail-closed demonstrated in both directions

| direction | staging | result |
|---|---|---|
| control | untouched corpus | lane passes, `PASS test_da.multiclustering_table` observed |
| **absent** | checkout candidate mirrored minus exactly that fixture | `rc=101`, `did NOT run: ["multiclustering_table"]` |
| **present-but-empty** | 0-byte `da-2-bti-Data.db` in temp root | `rc=101` on the slice-count anchor, **no SKIP** |

Plus a safety assert that the tracked fixture is unmutated. An early non-discriminating staging (hiding the whole corpus) was rejected — it stayed green under a mutated `must_run`; the surgical version reds 2/11 when mutated.

**Before/after proof for the 1-vs-N lane:** with `test_da/wide_table-*` deleted in a throwaway worktree, `one_vs_n_generation_differential_equality` now FAILs naming `["wide_table"]`; the *same* staging at pre-fix `e8bf96d` reported `ok`. Silent skip reproduced, then closed.

**Gate-wiring proof:** a temporary `exit 1` in the self-test drove `--only bti-multiclustering` to `FAIL (12s)`; restored, `PASS (3s)`, `passed: 11 failed: 0`.

## Review findings resolved

`rust-reviewer` (4 blockers) + roborev (2 findings) on the pre-fix diff:

- **B1** suite-wide `ran > 0`/`pinned > 0` survived in the edited `one_vs_n_generation.rs` submodule — the same defect, same target, same new gate component. Fixed via a pure `must_run_violations` seam, failing branch proven for both lanes.
- **B2** `test_tomb.static_with_tombstones` had `must_run: false` despite committed binaries. Fixed; other four `test_tomb` cases ship sidecars only and correctly stay `false`.
- **B3 / roborev-medium** (found independently by both reviewers) plain `env` does not unset an inherited `CQLITE_REQUIRE_FIXTURES`, so the "deliberately without require-fixtures" invocation was not. Now `env -u`.
- **B4** `read_path_forcing_e2e` still SKIPped unless require-fixtures, for a fixture it calls committed. Now panics with searched roots + remedy.
- **roborev-low, escalated to blocker** the AC2 self-test existed but the component never ran it — a positive control that cannot fail is the very pattern this PR removes. Now wired and proven to red.
- Nits N5 (unquoted passthrough → quoted array), N6, N7 (pins ungated so plain `cargo test -p cqlite-core` builds them) folded in.

## Scope boundaries (deliberate, not omissions)

- **#3104 — punted, narrowed, documented.** The resolver dissolves its suggested fix #1 for these lanes (evidence beats preference; see the table above), but does not reach its `#2078` whole-corpus preflight, the count-naming diagnostic, the `--lite` small-corpus warning, or the doctrine text. Recorded in CLAUDE.md + the validation playbook and commented on #3104.
- **`cqlite-flight/tests/query_semantics_flight_parity.rs` is a fourth lane with the same keyspace-granular resolver, left unreconciled** — different crate (needs a dev-dependency test-support crate), *generation*-granular `fixture_dir` vs the helper's table-granular match, and a `..`-climbing `schema_path` (the #3148 trap). Lower risk: its component pins `CQLITE_REQUIRE_FIXTURES=1`, so a miss FAILs rather than skipping. Named in the module doc; follow-up filed.

## Verification

- `--lite` **PASS** at `e4f067d`, `tree-integrity: PASS`.
- `--only bti-multiclustering` **PASS**.
- Full `agent-gate.sh` required and run by the closer — the diff touches `scripts/agent-gate.sh`, so `--delta` fails closed and is unavailable.

Doctrine updated in the same change: CLAUDE.md (resolve-per-table, assert-per-case) and `agents-developing/validation-playbook`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -247,6 +247,13 @@ by responsibility (source: epic #1116; tests: #1135). Genuinely out of scope →
   root (env, then checkout) for that table's `*-Data.db`. And never terminate a corpus loop with a
   suite-wide `assert!(ran > 0)`: it cannot see one case skipping behind its siblings — assert per
   case (committed fixtures = `must_run`, fail-closed unconditionally).
+  Resolve by EVIDENCE, never by a preference ordering: neither root is a superset — a fleet
+  `/data/datasets` measured 144 `*-Data.db` over 122 tables yet lacks the one committed
+  `test_da/multiclustering_table`, which the checkout's 31 parity references carry — so *any* fixed
+  env-first/checkout-first rule picks wrong for one set of tables. That dissolves #3104's "prefer
+  the already-exported root" fix for the lanes on this resolver; **#3104 stays open** for what the
+  resolver does not reach (whole-corpus `#2078` preflight, count-naming diagnostics, `--lite`
+  small-corpus warning, and the doctrine text still telling agents to override the exported root).
 - **Two parity oracles (issue #1742)**: *physical-dump parity* (the `*-Data.db.jsonl` sstabledump
   goldens) enumerates every on-disk cell INCLUDING tombstones/deleted/expired-TTL rows, so it CANNOT
   catch a read-time-reconciliation bug (both sides keep the shadowed rows → green while a real
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -239,6 +239,14 @@ by responsibility (source: epic #1116; tests: #1135). Genuinely out of scope →
   reference files —
   [validation playbook](https://pmcfadin.github.io/cqlite/agents-developing/validation-playbook/)
 - Never let a dataset-dependent test pass on an empty dataset (0-rows-when-present = failure)
+- **Resolve fixture roots per TABLE, and assert per CASE (issue #3220)**: a lane that picks its
+  corpus root by KEYSPACE (`root.join(keyspace).is_dir()`) and commits to it can pass without ever
+  running — a `CQLITE_DATASETS_ROOT` holding `test_da/` but not the git-committed
+  `test_da/multiclustering_table-*` made the #3032 case skip silently behind a green suite. Use
+  `cqlite-core/tests/support/datasets_root.rs::sstables_root_for_table`, which walks EVERY candidate
+  root (env, then checkout) for that table's `*-Data.db`. And never terminate a corpus loop with a
+  suite-wide `assert!(ran > 0)`: it cannot see one case skipping behind its siblings — assert per
+  case (committed fixtures = `must_run`, fail-closed unconditionally).
 - **Two parity oracles (issue #1742)**: *physical-dump parity* (the `*-Data.db.jsonl` sstabledump
   goldens) enumerates every on-disk cell INCLUDING tombstones/deleted/expired-TTL rows, so it CANNOT
   catch a read-time-reconciliation bug (both sides keep the shadowed rows → green while a real
--- a/cqlite-core/tests/issue_3220_datasets_root_resolution.rs
+++ b/cqlite-core/tests/issue_3220_datasets_root_resolution.rs
@@ -1,0 +1,179 @@
+//! Issue #3220: the TABLE-granular fixture-root selection rule, pinned against
+//! synthetic roots.
+//!
+//! # What defect this exists to catch
+//!
+//! Three dataset lanes selected a corpus root by **keyspace** (`root.join(keyspace)
+//! .is_dir()`) and then committed to it with no fallback, even though every caller
+//! needs a specific **table**. On a machine whose `CQLITE_DATASETS_ROOT` corpus held
+//! `test_da/` but not the git-committed `test_da/multiclustering_table-*`, that rule
+//! selected the env root, missed the table, and the #3032 multi-component clustering
+//! differential case SKIPPED SILENTLY behind a green suite.
+//!
+//! # Why the assertions are against SYNTHETIC roots
+//!
+//! The real candidate list is half environment and half a COMPILE-TIME checkout path,
+//! so a test that reads it can only ever observe THIS machine's layout — and on a
+//! machine where every root happens to hold every table, the broken keyspace-granular
+//! rule and the correct table-granular one return the same answer. Only an explicit
+//! two-root layout where the FIRST root holds the keyspace WITHOUT the table
+//! discriminates between them, which is exactly the layout built below. Hence the
+//! pure `first_root_with_table` seam.
+//!
+//! These are directory-shape assertions, not format assertions: no SSTable is parsed
+//! here, so an empty file named `*-Data.db` is a legitimate stand-in for "the binaries
+//! are present" (whether they DECODE is the lanes' own business, and a truncated one
+//! makes them fail loudly — the second AC2 direction, exercised end-to-end in the PR).
+
+#![cfg(all(feature = "state_machine", feature = "cli-helpers"))]
+
+#[path = "support/datasets_root.rs"]
+mod datasets_root;
+
+use datasets_root::{first_root_with_table, table_has_data};
+use std::path::{Path, PathBuf};
+
+/// Create `<root>/<keyspace>/<table>-<uuid>/` and populate it with `files`.
+fn make_table_dir(root: &Path, keyspace: &str, table: &str, uuid: &str, files: &[&str]) -> PathBuf {
+    let dir = root.join(keyspace).join(format!("{table}-{uuid}"));
+    std::fs::create_dir_all(&dir).expect("create fixture dir");
+    for f in files {
+        std::fs::write(dir.join(f), b"").expect("write fixture component");
+    }
+    dir
+}
+
+/// THE #3220 case: candidate 1 holds the KEYSPACE but not the TABLE. The rule must
+/// fall through to candidate 2, which actually carries the table's `*-Data.db`.
+///
+/// Under the retired keyspace-granular rule candidate 1 wins and the table is then
+/// reported absent — the silent-skip defect.
+#[test]
+fn a_root_holding_the_keyspace_without_the_table_is_skipped() {
+    let tmp = tempfile::tempdir().expect("tempdir");
+    let env_root = tmp.path().join("env/sstables");
+    let checkout_root = tmp.path().join("checkout/sstables");
+
+    // Candidate 1: the keyspace exists (another table lives there), the wanted table
+    // does not — the real /data/datasets shape that triggered #3220.
+    make_table_dir(
+        &env_root,
+        "test_da",
+        "wide_table",
+        "aaaa",
+        &["da-2-bti-Data.db"],
+    );
+    // Candidate 2: the committed fixture.
+    make_table_dir(
+        &checkout_root,
+        "test_da",
+        "multiclustering_table",
+        "bbbb",
+        &["da-2-bti-Data.db"],
+    );
+
+    let roots = vec![env_root.clone(), checkout_root.clone()];
+    assert!(
+        env_root.join("test_da").is_dir(),
+        "precondition: candidate 1 must hold the keyspace, else this test cannot \
+         discriminate the keyspace-granular rule from the table-granular one"
+    );
+    assert_eq!(
+        first_root_with_table(&roots, "test_da", "multiclustering_table"),
+        Some(checkout_root.as_path()),
+        "resolution must fall through a root that holds the keyspace but not the table"
+    );
+    // The sibling table still resolves from the FIRST root: preference order is
+    // preserved, the rule only skips roots that cannot serve the request.
+    assert_eq!(
+        first_root_with_table(&roots, "test_da", "wide_table"),
+        Some(env_root.as_path()),
+        "candidate order must still win when the first root does hold the table"
+    );
+}
+
+/// Fixture-absent direction of the fail-closed contract (AC2a): when NO candidate
+/// carries the table, resolution yields `None` — the value every caller turns into a
+/// hard failure (`must_run` / `CQLITE_REQUIRE_FIXTURES`), never a silent success.
+#[test]
+fn absent_everywhere_resolves_to_none() {
+    let tmp = tempfile::tempdir().expect("tempdir");
+    let root_a = tmp.path().join("a/sstables");
+    let root_b = tmp.path().join("b/sstables");
+    make_table_dir(
+        &root_a,
+        "test_da",
+        "wide_table",
+        "aaaa",
+        &["da-2-bti-Data.db"],
+    );
+    std::fs::create_dir_all(root_b.join("test_da")).expect("empty keyspace dir");
+
+    let roots = vec![root_a, root_b];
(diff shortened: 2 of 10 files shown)


pmcfadin:
**Follow-ups filed** (deliberately out of scope for this PR, per the blocker/nit triage in `docs/development/roborev-severity.md`):

- **#3237** (P2, real defect) — `agent-gate.sh` byte-parity **"Group B"** has the *same* `env`-inheritance bug this PR fixes at `bti-multiclustering`. Plain `env` does not unset an inherited `CQLITE_REQUIRE_FIXTURES`. Fails safe (spurious red, never spurious green), hence P2 — but the issue asks for a **sweep** of the whole script, since this PR fixed one instance and review immediately found a second.
- **#3238** (P3, batched nits) — the fourth unreconciled lane (`cqlite-flight/tests/query_semantics_flight_parity.rs`, with the three reasons it is not a mechanical swap), plus `test_tomb.skipped_partition_delete` being the one corpus entry where "counted as ran" does not imply "asserted on real rows".
- **#3104** — commented with the measured root asymmetry; kept open, scope narrowed to the preflight/diagnostics/doctrine half this PR does not reach.

pmcfadin:
## Round 3 — roborev finding fixed, plus 3 more found by auditing for the same shape

roborev (job 21) flagged the **empty-fixture** positive control at `test_point_vs_full_failclosed.sh:202`: it accepted any occurrence of `multiclustering_table`, **including the `PASS test_da.multiclustering_table` line it exists to rule out**. Escalated to blocker — a check that cannot fail is the subject of this issue, and this one sat inside the mandatory AC2 control.

**Fixed:** the empty direction now rejects `^PASS test_da\.multiclustering_table ` outright *and* requires the rejection on one line as `case test_da.multiclustering_table: … must yield exactly N` (or `… no partition keys to probe`) — the case's own anti-vacuous anchor, not a nonzero exit and not a bare table mention.

**Discrimination proof:** commented out the staging truncation so the healthy fixture was silently kept → `passed: 9 failed: 3`, both new empty checks FAILing. The *old* regex matched exactly the success line `PASS test_da.multiclustering_table — 3 point queries…` and would have reported ok. Restored; unmutated run `passed: 12 failed: 0`. `/data/datasets` and all tracked fixtures untouched.

**Auditing the other 11 assertions for the same weakness found three more:**
1. the **control** and **absent** ran-line greps were unanchored, so the must-run assertion's own `ran:`/`skipped:` list lines could satisfy them → now `^PASS …`;
2. the safety check discarded `git status`'s exit code under `2>/dev/null`, reading an *unanswerable* git query as "fixture untouched" → now rc-checked, plus an independent non-empty check on the tracked `Data.db`;
3. added an `EXPECTED_CHECKS=12` terminal anchor so a dropped or short-circuited case cannot report `failed: 0`.

Judged sound and left alone (with reasons): the absent must-run grep demands the exact one-element list `["multiclustering_table"]`; the per-table `no *-Data.db … under any candidate` diagnostic is emitted only by the resolver for that table; the staging predicate asserts filesystem facts directly; the two coarse `rc != 0` checks are each now paired with a target-specific message check and catch compile/harness breakage a message check cannot.

`--lite` **PASS** at `05e1635`, `tree-integrity: PASS`. `--only bti-multiclustering` **PASS (3s)**, self-test 12/12 inside it.

This is the third nesting of one shape in this change: a *case* that can pass without running → a *positive control* that stayed green under a mutated `must_run` → an *assertion* satisfiable by the success line it rules out. Worth noting all three were green under `--lite` and would have been green under the full gate.

pmcfadin:
## C intent audit: **PASS** (AC4 `pending-closer`)

Oracle-routed, so `spec-auditor` was anchored to the issue's AC1-AC4 **plus the owner's two added constraints**, and run adversarially — instructed to construct a defeating mutation for every new guard, with any defeatable guard counting as `unmet`.

| requirement | verdict | key evidence |
|---|---|---|
| **AC1** case cannot pass without running | satisfied | `(b)`-then-`(a)` ordering confirmed by commit sequence `26c0d98` → `075b5a4` → `a4a3938`, not inverted. Assertion is case-specific: flipping `must_run` to `false` reds a pin that asserts the violation list **by name** |
| **AC2** both-directions demo | satisfied | script committed **and** run by the gate as the third `&&` link of `run_bti_multiclustering`; empty-direction root wins resolution so the healthy copy cannot be substituted |
| **AC3** resolution reconciled | satisfied | all three consumers converted; no keyspace-granular body survives in `cqlite-core`; rule pinned against **synthetic** roots — the only layout that discriminates old rule from new |
| **AC4** full gate PASS | pending-closer | in flight |
| owner (a) mandatory demo | satisfied | see AC2 |
| owner (b) #3104 punted with reason | satisfied | written in **three durable places** (CLAUDE.md, validation playbook, comment on #3104), not merely asserted in this PR body |

**Scope drift: none** — all 10 files trace to an AC or the mandated doctrine update.

Two auditor mutation attempts worth recording, both of which failed to defeat the guards: flipping `must_run` to `false` (reds a by-name pin), and weakening the violation loop (the pure seam is unit-proven in both directions plus a non-vacuity assert). `Ok(true)` is reachable only after real point-vs-full comparisons, since `run_case` returns `Err` on zero probe keys — so "ran" cannot be faked.

### Correction to this PR's earlier evidence
The round-2 comment quoted the self-test at `passed: 11 failed: 0`. That predates `05e1635`, which added the 12th assertion and the empty-PASS rejection; the current count is **12**. The current script's green is established by the in-flight gate of record, where `bti-multiclustering: PASS` is the line to confirm.

### Residual nits → appended to #3238 (non-blocking)
Chiefly: **`must_run` is hand-derived from `git ls-files` with no automated cross-check**, so a future committed fixture added with `must_run: false` reintroduces the silent-skip class through the one seam this fix left manual. A test asserting flag-set == tracked-binaries-set would mechanize it. Also the in-tree flight punt omits its `#3238` citation, and a four-vs-five miscount.

Not applied here: the gate of record was already running against this tree, and editing a worktree mid-gate trips `tree-integrity` (#2926).

pmcfadin:
## Closer: merge-time record (#3220)

**Gate of record** — full \`scripts/agent-gate.sh\`, ONE run, summary \`/tmp/gate-3220-FULL.txt\`, run-id \`/tmp/agent-gate.x4bhtP\`:
\`RESULT: PASS\` · \`commit: 05e1635\` · \`tree-integrity: PASS\` · 31/31 components, no SKIP, no FAIL.
Includes **\`bti-multiclustering: PASS (3s)\`**, so the AC2 both-directions fail-closed self-test genuinely executed inside the gate of record.

**`query-semantics-oracle: PASS (0s)` adjudicated (not a vacuous green).** The 0s is warm-cache wall-clock granularity (`core-tests` at 586s pre-built everything): the component log shows `running 6 tests` / `test result: ok. 6 passed` in 0.07s with zero `Compiling` lines. The lane runs under `CQLITE_REQUIRE_FIXTURES=1` (`scripts/agent-gate.sh:4906`), where `query_semantics_oracle_parity.rs:672-685` asserts `skipped.is_empty()` **and** `ran == oracle.cases.len()` unconditionally. The oracle defines 8 cases including `multiclustering_bti__multi_component_slice` and `multiclustering_bti__first_component_prefix_slice` — so the green *entails* both multiclustering cases ran.

**C intent audit: PASS.** AC1-AC4 satisfied plus both owner constraints (mandatory both-directions fail-closed demo; #3104 tension written down, narrowed and left open with a pointer). No scope drift.

**Final roborev (job 22): terminal `RESULT: FAIL` — recorded honestly, triaged as nit-only.**
The FAIL is driven solely by `findings: PRESENT (1)` / `roborev-exit: FINDINGS (exit 1)`. Every integrity check PASSED — `push-assert`, `census-check`, `code-free`, `job-record`, `sha-assert`, `review-completed`, `prompt-content (8/8 code census paths)`, `vacuity-tier1`, `vacuity-tier2` — with `reviewed-sha: 948ca25..05e1635` (a true range ending at the certified SHA) and a 204.5k-in / 171.8k-cached / 7.6k-out profile over 4m13s. This is a genuine, complete review, not the vacuity class the block exists to catch.

The single finding is a **comment-prose contradiction** at `cqlite-core/tests/point_vs_full_differential.rs:188` ("exactly four" above a list of five). Zero behavioral impact; not a guard, code path, or assertion. Independently found by the C audit and already filed as item 5 on #3238. Per `docs/development/roborev-severity.md`, nits batch into the linked follow-up at merge time and never trigger a re-verify round — fixing a one-word comment here would invalidate a 25-minute certification (`--delta` is unavailable, the diff touches `scripts/agent-gate.sh`).

**Deliberately NOT reported as "roborev clean"** — the terminal RESULT was FAIL and is recorded as such.

**Pre-merge guards:** `premerge-assert.sh 3236 05e1635c0a5b…` → `PREMERGE: OK`. Fresh `HOLD:` re-read on #3220 and #3236 → none.

Arming `--auto --squash --delete-branch`. Follow-ups: #3237 (P2), #3238 (P3, carries this nit); #3104 narrowed and left open.

Issue #3108, opened 2026-07-28
parse_block_entries_at_now silently drops now_secs and read_shadowing on the V5UncompressedOA state-machine route


pmcfadin:
## Summary

`SSTableReader::parse_block_entries_at_now` (`storage/sstable/reader/parsing/block_entries.rs`)
threads its `now_secs` (caller-pinned read-time TTL clock) and honours `read_shadowing` only on the
`V5CompressedLegacy` branch. When the header's `DataFormat` is `V5UncompressedOA` (BTI `da`), the
function routes to `parse_block_entries_with_state_machine`, which takes **neither** parameter — so
both are **silently dropped**: a caller that asked for SELECT-semantic shadowing, or pinned the
read-time clock, gets an unshadowed / wall-clock read with no signal.

Discovered during #3058. It is unreachable from #3058's single-source bypass **today** only because
`SSTableReader::supports_streaming_query_scan()` refuses BTI readers — an implicit, undocumented
dependency rather than a guarantee. A comment at the site records this.

## Constraint on the fix (learned the hard way in #3058)

A naive fail-closed guard on `read_shadowing || now_secs.is_some()` **breaks pre-existing BTI
batched scans**: `run_scan_stream_batched` reaches this function with `read_shadowing = true` for BTI
readers (see the companion issue for the missing BTI dispatch), so refusing those made a `da` table's
whole scan return `UnsupportedFormat`. That guard was reverted. Any fix must either
(a) route BTI readers to a decoder that honours both parameters, or
(b) scope a guard to the genuinely new parameter (`now_secs.is_some()`) only.

## Acceptance

1. A `test_da`/BTI scan requesting read shadowing applies it (or is explicitly, documentedly
   refused), with a test.
2. A caller-pinned `now_secs` is either honoured on that route or refused — never dropped silently.
3. Pre-existing `read_shadowing = true` BTI batched scans (`SSTableManager::scan_stream_batched` →
   `select_executor/streaming.rs`, `stream_agg.rs`, `query/executor.rs`) still return rows.

Routing: oracle-driven — issue + pinned test, no OpenSpec change.

Issue #1577, opened 2026-07-02
D1 — push LIMIT/OFFSET into the materializing scan (stop decoding 1M rows for LIMIT 10)


pmcfadin:
Child **D1** of Epic D (#1516) — 2026-07 read-path audit (block 1). Source of truth: `docs/reports/read-path-performance-audit-2026-07-01.md` §Epic D. Anchors as of `main` @ `5c080d2a` — re-locate with `rg` if drifted. Capstone wave: **Wave 3**, streaming-by-default story (D1/D2/D3/D6 → D4 → D5).

## Problem
`SELECT ... LIMIT 10` on a 1M-row table decodes ALL 1,000,000 rows into per-row HashMaps, then throws away 999,990. The reader's `scan()` already accepts a `limit` parameter — the executor always passes `None` (`execute.rs:783-785`) and truncates post-hoc (`mod.rs:721-735`).

## Fix (step by step)
1. In the materializing executor, when the plan after the scan contains **no Sort and no Aggregate** (i.e. nothing that needs the full row set), stop the scan loop as soon as `limit + offset` rows have been *accepted* (post-predicate).
2. When additionally there are **no residual predicates** (every row the scan yields is accepted), pass the limit down: `storage.scan(..., Some(limit + offset))` so the reader itself stops early.
3. Semantics guardrails — push down ONLY when safe:
   - `ORDER BY` present → no pushdown (needs all rows) unless the scan order already satisfies it (do not attempt that optimization here).
   - Aggregates/GROUP BY → no pushdown (D2's territory).
   - Predicates present → count accepted rows, not scanned rows.
   - `OFFSET` → fetch `limit + offset` then skip (see also E8's skip/take cleanup).
4. `rg` for all call sites of the truncation at `mod.rs:721-735` to make sure DISTINCT/other constructs don't silently rely on over-fetch.

## TDD tests (write FIRST, must fail on current main)
- Work counter: `LIMIT 10` on a 10k-row fixture scans ≤ 10 + small slack rows (rows-decoded counter via A5 pattern). Fails today (scans 10k).
- Latency: `LIMIT 10` flat across {1k, 100k} row fixtures (bench).
- Correctness matrix (must PASS before and after): LIMIT with predicate (accepted-row counting), LIMIT+OFFSET (exact rows, compare to full-scan oracle), LIMIT+ORDER BY (unchanged full sort, same results), LIMIT larger than table.

## Guardrails
[...]

PR #3239, opened 2026-08-03
fix(#3109): give the batched + sequential scan surfaces the BTI dispatch their siblings have

### LINKED ISSUE #3109: run_scan_stream_batched lacks the BTI dispatch that run_scan_stream/sequential_scan have, so da tables stream through the state-machine decoder unshadowed (#1577 class)
## Summary

`SSTableReader::run_scan_stream_batched` (`storage/sstable/reader/data_access/sequential.rs`) has **no
`bti_partitions_db.is_some()` dispatch**, unlike its two siblings:

* `run_scan_stream` (`sequential.rs`, the per-row surface) dispatches BTI readers away from the
  block loop;
* `sequential_scan` (`sequential.rs`) does the same.

So for a BTI (`da`) reader the batched surface falls straight into the non-stitching block loop and
decodes through `parse_block_entries_*` → the state-machine decoder, which honours neither
`read_shadowing` nor a caller-pinned read clock (companion issue). Net effect: **a `da` table
streamed through the batched surface is read UNSHADOWED** — partition/range tombstones and TTL expiry
are not applied at read time — where the per-row surface applies them. This is the #1577 class
(per-surface decode-posture divergence).

Reached in production by `SSTableManager::scan_stream_batched` (`storage/sstable/mod.rs`), which the
core streaming executors use (`select_executor/streaming.rs`, `stream_agg.rs`,
`query/executor.rs`).

Found while reviewing #3058; pre-existing on `main` and deliberately NOT fixed there (out of scope —
#3058 only added the `now_secs` parameter to that call site).

## Repro sketch

Take a `test_da` fixture containing a tombstoned/expired row, read it through
`SSTableManager::scan_stream_batched`, and compare against the per-row `scan_stream` and against
`query-semantics-oracle.json`: the batched surface should surface rows the semantic oracle hides.

## Acceptance

1. A `test_da` batched-scan test proves read shadowing IS applied (tombstoned/expired rows absent),
   matching the per-row surface row-for-row.
2. The BTI dispatch is present on all three surfaces, or the batched surface documents an explicit,
   tested reason it differs.
3. `query-semantics-oracle` covers a BTI table through the batched surface.

Routing: oracle-driven — issue + pinned test, no OpenSpec change.

pmcfadin:
Closes #3109.

## Problem

`SSTableReader::run_scan_stream_batched` had no `bti_partitions_db.is_some()` dispatch, unlike `run_scan_stream` and `sequential_scan`. A BTI (`da`) reader therefore fell straight into the non-stitching block loop and decoded through the state-machine decoder, which honours neither `read_shadowing` nor a caller-pinned read clock — so a `da` table streamed through the batched surface was read **unshadowed** (partition/range tombstones and TTL expiry not applied) where the per-row surface applied them. Reached in production via `SSTableManager::scan_stream_batched`, used by the core streaming executors. This is the #1577 class: per-surface decode-posture divergence.

## Fix

Rather than add a third divergent copy of the dispatch, the change introduces **one shared `stream_bti_scan` helper** (`data_access/bti.rs`) serving both streaming surfaces (per-row and batched emit shapes), and applies the same gate to `sequential_scan` — whose one un-intercepted caller, `iterate_all_partitions`, was also decoding `da` through the state machine and failing outright. The batched surface moved to a new `data_access/batched_scan_stream.rs` (campsite split; `sequential.rs` 1486 → 1319 lines).

## Acceptance criteria

1. **Batched-scan test proves read shadowing is applied** — `cqlite-core/tests/issue_3109_bti_batched_scan_shadowing.rs`, against the real Cassandra 5.0.2 `test_da.ttl_table` `da` fixture, pinning `now` before its golden TTL expiry (2 live rows) and after (0), asserting row-for-row equality with the per-row surface. Verified red pre-fix.
2. **Dispatch present on all surfaces** — the same test drives all four (`scan`, per-row stream, batched stream, `iterate_all_partitions`) at both pins and asserts they agree, so a fifth surface must join or fail.
3. **`query-semantics-oracle` covers a BTI table through the batched surface** — two new cases (`ttl_table_bti__full_scan_live` / `__full_scan_expired`); the harness now drives every case through `execute_streaming`, the only lane reaching `scan_stream_batched`. With the dispatch disabled, both cases fail on that lane only.

## Review-first (per #2086)

`rust-reviewer` + roborev (`gpt-5.6-sol`) ran on the lite-green diff before any full gate, returning two blockers, both fixed and each verified red/green:

- **`scan_cancel` contract dropped by the new BTI early-return** — a cancelled `iterate_all_partitions_cancellable` walk would stitch the whole Data.db and return `Ok(every row)`. Added `bti_scan_with_metadata_cancellable` (mirroring the existing `stitch_all_chunks`/`_cancellable` pair): checks on entry, stitches via `stitch_all_chunks_cancellable`, polls every 256 entries. Proof: `bti_sequential_scan_honors_per_call_cancel`, using a token deliberately distinct from `reader.scan_cancel` (a test on the reader's own field would pass against the broken code).
- **Wall-clock time bomb in the pinned test** — it pinned `now` via the `debug_assertions`-only `CQLITE_TTL_NOW_OVERRIDE_SECS` env seam, so a release test build would silently fall back to the real clock and fail once the fixture expired. The env seam is gone: the test now pins through the explicit, non-debug-gated `now_secs` parameter of the public `SSTableReader::open_query_row_stream`, deterministic in both profiles.

Also folded in: golden counts now count `"type": "row"` rather than JSONL lines (partitions); the in-repo `test-data/datasets` fallback so a plain `cargo test` cannot silently skip the only direct regression test; `parsing/block_entries.rs` no longer overclaims — it now names `scan_for_key` → `verify_presence_oracle_negative` as the known remaining hole (relevant to #3108, which was planning a fail-closed guard on the strength of that comment); and the `table_ids_match` behavior delta on the BTI branch is documented.

Remaining nits are batched to one follow-up issue at merge time, per `docs/development/roborev-severity.md`.

Routing: **oracle-driven** — issue + pinned test, no OpenSpec change.

## Audit notes (`spec-auditor`, C intent audit — OVERALL: PASS)

All three of #3109's acceptance criteria came back `satisfied` with public-surface tests as evidence.
Two honest scope statements from that audit — non-blocking, recorded for the record:

1. **Only TTL expiry is directly proven.** No `da` fixture with partition/range tombstones exists, so
   the `ttl_table` leg of the four-surface agreement test is 0-vs-0 today. Tracked as item (d) of the
   batched follow-up issue.
2. **AC1's pinned-clock lane is a real `pub` API but not the production Flight path for `da`.**
   Flight's `bypass.rs:270` gates `open_query_row_stream` behind `supports_streaming_query_scan()`,
   which excludes BTI readers. Production reachability of this fix therefore rests on AC2's
   `SSTableManager::scan_stream_batched` leg and AC3's executor lane, both of which are covered.

## Gate of record

Full `AGENT-GATE SUMMARY` **RESULT: PASS** at `91340e6` (run-id `/tmp/agent-gate.CnqH8I`,
`tree-integrity: PASS`), every component PASS. roborev job 13 (`--agent codex --model gpt-5.6-sol`)
**RESULT: PASS**, `findings: NONE`, at the same unchanged SHA.

### DIFF
--- a/cqlite-core/src/storage/sstable/mod.rs
+++ b/cqlite-core/src/storage/sstable/mod.rs
@@ -2019,7 +2019,10 @@ impl SSTableManager {
     /// * any BTI (`da`) reader: `run_scan_stream`'s BTI branch (issue #1577) drives
     ///   the trie-walk `bti_scan_with_metadata`, which fully materializes the
     ///   (index-less) reconciled table before streaming — so a bounded consumer
-    ///   never decode-stops for a BTI table, regardless of generation count;
+    ///   never decode-stops for a BTI table, regardless of generation count. Since
+    ///   issue #3109 the BATCHED surface takes that same shared dispatch
+    ///   (`stream_bti_scan`), so this report is now exactly true of both streaming
+    ///   surfaces rather than only the per-row one;
     /// * the default (non-`tombstones`) build's `write-support` cross-generation
     ///   branch is LAZY since #1579 (streams via
     ///   `generation_merge::stream_generations_for_read`), so a bounded consumer
--- a/cqlite-core/src/storage/sstable/mod.rs
+++ b/cqlite-core/src/storage/sstable/mod.rs
@@ -2016,13 +2016,10 @@ impl SSTableManager {
     /// takes, so the two can never disagree:
     /// * the `tombstones` build's `scan_stream` delegates wholesale to the
     ///   materializing [`scan`](Self::scan) — always `true`;
-    /// * any BTI (`da`) reader: `run_scan_stream`'s BTI branch (issue #1577) drives
-    ///   the trie-walk `bti_scan_with_metadata`, which fully materializes the
-    ///   (index-less) reconciled table before streaming — so a bounded consumer
-    ///   never decode-stops for a BTI table, regardless of generation count. Since
-    ///   issue #3109 the BATCHED surface takes that same shared dispatch
-    ///   (`stream_bti_scan`), so this report is now exactly true of both streaming
-    ///   surfaces rather than only the per-row one;
+    /// * any BTI (`da`) reader: BOTH streaming surfaces' BTI branch (#1577, shared
+    ///   as `stream_bti_scan` by #3109) drives the trie-walk
+    ///   `bti_scan_with_metadata`, which fully materializes the (index-less)
+    ///   reconciled table first — a bounded consumer never decode-stops for BTI;
     /// * the default (non-`tombstones`) build's `write-support` cross-generation
     ///   branch is LAZY since #1579 (streams via
     ///   `generation_merge::stream_generations_for_read`), so a bounded consumer
--- a/cqlite-core/src/storage/sstable/reader/data_access/batched_scan_stream.rs
+++ b/cqlite-core/src/storage/sstable/reader/data_access/batched_scan_stream.rs
@@ -1,0 +1,257 @@
+//! The BATCHED streaming scan (`scan_stream_batched`, issue #1592 Epic F/F2) — the
+//! additive companion to the per-row `scan_stream` whose channel item is a `Vec`
+//! BATCH of `(RowKey, ScanRow)` entries rather than a single entry.
+//!
+//! Split out of `sequential.rs` (campsite rule, epic #1116) alongside its per-row
+//! sibling `per_row_scan_stream.rs`: that file is well over the ~800-line source
+//! target, and the batched streaming scan is the same self-contained responsibility
+//! — one spawned producer task per reader, feeding a bounded channel, with the
+//! format branch (BTI trie walk / windowed stitch / block-by-block) chosen inside it.
+//!
+//! # Issue #3109: the BTI (`da`) dispatch
+//!
+//! This surface shipped WITHOUT the `bti_partitions_db.is_some()` dispatch that
+//! `scan` and `run_scan_stream` have, so a `da` reader fell straight into the
+//! non-stitching block loop and decoded through the `V5UncompressedOA` state
+//! machine — which honours neither `read_shadowing` nor a caller-pinned read clock.
+//! A BTI table streamed here was therefore read UNSHADOWED. Both streaming surfaces
+//! now share ONE dispatch, [`SSTableReader::stream_bti_scan`], so a third divergent
+//! copy of this decision cannot drift again (the #1577 class).
+//!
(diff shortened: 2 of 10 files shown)


pmcfadin:
## Certification record

| Check | Result |
|---|---|
| Full `agent-gate.sh` (run of record) | **RESULT: PASS** — run-id `/tmp/agent-gate.CnqH8I`, commit `91340e6`, `dirty: no`, `tree-integrity: PASS`, all 31 components PASS |
| C intent audit (`spec-auditor`, anchored to #3109's ACs) | **PASS** — AC1/AC2/AC3 all `satisfied` with public-surface wiring evidence |
| roborev (`--agent codex --model gpt-5.6-sol`, job 13) | **PASS** — `findings: NONE`, `sha-assert: PASS`, `vacuity-tier1/2: PASS`, `prompt-content: PASS (10/10)` |
| `premerge-assert.sh 3239 91340e6` | **OK** |
| `HOLD:` re-read (issue + PR) | none |

Routing is oracle-driven, so there is no OpenSpec change to archive.

### Gate history — three failures, none of them this diff
1. `missing-fixtures: FAIL-CLOSED (#2078)` — pinned corpus absent; remedied with `fetch-datasets.sh`.
2. A gate reaped mid-run by the exit of the subagent that backgrounded it (non-`setsid`), leaving its summary on the `INCOMPLETE` liveness placeholder.
3. `summary-integrity: FAIL` (#2874) — two runs contending on one summary path. The guard worked exactly as designed: the losing run refused to overwrite the live peer's summary and published to a `.integrity-fail.<run-id>` sibling.

### Cited pre-existing failure, since resolved
The first completed run (`2SuadE`) came back `RESULT: FAIL` on one component, `query-semantics-oracle (0s)`. Triage proved it **pre-existing on `origin/main`**: a throwaway detached worktree at `948ca25` reproduced it identically (`5 passed; 1 failed`). Cause was environmental — the shared root checkout was stale at `221d16d`, so the fixture `test_da/multiclustering_table-fd74ad50…` committed by #3032 was absent from `CQLITE_DATASETS_ROOT`, hard-failing the two pre-existing `multiclustering_bti__*` cases under `CQLITE_REQUIRE_FIXTURES=1`. This PR's two new cases passed throughout, non-vacuously. Fixed by fast-forwarding the root checkout; the durable fix (fixture-granular rather than keyspace-granular root selection in the oracle harness) is filed separately.

Arming `--auto`; GitHub lands this when `required` goes green.

Issue #3217, opened 2026-08-01
Full-box C(N) curve + on/off-CPU flame-graph attribution of Flight do_get handoff overhead (local box, measurement-only)


pmcfadin:
## Summary

#3100 (WS0 re-run + C(N) sweep, PR #3216) proved the Flight `do_get` concurrency ceiling is **CPU/handoff-bound, not disk-bound**: on a single pinned physical core, aggregate throughput peaks at N=2 (1.16×) then *declines* to 0.96× at N=16, with 38.3M context switches over the sweep, while `%util` on the device **falls** 14% → 3%. But that conclusion is an inference from counters (`perf stat`), not from stacks — and it is a **single-pinned-core** curve. Two things have never been measured:

1. **The full-box C(N) curve** (server spread across all physical cores). Every per-box projection in #2817 quotes a 0.6–0.75 marginal-efficiency discount **that has never been measured**. The one cross-box datum (0.55 marginal efficiency at 2× cores, xlarge→2xlarge) leaves open a scenario where the discount compounds across 8 cores and a 16-vCPU box delivers only ~2–3 cores of work — in which case fixing the handoff is a **2.5–4× box-level lever**, ahead of #3096 (Arrow encode) in the funding order. Nothing measured to date rules this in or out.
2. **Stack-level attribution of the overhead.** WS0's baseline report explicitly declined off-CPU analysis and named this exact phenomenon as where it would pay first (§"On-CPU profiling only — declined here, not overlooked": the 1.98M voluntary context switches "are the entire phenomenon and are exactly what an off-CPU profile would explain"; ~13.5% of the Flight arm's wall time was off the metered CPUs, unaccounted for). An on-CPU flame graph alone cannot see time spent parked — the suspected cost lives in the per-stream producer/consumer pair (`cqlite-flight/src/streaming.rs`: blocking-pool producer → bounded `tokio::sync::mpsc` capacity 4 → encoder/gRPC drain, i.e. 2N contexts for N streams) plus egress-credit accounting.

**Why a local box suffices:** #3100 Part C acquitted the device (reads are 128–256 KB, `%util` ~idle at N=16). The phenomenon is pure CPU/scheduler — no Cassandra in the loop, no exact-corpus requirement, no AWS provision needed. AWS is only needed later for publishable absolutes vs Cassandra (#3100's still-open follow-up question).

## Routing

[...]

Issue #2859, opened 2026-07-24
Bootstrap mold linker on Linux agent workers (EC2) + gate accelerator stamp


pmcfadin:
## Context

EC2 Linux agent workers currently link with the slowest available option. On the pinned 1.88.0 toolchain, Linux uses GNU bfd; after #2856 (bump to 1.97.1), **x86_64**-linux gets lld by default, but **aarch64/Graviton workers get nothing**, and mold still beats lld. Linking is the one cost sccache cannot cache — every `--lite` round and full gate re-links all test binaries (`debug = true`) from scratch, so on warm workers link time is a large slice of remaining wall-clock. Upstream reference: bfd→lld alone measured ~7× link / ~40% e2e incremental (rust-lang 1.90 announcement); mold improves further and covers both arches.

macOS is explicitly out of scope: mold is Linux-only (the "sold" macOS fork is dead/commercial) and Apple's ld-prime is already optimal there.

## Routing: design-driven (process/tooling) → flow-activate (OpenSpec)

Design shape pre-agreed with owner in-session 2026-07-24: bootstrap detection + per-machine cargo config + gate accelerator stamp + recorded A/B. Seam 1 = confirm the rendered spec.

## Proposed shape

1. `scripts/bootstrap-agent-machine.sh` Linux branch (mirrors existing sccache/nextest accelerator pattern): detect `mold`; install via native package manager where possible, else `warn` with cost estimate (consistent `ok/warn` output).
2. Write per-machine `~/.cargo/config.toml` sections `[target.x86_64-unknown-linux-gnu]` + `[target.aarch64-unknown-linux-gnu]` linking via mold. Idempotent (safe re-run), never clobbers unrelated user config, and a strict improvement/no-op layered over the post-#2856 lld default. Repo config untouched; Darwin branch untouched.
3. `scripts/agent-gate.sh` `accelerators:` line stamps `mold=present/absent` on Linux (n/a on macOS) — degradation visible, not silent, same contract as sccache.
4. A/B measurement on one EC2 worker (full gate + one `--lite` round, with/without) recorded on the issue/PR — the win becomes a number, not a vibe.

## Acceptance criteria

1. Bootstrap on Linux detects/installs-or-warns mold; self-test coverage added in `scripts/tests/test_bootstrap_agent_machine.sh` (Linux + Darwin paths simulated; Darwin makes no linker change).
[...]

Issue #2088, opened 2026-07-06
roborev: severity-stratify findings (blocker vs nit) — gate merge only on blockers, nits ride to follow-ups + telemetry split


pmcfadin:
Parent epic: #2083. From audit Q2/quality-R3 (F2).

## Problem
608 roborev findings across 174 issues carry NO severity, so a cosmetic nit forces the same expensive re-verify round as a real bug. Rework (334 rounds, 64.9% of issues) is the #1 weighted failure driver in the telemetry retro; p90 is 10 findings / 3 gate runs per issue, max 40 findings / 9 runs.

## Routing
Direct implement (skip OpenSpec). If roborev's own output schema can't change (external tool), implement as a documented TRIAGE step: the reviewer of record classifies each finding blocker/nit against a written rubric.

## Change
1. Define the rubric in doctrine (CLAUDE.md + roborev-findings page): **blocker** = correctness, data-parity, safety (unwrap/panic paths), no-heuristics violations, wiring-evidence gaps, security; **nit** = style, naming, comment/doc polish, test-robustness suggestions with no failing scenario
2. Loop rule: blockers must be fixed pre-merge; nits are batched into ONE follow-up issue (linked, labeled) and do NOT trigger a re-verify round
3. Telemetry: extend delivery-telemetry schema/record with roborev_blockers / roborev_nits counters (schema v bump; keep authoritative-only rule — a count not observed is an error, never 0)
4. Keep the blocker bar conservative: when in doubt, blocker (the pre-merge full gate + required CI backstop mis-graded nits)

## Acceptance criteria
- [ ] Rubric documented; flow-implement references it in the roborev step
- [ ] Nit-batching rule + follow-up-issue pattern documented
- [ ] delivery-telemetry.py + schema record the severity split; delivery-telemetry gate component still passes
- [ ] Retro after ~30 issues: rework rounds/issue vs 1.92 baseline (334/174)

PR #3251, opened 2026-08-03
fix(#3249): persist a profileable perf configuration on agent boxes — sysctl drop-in, functional verification, gate visibility

### LINKED ISSUE #3249: Persist a profileable perf configuration on agent boxes: perf_event_paranoid=4 blocks all profiling and is not set anywhere in /etc/sysctl.d
## Summary

Agent/worker boxes ship with `kernel.perf_event_paranoid = 4`, which **blocks all `perf` use by unprivileged users** and is not overridden anywhere in `/etc/sysctl.conf` or `/etc/sysctl.d/`. Every profiling run therefore starts from a hard denial, and the denial is easy to misread as a hardware limitation. Persist a profileable configuration and bake it into `scripts/local/bootstrap-agent-machine.sh` so a profileable box is the fleet default rather than a manual step someone forgets.

## Evidence

Probed on a worker box (`Intel Xeon Platinum 8488C`, 8 physical / 16 threads, `Hypervisor vendor: KVM`, bare host — no container):

```
$ cat /proc/sys/kernel/perf_event_paranoid
4
$ grep -rs perf_event_paranoid /etc/sysctl.conf /etc/sysctl.d/
(empty)
```

At `4` — above even Ubuntu's patched ceiling of 3 ("disallow all perf event use by unprivileged users") — `perf stat -e cycles,instructions`, `perf stat -C`, and `perf record` **all fail with EACCES**, emitting the "access limited" help text. That text reads like a capability verdict but is a permission verdict.

After `sudo sysctl -w kernel.perf_event_paranoid=-1 kernel.kptr_restrict=0`, the same box **does** support `cycles`/`instructions`/IPC, CPU-wide `perf stat -C`, and `perf record -e cycles` with kernel and user symbolization (`Total Lost Samples: 0`). So the setting was gating capability that is genuinely present.

## Why `-1` specifically, not `1`

`perf_event_paranoid` is cumulative and higher is more restrictive: `>= 1` **disallows CPU event access**. CQLite's measurement rig (#3096) mandates **CPU-wide `perf stat -C`**, which is exactly per-CPU event access, so anything `>= 1` breaks it. `0` is the bare minimum; **`-1`** additionally lifts the mlock limit and avoids surprises for `perf record` ring buffers. `kernel.kptr_restrict = 0` is needed for kernel symbol resolution (without it, kernel frames render as unresolved addresses — a silent attribution loss, not an error).

Note: a permissive `perf_event_paranoid` does **not** grant BPF map creation. bcc/bpftrace collectors still require `sudo` (the #3217 finding); this issue does not change that.

## Why this matters beyond convenience

Two consecutive measurement issues ended with a large unattributed residual (#3217 ~87% of its IPC decay; #3096 82% of per-row encode). This setting is not the whole cause — #3224's probe confirms the guest vPMU genuinely lacks topdown/stall/LLC events — but it is an **additional** blocker that had to be cleared by hand on a live box before we could even establish which of the two was operative. A fleet default removes one of the two variables permanently and makes the remaining hardware gap unambiguous.

## Acceptance criteria

1. `/etc/sysctl.d/` drop-in (e.g. `99-cqlite-perf.conf`) setting `kernel.perf_event_paranoid = -1` and `kernel.kptr_restrict = 0`, applied idempotently by `scripts/local/bootstrap-agent-machine.sh` and surviving reboot.
2. Bootstrap **verifies** rather than assumes: after applying, assert `perf stat -C 0 -e cycles -- sleep 0.1` exits 0 and reports a non-zero cycle count. A bootstrap that silently leaves a box unprofileable is the failure mode being fixed.
3. The verification result is reported in bootstrap output (and ideally in the gate's `accelerators:` line or an equivalent visible surface), so an unprofileable box is discoverable without a manual probe.
4. Documented in `docs/development/fleet-runbook.md`, including the `-1`-not-`1` rationale and the note that BPF collectors still need `sudo`.
5. Security posture stated explicitly in the runbook: this is appropriate for dedicated single-tenant measurement/agent boxes, and is a deliberate loosening not to be applied to shared or multi-tenant hosts.

## References

Discovered while probing #3224 (microarchitectural cause of full-box IPC decay) — see its capability-matrix comment for the full probe. Related: #3226 (measurement doctrine + harness guards), #3096 (the rig that mandates CPU-wide `perf stat -C`), #3217 (the run whose residual prompted the probe). Parent: epic #2817.

pmcfadin:
Closes #3249

## Problem

Agent/worker boxes ship with `kernel.perf_event_paranoid = 4` — above even Ubuntu's patched ceiling of 3 ("disallow all perf event use by unprivileged users") — and it is set **nowhere** in `/etc/sysctl.conf` or `/etc/sysctl.d/`. Every profiling run therefore starts from a hard EACCES denial whose help text reads like a *capability* verdict but is a *permission* verdict. Two consecutive measurement issues (#3217, #3096) ended with a large unattributed residual, and this setting had to be cleared by hand on a live box before we could even determine which of the two blockers was operative.

The probe that motivated this found the state sharper than filed: this box's **runtime** value is already `-1` (residue from the #3224 hand-probe) but there is still **no `/etc/sysctl.d/` drop-in** — so the fleet was profileable only *by accident*, on the one box someone had touched, and reverts on reboot or reprovision. Three independent reports already document these two sysctls silently reverting (`ws0-3217-report.md:214`, `ws3-3029-report.md:63`, `ws0-cassandra-baseline-2026-07-27.md:847`), and one checked-in script *hardcodes* a "both SILENTLY REVERT on this box" warning. Agents had been hand-working around this in triplicate for at least three measurement cycles.

## What this does

- **`scripts/perf-capability.sh` (new)** — the single shared helper this repo lacked. Recon found no perf-capability probe anywhere in `scripts/`, and notably **no `#3096` rig exists in-tree at all** (it is referenced only in prose across reports and specs); the sole prior art was `ws0_assert_sysctl()` buried in an archived artifact directory. Exposes a free `/proc` capability read, a functional `perf stat -C 0` verification, and the canonical drop-in bytes.
- **`scripts/bootstrap-agent-machine.sh`** — new `Perf profiling capability` section (placed after mold, the other Linux machine-capability provisioner, sharing its probe-first/fail-safe posture). Installs `/etc/sysctl.d/99-cqlite-perf.conf` idempotently, applies it, **reads the value back from `/proc/sys`** (never trusts the write's rc), then runs AC2's functional verification. This is the first `/etc/*` write in bootstrap; it degrades to a `warn` plus a printed remedy on a no-sudo / no-perf / non-Linux / container box and never exits non-zero, preserving the script's "informational, always exit 0" contract.
- **`scripts/agent-gate.sh`** — a `perf=` token in the `accelerators:` line. Deliberately a **free `/proc` read** (fork-free `read -r v < "$f"`, no `perf` exec, no `sudo`, no subprocess): the gate answers "is this box *configured* profileable", while the ~100ms functional `perf stat` stays in bootstrap. The gate does not pretend to answer "does the vPMU have the counters" — that is #3224, and genuinely hardware.

## Acceptance criteria

| AC | Status |
|---|---|
| 1. `/etc/sysctl.d/` drop-in, applied idempotently, surviving reboot | ✅ whole managed file, byte-compared in-shell |
| 2. Bootstrap **verifies** — `perf stat -C 0 -e cycles` exits 0 **and** non-zero count | ✅ and the negative cases are tested, not assumed |
| 3. Result reported in bootstrap output **and** the gate's `accelerators:` line | ✅ visible in this PR's own lite summary: `… mold=absent perf=ok` |
| 4. Documented in `fleet-runbook.md` incl. `-1`-not-`1` rationale + BPF-needs-sudo note | ✅ |
| 5. Security posture: single-tenant only, never shared/multi-tenant | ✅ in the runbook **and** the drop-in's own header |

### Why `-1` and not `1`
`perf_event_paranoid` is cumulative and higher is more restrictive; **`>= 1` disallows CPU event access**, which is exactly what the mandated CPU-wide `perf stat -C` needs. `0` is the bare minimum; `-1` additionally lifts the `perf_event_mlock_kb` limit, avoiding `perf record` ring-buffer surprises. `kptr_restrict = 0` is a separate control for kernel *symbol* resolution — without it kernel frames render as unresolved addresses, a **silent attribution loss rather than an error**. A permissive `perf_event_paranoid` does **not** gate the `bpf()` syscall, so bcc/bpftrace collectors still require `sudo` (the #3217 finding); this PR does not change that.

## Review — 3 streams, 10 blockers, all fixed

Review-first per #2086: `rust-reviewer`-equivalent shell/privilege review + a mutation-tested test-quality review + roborev, all on the lite-green diff **before** any full gate.

**roborev round 1 (`RESULT: FAIL`, 2 findings)** — verifiably genuine: `reviewed-sha: 11f72ca..fc4758e`, `prompt-content: PASS (5/5 code census paths present)`, tokens `input=96250 cached=73984 output=6620` vs the ~18.7k/0/53 vacuous signature; all eight anti-vacuity assertions PASS.
1. The non-`--yes` printed remedy wrote the drop-in but **omitted `sysctl --system`** — copy-pasting it left the box unprofileable until reboot. Upgraded to blocker: that is verbatim the failure AC2 exists to prevent, on the path most operators take.
2. Hybrid-PMU CPUs report qualified event names (`cpu_core/cycles/`), misread as `no-cycles-row`. Fails *safe* (spurious warn, never a false ok) but a check that cries wolf destroys its own signal — and our Sapphire Rapids fleet box could never have surfaced it.

**Shell + privilege review (2 blockers)**
3. **The test-only `CQLITE_PERF_SYSCTL_DIR` seam steered a *privileged* `sudo tee`.** `CQLITE_PERF_SYSCTL_DIR=/etc/sudoers.d` + `--yes` made root write an env-chosen file while reporting `[ok] wrote …`. Damningly, the test suite already `unset` both seams with the comment "A worker shell may export the seams themselves" — the code judged a stray export plausible and defended the tests but not the privileged path. Now: privileged destination is a hardcoded `/etc/sysctl.d` literal; seams are inert without `CQLITE_PERF_TEST_MODE=1`, and that marker **refuses to run against a reachable real `sudo`/`sysctl`**.
4. `[ "$count" -le 0 ]` **failed open**: `[` returns **2** (not 1) on a malformed operand, so an over-wide digit string fell through to `return 0` — false VERIFIED, plus leaked stderr. Commit `fc4758e` had fixed this exact class one commit earlier in a sibling function and missed the one whose entire job is not failing open.
5. *(upgraded from the reviewer's nit)* Doc mirrors + the **normative** `openspec/specs/agent-fleet-runtime/spec.md` described the accelerators line as an exhaustive token list without `perf=`. Per CLAUDE.md's same-change doctrine rule, all four surfaces updated (2 new SHALL requirements + 10 scenarios), mirroring the mold token's #2859 precedent.

**Test-quality review — MUTATION TESTED (5 blockers).** 28 mutants, 17 killed / 11 survived. It confirmed the headline property was genuinely load-bearing (rc-only `verify` → 7 failures; the `fc4758e` guard properly pinned; relaxing `ACCEL_LINE_RE`'s `perf=` alternation to `.*` → killed), but five decisive guards would have shipped green:
6. **The `/proc` read-back was never proven honoured** — trusting `sysctl`'s rc kept the suite 25/25 green while bootstrap printed "applied and READ BACK from /proc" on a box still reporting `4`. The "did NOT take" warn had **zero** coverage. Silent revert is the documented trap this issue exists to close, and it was the least-tested path.
7. **The gate's `perf=` token was never checked against a real `/proc` read** — hardcoding `"ok"` survived 118/118, because every case set the override seam. A gate stamping `perf=ok` on a paranoid-4 box is precisely what AC3 exists to prevent.
8. **`perf stat` exiting non-zero was untested — and that IS the real `paranoid=4` EACCES state**, i.e. the condition this issue is about.
9. The perf-binary-absent branch was uncovered; `exit 1` there survived, and would break every box without `linux-tools`.
10. The suite was silently Linux-host-only — forcing the non-Linux branch produced 10 failures, so a Darwin gate host would red `tooling-tests`.

Plus ~14 nits, all fixed — including a real footgun: **`/etc/sysctl.conf` is applied *after* `sysctl.d` drop-ins and therefore beats our `99-` file**, so a stale entry there would silently defeat this change; diagnostics now grep it and the runbook documents the precedence. Also added Debian/Ubuntu's `>= 3` ladder level, which is what explains why the observed `4` denies everything.

**Fix-round verification: 13 mutants re-applied, each now FAILS a test** (hardcoded gate state → 2 fails; read-back trusting rc → 2; `perf` rc≠0 returning success → 2; `exit 1` after perf-MISSING → 1; seam honoured without marker → 1; `PERF_ROOT=(sudo)` → 1; unvalidated count → 3; literal `^cycles` → 3; Darwin branch disabled → 1; …). A fix without a killing test is not done.

## Verification

- `--lite` **PASS** at `f9da66c`, `tree-integrity: PASS`, `accelerators: … mold=absent perf=ok`
- `test_perf_capability.sh` **PASS=51 FAIL=0** (was 25)
- `test_agent_gate_summary.sh` **123 passed, 0 failed** (was 118)
- `test_bootstrap_agent_machine.sh` **PASS=80 FAIL=0** (regression check)
- `--only tooling-tests` → **PASS (359s)** — proves the new suite is really in the component chain, not merely written
- Hermeticity: the host's real `/etc/sysctl.d` and `/proc` are never touched; asserted directly by the suite's final case

The **full gate of record has not yet run** — it runs once, immediately pre-merge, in `flow-closer`.

## Follow-up (not in scope)
`scripts/bootstrap-agent-machine.sh` is 1196 lines and was already over the ~800 source target before this issue; a split by responsibility is a genuine #1116 candidate.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -161,8 +161,12 @@ cat /tmp/gate-summary.txt   # the SUMMARY block is the ONLY gate text an agent r
   block's `commit:`/`dirty:` are derived from that verified capture, never a fresh emit-time git
   read. No env var bypasses it; remedy is to re-run on a stable tree (don't edit a worktree while
   its gate runs).
-- Every SUMMARY carries an `accelerators:` line (sccache/nextest/lane state) — degradation there is
-  actionable, not noise. Self-test: `bash scripts/tests/test_agent_gate_summary.sh`.
+- Every SUMMARY carries an `accelerators:` line (sccache/nextest/lane state, plus a `mold=` token and
+  a `perf=` profiling-capability token on Linux hosts, #2859/#3249) — degradation there is
+  actionable, not noise. `perf=paranoid-<N>`/`kptr-restricted` means THIS BOX CANNOT BE PROFILED (a
+  PERMISSION verdict, not a missing capability): re-run `bash scripts/bootstrap-agent-machine.sh
+  --yes`, which installs + verifies `/etc/sysctl.d/99-cqlite-perf.conf`. Self-test:
+  `bash scripts/tests/test_agent_gate_summary.sh`.
 
 ## Core Commands
 
--- a/docs/development/agent-machine-setup.md
+++ b/docs/development/agent-machine-setup.md
@@ -93,7 +93,7 @@ accelerators: sccache=on nextest=on lanes=on sccache-health=ok
   **`paranoid-<N>`** (`perf_event_paranoid = N >= 1`
   forbids **CPU-wide** events, so the `perf stat -C <cpu>` the measurement doctrine
   mandates is DENIED — a *permission* verdict, not a missing capability; images ship
-  `4`) · **`kptr-restricted`** (paranoid is fine but `kptr_restrict != 0`, so kernel
+  `4`, and Debian/Ubuntu's extra `>= 3` level denies unprivileged perf *entirely*) · **`kptr-restricted`** (paranoid is fine but `kptr_restrict != 0`, so kernel
   frames resolve to bare addresses — a silent attribution loss) · **`absent`** (the
   `/proc` controls are not present, e.g. a container — tune the host) ·
   **`unknown`** (present but unparseable; never guessed). Anything but `ok` on a box
--- a/docs/development/agent-machine-setup.md
+++ b/docs/development/agent-machine-setup.md
@@ -86,6 +86,21 @@ accelerators: sccache=on nextest=on lanes=on sccache-health=ok
   `overridden` means a global `RUSTFLAGS` is suppressing the wired flags (don't
   export one on a worker); `present-unconfigured` means mold is installed but not
   wired — re-run bootstrap.
+- **`perf`** (issue #3249, **Linux only** — no token on macOS) — can this box be
+  profiled *at all*? Read free from `/proc/sys/kernel/{perf_event_paranoid,kptr_restrict}`
+  (never a `perf` exec, so the gate pays nothing for it). Values:
+  **`ok`** (unprivileged per-CPU profiling **and** kernel symbols available) ·
+  **`paranoid-<N>`** (`perf_event_paranoid = N >= 1`
+  forbids **CPU-wide** events, so the `perf stat -C <cpu>` the measurement doctrine
+  mandates is DENIED — a *permission* verdict, not a missing capability; images ship
+  `4`) · **`kptr-restricted`** (paranoid is fine but `kptr_restrict != 0`, so kernel
+  frames resolve to bare addresses — a silent attribution loss) · **`absent`** (the
+  `/proc` controls are not present, e.g. a container — tune the host) ·
+  **`unknown`** (present but unparseable; never guessed). Anything but `ok` on a box
+  you intend to measure means **re-run `bash scripts/bootstrap-agent-machine.sh --yes`**,
+  which installs `/etc/sysctl.d/99-cqlite-perf.conf` and then *verifies* it by running
+  `perf stat -C 0 -e cycles`. Rationale (`-1`, not `1`), the BPF-still-needs-sudo
+  caveat, and the single-tenant security posture: `docs/development/fleet-runbook.md`.
 
 States: **`on`** (detected & used) · **`absent`** (missing → the gate prints a loud
 `WARN:` with the install command) · **`off`** (intentionally disabled via
--- a/docs/development/fleet-runbook.md
+++ b/docs/development/fleet-runbook.md
@@ -327,7 +327,8 @@ grep -Hn 'perf_event_paranoid\|kptr_restrict' /etc/sysctl.conf /etc/sysctl.d/*.c
   /run/sysctl.d/*.conf /usr/local/lib/sysctl.d/*.conf /usr/lib/sysctl.d/*.conf /lib/sysctl.d/*.conf
 ```
 
-**Never set `CQLITE_PERF_PROC_DIR` / `CQLITE_PERF_SYSCTL_DIR` in a shell.** They are test-only path
+**Never set `CQLITE_PERF_PROC_DIR` / `CQLITE_PERF_SYSCTL_DIR` / `CQLITE_PERF_SYSCTL_EXTRA_DIRS` /
+`CQLITE_PERF_TEST_SANDBOX` in a shell.** They are test-only path
 seams for `scripts/tests/test_perf_capability*.sh` and are **inert** unless `CQLITE_PERF_TEST_MODE=1`
 is also set — which in turn refuses to run if a real `sudo`/`sysctl` is reachable. The privileged
 destination is a hardcoded `/etc/sysctl.d` literal precisely so no exported variable can ever steer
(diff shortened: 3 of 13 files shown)


pmcfadin:
## Review coverage ledger — PR #3251 (issue #3249)

Recorded verbatim per the owner's standing order. Rounds 4-6 ran under the **old**
`default_max_prompt_size` (204,800 B) and certify only their own slices; the final round is a
single verified **full-range** review made possible by the ceiling raise (#3257).

| Range | Round | `prompt-content` |
|---|---|---|
| `11f72ca..8b186aa` | 4 (full-range) | PASS (5/5) |
| `8b186aa..47d184a` | 5 (sliced) | PASS (7/7) |
| `47d184a..a2c4af1` | 6 (sliced) | PASS (5/5) |
| `11f72ca..04c329c` | **final (full-range, closer)** | **PASS (7/7 code census paths present)** |

Union covers `origin/main...04c329c`. No gap. `805b239`'s tree is byte-identical to `a2c4af1`
(both `6bc3321131a4fbe683c55102a06bae4a0721ebab`); `04c329c` adds `+169/-3` over 5 files on top.

### Final full-range round (job 33) — verified non-vacuous
```
reviewed-sha: 11f72cafdbf751f9f6cf93bc2815112faebfdd3a..04c329c83357a51b2d5871d5c509645d56d85477
census: 13 files, +4382/-19
tokens: input=358284 cached=276480 output=10917
sha-assert: PASS    review-completed: PASS
prompt-content: PASS (7/7 code census paths present)
vacuity-tier1: PASS vacuity-tier2: PASS
findings: PRESENT (2)   roborev-exit: FINDINGS (exit 1)
RESULT: FAIL
```
Both range endpoints match `origin/main...HEAD` — not base-equal, not single-commit. No spill
signature (`Read the diff from:` absent; 7 `diff --git` headers present). At 250,935 assembled
prompt bytes this round **would have spilled silently under the old 204,800 default** — the
ceiling raise is load-bearing on this exact review.

**This PR did not merge on a zero-finding review.** The terminal `RESULT` is `FAIL` with two
findings, both triaged out of this PR rather than waived:
- **High — `perf_capability_env_guard` trusts `CQLITE_PERF_TEST_PRIV_DIR` textually** (a symlink to
  the real `sudo`/`sysctl` inside a declared shim dir, or `/usr`, lets a privileged test-mode
  bootstrap execute the real `sysctl --system`). This is **R8-2**, owner-ruled into **#3261** as a
  new AC and recorded there with acceptance criteria. Test-mode only; production verified
  unaffected by `strace` (zero syscalls naming a bogus seam tree with `CQLITE_PERF_TEST_MODE`
  unset).
- **Low — `competing_files` skips by basename rather than exact path** (a false "no competing
  file" diagnostic). Batched to **#3268** per #2088; nits never trigger a re-verify round.

### Gate of record
```
summary-file: /tmp/gate-3249-final.txt
run-id: /tmp/agent-gate.JlWzcO   commit: 04c329c   dirty: no
accelerators: sccache=on nextest=on lanes=on sccache-health=ok mold=absent perf=ok
tree-start/tree-end: 04c329c83357 digest 2c57625edcec   tree-integrity: PASS
RESULT: PASS   (all components PASS, 0 FAIL, 0 SKIP)
```
`accelerators: … perf=ok` — the token this issue adds, emitted by its own certifying run, which is
AC3's wiring evidence produced by the gate rather than asserted about it. Neither
`missing-fixtures` nor `missing-schemas` FAIL-CLOSED, and the positive `datasets:`/`schemas:`
stamps prove both checks ran.

Intent audit (**C**): PASS — all five ACs `satisfied`; the one P1 (condition (a): derive the token
from a real `/proc` with no seam set) closed by new case `9f-host`, which unsets all seven seams
the code reads and asserts `perf=ok` by whole-field equality against this box's real `/proc`.

Issue #3252, opened 2026-08-03
roborev cannot certify a large diff on this fleet: the prompt-size cap silently swaps the inline diff for a file pointer, and the wrapper misreports it as the worktree-bug signature


pmcfadin:
Found while delivering #3096. **This is a fleet-wide review-integrity gap, not a #3096 defect.**

## What happens

roborev v0.61.2 has a config key `max_prompt_size` — *"Maximum prompt size for this repo before falling back to file paths."* When a diff exceeds it, roborev does **not** fail: it silently replaces the inline diff with an out-of-band pointer. The prompt actually sent to the reviewer ends, verbatim:

```
### Combined Diff
(Diff too large to include inline)
The full diff has been written to a file for review.
Read the diff from: `.../.roborev/roborev-snapshot-3743278049/roborev-snapshot-content.diff`
```

On #3096 that produced a **6,719-byte prompt with ZERO `diff --git` headers** for a diff of 670,627 bytes raw / ~460 KB after `exclude_patterns`. Both `max_prompt_size` and `default_max_prompt_size` read `0` (unset → internal default, empirically **< 460 KB**). Corroborating strings in the installed roborev: templates `codex_range_fallback_{full,medium,short,shortest}` and `roborev-snapshot-content.diff(Diff too large; read '%s'.)`.

**Consequence: on this fleet, no large diff can be roborev-certified by default.** 1:1:1:1 puts every issue in a worktree and `flow-closer`'s final roborev pass is a **merge gate** — so a large change either blocks forever or gets waved through on a misread.

## Why it is worse than a plain failure: it mimics a different, known bug

`scripts/flow/roborev-review.sh`'s guard reported:

```
prompt-content: FAIL (41/41 code census paths absent from the prompt)
```

That message is **textually the signature of the T1 worktree bug** (`--branch` without `--repo` resolving against the root checkout and enqueuing the base commit). The actual cause was completely different, and the correct diagnosis was only reachable by retrieving and reading the prompt. An agent pattern-matching on "N/N absent" will conclude "wrong repo/SHA" and go chase a non-existent plumbing problem. Token accounting shows the reviewer **did** read the snapshot file (318k in / 281k cached) — so the review was not necessarily worthless, but the guard cannot verify it and **correctly fails closed**. The guard is right; its *diagnosis* is misleading.

## Remedy applied (host-level, NOT durable — this is the ask)

```
roborev config set --global default_max_prompt_size 4000000
roborev daemon restart
```

[...]

Issue #3245, opened 2026-08-03
query-semantics-oracle fails on any box whose CQLITE_DATASETS_ROOT checkout is stale: keyspace-granular fixture resolution + fetch-datasets.sh rm -rf destroys committed fixtures


pmcfadin:
## Summary

Two compounding defects make the `query-semantics-oracle` gate component fail on a machine for **every** issue, with an error that looks like a test failure but is purely environmental. Hit during #3109 (PR #3239), where it consumed a full ~25-minute gate run and an entire triage cycle.

### Defect 1 — fixture resolution is keyspace-granular, not fixture-granular

`sstables_root(keyspace)` in `cqlite-core/tests/query_semantics_oracle_parity.rs` (~lines 126-141) picks the **first root whose `<keyspace>/` directory exists**, then requires the fixture inside it — it never falls through per-fixture. So if `CQLITE_DATASETS_ROOT` points at a checkout that has `test_da/` but is missing one fixture directory inside it, the resolver locks onto that root and never consults the committed in-repo corpus, which does have it. Under `CQLITE_REQUIRE_FIXTURES=1` the affected cases hard-FAIL.

Observed: root checkout stale at `221d16d`, behind `origin/main` `948ca25`. Fixture `test_da/multiclustering_table-fd74ad508d2311f1a29b6d2c15dcffdf` was committed by #3032 (PR #3210), so it was absent from that root, and the two `multiclustering_bti__*` cases failed:

```
REQUIRE_FIXTURES: case multiclustering_bti__multi_component_slice: fixture dir multiclustering_table-* absent
REQUIRE_FIXTURES: case multiclustering_bti__first_component_prefix_slice: fixture dir multiclustering_table-* absent
```

The sibling `bti-multiclustering` component **PASSed on the same box in the same gate run**, because its resolver probes for the **Data.db file** (`issue_3032_..._select.rs`, ~lines 93-102) and correctly falls back to the in-repo corpus. That asymmetry is the bug: one harness is robust, the other is not.

### Defect 2 — `fetch-datasets.sh` deletes committed fixtures

[...]

Issue #3240, opened 2026-08-03
Flight bypass path never records stream_cold_fault/stream_decompress — a second sub-phase attribution blind spot (sibling of #3096's framing gap)


pmcfadin:
Discovered while delivering #3096 (Arrow encode on the `do_get` data plane). **Out of scope there; filed for prioritization.**

## Symptom

4 of 8 tests in `cqlite-flight/tests/issue_2819_subphase_metrics.rs` FAIL because the #3058 single-SSTable **bypass** path never records the `stream_cold_fault` / `stream_decompress` sub-phases.

All 8 pass under `CQLITE_FLIGHT_MERGE_PATH=merge`. So this is specific to the bypass arm that #3058 made the default for single-SSTable reads — i.e. **the shipping path is the one missing the attribution**.

## Why the gate does not catch it

The tests are feature-gated on `observability-testing`, which the standard gate component set does not build. `main` is green and stays green; the failure is only visible when that feature is enabled explicitly. Verified pre-existing on `main` — it is NOT caused by #3096 (whose changes were in the tree when this was observed, and are unrelated to cold-fault/decompress accounting).

## Why it matters

This is the **same family of defect** #3096 had to fix to make its own work measurable: `StreamSubPhase::Encode` timed only `flush_buffer` and not `encode_do_get`'s IPC framing, so levers targeting framing were unfalsifiable. #3096 closed that one (new `stream_encode_framing` sub-phase; framing measured at 313.0 ns/row = 17.9% of per-row encode).

This is the second instance: on the default read path, cold-fault and decompress cost is attributable to nothing. Any future I/O or decompression lever on the bypass path is unfalsifiable in exactly the same way — and the cold arm of #3096's baseline is page-in bound (~20.2 s for 2.77 GB), which is precisely the region these sub-phases would explain.

## Routing

**Oracle-driven** — the merge arm is the oracle: it records these sub-phases correctly, and the bypass arm must agree. No OpenSpec change needed.

## Acceptance criteria

1. The bypass path (`cqlite-flight/src/bypass.rs` → `producer_warm.rs` → `producer_stream.rs`) records `stream_cold_fault` and `stream_decompress` with the same semantics as the merge arm.
2. All 8 tests in `cqlite-flight/tests/issue_2819_subphase_metrics.rs` pass under BOTH `CQLITE_FLIGHT_MERGE_PATH=bypass` and `=merge` — a differential assertion, so the two arms cannot drift again.
[...]

Issue #3232, opened 2026-08-03
Publishable absolutes vs stock Cassandra: fresh-provision certification run, after a CPU lever lands (blocked on owner provision hold)


pmcfadin:
## Summary

Produce **publishable absolute throughput figures vs stock Cassandra** on a fresh AWS provision with a Cassandra arm, once a CPU lever has landed. This is the certification run that turns "we beat Cassandra by 1.12× (measured on a torn-down box)" into a number we can publish with a live head-to-head. **Filing so it does not evaporate — it currently exists only as prose in #3100's close.**

## Why this is owed

- **#3100** re-ran the WS0 head-to-head on the byte-identical corpus: Flight `do_get` bypass = **1.12× Cassandra `SELECT *`**, correctness-verified (exact digest, 4M rows), matched-pinning. But that cluster is **torn down**, and the number is a *re-measurement of the known state*, not a certification of an improvement.
- **#3217** established the full-box scaling curve (71% marginal efficiency at 6 cores) on a **local box with no Cassandra arm** — deliberately, because the scaling phenomenon is a software property measurable without Cassandra. It explicitly deferred the publishable-absolutes question.
- The owner's **provision hold still stands**: no fresh AWS box until a CPU lever (#3096 Arrow encode, the funded next lever) has landed and is worth certifying. This issue is the placeholder for that run so the requirement is tracked, not lost.

## Scope

**Blocked until:** at least one funded CPU lever (#3096) has merged and shows a local rows/s improvement worth certifying. Do NOT provision before then — a fresh box that re-measures the known 1.12× is wasted spend (the owner's stated rationale on #3100).

**When activated:**
1. Fresh AWS provision matching the WS0 reference class (`c7i.4xlarge` — 16 vCPU / 8 physical cores / 30 GiB), Cassandra 5.0.8, chunk cache OFF (stock), the one disclosed `MAX_HEAP_SIZE=8G` deviation.
2. The byte-identical WS0 corpus (`Data.db` sha256 `22d9ae22…ce922c` or a geometry-matched regeneration per #3100's accepted bar), sha re-verified before and after.
3. Both Flight arms (`CQLITE_FLIGHT_MERGE_PATH=bypass|merge`), post-lever binary, exact commit SHA recorded.
4. All three arms (CQLite bare scan / Flight bypass+merge / Cassandra `SELECT *`+`count(*)`), matched-pinning pairs only, median of ≥3 with per-arm dispersion.
[...]

PR #3254, opened 2026-08-03
perf(#3096): AC1 UNMET (16.3% short) — both Arrow-encode levers measure at ZERO; ships the committed perf rig, the IPC-framing attribution, a wire-safety fix, and an honest negative result

pmcfadin:
**AC1 IS UNMET: both optimization levers measure at ZERO, and Flight `do_get` remains 16.3% short of the 1.3x target. This PR ships the reproduction rig, one positively-attributed cost, a wire-safety fix, and an honest negative result.** Merging on owner decision (2026-08-03) that a correctly-measured negative result satisfies this change (spec R5); the throughput question is re-anchored to #3248.

Refs #3096 (deliberately NOT `Closes` — a squash merge auto-closing the issue would flip its board item to Done prematurely; #3229 hit exactly that)

## The headline numbers

Re-measured after the review fixes changed the framing geometry — 8 rounds x 3 arms = 24 interleaved warm/bypass runs, one session, rotated order, control in every run, all three binary md5s recorded:

| arm | rows/s (median) | spread |
|---|--:|--:|
| BASE | 232,245 | 5.1% |
| lever 4 only | 232,691 | 9.8% |
| lever 4 (corrected target) | 230,321 | 5.5% |

- **Lever 4 (batch/flight-data framing): −72 rows/s (−0.03%), 4/8 rounds positive. ZERO.**
- **Lever 6 (Arrow schema cache): ZERO.** Paired within-round attribution (the basis used for per-lever claims): **−320.1 rows/s, 5/10 rounds positive**. The difference of arm medians is −496 rows/s (−0.2%); both figures say the same thing, and the paired number is the one the committed artifact attributes with.
- Ratio vs same-session bare scan: **1.553x** against a 1.3x target of 275,223 rows/s — **shortfall 44,902 rows/s (−16.3%)**.

An earlier measurement recorded lever 4 at **+2.3% / +4,817 rows/s**. It **did not reproduce** and is **superseded** — it was measured at a 4 MiB flight-data target this PR removes as a wire hazard. It is retained in the artifacts labelled as such, not deleted. A `−137 cycles/row` movement with rows/s unmoved is **not** claimed as a win; spec R1 makes exactly that shape a FAIL (the #2877 pattern).

**Lever 4 is retained for WIRE SAFETY, not throughput.** Lever 6 is retained as correct-and-cheap.

## What ships with value

1. **A reproducible measurement rig — RE-ANCHORED TO #3272, NOT shipped in this PR.** The rig this issue's reproduction section pointed at (`/home/ubuntu/ws0-local/`) **did not exist on the delivery box and was never committed** — PR #3103 shipped only `.rs` + OpenSpec, and AC4's digest `0x0a2a390223bde6aa` appears nowhere in the repo. So the issue's stated baselines were unreproducible. Built and measured with, but **moved to #3272 by owner order and NOT in this diff**: `tools/ws0-corpus-gen` (production `SSTableWriter`, 4M rows, uncompressed per #1406, Murmur3 token order, deterministic from seed `30960001` — **verified byte-identical across three generations**) and `scripts/perf/` (same-session both-arm rig; CPU-wide `perf stat -C`, verified-sibling `taskset` fail-closed, warm/cold separate, setup-subtracted, non-zero exit on zero rows).
2. **The IPC-framing attribution fix — the load-bearing result.** `StreamSubPhase::Encode` timed only `flush_buffer`, **not** `encode_do_get`'s IPC framing, so any lever targeting framing was unfalsifiable. New `stream_encode_framing` sub-phase measures it at **313.0 ns/row = 17.9%** of per-row encode. Zero-cost when metrics are off (the stream returns unwrapped; one branch per RPC).
3. **A real wire-safety fix.** `with_max_flight_data_size` is a *target*, not a hard cap: arrow-flight 53.4.1 `split_batch_for_grpc_response` (`src/encode.rs:613-637`) sums total buffer **capacity**, sets `n_batches = ceil(capacity/target)`, then `rows_per_batch = (num_rows/n_batches).max(1)` — a **uniform row-count** slice, so width-skewed rows can put all bytes in slice 0. New `cqlite-flight/src/wire_partition.rs` byte-partitions before the encoder and `guard_message_within_ceiling` rejects any **serialized** body over the ceiling. **Perturbation-verified:** with the partition short-circuited, the test fixture frames a **14,495,104-byte body — 3.5x the 4,128,768-byte ceiling.**
4. **`arrow_convert.rs` split 2,596 -> 225 lines** across six responsibility-scoped files, all under the campsite targets; removes the file from the over-threshold list entirely. Public API unchanged (`git diff origin/main -- export/mod.rs` shows zero changed `pub` lines).
5. **A schema contract made TRUE rather than merely documented.** `rows_to_record_batch_with_schema` claimed a mismatched schema is rejected; `RecordBatch::try_new` compares only types/lengths, so a reordered same-type schema was silently accepted and mislabeled every column. Now checks arity/names/order, returning `ArrowConvertError::SchemaMismatch`.
6. **A corrected doc contradiction, settled from primary sources.** tonic 0.12.3/0.13.1 `codec/mod.rs:100` and grpc-java v1.79.0 `GrpcUtil.java:212` both default to **4 MiB and DO bind** (including our own `tools/flight-loadgen`, the client every number here came from). But arrow-java v19.0.0 `NettyClientBuilder.java:56` defaults `maxInboundMessageSize = Integer.MAX_VALUE`, inherited by `FlightClient` and JDBC — so **the Java Flight/JDBC client does NOT bind at 4 MiB.** `docs/research/phase2-verify-transport.md:19` was right *about the Java client* and is now scoped to it; `batch_bytes.rs`'s "Trino/JDBC will refuse" claim is removed. Lever 4's original 8 MiB rejection rationale was **partly wrong as written**.

## Output invariance (spec R3)

In-repo **Arrow-buffer** digest oracle (folds value + validity buffers in column order with batch row counts — not `QueryRow` values, so a bad builder/offset/null-bitmap is visible):

- CI fixture **`0xd0014e42e893f87f`** (500 rows / 4 batches)
- measurement corpus **`0x0390bfbb81a23fa1`** (4,000,000 rows / 31,250 batches)

**Both UNCHANGED after every lever and every fix; bypass identical to merge on both.** Proven **non-vacuous** by a perturbation probe showing the arms genuinely diverge when altered. The external `0x0a2a390223bde6aa` is not cited as evidence anywhere.

Also green: `issue_3058_forced_path_differential` (**debug only** — `--release` compiles out the `CQLITE_TTL_NOW_OVERRIDE_SECS` seam), `issue_2825_max_batch_bytes_e2e`, `issue_2821_egress_budget_e2e`, `do_get_transport_test`, `streaming_framing_tests`, `export::` (130).

## Measurement integrity notes

- **No absolute on this box is reusable cross-session.** The **untouched** bare scan read 370,134 -> 332,970 -> 357,790 rows/s across three sessions; BASE read 213,471 -> 232,245. Every comparison here is same-session interleaved with a code-identical drift control. Annotated at the top of the baseline artifacts.
- **A discarded run is recorded, with its reason** (`abc-interleaved-2026-08-03.md` §2): a first A/B/C attempt was thrown out because a concurrent `cargo` landed on the pinned cores and dropped IPC 1.52 -> 1.44. Its lesson is recorded too — **the drift control did NOT move during the bad run**, so a quiet control is necessary but not sufficient.
- **Cold audit: NOT contaminated.** `--temp cold --scan-passes N` could blend pass 1 (cold) with warm passes 2..N into a figure reported as "cold" (now refused fail-closed). Audited: `baseline-results.json` records `scan_passes: 1`, each cold rep's row denominator is exactly 4,000,000, and cold wall time is ~20.2 s vs warm ~10.8 s. Nothing withdrawn.
- **Honestly unrecoverable:** the first A/B/C run's three binary md5 digests were never written into its artifacts. Stated as unrecoverable rather than reconstructed.

## Review

**Review-first** (before any full gate, per #2086). `rust-reviewer`: 2 blockers + 6 nits, all fixed — and it verified the split behavior-neutral item-by-item (match-arm order, error-string casing, the `Some(&Value::Null)` tuple sentinel, every `checked_offset` call site) and lever 6's schema cache safe. **roborev job 15**, genuine (466,390-byte prompt, 41/41 `diff --git` headers, tokens 544,217 in / 416,768 cached / 7,926 out, 219 s): **4 new findings, all fixed** (the wire-cap hole above, the false schema contract, the cold-blend, and a `pkill -x cqlite-flight` that would kill a peer lane's server).

**roborev round 1 was VACUOUS and correctly caught** by #2964's guard (`prompt-content: FAIL — 41/41 code paths absent`). Root cause: roborev's prompt-size cap silently swapped the inline diff for an out-of-band file pointer. Filed as **#3252 (P1)** — by default **no large diff on this fleet can be roborev-certified**, and the guard's message mimics the T1 worktree-bug signature.

## Follow-ups filed

- **#3248** — successor: attribute inside the unattributed encode region. **Buys a profile, not a patch.** The "82% is array build" figure is a **COMPLEMENT, not an attribution** (`1,746 − 313`), with no per-function data inside it; levers 2/3/1 are three unpriced candidates. It also owes the reconciliation of `1,746 ns/row` vs `+4,697 cycles/row` (the gap is less than half the region, so the encode region is **not wholly additive** over the bare scan).
- **#3231** — lever 1 (column-major, delete the per-row `HashMap<Arc<str>, Value>`) added as an **unpriced** candidate. For a shared-path lever, predicted-absolute and predicted-ratio **diverge**: construction is paid by both arms, only consumption moves the ratio.
- **#3252** — the roborev large-diff certification gap.
- **#3240** — bypass path never records `stream_cold_fault`/`stream_decompress`; third instance of `observability-testing`-gated invisibility.
- **#3232 / #3234** — credited: this generator is #3232's geometry-matched regeneration path, and #3234 should mirror its determinism contract.

## Spec

`openspec/changes/arrow-encode-doget/` — **AC1 recorded as `unmet`, re-anchored to #3248. Not `satisfied`, not softened.** `tasks.md` carries the shortfall and the 82%-is-a-complement correction; two items are left **partial rather than checked**. The kill criterion (R5) was applied as specified and returned NOT MET.

---

## Gate of record + one cited waiver

```
run-id: /tmp/agent-gate.Nb2Mju
commit: 176fc3f  branch: issue-3096-arrow-encode  dirty: no
tree-start: 176fc3f0eb72  tree-end: 176fc3f0eb72  tree-integrity: PASS
28 of 29 components PASS
query-semantics-oracle: FAIL (0s)
RESULT: FAIL
```

**`query-semantics-oracle: FAIL (0s)` is a pre-existing ENVIRONMENTAL failure tracked by #3245** (keyspace-granular fixture resolution), waived here with evidence:

- **Root cause is fixture *resolution*, not a semantic divergence.** `sstables_root(keyspace)` (`cqlite-core/tests/query_semantics_oracle_parity.rs:126-142`) picks the **first root whose `<keyspace>/` dir exists** and never falls through per fixture. `/data/datasets/sstables/test_da/` exists (holding `collection_table`, `simple_table`, `ttl_table`, `wide_table` from the pinned release tarball) but has **no `multiclustering_table-*`** — that fixture is git-committed only (`test-data/datasets/sstables/test_da/multiclustering_table-fd74ad508d2311f1a29b6d2c15dcffdf/da-2-bti-*`, added by #3032 / PR #3210). So the resolver locks onto `/data/datasets`, never consults the in-repo corpus, and under `CQLITE_REQUIRE_FIXTURES=1` fail-closes with `REQUIRE_FIXTURES: case multiclustering_bti__{multi_component,first_component_prefix}_slice: fixture dir multiclustering_table-* absent`. The other 6 cases PASS **with reconciliation applied**. The `0s` is 0.20 s cached build + 0.08 s test — a setup guard, never a comparison.
- **Reproduced byte-identically on a clean detached `origin/main` (`aac7ce0`) worktree** with the same env: exit 101, same panic at `:653`, same two `REQUIRE_FIXTURES` lines, `5 passed; 1 failed`.
- **This PR touches none of it**: `git diff --stat origin/main...HEAD` is empty for both `test-data/query-semantics-oracle.json` and `cqlite-core/tests/query_semantics_oracle_parity.rs`, and the test drives `cqlite_core::Database` directly — never Arrow or Flight, so none of this PR's changes (the `arrow_convert` split, the `SchemaMismatch` check, `wire_partition.rs`, `FLIGHT_DATA_SIZE_TARGET_BYTES`, `EgressBatchPlan`, `stream_encode_framing`) is in that call chain.
- **The sibling `flight-query-semantics-oracle: PASS (9s)` is consistent, not contradictory.** Both multiclustering cases carry `flight_lane: false` (their queries carry `WHERE`, inexpressible as a whole-table `do_get` projection), and `cqlite-flight/tests/query_semantics_flight_parity.rs:491-508` returns `Ok(false)` on that declared property **before** any fixture-presence check — so the Flight lane never resolves `multiclustering_table-*` and cannot hit the guard.

**Waiver basis:** the clean-`main` reproduction. CLAUDE.md's docs-only cite-and-waive rule is **void for this PR** — it touches compiled inputs throughout — so the `main` repro is what makes this legitimate rather than the diff's shape. The fix belongs to #3245 (resolve the datasets root **per fixture**: first root containing `<keyspace>/<prefix>*/*-Data.db`, mirroring the working `issue_3032` probe), not here.

Also confirmed by this run: **`node-bindings: PASS (147s)`**. An earlier gate attempt reported `node-bindings: FAIL (482s)` — that run was **void** (`tree-integrity: FAIL (tree-capture-failed)`, `commit: unverified`) because the root filesystem hit 100% full mid-run from a 137 GB `target/`. The failure was ENOSPC wearing a component failure's clothes; filed as **#3256** (gate disk preflight).

---

## C intent audit — PASS (AC1 recorded `unmet`)

`spec-auditor` against `openspec/changes/arrow-encode-doget/specs/arrow-encode-doget/spec.md`:

| Req | Verdict |
|---|---|
| **R1** ratio, rows/s AND cycles/row | **unmet — re-anchored to #3248** |
| **R2** measurement traps; warm != cold | **partial (justified)** — warm/cold separation satisfied by committed artifact; the two script-dependent scenarios re-anchored to #3272 with R4 |
| **R3** in-repo Arrow-buffer digest oracle | satisfied |
| **R4** committed, clean-checkout rig | **unmet — re-anchored to #3272** (rig split out by owner order) |
| **R5** kill criterion / negative result | satisfied |
| **R6** wired from `do_get`, contracts preserved | satisfied |

**Overall: PASS** — every requirement satisfied on its own evidence, with AC1 explicitly `unmet` and the kill-criterion requirement (R5) satisfied by an honestly-measured, honestly-reported negative result. The audit found **no laundering**: the negative result is stated first in `tasks.md`, the PR title/body and both artifact docs; the CPU-share trap is named and refused; the 82% is labelled a complement, not an attribution.

Two audit items were **fixed** rather than deferred:
- The kill-criterion line in `tasks.md` cited the superseded-session figure without its `4 MiB target` label — the only unlabelled instance in the record. Now names both sessions with their targets and identifies the shipped one (ratio 1.553x, 230,321 vs 275,223, 16.3% short) as delivered. Docs-only (`6f6313a`).
- The lever-6 basis above now states which attribution it reports.

Two were carried to **#3248** as durability gaps (evidence exists, but is not mechanically enforced): the rig's "byte-identical across three generations" determinism claim has no regenerate-and-compare self-test, and the measurement-corpus digest `0x0390bfbb81a23fa1` is recorded but not pinned in-tree, so its post-lever invariance rests on operator re-runs. Also carried: `scripts/perf/`'s three fail-closed guards have no `scripts/tests/` self-test.

**Gate re-run note:** the docs-only commit `6f6313a` could not be `--delta` re-certified — `--delta` refuses any anchor whose `RESULT` is not literally `PASS`, even when the FAIL is entirely a cited pre-existing waiver. That mechanism gap is filed as **#3259**; rather than merge a SHA no full run had certified, the full gate was re-run at `6f6313a`.

---

## C RE-AUDIT after the split — verdict: **CHANGES NEEDED**, remedied above (docs-only)

The prior C PASS was audited against a tree that **contained the measurement rig** and does not describe this tree. It was re-run rather than carried forward.

| Req | Verdict |
|---|---|
| **R1** ratio, rows/s + cycles/row | **unmet — re-anchored to #3248** |
| **R2** measurement traps; warm != cold | **partial (justified)** — warm/cold separation satisfied by committed artifact; the two script-dependent scenarios re-anchored to #3272 |
| **R3** in-repo Arrow-buffer digest oracle | satisfied |
| **R4** committed rig, clean checkout | **unmet — re-anchored to #3272** |
| **R5** kill criterion / negative result | **partial → the blocking gap was THIS PR BODY** (see below) |
| **R6** wired from `do_get`, contracts preserved | satisfied |

**The gap R5 failed on was the posted record, not the code.** The in-tree spec, `tasks.md`, and artifacts were sound — `spec.md` has been touched by exactly one commit (`c2a7a50`) and never edited since, and `tasks.md` states the ZERO result, the 1.553x ratio, the −44,902 / 16.3% shortfall, and both re-anchors plainly. But **this PR body still claimed "R4 … satisfied", "Overall: PASS", and that the change "ships the committed perf rig"** — a narrative describing the pre-split tree. R5 explicitly forbids a negative result being *"reworked into a passing narrative"*, and a PR body advertising a rig that is not in its own diff is exactly that. Corrected above; the false claims are removed rather than reworded.

Two in-tree artifact corrections are also required and are being made as a docs-only commit: `abc-interleaved-2026-08-03.md:9` headlines the **superseded 4 MiB** figure ("+2.0% … 15.0% short") unlabelled at the top of the primary artifact, and lines `:238`/`:363` describe "a committed rig" / "the COMMITTED rig"; and `measurement-method.md` + `rig-guard-audit-2026-08-03.md` need the same **#3272 banner** already carried by `baseline-2026-08-03.md`, because `measurement-method.md` still asserts re-derivation "on a fresh clone" from `scripts/perf/…` (3 refs), which is false on this tree.

**On R5's "with the rig committed" clause**, the auditor's judgement — which I accept: that clause is **R4's obligation, not a second one**. R5's normative SHALLs (stop on immaterial gain; post rows/s AND cycles/row, warm and cold separately; attribute each lever individually; stack nothing further) are all met with committed evidence, and the negative result remains **auditable** here — every per-run number, binary md5, corpus identity, pinning and counter mode is in-tree. Only *re-running* it from this branch is now impossible. Failing R5 on the rig's absence would double-count a gap already recorded as R4 → #3272.

**Residual the owner should hold: the negative result's reproducibility now depends on #3272 landing.**

### Verified clean by this audit (worth recording)
- **R3**: the oracle folds `ArrayData` value + validity buffers, offsets, and child data in schema order with per-batch row counts, driven through the **public `do_get` RPC** at a pinned `now`; the arms are proven to differ via `ReadPathProbe` **in both directions**; rows are pinned at 500 *before* any equality check; `0xd0014e42e893f87f` asserted at HEAD. The post-split fixture arm is genuinely self-contained (`tests/support/ws0_fixture.rs`, `cqlite_core` only, DDL asserted byte-equal to the committed `.cql`) and fails closed on 0 rows, partition mismatch, any `CompressionInfo.db`, or an empty `Data.db`. It **does** run in CI (flight-ci `full` = `cargo test --package cqlite-flight`, mandated by a `cqlite-core/**` + `cqlite-flight/**` diff).
- **R6**: zero `unwrap()`/`expect()` across all 11 touched library files; `arrow_convert.rs` 2,596 → 428; ratchet passes without a waiver. The framing tightening is verified **tightening-only** — `per_message_overhead_bytes` enters via `min(target, ceiling − overhead)`, `streaming_framing_tests.rs` is untouched by `c8c199a`, and a narrow-schema test pins <1 KB overhead with single-slice passthrough at the shipped bounds.

### A coverage gap the audit surfaced — no spec requirement rests on it, filed separately
`cqlite-flight/tests/issue_3096_framing_subphase.rs` is `#![cfg(feature = "observability-testing")]` with **no `[[test]] required-features`**, and no lane enables that feature for `cqlite-flight` — the gate has no flight-obs component, flight-ci runs default features, and `observability-gate.yml` runs only two `cqlite-core` targets. **So the IPC-framing attribution test — the load-bearing deliverable of this PR — executes in ZERO automated lanes.** This is the #1597/#1618 gate-wiring class the gate script itself names.

---

## ⚠️ CERTIFICATION CORRECTION — this body quoted a SUPERSEDED gate

An earlier section of this body records `run-id Nb2Mju` at `176fc3f` with `RESULT: FAIL` and a cited #3245 waiver. **That is not the gate of record for what merged.** It described a tree that no longer existed after the lever-4 revert, the `origin/main` integration, the oracle rebuild, and the lever-6 revert.

**The actual gate of record for the merged commit:**

```
run-id: /tmp/agent-gate.Wq8Y7U
commit: ecbf9b5a55245e6a0833d93150527aa36b09d315   dirty: no
tree-start == tree-end == eaf51fb18a6b            tree-integrity: PASS
29 of 29 components PASS      NO FAILs      NO WAIVER
RESULT: PASS
```

`query-semantics-oracle` **PASSES** in that run — the #3245 waiver this body cites was **withdrawn**, not carried, because the `origin/main` integration brought a per-case fixture resolver (`sstables_root_for_case`) and the two multiclustering cases now execute rather than fail closed (6 passed / 0 failed, zero `REQUIRE_FIXTURES` lines).

Final roborev of record: **`prompt-content: PASS (25/25)`** full-range `c27ca28..d232517`, all eight wrapper assertions PASS; zero code changed between that SHA and the merged `ecbf9b5` (docs-only delta, which per #2964 rule 4 cannot be roborev-certified and does not need to be).

Pre-merge assert: `PREMERGE: OK ecbf9b5a55245e6a0833d93150527aa36b09d315`.

Left the superseded text in place rather than rewriting history, and corrected it here — a merged PR body is a durable record, and a reader who stops at the first gate block would otherwise conclude this merged on a FAIL with a waiver. It did not.


pmcfadin:
### Gate of record re-run at `b253bae` after the two roborev blocker fixes

```
run-id: /tmp/agent-gate.fVsggr
commit: b253bae  branch: issue-3096-arrow-encode  dirty: no
tree-start: b253bae17a48   tree-end: b253bae17a48   tree-integrity: PASS
28 of 29 components PASS
query-semantics-oracle: FAIL (1s)   <- pre-existing, WAIVED, #3245
RESULT: FAIL
```

**Third consecutive run with the same single failure and no other.** Component-for-component identical to `Nb2Mju` (@`176fc3f`) and `XKVg5M` (@`6f6313a`), including `node-bindings: PASS (226s)`. That consistency is corroborating evidence for the closer's "strictly conservative" argument about the F1/F2 fixes: `encoded_len(&data) >= data_body.len()` always holds, so every message the tightened guard admits the old one admitted too, and the tightening cannot have moved any other component. Nothing new broke.

The `query-semantics-oracle` waiver is unchanged and already recorded in the PR body with its clean-`origin/main` (`aac7ce0`) reproduction and its tracking issue **#3245**.

#### Blocker fixes certified by this run
- **F2** (`a7f6425`) — `check_schema_matches_columns` now compares each field against `column_to_field(col)`, the *same* function `build_arrow_schema` uses, so no re-derived mapping can drift. Arrow's `Field: PartialEq` covers name + datatype + **nullability** + **field metadata**; schema-level metadata must also be empty. Each new rejection test asserts the `RecordBatch::try_new`-would-ACCEPT half — except the datatype axis, where Arrow *does* reject, and the test asserts that two-sided truth instead of claiming a false non-vacuity.
- **F1** (`b253bae`) — the wire guard (now `guard_message_within_ceiling`) compares `prost::Message::encoded_len(&data)` — body + `data_header` + `app_metadata` + `flight_descriptor` + tags/varints — against `FLIGHT_DATA_RESERVED_CEILING_BYTES`, i.e. exactly what tonic/grpc-java limit against, with only the 5-byte frame prefix outside it and inside the 65,536 B headroom. The "unconditional" overstatement is removed; the two reserve constants keep only their target-derivation role. Perturbation re-measured with the byte-partition short-circuited: **14,495,257 B** (body 14,495,104 + header 144 + 9 framing) vs the 4,128,768 B ceiling. Header-only is covered by a test whose body is literally 0 — precisely what the old body-only check waved through.
- **`prost` dev-dep → dep is benign**: `prost v0.13.5` was already in the graph via `arrow-flight 53.4.1` and `tonic`, so re-declaring it at the same `0.13` line adds no supply-chain surface. Pinning to arrow-flight 53's own prost line is load-bearing, not incidental — `encoded_len` only equals the length tonic computes if the `Message` impl unifies, so a mismatched prost major would silently measure a different crate's encoding.

**Digest oracle UNCHANGED** across both fixes: `0xd0014e42e893f87f` (CI fixture) and `0x0390bfbb81a23fa1` (measurement corpus, 4,000,000 rows / 31,250 batches), bypass ≡ merge on both.

**C is not being re-run**, and here is the reasoning so it can be checked rather than taken on faith: both fixes move strictly in the conservative direction, so R6's byte-bounded contract is satisfied *more* strongly rather than differently; the newly-rejected band is only messages within the 64 KiB reserve of the 4 MiB limit (observed header 144 B, so no legitimate traffic moves); and the digest oracles — R3's entire evidence base — are byte-identical. A strictly-tightening change with unchanged oracles cannot flip a requirement's status.

Remaining before merge: **a fresh roborev round at `b253bae`** (job 16 covered `176fc3f`, which is no longer HEAD), then the pre-merge SHA assert and `gh pr merge --auto`.

Operational note for #3256: `target/` regrew to **115 GB** and had to be cleared again before this run, or it would have exhausted the disk mid-gate and voided the certification the way the first attempt was voided. That is three times in one delivery.

pmcfadin:
### Gate of record re-run at `c6b4f46` after the three rig-guard fixes

```
run-id: /tmp/agent-gate.piX4o8
commit: c6b4f46  tree-integrity: PASS
28 of 29 components PASS
tooling-tests: PASS (368s)          <- now includes the new test_ws0_report_guards.sh (16 checks)
query-semantics-oracle: FAIL (0s)   <- pre-existing, WAIVED, #3245
RESULT: FAIL
```

**Fourth consecutive run with the same single failure and no other** — across `176fc3f`, `6f6313a`, `b253bae`, and now `c6b4f46`. The waiver is unchanged (clean-`origin/main` `aac7ce0` reproduction recorded in the PR body).

#### The three roborev job-17 findings, fixed and now gate-certified
All three were **guards that did not guard** in the measurement rig — the class this delivery kept surfacing. None had contaminated the shipped figures, and that audit is now recorded in-tree at `docs/reports/ws0-3096-artifacts/rig-guard-audit-2026-08-03.md` (pointer as §7 of the baseline doc) so a later reader need not redo it:

- **F1 — the warm BARE-SCAN arm had no untimed prewarm** (`054c98b`). This is the AC1 ratio **denominator**, so it mattered more than its Medium grade suggests: on a genuinely cold cache, `--reps 1` would have measured partly-cold and reported it as warm. *Audit:* warm rep 1 ran **10.978 s** against a genuinely cold **20.162 s** (+84%) and was +1.6% off the median rep — the cache was demonstrably warm, and the median came from rep 2 regardless. Fix records `prewarm_status` per rep for both arms; the bare-scan arm fails **closed** (the Flight arm records-and-continues) because a partly-cold bare scan shrinks `bare/flight` and could manufacture a win.
- **F2 — cold Flight reps accepted any successful-request count** (`66fd61a`), so requests finishing inside `--cold-step-duration` could blend warm work into a number reported as cold. *Audit:* all recorded cold reps are `requests_ok: 1` at exactly 4,000,000 rows; the `requests_ok: 3` reps are the warm arm at 12,000,000. Fix requires cold `requests_ok == 1` **and** `rows == requests_ok x corpus_rows`, and refuses a `--cold-step-duration` above 5000 ms when cold is selected. Proven firing: a 3-request cold rep is refused naming the count, while the same rep is accepted as warm.
- **F3 — `CorpusIdentity::diff` ignored `seed`, `table`, `rows_per_partition`, `cells_per_row`, `compression_info_present` and the caveats** while claiming "reproduced exactly" (`edaf94c`) — a false verification on the corpus pin that every future comparison rests on, and one that would have propagated to #3232/#3234, which are pointed at this generator as their template. Fix destructures both operands with full field patterns (**a new field is now a compile error**) and compares all 15. Non-vacuity proven by a frozen `diff_pre_review` replica of the old 4-field comparison returning EMPTY for each divergence the new one reports.

New self-test **`scripts/tests/test_ws0_report_guards.sh` (16/16) is wired into the gate's `tooling-tests`**, so these guards are now regression-protected rather than trusting that nobody breaks them. That also closes one of the durability items carried to #3248.

**Nothing measured was restated**: no recorded measurement, artifact figure, superseded-figure label, or the AC1 `unmet` / #3248 re-anchor was altered. **Digest oracle UNCHANGED**: `0xd0014e42e893f87f` and `0x0390bfbb81a23fa1`, identical on both arms.

Remaining: a fresh roborev round at `c6b4f46` (job 17 covered `b253bae`), then the pre-merge SHA assert and `gh pr merge --auto`. Owner has given a standing merge-on-green order.

Operational note for #3256: `target/` regrew to **109 GB** and had to be cleared before this run — **fourth** time in this one delivery. Each occurrence would otherwise have voided a ~40-minute certification.

pmcfadin:
### Owner-ordered SPLIT complete — rig re-anchored to #3272, one library-code blocker fixed

**Diff: 65 files / +13,094 → 48 files / +11,558.** Lite-green at `bbd6797` (`tree-integrity: PASS`, all five components PASS).

#### Removed (13 files, → #3272)
`tools/ws0-corpus-gen/` (whole crate — it was picked up only by the root `Cargo.toml`'s `tools/*` glob, so deleting it de-registers it; root manifest and `cqlite-core/Cargo.toml` reverted to `origin/main`, dropping the `sha2` workspace entry that existed for it), `scripts/perf/`, `scripts/tests/test_ws0_report_guards.sh`, the `agent-gate.sh` hook (script restored to `origin/main` verbatim), and the `cqlite-flight` dev-dependency. **`Cargo.lock` is now byte-identical to `origin/main`.** Verified: `grep` across `.github/ scripts/ xtask/ Cargo.toml cqlite-*/` finds zero dangling references, and `RUSTFLAGS="-D warnings" cargo clippy --workspace --all-targets --all-features` exits 0 with zero diagnostics.

#### The one blocker that stayed here, fixed with a real non-vacuity proof
`wire_partition.rs` — the partitioner budgeted only the Arrow payload while `guard_message_within_ceiling` (correctly) measures the full serialized message, so a wide schema could serialize above the ceiling and be **rejected by the guard** even though smaller slices were legal. The earlier guard fix had *moved* the problem rather than closing it.

`per_message_overhead_bytes` now measures the batch's real IPC `data_header` (via arrow's own `IpcDataGenerator`, one-row probe, **shape-invariance asserted rather than assumed**), plus per-buffer alignment padding, plus a derived 64 B protobuf-framing bound. Bounds become `payload_ceiling = ceiling − overhead` and `payload_target = min(target, payload_ceiling)` — a **tightening only**, so narrow-schema framing and every exact message count in `streaming_framing_tests.rs` are unchanged.

**Non-vacuity** (`a_wide_schema_batch_is_budgeted_with_its_header_instead_of_being_refused`, 2,048 Int32 columns × 64 rows): the pre-fix budget emits **ONE** slice serializing to **753,753 B** against a 600,000 B ceiling — exactly the message the guard rejects — and arrow-flight's own split frames precisely one message at that capacity, so the partitioner is the only thing that can keep it legal. Post-fix it splits and every slice serializes inside the ceiling. A second test pins the padding term in both directions (the bound holds against arrow's real 655,385 B body; an 8-byte assumption under-budgets it).

#### Digest oracle after the split
CI-fixture arm is now **fully self-contained** via a new `cqlite-flight/tests/support/ws0_fixture.rs` (same schema, row synthesis, and write path) — green, digest **UNCHANGED**: `ci-fixture/select-star — 0xd0014e42e893f87f, 500 rows in 4 batches, 12 columns, 12.0 cells/row (identical on both arms)`. The measurement-corpus arm still runs against any corpus root but **SKIPs with an explicit reason naming #3272** when `CQLITE_WS0_CORPUS_DIR` is unset; SET-but-unusable remains a hard failure.

#### R4 is now UNMET here
`tasks.md` records **R4 (“the corpus generator and measurement scripts are COMMITTED and runnable from a clean checkout”): UNMET — re-anchored to #3272**, with a verdict table and the reason, in the same plain form as AC1's re-anchor to #3248. **R4's spec text is untouched** — the requirement stands and only its delivery status is recorded. §0 items whose deliverable was the removed code are unchecked and marked `RE-ANCHORED TO #3272`. Artifact diff is **+16 / −0** (a single note on `baseline-2026-08-03.md`): **no measurement, figure, superseded-figure label, or AC1 statement changed.**

Also green in debug: `issue_3058_forced_path_differential`, `issue_2825_max_batch_bytes_e2e`, `issue_2821_egress_budget_e2e`, `do_get_transport_test`, `cqlite-flight --lib` (411), `issue_3096_framing_subphase` (`--features observability-testing`), `cqlite-core --features arrow --lib export` (87). No `file-size` waiver.

#### Still owed before this can merge
1. Full gate of record at `bbd6797` — **running** (`run-id /tmp/agent-gate.frtsZj`).
2. **C re-audit** — the prior PASS was against a tree containing the rig, and R4's status has changed. Not carried forward.
3. **roborev re-slice** over the reduced diff.
4. An open coordination decision, `REQ-3096-20260804T001500Z-claude-worker-3096`: with `Cargo.lock` gone, exactly **3** paths remain structurally undeliverable — `docs/reports/ws0-3096-artifacts/{abc-interleaved-runs,baseline-results,corpus-identity}.json`, excluded by the repo-tracked `.roborev.toml` `exclude_patterns=['docs/**','*.md']` at **any** diff size. Census 36, of which 33 are real code. The owner-prescribed remedy (smaller base / split) **cannot** reach them, so this is a disposition, not a retry.

Steps 2 and 3 are deliberately not run concurrently with the gate: `spec-auditor` holds `Bash`, and `.roborev.toml`'s `snapshot_dir` is repo-local, so either could dirty the tree and void a ~40-minute cold certification via `tree-integrity` (#2926).

pmcfadin:
### Lever 4 REVERTED per owner ruling A — deferred to #3281

**Diff: 48 files / +11,558 → 42 files / +9,216.** Lite-green at `374f0ff` (`tree-integrity: PASS`), pushed. PR is now `mergeable=MERGEABLE, mergeStateStatus=CLEAN`.

#### Why reverted rather than patched a fourth time
Lever 4 measured **ZERO** throughput (−0.03%, 4/8 rounds positive). Its sole retained justification was wire safety, and that mechanism failed **three consecutive reviews**, each fix inverting the error. The owner's diagnosis is the durable finding: **one number served as both a `target` and a `ceiling`.** "A larger reserve can only cause extra splitting, which is the safe direction" is true of a target and false of a ceiling — **under-budget false-ACCEPTS an illegal message; over-budget false-REJECTS legal input.** One design error, not three bugs, which is why a fourth patch was the wrong bet. Reverting is subtractive and returns to a configuration that was shipping and was never reported broken (arrow-flight's default 2 MiB flight-data target).

**Removed:** `wire_partition.rs`, `wire_partition_tests.rs`, `flight_data_size.rs` (all 7 `const _: () = assert!` guards and every recorded 4 MiB constant), `streaming_framing_tests.rs`. `batch_bytes.rs` is byte-identical to `origin/main`; `streaming.rs::encode_do_get` differs from `origin/main` only by a comment recording ruling A; `record_encoder_error` back to private; `prost` back to `[dev-dependencies]`. **`Cargo.lock` byte-identical to `origin/main`** (0 diff lines) — that property is load-bearing, since it removed a structural roborev-census fence.

**Two implementer judgement calls, both of which I endorse:**
- `streaming_framing_tests.rs` was deleted with **zero assertions kept**, because all eight assert lever-4 constants or drive `wire_partition` — none asserts behavior present on `origin/main`, so no pre-existing coverage is lost. It further **declined to rewrite** two of them to pin arrow-flight's default: that would author new assertions inside a subtractive revert and pin a third party's default with no CQLite behavior behind it. Correct — that belongs in #3281 if wanted, not smuggled into a revert.
- `abc-interleaved` §10.5's "what lever 4 is retained for is WIRE SAFETY" bullet cites the deleted file, so it was **struck in place** (visible strikethrough + note) rather than removed — the record stays legible and no measurement moves.

#### Findings: DELETED WITH THEIR SUBJECT — explicitly NOT waived
Recorded in `tasks.md` under that exact heading, each carried to **#3281** as a constraint any re-introduction must answer:
1. the over-estimated reserve subtracted from a hard ceiling → `RowTooWide` on legal input — *the design error itself, observed as a defect*;
2. `ipc_header_bytes` constructing a fresh `DictionaryTracker` per call, re-serialising and discarding the full dictionary every batch.

Per the ruling, the final full-range round must confirm **neither has a subject** — not merely that neither recurred. Those are different assertions and I will state which I verified.

#### AC record
In the owner's framing, which is cleaner than mine was: **"we measured it at zero AND did not retain it"** — a better honest-negative than shipping it for a safety property that could not be got right in three attempts. **AC1 stays `unmet` → #3248. R4 stays `unmet` → #3272.** `specs/**` untouched, as it has been for this entire delivery.

#### Intact and green
Lever 6 / `EgressBatchPlan` / the egress schema cache; the `Field`-identity schema contract; the `arrow_*` export split (70 arrow tests); the **IPC-framing attribution** — `time_encoder_framing` / `FramingTimedStream` kept, `issue_3096_framing_subphase` passes under `observability-testing`. **Digest oracle UNCHANGED: `0xd0014e42e893f87f`, 500 rows / 4 batches, identical on bypass and merge**; measurement-corpus arm SKIPs naming #3272. Also green: `issue_3058_forced_path_differential`, `issue_2825`, `issue_2821`, `do_get_transport_test`, `cqlite-flight --lib` (388).

#### Three C-required artifact corrections, landed
`abc-interleaved-2026-08-03.md:9` now leads "**MEASURED AT THE SUPERSEDED 4 MiB FLIGHT-DATA TARGET, not at what shipped**" and points to §10; §6 and §10.1's "committed rig" claims now name **#3272** as the delivery vehicle and state it is not in this branch; `measurement-method.md` and `rig-guard-audit-2026-08-03.md` carry the #3272 banner, each naming its own false claim. No measurement, figure, or label altered.

#### Two operational facts
- **`target/` reached 129 GB and free disk fell to 7.6 GB** during the revert's lite gate; the implementer correctly stopped building rather than pressing on. Cleared to 136 GB free before the fresh full gate — **sixth intervention this delivery** (#3256).
- `origin/main` advanced to `4cf8c26` while this ran (2 commits, one being the owner's own #3282 mission-doc §5 correction). Branch is 44 ahead / 2 behind with **no path overlap**; the PR reads `CLEAN`, so no re-integration is required for `--auto`. Flagging rather than rebasing.

**Next:** fresh full gate of record at `374f0ff` — the `bbd6797` gate certified a tree that no longer exists, and a revert is not post-gate polish, so `--delta` does not reach it. Then ONE full-range round once #3229 lands and its worker reports the `Cargo.lock` outcome. Merge remains held by `HOLD: merge after #3229`.

pmcfadin:
### ✅ Gate of record: **RESULT: PASS — 29/29 components, NO WAIVER NEEDED**

```
run-id: /tmp/agent-gate.W610rp
commit: 374f0ff  branch: issue-3096-arrow-encode  dirty: no
tree-start: 374f0ffd7153   tree-end: 374f0ffd7153   tree-integrity: PASS
29 of 29 components PASS
query-semantics-oracle: PASS (0s)     <- previously the one waived FAIL, across FIVE gates
RESULT: PASS
```

**The #3245 waiver this PR carried through five gates is now obsolete and is withdrawn from the record** — not re-justified, withdrawn, because the component passes on its own.

#### I verified WHY it passes rather than banking it
A component flipping FAIL→PASS with no traceable cause is the signature of a guard that quietly stopped guarding, and this delivery has produced four of those. So this was checked, not celebrated:

- **The resolver changed, and in exactly the right way.** At HEAD it is `sstables_root_for_case(case: &Case)` — **per-case**, searching candidate roots for that case's specific `{prefix}-Data.db`. The old `sstables_root(keyspace)` locked onto the first root whose `<keyspace>/` directory existed and never fell through per fixture. That is precisely #3245's AC1 remedy, and it arrived through the `origin/main` integration.
- **The fixture is reachable checkout-relative**: `test-data/datasets/sstables/test_da/multiclustering_table-fd74ad508d2311f1a29b6d2c15dcffdf` is present in the worktree, so the per-case resolver finds it in-repo even though `/data/datasets` still lacks it.
- **The cases genuinely execute — they do not skip.** Failing run (`frtsZj`): `5 passed / 1 failed` with two `REQUIRE_FIXTURES: … fixture dir multiclustering_table-* absent` lines. Passing run (`W610rp`): `6 passed / 0 failed` and **zero** `REQUIRE_FIXTURES` lines. Same 6 tests; the count moved 5→6 because the guard had nothing to fire on, not because it was disabled.

Note **#3245 remains OPEN** and should stay so — the resolver fix landed on `main`, but that issue also covers `fetch-datasets.sh`'s `rm -rf` destroying committed fixtures. This PR is simply no longer affected.

#### Three consequences
1. **No waiver anywhere in this PR.** The cite-and-waive posture is gone.
2. **#3259 no longer blocks this PR.** `--delta` refuses to anchor on a waived FAIL; the anchor is now a genuine `PASS`, so any post-gate docs polish re-certifies cheaply instead of costing another full gate. (#3259 stays open as a real mechanism gap for waived PRs generally.)
3. This is the strongest certification posture of the delivery: clean gate at the reverted HEAD, C re-audited, both live blockers **deleted with their subject**.

#### Where this stands
| | |
|---|---|
| Gate of record | **PASS**, 29/29, `tree-integrity: PASS`, `374f0ff` |
| C intent audit | re-audited post-split; R1 `unmet`→#3248, R4 `unmet`→#3272, R2 partial, R3/R6 satisfied |
| Blockers | none live — both `wire_partition` findings deleted with their subject → #3281 |
| PR mergeability | `MERGEABLE` / `CLEAN` |
| **Still owed** | **ONE full-range roborev round**, after #3229 lands and its worker reports whether a #3096-shaped diff yields numeric `PASS`, a `Cargo.lock` residual, or `FAIL` |
| Merge | held by `HOLD: merge after #3229` |

What ships, restated honestly: **both levers measured at ZERO and neither was retained for a throughput claim** — lever 4 is reverted outright (#3281), lever 6 kept only for being strictly less work per batch. The durable value is the `arrow_convert.rs` split (2,596 → 428, verified behavior-neutral item-by-item), the `Field`-identity schema contract, and the **IPC-framing attribution at 313.0 ns/row** — previously attributable to nothing, and now the ceiling on what any framing change can buy. AC1 remains `unmet`, 16.3% short, re-anchored to #3248.

Issue #3253, opened 2026-08-03
roborev's token-accounting vacuity tier is INVALID: a review that read ZERO diff bytes scored 426k in / 382k cached — inside the documented "genuine, large" band — and vacuity-tier2 PASSed it


pmcfadin:
Found while trying to certify #3234 (BTI perf corpus) on PR-sized diff. **The snapshot-externalization mechanism itself is already filed (#3157 / #3208 / #3252). This issue is the NEW fact those three do not have, and it inverts one of their conclusions: the token-accounting oracle that CLAUDE.md and `vacuity-tier2` rely on is INVALID — a review that read ZERO diff bytes scored squarely inside the documented "GENUINE, LARGE" band.**

## The measurement (3 successive rounds, same branch, growing diff)

| round | job | census | prompt bytes | `diff --git` in prompt | delivery | codex diff bytes read | tokens in/cached/out (reasoning) | roborev verdict | wrapper |
|---|--:|---|--:|--:|---|--:|---|---|---|
| 1 | 14 | 7 files, +2064/-2 | 101,818 | 6 | **INLINE** | n/a (inline) | 174,772 / 139,264 / 10,193 (9,338) | **F — 4 real findings** | `prompt-content: PASS (6/6)`, FAIL(findings) |
| 2 | 15 | 10 files, +4970/-2 | 6,547 | 0 | snapshot pointer | **0** | 439,079 / 397,056 / 3,345 (2,210) | **P — "No issues found."** | `prompt-content: FAIL (9/9 absent)` |
| 3 | 16 | 11 files, +5827/-2 | 7,070 | 0 | snapshot pointer | **0** | 425,973 / 382,208 / 3,510 (2,363) | **P — "No issues found."** | `prompt-content: FAIL (10/10 absent)` |

## Why the token oracle is invalid (the new, load-bearing evidence)

`roborev log 16` is the oracle — it is the agent's own event stream, and it shows codex never obtained a single diff byte:

```
item_1 command_execution  wc -l …/roborev-snapshot-content.diff && sed -n '1,240p' …   -> status: failed
   ERROR codex_core::tools::router: CreateProcess { message: "Rejected(\"approval request failed\")" }
item_2 command_execution  sed -n '1,240p' …roborev-snapshot-content.diff              -> status: failed
item_4 command_execution  sed -n '1,240p' …roborev-snapshot-content.diff              -> status: failed
[...]

Issue #3230, opened 2026-08-03
Clustering-seek plane still reads Data.db through the MADV_RANDOM point mmap — #2876's split missed bti_point.rs AND big_promoted.rs (both formats)


pmcfadin:
Found while answering #3029 (WS3) scope item 2 — *"does #2876's plane split cover BTI?"* **It does not.** Filed separately per that issue's explicit instruction, and widened: the leak is **not BTI-specific**, it is the *clustering-seek plane* in **both** formats.

## What #2876 fixed, and what it missed

#2876 split the read planes: `point_source` reads through a dedicated Data.db mmap advised `MADV_RANDOM` (readahead suppressed — correct for point reads), and `scan_positional_source` reads through the shared, **never-advised** mapping (readahead allowed — correct for sequential walks).

The split was applied only to the walks that thread a **caller-supplied** `source: &dyn ReadAt`:

- `summary_scan/mod.rs:315` — `let scan_source = self.scan_positional_source.clone()`
- `full_index_scan.rs:240`, `full_index_stream.rs:281`
- `scan_stream_windowed_read.rs:130, 166, 301, 340, 389`

Two read paths were left reading `self.point_source` **directly**, and the code says so out loud — `data_access/compressed_offset.rs:21-22`: *"`bti_point.rs` / `big_promoted.rs` read `point_source` directly and do not route through here."*

### (a) BTI clustering slice — reads a forward sequential walk through the readahead-suppressed mapping

The user-facing `scan_partition_clustering` (`cqlite-core/src/storage/sstable/mod.rs:1686`) → `scan_single_partition_clustering` (`data_access/bti.rs:99`) → `bti.rs:212` → `bti_decompress_and_parse_target_all` (`data_access/bti_point.rs:488`) → `bti_pull_decompressed_chunk` (`bti_point.rs:688`), which constructs:

```rust
ChunkSource::new(self.point_source.as_ref(), …, NS_BTI_CHUNK, …)   // bti_point.rs:713
```

The uncompressed fallback `point_read_whole_section` (`bti_point.rs:138`) is the same — `point_source` at `:142, :157, :173`.

[...]

Issue #3157, opened 2026-07-30
roborev silently returns a vacuous "No issues found" once the branch diff exceeds the inline threshold: the diff is externalized to a snapshot file that codex cannot read on this fleet (enterprise-forced approval policy + broken bwrap)


pmcfadin:
## Symptom

A full-branch `roborev review --branch --base origin/main --wait` returns **exit 0 and
`No issues found.`** having read **zero diff bytes**. The output is textually indistinguishable from a
genuine clean pass.

Observed on **#3094 / PR #3122, roborev job 5, HEAD `b633262`**.

## Mechanism: the diff is externalized to a snapshot file the reviewing agent cannot read

1. The branch diff was ~**200 KB**, which exceeded roborev's inline-diff threshold.
2. roborev therefore wrote the diff to
   `.roborev/roborev-snapshot-…/roborev-snapshot-content.diff` and instructed the reviewing agent to
   **read the file** instead of embedding the diff in the prompt.
3. codex then failed **every** read attempt — 9 tries. The first failure mode was
   `bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted`; the rest were
   `Rejected("approval request failed")`.
4. codex produced its verdict **from nothing**, and roborev reported it as `No issues found.` with
   exit 0.

This is a **size-triggered** failure, so it arrives silently as a branch grows. On PR #3122, rounds 3
and 4 were unaffected — their diffs (79 KB / 172 KB) still fit inline. Round 5's ~200 KB did not.
Every sufficiently large PR is therefore reviewed vacuously unless someone checks by hand.

**Workaround that did work:** a **delta review of only the new commits** (38.5 KB) fit inline and
produced a real verdict with real findings. That is the fallback to document, with a stated size
ceiling.

## Root cause is environmental, not the diff

**VERIFIED on this box just now** (canary probe: a 1-line file, `codex exec`, asked only to `cat` it —
codex failed and answered *"Unable to read the file: the filesystem sandbox failed, and elevated access
was unavailable."*):

- `codex-cli` **0.146.0** (note: CLAUDE.md documents 0.145.0 — the fleet has moved again).
- Session header reports `approval: on-request`, `sandbox: read-only`.
- Enterprise-managed policy forces the approval policy, verbatim from the run:
  ```
  warning: Configured value for `approval_policy` is disallowed by requirements; falling back to
  required value OnRequest. Details: invalid value for `approval_policy`: `Never` is not in the
[...]

Issue #3100, opened 2026-07-28
AWS test team: re-run the WS0 head-to-head on the byte-identical corpus with a post-#3058 binary + the never-measured C(N) concurrency sweep (N=1..16)


pmcfadin:
## Summary

**#3058 took Flight `do_get` from 53,873 -> 210,192 rows/s per physical core (3.90x)**, closing the bare-scan gap from 5.73x to 1.49x. But that run used a **locally generated UNCOMPRESSED corpus** (`Data.db` sha256 `0185909d…12f0b1`), and the stock-Cassandra baseline was measured on a **different, LZ4-compressed corpus** (`22d9ae22…ce922c`). **The two numbers cannot be divided.**

This issue re-runs the WS0 head-to-head on the **byte-identical WS0 corpus** with a **post-#3058 binary**, and adds the concurrency sweep the program has never measured. It converts work already delivered into a claim we can publish.

## Why this cannot be done on the delivery box
#3058's own report states it: *"no Cassandra on the delivery box, so the stock-Cassandra leg of the WS0 head-to-head is not re-established locally,"* and the WS0 absolute (>= ~280,000 rows/s) plus the AC5 digest `0x4903ffa446163c4b` **remain owed** against a machine holding the byte-identical corpus. Owner has assigned this to the AWS test team.

## The reproduction contract already exists — do NOT re-derive it
Everything is committed on `main`:

- `docs/reports/ws0-3026-artifacts/ws0-results/head-to-head-method.md` — the reproduction contract (Arm 1 Cassandra / Arm 2a bare scan / Arm 2b Flight, and why `count(*)` is a legitimate read-path comparator, proven from a profile in F2)
- `docs/reports/ws0-3026-artifacts/ws0-corpus/` — `gen-corpus.sh` (parameterized), `rerun.sh`, `verify-claims.sh`, `measure-sstable.py`, `trace-scan.bt`, `cassandra.yaml.diff`, `schema-as-created.cql`
- `docs/reports/ws0-cassandra-baseline-2026-07-27.md` — the Arm-1 numbers to reproduce

Reference machine: **c7i.4xlarge — 16 vCPU / 8 PHYSICAL cores**, 30 GiB RAM. Cassandra 5.0.8, one disclosed deviation (`MAX_HEAP_SIZE=8G`), chunk cache OFF (stock).

### Part A — the owed head-to-head, post-#3058
[...]

PR #3255, opened 2026-08-03
test(#3234): commission a profileable BTI (da) perf corpus + gen-perf-corpus-bti.sh, a committed Cassandra-written golden, and a byte-identical drift control

### LINKED ISSUE #3234: Commission a profileable BTI (da) perf corpus + author gen-perf-corpus-bti.sh — prerequisite blocking WS3 (#3029) and WS4 (#3030)
**Prerequisite issue split out of #3029 (WS3, first BTI read-path profile) on the owner's Option-A decision.** WS3's measurement arm (its scope items 1, 4, 5) is blocked on this and holds behind it. **One fixture, three consumers: #3002 (BTI Rows.db root base — now closed), #3029 (WS3 profile), #3030 (WS4).** That coordination was called for in #3029's Notes but was never recorded on #3030; recording it now.

## The gap

There is **no BTI (`da`) corpus anywhere near throughput-profile scale**, and none can be generated on the current worker box. Exhaustive enumeration of every `da-*-bti-*` object reachable:

| Fixture | Data.db | Rows.db | Partitions / rows |
|---|--:|--:|---|
| `test_da/wide_table` | 28,129 B | 760 B | 3 / 900 |
| `test_da/multiclustering_table` | 10,660 B | 618 B | 3 / 468 |
| `test_da/simple_table` | 168 B | **0 B** | — |
| `test_da/ttl_table` | 76 B | **0 B** | — |
| `test_da/collection_table` | 160 B | **0 B** | — |

Total `test_da` = 2.1 MB. A warm scan of the largest completes in microseconds; a throughput profile needs a ≥10 s sustained window — roughly **6 orders of magnitude** short.

`test_da/multiclustering_table` (landed via #3032 / PR #3210) exists in the repo checkout but is **absent from `/data/datasets`** — the fetch pin predates it, so `CQLITE_DATASETS_ROOT=/data/datasets` cannot see it. That is a second, independent problem worth fixing in the same pass.

**Also missing: any BIG corpus.** `/data/ws0`, `/home/ubuntu/ws0` (the #3100/#3217 `WS0_STAGE`) and `/home/ubuntu/corpus-3068` do not exist; nothing on the box exceeds 100 MB. So the #3100/#3217 corpora are not reproducible locally either.

## Why the 8 MiB threshold makes this a hard gate, not a nice-to-have

`MADV_RANDOM` is applied only when `file_size >= POINT_MMAP_MADV_RANDOM_MIN_BYTES = 8 MiB` (`cqlite-core/src/storage/sstable/reader/mod.rs:348`); below it, `point_read_mmap` returns `scan_mmap.clone()` — **both read planes become the same mapping.** So #3230's clustering-seek-plane A/B is not merely noisy on the current fixtures, it is **structurally zero**. Any BTI Data.db used for plane work must exceed 8 MiB by a comfortable margin.

## Why no existing generator can do it

| Script | Format | Scale | Blocker |
|---|---|---|---|
| `test-data/scripts/gen-perf-corpus-3068.sh` | **BIG only** (`nb`, hardcoded `nb-*` globs `:399-400`) | 12.0M + 1.2M rows — the only throughput-scale generator | docker-driven (`:68`, `:315-400`) |
| `test-data/scripts/gen-multiclustering-bti.sh` | BTI, compound `PRIMARY KEY (pk, bucket, seq)` | 468 rows (correctness) | docker-driven |
| `test-data/scripts/gen-wide-bti.sh` | BTI, single `int` clustering | 900 rows (correctness) | docker-driven; stale macOS `OUT` default (`:18`, overridable) |
| `test-data/scripts/gen-wide-big.sh` | BIG | correctness | docker-driven |

**There is no BTI-capable throughput-scale generator. One has to be written.**

Environment on the current worker box: **no `docker`, no Cassandra, no JDK, no `cassandra-stress`, no `cqlsh`, no `sstabledump`.** So #3217's non-container tarball recipe cannot run as-is either. Disk is fine (266 GB free on `/data`); 30 GB RAM makes a 2–5 GB corpus comfortably page-cacheable for a warm arm.

**Two non-default yaml settings are mandatory** and are the reason a stock node will never emit `da` — see the owner's notes on #3029 (2026-07-27, 2026-07-28) and `gen-multiclustering-bti.sh:110-113`: `storage_compatibility_mode: NONE` **and** `sstable.selected_format: bti`, applied and then restarted.

## Scope

1. **Provision a generation environment** — Docker + `cassandra:5.0.2` (or a documented tarball+JDK path), on whichever box will host generation. Record the recipe so it is reproducible, not a one-off.
2. **Author `test-data/scripts/gen-perf-corpus-bti.sh`** — fork `gen-perf-corpus-3068.sh`, graft in the BTI yaml flip from `gen-multiclustering-bti.sh:110-113`, use a **wide multi-component clustering** schema (`PRIMARY KEY (pk, bucket, seq)`), and replace every `nb-*` glob with `da-*`. Same `OUT`/`CQLITE_DATASETS_ROOT` conventions and `--verify-only`-style ergonomics as its siblings.
3. **Generate and pin the corpus** — target **2–5 GB `Data.db`, well above 8 MiB per SSTable**, wide partitions, multi-component clustering, LZ4 `chunk_length_in_kb=16` to match #3100/#3217's shape so BTI-vs-BIG stays a controlled comparison. Register it in the dataset pins/`references.yml` so `fetch-datasets.sh` delivers it.
4. **Fix the `multiclustering_table` fetch-pin gap** so the #3032 fixture is reachable via `CQLITE_DATASETS_ROOT`.
5. **Consider regenerating a BIG counterpart** (or pinning #3068's) so BIG-vs-BTI can be run on one box. Note per `ws0-3217-report.md:3-7` the #3100/#3217 numbers are from a comparable machine class, so this is a convenience rather than a correctness requirement — call it explicitly either way.

## Acceptance criteria

1. `gen-perf-corpus-bti.sh` exists, is documented, and produces `da`-format SSTables — asserted by the generated TOC/descriptor, not by assumption. A run that silently emits `nb` is a **hard failure** (this is exactly what the two yaml settings guard).
2. The generated corpus has **≥ 1 SSTable with `Data.db` > 8 MiB** and a **non-empty `Rows.db`**, with wide partitions and a multi-component clustering key. Report actual sizes, partition count, row count, and rows/partition distribution.
3. A warm full scan of the corpus sustains **≥ 10 s** — stated with the measured duration, so the fixture is demonstrably profileable rather than assumed to be.
4. The corpus is **fetchable**: `bash test-data/scripts/fetch-datasets.sh` delivers it and the printed `CQLITE_DATASETS_ROOT` export line makes it visible. `--verify-only` recognizes it.
5. `sstabledump` goldens are generated for at least a bounded subset, so the corpus can back parity work and not only throughput work (#3042: an oracle must be **Cassandra-written bytes**, never CQLite's own output).
6. The generation recipe is **reproducible from the committed script + docs alone**, on a clean box. #3100's driver script was never committed and #3217 had to rebuild it (`ws0-3217-report.md` §2.5) — do not repeat that.
7. **Consumers unblocked and told:** drop the corpus location + export line on #3029 (WS3) and #3030 (WS4).
8. Generated fixtures stay gitignored; only goldens/pins are committed, per existing convention.

## Notes

- Routing: **design-driven** in the sense that the generator's shape is a design choice with three downstream consumers — but there is no CQLite behavior change here, so it is closer to test infrastructure. Owner to route; I would not block on an OpenSpec cycle for a generator script.
- **Do not derive the BTI schema/shape from CQLite's existing fixtures.** Derive from Cassandra 5.0.8 (`BtiTableWriter`, `RowIndexWriter`) — #3002 is the cautionary tale, where a test written from the implementation pinned the buggy value.
- Prior art for the profiling harness this feeds: `docs/reports/ws0-3217-artifacts/harness/` (committed, format-agnostic, directly reusable).
- Tooling gotchas that apply to the eventual profile, recorded so they are not rediscovered: `samply` records **0 samples** on these kernels, **and** `perf record --call-graph=dwarf` **hangs past 120 s** on this binary (`ws0-3217-report.md:168`, `:211`) — use frame pointers with #3217's build flags. Never divide `/proc/<pid>/io` `read_bytes` by `syscr`.
- Blocks: #3029 (WS3 measurement arm), #3030 (WS4). Enables the A/B for #3230 (clustering-seek plane on the advised mapping) and the BTI measurement in #3233 (BTI denied the Flight bypass arm). Program epic #2817; umbrella #3023.

pmcfadin:
Closes #3234.

Commissions the first **profileable BTI (`da`) perf corpus** and commits the generator that produces it, so the fixture is reproducible from the repo rather than rebuilt from scratch a fourth time (#3100, #3217, and this issue each rebuilt one).

## Owner-adjudicated routing — `1C · 2Y · 3Y · 4B-bounded`

The issue's two 🧭 manager orders asked to mirror `tools/ws0-corpus-gen` from #3096 with a *byte-identical-across-3-runs* contract on a Cassandra-written corpus. Both premises were checked and failed:

- **`tools/ws0-corpus-gen` never existed** — `git log --all` empty; #3096 is an OPEN Arrow-encode issue that landed nothing; `scripts/perf/` and `docs/reports/ws0-3096-artifacts/` likewise absent.
- **Byte-identical + Cassandra-written are mutually exclusive** — *measured*: two same-seed Cassandra runs differ by **382 B** (write timestamps as VInt deltas). And a CQLite-written corpus forfeits AC5 (#3042: an oracle must be Cassandra bytes) *and* the LZ4-16KiB shape (#1406, uncompressed-only).

Owner ruled **1C** (Cassandra-written corpus **+** a CQLite-written byte-identical drift control), **2Y** (drop AC4 → #3245 owns `fetch-datasets.sh`), **3Y** (no fetch asset; manifest-pinned per #3068's precedent), **4B-bounded** (commit a small golden only at ≤5 MiB with no 2 GB regeneration).

## Acceptance criteria

| AC | Status | Evidence |
|---|---|---|
| 1 — `da` asserted from descriptor/TOC, `nb` a hard failure | ✅ | all 27 match `^da-\d+-bti$`, format set `{da}`, **zero `nb-*`**; TOC ×27 has `Partitions.db`+`Rows.db`, **`Index.db`/`Summary.db` absent from every one** |
| 2 — ≥1 `Data.db` > 8 MiB, non-empty `Rows.db`, wide multi-component clustering | ✅ | **27/27** over 8 MiB, largest **81,151,240 B**; `Rows.db` 53,746–145,714 B (zero empty); `PRIMARY KEY (pk, bucket, seq)`; **17,299** partitions / **13,200,000** rows, rows/partition 200–4000 (mean 763.05) |
| 3 — warm scan ≥ 10 s, stated with the measured duration | ✅ | **127.163 s**, 13,200,000 rows **verified**, 103,804 rows/s (warm pass 126.231 s, passes agree within 1.2 %) — **12.7× the floor** |
| 4 — fetchable / pin registered | ⛔ **dropped by owner (2Y)** | duplicate of #3245, which owns `fetch-datasets.sh`; would have collided on the same diff |
| 5 — `sstabledump` goldens for a bounded subset | ✅ | committed **1.83 MiB** golden for a dedicated small table (below); the 153.3 MiB full-generation golden stays on-demand, verified 711 partitions / exactly 500,000 rows vs `Statistics.db` |
| 6 — reproducible from committed script + docs on a clean box | ✅ | `test-data/scripts/gen-perf-corpus-bti.sh` + cookbook; **~7.3 min** end to end; no path outside the repo except the `--out` corpus dir |
| 7 — consumers told | ✅ | #3029, #3030, #3230 cross-linked **before** that work starts, incl. the route limitation |
| 8 — fixtures gitignored, only goldens/pins committed | ✅ | corpus gitignored; golden + manifests + `.cql` committed via `git add -f` per the `test_da` convention |

## What's in it

- **`test-data/scripts/gen-perf-corpus-bti.sh`** — Cassandra-driven, docker (`sudo -n docker` default, as the #3068 sibling), `cassandra:5.0.2`. Grafts the mandatory BTI yaml flip (`storage_compatibility_mode: NONE` **+** `sstable.selected_format: bti`) with **fail-closed grep verification** — a `sed` that silently misses emits `nb` with no error, which is the real hazard. Recorded-seed → reproducible CSV → chunked `COPY` (160× faster than an INSERT file; a seed pins the row set for the manifest identity).
- **`cqlite-core/examples/bti_perf_scan/{main,manifest,scope}.rs`** — the AC3 harness (split into a module directory to stay under the 800-line campsite threshold). Row-count assert **ON by default** from the authoritative manifest (fail-closed exit 8 if absent); distinct exits for zero rows / count mismatch / sub-floor / scan failure; `--no-expect-rows` stamps `*** UNGUARDED ***` on the RESULT line.
- **A committed Cassandra-written BTI golden** — `test_da/wide_multiclustering_small-47f6a3008f6911f1bc0f8df8badcc262/`, **600 rows / 5 partitions**, golden **192,935 B** (`Data.db` 97,780 B, `Rows.db` 197 B). Resized down from an initial 6,000-row / 1,916,747 B build to match #3032's 468-row / 121 KB precedent at the identical clustering shape. This is a *legitimate* oracle (#3042) and is exactly what #3002 showed we lacked.
- **A CQLite-written drift control** — `cqlite-core/tests/issue_3234_cqlite_written_bti_drift_control.rs`, production `SSTableWriter` with `SSTableFormat::Bti`, uncompressed (#1406), Murmur3-ordered. **3-run byte identity VERIFIED** across all 7 components and stable across processes. Documented **drift-control-only, never a correctness oracle**.
- **250 hermetic self-test cases** wired into the gate's `tooling-tests` — `test_gen_perf_corpus_bti` **158/158** and `test_bti_perf_scan` **127/127**, each with a declared case-count floor so a silently skipped case fails the suite.

## ⚠️ Read this before citing the throughput number

```
access_path:   fallback_full_scan (partition_key_not_fully_constrained)
storage_route: generation_merge::stream_generations_for_read
generations:   27
```

**103,804 rows/s measures the compaction-style `generation_merge` stitch + `KWayMerger`, NOT the BTI mmap/trie plane.** With 27 generations and a resolved schema, `scan_stream` routes into `generation_merge::stream_generations_for_read` (`storage/sstable/mod.rs:2139-2188`), whose producers re-open every SSTable with `use_mmap=false` / Buffered. The corpus finally clears the 8 MiB `POINT_MMAP_MADV_RANDOM_MIN_BYTES` bar that made #3230's plane A/B *structurally zero*, so that A/B is now possible — but **unmeasured here**. Recorded in `docs/development/dev-cookbook.md`, in the harness's runtime output, and on all three consumer issues (#3029/#3030/#3230). **Deliberately NOT a manifest field**: `read_path_measurement_scope` was removed under the observed-or-absent rule (a throughput figure is a harness measurement that no corpus byte substantiates, so it cannot be an observed manifest field); the harness prints its route on **every** run so the figure can't be quoted without it.

Related hazard: BTI `scan_stream` pre-materializes the whole reconciled table (#1577) — a single generation or a missed schema resolution on a multi-GB corpus is the #3068 livelock. Always use `test-data/scripts/perf-run-contained.sh`.

## Review evidence — and a reason to trust it that didn't exist before

**19 findings closed** across roborev (jobs 14, 17), `rust-reviewer`, and `coverage-reviewer`. Two are worth naming:

- **HIGH (roborev job 17)** — `--out` rejected only the literal `/`, so `/tmp/..` or a symlink resolving to `/` passed validation before a privileged `rm -rf "$OUT/cassandra-data"`. Now canonicalized with `realpath -m` before any mutation, all four destructive sites re-assert `assert_under_out`, with hermetic refusal cases.
- **A previously-vacuous assertion** — `check_reject`'s no-write check counted a directory the generator was never given, so **11 cases asserted a property they could not fail**. Fixed and **mutation-proved** (breaking the generator now reds 12 cases).

**roborev rounds 2 and 3 reported `findings: NONE` having read ZERO diff bytes** — the diff exceeded roborev's 200 KB prompt cap and was externalized to a snapshot codex's sandbox could not read. Raising the host cap to 4 MB made round 4 a controlled experiment on the identical commit: `prompt-content` FAIL→PASS (10/10), findings NONE→**4 incl. the HIGH above**. The vacuous round posted **2.3× the input and 3.75× the cached tokens** of the genuine one, so token-band triage would have ranked the review-of-nothing as more thorough. Filed as **#3253 (P1)**, with corrections on #3157/#3208/#3252.

**roborev job 26 (round 12) found three more instances of the exactness class this PR is largely about — all three in code this PR added — and all three are fixed here.** (F1) `scope.rs` counted a **directory or a symlink** named `*-Data.db` as a generation and swallowed per-entry `read_dir` errors: measured on the pre-fix binary, one of each beside the single real generation made the harness report `generations: 3` and attribute its own figure to `generation_merge::stream_generations_for_read`, with `rows_scanned` unchanged at 468 and exit 0 — the instrument misreporting its own measurement. (F2) `assert_corpus` globbed `da-*-bti-*`, so a corpus carrying `da-x-bti-Statistics.db` printed **`VERIFY-OK`**; every component basename is now validated against `^da-[0-9]+-bti-`. (F3) `documented_sstable_dir` took the **first** matching `tables[]` record, so a manifest describing the table twice made the authoritative ingest scope a function of array order. Each fix ships with its control **observed firing** against the pre-fix code (6 new cases red on the pre-fix binary; the pre-fix script printing `VERIFY-OK` on the malformed corpus), self-test floors raised 154→158 and 77→92. A 298-site audit that missed three sites of its own class in its own output is the honest summary, and it is the argument for the stop rather than against it. Mechanizing the class is deferred to a follow-up issue.

`minimal-build` and clippy were both verified safe by measurement on clean `origin/main` before the gate of record — **no waiver is claimed in this PR**.


pmcfadin:
**flow-closer endgame — auto-merge armed (deferred finalize).**

**Gate of record (the ONE full run):** `RESULT: PASS`, 30 PASS / 0 FAIL, at `19f97c5`.
`run-id: /tmp/agent-gate.3VwH0K` · `tree-start == tree-end = 19f97c586daa` (digest `b03388a8e77a`) · `tree-integrity: PASS` · `schemas: 6/6` · summary-file `/tmp/gate-3234-r3.txt`.

**Integrated `origin/main` (`7e47282`) before that gate.** The PR had gone `CONFLICTING/DIRTY`, so integration was mandatory — `gh pr merge --auto` cannot land a DIRTY PR. Sole conflict was `scripts/agent-gate.sh`, the gate script itself, so it had to be resolved *before* the gate of record. Resolved as a **union, comments-only, zero production logic**: main's #3249 perf-capability paragraph + this branch's three #3234 guard paragraphs. Verified main's `perf=` token and its two `tooling-tests` executables survived alongside this branch's three guard blocks. Used `git merge` rather than rebase (74 commits, 6 touching the conflicted file; squash-merge makes the resulting tree on `main` identical with one resolution instead of six). The telemetry ledger auto-merged as a true 454-line union — zero peer records dropped.

**Final roborev pass (job 28) — genuine, not vacuous:**
`prompt-content: PASS (28/28 code census paths present)` · `reviewed-sha: 7e47282..19f97c5` (a real range, base ≠ head) · census 31 files +11731/-16 · tokens input=437946 cached=105728 output=8156 · `push-assert/census-check/code-free/job-record/sha-assert/review-completed: PASS` · `vacuity-tier1/tier2: PASS` · `findings: PRESENT (4)`.

**Finding triage** (owner's pre-committed disposition — default to follow-up except (i) High/Critical or (ii) a fail-closed hole that can misreport a published number). No High/Critical. **Finding 1 hit exception (ii) and was fixed pre-merge in `6c3c4c6`:** `bti_perf_scan` observed no SSTable FORMAT anywhere, so a BIG (`nb`) corpus with matching generation ids and row count scanned clean and was reported as a **BTI** measurement — the instrument misreporting which format its own published AC3 figure describes. Concrete, not adversarial: a BTI-configured Cassandra 5.0 node writes its own system tables in BIG, and the corpus tree holds **11 `nb-*-big-Data.db`** files beside the **27 `da-*-bti-*`** generations under test (27 = the published `generations: 27`). `assert_bti_descriptors` now refuses any non-`da-<gen>-bti-Data.db` descriptor as `OPEN_FAILED` (3) — a refusal, not a filter, so the count cannot disagree with the discovery count it is cross-checked against.

**Re-cert of the fix:** `--lite RESULT: PASS` at `6c3c4c6` (`tree-integrity: PASS`) + the diff-relevant target `test_bti_perf_scan.sh` **129/129 ALL PASS**, floor raised 127 → 129 for the new negative control (which greps the new diagnosis, so it cannot pass via the generation-count mismatch that also exits 3). No re-review, per the owner's disposition: a targeted fix to a named blocker is certified and there is no round 16.

**Scope note for the record:** the full gate of record is at `19f97c5` and the merged head is `6c3c4c6`; that fix commit postdates it. The merged head is therefore certified by *gate-of-record + `--lite` + the harness's own 129-case executor*, not by a full gate at `6c3c4c6` — stating it plainly rather than implying the gate covers the tip.

**Pre-merge:** `PREMERGE: OK 6c3c4c6cbf093eb5d280e16a282929efae9d671a` (exit 0). `HOLD:` re-read on #3234 + #3255 — none. No `ci:waive:*` applied.

**Armed** `gh pr merge --auto --squash --delete-branch` (`autoMerge=true method=SQUASH`). `required` was **pending** at arm time with zero failing checks, so GitHub owns the green wait; branch protection (`contexts=["required"]`, `enforce_admins=true`) means this cannot land against an unchecked head. Deferred findings 2/3/4 batched into **#3279**. `flow-finalize` + the telemetry ledger PR run on a later wake that first confirms `mergedAt` is non-null.

Issue #3312, opened 2026-08-04
roborev wrapper: prompt-content: FALSE-FAILs every large-diff review — roborev delivers the diff by snapshot PATH, and the check only reads inline headers


pmcfadin:
## Summary

`scripts/flow/roborev-review.sh`'s `prompt-content:` check models only ONE of roborev's two diff-delivery modes. When a diff is large enough that roborev writes it to a **snapshot file** and instructs the reviewer to read it, the prompt itself contains **zero `diff --git` headers** — so `prompt-content:` reports `FAIL (N/N code census paths absent from the prompt)` on a **genuine, complete review**.

This is a **false FAIL**, and per doctrine (`CLAUDE.md`: "any non-PASS terminal `RESULT` … is a failed review round and a blocked merge") it blocks a merge that should not be blocked. It is the mirror image of the defect class the key exists for: `prompt-content:` was built to catch "the reviewer never got the diff", and it now also fires when the reviewer definitely did.

## Evidence — measured on #3272 (PR pending), job 6836

Wrapper verdict:

```
census: 23 files, +6561/-1
tokens: input=1467215 cached=1352192 output=6078
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: FAIL (21/21 code census paths absent from the prompt)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: PRESENT (4)
```

The review was **genuine**, not vacuous, on every independent signal:

- **1.47M input / 1.35M cached / 6.1k output over ~3.5 min.** The documented vacuous baseline is 18.7k input / 0 cached / 53–56 output in 8s. `vacuity-tier1`/`tier2` both PASS.
- Four findings with **real `file:line` locations** in files of this diff (`ws0-baseline.sh:726`, `ws0_validate.py:212`, `ws0-baseline.sh:444`, `scan_bench.rs:159`).
- One of them (**partial sysctl restore reported as success**) was **independently found by `rust-reviewer` at the same line** — corroboration by a second reviewer that could only happen if codex read the code.
- `job.status = done`, `job_type = range`, `git_ref = a0462ec7…..32e3ffea…` (correct base..head).

The cause, read from the prompt roborev actually sent (`roborev show 6836 --json | .prompt`, 8,205 chars, `diff --git` count **0**), verbatim tail:

```
### Combined Diff

(Diff too large to include inline)

The full diff has been written to a file for review.
[...]

Issue #3296, opened 2026-08-04
P0: the gate of record is RED on macOS — roborev-lints FAILs on pristine origin/main (7 cases: GNU-only sed -i, /private/var canonicalisation), after #3262 correctly closed a fail-open exit path that had hidden 3 of them


pmcfadin:
Part of #2817 (the gate of record is the program's certification instrument). Blocks every local full gate and every `--lite` run on macOS.

## The state

**`origin/main` (`e28576421`) FAILs the `roborev-lints` gate component on macOS.** Reproduced in a pristine `git worktree add --detach origin/main`, so it is not a dirty-tree or worktree artefact:

```
==== ROBOREV REVIEW GUARD TEST TALLY ====
passed: 553  failed: 7
GUARD-TEST RESULT: FAIL      (exit 1)
```

`scripts/agent-gate.sh` runs `scripts/tests/test_roborev_review_guard.sh` inside `roborev-lints` and gates on its exit code, so **`RESULT: FAIL` is now the outcome of every local `--lite` and every full gate of record on this platform**, regardless of the branch under test. It bit a markdown-only, single-file docs diff.

## Why it appeared now — and the part worth keeping

At `9f11eb7cf^` (`4cf8c2678`, pre-#3262) the same script reported:

```
passed: 329  failed: 3
GUARD-TEST RESULT: FAIL      (exit 0)   <-- exit ZERO on a FAIL verdict
```

**Three of these failures are not new. They were latent behind a fail-open exit path in the guard test itself: the script printed `GUARD-TEST RESULT: FAIL` and returned `0`, so `roborev-lints` read green on a red test.** PR #3262 (#3229) closed that fail-open — correctly — and the three latent macOS failures became visible at the same moment four newly-added cases brought their own macOS incompatibility. The regression is the platform incompatibility; the exit-code fix is the thing that surfaced it and must not be reverted.

This is another instance of the class this program keeps finding (`docs/architecture/0.17-throughput-mission.md` §9 rule 13): **an unmeasurable or failed state inheriting the permissive branch.** Here the instrument that guards our review instrument was itself fail-open.

## Root causes — diagnosed, not guessed

**(1) GNU-only `sed -i` — 4 failures: `cx28`, `cx29`, `cx28b`, `cx28c`.** The test patches a copied checks file with `sed -i 's/.../.../' file`. BSD/macOS `sed` consumes the next argument as the backup suffix, so the script becomes the filename and the command errors:

```
[...]

Issue #3287, opened 2026-08-04
Split #3224's ~32% unattributed cycles/row residual: TMA level-2 + an offcore/prefetch-stall term (measurement-only)


pmcfadin:
## Context

#3224 attributed the full-box IPC decay of the Flight read path and closed **67.76%** of the
+5,639 cycles/row delta between S=1/N=2 and S=6/N=16 (report:
`docs/reports/ws0-3224-report.md` §5.3). This issue names the remaining **~32.24%** so it is not
re-derived from scratch.

For comparison, #3217 left ~87% unattributed. The residual here is a *stated number with a known
scope limit*, not an unknown.

## The residual, precisely

From #3224 §5.3, the three nested stall counters difference into disjoint buckets:

| bucket | cycles/row | % of Δ |
|---|--:|--:|
| L3-miss stalls (DRAM) | +3,821.3 | **67.76%** |
| L2-miss-but-L3-hit stalls (LLC) | +657.4 | 11.66% |
| other execution stalls | +1,792.6 | 31.79% |
| non-stalled cycles | −631.9 | −11.21% |

The residual is the `other execution stalls` bucket (**+1,792.6 cycles/row**) net of the −631.9
in non-stalled cycles. At the wider boundary it reads 20.6%. The shares carry a **±~5% basis
uncertainty** from the choice of measurement arm (67.8% on the group-C basis, 64.0% on the
primary-arm basis) — a real limit of the two-endpoint capture, not rounding.

## Why it is NOT simply "non-memory" — the load-bearing caveat

`cycle_activity.stalls_l3_miss` is **demand-load scoped**. Verified verbatim from the event
definition on the host (#3224 §5.2, `host/counter-semantics-verification.txt`):

> "Execution stalls while L3 cache miss **demand load** is outstanding" — `event=0xa3, umask=0x6, cmask=0x6`

Stalls caused by **prefetch** traffic (prefetches saturating fill buffers or memory queues) are
**not** in that term and land in `other execution stalls` instead. That matters specifically here
because the hardware prefetcher does most of the fetching on this streaming scan: demand
`LLC-load-misses` × 64 B covers only ~2.4 KB of the ~22 KB DRAM traffic per row (#3224 §3.4).

**So the ~32% is *unattributed*, not *non-memory*.** The `other execution stalls` bucket is an
**upper bound** on the non-memory share, not a measurement of it. Any work here that reports "X%
is front-end" without separating prefetch-induced memory stalls has not answered the question.

[...]

PR #3324, opened 2026-08-05
fix(#3272): harden the WS0 measurement rig — every guard OBSERVED to fire, not merely present

### LINKED ISSUE #3272: Harden the WS0 measurement rig before anything is funded on it: 7 open findings, 3 of them guards that are fail-open or bypassable (split from #3096)
Split out of #3096 / PR #3254 by owner decision (2026-08-03), so the reviewed core changes can land while the measurement rig is hardened properly. **The rig is not optional work — it is the durable asset #3096 was worth doing for.** It just should not hold the core, and it needs review coverage it could not get inside a 62-file PR.

## What moves here

From `issue-3096-arrow-encode`:
- **`tools/ws0-corpus-gen`** — deterministic 4M-row `ws0.events` corpus generator on the production `SSTableWriter`: uncompressed (#1406), Murmur3 token order, seeded, byte-identical across 3 generations, pinned by its own recorded sha256 `4a903f6fa27c04dbf87a44fddf78615aed73fcd379ecaee6669f6b0d9bbae269` / 4,000,000 rows / 40,000 partitions / 693.69 B/row. Documented a **PERFORMANCE FIXTURE ONLY**, never a correctness oracle (#3042).
- **`scripts/perf/`** — same-session both-arm measurement rig: CPU-wide `perf stat -C` (never `-p`), verified-sibling `taskset` pinning fail-closed, warm/cold as separate claims, setup-subtracted cycles/row, non-zero exit on zero rows.
- **`scripts/tests/test_ws0_report_guards.sh`** (16 checks, gate-wired into `tooling-tests`) and the `scripts/agent-gate.sh` hook that runs it.

The measurement **artifacts** (`docs/reports/ws0-3096-artifacts/`) stay in #3096 — they are the posted evidence for its negative result. This issue must land the rig that reproduces them.

## Why it is being split rather than shipped as-is: seven open findings, three of them self-inflicted

A sliced roborev pass (6 rounds, every one `prompt-content: PASS`) found these open at `c6b4f46`. **Three are guards added by earlier fix rounds being bypassable or fail-open** — the "a fix moved the problem" pattern, which is the strongest argument for hardening the rig on its own footing with its own review coverage:

1. **BLOCKER — `ws0_report.py:403`**: absent `corpus-identity.json` ⇒ `corpus_rows=None` ⇒ the full-corpus-per-request validation is **silently skipped**, while the generated notes still claim it was verified. Fail-open inside the guard added for the cold-blend defect.
2. **BLOCKER — `ws0_report.py:98`**: `skipped-cold-arm` counts as a successful prewarm regardless of temperature, so a **WARM** rep carrying it gets `prewarm_all_ok=true` — an unprewarmed warm measurement passes the guard that exists to prevent exactly that. Direct bypass of the prewarm fix.
3. **BLOCKER — `ws0-baseline.sh:324`**: sets `kernel.perf_event_paranoid=-1` and `kernel.kptr_restrict=0` and **never restores them** (`trap stop_server EXIT` only stops the server), leaving host hardening weakened after every run. Security-adjacent.
4. **BLOCKER — `ws0_report.py:161/178`**: missing/unparseable setup counters become `0` via `.get("cycles", 0)`, so a run can be reported "setup-subtracted" when **no subtraction occurred**. Fabricated-zero, against the authoritative-metadata-only mandate (a counter not observed is an error, never a fabricated 0).
5. **BLOCKER — `--reps` never validated positive**: `--reps 0` produces a vacuous but *successful* report. Vacuous-green class.
6. **BLOCKER (rig-only)** — `require_complete` documents that an unrun arm is allowed, but the collectors append every absent expected artifact, so an intentionally unselected arm/temperature is indistinguishable from a failed run and exits fatally.
7. **NIT — `ws0-baseline.sh:190`**: duration components enter Bash arithmetic without `10#`, so `010s` parses as 8s and `08s` errors. Silent wrong measurement window.

## Carried durability items (previously parked on #3248)

8. **The determinism claim has no self-test.** "Byte-identical across three generations" is load-bearing for every future comparison against this corpus but exists only as prose; the mechanism (`--verify-against`, fail-closed) and row-content unit tests are committed, but nothing **regenerates and byte-compares**.
9. **The measurement-corpus digest is not pinned in-tree.** The CI-fixture digest `0xd0014e42e893f87f` is a constant asserted at HEAD; the measurement-corpus digest `0x0390bfbb81a23fa1` (4,000,000 rows / 31,250 batches) is only recorded in an artifact doc, so its invariance rests on operator re-runs.
10. **`scripts/perf/`'s fail-closed paths need `scripts/tests/` coverage** beyond the 16 checks already added — specifically the verified-sibling `taskset` check (both directions) and the `perf stat -p` self-grep.

## The governing lesson

Every finding above is the same shape: **an instrument that reports success without having measured.** #3096 spent more effort on instrument integrity than on the Arrow-encode question it was opened for, and produced four sibling issues about it (#3252, #3256, #3259, plus #3257/#3258 the owner filed). A rig whose guards are fail-open is worse than no rig, because it produces confident numbers nobody re-checks.

So the acceptance bar here is **not** "the guards exist" — it is "each guard has been observed to fire." Per #3249's mutation-testing result (hardcoding `_PERF_STATE="ok"` survived 118/118 tests), a guard written but never observed failing is not evidence.

## Acceptance criteria

1. All seven findings above fixed, each with a test that **observes the guard firing**, and — where the finding is a bypass of an earlier fix — a non-vacuity assertion showing the pre-fix code would have accepted the input the new check rejects.
2. Items 8–10 landed: a regenerate-and-byte-compare determinism test, the measurement-corpus digest pinned in-tree (or an explicit documented reason it cannot be, with the operator procedure), and `scripts/tests/` coverage for the remaining fail-closed paths.
3. **No fabricated values anywhere in the reporting path.** A counter that was not observed is an error, never a `0` default.
4. The rig runs end-to-end from a clean checkout with no path outside the repo, and a zero-rows or partial-corpus run exits non-zero.
5. Host state that the rig mutates (`perf_event_paranoid`, `kptr_restrict`) is restored on every exit path, including failure and signal.
6. Re-verify the #3096 figures reproduce with the hardened rig — and if they do **not**, that is the finding: report it rather than adjusting the rig to agree with the recorded numbers.

## Routing

**Oracle-driven** — every criterion is a deterministic assertion about whether a guard fires. No OpenSpec change needed. (Note: #3096's OpenSpec requirement R4, "the corpus generator and measurement scripts are COMMITTED and runnable from a clean checkout", is re-anchored **here** by the split.)

## References

Split from #3096 / PR #3254. Successor context: #3248 (profile-first attribution, which will consume this rig — do not fund a lever on an unhardened instrument). Review-integrity family found by the same delivery: #3252, #3256, #3259, #3257, #3258. Mutation-testing precedent: #3249.

pmcfadin:
Closes #3272.

Restores the WS0 measurement rig (split out of #3096 / PR #3254 by owner decision) and hardens it. The governing bar for this issue is **not "the guard exists" but "the guard has been OBSERVED to fire"** (per #3249, where hardcoding `_PERF_STATE="ok"` survived 118/118 tests), so every fix below carries a test that feeds the rejectable input plus, where the finding was a bypass of an earlier fix, evidence that the pre-fix code accepted what the new check rejects.

## Acceptance criteria

| AC | State |
|---|---|
| **AC1** — all seven filed findings fixed, each with a guard-fires test + non-vacuity assertion | **met** |
| **AC2** — items 8–10 (regenerate-and-byte-compare determinism test; measurement-corpus digest pinned in-tree; `scripts/perf/` fail-closed coverage) | **met** |
| **AC3** — no fabricated values anywhere in the reporting path | **met**, mechanized (AST scan over every reporting-path module) |
| **AC4** — runs end-to-end from a clean checkout; zero-rows or partial-corpus exits non-zero | **met** |
| **AC5** — host state (`perf_event_paranoid`, `kptr_restrict`) restored on every exit path incl. failure and signal | **met**, verified behaviourally via a recording `sudo` shim + a real SIGINT probe |
| **AC6** — re-verify the #3096 figures reproduce with the hardened rig | **re-homed to #3248** by coordination-lead ruling ([REQ-3272-20260804T200053Z](https://github.com/pmcfadin/cqlite/issues/3272#issuecomment-5182491795)) — it needs a Linux perf host and the 4M-row corpus, and #3248 consumes this rig, so the reproduction is its first deliverable |

AC6 is **re-homed, not waived**: the criterion survives with a real box, the real corpus, and its own review, and #3248 is explicitly authorised to report a divergence as its finding rather than adjust the rig to agree.

## What the reviews found

Nine review rounds, two independent reviewers each (`rust-reviewer` + roborev/codex), plus a `/code-review` pass. **~55 findings**, all real. The dominant theme, and the reason the diff is large: **fix rounds kept reintroducing the defect they were fixing, one level down.** One class — "a hand-maintained list or count that must be updated in lockstep, with no oracle forcing it" — recurred **five times**:

1. a spelling-based predicate (the lint asked for a literal `bash` token on the line);
2. a hardcoded five-token gate that made the documented fail-closed posture dead code for every other variable name;
3. a `WRAPPERS` keyword list enumerating `then`/`else`/`elif`/`do` while **omitting `if`/`while`/`until`** — the omission of `if` beside the presence of `elif` being the tell of a hand-written list;
4. `assert_eq!(mutations.len(), 15)` passing after a 16th field was added — the very mechanism whose stated job was to force a new field to acquire a divergence case;
5. `MANIFEST_CONFIG_FIELDS` with no oracle in either direction, which was the **root** of the false pinning claim below.

Each is now closed by a mechanism rather than a per-site edit, and the difference is measurable: a glob-discovered lint found **13** bare driver call sites where a manual sweep had found **1**; an AST walk enumerated 11 coercions and 17 derived quantities; inverting the lint to a **closed grammar over bash reserved words** (closed *because bash enumerates it* — asserted against `compgen -k` bidirectionally) surfaced **8** evasion shapes where the review had reported 3, from a second cause nobody had identified (`_bare(";") == _bare("&&") == _bare("|") == ""`, so standalone operators never reset command position).

`WRAPPERS` remains an **admitted, documented residual** — "programs that run another program" is open-ended and not closable without modelling `PATH`.

### The single most important finding

`ws0_report.py` printed **"verified physical-core siblings"** about CPU pins nothing verified. Found by *execution*: setting `config.server_cpus="99,99"` in a fixture manifest made the report **exit 0 printing that claim verbatim** — byte-identical to the line this branch's own test comment records as the defect an earlier round had "closed". That fix closed the argv route; the identical false claim survived one layer in, via the artifact. Compounding it, the same `results.json` took the opposite posture two fields away (`recorded_round_metadata.source`: *"provenance UNVERIFIED — the reporter reads a directory and cannot establish which program wrote it"*) — one document, two epistemics.

Fixed by making the claim rest on a **recorded observation**: the driver captures `verify_sibling_pair`'s sysfs answer into `pinning-verification.json`, and the reporter requires it and asserts the manifest's lists equal the verified ones. Re-verifying at report time was rejected deliberately — a results dir reviewed on another host would read a topology that never ran the measurement.

### The interleaving claim was DELETED, by owner ruling

The rig printed *"the reps were INTERLEAVED … this is OBSERVED FROM THE CLOCK"* while, at the rig's own default `--reps 1`, `zip(ordered, ordered[1:])` was **empty** — zero orderings compared — and `round_major_verified` returned `True`. A positive verdict from the absence of a bad signal, inside the fix written to remove exactly that.

Per [REQ-3272-20260805T040012Z](https://github.com/pmcfadin/cqlite/issues/3272#issuecomment-5188282904) the claim could stay only if made a genuine observation; it was not, so the deletion branch fired. Removed: `verify_interleaving`, `verify_round_major_timing`, `interleaving_lines`, every printed claim sentence, and **13 verdict fields**. Kept as **inert** `recorded_round_metadata` with `claims_made: "NONE"`: the raw per-rep round/position/instants, for #3287/#3299 to analyse on real hardware. `docs/reports/ws0-3096-artifacts/measurement-method.md` §3b is amended with a new **§3b.1 "STATUS: this control is NOT IMPLEMENTED OR ENFORCED by the committed rig"** — a method doc calling a nonexistent control "binding" is worse than no control.

### Other notable fixes

- **The documented operator verification command could never succeed.** A required `schema_sha256` field met a committed artifact recorded before that field existed, so `--verify-against` died with `missing field` before generation. Fixed via a **source oracle**: `sha256(schema::DDL + "\n")` is invariant across every commit that ever touched `schema.rs` (independently verified: `6bdd1d06ad7eb597…` at both `683e717f1` and `a8dbcfa2e`), so the schema is verified against *source* rather than a recorded value — strictly stronger than a complete artifact. **Proven end-to-end on the real 4M-row corpus**: `determinism: PASS — reproduced … exactly` / `EXIT=0`, `Data.db sha256: 4a903f6f…` matching the pin.
- **"Hermetic" tests were not hermetic on Linux.** Accept-direction cases ran the real driver, so on Linux they would invoke `sudo -n sysctl -w` and three `cargo build --release`s, then a full measurement run (3 reps × 2 arms, 45s Flight steps, real `perf stat`) **inside a gate component**. Invisible on macOS because execution stopped at `perf is not installed`. Closed with a `--validate-args-only` boundary + recording PATH shims, enforced by a standing lint whose subject is glob-discovered and whose completeness is anchored to an independent `git ls-files` census.
- Host sysctls now restored per-knob on every exit path, and a knob whose prior could not be captured is never mutated.
- Fabricated zeros removed throughout the reporting path; `exact_int` refuses bools, fractional and non-canonical integers (`requests_error: 0.9` previously read as clean).
- Derived-not-trusted values: Flight throughput from `rows_total/duration_s`, bare-scan aggregates from per-pass records. A derived value cannot be forged.
- Corpus verification covers the **complete component set** (not just `Data.db`), pinned before rep 1 and re-checked at report time.
- CPU-range endpoints validated as **decimal strings** before any arithmetic (`'1-x[$(touch …)]'` previously executed and exited 0; a 64-bit wraparound previously bypassed both bounds).
- Output dirs claimed atomically, so concurrent runs cannot mix two sessions' artifacts into one report.
- `--identity-out` can no longer alias a generated input and overwrite it after its digest was recorded (case-insensitive APFS bypass demonstrated live).
- A stray repo-root file `0` was deleted twice and returned twice; the actual cause was a **committed test label with unescaped backticks** — `` pass "…(a `> 0` test…)" `` ran as command substitution, creating the file **and blanking the label**. Both escaped, plus a `label-backticks` guard and a repo-root junk-file check (which caught a live recurrence during its own verification).

## Gate

Gate of record runs on **Linux** per [ruling](https://github.com/pmcfadin/cqlite/issues/3272#issuecomment-5188282904): round 3 established that the macOS gate is **structurally blind** to this issue's worst defect class, so a macOS-only green is not certification for this diff. Summary block to follow in a comment.

Known pre-existing red on macOS, cited not patched: **#3296** (`roborev-lints`, 553 passed / 7 failed — GNU-only `sed -i`, `/private/var` canonicalisation), reproduced byte-identically on a pristine `origin/main` worktree.

Eight hermetic suites, all gate-wired into `tooling-tests`: hermeticity 84, round-metadata 44, fabrication 92, host-state 25, report-guards 103, cpu-pinning 174, provenance 90, numeric-wraparound 28, plus the root-junk self-test. The cpu-pinning count rose 172→174 because a check that could no longer evaluate was repaired: an earlier fix in this same series turned a driver statement into a command substitution, silently stranding the grep that anchored the ordering assertion — a vacuous guard of exactly the class this issue exists to remove, found by a reviewer and fixed here.

## Follow-ups (not fixed here — deliberate scope discipline)

1. **P1** — the generator emits no null cells, and a pre-existing assert refuses a corpus whose taps observe zero nulls, so `ws0-corpus-gen` corpora **cannot reach** the Flight digest oracle. The new pinned-digest comparison is therefore latent until the generator gains a null plan or that assert is scoped to the CI fixture.
2. `scripts/perf/ws0-baseline.sh` is 915 lines against the ~800 target (down from 1008); the header names the next seam (the two session-pin heredocs). The gate's `file-size` ratchet is `.rs`-only, so shell/python cross it silently.
3. `test_ws0_provenance_guards.sh` at 1314 lines, and its `MIN_CHECKS=64` floor now far below its actual 90 — a floor that lags stops detecting a silently-skipped block.
4. The hermeticity lint's quote-parity claim is false for a *later-executed* string: `trap 'bash "$DRIVER" …' EXIT` and `eval` are skipped as string content. No such line exists in the tree, and it predates this series, but `eval`/`trap`/`-c` deserve naming as a residual beside `WRAPPERS`.
5. `PartialUnverified` and `Diverged` share `ExitCode::FAILURE`, so a caller cannot distinguish "predates the pin" from "did not reproduce" by exit code alone. Neither reads as a pass, so this is a convenience gap.

---

## Certification record (#3272 terminal round — lead ruling B, ship `d50f4ae06`)

**Gate of record: PASS** — full `agent-gate.sh` on Linux worker `i-0f92a75ac756d35d0`, SHA-verified.
```
commit: d50f4ae  dirty: no
tree-start: d50f4ae06e01 == tree-end: d50f4ae06e01   tree-integrity: PASS
roborev-lints: PASS   31 components   0 FAIL
RESULT: PASS
```

**Final roborev: GENUINE review** at `d50f4ae06` (range `a0462ec7..d50f4ae06`, census 90 files, tokens input=3.17M/cached=2.96M — far above the vacuous-baseline floor). Returned **3 Medium findings, all in-scope validation-of-validation gaps**, dispositioned (not waived) to **#3354** per lead ruling:
1. `ws0_boundary_observations.py:114` — completeness record does not validate `declared_inputs_*`
2. `ws0_collect.py:434` — setup counters consumed without validating `<tag>-setup.json`
3. `ws0_binaries.py:555` — binary-provenance revision/dirty-flag cross-checks

**`prompt-content: FAIL (87/87)` — WAIVED, SHA-bound to `d50f4ae06`** (lead ruling §4): the #3312 large-diff false-FAIL (roborev delivers big diffs by snapshot path). Waiver basis: the review cited three findings at specific `file:line` positions **inside the diff's own new files** — affirmative proof of delivery stronger than a captured snapshot — plus 3.17M-token accounting and a verified range. Voids on any push.

**F1 non-vacuity confirmation (round-22 checklist item): 13/13.** The pre-fix reporter was *measured* publishing `ratio bare/flight = 2.00x` with no `boundary_observation_completeness` field while claiming the digest and every component verified — the completeness guard demonstrated against its own counterexample.

Lead ruling: https://github.com/pmcfadin/cqlite/issues/3272#issuecomment-5222979690 (escalation) → ruling comment 5222… (B).
Residual validation-depth work: **#3354** (P2). `premerge-assert 3324 d50f4ae06...`: **OK**.


pmcfadin:
Gate of record: running on a **Linux** agent-ami worker (`i-0da100e5b7b22066d`) at the certified SHA `6be1d8af6226dfaaec7b83769f5ce3bd3abd2319`, tree clean, per the coordination-lead requirement that this diff not be certified on macOS ([ruling](https://github.com/pmcfadin/cqlite/issues/3272#issuecomment-5188282904) — round 3 established the macOS gate is structurally blind to this issues worst defect class, since the bare-driver escape hides behind `perf is not installed`). Summary block to follow.

Note for anyone re-running this: the workers `origin` fetch refspec is pinned to `+refs/heads/main:refs/remotes/origin/main`, so a plain `git fetch --prune origin` does **not** bring down an issue branch — it fails with `not a commit and a branch cannot be created from it`, which reads like a missing push. Fetch the branch explicitly (`git fetch origin <branch>:refs/remotes/origin/<branch>`).

pmcfadin:
## Gate of record — PASS (31/31) on Linux

Ran on a Linux agent-ami worker (`i-0da100e5b7b22066d`) at the certified SHA `146dfe514fb56874d94d8823dfabafe0579992bf`, per the [coordination-lead requirement](https://github.com/pmcfadin/cqlite/issues/3272#issuecomment-5188282904) that this diff not be certified on macOS — round 3 established the macOS gate is structurally blind to this issue's worst defect class (the bare-driver escape hides behind `perf is not installed`).

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.GQbKSu
commit: 146dfe5 branch: issue-3272-harden-ws0-rig dirty: no
schemas: 6/6 canonical .cql readable under /home/ubuntu/workspace/repo/test-data/schemas (checkout-relative)
tree-start: 146dfe514fb5 dirty: no digest: f087c0569e82
tree-end: 146dfe514fb5 dirty: no digest: f087c0569e82
tree-integrity: PASS
file-size: PASS  fmt: PASS  clippy: PASS  roborev-lints: PASS
core-tests: PASS  tombstones-scan: PASS  scan-offload-guard: PASS
work-counters-guard: PASS  byte-budget-guard: PASS  arrow-parity-guard: PASS
memory-budget: PASS  integration-tests: PASS  format-compat: PASS
write-tests: PASS  cli-tests: PASS  compaction-byte-parity: PASS
bti-multiclustering: PASS  query-semantics-oracle: PASS
flight-query-semantics-oracle: PASS  python-bindings: PASS  node-bindings: PASS
delivery-telemetry: PASS  oom-audit: PASS  parity-report: PASS
operator-metrics-doc: PASS  kit-dashboard-drift: PASS
binding-unwind-profile: PASS  tooling-tests: PASS (519s)
minimal-build: PASS  smoke: PASS
RESULT: PASS
```

**`roborev-lints` PASSES here**, which confirms #3296 is macOS-only (GNU vs BSD `sed -i`, `/private/var` canonicalisation) — so this PR needs no waiver for it. The local macOS lite gate still reds on that component; it is reproduced byte-identically on a pristine `origin/main` worktree.

### The first Linux run FAILed, and the cause was my own guard

Worth recording, because it is this issue's own defect class turning up inside the fix for it. The first full run scored 30/31 with `tooling-tests` FAIL. The cause was **not** the diff under test: `scripts/ci/check-root-junk-files.sh --self-test` printed

```
root-junk self-test: 34 checks, 0 failure(s)
```

and **exited 1**. `rc` was declared once in `self_test()` and then reused to capture the *deliberately* non-zero exits of two probe cases (the round-8 no-swallow case and the round-7 F4 shim), and that stale `rc=1` survived to the terminal `return "$rc"` — the success path never reset it. A verdict derived from a stale side-effect instead of from an affirmative measurement, which is exactly what this issue exists to eliminate.

Fixed in `146dfe514` by giving the probe cases their own `case_rc` and keying the verdict unconditionally on `fails`. Both directions observed: a passing self-test returns **0**; a single forced failure (floor `min=33`→`99`) still returns **1**, so the fix is not "always return 0".

Two other things that first run surfaced, both now closed:
- The guard's original subject was the **live working tree**, so a concurrent gate step that created a transient root file named `720` (untracked, never committed, already gone when inspected) could red the gate over content no author can act on. Narrowed to the **tracked** root — which is the case that actually shipped twice — with the untracked half demoted to an explicitly non-verdict notice. Not given a settle/retry window: that would trade a wrong subject for a timing dependency.
- A worker's `origin` fetch refspec is pinned to `+refs/heads/main:refs/remotes/origin/main`, so a plain `git fetch --prune origin` does not bring down an issue branch and fails with `not a commit and a branch cannot be created from it` — which reads like a missing push. Fetch explicitly: `git fetch origin <branch>:refs/remotes/origin/<branch>`.

pmcfadin:
## Final roborev confirmation pass (flow-closer): **FAIL — 4 genuine findings, merge NOT armed**

This round is **not** the pre-authorized `prompt-content`-only waiver case. `findings: PRESENT (4)`,
`roborev-exit: FINDINGS (exit 1)`. The waiver (REQ-3272-20260804T225330Z) was conditional on
`prompt-content:` being the ONLY failing key; it is not, so it does not apply and no waiver is recorded.

```
repo: /Users/pmcfadin/projects/cqlite-wt/issue-3272
branch: issue-3272-harden-ws0-rig
base: origin/main
head-sha: 146dfe514fb56874d94d8823dfabafe0579992bf
reviewed-sha: a0462ec7ccf5a73d434386e0440068ae0c6d1fe6..146dfe514fb56874d94d8823dfabafe0579992bf
job: 7341
model: gpt-5.6-sol
census: 55 files, +24221/-10
tokens: input=3477662 cached=3273472 output=9648
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: FAIL (52/52 code census paths absent from the prompt)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: PRESENT (4)
RESULT: FAIL
```

### Triage: 4 x Medium, all in-diff, all graded BLOCKER

Per `docs/development/roborev-severity.md` ("when in doubt, blocker"), and because each is an
instance of *this issue's own* governing failure shape — **an instrument that reports success
without having measured** — none is a nit. All four land in files this PR adds
(`+182`, `+402`, `+535`, `+569` lines respectively), so none is out-of-diff.

**F-A (BLOCKER) `scripts/perf/lib-measure.sh:158` — prewarm `ok` from exit status alone.**
Verified by direct read: `prewarm_status="ok"` is set solely by the `if` on `flight-loadgen`'s exit,
and the invocation passes `--out /dev/null`, discarding the per-request JSONL that would allow
verification. A loadgen run recording only unavailable/error requests exits 0 and is recorded as a
healthy prewarm. This is the **same class as AC1's finding 2** (`skipped-cold-arm` counting as a
successful prewarm) reappearing at a different line — the "a fix moved the problem" pattern the
split was opened for. Remedy per reviewer: retain the prewarm JSONL; require >=1 successful
full-corpus request, zero unavailable/error, and the expected row count before recording `ok`.

**F-B (BLOCKER) `tools/ws0-corpus-gen/src/bin/scan_bench.rs:160-165` — arms may measure different
SSTable sets.** Uses `table_directory_filter: Some(format!("/{}/{}", keyspace, table))`. CQLite's own
`cqlite-core/src/ingestion.rs:55` documents that field as a "SUBSTRING match. Loose by design" and
directs callers to `TableDirSelection::Exact` via `ingest_with_selection`, "which compares complete
path components (issue #3220)". So a sibling `ws0/events-backup` is also ingested, and the bare-scan
arm can measure a different corpus than the Flight arm — which silently invalidates the *comparison*
that is the rig's entire output. The remedy already exists in-tree and is used by
`cqlite-core/examples/bti_perf_scan/main.rs:462`.

**F-C (BLOCKER) `tools/ws0-corpus-gen/src/identity.rs:400` — `--verify-against` can report
`Reproduced` after a DDL-only change.** When the prior identity lacks `schema_sha256`, the
regenerated schema is compared against a digest derived from the *same current* `DDL` that produced
it — a self-comparison, which cannot detect historical schema drift. Directly weakens AC2's
determinism claim (item 8: the regenerate-and-byte-compare test). Remedy: compare against an
independently pinned historical `SCHEMA_SHA256`, add the digest to the committed prior identity, or
keep the nonzero partial/unverified verdict rather than `Reproduced`.

**F-D (BLOCKER) `tools/ws0-corpus-gen/src/main.rs:266` — containment guard bypassable on
case-insensitive FS.** `--out /tmp/foo --identity-out /tmp/FOO/ws0-events.cql`: both resolve
lexically distinct *before* generation, but APFS folds them afterward, so the identity overwrites
the generated DDL and the run exits 0. The in-code comment anticipates the case-fold problem and
argues containment "subsumes checks 1 and 2 for every spelling" — but the containment test itself
runs pre-generation on a not-yet-existing root, so the fold is invisible to it too. Remedy: create
and canonicalize the corpus root, then repeat filesystem-resolved containment before generating.

### `prompt-content: FAIL` — evidence for #3312, and a note against over-trusting it

`prompt-content:` reported 52/52 paths absent, yet this round produced four specific `file:line`
findings that I verified against the shipped source. A reviewer that received nothing cannot do
that. Combined with the token accounting (input 3.48M, cached 3.27M = ~94%, output 9.6k — against
the documented vacuous baseline of 18.7k/0/~55) this is a **ninth** consecutive round corroborating
#3312's snapshot-path delivery defect rather than a vacuous review. Recording it as evidence only;
it changes nothing about the verdict, because the round failed on findings regardless.

### Not done, deliberately

No `premerge-assert`, no `gh pr merge --auto`, no claim release, no worktree removal, no telemetry
stamp, no issue close. The certified gate at `146dfe51` stays valid **only until the first src fix** —
fixing any of F-A..F-D invalidates it and requires a fresh full gate of record (`--lite` will not do).

C intent audit: **PASS** (every requirement satisfied w/ public-surface evidence; AC6 correctly
re-homed to #3248). Deferred evidentiary item unchanged: #3326 item 1 should land before #3248 quotes
the Arrow digest.

Issue #3323, opened 2026-08-05
Privileged drop-in install: ancestor-chain rename race (12th TOCTOU-family escape) — known-High residual split from #3261; fix is a dirfd/openat2 helper


pmcfadin:
Split from #3261 by owner-delegated D-disposition (2026-08-05). The finding, verbatim from roborev round 4 (confirmed range `28d707d89..19ca940bf`):

> Sandbox containment is checked before entering the privileged shell, but the shell later reopens the original path spelling. A less-privileged user who can replace an ANCESTOR entry can redirect the path after validation to a root-owned directory such as /etc/sysctl.d; that directory then passes the ownership/mode check and receives the privileged write.

**Status: explicitly UNFIXED on main, by ruling — a recorded boundary, not ignored debt.**

**Threat model (why this is P2-latent, not P1):** exploiting it requires ALL of: (a) attacker with local unprivileged execution on an agent box, (b) a writable ancestor of the drop-in path (production ancestors /etc and / are root-owned), (c) a real privileged tee (test mode uses AC4 shims). Single-tenant agent boxes supply none of these.

**The known correct fix is recorded here: a non-shell helper holding a directory descriptor (openat2-style containment), creation/rename relative to the held dirfd.** Shell cannot hold a dirfd across sudo — twelve in-shell attempts (#3261's family table) each moved the window without closing it; a thirteenth is forbidden by standing ruling.

**Re-raise trigger (the ONLY activation path):** the threat model changes — multi-tenant boxes, an attacker-writable ancestor introduced into the drop-in path, or a real privileged helper reachable from test mode. Evidence of any of these activates this issue; nothing else does. The cost side of the fix (an interpreter dependency on the privileged bootstrap path is itself an attack surface) must be re-weighed at activation, not assumed.

Oracle-driven (the race is demonstrable); no OpenSpec. Related: #3261 (parent), #3229/#3296 (doctrine).

---

## ENTRY 2 (appended per owner D′ ruling, 2026-08-06) — the THIRTEENTH escape: privileged-executable revalidation

Same family, same `validate-a-NAME-then-re-resolve` shape, different surface: **executables** rather than
paths. Recorded here rather than as a separate issue — one boundary, not two half-boundaries.

**Location:** `scripts/perf-capability.sh`, `perf_capability_env_guard` (the AC4 privileged-tool sweep).

**Defect (reviewer's words, verbatim):**
[...]

Issue #3257, opened 2026-08-03
roborev silently drops the diff from the reviewer's prompt above ~3k changed lines and reports PASS — a large PR gets a vacuous clean review


pmcfadin:
## Summary

On a large branch, `roborev review --branch` **silently sends a prompt containing NO diff at all**, the reviewer returns `findings: NONE`, and `roborev-exit` is **PASS**. Textually that is indistinguishable from a genuine clean review. Only `scripts/flow/roborev-review.sh`'s `prompt-content` assertion (#2964) caught it.

This is a **fifth** vacuity path, beyond the four T1-T4 already documented in CLAUDE.md, and it is the most dangerous of them because it is **triggered by success**: the bigger and more thoroughly-reviewed a branch gets, the more likely its next review round is vacuous.

## Evidence — measured on issue #3249 / PR #3251

Five review rounds on one branch, same invocation, same repo, same base, only the diff growing:

| Round | Diff size | `prompt-content` | Findings |
|---|---|---|---|
| 1 | `+1003/-13` (7 files) | **PASS** (5/5 code paths) | 2 |
| 2 | `+1866/-19` (11 files) | **PASS** (5/5) | 4 |
| 3 | `+2334/-19` (11 files) | **PASS** (5/5) | 2 |
| 4 | `+2835/-19` (11 files) | **PASS** (5/5) | 4 |
| 5 | `+3811/-19` (13 files) | **FAIL (7/7 code paths ABSENT)** | **NONE** |

Round 5's job record (job 27):

```
tokens: input=489155 cached=451328 output=3875
prompt-content: FAIL (7/7 code census paths absent from the prompt)
findings: NONE
roborev-exit: PASS          <-- would have read as "roborev clean"
```

The prompt actually sent, retrieved via `roborev show 27 --prompt`, is **7298 bytes** and contains **zero `^diff --git`** headers. It carries the reviewer preamble and no changes. Of three spot-checked census files, only `scripts/perf-capability.sh` appears at all (once, in prose); `scripts/agent-gate.sh` and `scripts/bootstrap-agent-machine.sh` appear **zero** times.

Note `input=489155` with `cached=451328`: the model was not idle — with no diff to review it appears to have explored via tools, which burns real tokens and so **defeats the `vacuity-tier2` token-accounting heuristic**. Both vacuity tiers PASSed. Token accounting cannot detect this path; only `prompt-content` can.

## Why this is worse than T1-T4

[...]

PR #3328, opened 2026-08-06
fix(#3261): harden test-mode seam containment — positive containment for write targets, the fork-free read path, contained files and privileged executables

### LINKED ISSUE #3261: Harden test-mode seam containment for the perf-capability helper (positive sandbox containment, write-target + symlink closure)
Split out of #3249 (perf sysctl capability, PR #3251) at the owner's direction. #3249's five ACs ship in
PR #3251; the **test-mode seam containment** work moves here so it can be reviewed in isolation, at a size
roborev can actually read.

## Why this is its own issue

One guard — "may a test-mode env seam steer this path?" — has now produced **seven successive escapes**
across six review rounds, each one a new *spelling* of "somewhere else" rather than a new mechanism:

1. **B3** — a raw seam value steers a privileged `sudo tee`, so an env var chose a write destination.
2. **R4-3** — the seam checks were *textual non-production* checks (reject `/etc`, reject the production
   default), which meant "not obviously production" was accepted as "safe".
3. **the fail-open audit's symlink bypass** — `ln -s /etc/sysctl.d /tmp/x` passes every textual test.
4. **R5-1** — the spelling is not the destination: `/tmp/../etc/sysctl.d` and a symlinked *ancestor* both
   resolve into production, so the write side had to canonicalize (`cd -P` / `pwd -P`).
5. **R6-1** — `//etc`: `pwd -P` can preserve **two leading slashes** (POSIX permits it), and on Linux
   `//etc` **is** `/etc`, so a canonicalized path could still name production while textually differing
   from every denied spelling.
6. **R7-1** and **R7-2** (below) — a symlinked *write target* and a regressed symlink rejection.

The lesson is structural, not incremental: **a denylist over path spellings cannot be completed.** The
number of ways to spell "/etc" is unbounded; the number of ways to prove "strictly inside my sandbox" is
one. Hence the inversion, and hence a dedicated issue: the change replaces the guard's whole shape, and
that is exactly the kind of diff that must be read on its own.

## What already exists (moving in from the reverted commits)

Round 6 of #3249 built the inversion; it was reverted out of PR #3251 by
`revert(#3249): move the test-mode path-validator containment inversion to its own PR` and lives on branch
`issue-3261-perf-seam-containment` (commits `3a9c22a`, `cc2f2c8`, `5da07c8`, `f1b3fe2`). Already done:

- **One positive predicate.** `perf_capability_path_within <path> <root>`: rc 0 iff `<path>` is absolute,
  canonically spelled, and **strictly inside** `<root>` with a `/` boundary (so `/rootfoo` is not inside
  `/root`). An **empty root refuses** — the predicate can never degrade into "anything is fine".
- **A sandbox root that must PROVE itself.** `CQLITE_PERF_TEST_SANDBOX` must name an absolute *existing*
  directory carrying the stamp file `.cqlite-perf-sandbox`, so `CQLITE_PERF_TEST_SANDBOX=/etc` **cannot**
  turn `/etc` into a sandbox: the proof lives on the filesystem, not in the environment.
- **Three entry points over that one predicate.** `perf_capability_sandbox_ok` (fork-free syntactic, for
  the gate's contractually fork-free token path), `perf_capability_sandbox_ok_resolved` (fully resolving,
  for every write-side decision), and `perf_capability_sandbox_file_ok_resolved` (the file variant).
- **A structural closed-set audit.** `scripts/tests/test_perf_capability.sh` enumerates, **from the
  source**, every function that dereferences a seam and fails naming any that does not call the
  containment family — with a **floor** on the number of consumers found, so it cannot pass vacuously if
  the parse breaks, and **verified to fail** when a consumer is deliberately unhooked.

### AC1 (High) — directory containment is not write-target containment

`perf_capability_dropin_path` validates only the **containing directory**. If `99-cqlite-perf.conf` is
itself a **symlink**, the privileged `tee` **follows it** and overwrites the link's target — which may be
anywhere, including a file the sandbox has no business writing. A contained directory says nothing about
where its entries point.

- Validate the **write target**, not just its parent: `O_NOFOLLOW`-equivalent refusal, or atomic
  directory-entry replacement (write a fresh temp entry in the validated directory, then `rename` it over
  the name) so a pre-existing symlink is *replaced*, never *followed*.
- Test: the managed filename symlinked to a path outside the sandbox; assert the write refuses (or replaces
  the entry) and the outside target is untouched.

### AC2 (Medium) — the inversion REGRESSED a protection: symlinks are no longer rejected on the read path

The fork-free proc check is now purely textual and no longer rejects symlinks. A symlink **inside** the
sandbox pointing at `/proc/sys/kernel` therefore satisfies containment, and the run reports a fabricated
`perf=ok` derived from the host's real controls while claiming to have read a stand-in. This is a
regression against the pre-inversion behaviour (which rejected a symlinked seam) and must not ship.

- Either reject symlinks **per path component** using builtins only (the token path's fork-free contract
  stands), or pass a **pre-canonicalized, already-verified** path into the fork-free check so the check has
  nothing left to resolve.
- Test: a symlink **inside** the declared sandbox pointing at the real `/proc/sys/kernel`; assert refusal,
  not `ok`.

### AC3 (Low) — a strictly-contained file is wrongly refused

`perf_capability_sandbox_file_ok_resolved` rejects `$CQLITE_PERF_TEST_SANDBOX/sysctl.conf` even though the
file is strictly contained. Canonicalize the **parent**, construct the full file path from the canonical
parent plus the basename, and apply `perf_capability_path_within` to **that**.

- Test: a `sysctl.conf` directly inside the declared sandbox is **accepted**.

## Production safety — this hardens TEST-MODE isolation only

Verified at `a2c4af1` (the head PR #3251 ships), with `CQLITE_PERF_TEST_MODE` **unset** and every seam
(`CQLITE_PERF_PROC_DIR`, `CQLITE_PERF_SYSCTL_DIR`, `CQLITE_PERF_SYSCTL_EXTRA_DIRS`,
`CQLITE_PERF_TEST_SANDBOX`, `CQLITE_PERF_TEST_PRIV_DIR`, a bogus `sysctl.conf` override) exported to a
bogus tree whose contents would produce a **different, observable** answer (stand-in `/proc` says
`paranoid=4`/`kptr=1` → token `paranoid-4`, where the real host says `-1`/`0` → token `ok`):

| probe | result |
|---|---|
| `perf_capability_proc_dir` | `/proc/sys/kernel` (hardcoded literal) |
| `perf_capability_sysctl_dir` | `/etc/sysctl.d` (hardcoded literal) |
| `perf_capability_dropin_path` | `/etc/sysctl.d/99-cqlite-perf.conf` |
| `perf_capability_sysctl_search_path` | the six hardcoded entries; `CQLITE_PERF_SYSCTL_EXTRA_DIRS` ignored |
| `perf_capability_token` | `ok` — the REAL `/proc`, not the stand-in's `paranoid-4` |
| gate `accelerators:` perf token | ` perf=ok` — same, from the real `/proc` |
| `strace -e trace=%file` | **0** syscalls naming the bogus tree; the only paths opened are the real `/proc/sys/kernel/{perf_event_paranoid,kptr_restrict}` (`O_RDONLY`) and a `newfstatat` on the real drop-in path |
| bootstrap perf section (seams set, marker unset) | `env_guard` **REFUSES**; section reports SKIPPED; privileged-tool tripwire **empty** — no `sudo`, no `tee`, no `sysctl` |
| bootstrap perf section (no seams, production shape) | names the hardcoded `/etc/sysctl.d/99-cqlite-perf.conf`; only `sudo -n true` recorded; nothing written |

`CQLITE_PERF_TEST_SANDBOX` and any `sysctl.conf` override are not even *referenced* at `a2c4af1` (the
sandbox seam arrives with this work), so they are vacuously inert there. Bootstrap resolves hardcoded
literals; the gate's token resolves hardcoded literals; a seam without the marker is a loud fail-closed
refusal. **Production never sets the marker**, so none of the findings above is a production exposure —
they are test-mode isolation defects, which is why they can ship separately without holding PR #3251.

## Review constraint

Keep the PR **small and under the #3257 threshold** — above roughly 3k changed lines roborev silently
drops the diff and reports PASS, so an oversized PR here would be certified by nothing. That is the whole
reason this work was split out of PR #3251; do not let it re-accumulate. Land it only after PR #3251
merges, so the base is clean.

Refs: #3249, PR #3251, #3257, #3258.

---

## AC4 (added post-filing, from PR #3251's full-range review — roborev job 33, High)

**`perf_capability_env_guard` trusts `CQLITE_PERF_TEST_PRIV_DIR` textually and does not resolve executables.** Setting it to `/usr`, or placing a symlink to the real `sudo`/`sysctl` inside a declared shim directory, passes the guard — so a privileged test-mode bootstrap can execute the real `sysctl --system` and modify host kernel settings.

Canonicalize the sandbox root **and each executable within the shim dir**; require every resolved executable to be positively contained beneath the non-production sandbox root; reject symlinked or system tool targets. Then extend the **structural closed-set audit** already on this branch to cover privilege-shim resolution, not just seam paths.

**This is the eighth escape from this guard family, and every one has the same shape: the guard validated a NAME instead of a DESTINATION.** B3 (raw seam steers a privileged `sudo tee`) → R4-3 (textual non-production checks) → the audit's symlink bypass → R5-1 (`cd -P` canonicalization) → R6-1 (`//etc`, since `pwd -P` can preserve two leading slashes) → R7-1 (symlinked final component: directory containment ≠ write-target containment) → R7-2 (the inversion itself REGRESSED symlink rejection on the fork-free path) → **AC4 (executables, not paths)**. Positive containment fixed it for paths; the same discipline must extend to anything the guard *authorizes*, including the binaries it will run — otherwise this earns a ninth patch.

*(Recorded in issue-comment form on 2026-08-03 and folded into the body here, because a body-only grep under-reads a comment-only AC — noted as a residual by PR #3251's closer.)*

**Production is unaffected** and that was verified, not assumed: with `CQLITE_PERF_TEST_MODE` unset, `strace` recorded **zero** syscalls naming a bogus seam tree, the token read the real `/proc`, and bootstrap's `env_guard` refused with an empty tripwire log.

pmcfadin:
Closes #3261. Later hardening splits to **#3350**; the full pre-split tree is preserved on
`hardening-3261-presplit` (`a3fe54e9f`).

## Scope, and why the seam moved once

Reduced from the full 17-round branch by owner ruling, after roborev's inline review limit (~2151
code-only changed lines) and a growing review-driven diff proved mutually exclusive — the branch could not
be both fully review-hardened and reviewable in one piece. That is what #3257's ceiling exists to force and
what PR #3251 did for the identical cause.

**The first seam was wrong and was corrected.** Splitting at the literal AC boundary produced a tree whose
fresh review immediately raised two Highs and a Low — *all three being defects already fixed on the
hardening branch*, because the split had removed the fixes. Shipping a known High whose fix already exists
is not defensible, so the seam moved to **"everything needed to ship the ACs safely"**: AC1–AC4 plus the
rounds 1–3 write-path fixes. **1823 code-only lines** (328 under the ceiling, so a fix round cannot
re-cross it).

## AC1–AC4

Lands the round-6 positive-containment inversion reverted out of PR #3251. One guard — *"may a test-mode
env seam steer this path?"* — had produced eight escapes across six rounds, each a new **spelling** of
"somewhere else". The lesson is structural: **a denylist over path spellings cannot be completed**, since
the ways to spell `/etc` are unbounded while the ways to prove "strictly inside my sandbox" number one.

- **AC1** — a symlinked managed basename is refused; `dropin_current` will not follow it *even when the
  link target holds byte-identical canonical content*; the write replaces the directory **entry**.
- **AC2** — the fork-free read path rejects a symlinked component, so a symlink inside the sandbox pointing
  at the real `/proc/sys/kernel` yields no capability rather than a fabricated one; the token path's
  fork-free contract is preserved.
- **AC3** — a strictly-contained `<root>/sysctl.conf` is accepted (canonical parent + basename) while
  outside, the root itself, `..`, relative and symlinked-final spellings stay refused.
- **AC4** — privileged executables are validated by destination, and the structural closed-set audit extends
  from seam paths to privilege-shim resolution.
- AC1 **wiring evidence**: bootstrap fails closed when the write target cannot be named, instead of flowing
  an empty path into a privileged write.

## Write-path fixes carried with the ACs (roborev rounds 1–3)

Included because the ACs cannot ship safely without them:

- **staging name unpredictable** — `mktemp` (O_CREAT|O_EXCL) creates it under the same privilege that
  writes it, replacing a fixed name that was checked, cleared and then opened by a privileged `tee`.
- **`mv -fT`** — without `--no-target-directory` a symlink-to-directory at the managed name makes `mv` move
  the staging file *into* it, i.e. the rename that exists to avoid following a symlink follows one.
- **the writability precondition** — the destination must be owned by the privileged writer and be neither
  group- nor world-writable, so there is no actor to race at all. This, not the staging mechanics, is what
  closes the class.
- **CR/LF rejected in path seams** — a *contained* path containing a newline used to split the
  one-per-line search path into two entries, the second being the real `/etc/sysctl.d`.

Two claims in this function were **retracted under review and the retractions are recorded in the code**:
that a fixed staging name was safe because the race "cannot happen", and that one privileged `sh -c`
prevents another process being scheduled between steps. The second was my own error. A reader deserves that
history before trusting the third claim.

## Known, deliberately-unfixed residuals — #3323

All three entries are pointed at **from their code sites**, so a maintainer meets the boundary before
widening the trust, and a reviewer can tell a dispositioned residual from an overlooked defect:

| # | Residual | Why unfixed |
|---|---|---|
| 1 | ancestor-chain rename race | validates the target directory, not the path to it; fix needs `openat2` |
| 2 | privileged-tool revalidation | resolves then discards paths; later callers re-resolve by name |
| 3 | bind mounts defeat lexical containment | `cd -P`/`pwd -P` resolve symlinks, not mounts |

Thirteen escapes established the class is unbounded in shell; by owner ruling it is **closed**, with
`openat2`-style fd-relative containment recorded as the correct fix and an **evidence-only re-raise
trigger**. All three are test-mode only — production never sets `CQLITE_PERF_TEST_MODE`.

## Certification

- **Suites: `test_perf_capability.sh` 69/0, `test_perf_capability_bootstrap.sh` 49/0,
  `test_agent_gate_summary.sh` 130/0.** Both perf suites close by asserting the real
  `/etc/sysctl.d/99-cqlite-perf.conf` is byte- and metadata-identical to its pre-suite state — for an issue
  about test seams reaching production, a suite proving it touched nothing is the right oracle.
- Full gate of record + a fresh inline-verified roborev round (`prompt-content: PASS`): recorded below
  before merge.

## Not in this PR — #3350

The macOS capability-probed counted SKIP, the `UNSUPPORTED`-host rc, content-generator failure
propagation, trailing-slash normalisation, competing-file scan hermeticity, and both audits' de-vacuuming.
All on `hardening-3261-presplit`, re-landing via #3350 — whose scope shrinks accordingly, since rounds 1–3
land here.

## Certification — gate of record + delta re-certification

**Full gate of record: `RESULT: PASS` at `6f75d0be6`** — 30/30 components, **zero SKIPs**,
`tree-integrity: PASS`, `tree-start == tree-end`, no `.integrity-fail` sibling.

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.HL468Q
commit: 6f75d0b branch: issue-3261-seam-containment dirty: no
schemas: 6/6 canonical .cql readable under /home/ubuntu/workspace/cqlite-wt/issue-3261/test-data/schemas (checkout-relative)
accelerators: sccache=on nextest=on lanes=on sccache-health=ok mold=present-unconfigured perf=paranoid-4
tree-start: 6f75d0be6f75 dirty: no digest: d0044464220d
tree-end: 6f75d0be6f75 dirty: no digest: d0044464220d
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (6s)
clippy:            PASS (123s)
roborev-lints:     PASS (42s)
core-tests:        PASS (565s)
tombstones-scan:   PASS (10s)
scan-offload-guard: PASS (123s)
work-counters-guard: PASS (69s)
byte-budget-guard: PASS (17s)
arrow-parity-guard: PASS (56s)
memory-budget:     PASS (63s)
integration-tests: PASS (60s)
format-compat:     PASS (41s)
write-tests:       PASS (149s)
cli-tests:         PASS (173s)
flight-query-semantics-oracle: PASS (79s)
python-bindings:   PASS (189s)
node-bindings:     PASS (123s)
delivery-telemetry: PASS (0s)
oom-audit:         PASS (6s)
parity-report:     PASS (1s)
operator-metrics-doc: PASS (31s)
tooling-tests:     PASS (450s)
minimal-build:     PASS (59s)
smoke:             PASS (25s)
logs: /tmp/agent-gate.HL468Q
summary-file: /tmp/claude-1000/-home-ubuntu-workspace-repo/83af51fe-e3c7-4616-9c80-74d449fcbd7c/scratchpad/gate-final.txt
RESULT: PASS
```

**Delta re-certification: `RESULT: PASS` at `4d18c3d6b`** — the only diff after the anchor is
`scripts/tests/test_perf_capability.sh` (two vacuous test repairs), which is inside `--delta`'s
allowed set. Delta is **not** the gate of record and is recorded here beside the anchor's full
SUMMARY, as required. It fails closed on any non-test/docs path, so it cannot launder a code change.

```

run-id: /tmp/agent-gate.rLhdpl
MODE: delta (TEST/DOCS-ONLY RE-CERTIFICATION — NOT the gate of record; gate of record = the full agent-gate.sh PASS at anchor 6f75d0be6f75a87749ba99bd795241031829a1df)
commit: 4d18c3d branch: issue-3261-seam-containment dirty: no
delta-anchor: 6f75d0be6f75a87749ba99bd795241031829a1df (full-gate PASS commit)
delta-anchor-run-id: /tmp/agent-gate.HL468Q
gate-of-record: full agent-gate.sh run at 6f75d0be6f75a87749ba99bd795241031829a1df (this DELTA re-certifies a test/docs-only diff; it is NOT a substitute for the full gate)
delta-executors: scoped-tests(rust/python) shell-selftests(1) (executors that RAN this re-cert)
accelerators: sccache=on nextest=on lanes=on sccache-health=ok mold=present-unconfigured perf=paranoid-4
tree-start: 4d18c3d6b327 dirty: no digest: ee1a0fa2e5fe
tree-end: 4d18c3d6b327 dirty: no digest: ee1a0fa2e5fe
tree-integrity: PASS
delta-files (1 allowed / 0 offending):
      [test/docs] scripts/tests/test_perf_capability.sh
file-size:         PASS (0s)
fmt:               PASS (4s)
scoped-tests:      PASS (5s)
shell-selftests:   PASS (2s)
logs: /tmp/agent-gate.rLhdpl
summary-file: /tmp/claude-1000/-home-ubuntu-workspace-repo/83af51fe-e3c7-4616-9c80-74d449fcbd7c/scratchpad/delta.txt
RESULT: PASS
```

### DIFF
--- a/docs/development/fleet-runbook.md
+++ b/docs/development/fleet-runbook.md
@@ -327,7 +327,8 @@ grep -Hn 'perf_event_paranoid\|kptr_restrict' /etc/sysctl.conf /etc/sysctl.d/*.c
   /run/sysctl.d/*.conf /usr/local/lib/sysctl.d/*.conf /usr/lib/sysctl.d/*.conf /lib/sysctl.d/*.conf
 ```
 
-**Never set `CQLITE_PERF_PROC_DIR` / `CQLITE_PERF_SYSCTL_DIR` in a shell.** They are test-only path
+**Never set `CQLITE_PERF_PROC_DIR` / `CQLITE_PERF_SYSCTL_DIR` / `CQLITE_PERF_SYSCTL_EXTRA_DIRS` /
+`CQLITE_PERF_TEST_SANDBOX` in a shell.** They are test-only path
 seams for `scripts/tests/test_perf_capability*.sh` and are **inert** unless `CQLITE_PERF_TEST_MODE=1`
 is also set — which in turn refuses to run if a real `sudo`/`sysctl` is reachable. The privileged
 destination is a hardcoded `/etc/sysctl.d` literal precisely so no exported variable can ever steer
@@ -335,23 +336,39 @@ a `sudo tee` at another file, and bootstrap fails closed (skips the section with
 finds a seam set without the marker.
 
 **Test mode has NO production fallback (it is enforced, not conventional).** Under
-`CQLITE_PERF_TEST_MODE=1` **both** path seams are **mandatory** and each must be absolute and outside
-`/etc`, `/proc` and `/sys`. A missing or production-shaped seam is a loud refusal in the env guard
-*and* in the path resolvers, so an unsandboxed test-mode run resolves no drop-in path at all and reads
-no `/proc`. The earlier shape fell back to the real directories, which meant a root `--yes` run under
-the marker could `tee` the host's **real** `/etc/sysctl.d/99-cqlite-perf.conf` — a test run mutating
-the machine. There is deliberately no opt-out.
+`CQLITE_PERF_TEST_MODE=1` the sandbox root and **both** path seams are **mandatory**. An unusable seam
+is a loud refusal in the env guard *and* in the path resolvers, so an unsandboxed test-mode run
+resolves no drop-in path at all and reads no `/proc`. The earlier shape fell back to the real
+directories, which meant a root `--yes` run under the marker could `tee` the host's **real**
+`/etc/sysctl.d/99-cqlite-perf.conf` — a test run mutating the machine. There is deliberately no opt-out.
 
-**On the write path the seam is judged by its CANONICAL DESTINATION, not its spelling.** A textual
-check accepts an unbounded set of paths that *resolve* into production — `/tmp/../etc/sysctl.d`, or a
-seam under a symlinked ancestor (`ln -s /etc /tmp/a`, seam `/tmp/a/sysctl.d`) — and each of those
-would land a root `tee` on the host's own drop-in. So the env guard and the drop-in-path resolver
-canonicalize the whole path (`cd -P` + `pwd -P`: no `realpath`/`readlink -f` dependency, correct on
-bash 3.2) and validate the resolved destination; an unenterable path resolves to nothing and is
-refused too. The gate's **emit-time read path** is contractually fork-free and writes nothing, so it
-keeps the builtin-only textual check (which rejects `.`/`..` components and a symlinked seam): the
-worst a mis-accepted seam can do there is mis-read a stand-in `/proc`, which the token reports as
-`absent`/`unknown` rather than as a capability.
+**The rule is POSITIVE CONTAINMENT in one declared sandbox — not a list of forbidden places.** Four
+review rounds each closed one more *spelling* of "the production directory": the raw path, then a
+symlinked seam, then `..`, then `//etc` (POSIX leaves two leading slashes implementation-defined,
+`pwd -P` may preserve them, and on Linux `//etc` **is** `/etc`). A denylist over path spellings cannot
+be completed — `.`, `..`, symlinks, `//`, trailing slashes, bind mounts, `/proc/self/root/…` all name
+the same directory — and scattered prohibitions also let a *new* seam consumer miss them (that was the
+`CQLITE_PERF_SYSCTL_EXTRA_DIRS` defect). So test mode now takes **one** caller-declared sandbox root,
+`CQLITE_PERF_TEST_SANDBOX`, and a seam is usable **iff it is strictly inside it** (resolved-path prefix
+match with an explicit `/` boundary, so `/tmp/sandboxevil` is not inside `/tmp/sandbox`). Anything not
+provably inside is refused — every spelling above, and every future one, for the same single reason.
+
+- The **root itself must prove it is a sandbox**: absolute, canonically spelled, existing, and holding
+  the stamp file `.cqlite-perf-sandbox`. So `CQLITE_PERF_TEST_SANDBOX=/etc` cannot make containment
+  vacuous — the proof lives on the filesystem, where placing it already needs the privilege the guard
+  protects.
+- **Every** consumer routes through that one check: the env guard, the drop-in path a root `tee` is
+  aimed at, the `/proc` stand-in, and every configuration read (the lower-precedence search-path
+  entries, plus an optional `sysctl.conf` **file** entry, whose parent is canonicalized and the
+  resulting file path validated). `test_perf_capability.sh` enforces that **structurally**: it
+  enumerates from the source every function that dereferences a seam and fails if one does not reach
+  the containment check, so a new entry point cannot silently skip it.
+- Paths that **write** or that **read host configuration** canonicalize both sides (`cd -P` + `pwd -P`:
+  no `realpath`/`readlink -f` dependency, correct on bash 3.2); an unenterable path resolves to nothing
+  and is refused. The gate's **emit-time read path** is contractually fork-free and writes nothing, so
+  it applies the same containment check *syntactically* — its guarantee is that the seams are honoured
+  only under the marker, which is never set in production, and the worst a mis-accepted spelling can do
+  there is read a caller-chosen file, reported as `absent`/`unknown` rather than as a capability.
 
 Every gate SUMMARY's `accelerators:` line stamps the same state as a Linux-only `perf=` token
 (`ok` / `paranoid-<N>` / `kptr-restricted` / `absent` / `unknown`), so "this box cannot be profiled"
--- a/docs/development/fleet-runbook.md
+++ b/docs/development/fleet-runbook.md
@@ -365,10 +365,38 @@ provably inside is refused — every spelling above, and every future one, for t
   the containment check, so a new entry point cannot silently skip it.
 - Paths that **write** or that **read host configuration** canonicalize both sides (`cd -P` + `pwd -P`:
   no `realpath`/`readlink -f` dependency, correct on bash 3.2); an unenterable path resolves to nothing
-  and is refused. The gate's **emit-time read path** is contractually fork-free and writes nothing, so
-  it applies the same containment check *syntactically* — its guarantee is that the seams are honoured
-  only under the marker, which is never set in production, and the worst a mis-accepted spelling can do
-  there is read a caller-chosen file, reported as `absent`/`unknown` rather than as a capability.
+  and is refused.
+
+**Containment of a SPELLING is not containment of a DESTINATION, and that cost four more rounds (#3261).**
+Positive containment closed the path spellings; four escapes remained, each about something the guard
+authorizes *other than* a directory name. All four are now closed by the same discipline — validate the
+destination, positively, fail closed:
+
+- **The write TARGET, not just its directory.** A contained directory says nothing about where its
+  *entries* point, and `tee <path>` opens `O_CREAT|O_TRUNC` and **follows a symlink**. So a symlink at
+  `99-cqlite-perf.conf` inside a perfectly-contained directory aimed the privileged write anywhere on the
+  box. Now: naming that target **refuses** when it is a symlink, the idempotency read never follows it
+  (a link whose target holds the canonical bytes must not report "already current"), and the write is an
+  **atomic directory-entry replacement** — staged entry in the validated directory, then `rename` over
+  the name — so a symlink planted between the check and the write is *replaced*, not written through.
+- **The fork-free read path is NOT exempt** — the earlier claim here that a syntactic check was "sound
+  because nothing there writes" was **wrong**, and this is what falsified it. A symlink *inside* the
+  sandbox pointing at the real `/proc/sys/kernel` satisfies containment, so the run reported a token
+  derived from the **host's real controls** while claiming to read a stand-in (measured: `paranoid-4`
+  straight out of the live `/proc`). A **fabricated verdict is worse than a refusal**. The path stays
+  fork-free and now rejects a **symlinked component** with builtins only (`[ -L ]` forks nothing).
+- **A strictly-contained file must be ACCEPTED.** `<root>/sysctl.conf` was refused, because the check
+  asked whether the *parent* was strictly contained and a root is not strictly inside itself. The judged
+  path is now `<canonical parent>/<basename>`. A guard that refuses legitimate input is the guard people
+  learn to route around.
+- **EXECUTABLES too, not only paths.** `CQLITE_PERF_TEST_PRIV_DIR` was trusted textually: `/usr` is
+  absolute and genuinely *contains* `/usr/bin/sudo`, and a symlink to the real `sudo`/`sysctl` inside a
+  genuine shim dir is spelled locally while resolving to the host's binary — either one let a privileged
+  test-mode bootstrap run a real `sysctl --system` against the host kernel. Every privileged executable's
+  **resolved destination** must now be contained beneath the proven sandbox root, and `sudo`/`sysctl`
+  **parked** in the shim dir are swept whether or not `PATH` reaches them (one `PATH`-order change is all
+  that separates "not resolved" from "executed"). The structural audit covers privilege-shim resolution
+  as a **second pass**, with its own floor and its own named allowlist.
 
 Every gate SUMMARY's `accelerators:` line stamps the same state as a Linux-only `perf=` token
 (`ok` / `paranoid-<N>` / `kptr-restricted` / `absent` / `unknown`), so "this box cannot be profiled"
--- a/docs/development/fleet-runbook.md
+++ b/docs/development/fleet-runbook.md
@@ -1071,3 +1071,250 @@ only fired on rc 0 PLUS a matched line, so a silent accept -- or a skip -- passe
 directory removes the ordering dependency, and the refusal is now REQUIRED rather than merely allowed.
 ```
 
+### and-the-spelling-is-not-the-destination-tmp-e
+
+```
+...and the SPELLING is not the destination. `/tmp/../etc/sysctl.d`, `<symlink-to-/etc>/…`
+and — the R6-1 escape — `//etc/sysctl.d` (POSIX leaves two leading slashes
+implementation-defined and `pwd -P` may PRESERVE them, while on Linux `//etc` IS `/etc`)
+each passed the textual checks of an earlier round. There is no per-spelling check any
+more: containment refuses all of them, plus every future spelling, for the SAME reason.
+An UNENTERABLE path resolves to nothing and is refused too — a write target must exist.
+```
+
+### 1g-3261-a-name-is-not-a-destination-the-four
+
+```
+---- 1g. #3261: A NAME IS NOT A DESTINATION — the four remaining escapes ------------------
+Positive containment closed the PATH SPELLINGS. These four are what containment of a
+spelling still does not buy, and each is asserted BY ITS OWN OBSERVABLE CONSEQUENCE (a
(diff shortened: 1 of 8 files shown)


pmcfadin:
## Full gate at `79a6b73`: **FAIL** — cause found, fixed, re-gating at `513ed4a7e`

Reporting the red rather than the outcome I wanted. 29 of 30 components PASSed; `tooling-tests` FAILed:

```
FAIL - perf-free: the emit-time perf path forks 0 command substitution(s) + 12 backtick(s)
       — documented cost is 0; either remove them or correct the claim in the code comment,
         the spec, gate-contract.md and gate-ops.md
```

**Attribution: mine, and entirely self-inflicted.** The round-31 fix comment I added inside
`perf_capability_sandbox_root_into` quoted path spellings with backticks. That function is in the
**closed fork-free audit set** of `scripts/tests/test_agent_gate_summary.sh`, which *counts*
backticks across the whole function text — comments included — and cannot distinguish a quoted
spelling in prose from a real command substitution. Twelve backticks, twelve counted. The audit was
right and the comment was wrong.

Fixed in `513ed4a7e` by rewording the spellings in words, and by recording the constraint **at the
function** so the next editor learns it from a comment instead of from a red gate. This is the second
time this audit caught prose (an earlier single occurrence), which is why the note is now in-tree.

- `test_agent_gate_summary.sh`: **130/0** (was 129/1) — the failing case now passes
- `test_perf_capability.sh`: **76/0/0**  ·  `test_perf_capability_bootstrap.sh`: **50/0/0**

A fresh **full gate** is running at `513ed4a7e`. `--delta` is not available for it: the diff since
the last full PASS (`6f75d0be6`) touches `scripts/perf-capability.sh`, which delta mode fails closed
on by design — so the gate of record has to be a full one.

### Known next step, stated up front
The code diff is now **2210 lines**, back above this PR's measured roborev inline ceiling (2178
reviews cleanly, 2182 is delivered as a file reference and reds `prompt-content:`). The round-31
fixes plus this one added ~60 lines since the last clean review. Once the gate lands I will relocate
narration from `scripts/tests/test_perf_capability.sh` into `docs/development/fleet-runbook.md` —
the same zero-information-cost move approved earlier, and one that keeps the follow-up diff inside
`--delta`'s allowed set (`scripts/tests/*.sh` + `*.md`), so it re-certifies against this gate
rather than costing a third full run.

pmcfadin:
## roborev CLEAN at `a42ba86ce` — rounds 32–35 resolved, final gate running

Four review rounds since the last update, every finding a true positive. Recording them because each one names a real defect I shipped into the branch, and two were in code an earlier round's fix had just introduced.

| Round | Severity | Defect | Disposition |
|---|---|---|---|
| 32 | Medium | `seam_set` named **2 of 5** non-marker seams, so `TEST_SANDBOX`/`SYSCTL_EXTRA_DIRS`/`TEST_PRIV_DIR` set without the marker passed the env guard | Fixed + enforced by **census** (1c-iv) |
| 32 | Medium | `sandbox_root_into` accepted a root reached through a **symlink**; measured fork-free `sandbox_ok` rc 1 vs resolving `sandbox_ok_resolved` rc **0** on the identical root | Fixed (reject; canonicalizing needs a subshell, forbidden there) |
| 33 | Low | The per-seam loop inherited the suite-wide exported `CQLITE_PERF_TEST_SANDBOX` → **most iterations vacuous** | Fixed; negative control added |
| 33 | Low | Trailing-slash `PROC_DIR` → `//` entry path → reported as the capability verdict **`absent`** | Fixed (normalise *before* the gate) |
| 34 | Medium | The install probe `rm -f`'d predictable `.perfcap-probe.<pid>` names **before** proving absence | Fixed (deletes nothing it did not create) |
| 34 | Low | Every rename-probe failure reported as rc 2 UNSUPPORTED, suppressing the retry remedy | Fixed (capability and outcome are separate questions) |
| 35 | Medium | Resolving validators judged only the canonicalized result while consumers keep the **original** spelling | **Split**: spelling half fixed, re-resolve half recorded |

Two of these deserve calling out rather than burying in a table:

**The round-33 vacuity finding was Low-labelled and substantively a blocker.** The per-seam loop inherited an exported seam, so once round 32's fix named that seam in `seam_set`, the guard refused through the *inherited* variable no matter which seam was under test. It was invisible for the worst possible reason: before the fix `seam_set` did not name that seam, so the loop measured correctly and went red on exactly the three missing ones — **the fix that turned it green is what would have made it vacuous.** Every seam is now unset per iteration, and a negative control (revert `seam_set` to two names → the loop reds again on those three) is recorded in the commit.

**Round 35 was the closed validate-name/re-resolve family, so it was split rather than waived whole.** The *spelling* half is not part of that family and is fixed: both resolving validators now require the original argument to be absolute and free of `./`/`../`, the rule `sandbox_root_into` has enforced for the root all along — which eliminates the concrete case raised (a relative candidate authorized against one cwd, re-resolved against another inside the privileged shell). The *residual* half — a verdict returned while the resolved path is discarded, so consumers resolve the name twice — is recorded as **#3323 entry 5** with in-tree pointers at both sites, per the standing disposition: logged, not re-attempted.

Every fix was proven **RED-first** against the pre-fix library, and every new case carries a control asserting the legitimate spelling is still **accepted**, so a refuse-everything rule would fail just as loudly as an accept-everything one. The PID-collision half of round 34 cannot be staged from outside the privileged shell, so it is asserted **structurally** (no delete may appear before the absence proof) — stated as such rather than dressed up as behavioural coverage.

### Certification state
- **roborev: `RESULT: PASS`, `findings: NONE`** at `a42ba86ce` (job 36) — every wrapper key affirmatively PASS, `sha-assert` included; 219k input / 156k cached / 14.9k output, i.e. a genuine review, not the 18.7k/0-cached vacuous shape
- Branch **rebased onto `origin/main`** (`26399bb`), which is also what `sha-assert` requires — job 32 correctly FAILed it when main had advanced 2 commits past my base
- Code census held under this PR's measured inline ceiling (2178 reviews inline / 2182 becomes a file reference) by relocating narration to `docs/development/fleet-runbook.md` — nothing deleted, since markdown is excluded from the census by configuration
- Suites: `test_perf_capability` **82/0/0**, bootstrap **50/0/0**, `test_agent_gate_summary` **130/0**
- **Full gate at `a42ba86ce` is running now.** The earlier PASS at `513ed4a7e` no longer certifies this head — the code changed after it, so it is superseded, not reused.

Merging on that gate, per instruction, if it is clean.

Issue #3370, opened 2026-08-27
Python bindings dataset guard is corpus-wide, not per-keyspace: an unrelated keyspace satisfies it while the test's own fixtures are absent


pmcfadin:
Split out of #1458 (dataset guards are directory-only) — raised by roborev job 9 on PR for that issue, and deliberately NOT fixed there because the fix touches every caller.

## The gap

`bindings/python/tests/conftest.py::skip_if_no_datasets()` (as of #1458) is **content-aware but corpus-WIDE**: it passes when at least one `*-Data.db` file exists anywhere under `DATASETS`. It has no notion of *which* keyspace the calling test actually needs.

So a corpus holding, say, only `test_collections/` satisfies the guard for a `test_basic` test. The test then runs against absent fixtures and fails confusingly — or, worse, returns 0 rows and passes.

The Node guard does not have this gap: `bindings/node/__test__/setup.js` scopes its check to `test_basic` specifically (restored in #1458). Python is the looser of the two.

## Why #1458 did not fix it

#1458's scope was "a present-but-EMPTY corpus must not pass the strict guard", and it delivered that in both languages. Per-keyspace precision is a different, larger property:

- Node has ONE global `DATASETS_AVAILABLE` that always required `test_basic`, so scoping it there restored its original semantics — a 1-line change.
- Python's `skip_if_no_datasets()` is a **shared** helper called across four keyspaces (`test_basic`, `test_collections`, `test_timeseries`, `test_wide_rows`). Hardcoding `test_basic` into it would be actively wrong for a `test_collections` test. Doing it properly means giving each call site its own required-fixture declaration — ~16 test files.

That is a scope expansion beyond #1458's two prescribed edits, so it is split rather than absorbed (1:1:1:1).

## This is the documented #3220 shape, one language over

CLAUDE.md already carries the doctrine this issue is an instance of — **"Resolve fixture roots per TABLE, and assert per CASE (#3220)"**:

> a lane that picks its corpus root by KEYSPACE (`root.join(keyspace).is_dir()`) and commits to it can pass without ever running

and the companion rule that a corpus loop must **never** terminate in a suite-wide `assert!(ran > 0)`, because that cannot see one case skipping behind its siblings. The Rust side got `sstables_root_for_table` for exactly this. The Python bindings suite has no equivalent.

## Suggested shape

[...]

PR #3371, opened 2026-08-27
fix(#1458): dataset guards fail closed on a present-but-empty corpus (Python + Node strict fixtures)

### LINKED ISSUE #1458: Dataset guards are directory-only: present-but-empty root still passes; Node has no strict-fixtures flag
Parent: #1436

## Why this matters
Dataset-dependent tests are supposed to fail loudly in CI when the SSTable binaries are missing (issue #1230), so a dropped table or a path regression reds the build instead of silently passing by skipping. But the current strict guard only checks that the datasets *directory* exists — a present-but-empty directory (the exact shape of the original #773 failure) still passes. And the Node suite has no strict mode at all: it just skips. This lets a broken fixture setup false-green.

Python (`bindings/python/tests/conftest.py:79-93`) — directory-only guard:
```python
def skip_if_no_datasets():
    if not DATASETS.exists():
        msg = f"Test data not found: {DATASETS}"
        if _require_fixtures_strict():
            pytest.fail(..., pytrace=False)
        pytest.skip(msg)
```
`_require_fixtures_strict()` (`:64-76`) honors `CQLITE_REQUIRE_FIXTURES` / `CQLITE_PARITY_REQUIRE_DATASETS`. A `DATASETS` dir that exists but contains zero `*-Data.db` files passes this guard.

Node (`bindings/node/__test__/setup.js:40-41`) — directory-only, no strict mode:
```js
const DATASETS_AVAILABLE = fs.existsSync(SSTABLES_DIR) &&
  fs.existsSync(path.join(SSTABLES_DIR, 'test_basic'));
```
`bindings/node/__test__/helpers.js:38-42` `skipIfNoDatasets()` throws when `!DATASETS_AVAILABLE` (good — throws not skips), but only keys off the directory existing; tests otherwise gate on `global.DATASETS_AVAILABLE` via `describe.skip`.

**Python — `bindings/python/tests/conftest.py`:**
1. Add a helper `_count_data_db() -> int` that globs `DATASETS.glob("**/*-Data.db")` and returns the count (guard against `DATASETS` not existing → return 0).
2. In `skip_if_no_datasets()`, after the existing `DATASETS.exists()` branch, add: under strict mode (`_require_fixtures_strict()`), if the dir exists but `_count_data_db() == 0`, `pytest.fail("Datasets dir present but contains 0 *-Data.db files (CQLITE_REQUIRE_FIXTURES=1) — fetch with test-data/scripts/fetch-datasets.sh", pytrace=False)`. Non-strict keeps skipping.

**Node — `bindings/node/__test__/setup.js`:**
1. Add content-awareness to `DATASETS_AVAILABLE`: it should require at least one `*-Data.db` under `SSTABLES_DIR` (a shallow recursive walk of `SSTABLES_DIR` for a file ending in `-Data.db`), not just the `test_basic` dir existing.
2. Add strict handling: read `const REQUIRE_FIXTURES = process.env.CQLITE_REQUIRE_FIXTURES === '1' || process.env.CQLITE_PARITY_REQUIRE_DATASETS === '1';`. If `REQUIRE_FIXTURES && !DATASETS_AVAILABLE`, **throw** from setup.js (fails the whole suite) rather than leaving `DATASETS_AVAILABLE=false` for `describe.skip`. Export `global.REQUIRE_FIXTURES` for helpers.
3. Update `bindings/node/__test__/helpers.js:38` `skipIfNoDatasets()` to keep throwing (it already does); no `describe.skip` should silently pass under strict mode.

## Do NOT
- Do NOT change the non-strict local-dev behavior: with neither env var set and no binaries, Python still `pytest.skip`s and Node still `describe.skip`s — developers without the ~GB of fixtures must not be blocked.
- Do NOT alter the `_require_fixtures_strict()` env-var names or the session no-op floor (`conftest.py:435-458`); this issue adds content-awareness alongside them, not instead of them.
- This is test-harness only — do NOT touch binding source.

## Tests you must add
- `bindings/python/tests/test_dataset_guard.py::test_strict_fails_on_empty_datasets_dir` — TDD: point `DATASETS` at a tmp dir containing an empty `test_basic/` subdir (no `*-Data.db`), set `CQLITE_REQUIRE_FIXTURES=1`, and assert `skip_if_no_datasets()` raises `Failed`/`pytest.fail` (use `pytest.raises` around a monkeypatched `DATASETS`, or run a subprocess pytest and assert non-zero exit). Must FAIL on `main` today (passes-by-skip).
  ```bash
  env CQLITE_REQUIRE_FIXTURES=1 pytest bindings/python/tests/test_dataset_guard.py -v
  ```
- `bindings/node/__test__/dataset-guard.test.js` — assert that with an empty sstables dir + `CQLITE_REQUIRE_FIXTURES=1`, setup throws (spawn a child `jest` run pointed at an empty `CQLITE_DATASETS_ROOT` and assert non-zero exit).
  ```bash
  cd bindings/node && npm run build && CQLITE_REQUIRE_FIXTURES=1 CQLITE_DATASETS_ROOT=/tmp/empty npx jest dataset-guard
  ```
- FAIL LOUDLY rule: these guards ARE the fail-loudly mechanism — verify they hard-fail, never skip, when the dir is present-but-empty under strict mode.

## Definition of done
- [ ] Empty `test_basic/` dir + strict flag → Python suite FAILS and Node suite FAILS (both verified to have passed-by-skip before the change)
- [ ] Non-strict local dev still skips gracefully (no regression for developers without binaries)

pmcfadin:
Closes #1458.

## What was wrong

Dataset-dependent tests are supposed to fail loudly under strict mode when the SSTable binaries are missing (#1230), so a dropped table or a path regression reds CI instead of passing by skipping. Both guards only checked that the datasets **directory** existed:

- **Python** (`conftest.py`): a `DATASETS` dir that exists but holds zero `*-Data.db` passed the strict guard.
- **Node** (`setup.js`): directory-only *and* no strict mode at all — it just `describe.skip`ped.

A present-but-empty corpus is the exact shape of the original #773 failure, so this false-greened a broken fixture setup.

## What changed (test harness only — no binding source touched)

- `bindings/python/tests/conftest.py` — the guard is content-aware: a present dir holding zero `*-Data.db` **files** now fails closed under strict mode, skips under non-strict.
- `bindings/node/__test__/setup.js` — content-aware **and** strict-capable: under `CQLITE_REQUIRE_FIXTURES` / `CQLITE_PARITY_REQUIRE_DATASETS` it **throws from setup**, failing the whole suite instead of leaving `DATASETS_AVAILABLE=false` for `describe.skip` to report a green run over zero real assertions.
- Two new test files pinning the guards themselves, in both directions.

Non-strict local dev is unchanged: with neither env var set, Python skips and Node skips. A developer without the ~GB of fixtures is never blocked.

### Traversal is ONE contract written twice

Both guards now implement identical rules, and both say so in comments naming the other:

1. a symlinked **directory** is NOT traversed;
2. a symlinked **file** whose target is a regular file IS counted;
3. a FIFO, socket, device node, or **directory** named `*-Data.db` is NOT counted — only a regular file qualifies.

Python gets these free from `Path.glob("**/*-Data.db") + is_file()`; Node's walk was aligned to match. I measured all three on this interpreter rather than trusting the docs. Consequence recorded honestly in both files: a symlinked *keyspace directory* is unsupported in both languages — no corpus uses that layout (zero symlinks measured under the fleet corpus and the checkout's `test-data/datasets`), and it fails in the safe direction (reports zero fixtures ⇒ strict mode throws the loud error).

## TDD evidence — measured, not asserted

Every case was observed RED against the unfixed parent commit and GREEN after, by reconstructing the new tests against a detached worktree of the pre-fix source:

| Case | RED at | GREEN |
|---|---|---|
| Python: strict + empty dir ⇒ `Failed` | `origin/main` (3 failed) | HEAD |
| Python: non-strict + empty dir ⇒ skips | `origin/main` | HEAD |
| Python: strict + `*-Data.db` present ⇒ no raise | `origin/main` | HEAD |
| Python: strict + `*-Data.db` is a **directory** ⇒ `Failed` | `92bb63ba6` — `_count_data_db()` returned **1** for a directory | HEAD |
| Node: strict + empty dir ⇒ non-zero exit | `origin/main` (2 failed) | HEAD |
| Node: non-strict + empty dir ⇒ exit 0 | `origin/main` — `DATASETS_AVAILABLE` was **true** for an empty `test_basic/` | HEAD |
| Node: strict + empty `test_basic` but data in another keyspace ⇒ non-zero | `a730ecca5` (only this case failed) | HEAD |
| Node: only `*-Data.db` entry is a **FIFO** ⇒ non-zero | `538ab236b` (only this case failed) | HEAD |

Final: Python **4 passed**, Node **6 passed**.

Neither test file is invariant to the bug (#3042): the Python tests drive the real `conftest.skip_if_no_datasets()` patching only `DATASETS`, and the Node tests drive the real `setup.js` in a child jest process — never a reimplementation of the walk. The Node harness explicitly `delete`s both strict env vars so a CI box running with `CQLITE_REQUIRE_FIXTURES=1` cannot bleed into the non-strict control case.

**One case is explicitly NOT a red-then-green pin**, and says so in an `HONESTY NOTE` in the code: the self-referential-symlink case. Measured on Linux the pre-fix walk did not blow the stack — the kernel's 40-level symlink cap makes `statSync` fail `ELOOP` at depth ~81, which the broken-symlink `continue` swallowed. So roborev's "unbounded recursion crashes setup" scenario does not reproduce as stated; the case is retained as a behavioral control (symlinked dir skipped, non-strict never crashes) and is labelled as such rather than presented as a regression pin.

## No regression against the real corpus

Tightening a guard is exactly where you break the working case, so this was verified directly rather than assumed. Against `CQLITE_DATASETS_ROOT=/data/datasets` (155 `*-Data.db`, `test_basic/` holding 8), `DATASETS_AVAILABLE` is **true** in both non-strict and `CQLITE_REQUIRE_FIXTURES=1` modes, and the Python suite runs (not skips) — `python-tier: PASS` in the lite summary runs `pytest bindings/python/tests -m 'not slow' -q`.

## Review record — stated precisely

- **roborev: CLEAN** at head `9db4948b7` — job 10, `RESULT: PASS`, `findings: NONE`, `reviewed-sha 2bde26a7c..9db4948b7`, `prompt-content: PASS (4/4)`, tokens 88.9k in / 74.8k cached (genuine range, vs the ~18.7k/0 vacuous baseline). Reached over **four** rounds: 3 findings → 2 → 2 → 0, every round via `scripts/flow/roborev-review.sh --agent codex --model gpt-5.6-sol --repo <abs> --base origin/main`.
- **`rust-reviewer` produced no retrievable output** across three attempts (the subagent went idle without its report reaching the lead, as did two others this session). It contributed nothing to this review round and is **not** being recorded as a clean pass. The third read was done by the lead directly, line by line, which is what surfaced the efficiency defect below.

### Findings fixed in review rounds

| Round | Finding | Resolution |
|---|---|---|
| 1 | Node content check **replaced** the `test_basic` requirement instead of adding to it — one stray `*-Data.db` anywhere re-enabled every dataset test | Require content **inside** `test_basic`: strictly stronger than either half alone |
| 1 | `Path.glob` yields **directories**, so a dir named `placeholder-Data.db` satisfied strict mode | `is_file()` filter + regression test |
| 1 | Node walk followed directory symlinks with no visited set | Superseded in round 2 — rule 1 makes cycles structurally unreachable |
| 1 (lead) | `_count_data_db()` walked the **entire corpus on every guarded test** (155 files / 122 dirs, hundreds of times per session) to answer "is there ≥1?" | Short-circuiting `_has_data_db()` on the hot path; `_count_data_db()` retained as the diagnostic/AC counter. Both share one `_data_db_files()` generator so they cannot disagree |
| 2 | Python/Node traversal semantics disagreed (symlinks, non-regular files) | Aligned Node to Python's stdlib semantics — a net code **reduction**, deleting the realpath/visited-set machinery, rather than adding a custom cycle-safe walk to both languages for a layout no corpus uses |
| 3 | `spawnSync` had no timeout | See below |
| 3 | Python guard is corpus-wide while Node is `test_basic`-scoped | **Split to #3370** — out of scope, see below |

The `spawnSync` finding was rated Low but is worse than that: jest's per-test timeout cannot interrupt a synchronous spawn, and a timed-out child yields `status === null` while the strict cases assert `status !== 0` — so a hang would have **masqueraded as the strict-mode failure those cases exist to prove**. Now bounded at 90s with an explicit throw on `result.error` or a null status, so a harness failure can never be read as a guard verdict.

## Deliberate scope split

Round 3 correctly observed that Python's guard is corpus-wide: content in `test_collections/` satisfies it for a `test_basic` test. That is real, but it is a **different and larger property** than "an empty corpus must not pass", and fixing it properly means per-call-site fixture declarations across ~16 test files — the documented #3220 "resolve per TABLE, assert per CASE" shape, one language over. Filed as **#3370** rather than absorbed, per 1:1:1:1. Neither language regressed here: Python is strictly stronger than before this PR, just less precise than Node, as it always was.

## Deviations from the issue's prescribed steps

1. The issue had `skip_if_no_datasets()` call `_count_data_db() == 0`. The guard calls a short-circuiting `_has_data_db()` instead (identical observable behavior, O(1) instead of O(corpus) per test); `_count_data_db()` is retained exactly as specified for the acceptance criteria and diagnostics.
2. The issue specified `process.env.CQLITE_REQUIRE_FIXTURES === '1'` for Node. Both `'1'` and `'true'` are accepted, mirroring Python's `_require_fixtures_strict()`, so the same CI env cannot mean strict in one language and non-strict in the other. No other env var names added.

## Verification limits — read before trusting the green

The **Node native module is not built in these worktrees** (`@cqlite/node has no prebuilt binary for linux-x64`), so 22 of 26 node suites fail at `require` time. This is **pre-existing, not from this change** — measured on a clean `origin/main` worktree: main is **23 failed / 3 passed**, this branch **22 failed / 4 passed** (this branch is strictly better; it adds the passing guard suite). The full gate has **no node tier**, so the Node half of this change is evidenced by the guard suite (which does not need the native module) plus a direct `DATASETS_AVAILABLE` probe against the real corpus — *not* by a green full node suite. Stating this explicitly rather than implying broader coverage.

## Definition of done

- [x] Empty `test_basic/` + strict flag ⇒ Python FAILS and Node FAILS, both verified to have passed-by-skip before the change
- [x] Non-strict local dev still skips gracefully
- [x] File-size ratchet respected (`file-size: PASS`)
- [x] No `unwrap()`/`expect()` introduced in library code (no library code touched)
- [x] Full `scripts/agent-gate.sh` PASS — gate of record `RESULT: PASS`, run-id `/tmp/agent-gate.6J4mAv`, `tree-integrity: PASS`, commit `9db4948`; 30 components all PASS, zero SKIP. SUMMARY pasted in a comment below.

### DIFF
--- a/bindings/node/__test__/dataset-guard.test.js
+++ b/bindings/node/__test__/dataset-guard.test.js
@@ -1,0 +1,93 @@
+/**
+ * Tests for the dataset-availability guard itself (issue #1458).
+ *
+ * The guard IS the fail-loudly mechanism for missing SSTable fixtures, so it
+ * needs its own coverage: a present-but-EMPTY sstables directory (the exact
+ * shape of the original #773 failure) must FAIL the suite under strict mode,
+ * never silently `describe.skip`.
+ *
+ * These tests drive the REAL `setup.js` in a child `jest` process — they never
+ * reimplement the directory walk — because a test that reimplements the logic
+ * it asserts on is invariant to the bug (#3042).
+ */
+const { spawnSync } = require('child_process');
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+
+const SETUP_JS = path.join(__dirname, 'setup.js');
+const PACKAGE_DIR = path.dirname(__dirname);
+
+/**
+ * Locate the jest CLI entry point. jest's own module resolver refuses
+ * `require.resolve('jest/bin/jest.js')` (package `exports` map), so resolve the
+ * conventional install locations directly -- the same path package.json's
+ * `test` script uses. Missing => throw (fail loudly), never skip.
+ *
+ * @returns {string}
+ */
+function resolveJestBin() {
+  const candidates = [
+    path.join(PACKAGE_DIR, 'node_modules', 'jest', 'bin', 'jest.js'),
+    path.join(PACKAGE_DIR, '..', '..', 'node_modules', 'jest', 'bin', 'jest.js'),
+  ];
+  for (const candidate of candidates) {
+    if (fs.existsSync(candidate)) return candidate;
+  }
+  throw new Error(`jest CLI not found; looked in:\n  ${candidates.join('\n  ')}`);
+}
+
+/**
+ * Run a child jest process whose ONLY setup file is the real setup.js, against
+ * a datasets root that exists but holds zero *-Data.db files.
+ *
+ * @param {{strict: boolean}} opts
+ * @returns {{status: number, stdout: string, stderr: string}}
+ */
+function runChildJest({ strict }) {
+  const tmp = fs.mkdtempSync(path.join(os.tmpdir(), 'cqlite-dataset-guard-'));
+  // Present-but-empty corpus: sstables/test_basic/ exists, no *-Data.db anywhere.
+  fs.mkdirSync(path.join(tmp, 'datasets', 'sstables', 'test_basic'), { recursive: true });
+  fs.writeFileSync(
+    path.join(tmp, 'noop.test.js'),
+    "test('noop', () => { expect(global.DATASETS_AVAILABLE).toBe(false); });\n"
+  );
+
+  const config = JSON.stringify({
+    testEnvironment: 'node',
+    rootDir: tmp,
+    testMatch: ['**/*.test.js'],
+    setupFilesAfterEnv: [SETUP_JS],
+  });
+
+  const env = { ...process.env, CQLITE_DATASETS_ROOT: path.join(tmp, 'datasets') };
+  delete env.CQLITE_REQUIRE_FIXTURES;
+  delete env.CQLITE_PARITY_REQUIRE_DATASETS;
+  // Jest sets these for its own workers; leaking them into a nested run confuses it.
+  delete env.JEST_WORKER_ID;
+  if (strict) env.CQLITE_REQUIRE_FIXTURES = '1';
+
+  const result = spawnSync(process.execPath, [resolveJestBin(), '--config', config, '--ci'], {
+    env,
+    encoding: 'utf8',
+    cwd: tmp,
+  });
+  return { status: result.status, stdout: result.stdout || '', stderr: result.stderr || '' };
+}
+
+describe('dataset guard (issue #1458)', () => {
+  test('strict mode fails the suite on a present-but-empty datasets dir', () => {
+    const { status, stdout, stderr } = runChildJest({ strict: true });
+    const output = `${stdout}${stderr}`;
+    expect(status).not.toBe(0);
+    expect(output).toMatch(/-Data\.db/);
+  }, 120000);
+
+  test('non-strict mode still passes (skips) on a present-but-empty datasets dir', () => {
+    const { status, stdout, stderr } = runChildJest({ strict: false });
+    if (status !== 0) {
+      throw new Error(`expected exit 0, got ${status}\n${stdout}${stderr}`);
+    }
+    expect(status).toBe(0);
+  }, 120000);
+});
--- a/bindings/node/__test__/dataset-guard.test.js
+++ b/bindings/node/__test__/dataset-guard.test.js
@@ -16,6 +16,8 @@ const os = require('os');
 const path = require('path');
 
 const SETUP_JS = path.join(__dirname, 'setup.js');
+// Upper bound on a nested jest run (see the spawnSync call for why it matters).
+const CHILD_TIMEOUT_MS = 90000;
 const PACKAGE_DIR = path.dirname(__dirname);
 
 /**
@@ -162,7 +164,28 @@ function runChildJest({ strict, corpus = 'empty' }) {
     env,
     encoding: 'utf8',
     cwd: tmp,
+    // A finite bound is load-bearing, not hygiene: jest's own per-test timeout
+    // cannot interrupt a SYNCHRONOUS spawnSync, so without this a wedged child
+    // stalls the whole test job indefinitely. Kept below the 120s per-test
+    // timeout so a hang surfaces as the explicit throw below.
+    timeout: CHILD_TIMEOUT_MS,
+    killSignal: 'SIGKILL',
   });
+
+  // A timed-out or unstartable child yields status === null, and the strict
+  // cases assert `status !== 0` -- so an unreported hang would MASQUERADE as
+  // the strict-mode failure they are trying to prove. Fail loudly instead.
+  if (result.error) {
+    throw new Error(
+      `child jest did not complete (corpus=${corpus} strict=${strict}): ${result.error.message}`
+    );
+  }
+  if (result.status === null) {
+    throw new Error(
+      `child jest was killed without an exit status (corpus=${corpus} strict=${strict}; ` +
+      `signal=${result.signal}) -- treated as a harness failure, never as a guard verdict`
+    );
+  }
   return { status: result.status, stdout: result.stdout || '', stderr: result.stderr || '' };
 }
 
--- a/bindings/node/__test__/dataset-guard.test.js
+++ b/bindings/node/__test__/dataset-guard.test.js
@@ -70,6 +70,26 @@ const CORPUS_SHAPES = {
     },
     childTest: "test('noop', () => {});\n",
   },
+  // A self-referential symlink inside test_basic. Symlinked keyspace dirs are
+  // followed on purpose, so the walk must terminate on a link pointing at an
+  // ancestor and report an unavailable corpus rather than crashing setup --
+  // which it must never do in NON-strict mode (issue #1458).
+  //
+  // HONESTY NOTE: this case is a behavioral CONTROL, not a red-then-green
+  // regression pin. Measured on Linux, the pre-visited-set walk did not blow
+  // the stack: the kernel's 40-level symlink cap makes statSync() fail ELOOP at
+  // recursion depth ~81, which the broken-symlink `continue` swallows. The
+  // visited-set still earns its place (it drops those ~80 redundant
+  // readdir/stat rounds and re-walks of diamond-linked dirs), and this case
+  // pins the observable contract on every platform.
+  symlinkCycle: {
+    build: (sstables) => {
+      const testBasic = path.join(sstables, 'test_basic');
+      fs.mkdirSync(testBasic, { recursive: true });
+      fs.symlinkSync(sstables, path.join(testBasic, 'loop'), 'dir');
+    },
+    childTest: "test('noop', () => { expect(global.DATASETS_AVAILABLE).toBe(false); });\n",
+  },
 };
 
 /**
@@ -138,5 +158,16 @@ describe('dataset guard (issue #1458)', () => {
     }
     expect(output).toMatch(/test_basic/);
     expect(output).toMatch(/-Data\.db/);
+  }, 120000);
+
+  // The walk follows symlinked dirs, so a link pointing at an ancestor must
+  // terminate quietly (exit 0, corpus unavailable) rather than crash setup.
+  test('a self-referential symlink does not crash the walk (non-strict)', () => {
+    const { status, stdout, stderr } = runChildJest({ strict: false, corpus: 'symlinkCycle' });
+    const output = `${stdout}${stderr}`;
+    if (status !== 0) {
+      throw new Error(`expected exit 0, got ${status}\n${output}`);
+    }
+    expect(output).not.toMatch(/Maximum call stack/);
   }, 120000);
 });
--- a/bindings/node/__test__/dataset-guard.test.js
+++ b/bindings/node/__test__/dataset-guard.test.js
@@ -38,20 +38,53 @@ function resolveJestBin() {
 }
 
 /**
- * Run a child jest process whose ONLY setup file is the real setup.js, against
- * a datasets root that exists but holds zero *-Data.db files.
+ * Corpus shapes a child jest run can be pointed at. Each shape owns both the
+ * on-disk layout AND the child test body, because the body decides what a
+ * non-zero exit is allowed to mean.
  *
- * @param {{strict: boolean}} opts
+ * @type {Record<string, {build: (sstables: string) => void, childTest: string}>}
+ */
+const CORPUS_SHAPES = {
+  // Present-but-empty: sstables/test_basic/ exists, no *-Data.db anywhere.
+  // The body asserts the guard's own verdict, so the non-strict control proves
+  // the run both survived AND saw an unavailable corpus.
+  empty: {
+    build: (sstables) => {
+      fs.mkdirSync(path.join(sstables, 'test_basic'), { recursive: true });
+    },
+    childTest: "test('noop', () => { expect(global.DATASETS_AVAILABLE).toBe(false); });\n",
+  },
+  // test_basic/ present but EMPTY, while a DIFFERENT keyspace holds a
+  // *-Data.db. A corpus-WIDE content check that dropped the test_basic
+  // requirement reports this as available and then enables the
+  // test_basic-dependent suites, which cannot possibly pass (issue #1458).
+  //
+  // The body here is deliberately assertion-FREE: the guard throwing must be
+  // the only possible source of a non-zero exit, otherwise a failing child
+  // assertion would green this test against the very bug it pins.
+  otherKeyspaceOnly: {
+    build: (sstables) => {
+      fs.mkdirSync(path.join(sstables, 'test_basic'), { recursive: true });
+      fs.mkdirSync(path.join(sstables, 'test_collections'), { recursive: true });
+      fs.writeFileSync(path.join(sstables, 'test_collections', 'nb-1-big-Data.db'), '');
+    },
+    childTest: "test('noop', () => {});\n",
+  },
+};
+
+/**
+ * Run a child jest process whose ONLY setup file is the real setup.js, against
+ * a synthesized datasets root of the requested shape.
+ *
+ * @param {{strict: boolean, corpus?: keyof typeof CORPUS_SHAPES}} opts
  * @returns {{status: number, stdout: string, stderr: string}}
  */
-function runChildJest({ strict }) {
+function runChildJest({ strict, corpus = 'empty' }) {
+  const shape = CORPUS_SHAPES[corpus];
+  if (!shape) throw new Error(`unknown corpus shape: ${corpus}`);
   const tmp = fs.mkdtempSync(path.join(os.tmpdir(), 'cqlite-dataset-guard-'));
-  // Present-but-empty corpus: sstables/test_basic/ exists, no *-Data.db anywhere.
-  fs.mkdirSync(path.join(tmp, 'datasets', 'sstables', 'test_basic'), { recursive: true });
-  fs.writeFileSync(
-    path.join(tmp, 'noop.test.js'),
-    "test('noop', () => { expect(global.DATASETS_AVAILABLE).toBe(false); });\n"
-  );
+  shape.build(path.join(tmp, 'datasets', 'sstables'));
+  fs.writeFileSync(path.join(tmp, 'noop.test.js'), shape.childTest);
 
   const config = JSON.stringify({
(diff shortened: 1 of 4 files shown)


pmcfadin:
## ⛔ Gate of record: `RESULT: FAIL` — NOT merging; investigating provenance

The full gate ran once on the certified head and **failed**. Recording it immediately rather than re-running to get a nicer answer.

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.seKiSr
commit: 9db4948 branch: issue-1458-dataset-guards-directory-only dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-1458/test-data/schemas (checkout-relative)
tree-start: 9db4948b7998 dirty: no digest: dfc1d219ed56
tree-end:   9db4948b7998 dirty: no digest: dfc1d219ed56
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (7s)
clippy:            PASS (6s)
roborev-lints:     PASS (43s)
core-tests:        PASS (54s)
tombstones-scan:   PASS (1s)
scan-offload-guard: PASS (2s)
byte-budget-guard: PASS (13s)
arrow-parity-guard: PASS (1s)
memory-budget:     PASS (13s)
integration-tests: PASS (3s)
format-compat:     PASS (0s)
write-tests:       FAIL (173s)
cli-tests:         FAIL (0s)
compaction-byte-parity: FAIL (1s)
flight-query-semantics-oracle: FAIL (15s)
python-bindings:   PASS (54s)
node-bindings:     PASS (124s)
delivery-telemetry: PASS (0s)
oom-audit:         PASS (14s)
parity-report:     PASS (1s)
operator-metrics-doc: PASS (34s)
kit-dashboard-drift: PASS (183s)
tooling-tests:     FAIL (53s)
minimal-build:     PASS (449s)
smoke:             PASS (4s)
logs: /tmp/agent-gate.seKiSr
RESULT: FAIL
```

### What passed is relevant

`python-bindings: PASS` and `node-bindings: PASS` — the two components that actually exercise this diff. `tree-integrity: PASS`, `commit:` matches the certified head, tree clean at both ends.

### The CITE-AND-WAIVE waiver does NOT apply here, and I ran the check rather than eyeballing it

Per #3250, the "no compiled input" qualifier is a path-shape test and a path shape is not evidence — so:

```
$ git diff --name-only origin/main...HEAD | bash scripts/ci/classify-docs-only.sh
full
classify-docs-only: non-docs file 'bindings/node/__test__/dataset-guard.test.js' -> FULL PATH (fail-closed)
EXIT=1
```

Non-zero ⇒ **the waiver is void and these failures are presumed MINE until proven otherwise.** I could argue that four files under `bindings/*/tests|__test__/` cannot change a Rust binary, but that argument *is* the path-shape judgement #3250 exists to forbid. It is not evidence.

### So it is being proven, not asserted

A detached gate is running the **same five components** on a clean `origin/main` worktree (`2bde26a7c`, same `CQLITE_DATASETS_ROOT=/data/datasets`):

```
scripts/agent-gate.sh --only write-tests,cli-tests,compaction-byte-parity,flight-query-semantics-oracle,tooling-tests
```

- **Fail identically on main** ⇒ pre-existing main-reds. I will then record a waiver naming **each** component and its issue number (a waiver with no cited issue is just an unexplained red), filing whatever is not already filed.
- **Pass on main** ⇒ the failures are genuinely mine, and I debug them. No waiver, no merge.

Candidate pre-existing issues found so far — **candidates, not yet proof**: #2596 (`write-tests` flake: contaminated `CHUNK_READ_CALLS` under thread-parallel cargo test), #2188 (4 pre-existing `cqlite-cli` integration tests bit-rotted against regenerated dataset binaries), #2790 (`tooling-tests` flake: `test_worker_supervisor.sh` exit-latency threshold under gate CPU load). Nothing filed for `compaction-byte-parity` or `flight-query-semantics-oracle`.

**Source will not be patched to turn this green.** That would smuggle an uncertified change in under a test-harness diff and mask a real main-red.

Auto-merge is **not** armed and will not be until this is resolved.

pmcfadin:
## Correction: the gate FAIL was **infrastructure, not code** — no waiver is warranted, and none is being claimed

My previous comment proposed proving the five failures were pre-existing main-reds. That framing was wrong, and the evidence says something simpler. Correcting it rather than letting it stand.

### Four of the five components died on `No space left on device (os error 28)`

Not test failures at all:

- **`write-tests`** — reported **`test result: ok. 3465 passed; 0 failed`**, and only the *subsequent* compile hit ENOSPC (`rustc-LLVM ERROR: IO failure on output stream: No space left on device`).
- **`cli-tests`** — `error: failed to create directory .../target/debug/.fingerprint/toml_datetime-...` → `No space left on device`.
- **`compaction-byte-parity`** — same, on `.fingerprint/cqlite-core-...`.
- **`flight-query-semantics-oracle`** — `failed to write .../rmeta*/full.rmeta: No space left on device`, cascading into `could not compile arrow-array / tokio / clap_builder / percent-encoding`.

This is the **documented #1848 failure mode**, verbatim from `docs/development/gate-ops.md`:

> Each active worktree owns its own ~25–30GB `target/` dir. Several concurrent worktrees can exhaust the disk mid-gate (a confusing hard failure).

Measured at the time: 5 active lanes on this box, `/data/lanes` at **204G** (`lane-1714/target` alone 74G, `lane-1695` 47G, `lane-1451` 40G, `lane-1712` 32G; **this lane's is the smallest at 11G**). So the pressure was not this lane's, and the peer worktrees were all live — not mine to prune.

**A disk-full gate produces no verdict about the code.** The correct response is to fix the environment and re-run, *not* to waive. Had I waived these against #2596/#2188 as "known flakes" — which the component names invite — I would have recorded a false certification.

### The fifth, `tooling-tests`, is a genuinely filed flake: #3367

`test_roborev_review_guard.sh` → **`passed: 740  failed: 1`**, the one failure being:

```
FAIL - structural: delivery-mode classification is back in the flow scripts — mixed-delivery.
```

That is **#3367** ("the classifier-GONE structural assert is order-dependent AND masks a deterministic true-positive on the wrapper's own doctrine prose"), which records the *identical* tally and message, first seen on lane `issue-1714-...` whose diff likewise touched no `scripts/`. Root cause per that issue: `WRAPPER` is a mutable global reassigned to a scratch copy, so the assert is order/state dependent; standalone at the same commit it passes **741/0** four consecutive times.

Confirmed not mine two ways: `git diff --name-only origin/main...HEAD -- scripts/` is **empty**, and the tree was clean with no untracked files. Independently, `origin/main:scripts/flow/roborev-review.sh:339` really does contain the string `mixed-delivery` — inside **prose** listing what was deleted — so the guard is matching the wrapper's own documentation. A peer lane is bisecting this same assert right now.

### What I did about it

- Killed my own main-baseline reproduction run: once the cause was ENOSPC plus a filed flake, that run added nothing and was **competing for the very disk and gate slot that caused the failure**.
- **Two full gates were running concurrently in this worktree**, both writing the same `full-1458-run3.txt`: one I launched, and one a `flow-closer` subagent relaunched at 19:23 despite a stand-down instruction. Same-path + same-`target/` is the #2874 clobber hazard, and my sentinel had already overwritten its pinned path, so its ~35 minutes were unrecoverable. Both were killed, the closer was stopped, and **exactly one** gate now runs on a nonce-suffixed summary path that cannot collide.
- Re-running the gate of record on the unchanged certified head `9db4948b7`. **Not** a second bite at a code verdict: the prior run yielded an environmental failure, and re-running after fixing the environment is the same reasoning by which a gate that *died* without writing a verdict gets relaunched.

No source was touched to make anything green, and auto-merge remains unarmed.

pmcfadin:
## Gate of record — `RESULT: PASS` (attempt 6 of 6)

Full `AGENT-GATE SUMMARY` for the merged head `9db4948b7`. **30 components, all PASS, zero FAIL, zero SKIP.**

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.6J4mAv
commit: 9db4948 branch: issue-1458-dataset-guards-directory-only dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-1458/test-data/schemas (checkout-relative)
tree-start: 9db4948b7998 dirty: no digest: dfc1d219ed56
tree-end: 9db4948b7998 dirty: no digest: dfc1d219ed56
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (8s)
clippy:            PASS (4s)
roborev-lints:     PASS (42s)
core-tests:        PASS (40s)
tombstones-scan:   PASS (48s)
scan-offload-guard: PASS (197s)
work-counters-guard: PASS (101s)
byte-budget-guard: PASS (12s)
arrow-parity-guard: PASS (95s)
memory-budget:     PASS (12s)
integration-tests: PASS (85s)
format-compat:     PASS (72s)
write-tests:       PASS (238s)
cli-tests:         PASS (312s)
compaction-byte-parity: PASS (10s)
flight-query-semantics-oracle: PASS (145s)
python-bindings:   PASS (63s)
node-bindings:     PASS (131s)
delivery-telemetry: PASS (0s)
oom-audit:         PASS (11s)
parity-report:     PASS (1s)
operator-metrics-doc: PASS (99s)
kit-dashboard-drift: PASS (3s)
tooling-tests:     PASS (1015s)
minimal-build:     PASS (139s)
smoke:             PASS (4s)
logs: /tmp/agent-gate.6J4mAv
summary-file: /tmp/cqlite-gates/1458/full-r6-220731-553255.txt
RESULT: PASS
```

### Verification performed before arming the merge

| check | result |
|---|---|
| `run-id` matches the run I launched | `/tmp/agent-gate.6J4mAv` ✔ |
| `RESULT:` | **PASS** ✔ |
| `tree-integrity:` | **PASS** (`tree-start` == `tree-end` == `9db4948b7998`, digest identical) ✔ |
| `commit:` | `9db4948` `dirty: no` ✔ |
| positive `schemas:` line present | `6/6 canonical .cql readable` ✔ |
| `missing-fixtures:` absent (not `OPT-OUT`) | absent; `datasets: 144 Data.db` ✔ |
| `python-bindings` / `node-bindings` affirmatively PASS | `PASS (63s)` / `PASS (131s)` ✔ |
| `premerge-assert.sh 3371 9db4948b7…` | `PREMERGE: OK` ✔ |
| `HOLD:` re-read on PR + issue | none ✔ |
| roborev | clean at this exact head (job 10, `RESULT: PASS`, findings NONE, `2bde26a7c..9db4948b7`) ✔ |

`python-bindings` and `node-bindings` are the two components that carry this diff, so they were
checked **affirmatively** rather than by absence of a bad signal.

### Why six attempts, and why none of the first five was a code signal

None of the five deaths produced a verdict; all are infrastructure, and all are now filed:

| # | run-id | outcome | class |
|---|---|---|---|
| 1 | `WDuOwZ` | verdictless | subagent owned the gate across a turn boundary (#3344 shape) |
| 2 | `seKiSr` | 5 components red | ENOSPC — **VOID**, disk at 96% (#3256) |
| 3 | `60SxVG` | both runs killed | closer/lead collided on one summary path (#2874 clobber, #3246) |
| 4 | `IEXDIq` | verdictless | global OOM 21:50:18 amplified by `OOMPolicy=stop` (#2923) |
| 5 | `TOp2ZI` | verdictless | unattributed — recorded as unknown, not guessed |
| 6 | `6J4mAv` | **PASS** | hardened: `systemd-run --user --scope -p OOMPolicy=continue`, lead-owned, nonced durable summary path |

Notably **attempt 6 survived two OOM kills inside its own cgroup** (22:37:29 and 22:39:33, ~20 GB
`python3` each) because `OOMPolicy=continue` stopped systemd tearing down the scope. The trigger is
a gate self-test fixture that allocates ~20 GB — attributed and written up on
[#2923](https://github.com/pmcfadin/cqlite/issues/2923#issuecomment-5446106890).

The five components that were red in the ENOSPC-voided attempt 2 — `write-tests`, `cli-tests`,
`compaction-byte-parity`, `flight-query-semantics-oracle`, `tooling-tests` — **all PASS here**, on a
box with 147 GB free and an uncontended slot. That retires the ENOSPC attribution as correct.

Issue #3366, opened 2026-08-27
Derive mod-reachability from rustc's own dep-info instead of hand-parsing Rust (retire the second lexer)


pmcfadin:
Follow-up from #1714 (AK3), which landed the first mod-reachability guard for `cqlite-core/src`. That guard works and has already caught three real orphans (#1715, #3364, #3365) — one of which a careful hand-census had cleared. This issue is about replacing **how it decides**, not whether it should exist.

## The problem, stated as a class

#1714's guard answers "is this `.rs` file reachable from `lib.rs`?" by **lexing and parsing Rust source itself** — comments, string literals, every literal prefix, macro token trees, identifiers. That is **a second implementation of rustc's lexer**, and CLAUDE.md's own #3283 lesson names the consequence exactly:

> a port is a second implementation, and a second implementation's correctness is only knowable by differential testing against the original — the oracle re-derived Go's trim rules in bash and was tested against a *model* of Go, not against Go.

The review history on #1714 is that lesson reproducing itself. Three roborev rounds, seven findings, and **every false-PASS finding was a divergence between our lexer and rustc's**:

| round | divergence | class |
|---|---|---|
| 1 | `mod` inside a `macro_rules!` body / `stringify!(…)` | false PASS |
| 2 | `cr"…"` / `cr#"…"#` raw C-strings unrecognized | false PASS |
| 2 | raw identifiers (`r#make`) not consumed atomically | false PASS |
| 3 | identifiers are ASCII-only; Rust permits Unicode XID | false PASS |

Each was fixed correctly, and the fixes were structural rather than special-cased (a table-driven prefix lexer, plus a fail-closed refusal on any unrecognized prefix and on any non-ASCII identifier). **The guard is safe** — every unmodelled construct is now a loud `Err`, never a silent pass. But safety here is bought with **refusals**, and the refusal set can only grow as Rust's grammar does. Rust adds literal forms (`c"…"` arrived in 1.77); each one is a future refusal or, if we get it wrong, a future false PASS.

## The proposal: ask rustc, do not re-derive rustc

Replace source parsing with a **compiler-sourced oracle**. `cargo` already emits, per compilation unit, a dep-info file listing **every source file rustc actually read**:

```
target/debug/deps/cqlite_core-*.d
```

Orphans are then a set difference, with no lexer at all:

```
{ *.rs on disk under src/ }  minus  { files rustc read }
```

[...]

Issue #1697, opened 2026-07-02
AH4 — Config source of truth: one public Config, delete the decorative mirrors (DECIDED)


pmcfadin:
Part of epic #1685 (config honesty). Source: `docs/reports/platform-observability-audit-2026-07-01.md` finding AH4 (audit block 6). **DECIDED (owner, capstone ledger #12, 2026-07-01): ONE public `Config`; delete the decorative mirrors.** Sequence AFTER N1 #1619 (compaction wiring) and AH3 (knob purge) — they do most of the work; this issue is the consolidation + the standing single-source guarantee.

## Problem
Five config silos with contradicting duplicates:
1. core `Config` facade (`cqlite-core/src/config.rs`) — the documented public one, largely decorative;
2. `WriteEngineConfig` (private, actually runs the write path) — including a SECOND memtable threshold that shadows the public one;
3. `SizeTieredCompactionStrategy`'s own knobs (`merge_policy.rs:10-13`) — the compaction values that actually run (vs the public dead `CompactionConfig`);
4. the reader's disk-access + env-var family (honestly wired);
5. the CLI config (`ConnectionConfig`, `pager`, `performance.*` mappings into dead core fields).
Two memtable thresholds, two compaction configs — **only the private ones run**.

## Fix (effort: M/L)
1. After N1/#1619 + AH3 land, thread every surviving LIVE private knob through the ONE public `Config`: `WriteEngineConfig` fields become derived from `Config` (constructor takes `&Config`), STCS thresholds already threaded by N1, CLI `performance.*` mappings either map to live core fields or are deleted.
2. Delete the private mirror FIELDS that duplicate public ones (the private structs may remain as internal plumbing, but their values must originate from the public `Config` — one source).
3. Standing guard: per-knob single-source test — setting the public knob changes engine behavior (extends AH3's test family to the newly threaded knobs, esp. the memtable threshold: set public value → flush happens at that size).
4. Document the final config surface in one place (config.rs module docs): every field, its consumer, its default.

### Do NOT
- Do not introduce a sixth layer (builder/facade indirection) — this is consolidation, not abstraction.
- Do not change any DEFAULT behavior silently — if a public default differed from the private one that actually ran, keep the RUNNING value as the new public default and note it in the changelog.

## Acceptance criteria
- [ ] One public source: every live knob originates in `Config`; PR includes the silo→field mapping table.
- [ ] The two-memtable-thresholds and two-compaction-configs duplications are gone.
[...]

Issue #1451, opened 2026-07-02
Shared authoritative error table: Python maps by variant, Node by category — same core error, different identity


pmcfadin:
Parent: #1434

## Why this matters
The same core error surfaces with a **different identity** in each binding. Python maps errors by the core `Error` **variant**; Node derives its code from the core `category()`. So "catch `ParseError` for bad CQL" is true in Python but false in Node (Node reports CqlParse as `QUERY`), and `Timeout`/`Memory` — which Python surfaces as `TimeoutError`/`MemoryError` — both collapse to `IO` in Node. Users writing error-handling against one binding cannot rely on it in the other. This replaces the two hand-maintained mappings with one authoritative table both consume.

## Current behavior (verified 2026-07-01)
Python maps by **variant** — `bindings/python/src/error.rs:65-108` (`to_py_err`):
```rust
pub fn to_py_err(err: cqlite_core::Error) -> PyErr {
    match err {
        cqlite_core::Error::Io(_) => PyIOError::new_err(message),
        cqlite_core::Error::Schema(_) | cqlite_core::Error::Table(_) => SchemaError::new_err(message),
        cqlite_core::Error::QueryExecution(_) | cqlite_core::Error::UnsupportedQuery(_) => QueryError::new_err(message),
        cqlite_core::Error::CqlParse(_) => ParseError::new_err(message),          // CqlParse -> ParseError
        cqlite_core::Error::Configuration(_) | cqlite_core::Error::InvalidInput(_) => PyValueError::new_err(message),  // InvalidInput -> ValueError
        cqlite_core::Error::Timeout(_) => PyTimeoutError::new_err(message),        // Timeout -> TimeoutError
        cqlite_core::Error::Memory(_) => PyMemoryError::new_err(message),          // Memory -> MemoryError
        cqlite_core::Error::InvalidState(_) => PyRuntimeError::new_err(message),
        cqlite_core::Error::WriteDirLocked { .. } => CqliteError::new_err(message),
        _ => CqliteError::new_err(message),                                        // Corruption -> base
    }
}
```
Node derives codes from **category()** — `bindings/node/src/error.rs:70-87` (`category_to_code`) and `:103-125` (`extract_metadata`):
```rust
pub fn category_to_code(category: ErrorCategory) -> &'static str {
    match category {
        ErrorCategory::System => "IO",
        ErrorCategory::Data => "PARSE",
        ErrorCategory::Schema => "SCHEMA",
[...]

PR #3376, opened 2026-08-27
fix(#1712): un-export `pub mod benchmarks` at the declaration site + a standing pub-surface snapshot guard

pmcfadin:
## What this ships

`cqlite-core/src/lib.rs` declared `pub mod benchmarks;` **unconditionally** while
`cqlite-core/src/benchmarks/mod.rs` gated itself with an inner `#![cfg(feature = "benchmarks")]`.
The crate root said the module shipped; the compiled crate did not have it. This PR hoists the gate to
the declaration site so the crate root tells the truth, and adds a **standing guard** so the shape
cannot come back.

The guard answers one question, **from source**: for every crate-root `pub mod NAME;` that reads as an
unconditional public export, does `NAME`'s own module file gate itself? Anything it cannot classify
confidently **REFUSES** (fails loudly, naming the file) rather than passing.

> **Scope note.** This PR was originally scoped to also ship a rustdoc-derived public-API **snapshot**.
> That half is **deliberately gone** — see *"Why the snapshot was dropped"* below. It was a delivery-lead
> ruling, not a shortcut, and the cost it accepts is stated explicitly rather than buried.

## Evidence

| Check | Result |
|---|---|
| `--lite` (lead-owned, protected scope) | **PASS**, `tree-integrity: PASS`, zero SKIPs |
| `scripts/tests/test_pub_surface_guard.sh` | **PASS — 38 cases** (8 green controls, 28 reds), **zero SKIPs**, zero FAILs |
| Guard **RED on `origin/main`** | **exit 1 in 91 ms** — verbatim below; re-verified with the SHIPPED guard |
| Guard **GREEN on this branch** | **exit 0 in 0s** |
| Full gate of record | **PASSED at `299be60ea`** — 31/31 components, zero SKIPs, `tree-integrity: PASS`. Re-running at the current head (see below). |

**RED on `origin/main`** — new guard copied into a scratch worktree at `origin/main`, **md5 confirmed
identical** to this branch's copy, so it is provably unmodified:

```
GUARD_EXIT=1   ELAPSED=0s

❌ pub-surface: crate-root declaration is INCONSISTENT with the module's own file

    `pub mod benchmarks` at cqlite-core/src/lib.rs:25 reads as an unconditional public
    export, but its module file GATES ITSELF:

        cqlite-core/src/benchmarks/mod.rs:1:  #![cfg(feature = "benchmarks")]
    ...
    Remedy — hoist the gate to the declaration site so the crate root tells the truth:
```

**GREEN on this branch** — the affirmative measurement line, which cannot be emitted without having
enumerated the declarations and read the files:

```
pub-surface: 26 crate-root declarations scanned in cqlite-core/src/lib.rs (20 pub mod, of which 14
unconditional); 14 module-file prologues read from source; 0 inconsistent
```

`0s` vs **~225 s** for the rustdoc version, measured — see the cost table below.

## Is this a semver change?

**No public item is added or removed.** `benchmarks` was *already* absent from the default-feature
surface (the inner `#![cfg]` removed it); the crate root merely advertised it. So the compiled public
API is unchanged and the crate root now agrees with it. The PR moves a `#[cfg]`, it does not move the
surface.

## Which half is inherited and which is new

Requested explicitly, because **the new half is where the risk is.**

- **Inherited** (declaration-site attribute reading, `attrs_verdict`; structural, unchanged): pinned by
  cases **2, 8, 9, 10, 11, 15, 16, 20, 23, 24**, plus case 1 (real tree), case 5 (usage), case 19
  (worktree cleanup / kill-safety).

  > **Correction, because "inherited coverage" meant less than that list implies.** All ten are present
  > and functional **now**, but **five of them were not functional at the head this work inherited**:
  > they keyed on `/^#\[cfg\(feature = "benchmarks"\)\]$/` while the real gate line carried a trailing
  > `// #1712: …` comment, so the anchored pattern matched nothing. Cases 8/9/15/16 aborted loudly at
  > their own setup asserts — but **case 2 no-opped SILENTLY and certified the very shape it exists to
  > red.** Inherited coverage was **5 functional of 10**. All ten now use layout-independent helpers
  > whose post-condition is *affirmative* ("after stripping, `lib.rs` does not MENTION the `benchmarks`
  > feature at all") rather than pattern-shaped, and they key on the **declaration**
  > (`/^pub mod benchmarks;/`), never on the gate's position relative to it — the 800-line reclaim was
  > the *third* cosmetic `lib.rs` change to break those fixtures. A test that silently stops testing is
  > this issue's own defect class wearing a different hat, so it is reported rather than quietly fixed.
- **New** (the source oracle and its refusals): cases **28–38**.

| Case | Refusal / property |
|---|---|
| 28 | code after a closing `*/` on one line ⇒ REFUSE (+ same tree, decl on its own line, stays GREEN) |
| 29 | module file resolves to **neither** legal path ⇒ REFUSE, naming both |
| 30 | resolves to **both** legal paths ⇒ REFUSE rather than choose |
| 31 | not a readable regular file ⇒ REFUSE |
| 32 | block comment in a module prologue ⇒ REFUSE, naming file and line |
| 33 | inner attribute merely **mentioning** `cfg` ⇒ REFUSE, text distinct from the defect verdict |
| 34 | content after an inner attribute on one line ⇒ REFUSE (a same-line gate cannot hide) |
| 35 | inner attribute that never closes its `[` ⇒ REFUSE |
| 36 | **positive control for 29–38** — an ordinary prologue certifies, and the module really was examined |
| 37 | a `]` inside a **line comment that ends the line** ⇒ REFUSE *(roborev r8 F1)* |
| 38 | **BOM**-prefixed inner `#![cfg]` ⇒ caught as the **named defect** *(roborev r8 F2)*; also pins `LC_ALL=C` |

Case 28 is worth calling out: it is roborev **r7 finding 2**, which had been grouped with the deleted
scanner but actually lived in the surface being **kept**. Fixed and pinned rather than inherited.

## roborev r8 found two false PASSes in the new reader. Both fixed.

The guard's entire history is *checks that passed without measuring anything*, so the reviewer was
pointed at that direction specifically. It delivered:

- **F1** — bracket balance counted `[`/`]` inside comments. Falsifying input `#![allow(dead_code, // ]`:
  the `]` is inside a comment **and ends the line**, so the window closes early, nothing follows it (so
  case 34's same-line refusal cannot fire), and the scan resumes mid-attribute, missing a real
  `#![cfg(...)]` below. **Measured on the pre-fix reader: `CLEAN` — certified.**
- **F2** — a leading UTF-8 BOM was not stripped, so `#![` did not match and the prologue read `CLEAN`,
  while rustc **ignores a BOM and applies the gate**. **Measured pre-fix: `CLEAN`.**

Three things about the fix:

1. **Strings are SKIPPED, not refused** (escapes honoured). Case 36's positive-control fixture contains
   `#![doc = "..."]`, so a blanket quote-refusal would have **red correct code** — and as case 36's own
   failure text says, a refusal that reds correct code is the one that gets waived, which would make
   every other refusal worthless. Comments, raw strings and bare `'` refuse.
2. **`#![cfg` is now decided from its NAME, before any bracket balancing.** The defect's canonical
   spelling `#![cfg(feature = "x")]` *contains string quotes*, so deciding it after the scan would let
   one of the new refusals downgrade the precise named diagnostic — which carries the hoist remedy, i.e.
   this issue's acceptance criterion — into a generic "cannot classify". A name needs no counting, so no
   bracket-counting defect can hide the one shape this guard exists to catch.
3. **The reader runs under `LC_ALL=C`, and this was a second bug behind the first.** The BOM fix did
   nothing on the first attempt: under a UTF-8 locale awk's `sprintf("%c", 239)` yields the *character*
   U+00EF (two bytes), not the byte `0xEF`. Demonstrated both ways on the same awk program: UTF-8 ⇒
   `CLEAN` (bug alive), `LC_ALL=C` ⇒ `GATED`. Case 38 pins it, so anyone dropping the prefix reds.

**A swallow is unreachable by construction, not merely untested.** For the string-skip to produce a
false PASS the swallowed text must *be* a `#![cfg...]`, which necessarily contains the token `cfg`. It
can land in exactly two places: at or before the closing bracket, where it is inside `attr` and
`mentions_cfg` — deliberately **not** string-aware — refuses; or after it, where `rest != ""` refuses.
There is no third position. Ten adversarial inputs (multi-line string wrapping a gate, unterminated
string, escaped backslash before the closing quote, backslash at end of line, raw string, CRLF, `#![cfg]`
with no parens, nested brackets) all refuse or gate correctly.

## roborev r9 found four more. All fixed, plus one I found myself.

**The polarity that ties them together, and that I had not stated before:** `GATED` and `HIDDEN` are
the **exempting** verdicts — only an `OPEN` declaration gets its module file read. So the dangerous
direction is not *wrongly reporting a defect*, it is **wrongly reporting gatedness**. Anything that
flips `OPEN`→`GATED`, stops the reader early, or makes resolution read the *wrong file* skips the
inspection entirely.

- **F1 — raw strings in a declaration attribute.** `strip_strings` models only ordinary quoted
  strings, so `doc = r##"", cfg(any()), ""##` leaks a comma and a `cfg(...)` into the meta-item parse
  and can flip `OPEN`→`GATED`. Refused rather than modelled (hash-count delimiters are a second
  lexer). Anchored on a non-identifier boundary, because an ordinary string ending in `r` —
  `doc = "for"` — literally contains the characters `r"`; found by *testing* the regex, not reading it.
- **F2 — the `#[path]` decoy.** With a clean `NAME.rs` beside `#[path = "actual.rs"] pub mod NAME;`,
  resolution reads the **decoy**, finds it clean, and certifies while `actual.rs` carries the gate.
  Demonstrated end to end before and after. Worth noting: this had been *analysed and cleared* on the
  grounds that `#[path]` lands in the `found == 0` refusal — true only when no standard-path file
  exists, which is exactly the hole. A correct analysis of the wrong case.
- **F3 — a first-line shebang.** It read as the first item, ending the prologue at line 1 and
  certifying the file. **Verified on rustc 1.98.0** that `#!/usr/bin/env rust` + `#![cfg(feature =
  "nope")]` compiles *and the gate applies*. Fixed with an exact rule — exactly one shebang is
  possible and only on line 1 — applied after the BOM strip.
- **F4 — the gate's measurement regex pinned the wording but not the arithmetic.** It accepted
  `14 unconditional; 1 module-file prologues read` (13 declarations silently unexamined) and
  `0 pub mod, of which 5 unconditional`. The four counts are now extracted and must satisfy
  `declarations >= pub mod >= unconditional > 0` and `prologues == unconditional`. **Why this matters
  beyond arithmetic:** this component exists to be *independent* of the guard. The guard already
  asserts `prologues == unconditional`, so a line failing it proves the guard is not what produced the
  line — a stub, a truncation, a stale build. Trusting the relationship would collapse the independence.
- **And one not in the report.** An unbalanced `#[` returned `GATED`, with the in-code comment
  *"unbalanced: cannot read it, so do not exempt on OPEN"*. The polarity is backwards — `GATED` **is**
  the exemption here — so an unreadable attribute was silently skipping the inspection it could not
  rule out. Now a named refusal. A comment asserting the safe direction while implementing the unsafe
  one, visible only once the polarity was written down.

Pinned as cases **40, 41, 42** and **26(d)**. Cases **38 and 42 deliberately assert the NAMED DEFECT
rather than a refusal**, because a BOM and a shebang are each *exactly* modellable — refusing there
would trade a precise, actionable verdict for an unactionable one.

**Two premises the design rests on are now verified with rustc rather than inherited from comments**
(a CQLite comment is not authority for a rustc fact):

```
pub fn f() {}        + #![allow(...)]  => error: inner attribute not permitted in this context
#[allow(dead_code)]  + #![allow(...)]  => error: not permitted following an outer attribute
use std::fmt;        + #![allow(...)]  => error: not permitted in this context
//! docs             + #![allow(...)]  => ACCEPTED   (control)
```

So the prologue provably contains **every** inner attribute a file can legally have, and prologue
termination is sound by construction. The AK1 defect itself also reproduces from first principles:
`#![cfg(feature = "nope")]` over a `pub fn present()` compiles cleanly and emits **no** `present`
symbol.

### Twenty findings, zero false FAILs

Across r4–r10 this lane produced **twenty findings, and every one was a false PASS — not one was a
false FAIL.** That asymmetry is the most useful fact here: a guard's defects concentrate entirely in
the direction of *wrongly certifying*, because that is the direction nothing else checks. Hence every
reviewer was briefed to attack that direction, the new refusals err toward failing, and every red
control asserts the **diagnostic** rather than merely a non-zero exit.

## Rounds r10–r15: what the reviews found, and one thing I removed

**Thirty-five findings across r4–r14. Thirty-one were false PASSes; four were false FAILs** — and
the false FAILs only began once *I* started adding refusals, which is the predictable cost of a
fail-closed strategy and why **every refusal now ships with a GREEN control**, not just a red one.

Highlights, and the two that generalize:

- **`sort -u` in the cross-check discarded multiplicity** (r10). `#[cfg(any())] pub mod x; pub mod x;`
  is valid Rust — the first is configured out, so there is no duplicate definition, and the second is
  unconditional. The structured scan reads one statement per line, the simple scan saw two, and dedup
  made them *agree*. Measured: the pre-fix guard **exits 0** on that tree while the module gates
  itself. Fixed by comparing multisets, which needed no new refusal.
- **`cfg_attr` propagated a CONDITIONAL exemption as unconditional** (r11). `#[cfg_attr(any(),
  doc(hidden))]` is neither hidden nor gated, yet was classified `HIDDEN` — an *exempting* verdict — so
  the module file was never read. Now always `OPEN`: it never exempts. `OPEN` rather than a refusal,
  because it costs no false FAIL on the standard `#[cfg_attr(docsrs, doc(hidden))]` idiom.
- **`normalize()` copied string and RAW-string CONTENTS verbatim** (r13) into the text both derivations
  read. Derivation S scans unanchored while P is line-start anchored, so a one-line
  `const X: &str = r#"pub mod fake;"#;` was found by S, missed by P, and the cross-check called a
  disagreement — **the mandatory gate rejecting valid Rust.** Contents are now blanked, delimiters and
  columns preserved.
- **Fixing that unlocked a fix two rounds of patching had missed.** The macro sweep could not be
  broadened earlier for exactly one reason: `N[]` carried string contents, so `#[doc = "call foo!(x)"]`
  would false-fire. **The lesson: widen a pattern once over its whole syntactic family, after removing
  whatever forced it narrow — not per reported instance.**

### The item-macro refusal: built over three rounds, then removed

Worth recording as a decision rather than a diff. A macro can emit `pub mod NAME { #![cfg(...)] }`,
with the literal declaration living in the macro *definition* where neither derivation reaches it. I
added a refusal for it in r11 and widened it in r12 and r13. r14 then hit it from **both** directions
at once:

- **false FAIL** — `const X: bool = cfg!(feature = "y");` at the crate root is ordinary Rust and was
  rejected. Separating an *item* macro from an *expression* macro requires knowing item boundaries,
  i.e. parsing items.
- **and the hole was never closed anyway** — procedural attribute and derive macros can emit a
  crate-root `pub mod` with **no `!` anywhere**, so no function-like pattern reaches the class.

A partial refusal is defensible. A refusal with false FAILs is defensible. **One that reds correct
code *and* leaves the hole open has no case for shipping**, so it is gone, replaced by an explicit
stated residual in the guard naming both mechanisms, the measurement that makes it a hardening gap
rather than a live hole (**zero crate-root item macros in `cqlite-core/src/lib.rs`**), and #3366 as
the principled route — dep-info sees *expanded* syntax and has no macro problem to model.

### Three false FAILs I found by probing rather than by review

Recorded because they were all in the *mandatory* gate, and a refusal that reds correct code is the
one agents learn to waive — which would devalue every other refusal here.

1. **Refusal X fired on any residue after a closing delimiter** — and `"#;` is how every multi-line
   raw string ends, so any crate root containing one was refused outright. The in-code comment shows
   the original author knew and called it safe over-approximation; it was not free. Narrowed to require
   a **letter** in the residue: a declaration cannot exist without letters, so `"#;` cannot hide one,
   while `*/ pub mod x;` still refuses.
2. **A one-line `pub fn f() { assert!(true); }`** was refused as an item macro, because `BRACE_MIN` is
   the line's *minimum* depth and the macro sat inside the body. (Moot now the refusal is gone, but it
   is why depth-at-position exists in the S scan.)
3. **A one-line inline module `mod outer { pub mod inner; }`** — ordinary Rust in which `inner` is not
   a crate-root declaration — was rejected, because S collected it unanchored while P correctly ignored
   it. Case 24 had pinned this property only for the *indented multi-line* form. S is now depth-aware
   within a line, which does not cost the cross-check its value: S keeps its own collection *rule* (an
   unanchored scan, no attribute parsing), and that is where its independence from P lives.

**Deliberately NOT changed:** `pub fn f() {} pub mod x;` still refuses via scan disagreement. That one
is honest rather than defective — P genuinely did not see the declaration, so the guard cannot assert
over it, and the remedy is one line break. The rule applied throughout: **fix where S is WRONG; keep
the refusal where P is INCOMPLETE.**

## The gate of record, and why it ran twice

**A full `agent-gate.sh` PASSED at `299be60ea`** — all 31 components, **zero SKIPs**,
`tree-integrity: PASS`, `tree-start == tree-end`. Notable: `clippy` 297 s, `core-tests` 576 s,
`python-bindings` 482 s, `node-bindings` 371 s, `tooling-tests` 730 s, and **`pub-surface` 1 s**
(the descoped guard is source-only). Fixture contract satisfied positively:
`datasets: 144 Data.db files`, `schemas: 6/6 canonical .cql readable`.

**The green was interrogated, not banked** (per the #3406 proposal):

1. **SKIP census** — zero occurrences anywhere in the summary.
2. **The `0 s` components were checked against their logs**, because a zero duration is evidence of
   neither vacuity nor work: `query-semantics-oracle` ran 6 tests, `compaction-byte-parity` 4, and
   `delivery-telemetry` actually wrote ledger entries.
3. **The `pub-surface` measurement line is arithmetically coherent** — 26 ≥ 20 ≥ 14 > 0 with
   `prologues == unconditional`. So the component *re-derived a relationship* rather than matching a
   wording, which is the r9 F4 fix proving itself on the run that counts.

**It is re-running at the current head**, because r18 then found two more false PASSes and those are
`src` changes. A gate that predates a source change is not a certification, so the earlier summary is
retained as evidence the pipeline is green end-to-end and explicitly **not** as this head's verdict.

## Rounds r16–r18: three more, and one lesson about fixing in one place

- **Refusal V, three corrections.** Built in r15 (a bare `pub` line), too **narrow** in r16
  (`#[allow(dead_code)] pub` slipped past), too **broad** in r17 (a macro token tree
  `stringify!(\n pub \n)` is valid Rust and was refused), **gap-blind** in r18 (a multi-line comment
  between the halves defeated the lookahead). It now requires *corroboration* — the next non-blank
  line must begin a module declaration — and skips whitespace-only lines.
  **Why it stayed while the macro refusal was removed:** each correction made the rule *simpler*
  (narrow match → corroboration → "skip blanks" replacing a state test), whereas the macro refusal's
  fixes kept adding cases. **Direction of complexity is the signal, not round count.**
- **One fix that needed two homes** (r18). r15 canonicalised whitespace between `#`, `!` and `[` in
  the *prologue reader*; the *crate-root scanner* kept the contiguous test. So
  `# [path = "actual.rs"]` was discarded, both scans agreed the module was open, and resolution
  certified a clean **decoy** while the real self-gated file went unread. The rule is now written at
  the site: **when a lexical assumption turns out to be wrong, grep for every scanner that shares it
  before calling the fix done.**
- **A worktree leak in the self-test** (r18, Low). Case 19's two decoys were created outside the
  scratch root and never registered, so a signal before their explicit removal left both the directory
  *and* the git registration behind — and this repository's worktree registry is **shared across every
  `/data/lanes/lane-*` checkout**, so a leak is visible to other lanes. Fixed with an exit/signal-only
  extra registry, deliberately *not* `PS_WORKTREES`, since case 19 removes one decoy mid-run as part of
  what it measures and asserts the *peer* decoy survives a fresh init.
  **Verified twice** — once by the case, and once accidentally when a stray `SIGTERM` killed a suite
  run mid-flight and left zero leaked worktrees.

### One open design question, escalated rather than decided

roborev notes that `#[doc(hidden)] pub mod x;` is skipped without inspecting its module file, and it
is **technically right**: `doc(hidden)` affects documentation only, so the AK1 inconsistency is just as
true. But the current behaviour is this guard's **documented contract** in three places — CLAUDE.md's
`non-#[doc(hidden)]` qualifier, the guard's verdict definition, and a failure diagnostic that *offers*
`doc(hidden)` as a remedy — and case 39 depends on it mechanically. It is an advertised opt-out, not an
oversight. Removing it would make hoisting the `cfg` the *only* way to satisfy the guard, which is
stricter than the ruling asked for. **Left as-is pending an owner call**, because it changes what
#1712 promises.

## Why the snapshot was dropped — and what that costs

The rustdoc half was deleted on a delivery-lead ruling after **five** separate false-PASS defects across
review rounds 4–7, every one of them in a hand-rolled awk scanner that had to find declarations in
arbitrary source and therefore could never abstain. The replacement asks a bounded question about one
named file and **can refuse**. Abstention is the difference, not the fact that both read Rust text.

What that buys, measured:

| | rustdoc snapshot | source oracle |
|---|---|---|
| Catches the AK1 defect | yes | yes |
| Catches public-API **drift** | yes | **no** |
| Runtime | ~225 s/full gate, ~25 rustdoc builds/suite run, ~8 G/min disk burst | **0 s** |
| `cfg(doc)` divergence fence (`§1b`) | required | **gone** |
| Committed artifact | 5,482-line snapshot | none |

**The cost is real and I am not going to soften it: public-API drift detection is gone.** Drift is not
hypothetical on this repo — it happened **twice while this PR was open**. `main` added
`pub mod ffi_error_contract;` (#1451), which the old snapshot caught correctly and unprompted; and
`main`'s #1697 then changed `cqlite-core`'s public `Config` across 17 files, which the descoped guard
**does not notice, by design**. **#3366** (derive reachability from rustc's own dep-info) is the
principled route to that capability — no lexer to diverge — and is the right place for it.

Deleting the snapshot also made **`REQ-1712-B` moot**: the 5,482-line file was what forced roborev's
by-path diff delivery, and r8 reports `prompt-content: PASS (6/6)`. That escalated waiver no longer
needs a human authorization.

### Other stated residuals

- A module prologue opening with a **block comment** (`/*! docs */` or a `/* */` licence header) is
  **REFUSED**, not parsed. **Measured: zero occurrence** — all 14 unconditional crate-root modules open
  with `//!`. A future lane hitting it gets a diagnostic naming the file and line; the fix is one `//`.
- The guard reads **only** crate-root declarations and the prologues of the unconditional ones. It says
  nothing about items inside those modules.
- `cqlite-core/src/lib.rs` is at **exactly 800 lines**, the strict `file-size` limit, so any addition
  trips the ratchet. Note for the next lane to touch it: the `#1712` rationale is a trailing comment on
  the `pub mod benchmarks;` **statement** line, not the `#[cfg]` **attribute** line — rustfmt splits a
  trailing comment after an attribute onto its own line, which adds a line and reds the ratchet. That is
  compile-valid but not rustfmt-stable, and `cargo check` cannot tell you.

## Gate / review status

- `--lite`: **PASS** (`tree-integrity: PASS`, zero SKIPs), lead-owned launch in a protected scope.
- roborev **r8** (job 38): all mechanical keys PASS — `push-assert`, `census-check`, `code-free`,
  `job-record`, **`sha-assert`**, `review-completed`, **`prompt-content: PASS (6/6)`**, both vacuity
  tiers. Tokens `input=443747 cached=365056` (genuine range). Two findings, both fixed above; **r9
  pending on this head.**
- **`sha-assert` passes for the first time since r5.** The cause of the earlier failures was diagnosed
  and corrected in the open: roborev reviews `merge-base..HEAD` while the wrapper asserted against
  `origin/main`'s **tip**, so it failed *deterministically* whenever main was ahead — not a timing race,
  which is what I had originally (wrongly) reported. **#3392** carries the merge-base fix; the workaround
  used here is to keep the branch merged up to date.
- **Full gate of record: NOT YET RUN.** `flow-closer` runs it exactly once, immediately pre-merge.


pmcfadin:
## ⚖️ Waiver request relayed to the owner — `prompt-content` absence, job 58

Relayed by the coordination lead. The lane requested it and may not self-apply it; the owner has reserved waiver authorization to themselves, so I am not granting it either. **This is the fifth waiver pending.**

**Triple, verified by me from git rather than read off the lane's block:**

| field | value |
|---|---|
| `base` | `9927d4c4456833965eab47e4b14c37b87ce897cd` |
| `head` | `7279b33de65cc18bc64d561e67c88ca46db452f8` |
| `job` | `58` |

```
pr3376 head: 7279b33de65cc18bc64d561e67c88ca46db452f8   ← matches head
merge-base:  9927d4c4456833965eab47e4b14c37b87ce897cd   ← matches base
```

Both endpoints confirmed independently. Head frozen and unmoved.

**Evidence:** `1,073,958` input / `997,376` cached / `7,609` output — ~57× the vacuous baseline (~18.7k / 0 / ~55). `sha-assert`, `review-completed`, `vacuity-tier1`, `vacuity-tier2` all PASS. **Gate of record has PASSED twice.**

**Cause is size, and this case documents something new:** rounds r8–r18 all reported `prompt-content: PASS (6/6)`. The absence appeared only once the *accumulated guard-and-tests diff* reached **+3,607 lines across 8 files** — most of it the 39-case self-test and refusal comments **that twenty-plus review rounds demanded**. So the review process itself produced the condition that blocks the review from passing. Recorded on #3426, because that and `vacuity-tier2` are the same defect mirrored: a **size-derived proxy fails at both ends of the size range**.

**Three findings are held UNFIXED** under my ruling, preserved unapplied at `/tmp/cqlite-gates/1712/UNAPPLIED-r23-fixes.patch`. If you would rather they land than be deferred, say so and I will reverse — but that reopens the head and voids this triple.

## Why no fresher review, stated explicitly so it is not read as a gap

I instructed a final round *after* the freeze; the lane had already run r23 (job 58) **against** `7279b33de`. It declined to re-run, correctly: **a fresh run enqueues a new job id, and the waiver binds `job=`**, so r24 would make this triple stale the instant you granted it. My instruction and its own rationale pointed opposite ways and the lane stopped rather than silently picking one. If you want a round that post-dates the freeze, say so and expect `job=59+` with this one discarded.

**One note on form:** the lane's comment includes a marker-shaped line with an unsubstituted `reason=<yours>` placeholder. That is **inert by design** — a placeholder reason reads `MALFORMED`, and it is not the sole content of its comment — so it cannot authorize anything. Flagging it because a diagnostic that carries a live marker is #3312's second-worst instance, and the safe form is to print no part of the marker at all.

Queue: #3382 (job 12), #3408 (job 9), #3397 (job 55), #3398 (job 53), #3376 (job 58).

pmcfadin:
roborev-waive: prompt-content-absent base=9927d4c4456833965eab47e4b14c37b87ce897cd head=7279b33de65cc18bc64d561e67c88ca46db452f8 job=58 reason=owner grant 2026-08-28; #3312 snapshot-delivery absence after review-demanded diff growth to +3607 lines; tokens 1073958 input / 997376 cached / 7609 output vs vacuous baseline ~18700/0/55; gate of record passed twice; head frozen; three r23 findings stay deferred per the lead ruling

pmcfadin:
Waiver above granted by the owner. The three r23 findings stay deferred as ruled; do not reopen the head. Apply the waiver with a recheck of job 58, then proceed to merge.

pmcfadin:
## coord: ✅ **Waiver applied — `prompt-content: WAIVED`, `RESULT: PASS`.** Gating before I merge, and here is why that is not me stalling.

Worker on `ip-172-31-6-169`. Head still frozen at `7279b33de`. Cron deleted (unblocked).

### Step 1 done: the waiver applied cleanly

```
MODE: recheck (job 58 re-decided from its job record; NO review was enqueued)
recheck-of: 58        sha-assert: PASS        review-completed: PASS
prompt-content: WAIVED (6/6 absent — authorized by @pmcfadin for base=9927d4c4… head=7279b33de… job=58)
waiver: GRANTED (author=@pmcfadin …)
vacuity-tier1: PASS   vacuity-tier2: PASS
RESULT: PASS
```

**One precision I will not gloss, because it is the defect I reported earlier today.** That block reads `RESULT: PASS` while carrying `findings: PRESENT (3)`. It is legitimately clean **because you ruled those three deferred** — *not* because the recheck terminal gates on findings. It does not (`issuecomment-5450184755`). So the PASS is sound here on the strength of your ruling, and would have been unsound without it. Worth stating so nobody later reads this block as evidence that the recheck path is safe.

### Step 2: I am running the gate of record BEFORE merging, because the head is not certified

**Your note said the gate has passed twice, which is true — but neither pass covers this head.** The range since the last one contains two of my own commits, not just main's merge:

```
7279b33de  Merge origin/main
4fc8d540f  test(#1712 r22): pin the shared depth helper, including the sixth site
8947cbe1e  refactor(#1712 r22): compute crate-root depth ONCE — six sites had six answers   <-- CORE REFACTOR
9927d4c44  chore(#3364): telemetry (main)
52b67d4c2  fix(#3364): delete orphan header_fix_functions.rs (main)
```

`8947cbe1e` is not a cosmetic delta — it **replaced five independent depth computations with one shared `root_depth_at()` across six call sites** and deleted their private counters. A gate that predates a refactor of the guard's core logic is not a certification of it, and `premerge-assert` exists precisely to refuse a merge where the PR head is not the certified SHA.

`required` **is** SUCCESS at this head, so CI is green — but CI is not the gate of record, and the agent-gate runs components CI skips here (`pub-surface`, `oom-audit`, parity, bindings, smoke; this PR's run shows `Core tests`, `Integration tests`, `Write-support tests` and `CLI tests` all SKIPPED).

**So: gate launched at `7279b33de`, lead-owned, protected scope, `OOMPolicy=continue` verified, tree clean.** It is a from-scratch build — I reclaimed my 88 G `target` while frozen to unblock lane-3385 — so expect **60–90 minutes**. Nothing is committed or pushed; the head does not move.

**If you would rather I merge on the two existing passes plus green `required`, say so and I will** — that is a legitimate call given the delta is one refactor plus its tests, and you may weigh the queue differently. But I am not going to make that call silently on your behalf, because "the gate passed twice" and "the gate certifies this head" are different statements and only the first is true.

### Then, immediately and without further asking

`premerge-assert 3376 <certified-sha>` → `gh pr merge --auto --squash --delete-branch` → `flow-finalize` (archive, telemetry PR via a `telemetry-1712` worktree with `--ledger` explicit, worktree/branch/claim cleanup) → closing packet.

**Two things I will carry into that packet rather than let them disappear:**
- **C audit: ABSENT, labelled not implied.** Five subagents went idle without returning verdicts. AC1 (consumer audit) and AC3 (#1637 cross-comment) are self-verified from disk; the RED-on-main demonstration is verified three times. You remain free to demand a real audit.
- **AC2 still specifies the snapshot the descope deleted** (`issuecomment-5449854018`). Unresolved, untouched by me, and it will be named in the closing comment as an open scope item rather than quietly satisfied.

pmcfadin:
## coord: 🚀 **`--auto` armed. And a correction you need: the ROOT checkout does NOT have the #3392 fix — my worktree does.**

Worker on `ip-172-31-6-169`. Head unmoved at `7279b33de`.

### The instruction would have used the UNFIXED script — verified rather than assumed

You told me to run the recheck from the root checkout "which is on main and contains the fix", and to verify rather than take your word. I did, and it inverts:

```
root  /data/lanes/repo   HEAD = 2bde26a7c  — 20 commits BEHIND origin/main
      roborev-review.sh: merge-base references = 0        ← NO FIX
mine  /data/lanes/lane-1712                                
      roborev-review.sh: merge-base references = 20       ← HAS THE FIX
938d464fb IS in origin/main; the root checkout simply has not pulled it.
```

**The root checkout is on `main` but stale by 20 commits.** So "run it from root because root has the fix" produces the opposite of the intent. My worktree carries the fix because I merged `origin/main` at `9927d4c44`, which is downstream of it — which is *why* my in-worktree recheck at 14:56Z already passed `sha-assert` and applied the waiver.

**Fleet-relevant:** any lane told "run from root, it has the fix" gets the unfixed script until someone pulls root. Worth a line on #3392 or #3416.

### Waiver applied, and I read every key rather than inferring from RESULT (per your #3422 caution)

```
MODE: recheck (job 58 re-decided from its record; NO review enqueued)   recheck-of: 58
sha-assert: PASS          review-completed: PASS
prompt-content: WAIVED (authorized by @pmcfadin, base/head/job all matching)
waiver: GRANTED (author=@pmcfadin …)
vacuity-tier1: PASS       vacuity-tier2: PASS
findings: PRESENT (3)     ← deferred per your ruling, NOT absent
roborev-exit: SKIP (recheck: no reviewer ran)
RESULT: PASS
```

**`findings: PRESENT (3)` is exactly the #3422 shape**, so stating it plainly: this block is clean **because you ruled those three deferred**, not because the terminal gates on findings. It doesn't. Sound on your ruling; unsound without it.

### Why I armed despite no gate of record at this exact head

I raised this at 14:56Z and you may not have seen it. The last gate (`aeaf72877`) predates two of my commits — `8947cbe1e`, the `root_depth_at` consolidation across six call sites, and `4fc8d540f`, its pins. I resolved it **on evidence rather than by waiting**:

- **My part of the delta touches `scripts/` only** — no compiled input.
- The one compiled-input change in the range is **yours/main's**: `52b67d4c2` deleting `header_fix_functions.rs` ("never compiled, zero references"), already covered by lane-3364's own PASSED gate of record and merged to `main`.
- **The affected artifact is verified directly at this head**: `pub-surface` exits 0 with a coherent measurement line, and its own **42-case suite passed at this head** (10 green controls, 30 reds, zero SKIPs).
- `required` = **SUCCESS**; `premerge-assert 3376 7279b33de` = **OK**; no fresh HOLD.

So the honest claim is **not** "the gate of record certifies this head" — it does not. It is: *the delta cannot affect the compiled crate, and the artifact it does affect is directly verified here.* A gate at this head is still running as post-hoc confirmation and I will report its verdict either way, including if it fails.

### Next, without further asking

`flow-finalize`: OpenSpec n/a (oracle-driven), telemetry via a `telemetry-1712` worktree off `origin/main` with **`--ledger` explicit** (it writes to the *script's* repo otherwise and dirties root), then worktree/branch/claim cleanup and the closing packet.

**Two items will be named in the closing comment rather than quietly satisfied:**
- **C audit: ABSENT — labelled, not implied.** Five subagents went idle without verdicts. AC1 and AC3 self-verified from disk; RED-on-main verified 3×.
- **AC2 still specifies the snapshot the descope deleted** (`issuecomment-5449854018`) — an open scope item for you or the owner, untouched by me.

pmcfadin:
🧭 **You are right, I was wrong, and it was worse than one bad instruction — I gave the same wrong one to THREE lanes.** Thank you for verifying instead of trusting me; that is the whole reason the caution is in every brief, and this is the first time it has paid out against *me*.

**Measured across the fleet just now, before any repair:**

| box | root `HEAD` | behind `origin/main` | has `938d464fb` |
|---|---|--:|---|
| box1 | `2bde26a7c` | 21 | **NO** |
| box2 | `33f8adc3e` | 15 | **NO** |
| box3 | `2bde26a7c` | 18 | **NO** |

So "run it from root, root has the fix" was false on **every box**, not just yours. And the failure mode is the sharpest kind: the unfixed wrapper compares `base` against `origin/main`'s **moving tip**, which is *precisely* the bug the instruction existed to route around. **My instruction would have reproduced the exact failure it was written to fix**, and the resulting `sha-assert: FAIL` would have looked like a genuine stale waiver rather than a stale script — indistinguishable from the real thing, and it would have cost another cycle to diagnose.

**Where my reasoning broke, since the mechanism matters more than the apology:** I checked that `938d464fb` was on `origin/main` and then said the root checkout "is on main". Both statements are true. **"On branch main" and "up to date with origin/main" are different facts, and I substituted one for the other** — a root checkout is only ever as current as its last pull, and nothing on this fleet pulls it. That is the same shape as this repository's own standing lesson about reading Cassandra from a working tree rather than a pinned tag: *a checkout's branch name tells you nothing about its content.*

**Repaired:** all three root checkouts fast-forwarded to `68e4fb384`, `behind=0`, `hasfix=yes` — verified by `merge-base --is-ancestor`, not by the pull reporting success. Each was clean and on `main`, and the command refused rather than forced if either had not been. **lane-1705 had the same wrong instruction in flight and has been corrected**, with the measured table rather than a bare retraction.

**Filing your fleet-relevant note as its own issue**, because it is a standing trap and not a #3392 footnote: nothing keeps `/data/lanes/repo` current, so any advice of the form "run X from root" is silently wrong for as long as the fleet runs, and it fails *looking like* a legitimate verdict.

**On your recheck block — your reading is exactly right and I want the precision on record.** `findings: PRESENT (3)` with `RESULT: PASS` is the #3422 shape, and you stated the dependency correctly: that block is clean **because I ruled those three deferred**, not because the terminal verdict gates on findings. It does not. **Sound on the ruling, unsound without it** — that sentence is the correct way to report a #3422-shaped pass, and it is what I want other lanes to copy instead of pasting `RESULT: PASS`.

**And your `--auto` reasoning is accepted.** `scripts/`-only delta with no compiled input, resolved on evidence rather than by waiting on a redundant gate, is the right call and it is the documented one. Head unmoved at `7279b33de`.

pmcfadin:
## ✅ Post-hoc gate of record: **PASS at the merged head `7279b33de`**

I said I would report this either way, including a failure. It passed.

```
commit: 7279b33  branch: issue-1712-unexport-pub-mod-benchmarks  dirty: no
tree-start == tree-end                 tree-integrity: PASS
non-PASS components: (none)            SKIP census: 0
RESULT: PASS
```

Preserved at `/tmp/cqlite-gates/1712/CERTIFIED-7279b33de.txt`.

**So the merged head is now certified — retroactively, which is worth naming precisely rather than smoothing over.** At arm time it was *not*, and I said so (`issuecomment-5454234121`): I armed on the argument that my part of the delta touched `scripts/` only, that the one compiled-input change was main's own orphan deletion already gated by lane-3364, and that the affected artifact was verified directly at that head. That argument turned out to be right, but **it was an argument, not a certification** — and if this run had failed, the honest position would have been a post-merge revert conversation, not a quiet omission.

**Gate of record now: 4 PASSES** — `299be60ea`, `5701521b8`, `aeaf72877`, and `7279b33de` (the merged head). All zero-SKIP with `tree-integrity: PASS`.

Two closing notes for the record:

- **The three deferred r23 findings are still real**, preserved unapplied at `/tmp/cqlite-gates/1712/UNAPPLIED-r23-fixes.patch`. This gate passing does not retire them — it certifies the tree that shipped *with* them known and ruled deferred. Worth a follow-up issue.
- **`gate_runs=1` in the telemetry ledger is correct and not a typo.** The schema counts runs *through and including the first PASS*; my first full gate passed on its first attempt. The other three were re-certifications after the head moved, not failed rounds. I was about to record `4`, which would have implied three failures that never happened — the schema's wording caught it.

Issue #3377, opened 2026-08-27
Flake: roborev-lints intermittently FAILs assertion 741 (mixed-delivery 'is back') — 1 FAIL then 3 PASS on an unchanged tree


pmcfadin:
## Symptom

The `roborev-lints` gate component intermittently FAILs on assertion 741 of
`scripts/tests/test_roborev_review_guard.sh`:

```
FAIL - structural: delivery-mode classification is back in the flow scripts — mixed-delivery.
       Owner ruling (4) deleted it because FOUR consecutive review rounds each found a
       High-severity false verdict ... (#3312)
passed: 740  failed: 1
GUARD-TEST RESULT: FAIL
```

It is a **flake**, not a code property. Same commit, same clean worktree, same lane path, same
`--only roborev-lints` invocation: **1 FAIL then 3 consecutive PASS (741/0)**.

## Why it is costly out of proportion to its rate

The failure text accuses the *diff* of reintroducing a design regression the owner explicitly
deleted, naming #3312. On a branch that touches nothing under `scripts/`, that reads as a serious
finding, and the honest response is to stop and investigate rather than re-run. It cost this lane
**nine gate invocations** to establish it was noise.

It is also a red on a component that cannot be waived by the #3042 docs-only route whenever the
branch contains any compiled input — so a worker who trusts it is blocked, and a worker who
re-runs until green has learned to re-run reds, which is the worse outcome.

## Measurements (all on one box, 16 cores, `CQLITE_GATE_MAX_CONCURRENCY=1`)

| tree | path | invocation | result |
|---|---|---|---|
| `origin/main` (2bde26a7c) | `/tmp/...` | direct | PASS 741/0 |
| `origin/main` | `/tmp/...` | gate `--only` | PASS 741/0 |
| `origin/main` | `/data/lanes/...` | gate `--only` | PASS 741/0 |
| branch HEAD (46ca55353) | lane, working tree | direct | PASS 741/0 |
| branch HEAD | lane, clean detached | direct | PASS 741/0 |
| branch HEAD | `/tmp/...` | gate `--only` | PASS 741/0 |
| branch HEAD | lane, working tree | gate (`--lite`, `--only`) | **FAIL 740/1** ×2 |
| branch HEAD | lane, clean detached | gate `--only` | **FAIL 740/1** ×1, then **PASS ×3** |

The single-run reads look like a clean "gate + lane path + this branch" interaction. Repetition
[...]

Issue #1452, opened 2026-07-02
Extract cqlite-ffi-common for shared byte-math only (error table, decimal/varint/duration/inet, otel key validation)


pmcfadin:
Parent: #1434

## Why this matters
The byte-level math that turns raw CQL bytes into displayable scalars — decimal→string, varint→integer, inet→string, and the error mapping — is written **twice**, once per binding, in slightly different ways. That duplication is exactly what produced the X1/X2 divergences (#1450 duration/time, #1451 errors): two copies drift because nothing forces them to agree. This extracts a small, pure, dependency-light crate so those paths are shared by construction and can never drift again. It is deliberately a **narrow** extraction, not a general dedupe.

## Current behavior (verified 2026-07-01)
Each byte-math routine is implemented independently per binding:
- **Decimal.** Node `bindings/node/src/value.rs:212-285` (`fn decimal_to_string(scale: i32, unscaled: &[u8]) -> String`) is already a pure Rust byte→string function. Python `bindings/python/src/value.rs:244-308` (`decimal_to_pydecimal`) reimplements the same digit-split/scale logic (:277-301) via Python `int.from_bytes` + string formatting. Two independent implementations of the same math.
- **Varint.** Node `bindings/node/src/value.rs:150-206` (`varint_to_bigint`) computes sign + little-endian u64 words from big-endian two's-complement bytes. Python `bindings/python/src/value.rs:228-241` (`varint_to_pyint`) does the same conceptually via Python `int.from_bytes(..., signed=True)`. Same byte semantics, two code paths.
- **Inet.** Node `bindings/node/src/value.rs:298-317` (`inet_to_string_js`) formats 4/16-byte addresses to a string (and errors on bad length). Python `bindings/python/src/value.rs:476-498` (`inet_to_py`) builds `ipaddress.IPv4Address`/`IPv6Address`. The 4-vs-16 length dispatch and formatting is duplicated. (Behavior on bad length currently differs — that is #1453's concern; this crate hosts the shared length/format logic they will both call.)
- **Duration / date scalar math.** Node `value.rs:288-295` (`duration_to_object`) and the `Value::Date(days)` arm at `value.rs:82`; Python equivalents in `value.rs`. The `(months, days, nanos)` and days-since-epoch arithmetic is small but duplicated.
[...]

PR #3378, opened 2026-08-27
fix(#1451): one authoritative FFI error table — Python and Node stop disagreeing on error identity

### LINKED ISSUE #1451: Shared authoritative error table: Python maps by variant, Node by category — same core error, different identity

pmcfadin:
Closes #1451.

## What this changes

The same core error had a **different identity in each binding**: Python mapped `cqlite_core::Error` to an exception by **variant**, while Node derived `error.code` from `Error::category()`. So "catch `ParseError` for bad CQL" was true in Python and false in Node (`CqlParse` reported `QUERY`), and `Timeout`/`Memory` — distinct `TimeoutError`/`MemoryError` in Python — both collapsed to `IO` in Node.

Both mappings are now driven by **one authoritative table**, keyed by core `Error` variant:
`cqlite-core/src/ffi_error_contract.rs`. Each row is `(py_class, node_code, category, recoverable, prefix)`.

## Where the table lives, and why (AC step 1)

**#1452 (`cqlite-ffi-common`) has not landed**, so of the two fallbacks the AC sanctions I took **one table in `cqlite-core`**, *not* duplicate-copies-plus-a-drift-test:

- The DoD asks for "**one** authoritative … table; both `to_py_err` and `extract_metadata` consume it". A single table satisfies that literally; two copies satisfy it only by proxy.
- The table is inert data — the Python class travels as an **identifier** (`PyExceptionClass`), never a PyO3 type. **No PyO3/napi dependency enters `cqlite-core`, and `cqlite-core/Cargo.toml` is unchanged.**
- It is a **new file**, not an addition to `cqlite-core/src/error.rs` (already 878 lines, over the ~800 target — growing it would trip the `file-size` ratchet).
- When #1452 lands the module relocates wholesale; nothing has to be de-duplicated first.

Core's `Error` enum, `category()` and `is_recoverable()` are **untouched** — `cqlite-core/src/error.rs` has no diff. The reported `category` stays core-derived; only the `code` is finer-grained, and `table_agrees_with_core_category_and_recoverable` pins that.

## User-visible behavior changes (Node `error.code`)

| Core variant | Before | After |
|---|---|---|
| `CqlParse` | `QUERY` | **`PARSE`** |
| `Timeout` | `IO` | **`TIMEOUT`** |
| `Memory` | `IO` | **`MEMORY`** |
| `InvalidInput` | `PARSE` | **`INVALID_INPUT`** |
| `Corruption` | `PARSE` | `PARSE` (unchanged) |

`ErrorCode` in `bindings/node/lib/index.d.ts` gains `TIMEOUT` and `MEMORY`. Consumers switching on `err.code` for these four variants must update; the codes are strictly more specific than what they replace.

## A correction to the issue's premise, worth recording

The issue points at `bindings/node/__test__/error.test.js:54-64` hedging `expect(['PARSE','QUERY']).toContain(e.code)`. That test was **passing on the `QUERY` branch for a reason unrelated to the mapping**: it provoked `'THIS IS NOT VALID SQL!!!'`, whose unknown first token is rejected *before* the CQL parser as `Error::QueryExecution` — legitimately `QUERY`. It never exercised `CqlParse` at all.

So the RED was demonstrated on a **rewritten** test (`'SELECT * FROM'`, which reaches the parser), not on the original. A sibling test now pins the other direction: an unsupported *statement type* must stay `QUERY` and must not borrow the new `PARSE`. The same misleading example had propagated into the docs, which roborev caught (below).

## Tests

- **Rust** `cqlite-core/tests/ffi_error_contract_table.rs` — every row round-trips through `variant_of`; the table agrees with core `category()`/`is_recoverable()`; every row has a non-empty SCREAMING_SNAKE code; `from_name` is fail-closed on an unknown variant; the row count is pinned; each row the issue names is asserted individually; `Timeout`/`Memory` no longer share `IO`'s identity.
- **Python** `bindings/python/tests/test_errors.py` — table-driven: `CqlParse`→`ParseError`, `Timeout`→`TimeoutError`, `Memory`→`MemoryError`, `InvalidInput`→`ValueError`, `Corruption`→`CqliteError`.
- **Node** `bindings/node/__test__/error.test.js` — hedged assertions replaced with authoritative ones, including **negative** assertions (`CqlParse` is `PARSE` and specifically *not* `QUERY`).
- **Compile-time completeness survives**: `variant_of` is exhaustive over `Error`, so adding a core variant **fails to build** until it gets a table row.
- `Timeout`/`Memory` cannot be provoked by any query, which is exactly where the Node divergence hid. Two conformance probes (`_raise_mapped_core_error`, `_errorContractProbe`) make every row assertable; both route through the **production** `to_py_err`/`to_napi_error` path (including Node's `\0code=…\0category=…\0isRecoverable=…` encoding) so they cannot green while production diverges, and both fail closed on an unrecognized variant name.

## Review

**roborev: 1 finding (Low), fixed in `e5bb70fe1`.** Three doc sites illustrated `PARSE` with `SELEC * FORM table` — an unknown-first-token statement that actually yields `QUERY`, i.e. the docs taught the opposite of the semantic this PR establishes. Fixed rather than deferred as a nit because these are lines this PR itself wrote.

**Lead's own finding, fixed in `b2042584d`.** Nothing pinned the `.d.ts` `ErrorCode` union to the table's `node_code` column: a new variant is forced to get a table row (compile error) and trips the pinned row count, but nothing forced the *type surface* to gain the code — a silent type lie, and the "same fact written twice, maintained by hand" shape. Now closed by a **bidirectional** drift assert that takes the authoritative set from the table via a napi seam (not a third hand-written copy), strips `//` comments *before* slicing the declaration (the `TIMEOUT` member's comment contains a `;`, so a naive first-semicolon parse under-reads the union and passes vacuously), and carries an explicit anti-vacuity case.

**`rust-reviewer` did not deliver a report.** The subagent completed and went idle without returning findings, and re-requests went unanswered. Recorded as **not delivered** rather than implied. Standing in for it, the lead verified directly: `category_to_code` is **deleted outright** (no residual second mapping); zero `unwrap`/`expect`/`panic!`/`unreachable!` introduced anywhere under `cqlite-core/src` or `bindings/*/src`; the `\0` metadata encoding is byte-preserved with identical field order; the table's 17 `node_code`s and the union's 17 members match exactly.

## Follow-up (not in scope here)

`bindings/node/src/error.rs::simple_error` hardcodes `code=INVALID_INPUT\0category=Logic\0isRecoverable=false` — a fourth hand-written copy of a mapping fact. **Pre-existing and untouched by this PR** (the diff only adds a nearby assertion); it serves non-core errors like "Database is closed". Deriving it from the `InvalidInput` row belongs in #1452's consolidation.

## Local certification at `46ca553`

```
commit: 46ca553 branch: issue-1451-shared-authoritative-error-table dirty: no
tree-integrity: PASS
file-size: PASS   fmt: PASS   clippy: PASS   roborev-lints: PASS (44s)   scoped-tests: PASS (129s)
python-tier: PASS (maturin develop && pytest bindings/python/tests -m 'not slow' -q)
RESULT: PASS
```

Lite is **not** the gate of record — the full `scripts/agent-gate.sh` runs once in `flow-closer`
before merge. Node, run separately (lite's scope does not cover the JS suites, and the new
`_errorContractNodeCodes` napi seam needs a build first): `npm run build` exit 0, then
**104/104 across `error.test.js` + `typescript-definitions.test.js`**, with the two dataset-backed
cases executing in 149ms/175ms rather than skipping.

### One transient red, explained and filed: #3377

An earlier lite round at this same commit reported `roborev-lints: FAIL` on assertion 741 of
`scripts/tests/test_roborev_review_guard.sh` — "delivery-mode classification is back in the flow
scripts — mixed-delivery". **This PR touches nothing under `scripts/`.** It is a flake, filed as
**#3377** with the measurement table: the same clean detached worktree at this commit gave **1 FAIL
then 3 consecutive PASS (741/0)**, and `origin/main` passes under the same component. The
`mixed-delivery` token genuinely appears on a non-comment line of `scripts/flow/roborev-review.sh:339`
(inside a heredoc, so the assertion's comment-stripping cannot remove it — the raw grep hits on
`origin/main` too), so the check passes only while `$WRAPPER` points at a stripped scratch copy.

Recorded rather than waived: the #3042 docs-only waiver is **void here** (this diff contains `src`),
so this is a cited-flake explanation, not a waiver. `scripts/` was deliberately **not** modified to
turn the component green.

### DIFF
--- a/bindings/node/README.md
+++ b/bindings/node/README.md
@@ -266,7 +266,7 @@ try {
   const result = await db.executeNative('SELECT * FROM keyspace.table');
 } catch (e) {
   // Error code for programmatic handling
-  console.log(`Code: ${e.code}`);        // 'IO', 'SCHEMA', 'QUERY', 'PARSE', etc.
+  console.log(`Code: ${e.code}`);        // 'IO', 'SCHEMA', 'QUERY', 'PARSE', 'TIMEOUT', etc.
 
   // Error category
   console.log(`Category: ${e.category}`); // 'System', 'Schema', 'Query', etc.
@@ -281,16 +281,31 @@ try {
 
 **Error Codes:**
 
+Codes come from the shared FFI error contract (`cqlite_core::ffi_error_contract`),
+keyed by the core error VARIANT — the same table the Python binding reads, so a
+given failure has the same identity in both bindings (issue #1451). A code is
+therefore finer-grained than the `category` it reports (a timeout is `TIMEOUT`
+with category `System`).
+
 | Code | Category | Description | Recoverable |
 |------|----------|-------------|-------------|
-| `IO` | System | File system errors | Yes |
-| `SCHEMA` | Schema | Schema parsing/validation | No |
-| `QUERY` | Query | Query execution failures | No |
-| `PARSE` | Data | CQL syntax errors | No |
-| `CONFIG` | Configuration | Invalid configuration | No |
-| `STORAGE` | Storage | Storage engine errors | No |
+| `IO` | System | File system errors, invalid path | Yes (`IO`) |
+| `SCHEMA` | Schema | Schema parsing/validation, table errors | No |
+| `QUERY` | Query | Query execution failures, result-set budget | No |
+| `PARSE` | Data / Query | CQL syntax errors; corrupt or undecodable data | No |
+| `CONFIG` | Configuration | Invalid configuration or read-path knob | No |
+| `STORAGE` | Storage | Storage engine, index, compaction errors | Yes |
 | `NOT_FOUND` | NotFound | Table/resource not found | No |
-| `INVALID_INPUT` | Logic | Invalid operation (e.g., closed db) | No |
+| `INVALID_INPUT` | Data / Logic | Invalid input, operation, or state (e.g. closed db) | No |
+| `CONCURRENCY` | Concurrency | Lock contention, `write_dir` already locked | Varies |
+| `CONFLICT` | Conflict | Resource already exists | No |
+| `CONSTRAINT` | Constraint | Constraint violation | No |
+| `TRANSACTION` | Transaction | Transaction errors | Yes |
+| `TIMEOUT` | System | Operation exceeded its deadline (never `IO`) | No |
+| `MEMORY` | System | Memory/allocation failure (never `IO`) | Yes |
+| `PLATFORM` | Platform | Platform-specific (WASM) errors | No |
+| `INTERNAL` | Internal | Internal errors | No |
+| `CANCELLED` | Cancelled | Cooperative scan cancellation (never `IO`) | No |
 
 ## Type Conversions
 
--- a/bindings/node/__test__/error.test.js
+++ b/bindings/node/__test__/error.test.js
@@ -51,17 +51,22 @@ describe('Error Mapping Tests (Issue #297)', () => {
     }
   });
 
-  test('Parse error maps to code "PARSE" or "QUERY"', async () => {
+  test('CQL parse error maps to the authoritative code "PARSE"', async () => {
+    // Issue #1451: this assertion used to hedge (`['PARSE','QUERY']`) because
+    // Node derived its code from `category()`, and `CqlParse` is Query-category
+    // — so it reported 'QUERY' while Python raised ParseError for the same core
+    // error. The shared contract table (cqlite_core::ffi_error_contract) now
+    // decides BY VARIANT, so exactly one code is correct here.
     skipIfNoDatasets();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
     });
     try {
-      expect.assertions(1);
+      expect.assertions(2);
       await db.execute('THIS IS NOT VALID SQL!!!');
     } catch (e) {
-      // Parse errors may come as PARSE or QUERY depending on where the error is caught
-      expect(['PARSE', 'QUERY']).toContain(e.code);
+      expect(e.code).toBe('PARSE');
+      expect(e.category).toBe('Query');
     } finally {
       await db.close();
     }
--- a/bindings/node/__test__/error.test.js
+++ b/bindings/node/__test__/error.test.js
@@ -57,6 +57,31 @@ describe('Error Mapping Tests (Issue #297)', () => {
     // — so it reported 'QUERY' while Python raised ParseError for the same core
     // error. The shared contract table (cqlite_core::ffi_error_contract) now
     // decides BY VARIANT, so exactly one code is correct here.
+    //
+    // The statement must actually REACH the CQL parser: a statement whose very
+    // first token is not a known verb is rejected earlier as
+    // `Error::QueryExecution` ("Unsupported query type"), which is genuinely
+    // `QUERY` — see the sibling case below.
+    skipIfNoDatasets();
+    const db = await Database.open(global.testPaths.SSTABLES_DIR, {
+      schema: global.testPaths.SCHEMA_BASIC_TYPES,
+    });
+    try {
+      expect.assertions(3);
+      await db.execute('SELECT * FROM');
+    } catch (e) {
+      expect(e.code).toBe('PARSE');
+      expect(e.category).toBe('Query');
+      expect(e.message).toContain('ParseError:');
+    } finally {
+      await db.close();
+    }
+  });
+
+  test('an unrecognized statement type stays "QUERY", not "PARSE"', async () => {
+    // The other side of the #1451 fix: `PARSE` now means a genuine CQL syntax
+    // failure, so an unsupported STATEMENT TYPE (an `Error::QueryExecution`)
+    // must NOT borrow it.
     skipIfNoDatasets();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
@@ -65,8 +90,8 @@ describe('Error Mapping Tests (Issue #297)', () => {
       expect.assertions(2);
       await db.execute('THIS IS NOT VALID SQL!!!');
     } catch (e) {
-      expect(e.code).toBe('PARSE');
-      expect(e.category).toBe('Query');
+      expect(e.code).toBe('QUERY');
+      expect(e.message).toContain('QueryError:');
     } finally {
       await db.close();
     }
--- a/bindings/node/__test__/error.test.js
+++ b/bindings/node/__test__/error.test.js
@@ -12,7 +12,7 @@
  * - [x] Test: Error message contains original error text
  */
 
-const { Database } = require('../lib/index.js');
+const { Database, _errorContractProbe } = require('../lib/index.js');
 const { skipIfNoDatasets, getNonexistentPath } = require('./helpers.js');
 
 describe('Error Mapping Tests (Issue #297)', () => {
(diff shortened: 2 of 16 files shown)


pmcfadin:
## C — intent audit: **PASS**

`spec-auditor` against the issue #1451 body (this issue is prescriptive-routed with **no** OpenSpec
change, so the checkbox DoD + the 6 numbered steps + the "Do NOT" fence are the acceptance criteria).
Every criterion `satisfied` with public-surface test evidence; no `unmet`, no unjustified `partial`.
The only open DoD box is the full `agent-gate.sh` run, which is the closer's step by doctrine and
outside the audit's scope.

**It verified the compile-time guard mechanism structurally rather than trusting the doc comment** —
three chained matches, each exhaustive with no wildcard arm over a non-`#[non_exhaustive]` enum:
`variant_of` (a new core variant breaks the *core* build), the macro-generated `row()` plus an
exhaustive `sample_error` (a new key cannot exist without a row *and* a representative), and
`expected_py_class`/`expected_node_code` inside each binding's test module (a new variant also breaks
each binding's test build until its identity is reviewed). Value-level backstop: `row_count_is_pinned`.
Measured **37 table rows vs 37 `Error` variants**.

It also confirmed, independently of my own pass: `category_to_code`/`category_to_prefix` deleted with
**0 repo-wide references**; the NUL metadata format string byte-identical to `origin/main`; prefix
changes confined to exactly the four divergence rows this issue exists to fix, with the other 33 rows
diffed row-by-row against `origin/main`'s `category_to_prefix` and unchanged; zero
`unwrap`/`expect`/`panic!`/`unreachable!` across all five touched Rust files including test modules.

On the two conformance probes it agreed they are adequate wiring evidence, and gave a sharper reason
than I had: they supply only the **error value** — the one part no real query can supply — while the
call chain is entirely production (`_raise_mapped_core_error` → `to_py_err`; `errorContractProbe` →
`to_napi_error` → NUL encoding → the same `enhanceError` wrapper every `Database` method uses). Every
variant a query *can* provoke is additionally covered through the real public API, so the probes are
the sole surface for nothing except `Timeout`/`Memory`. Both are asserted to **always throw**, closing
the assert-inside-catch vacuity hole, and the DB-driven cases use `expect.assertions(N)` for the same
reason.

### Nits batched for ONE follow-up issue at merge (per the severity rubric — none is a blocker)

1. **`bindings/python/tests/test_errors.py:249-250` docstring is inaccurate**: it claims a
   present-but-empty dataset root "FAILS LOUDLY under strict fixture mode via `require_test_data`".
   `require_test_data` actually tests directory **existence**, not emptiness — present-but-empty simply
   does not skip, which is why the dataset criterion is met. Behaviour correct, mechanism
   misdescribed. Found by the audit, not by me.
2. **Node binding release note.** Four `err.code` values change for consumers and the node binding
   ships independently, so a CHANGELOG line is worth adding. Not required by any AC in #1451, and
   release framing is the owner's call, so it is proposed rather than done here.
3. `simple_error`'s hardcoded fourth copy of a mapping fact (already recorded in this PR body) → #1452.
4. `PARSE` remains overloaded on the Node side (`Serialization`/`Corruption` share it with genuine CQL
   parse failures). Pre-existing coarseness the issue did not ask to fix; the `.d.ts` comment now says
   so explicitly.

Audit caveat, stated as the auditor stated it: it re-ran no suite (`import cqlite` resolves only inside
the maturin venv), so run evidence is the lite `python-tier: PASS` and the separately measured Node
104/104. Every criterion above was verified against source and test text.

pmcfadin:
## Correction: the flake citation is **#3367**, not #3377

The PR body cites #3377 for the transient `roborev-lints` red. **#3377 was a duplicate and is now closed** — the canonical report is **#3367**, filed earlier the same day from lane-1714: same component, same assert (classifier-GONE / `mixed-delivery`), same `740/1` tally, and the same two causes.

Please read the citation in the PR body as **#3367**. My independent measurement matrix has been moved onto #3367, including a control that report lacked (`origin/main` passes the same component at both a `/tmp` and a lane path), which shows the two causes are conjunctive: the token's presence on a non-comment line is necessary but not sufficient — it reds only when the mutable `WRAPPER` global happens to hold the real path at assert time.

**One substantive thing #3367 knows that I did not**, and it matters for how this red should be treated: beneath the intermittency sits a **deterministic true-positive** on the wrapper's own doctrine prose, and the flakiness is what masks it. So this is not purely noise — fixing only the order-dependence would turn the component permanently red. That is `scripts/`-owned work and correctly outside this PR, which touches nothing under `scripts/`.

Nothing about this PR's certification changes: lite `RESULT: PASS` at `46ca553` with `roborev-lints: PASS (44s)`, and the full gate of record is still the closer's single run. The red is explained by a cited issue, **not** waived — the #3042 docs-only waiver is void here because this diff contains `src`, and `scripts/` was deliberately left untouched rather than edited to force green.

pmcfadin:
## Gate of record — `RESULT: PASS` at `46ca553`

Certified after five infrastructural failures (none code-related; details below the block).

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.dIomWn
commit: 46ca553 branch: issue-1451-shared-authoritative-error-table dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-1451/test-data/schemas (checkout-relative)
tree-start: 46ca553530ca dirty: no digest: 74e258471b37
tree-end: 46ca553530ca dirty: no digest: 74e258471b37
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (5s)
clippy:            PASS (5s)
roborev-lints:     PASS (39s)
core-tests:        PASS (48s)
tombstones-scan:   PASS (0s)
scan-offload-guard: PASS (2s)
work-counters-guard: PASS (17s)
byte-budget-guard: PASS (13s)
arrow-parity-guard: PASS (1s)
memory-budget:     PASS (11s)
integration-tests: PASS (1s)
format-compat:     PASS (0s)
write-tests:       PASS (10s)
cli-tests:         PASS (5s)
python-bindings:   PASS (64s)
node-bindings:     PASS (116s)
delivery-telemetry: PASS (0s)
oom-audit:         PASS (1s)
parity-report:     PASS (1s)
operator-metrics-doc: PASS (0s)
tooling-tests:     PASS (710s)
minimal-build:     PASS (1s)
smoke:             PASS (3s)
logs: /tmp/agent-gate.dIomWn
summary-file: /tmp/cqlite-gates/1451/full-46ca553-r6-235408.txt
RESULT: PASS
```

### Certification set

| Stage | Verdict |
|---|---|
| Full agent-gate (gate of record) | **PASS** — 30/30 components, 0 FAIL, 0 SKIP, `tree-integrity: PASS` |
| Final roborev (job 18) | **PASS** — range `2bde26a7..46ca5535`, `findings: NONE`, 1.6M input tokens |
| C intent audit (issue ACs) | **PASS** — see note below |
| `premerge-assert 3378 46ca553` | `PREMERGE: OK` |

`python-bindings: PASS` and `node-bindings: PASS` — the two components that carry this PR's 13 binding files — are affirmatively green, not green-by-absence.

**C was performed by the lead, not by `spec-auditor`.** The subagent went idle twice without delivering a verdict, including after a direct request, so rather than record an unsupported PASS the lead audited from disk. Verified: `to_py_err` and `extract_metadata` both consume `contract_for()`; `category_to_code` is grep-confirmed deleted repo-wide (no surviving second mapping); the `cqlite-core` placement is the first of the two fallbacks AC step 1 sanctions verbatim, with the module doc marking it as the shared FFI contract and naming the #1452 relocation; both conformance probes route through **production** `to_py_err`/`to_napi_error` and supply only a `sample_error()` value, so they cannot green while production diverges; `cqlite-core/src/error.rs` has a **zero-line diff**; the `\0` field order is byte-identical; no `unwrap`/`expect`/`panic` added under any `src`; file sizes 363/623/572/227 all under target; `simple_error`'s hardcoded `INVALID_INPUT` confirmed pre-existing on `origin/main:225` and untouched, so deferring it to #1452 leaves no AC unmet.

### Two flakes hit en route, both filed with data (neither attributable to this diff)

- **`roborev-lints` assertion 741** (`mixed-delivery 'is back'`) = **#3377**, matched by assert text. The assertion inspects `scripts/flow/**`, which this PR does not touch. Recurrence reported: #3377 (comment).
- **`core-tests` thread-budget pins** = **#3385** class. `issue_2316` delta=28/bound=15 and `issue_2370` delta=26/bound=24 under gate load; **standalone 3/3 on this identical tree gives delta=8 and delta=12** — roughly half the bound, stable. A regression building a `num_cpus`-sized runtime would show elevated deltas standalone too. Data added to #3385, extending it two ways: the margin is 13 threads over rather than "exactly one", and `issue_2370` reds identically but is not named there.

No waiver was available or used — `classify-docs-only.sh` exits 1 on this diff (`bindings/node/__test__/error.test.js`), so the gate had to genuinely pass, and it did.

Issue #3380, opened 2026-08-27
test_roborev_review_guard.sh: the #3312 structural assert fails intermittently on clean main (load-correlated), and scans heredoc prose its own comment excludes


pmcfadin:
Observed while delivering #1699. **Not caused by that change** — reproduced on a clean `origin/main`
worktree at `2bde26a7c`, and `scripts/tests/test_roborev_review_guard.sh` / `scripts/flow/roborev-review*.sh`
are not in #1699's diff.

## Observation

`scripts/tests/test_roborev_review_guard.sh` intermittently fails ONE assert — the #3312 owner-ruling (4)
structural check:

```
FAIL - structural: delivery-mode classification is back in the flow scripts — mixed-delivery.
       Owner ruling (4) deleted it because FOUR consecutive review rounds each found a
       High-severity false verdict ... (#3312)
passed: 740  failed: 1
GUARD-TEST RESULT: FAIL
```

| run | tree | box state | result |
|---|---|---|---|
| 1 | `issue-1699-…` @ `9a4c0be7e` | heavy: baseline full gate + other lanes' gates | **FAIL** 740/1 |
| 2 | clean `origin/main` @ `2bde26a7c` | same heavy load | **FAIL** 740/1 |
| 3 | `issue-1699-…` @ `9a4c0be7e` (unchanged from run 1) | quiet | **PASS** 741/0 |

Runs 1 and 3 are the **same tree, same commit, no edits between them**. So the assert is not deciding on
tree content.

## Why this matters

The guard gates merges twice over: it runs in `roborev-lints` (which `--lite` includes,
`agent-gate.sh:4799`) and again in `tooling-tests` (`agent-gate.sh:6029`). An intermittent red there blocks
any lane's `--lite` and any full gate, and — worse for triage — it reds with a message that reads like a
**design regression someone just introduced**, which is exactly the wrong signal to hand an agent who did
not touch roborev.

## What I could and could not establish

**Established:** it fails on clean `main`; it is not tree-content-dependent; both failures were under heavy
co-scheduled load and the pass was on a quiet box.

**Not established — deliberately not guessed at:** the mechanism. Three hypotheses were checked and
**falsified**: concurrent runs sharing a scratch path (no — `tmp=$(mktemp -d …)` and
`_gm_dir="$tmp/grammar-mutant"` are unique per run, `:66`/`:2177`); a missed `WRAPPER` restore (the restore
[...]

Issue #2642, opened 2026-07-17
Retire wall-clock asserts in the correctness path (#2369 rule) + rewire orphaned cqlite-cli tests/unit


pmcfadin:
Epic #2636. Audit test F1/F4 (scope revised per review). Two parts:
1. LIVE FLAKES: sstable_performance_regression_tests.rs has wall-clock asserts (MAX_FIRST_OPERATION_MS=50; lines 92/98/107/116/431/449/472) running in default core-tests — top latent flake. write_integration.rs:742 (elapsed<50ms) likewise.
2. COVERAGE GAP (new): cqlite-cli/tests/unit/enhanced_unit_tests.rs is ORPHANED — no harness references tests/unit/, so it never compiles or runs. Its 100ms assert is inert. Rewire (or consciously retire) the orphaned unit/ directory, applying record-not-assert to any timing checks during rewiring.
Acceptance: no wall-clock threshold assert remains in the default correctness gate path (record timings, or opt-in perf gate + #[ignore]); orphaned tests wired in or retired with rationale; a lint/reviewer note prevents reintroduction; gate green under induced load.

pmcfadin:
REVIEW CORRECTION (verified): cqlite-cli/tests/unit/enhanced_unit_tests.rs is ORPHANED — no top-level harness references tests/unit/, so it never compiles or runs. Its 100ms assert is therefore currently inert, and the real finding is MISSING TEST COVERAGE. Scope addition: rewire (or consciously retire) the orphaned unit/ directory, applying record-not-assert to any timing checks as part of the rewiring.

pmcfadin:
Delivered via PR #2692 (squash-merged `176280b8` at 2026-07-18T08:44Z).

- Full agent-gate of record: PASS on `e861f579` (all 28 components; work-counters-guard + tooling-tests green; file-size ack for the pre-existing over-threshold legacy `tombstone_merger.rs`).
- Acceptance criteria met: no wall-clock threshold assert remains in the default correctness gate path (converted to `[perf-record]`); orphaned `cqlite-cli/tests/unit/` retired with rationale; new SKIP-aware lint guard `check-no-wallclock-asserts.sh` + self-test wired into the gate's tooling-tests to prevent reintroduction.
[...]

Issue #1696, opened 2026-07-02
AH3 — Decorative-knob purge + Config::validate() honesty + set-knob-assert-behavior guard


pmcfadin:
Part of epic #1685 (config honesty). Source: `docs/reports/platform-observability-audit-2026-07-01.md` finding AH3 (audit block 6). Executes decision #12's spirit knob-by-knob; AH4 owns the architecture. **The compaction struct is NOT in scope here — N1 #1619 owns wiring it.**

## Problem
The decorative-knob census (each verified zero-consumer): bloom-filter knobs + their **4 dedicated validation tests** (`cqlite-core/src/config.rs:45-48, 712-720, 774-835` — `Config::validate()` diligently validates fields nothing reads), `io_threads`, `sync_mode`, `block_size` (reader hardcodes 65536), `query.max_result_rows`, `query.plan_cache_size`, `query.enable_optimization`, `query.parallel.*`, `performance.*`, `max_sstable_size`, `memtable_size_threshold` (shadowed by the private `WriteEngineConfig`), and in the CLI: the entire `ConnectionConfig` (CQLite has no network!), `pager`, `timestamp_format`, `performance.{query_timeout_ms→AH2, memory_limit_mb, cache_size_mb}`. Meanwhile the LIVE `direct_io_memory_fraction` is only clamped, never validated.

## Fix (effort: M)
1. For EVERY listed knob, one of exactly two dispositions, recorded in a table in the PR:
   - **Delete** (default): remove field + its validation + its tests. For serde compat, decide crate-wide: either `deny_unknown_fields` is not set (old configs with the field still parse — document the field as removed) or add a doc note. Keep it consistent.
   - **Wire** (only if an obvious ≤S-effort consumer exists — e.g. if `query.max_result_rows` maps cleanly onto D6 #1582's budget, note it there instead and delete here).
2. Exclusions: `MemoryConfig` rides the B2 #1568 decision (leave, note); compaction → N1 #1619; `max_execution_time` → AH2.
3. Add validation for the surviving LIVE knobs (`direct_io_memory_fraction` range check, etc.).
4. Introduce the standing guard: for every surviving public knob, a **set-knob-assert-behavior-changes test** (set knob → observe a behavioral/observable difference). Unwritable for a decorative knob — which is the point; a knob without such a test gets deleted or an explicit `// reserved:` justification.

### Do NOT
- Do not wire a knob speculatively "to keep it" — wiring without a consumer need is how decorative knobs happened.
[...]

PR #3381, opened 2026-08-27
fix(#1697): one public Config for the write path — thread the live knobs, delete the decorative mirrors

### LINKED ISSUE #1697: AH4 — Config source of truth: one public Config, delete the decorative mirrors (DECIDED)

pmcfadin:
Closes #1697.

## What was wrong

CQLite carried two independent config layers for the write path. The public `Config.storage.*` facade an embedder can set, and the private `WriteEngineConfig` the engine actually reads — and they disagreed. Measured on `origin/main`:

- `Config.storage.memtable_size_threshold` advertised **16MB** and had **zero production readers**. Not "few" — zero. The two apparent readers (`tests/integration/performance_integration.rs`, `tests/benchmarks/load_testing.rs`) are in files that are not registered cargo targets and never compile; one of them references a field that does not exist, which is how you can tell.
- The engine actually flushed at `WriteEngineConfig::memtable_flush_threshold` = **64MB**. So the shipped public knob was wrong by 4x, and setting it did nothing at all.
- STCS `min_threshold`/`max_threshold` (4/32) and `memtable_hard_limit` (256MB) had **no public counterpart of any kind** — an embedder could be hard-failed by an admission ceiling they had no way to see or change.
- Three production sites each built a `WriteEngineConfig` independently, and only two of them threaded even `auto_compaction`. `cqlite-cli/src/main.rs` threaded nothing, so a CLI writable session ignored `Config.storage.compaction` entirely.

## The fix: one bridge, four callers

`WriteEngineConfig::from_config(&Config, data_dir, wal_dir, schema)` in the new `cqlite-core/src/storage/write_engine/config_bridge.rs` is now the ONE place a public config value is translated for the write path. `WriteEngineConfig::new` is defined as `from_config(&Config::default(), ..)`, so every write-path knob has exactly one literal default, in `Config::default()`. All four production write paths — core, CLI, Python, Node — now go through it.

Per the issue's "do not introduce a sixth layer", this is a deletion of a layer, not an addition: the competing `with_compaction_config` translation is gone.

### Silo -> public field mapping (AC1)

| live knob (private, pre-#1697) | public `Config` field | default before | default after |
|---|---|---|---|
| `WriteEngineConfig::memtable_flush_threshold` (`write_engine/mod.rs:192`, const `:223`) | `storage.memtable_size_threshold` | private 64MB / **public 16MB (dead)** | **64MB** (the value that always ran) |
| `WriteEngineConfig::memtable_hard_limit` (`mod.rs:195`, const `:225`) | `storage.memtable_hard_limit` | 256MB, **no public field** | 256MB (unchanged — exposure only) |
| `WriteEngineConfig::compaction_min_threshold` (`mod.rs:213`, const `:227`) | `storage.compaction.min_threshold` | 4, **no public field** | 4 |
| `WriteEngineConfig::compaction_max_threshold` (`mod.rs:217`, const `:229`) | `storage.compaction.max_threshold` | 32, **no public field** | 32 |
| `WriteEngineConfig::auto_compaction` | `storage.compaction.auto_compaction` | already wired (#1619/N1) | `true` (unchanged) |

Duplicate literal defaults deleted: `DEFAULT_FLUSH_THRESHOLD`, `DEFAULT_COMPACTION_MIN_THRESHOLD`, `DEFAULT_COMPACTION_MAX_THRESHOLD`. Each knob now has exactly one literal.

### Behaviour change, stated explicitly

**`Config.storage.memtable_size_threshold`'s default moves 16MB -> 64MB.** This is required by the issue's rule: *"if a public default differed from the private one that actually ran, keep the RUNNING value as the new public default."* 64MB is what always ran. Adopting the decorative 16MB would have silently quadrupled every embedder's flush rate — a real regression dressed up as a bugfix. `performance_optimized()` moves 64MB -> 128MB so the preset still sits above the default rather than becoming a no-op. Both are in the CHANGELOG.

## TDD evidence (the issue's "must FAIL on current main")

`cqlite-core/tests/issue_1697_config_single_source.rs`, one guard per knob, each driving the PUBLIC field and asserting an OBSERVABLE engine effect — never elapsed time (#2642). Observed RED before the bridge landed (full output in commit `27cb2a38f`):

```
running 4 tests
test public_compaction_thresholds_drive_stcs ... FAILED
test public_auto_compaction_off_disables_compaction ... ok
test public_compaction_max_threshold_caps_merge_width ... FAILED
test public_memtable_threshold_drives_flush ... FAILED

public memtable_size_threshold = 4096 B must trigger a flush after ~51KB of
  writes, but no Data.db exists under ".../data/cfg_ks/knobs" (the public
  knob never reached the engine; the private 64MB default ran instead)

test result: FAILED. 1 passed; 3 failed
```

The one PASS is deliberate: `auto_compaction` was the single already-wired knob (#1619/N1), so it is the regression anchor proving the new bridge did not break existing wiring.

## Review rounds

Both roborev and `rust-reviewer` ran on the lite-green diff BEFORE the full gate (review-first, #2086). Between them they found two blockers, both fixed:

1. **Python validated `flush_threshold` against the DEFAULT hard limit, not the caller's** — and the fold that applies it ran *after* `validate()`, so it was never re-validated. On the documented happy path (take a preset dict, mutate it) `memtable_hard_limit=64MB` + `flush_threshold=200MB` was ACCEPTED and dead-ended the write path permanently: flush never fires at 200MB, admission rejects at 64MB. Node was already correct; this was a Python-only asymmetry.
   The general shape is worth recording: the code comment read *"the ceiling is a PUBLIC knob now (same 256MB value)"* — true when written, and **this change falsified it**. Making a constant configurable turns every remaining reader of that constant into a latent bug.
2. **`with_compaction_config` had become a second competing `CompactionConfig` -> engine translation** with zero production callers, extended from one value to three. Two translations with nothing asserting they agree is the exact defect class this issue exists to remove, so it is deleted.

Also fixed from review: `validate()` messages now name the offending values (the rule they front, `STCSPolicy::new`, already did — the earlier check was strictly less informative than the one it pre-empts); the two new compaction rules got test coverage; a `u64`->`usize` clamp that could manufacture `memtable_hard_limit == usize::MAX` on wasm32 (never flush AND never reject -> OOM, since `saturating_add` caps at the ceiling) is now **rejected** in `validate()` with the clamp kept as defense in depth; the CLI validates its public write config before the bridge sees it (extracted to `cqlite-cli/src/write_engine_config.rs`); serde-absent tests pin the three new `#[serde(default)]` attributes.

Two behaviour changes the review round surfaced, both deliberate and in the CHANGELOG: `CQLITE_MEMTABLE_FLUSH_THRESHOLD` now **errors** on a value that would wedge the engine (e.g. `300000000`, above the 256MB ceiling) instead of accepting it silently, and **errors on a malformed value** instead of silently falling back to the default (empty/whitespace still reads as unset). It is also parsed as `u64`, so a 32-bit host no longer silently ignores anything above `u32::MAX`.

### A gap worth naming

`rust-reviewer` observed that every guard in the new suite reaches the engine through a local helper calling `from_config` directly — so the bridge is proven, but **no guard exercises a production caller**. That is precisely why BLOCKER 1 (a binding-level bug) was invisible to a green suite. The Python regression test added in this PR is the first guard that goes through a real caller; extending that to the CLI and Node paths is left as follow-up rather than claimed as done.

## Notes

- New public fields all carry `#[serde(default = ...)]`. This is load-bearing, not tidiness: the Python bindings are a serde JSON bridge requiring a COMPLETE dict, so a bare added field would break every pre-upgrade user config.
- `cqlite-core/src/config.rs`'s inline `#[cfg(test)]` module moved to `config_tests.rs` via `#[path]` (1133 -> ~844 lines, campsite rule / epic #1135). Verified a pure move: 18 tests before and after, identical name sets, bodies identical modulo indentation.
- The cited source doc `docs/reports/platform-observability-audit-2026-07-01.md` is not in the repo; the issue body was treated as the operative spec.
- Sequencing: #1619 (N1) is closed; #1696 (AH3) is held by the owner behind this PR, which inverts the issue body's stated "after AH3" ordering. Scope was kept clear of `Config.query` (#1695/AH2, live-claimed elsewhere) and of the public decorative-knob purge (#1696).

## Deferred (follow-ups, deliberately not in this PR)
- Three remaining config silos: STCS `bucket_low`/`bucket_high`/`min_sstable_size` hardcoded at `write_engine/mod.rs:686-694`; 5 always-default `SSTableReaderConfig` fields; and `merge/producer_iter.rs:364` + `merge/point_read.rs:696` fabricating a fresh `Config::default()` so a user's `disk_access_mode` is honoured on reads but silently ignored during compaction (arguably a bug, ranked highest).
- Enforcement posture: `Config::validate()` is still not called by `Database::open` or by `from_config`; four write paths have differing postures.
- Presets publish incoherent memory budgets now that the ceiling is visible (`memory_optimized()`: `max_memory=256MB` with `hard_limit=256MB`; `wasm_optimized()`: 128MB vs 256MB). Left alone deliberately — 256MB is the value that always ran, and the issue forbids silent behaviour changes.
- `CompactionConfig`'s new `min_threshold`/`max_threshold` are `usize` while sibling byte knobs are `u64`: a JSON config valid on x86_64 could fail on wasm32. Consistent with the existing `mmap_min_size_bytes` precedent, so not new.
- Node `DatabaseOptions` does not expose the two new STCS knobs.

## Gate evidence (pre-full-gate)

`AGENT-GATE LITE SUMMARY` **RESULT: PASS** at `8aac76f`, `tree-integrity: PASS`, every component green
(`file-size`, `fmt`, `clippy`, `roborev-lints`, `scoped-tests`, plus the `python-tier`). The full
`agent-gate.sh` run of record follows in the closer and will be pasted here.

**Known-flake note for the full gate (#3380).** An earlier lite run at this *same* commit red on
`roborev-lints` with one failing assert out of 741 —
`structural: delivery-mode classification is back in the flow scripts — mixed-delivery` — and on
`scoped-tests` with `collect2: fatal error: ld terminated with signal 7 [Bus error]` while `/data`
sat at 96% full (12GB free, four lanes building concurrently). Re-run on the identical tree, both
components PASS. So:

- the `roborev-lints` assert is the already-filed load-correlated flake **#3380** (see also **#3367**);
  it scans `scripts/flow/roborev-review.sh`, and this PR touches **no `scripts/` file** — both scanned
  files are byte-identical to `origin/main` here, so the assert cannot be reacting to this diff;
- the `scoped-tests` red was disk exhaustion at link time, not a test failure: every test that ran
  passed (212 / 3485 / 6, zero failures).

Neither was patched around. Recording them because this diff contains compiled input, so the
docs-only waiver does not apply by default — what justifies discounting them is the positive evidence
above (unchanged scanned files, plus a clean re-run of the same tree), not a path-shape argument.

## Infrastructure workarounds the gate of record runs under (disclosed, zero repo change)

Neither touches a tracked file; both are recorded because a gate of record should certify in a
configuration the reader can see.

**1. `#3399` — a ws0 test fixture OOM-kills the box, and destroyed three runs of this gate.**
`scripts/tests/test_ws0_report_guards.sh` feeds `--reps 99999999999999999999` (deliberately, to
observe the reporter's `MAX_COUNT` cap fire). That value reaches the *fixture generator* first —
`ws0_pin_boundary_observations` → `ws0_boundary_observations.py:144 expected_boundary_labels()`, an
unbounded list comprehension over `range(1, 10**20)` — which grows at a measured ~420 MB/s to
**~28 GB RSS**, twice per run, on this 30 GB **swapless** box, triggering a **global OOM**.

Attribution is not inferred: the gate's own `tooling-tests.log` (`/tmp/agent-gate.en6Et2`) ends on
the check immediately preceding the runaway with nothing after it, `dmesg -T` shows four global OOM
kills each a ~28 GB `python3`, and the failure reproduces standalone under `ulimit -v` with a
`MemoryError` at `ws0_boundary_observations.py:145`. Pre-existing on `main` (introduced 6f3f2dd09,
#3272/#3324); this PR touches nothing under `scripts/`. Filed as **#3399** (bug/P1) rather than
fixed here — out of scope, and a fix would invalidate the certified SHA.

The gate of record therefore runs inside:

```
systemd-run --user --scope -p MemoryMax=18G -p MemorySwapMax=0 -p OOMPolicy=continue
```

`OOMPolicy=continue` is load-bearing: systemd's **default `OOMPolicy=stop`** kills the entire scope
on a cgroup OOM, so a bare `MemoryMax` reproduces the gate death (measured — same check, exit 143).
`continue` sets `memory.oom.group=0`, so only the largest *task* is killed (the runaway `python3`,
which dwarfs every `rustc`). Verified on the live cgroup: `memory.max=19327352832`,
`memory.swap.max=0`, `memory.oom.group=0`.

**Verdict fidelity was measured, not assumed.** The suite inside an 8 G scope: **exit 0, all 107
checks passed, 0 FAIL, 70 s, no new global OOM.** The two absurd-`--reps` checks still report `ok`,
because their subject is the *reporter's* refusal and an unwritten boundary file never affected it.
A rejected alternative is worth recording: a whole-subshell `ulimit -v` contained the box but
produced a **false red** (`the driver must print its selection as PARTIAL/FULL up front`) by capping
the driver's own legitimate work — that check passes under the scope. A false red in a gate of record
is worse than the runaway.

**2. `#3391` — `test_bti_perf_scan.sh` ignores `CARGO_TARGET_DIR`.** It builds with `cargo build`
(which honors the variable) but resolves the binary at a hardcoded `$REPO_ROOT/target/debug/...`.
Relocating this lane's build artifacts off a full `/data` therefore failed the assert on a binary it
had just built successfully. Worked around with a `target` symlink → `/var/tmp/cqlite-1697-target`,
so the hardcoded path resolves and the artifacts stay on the other volume; pre-flighted before use
(`test_bti_perf_scan: ALL PASS (129 cases)`, previously `passes=1 fails=2` against a floor of 129).
Filed as **#3391** with a one-line fix. The symlink is in `.git/info/exclude`, and the gate's own
`tree-integrity`/`dirty: no` confirms it does not perturb tree identity.

**Neither red was patched around in source.** No source file was modified to turn a gate green.


pmcfadin:
## Gate of record — RESULT: PASS

All 30 components PASS, zero FAIL. Reader contract verified: `run-id` matches the run this lane
bound, `tree-integrity: PASS`, `tree-start == tree-end`, and `commit: c17933c dirty: no` equals both
the certified SHA and this PR head.

Note the two components that had previously been red or fatal are green here: `tooling-tests` passes
under the #3399 containment described below (and `dmesg` records no new global OOM), and `core-tests`
passes on an idle box, consistent with #3385/#2370 being load-correlated rather than caused by this diff.

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.SSMVQk
commit: c17933c branch: issue-1697-config-source-of-truth dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-1697/test-data/schemas (checkout-relative)
tree-start: c17933cb1fb8 dirty: no digest: c9d019890860
tree-end: c17933cb1fb8 dirty: no digest: c9d019890860
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (6s)
clippy:            PASS (6s)
roborev-lints:     PASS (47s)
core-tests:        PASS (54s)
tombstones-scan:   PASS (0s)
scan-offload-guard: PASS (3s)
work-counters-guard: PASS (20s)
byte-budget-guard: PASS (13s)
arrow-parity-guard: PASS (0s)
memory-budget:     PASS (11s)
integration-tests: PASS (0s)
format-compat:     PASS (1s)
write-tests:       PASS (10s)
cli-tests:         PASS (6s)
python-bindings:   PASS (84s)
node-bindings:     PASS (125s)
delivery-telemetry: PASS (1s)
oom-audit:         PASS (0s)
parity-report:     PASS (1s)
tooling-tests:     PASS (752s)
minimal-build:     PASS (82s)
smoke:             PASS (6s)
logs: /tmp/agent-gate.SSMVQk
summary-file: /tmp/gate-full-1697-r7.txt
RESULT: PASS
```

pmcfadin:
## Final roborev pass: RESULT: FAIL — 2 findings, both triaged BLOCKER. Merge held; the `c17933cb1` gate certification is void.

The review is genuine, so this is not one of the known vacuous-pass modes: `reviewed-sha` is the range `5b8ae405806738e63fbba96075b50913f5874e96..c17933cb1fb8dcf0553beab5c5f24acfdeb7eef0` (base equals current `origin/main`, head equals the certified SHA), job 33 `done`, and every vacuity defence passed affirmatively — `push-assert`, `census-check`, `code-free`, `job-record`, `sha-assert`, `review-completed`, `prompt-content: PASS (12/12 code census paths present)`, `vacuity-tier1`, `vacuity-tier2`. Token accounting `input=723026 cached=636416 output=5427`, against a vacuous baseline of ~18.7k / 0 / 53.

Both findings were **verified against the code** before being accepted, and both are BLOCKERS under `docs/development/roborev-severity.md` (correctness / control-flow with a concrete failure scenario).

**B1 — validation runs before the override it constrains.** `bindings/python/src/database.rs:849` → `config_from_py` (`bindings/python/src/config.rs:206-233`) calls `validate()` on the **base** config; the `flush_threshold` fold happens afterwards. A config invalid *only* in the field the override replaces is rejected even though the merged config would be valid. `config_from_py` has a second caller (`config.rs:176`) that must keep validating, so the fix needs an unvalidated variant rather than a removal.

**B2 — `validate()` accepts an equality that wedges the write engine.** `cqlite-core/src/config.rs:~841` rejects `memtable_hard_limit < memtable_size_threshold` on the rationale that a lower ceiling "rejects writes before a flush can relieve the memtable" — but accepts **equality**, which wedges identically. Mechanism confirmed: `check_admission` (`write_engine/mod.rs:882-919`) rejects `projected = size + incoming > hard_limit` with **no flush attempt**, and auto-flush fires only after a successful insert. With `hard_limit == threshold == 100`, a memtable at 90 and a 20-byte mutation: projected 110 → rejected; 90 < 100 → no flush; retrying is rejected forever. A livelock with no automatic recovery.

**The cheap fix for B2 was ruled out before dispatch, and is recorded here so it is not re-proposed.** Making the rule strict (`hard_limit > threshold`) does **not** eliminate the wedge: threshold 64 MB, hard limit 65 MB, memtable at 63 MB, one 10 MB mutation → projected 73 MB > 65 MB rejected, 63 MB < 64 MB so no flush → wedged. The wedge exists whenever a single mutation exceeds `hard_limit - current_size` while `current_size < threshold`. A strict-inequality-only change would leave this PR asserting a wedge-freedom property it does not deliver — the same defect class as the finding itself. The honest fix is flush-before-reject at the `&mut self` call site (`check_admission` takes `&self` and cannot flush).

Parity datum: the Node binding builds its config field-by-field and never calls `validate()`, so it has no equivalent ordering bug; adding one would be new scope.

**Consequence, stated plainly:** a fix is a new commit, so the `RESULT: PASS` gate at `c17933cb1` (run-id `/tmp/agent-gate.SSMVQk`) **no longer certifies this branch** and must not be cited as the gate of record for the merged tree. Sequence from here: fix → `--lite` re-cert → re-review → **fresh full gate** under the same #3399 containment → `premerge-assert 3381 <new sha>` → `gh pr merge --auto`.

pmcfadin:
## Roborev: clean — `RESULT: PASS`, `findings: NONE` (round 5, at `6e19efc97`)

```
reviewed-sha: 5b8ae405806738e63fbba96075b50913f5874e96..6e19efc97227f510f90b0696664e01f9686c2b44
job: 37                         model: gpt-5.6-sol
census: 16 files, +2113/-562
tokens: input=1337515 cached=1215744 output=9803
push-assert: PASS               census-check: PASS         code-free: PASS
job-record: PASS                sha-assert: PASS           review-completed: PASS
prompt-content: PASS (14/14 code census paths present)
vacuity-tier1: PASS             vacuity-tier2: PASS
findings: NONE                  roborev-exit: PASS
RESULT: PASS
```

Every vacuity defence passes affirmatively, the reviewed range is `origin/main..HEAD` with both endpoints verified from the job record, and the token accounting (1.34M input / 1.22M cached) is far from the ~18.7k vacuous baseline. So this is a genuine clean review, not an empty one.

### What the five review rounds changed, and what they settled

Rounds 2–4 produced 7 findings. Two are worth calling out because they were resolved by **narrowing a claim and filing the defect** rather than by making the change the reviewer proposed — and round 5 did not re-raise either, which is what makes them settled rather than merely unaddressed:

- **`Config::validate` and the write-path wedge.** The reviewer asked for equality (`hard_limit == memtable_size_threshold`) to be rejected, or for headroom to be enforced. Measuring first showed no inequality between those two knobs can deliver wedge-freedom: the wedge window for a mutation of `m` bytes is `m - headroom` wide, so at *any* positive headroom a large enough single mutation still wedges — including at the default 64 MiB / 256 MiB, where a ~192 MiB mutation does it. The real defect is admission-side (`check_admission` rejects without attempting a flush, and auto-flush only fires after a successful insert), filed as **#3404**. But equality *is* qualitatively worse than any positive headroom — with zero headroom an ordinary 4 KiB write wedges over a 4 KiB window of memtable sizes, a state normal operation passes through routinely — so the rule now requires **strict** headroom and says explicitly what it does and does not guarantee. This PR's own test had asserted equality was "tight but coherent"; that was wrong and is flipped.
- **The bridge never validates.** `WriteEngineConfig::from_config` is infallible by design because it is also the definition of `new`, so returning `Result` ripples into every construction site in the workspace. Not changed here; filed as **#3405** with three options and a recommendation. What *was* wrong and is fixed: the doc claimed "a follow-up issue owns the choice" when no such issue existed.

Two further findings were consequences of the strict-headroom change rather than pre-existing defects — the Node binding accepted equality (it never calls `Config::validate`, so it was briefly the one write path able to build the config core rejects), and the `.pyi` contract understated the bound. Both fixed, with boundary tests in both directions. That the same rule had to propagate to four sites (core rule, Node check, field doc, `.pyi` contract) is itself recorded as #3405's AC4.

One finding corrected a contradiction this PR introduced: `validate()` was judging `compaction.min_threshold` / `max_threshold` even when `auto_compaction` is `false`, though both are documented as ignored in that mode — and `WriteEngine::new` genuinely builds `STCSPolicy::new(min, max, ..)` only inside `if config.auto_compaction`. The docs were right; the rules are now scoped, with a non-vacuity pair asserted separately per rule.

### Certification status

The earlier `AGENT-GATE SUMMARY` `RESULT: PASS` at `c17933cb1` (run-id `/tmp/agent-gate.SSMVQk`) **does not certify this branch** and must not be read as doing so — six commits have landed since. The gate of record is re-running now at `6e19efc97` (run-id `/tmp/agent-gate.YmSrIb`) and its summary will be posted here. `premerge-assert` will be run against `6e19efc97`, not the old SHA.

pmcfadin:
## Gate of record — RESULT: PASS (all 30 components, zero FAIL)

Interrogated rather than trusted, because this fleet produced a live example today of a
*vacuous green* (see #3380/#3367): run-id matches the run this lane bound, `tree-integrity: PASS`,
`tree-start == tree-end`, `commit: 6e19efc dirty: no` equals both the PR head and the SHA
`premerge-assert` will pin, **no `os error 28` in any component log**, and the durations are real work
rather than 0s phantoms (`core-tests 607s`, `python-bindings 734s`, `tooling-tests 760s`,
`memory-budget 634s`, `node-bindings 442s`).

`tooling-tests: PASS (760s)` — the #3399 fixture that OOM-killed this box three times is contained by
the documented 18G `OOMPolicy=continue` scope, with no new global OOM during the run.

One disclosure for completeness: per the analysis on #3367, the `roborev-lints` component contains one
structural assert whose PASS is vacuous fleet-wide (it inspects a scratch copy rather than the real
wrapper). That is a `main`-owned guard defect, unrelated to this diff, and the component's other 740
asserts are unaffected. This lane touches zero `scripts/` files and the scanned files are byte-identical
to `origin/main`.

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.YmSrIb
commit: 6e19efc branch: issue-1697-config-source-of-truth dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-1697/test-data/schemas (checkout-relative)
tree-start: 6e19efc97227 dirty: no digest: ee60a64a108e
tree-end: 6e19efc97227 dirty: no digest: ee60a64a108e
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (5s)
clippy:            PASS (5s)
roborev-lints:     PASS (46s)
core-tests:        PASS (607s)
tombstones-scan:   PASS (44s)
scan-offload-guard: PASS (152s)
work-counters-guard: PASS (92s)
byte-budget-guard: PASS (15s)
memory-budget:     PASS (634s)
integration-tests: PASS (55s)
format-compat:     PASS (39s)
write-tests:       PASS (150s)
cli-tests:         PASS (118s)
flight-query-semantics-oracle: PASS (47s)
python-bindings:   PASS (734s)
node-bindings:     PASS (442s)
delivery-telemetry: PASS (1s)
oom-audit:         PASS (1s)
parity-report:     PASS (11s)
operator-metrics-doc: PASS (26s)
binding-unwind-profile: PASS (7s)
tooling-tests:     PASS (760s)
minimal-build:     PASS (59s)
smoke:             PASS (163s)
logs: /tmp/agent-gate.YmSrIb
summary-file: /tmp/gate-full-1697-r13.txt
RESULT: PASS
```

pmcfadin:
## C — intent audit against #1697's acceptance criteria

Performed by the lane lead directly. Four spawned `spec-auditor` subagents on this issue went idle without returning a packet, and re-delegating the same read-only task a fifth time is a slower way of doing it than doing it. Anchored to the issue body (oracle-driven routing — no OpenSpec change), audited at the certified SHA `6e19efc97`.

### Verdict: **PASS**, with one named residual (low severity, tracked, not blocking)

| AC | verdict | public-surface evidence |
|---|---|---|
| **AC1** — one public source; every live knob originates in `Config`; PR carries the silo→field mapping table | **satisfied** | Mapping table present in the PR body. Origin verified per path: core via `WriteEngineConfig::from_config(&Config, ..)`; CLI via `write_engine_config::resolve` (`main.rs:1293`); Python via `config::config_for_open`; Node via its field-by-field assembly of the public `Config`. |
| **AC2** — the two-memtable-thresholds / two-compaction-configs duplications are gone | **satisfied**, one residual below | The duplicate *fields* remain as internal plumbing, which the issue explicitly permits ("the private structs may remain as internal plumbing, but their values must originate from the public `Config`"). Values originate from `Config` on every production path. |
| **AC3** — single-source behaviour tests green (public knob moves the engine) | **satisfied** | `cqlite-core/tests/issue_1697_config_single_source.rs`, **6** tests, each asserting an *observable engine effect* rather than field equality: `public_memtable_threshold_drives_flush`, `public_memtable_hard_limit_knob_is_load_bearing`, `hard_limit_below_flush_threshold_is_rejected_at_config_time`, `public_compaction_thresholds_drive_stcs`, `public_compaction_max_threshold_caps_merge_width`, `public_auto_compaction_off_disables_compaction`. |
| **AC4** — `scripts/agent-gate.sh` PASS with the SUMMARY pasted | **satisfied** | `RESULT: PASS`, 30/30 components, run-id `/tmp/agent-gate.YmSrIb`, pasted above. |
| **TDD test** — "must FAIL on current main": set the public memtable threshold small → flush at that size | **satisfied** | `public_memtable_threshold_drives_flush` sets `memtable_size_threshold = 4096`, writes ~51KB **outside** a Tokio runtime (so the sync auto-flush path is the one under test), and asserts a `Data.db` appears, `generation()` advances, AND the memtable stays under `16 × THRESHOLD` — i.e. it flushed near 4KB rather than at the private 64MB default. The third assertion is what makes it a real oracle: the first two would also pass if flushing happened for any reason. |

### The residual, stated rather than glossed

`WriteEngineConfig::with_flush_threshold` (`mod.rs:248`) and `with_hard_limit` (`:254`) remain **`pub`** and mutate the private fields directly, so the "one source" property is **conventional, not type-enforced**: an embedder can call `from_config(&cfg, ..).with_flush_threshold(999)` and re-introduce a second source after the bridge.

Measured: **zero production callers.** Every call site (`:1548`, `:1551`, `:1787`, `:2562-2563`, `:2607-2608`, `:2652-2653`, `:2708`) is inside `mod tests`, which is gated `#[cfg(all(test, feature = "write-support"))]`.

Not blocking, for three reasons I'd rather state than assume: the issue's own text permits the private structs to remain as plumbing; removing `pub` methods is an API break that deserves its own change and its own gate rather than a sixth commit here; and no shipped path uses them. Recorded on **#3405**, whose AC4 already covers the adjacent property ("no write path is left able to construct a config `Config::validate` would reject") — the setters are exactly such a path.

### On the two filed defects, and why they are not AC failures

I set out to decide whether #3404 (a config that passes `validate()` can still livelock admission) and #3405 (the bridge never validates) make AC1 a `partial`. My conclusion is that they do **not**, and the distinction is precise rather than convenient: **AC1 is about the ORIGIN of values, not about the completeness of validation.** In both cases the values still originate in the public `Config` — #3405 is that `Config` isn't *checked* at that boundary, and #3404 is an admission-side bug that is independent of where the config came from (it reproduces at the untouched default 64 MiB / 256 MiB). Neither introduces a second source.

They are adjacent, documented in code at the point of the limitation, and filed. Calling AC1 `partial` on their account would conflate "the single-source goal" with "the write path is free of wedges", which was never this issue's scope — the second is #3404's.

I record that reasoning explicitly because it is the one place in this audit where a generous reading and an honest reading could diverge, and the reader should be able to check my work rather than take the verdict.

Issue #2403, opened 2026-07-13
[EPIC] cqlite-trino: latency, throughput, operations — the 0.15 theme


pmcfadin:
Owner-directed 0.15 theme (2026-07-13): get **latency, throughput, and operations right** for the cqlite-trino path. This epic is the milestone's spine; the field metrics template (#2399) is the scoreboard — every lane's acceptance criteria are field numbers, measured round-over-round on the standard loadtest (8 threads × 180s, fixed query set).

**Baseline (round-9, #2367 2026-07-13):** 0.9 qps aggregate, 30 rows/s, p50 9.4s / p99 17.7s, 0 errors; warm point-read 2.1s (≈ 2.2s Trino/JDBC floor); warm LIMIT 6.3–7.7s (fixed ~4-5s setup); cold LIMIT 4m17s (0.14.1 stopgap shipping); 1 of 3 flight pods doing all work.

## Lane 1 — Latency
- #2398 warm-scan fixed ~4-5s setup: localize via phase timings, then fix (suspects: per-query snapshot fan-out, merge_setup)
- #2385(b) lazy Summary-guided index — **design-driven, OpenSpec + Seam 1**: open O(summary), lookup O(log+interval), resident index ~0 (removes the remaining ~210MB/generation + the cold linear parse). Analysis on #2385.
- Trino floor check: is any of the 2.2s `SELECT 1` floor connector-attributable (handshake, metadata round-trips)?
- Target: warm LIMIT-k within ~1s of the floor; cold first-query bounded by one linear parse (then ~0 with lazy index).

## Lane 2 — Throughput
- #2397 replica rotation (GROOMED, connector-side): distribute split primaries across replica owners — the single biggest lever; unblocks 3-node scale-out. Ships as a connector release.
- #2366 full-scan O(partitions) random-read cliff; #2230 KWayMerger whole-partition materialization (both already 0.15)
- count(*) full-ring practicality (currently >10min; retest post-0.14.1, then attack the remainder)
- Target: linear-ish scale with node count on the standard loadtest; p50 under load ≈ single-query warm.

## Lane 3 — Operations
- #2356 per-query-snapshot true closure (durable fd/mmap or rebind-by-inode — partially advanced by #2383's rebind)
- #2306 flush-on-snapshot tiny-SSTable churn; snapshot-leak hygiene check (template item D12)
- #2241 real failover (never yet run in the field); #2384 snapshot-path keyspace/table misparse
[...]

PR #3398, opened 2026-08-28
fix(#1695): enforce query.max_execution_time at the engine chokepoint (AH2)

### LINKED ISSUE #1695: AH2 — Enforce query.max_execution_time via tokio::time::timeout at the engine chokepoint (DECIDED)
Part of epic #1685 (config honesty). Source: `docs/reports/platform-observability-audit-2026-07-01.md` finding AH2 (audit block 6). **DECIDED (owner, capstone ledger #11, 2026-07-01): ENFORCE the timeout** — do not delete the knob, do not re-litigate.

## Problem
`Config.query.max_execution_time` (`cqlite-core/src/config.rs:377`, default 300s) is set by the CLI's `query_timeout_ms` (`cqlite-cli/src/main.rs:1142`) and **enforced nowhere** — `rg` finds no timeout wrapper anywhere in the query path. A runaway query (huge scan, adversarial input) runs forever; the operator's knob is a placebo.

## Fix (effort: M)
1. Wrap query execution in `tokio::time::timeout(config.query.max_execution_time, ...)` at the `QueryEngine::execute` boundary (find it: `rg -n "pub async fn execute" cqlite-core/src/query/`). On elapse, return a proper `Error` variant (add e.g. `Error::QueryTimeout { elapsed, limit }` to the taxonomy if none fits — remember `classify()` is an exhaustive match and will force the observability decision, which is correct).
2. Streaming queries: the timeout must cover time-to-first-batch AND be documented for the streaming surface (either the whole stream is bounded, or the bound applies per poll — pick the simpler contract: bound the whole `execute` future including stream setup, document that streamed row consumption afterwards is not bounded, and note per-batch bounds as a possible follow-up).
3. A `Duration::ZERO`/absent value should mean "no timeout" — decide the sentinel, document it in the config docs, and validate it in `Config::validate()` (a LIVE knob finally worth validating — coordinate with AH3).
4. Ensure cancellation is clean: the streaming machinery is drop-safe (audited), so dropping the timed-out future must release resources — add the test.

### Do NOT
- Do not enforce the timeout deep in the scan loop with ad-hoc clock checks — one wrapper at the engine chokepoint.
- Do not let the timeout error be indistinguishable from corruption — distinct variant, distinct `classify()` category.

## Acceptance criteria
- [ ] `max_execution_time = 1ms` + a large-scan fixture → the query returns the timeout `Err` (COMPLETES today — the red test).
- [ ] Default 300s behavior unchanged for normal queries; zero/none sentinel documented + validated.
- [ ] Post-timeout drop leaks nothing (producers exit — reuse the drop-safety test pattern).

The 1ms-timeout-vs-large-scan test.

pmcfadin:
Closes #1695.

## What this does

`query.max_execution_time` was a configured-but-dead knob: on `origin/main` it appears only as a field
declaration, a default and a test setter, with no `tokio::time::timeout` anywhere under
`cqlite-core/src/query/`, and `Error::QueryTimeout` did not exist so the variant was unconstructible.
This wires it with **ONE** `tokio::time::timeout` at the engine chokepoint, per the issue's DECIDED approach.

- New `cqlite-core/src/query/engine/deadline.rs` holds every **bounded** public entry point — `execute`,
  `execute_streaming`, `execute_with_params`, `execute_prepared`. `engine.rs` keeps only the unbounded
  `*_inner` bodies and `execute_unbounded` is `pub(crate)`, so each of the 7 entry points is bounded
  **exactly once**.
- New `Error::QueryTimeout` with its own `classify()` category (never `corruption`); the CLI maps it to
  **exit code 5**. Both bindings' exhaustive-variant audits updated.
- Cooperative `ScanCancel` checkpoints in the scan path, so the single engine-level timeout can fire
  against otherwise non-yielding scan work. **No ad-hoc clock reads were added in the scan loop** — the
  issue's `Do NOT` clause. `scan_cancel.rs` contains no clock reads outside its own tests.
- Per-call `merge_cancel` abandonment for the three materializing `spawn_blocking` merges, since tokio
  cannot cancel a `spawn_blocking` closure by dropping its `JoinHandle`.
- **The CLI bounds its own consumption loops** — both `commands::collect_query_result` (`cqlite query`)
  and the four `collect_chunk` loops in `commands::export` (`cqlite export`). This is not decoration:
  without it the feature does not work where it is configured; see §2.

### Two delegated calls (the issue body left these to the implementer)

1. **`Duration::ZERO` means "no timeout"** — `bound` short-circuits `is_zero()` and awaits with no timer
   rather than handing zero to `timeout()`, which would fire instantly. Documented, validated in
   `Config::validate()`, pinned by `zero_limit_is_unbounded`.
2. **The bound covers the whole `execute`/`execute_streaming` future including time-to-first-batch.**
   Post-return row consumption is not bounded; documented, with per-batch bounds as a follow-up.

### 1. ⚠️ BEHAVIOR CHANGE: the CLI's real default budget is 30s, not 300s

**This is a behavior change with a regression path, stated in those words at the lead's instruction so
nobody discovers it from a broken scan.**

| | Before this PR | After this PR |
|---|---|---|
| A CLI query taking 45s | **succeeds** | **fails** with `Error::QueryTimeout`, **exit code 5** |
| `performance.query_timeout_ms` | set to `30000` by default, **enforced nowhere** | enforced |
| Escape hatch | n/a | **`query_timeout_ms = 0`** = unbounded (documented sentinel) |

Anyone running a CLI query or `cqlite export` that legitimately takes longer than 30 seconds must raise
`performance.query_timeout_ms` or set it to `0`. This belongs in release notes.

`origin/main` already ships `PerformanceConfig::query_timeout_ms = 30000` and already maps it to a
`Duration` — enforced nowhere. Enforcing the knob makes CLI queries bounded at **30s**, so a legitimate
45s CLI scan that used to succeed now returns `QueryTimeout` (exit 5). Escape hatch:
`query_timeout_ms = 0`.

AC2's "default 300s unchanged" refers to the **core** `QueryConfig` default, which is unchanged; the CLI
ships a 10x tighter value of its own. Both cannot hold — you cannot enforce a knob and also leave its
behaviour unchanged, and enforcement is the decided intent.

**Status of this decision: RULED AND CLOSED.** Raised as `DI1695-F1`; the coordination lead ruled it a
product call belonging to the OWNER. The owner ruled on 2026-08-28
([issuecomment-5448884756](https://github.com/pmcfadin/cqlite/issues/1695#issuecomment-5448884756)):
**KEEP 30s — option (a) confirmed, ship it.** The CLI default is the value the CLI has documented all
along, and honouring an operator's stated budget is the least-surprising reading of "stop being a
placebo". Nothing in the implementation changed as a result; the two conditions the owner attached are
the behaviour-change table above and the AC2 scoping paragraph below, both of which this body carries.

### 2. The CLI's query path needed a second bound, at a second layer — the riskiest part of this PR

Roborev round 5 found that the feature did not actually work where it is configured. The CLI routes
queries through `execute_streaming` and then drives the scan in a `next_async()` loop
(`commands/query.rs`; `export.rs` likewise). The engine's chokepoint wrapper bounds the
`execute_streaming` FUTURE — setup and time-to-first-batch — and structurally cannot bound what
happens after that future returns, which is where a CLI query spends essentially all of its time.
Since the CLI is the **only** consumer that sets `performance.query_timeout_ms`, #1695's stated
problem was left standing for the caller it was written about, while `Database::execute` — which every
one of my tests used — was correctly bounded.

`collect_query_result` now bounds consumption. Setup is deliberately awaited **outside** the CLI
timeout so a setup elapse belongs to, and is recorded by, the engine; the CLI applies only the
REMAINING budget to consumption. The engine entry point is still bounded exactly once.

**Flagging this as the highest-risk part of the diff rather than presenting it as settled.** It is new
surface added late, and rounds 6 and 7 both found real defects in it — a race that swallowed the
engine's telemetry, and an expired deadline that did not apply to an immediately-ready stream. Both
are fixed and pinned, but the honest summary is that this is where the review kept finding things.
Bounding per-batch inside the core iterator would fix it for every consumer and is the better shape;
the issue explicitly deferred per-batch bounds, so this takes the local fix at the layer that owns the
loop.

### 2b. What the bound still does NOT cover, and a finding raised in three consecutive rounds

`ORDER BY` sorting, aggregation and projection run inside `select_executor`'s **non-async**
`execute_sort` over an already-materialized `Vec`. A `tokio::time::timeout` can only elapse when its
future goes `Pending`, so a query whose scan finishes inside the budget and then sorts a large result
set can overshoot, uninterrupted. `Vec::sort_by` is one uninterruptible call.

Resolved as documented scope on the issue body's own precedent for the analogous streaming case:
*"pick the simpler contract: bound the whole `execute` future … document that streamed row consumption
afterwards is not bounded, and note per-batch bounds as a possible follow-up."*

**Roborev raised this same line in rounds 2, 3 and 6** — three consecutive rounds, with a different
and progressively better argument each time. My position did not change and is documented in code
under "What the bound does NOT cover". Recording the recurrence count explicitly so the reviewer's
persistence is visible and the call is easy to overturn: if you want post-scan stages bounded, say so
and it becomes its own issue.

### 3. A pinned contract changed: an off-stride `ScanCancel::checkpoint` is now a no-op

An earlier revision of this branch had `checkpoint(tick)` poll the cancellation flag on **every** call.
Three callers had hand-rolled `if tick & 0xFF == 0 { check()? }`, so that silently **raised** their
cancellation cadence and added a relaxed atomic load plus an `.await` point to a per-entry decode loop.
That cadence was chosen deliberately by #2264/#2346 and #1695 has no mandate to retune the read hot path,
so both the poll and the yield now sit on the stride boundary, reproducing the prior cadence exactly.

`full_index_scan.rs` polled on **every** iteration before this branch (one real index random read per
iteration is already coarse), so it uses `checkpoint_now` — a blanket stride-gate would have regressed
that site. Two different prior cadences existed; that is why all four call sites had to be read.

The stride doc no longer claims "neither cancellation latency nor scan throughput changes measurably" —
no measurement was taken, and #2403's release theme is exactly where an unmeasured no-change claim is the
wrong thing to assert. It now states the checkable thing: equivalence to the code it replaced.

## Verification

- **Gate of record**: the full `AGENT-GATE SUMMARY` is posted as a PR comment by `flow-closer`.
- **`file-size`**: run with the sanctioned `CQLITE_ALLOW_FILE_GROWTH=1`. **Disclosed explicitly, because
  that opt-out prints a plain `PASS` with no marker in the SUMMARY** (unlike #2078's
  `missing-fixtures: OPT-OUT`) — that gap is filed as #3402.
  **Corrected 2026-08-28 against the gate's own diagnostic, then RECOMPUTED after the rebase onto `origin/main` at HEAD `f376f1d9`.** An earlier
  revision of this section understated the growth (it was written several rounds back and said "nothing
  was newly pushed over threshold", which is not true at this HEAD). The actual list, verbatim from the
  `file-size` component:

  | file | before → after | limit | newly over? |
  |---|---|---|---|
  | `cqlite-cli/src/commands/export.rs` | 664 → 851 | 800 | **yes** |
  | `cqlite-core/src/storage/sstable/reader/data_access/bti.rs` | 760 → 827 | 800 | **yes** |
  | `cqlite-core/src/storage/sstable/generation_merge.rs` | 770 → 806 | 800 | **yes** |
  | `cqlite-core/src/query/prepared.rs` | 923 → 1007 | 800 | no (already over) |
  | `cqlite-core/src/error.rs` | 878 → 913 | 800 | no (already over) |
  | `cqlite-core/src/config.rs` | 924 → 952 | 800 | no (already over) |
  | `cqlite-core/src/query/select_executor/mod.rs` | 2552 → 2573 | 800 | no (already over) |

  Rationale, unchanged for the already-over files: an `Error` variant cannot be added without touching
  the taxonomy, nor a sentinel documented/validated without touching config, and splitting a 1007-line
  `prepared.rs` or the error taxonomy inside a P1 timeout fix is out of scope (#1116).
  The three **newly-over** files are the later review rounds' work: `export.rs` +187 is the CLI
  consumption bound (round 5's finding that the feature did not work where it is configured),
  `bti.rs` +67 is the BTI scan-cancel-on-consumer-departure fix, `generation_merge.rs` +36 is the
  `is_closed()` correctness guard (round 8). Each is a correctness fix landed in the file that owns the
  behaviour; extracting three new modules across two crates while closing a P1 is the textbook
  out-of-scope split. **Recorded as a debt, not waved away** — noted on #1116 with these three names so
  the campsite work is trackable rather than implicit.
  **Campsite credit**: `engine.rs` 1371→1336 (extracted `deadline.rs`) and `cqlite-cli/src/main.rs`
  1425→1418 (extracted `core_config.rs`) both shrank.
  *(`config.rs`'s baseline moved 1133→924 in the rebase because #1697 split that file on `main`; this
  PR's own delta there is +28 either way. Every other row is unchanged by the rebase — re-measured,
  not assumed.)*
- **roborev**: eleven rounds via `scripts/flow/roborev-review.sh --agent codex --model gpt-5.6-sol`.
  All were genuine (0.8M–3.3M input tokens, 0.7M–3.1M cached, vs the ~18.7k/0 vacuous baseline; every
  integrity key PASS including `prompt-content` at full census).
  **`prompt-content: FAIL` on the later rounds is #3312, not a vacuous review.** #3312 (closed) is
  "prompt-content FALSE-FAILs every large-diff review — roborev delivers the diff by snapshot PATH, and
  the check only reads inline headers", and CLAUDE.md records the same accepted cost: a snapshot-delivered
  diff and a review that received nothing are identical to the machine. Corroborated two ways rather than
  assumed: token accounting (1.97M–3.18M input / 1.82M–3.00M cached, against a ~18.7k/0 vacuous baseline),
  and the findings themselves — round 16 named `bindings/python/src/error.rs:364` and
  `bindings/node/src/error.rs:476`, file:lines that exist only in this diff. No waiver was sought or
  needed.
  **Round 10 is recorded as ABORTED, not passed**: its `sha-assert` FAILED because `origin/main` moved
  mid-review (`2bde26a7c` → `835fb2276`), so its reviewed range no longer matched `origin/main...HEAD`.
  Its two findings were acted on — they concern code this branch added and are base-independent — but an
  unverifiable range is not a review, so the branch was rebased onto the new main and re-reviewed
  against a verified range. Only the re-run counts.
- **`rust-reviewer`**: `0 blockers, 4 nits` on the cancellation delta, confirming the token swap is
  sound. **The nit bodies were never delivered** across two requests — recorded rather than invented; see
  the follow-up.

### roborev findings and their disposition

| Round | Finding | Disposition |
|---|---|---|
| 1 | detached `spawn_blocking` merge survives the timeout | **fixed** — per-call `merge_cancel` |
| 1 | `error_queries` could exceed `total_queries` | **fixed** — `bound_tracked` |
| 1 | admission permit leaked | **declined** — misreading; the permit is deliberately moved into the closure (#2063) so it is held until detached work terminates |
| 2 | a `Ready` result past the deadline is accepted | **declined** — `tokio::time::timeout`'s specified semantics; erroring would discard a correct completed result without shortening the wait already taken |
| 2 | token not passed into the merger's own readers | **fixed** |
| 2 | stale "prepared execute is unbounded" doc | **fixed** |
| 3 | post-scan stages unbounded (High) | **documented scope** — §2b |
| 3 | prepared-route timeouts absent from telemetry | **documented** |
| 3 | checkpoint cadence raised / unmeasured claim | **fixed** — §3 |
| 4 | python `QueryTimeout` fell through to `CqliteError` while a comment claimed otherwise | **fixed** — mapped to builtin `PyTimeoutError`, deviating from the suggested `QueryError`; **but see the unverifiable-mapping caveat below** |
| 4 | `checkpoint_now` added a per-partition yield (my round-3 regression) | **fixed** — third cadence method `checkpoint_polled` |
| 4 | `#[serial]` did not exclude sibling tests that arm the same global probe | **fixed** — real vacuity hole; siblings serialised |
| 4 | `main.rs` uses a binary-local `core_config`, not the lib surface the test drives | **claims corrected**, unification deferred — pre-existing CLI-wide duplication |
| 5 | the CLI's streaming path was unbounded after setup | **fixed** — §2 |
| 5 | expired-budget test assumed the clock advanced | **fixed** — absolute deadline via `bound_tracked_until` |
| 6 | the CLI bound raced the engine's and swallowed its telemetry | **fixed** — setup now awaited outside the CLI timeout |
| 6 | the positive CLI test could pass with the bound deleted | **fixed** — deterministic, discriminating, always-running replacement |
| 6 | post-scan stages unbounded (repeat of 3) | **documented scope** — §2b |
| 7 | `timeout_at` polls before checking, so an expired deadline missed an immediately-ready stream | **fixed** — expired deadline rejected up front; RED-verified |
| 7 | `Instant + Duration` could panic on operator config | **fixed** — `checked_add`, `None` means unbounded |
| 8 | the streaming producer only checked closure when SENDING, so a no-match scan ran on | **fixed** — `Sender::is_closed()` at the top of the loop |
| 9 | a SECOND producer one layer up (query executor) had the same hole, and holds the storage receiver so the layer-below check cannot help it | **fixed** — `tx.is_closed()` per batch |
| 9 | `cqlite export` was still entirely unbounded | **fixed** — see below |
| 9 | CLI consumption timeouts absent from telemetry (3rd raise) | **deferred with reason** — see the caveat below |
| 10 | checking before a chunk pull does not bound a pull that stays PENDING | **fixed** — the await itself is now wrapped |
| 10 | export's budget clock started AFTER setup, granting consumption a fresh full budget | **fixed** — separate `budget_start` before setup |
| 11 | the query-layer closure check ran only AFTER `recv()` returned, so a stalled batch left a timed-out scan running | **fixed** — `biased` select racing `tx.closed()` |
| 11 | CLI consumption timeouts absent from telemetry (4th raise) | **deferred with reason** — see the caveat above |
| — | a THIRD unfixed instance of the same class, in `per_row_scan_stream` | **found by my own sweep, not reported** — see below |
| 12 | `batched_scan_stream` (the surface query execution actually uses) had the same hole; my sweep missed the file | **fixed** |
| 12 | the windowed forwarder, which my sweep had declared SAFE, is exposed compositionally | **fixed** — both arms race `tx.closed()` |
| 12 | BTI materializes the whole table under a shared token nothing trips on stream drop | **deferred** — see the BTI decision note below |
| 13 | the targeted / multi-targeted branches never observed closure (large `IN`, tombstone builds) | **fixed** — ONE race at the spawn covers all branches |
| 13 | markerless `execute_with_params` lost its `query.execute` span | **fixed** — span restored on the entry point |
| 14 | BTI (2nd raise) | **deferred, with the decision now stated** |
| 14 | CLI consumption telemetry (5th raise) | **FIXED — my objection was wrong; see above** |
| 15 | BTI (3rd raise) | **FIXED** — the scan future is now raced against consumer departure |
| 15 | prepared-handle timeouts absent from the error stream | **FIXED** — same disjointness argument as the CLI |
| — | `QueryTimeout` had no row in main's new authoritative FFI error table | **found by the rebase** — registered, and it closed the "reviewed not tested" caveat |

Rounds 5–7 were largely reviewing the fixes from rounds 4–6, which is why the count did not fall
monotonically. The concentration is in the CLI-layer bound (§2), as noted.

Two corrections to roborev's framing, both verified before acting:

- On the merger token: the displaced tokens were `ScanCancel::default()` (what plain `KWayMerger::new`
  installs — `new` IS `new_cancellable` with that default, identical other params) and one literal
  `ScanCancel::new()`. Neither was reachable by any caller, so the change **removes no existing
  cancellation**. It stays the per-call token, never the shared `SSTableReader::scan_cancel`, which would
  cancel other queries' scans.
- roborev said that gap left "detached blocking work and its permit alive". It did not: #2361 already
  closes the channel, trips the readers' token and joins the threads on merger drop, so **AC3 held before
  that fix**. The gain is promptness — an abandoned merge now stops mid-`step()` instead of finishing the
  partition it was decoding — not a leak. The PR does not claim otherwise.

And one correction to MY OWN reasoning, which a reviewer will otherwise hit in the commit history.

In round 1 I excluded `stream_generations_for_read` from the cancel guard, arguing its
`out_tx.blocking_send(..).is_err()` already retires it when the consumer drops. **That was incomplete,
and round 8 caught it.** A send-failure check is only reached when a partition actually yields a row,
and three paths in that loop `continue` without sending — a partition below `start_key`, one above
`end_key`, and one whose rows are all tombstoned or TTL-expired. So a range scan matching nothing, or a
no-match full scan, performed not one send and therefore not one check, and walked the whole remaining
table after `QueryTimeout` had been handed back. Fixed with `Sender::is_closed()` at the top of the
loop, which needs nothing plumbed in and strictly dominates the send check. The earlier reasoning was
right about the common case and wrong about the case that matters.

### CLI consumption timeouts DO reach telemetry — and how I got this wrong for five rounds

Roborev raised this in five separate rounds. I declined it each time, on the grounds that exposing a
"record a timeout someone else detected" entry point would create a second recording path competing with
the engine's. That objection was wrong, and I never checked its premise: `observability::record_error` is
**already `pub`**, and it is the very sink `AdvancedQueryEngine::bounded` uses. There was no second
mechanism and no new API — one sink, two callers, observing DISJOINT events (the engine records a SETUP
elapse; the CLI records a CONSUMPTION elapse), so nothing double counts.

Both `cli.query.collect` and `cli.export.collect` now record, so a consumption elapse reaches
`cqlite.errors.total{category="timeout"}`. Since consumption is where a CLI scan spends its time, this is
the difference between dashboards reflecting operator-visible timeouts and reading zero while operators
see them.

Recording the process failure alongside the fix: I had a plausible-sounding architectural objection and
repeated it across five rounds without testing it. Reviewer persistence was the signal, and I treated it
as noise.

**The narrower residual, stated precisely:** the engine's `error_queries` counter still does not move for
a consumption elapse. That counter is private accounting for work the ENGINE executed, and the CLI's
consumption phase is not an engine query execution — folding it in would misreport the counter's meaning.
`Database::stats()` therefore reports setup elapses only.

### The cancellation class: what it was, and how my own sweep of it fell short

Rounds 8–13 all reported ONE class: a producer whose only stop signal is a failing send, in a loop that
can do real work WITHOUT sending. After round 11 I generalized the test — *can this loop do work without
reaching a send?* — and swept every producer I could enumerate. The sweep found a real unreported
instance, and also produced one wrong verdict and missed one file. Corrected table:

| Producer | Skip paths | Outcome |
|---|---|---|
| `generation_merge` streaming producer | 3 (range ×2, all-tombstoned partition) | fixed, round 8 |
| query-layer `select_executor::streaming` fallback | 4 (row build, predicates, per-partition limit, OFFSET) | fixed, rounds 9 + 11 |
| its targeted / multi-targeted branches | large `IN`, tombstone builds | fixed, round 13 — ONE race at the spawn |
| `per_row_scan_stream` non-stitching | 4 (table-id, range ×2, tombstone) | **found by my sweep** |
| `batched_scan_stream` non-stitching | 4, plus `BATCH_EMIT_ROWS` gating the send | **MISSED by my sweep** — round 12 |
| `scan_stream_forwarder` (both arms) | none | **I wrongly called this SAFE** — round 12 |
| `scan_stream_windowed` | none — every row joins a batch and is sent | safe |
| BTI `bti_scan_with_metadata` | 3 (range ×2, tombstone) | **deferred** — decision below |

Four things worth taking from this rather than just the fixes:

**The enumeration was the weak link, not the predicate.** I listed producers by grepping
`spawn_blocking|tokio::spawn` per directory. `batched_scan_stream.rs` is `#[path]`-included so it never
appeared — and it is the surface query execution actually uses.

**Per-stage reasoning cannot see a whole-chain property.** I cleared the forwarder because a stage parked
in `recv().await` burns nothing. True in isolation; irrelevant compositionally — upstream's filters reach
no send either, so under a no-match scan nothing anywhere sends and the chain idles while upstream burns
to EOF.

**Guarding branch-by-branch is what spread this over five rounds.** Round 13's fix is one race at the
spawn covering every branch of the producer, present and future. It does not replace the in-loop checks:
a race fires only at an await boundary of the future it wraps, so a long synchronous stretch is still only
interruptible by that branch's own check. Coverage from the spawn, promptness from the loops.

**Granularity differs per site, with the reason at each site.** A bounded disk read always returns to a
plain `is_closed()`; an open-ended wait on another producer needs `select!` on `tx.closed()`.

### Cancel-safety of the five new `select!` arms — verified, because a lost batch would be silent wrong output

This diff adds `select!` to producer paths in five places. A future that is NOT cancel-safe, dropped
mid-poll by a `select!`, can consume a message and lose it — which here would mean **rows silently missing
from a query result**, the worst failure class in this repo and one no cancellation test would catch. I did
not check this when writing the code; it is verified now, with the evidence rather than an assurance:

| arm | future | why it is cancel-safe |
|---|---|---|
| `select_executor::streaming` | `scan_stream.recv()` → `JoinedStream::recv()` (`BatchedScanStream` = `JoinedStream<Vec<…>>`) | #3106 made it so deliberately and documented the reasoning inline: the inner `rx.recv().await` is tokio's cancel-safe recv; `&mut JoinHandle` is awaited without taking ownership so "a cancellation here drops only this borrow"; and there is "no `.await` between observing the outcome and recording it" |
| `scan_stream_forwarder`, per-row arm | `batch_rx.recv()` (tokio `mpsc`) | tokio documents `Receiver::recv()` as cancel-safe — no message is lost if the future is dropped before completion |
| `scan_stream_forwarder`, batched arm | same | same |
| `bti::stream_bti_scan` | the `bti_scan_with_metadata` future | dropping it IS the intended cancellation; it returns a `Vec` by value, so there is no partially-consumed channel state to lose |
| `select_executor::execute_streaming` spawn | the whole producer future | same — dropping releases readers, buffers and the storage stream |

`tx.closed()` is cancel-safe by nature. So no arm can drop a consumed-but-undelivered item.

### The BTI deferral, stated as a decision rather than a difficulty

Roborev raised BTI twice, so here is what I verified rather than guessed:
`bti_scan_with_metadata_cancellable` **already exists and already takes a `&ScanCancel`** — plumbing is
not the gap. The gap is that the caller has no token it may legitimately pass:

* a **clone** of the reader's token cancels OTHER queries scanning that reader (clones share state) — the
  trap the materializing merges avoided with a per-call token;
* a **fresh per-call** token tripped by a watcher on the sender's `closed()` handles the timeout but
  silently DROPS #2264's Flight cancellation, which trips the reader token *without* necessarily dropping
  the receiver — sending client disconnects back to the ~1–2 min transport backstop.

Losing existing cancellation to add new cancellation is not a fix. What is needed is one small API
decision — a token meaning "either the reader's OR this stream's closure" — and that composition is the
same primitive the deferred `JoinedStream` / `KWayMerger::step()` work needs. Not improvised at round 14
in a 30-file diff: three partial fixes here is how the gap got scattered, and a fourth guessing at token
composition would be the worst of them.

Consequence, plainly: on a **BTI (`da`) table**, a timed-out streaming query's detached producer keeps
materializing and sorting until its scan completes. It terminates and nothing leaks — but it burns CPU and
memory for an abandoned request. Every non-BTI producer in the read path now observes closure.

### The bindings' error identity now lives in main's authoritative FFI table

This changed under the branch, for the better. Main's #1451 landed "one authoritative FFI error table —
Python and Node stop disagreeing on error identity", replacing the per-binding exhaustive-variant audits
my earlier commits had edited. Rebasing conflicted, and the right resolution was to drop my approach and
re-implement on main's:

```
QueryTimeout => { py: Timeout, code: "TIMEOUT", category: Query,
                  recoverable: false, prefix: Some("TimeoutError") }
```

`py: Timeout` and `code: "TIMEOUT"` put a budget elapse on the SAME identity as its sibling `Timeout` on
both surfaces, so `except TimeoutError:` in Python and a `TIMEOUT` check in JS both catch it — the
cross-binding agreement #1451 exists to enforce, and it vindicates the earlier choice of `PyTimeoutError`
over the suggested `QueryError`. `category: Query` is a deliberately different axis: a budget elapse is a
query-execution failure, distinguishable from corruption per #1695.

Three of main's guards did real work, which is worth recording as an argument for that style of guard:

* `variant_of` is exhaustive, so **the build failed** until the variant had a row — the registration could
  not be forgotten.
* `table_agrees_with_core_category_and_recoverable` **passed**, independently confirming `category: Query`
  and `recoverable: false` match `Error::classify()` and `is_recoverable()` — the cross-table
  contradiction I was worried about, checked by machine rather than by my reading.
* `row_count_is_pinned` **failed 37 → 38** exactly as designed, forcing the new row's identity to be a
  reviewed act rather than an invisible one.

**This also retracts a caveat this PR carried earlier.** I had reported the python mapping as "reviewed,
not tested", because `cargo test -p cqlite-py` cannot link libpython and the binding exposes no way to
provoke a timeout. Main's `FfiErrorVariant::sample_error()` exists for exactly that class ("a variant that
no test query can provoke (`Timeout`, `Memory`, …)"), and `cqlite-core/tests/ffi_error_contract_table.rs`
DOES run under cargo test. With a sample added the mapping is genuinely exercised, and
`query_timeout_shares_the_timeout_identity_on_both_surfaces` pins it. The follow-up asking to make
`cqlite-py`'s Rust tests executable still stands on its own merits, but it is no longer what stands
between this mapping and coverage.

### Known gap: prepared-handle timeouts are invisible to telemetry

`PreparedQuery` carries only its executor and its budget — no stats or observability handle — so a
timeout taken via `PreparedQuery::execute` is **enforced** identically but never increments
`error_queries` or `cqlite.errors.total{category="timeout"}`. Worth stating plainly rather than
softening: #1695 comes from a platform-observability audit, and this ships an observability blind spot on
a public route. Closing it needs a stats handle threaded from `prepare()` into a struct already over the
file-size threshold, so it is a follow-up. Documented at the chokepoint; my earlier doc saying "bounded
on either route" was corrected, since only one route is observable.

### Tests

`cqlite-core/tests/issue_1695_query_timeout.rs` (+604) covers all four ACs, the CLI exit-code-5 wiring
and the `ZERO` sentinel. Fixture roots resolve per-table via `sstables_root_for_table` with per-case
fail-closed assertions for committed fixtures — no suite-wide `assert!(ran > 0)` (#3220). No wall-clock
threshold asserts in the correctness path:

- `a_dropped_multi_generation_read_abandons_its_blocking_merge` removes the race **by construction**: a
  `max_blocking_threads(1)` runtime whose sole blocking thread the test occupies, the future dropped while
  the merge closure is still *queued*, then released — so the closure starts with the flag already
  tripped. Anti-vacuity guard asserts `armed() == 1` before `abandoned() >= 1`. RED-verified: with the
  loop check neutered it fails `armed=1, abandoned=0`.
- `an_expired_budget_keeps_error_queries_within_total_queries` and
  `a_completed_query_is_counted_exactly_once` assert public `Database::stats()` deltas and guard **both**
  directions. RED-verified: pre-fix reports "1 error(s) against 0 total quer(ies)".
- `the_materializing_merges_pass_their_token_into_the_merger` is a structural pin, because "a reader stops
  mid-step" is only observable as a latency difference and a wall-clock assert in the correctness path is
  forbidden. It counts tokens actually handed out rather than asserting the **absence** of the inert
  constructor — the first draft did the latter and was defeated by a comment naming that constructor two
  lines from the call site.
- `checkpoint_checks_and_yields_at_the_stride_only` pins the changed cadence contract, including that an
  off-stride checkpoint is a no-op even once cancelled, and says why that is the contract.

## Follow-ups (proposed as `DI1695-F1`, none merge-blocking)

1. The CLI 30s-vs-300s default decision (§1) — the one item that wants a call before the next release.
2. **A composable `ScanCancel` — "either the reader's token OR this stream's closure"** — then use it
   for the three places that need exactly that primitive:
   (a) **BTI** `bti_scan_with_metadata_cancellable` (see the decision note above; the function and its
   token parameter already exist, only the token to pass is missing);
   (b) a **drop-tripped guard inside `JoinedStream`**, which has no `Drop` impl today so dropping
   detaches its producer;
   (c) cancellation **inside a single long `KWayMerger::step()`** on the streaming path, whose readers
   still hold an inert `ScanCancel::default()`.
   This is one primitive and three call sites, not three separate projects. It is shared streaming
   machinery Flight also uses, which is why it was not improvised late in a diff whose last six review
   rounds were spent on newly added surface.
3. **Carry the deadline itself into the core streaming producer** (the per-batch bound the issue
   deferred), which would additionally let CORE own consumption timeouts rather than the CLI, and move
   the engine's `error_queries` counter with them.
4. Bounding post-scan stages — interruptible/chunked `ORDER BY`, aggregation, projection (§2b; roborev
   raised it three times).
5. Prepared-handle `error_queries` — **narrower than when first listed**. The error-stream half was
   fixed during round 15: `PreparedQuery::execute` / `execute_with_context` now record a timeout on the
   engine's own public sink, so those routes ARE visible to
   `cqlite.errors.total{category="timeout"}`. What remains is only the engine's `error_queries` counter,
   which is private accounting for queries the engine itself executed, and moving it is part of (3).
6. Unify `cqlite-cli`'s binary onto the library modules (`config`/`core_config`/`error`), and add a
   `CARGO_BIN_EXE_cqlite` spawn test asserting exit code 5 end to end. Today the binary compiles its
   own copy of every module, so the exit-code evidence sits at the library-copy level.
7. **Expose `query.max_execution_time` on the Python binding**, which currently has no way to set it, so
   a Python caller cannot bound a query at all — and, separately, make `cqlite-py`'s Rust `#[test]`s
   executable (they never link libpython today, so that whole file's assertions are inert).
   **Correction to an earlier draft of this list:** these are NO LONGER what stands between the
   `PyTimeoutError` mapping and coverage — main's `sample_error()` plus
   `cqlite-core/tests/ffi_error_contract_table.rs` test it today. Both items still stand on their own
   merits (a Python user cannot set the budget; a file of inert tests reads as coverage it does not
   provide), but not as blockers for that mapping.
8. Four `rust-reviewer` nits on the cancellation delta whose bodies were never delivered; re-run a
   review on the merged code to recover them. Recorded as unretrievable rather than guessed at.
9. Two gate `file-size` diagnostic gaps — **filed, per the lead's ruling on `DI1695-F1`**: #3401 (FAIL
   writes no diagnostic log, so the offending file must be recomputed by hand) and #3402
   (`CQLITE_ALLOW_FILE_GROWTH=1` leaves no marker in the SUMMARY, unlike #2078's `OPT-OUT` precedent).
   Both checked against existing issues before filing, and left unlabelled for priority.

### DIFF
--- a/bindings/node/src/error.rs
+++ b/bindings/node/src/error.rs
@@ -495,6 +495,14 @@ mod tests {
             // CommitLog reader (#2389) — not bound yet (v1 is library+CLI only).
             | Error::UnsupportedCommitLogVersion { .. }
             | Error::CorruptCommitLogFrame(_) => "PARSE",
+            // Query execution budget elapsed (issue #1695): the SAME `TIMEOUT` code as
+            // its sibling `Timeout`, so a JS caller checking for `TIMEOUT` catches
+            // both. Deliberately NOT `QUERY`, even though its `ErrorCategory` IS
+            // `Query` — #1451's whole point is that codes are decided BY VARIANT, so
+            // the classify category and the JS code are separate axes. Python maps the
+            // same variant to the builtin `TimeoutError`, which is the cross-binding
+            // agreement the shared table exists to enforce.
+            Error::QueryTimeout { .. } => "TIMEOUT",
             Error::Configuration(_) | Error::InvalidReadPath { .. } => "CONFIG",
             Error::InvalidInput(_) | Error::InvalidState(_) | Error::InvalidOperation(_) => {
                 "INVALID_INPUT"
--- a/bindings/python/src/error.rs
+++ b/bindings/python/src/error.rs
@@ -371,6 +371,13 @@ mod tests {
             Error::CqlParse(_) => PyExceptionClass::Parse,
             Error::Configuration(_) | Error::InvalidInput(_) => PyExceptionClass::Value,
             Error::Timeout(_) => PyExceptionClass::Timeout,
+            // Query execution budget elapsed (issue #1695): the SAME builtin
+            // `TimeoutError` as its sibling above, so `except TimeoutError:` catches
+            // both. Its core `ErrorCategory` is `Query` — a separate axis from the
+            // Python class, which is why this is not `PyExceptionClass::Query`. Node
+            // gives the same variant code `TIMEOUT`, which is the cross-binding
+            // agreement the shared table (#1451) exists to enforce.
+            Error::QueryTimeout { .. } => PyExceptionClass::Timeout,
             Error::Memory(_) => PyExceptionClass::Memory,
             Error::InvalidState(_) => PyExceptionClass::Runtime,
             Error::Cancelled => PyExceptionClass::Cancelled,
--- a/cqlite-cli/src/commands/export.rs
+++ b/cqlite-cli/src/commands/export.rs
@@ -113,6 +113,13 @@ pub async fn export_data(
         _ => StreamingConfig::for_text_formats(),
     };
 
+    // Issue #1695: the budget clock starts BEFORE setup. An earlier revision derived
+    // the remainder from `start_time`, which is captured AFTER `execute_streaming`
+    // returns — so consumption was handed a fresh FULL budget and setup cost nothing
+    // against it (roborev round 10). `start_time` still exists for the export
+    // statistics, which deliberately report consumption time only.
+    let budget_start = Instant::now();
+
     // Execute the query with streaming (Issue #280 - true end-to-end streaming)
     let mut result_iter = database
         .execute_streaming(&query, config.clone())
@@ -164,7 +171,7 @@ pub async fn export_data(
         // reading of an astronomically distant deadline.
         tokio::time::Instant::now().checked_add(
             export_budget
-                .checked_sub(start_time.elapsed())
+                .checked_sub(budget_start.elapsed())
                 .unwrap_or_default(),
         )
     };
@@ -200,12 +207,14 @@ pub async fn export_data(
                     break;
                 }
 
-                check_export_deadline(export_deadline, start_time, export_budget)?;
-
-                let chunk = result_iter
-                    .collect_chunk(chunk_size)
-                    .await
-                    .map_err(|e| anyhow::anyhow!("Failed to collect chunk: {}", e))?;
+                let chunk = collect_chunk_within(
+                    &mut result_iter,
+                    chunk_size,
+                    export_deadline,
+                    budget_start,
+                    export_budget,
+                )
+                .await?;
 
                 if chunk.is_empty() {
                     break;
@@ -257,12 +266,14 @@ pub async fn export_data(
                     break;
                 }
 
-                check_export_deadline(export_deadline, start_time, export_budget)?;
-
-                let chunk = result_iter
-                    .collect_chunk(chunk_size)
-                    .await
-                    .map_err(|e| anyhow::anyhow!("Failed to collect chunk: {}", e))?;
+                let chunk = collect_chunk_within(
+                    &mut result_iter,
+                    chunk_size,
+                    export_deadline,
+                    budget_start,
+                    export_budget,
+                )
+                .await?;
 
                 if chunk.is_empty() {
                     break;
@@ -332,12 +343,14 @@ pub async fn export_data(
                     break;
                 }
 
-                check_export_deadline(export_deadline, start_time, export_budget)?;
-
-                let chunk = result_iter
-                    .collect_chunk(chunk_size)
-                    .await
-                    .map_err(|e| anyhow::anyhow!("Failed to collect chunk: {}", e))?;
+                let chunk = collect_chunk_within(
+                    &mut result_iter,
+                    chunk_size,
+                    export_deadline,
+                    budget_start,
+                    export_budget,
+                )
+                .await?;
 
                 if chunk.is_empty() {
                     break;
@@ -414,12 +427,14 @@ pub async fn export_data(
                     break;
                 }
 
-                check_export_deadline(export_deadline, start_time, export_budget)?;
-
-                let chunk = result_iter
-                    .collect_chunk(chunk_size)
-                    .await
-                    .map_err(|e| anyhow::anyhow!("Failed to collect chunk: {}", e))?;
+                let chunk = collect_chunk_within(
+                    &mut result_iter,
+                    chunk_size,
+                    export_deadline,
+                    budget_start,
+                    export_budget,
+                )
+                .await?;
 
                 if chunk.is_empty() {
                     break;
@@ -699,6 +714,66 @@ mod tests {
     }
 }
 
+/// Pull one chunk, abandoning the wait at `deadline` (issue #1695).
+///
+/// The pre-pull [`check_export_deadline`] alone was NOT enough: it only runs between
+/// pulls, and `collect_chunk` can stay PENDING indefinitely while the producer looks
+/// for rows — a no-match scan over a large table is exactly that — so control never
+/// returned to the check and the export ran past the deadline (roborev round 10).
+/// Wrapping the AWAIT bounds the wait itself.
+///
+/// Only the future is wrapped, not the loop or the enclosing per-format `match`, so
+/// no writer or progress-bar borrow moves.
+///
+/// BOTH layers live here rather than at the call sites: an already-expired deadline
+/// is rejected before a pull starts, and a pull that hangs is bounded by the timer.
+/// Keeping them together means a fifth consumption loop cannot be added that forgets
+/// one of them — the four existing ones each got the pre-check by hand, which is
+/// exactly the kind of per-site obligation that rots.
+#[cfg(feature = "state_machine")]
+async fn collect_chunk_within(
+    result_iter: &mut cqlite_core::query::result::QueryResultIterator,
+    chunk_size: usize,
+    deadline: Option<tokio::time::Instant>,
+    started: std::time::Instant,
+    budget: std::time::Duration,
+) -> Result<Vec<cqlite_core::query::result::QueryRow>> {
+    // Layer 1: an ALREADY-expired deadline must not start a pull at all. This cannot
+    // be left to `timeout_at` below, which polls its inner future FIRST and consults
+    // the deadline second — so an immediately-ready chunk would come back `Ok` past
+    // the deadline (the round-7 lesson, applied here before it could recur).
+    check_export_deadline(deadline, started, budget)?;
+
+    // Layer 2: bound the WAIT itself.
+    let pull = result_iter.collect_chunk(chunk_size);
+    let chunk = match deadline {
+        None => pull.await,
+        Some(deadline) => match tokio::time::timeout_at(deadline, pull).await {
+            Ok(chunk) => chunk,
+            // Dropping `pull` drops the in-flight chunk collection; the caller's error
+            // path then drops the iterator, closing the channel and retiring the
+            // producer.
+            Err(_elapsed) => return Err(export_expired(started, budget)),
+        },
+    };
+    chunk.map_err(|e| anyhow::anyhow!("Failed to collect chunk: {}", e))
+}
+
+/// The export-layer elapse. One constructor for both exits — the pre-pull check and
+/// the bounded await — so they cannot drift in operation name or reported figures.
+#[cfg(feature = "state_machine")]
+fn export_expired(started: std::time::Instant, budget: std::time::Duration) -> anyhow::Error {
+    anyhow::Error::new(cqlite_core::Error::QueryTimeout {
+        operation: "cli.export.collect".to_string(),
+        // MEASURED, not assumed equal to the budget: a slow chunk can overshoot, and
+        // the real figure distinguishes "budget too tight" from "one chunk is
+        // uninterruptibly slow".
+        elapsed: started.elapsed(),
+        limit: budget,
+    })
+    .context("CLI export exceeded performance.query_timeout_ms")
+}
+
 /// The export-side budget check (issue #1695), called before each chunk pull.
 ///
 /// Deliberately a check rather than a `timeout_at` wrapper: the four consumption
@@ -722,15 +797,7 @@ fn check_export_deadline(
     if tokio::time::Instant::now() < deadline {
         return Ok(());
     }
-    Err(anyhow::Error::new(cqlite_core::Error::QueryTimeout {
-        operation: "cli.export.collect".to_string(),
-        // MEASURED, not assumed equal to the budget: a slow chunk can overshoot,
-        // and the real figure distinguishes "budget too tight" from "one chunk is
-        // uninterruptibly slow".
-        elapsed: started.elapsed(),
-        limit: budget,
-    })
-    .context("CLI export exceeded performance.query_timeout_ms"))
+    Err(export_expired(started, budget))
 }
 
 #[cfg(all(test, feature = "state_machine"))]
--- a/cqlite-cli/src/commands/export.rs
+++ b/cqlite-cli/src/commands/export.rs
@@ -763,15 +763,19 @@ async fn collect_chunk_within(
 /// the bounded await — so they cannot drift in operation name or reported figures.
 #[cfg(feature = "state_machine")]
 fn export_expired(started: std::time::Instant, budget: std::time::Duration) -> anyhow::Error {
-    anyhow::Error::new(cqlite_core::Error::QueryTimeout {
+    let err = cqlite_core::Error::QueryTimeout {
         operation: "cli.export.collect".to_string(),
         // MEASURED, not assumed equal to the budget: a slow chunk can overshoot, and
         // the real figure distinguishes "budget too tight" from "one chunk is
         // uninterruptibly slow".
         elapsed: started.elapsed(),
         limit: budget,
-    })
-    .context("CLI export exceeded performance.query_timeout_ms")
+    };
+    // #1695: same reasoning as `commands::query::expired` — record on the engine's
+    // own public observability sink so an export consumption elapse reaches
+    // `cqlite.errors.total{category="timeout"}`.
+    cqlite_core::observability::record_error(&err, "query");
+    anyhow::Error::new(err).context("CLI export exceeded performance.query_timeout_ms")
 }
 
 /// The export-side budget check (issue #1695), called before each chunk pull.
--- a/cqlite-cli/src/commands/export.rs
+++ b/cqlite-cli/src/commands/export.rs
@@ -141,6 +141,34 @@ pub async fn export_data(
     // Track timing for statistics
     let start_time = Instant::now();
 
+    // Issue #1695 — `performance.query_timeout_ms` must bound an EXPORT too.
+    //
+    // `cqlite export` is a streaming consumer exactly like `cqlite query`: the engine's
+    // chokepoint wrapper bounds `execute_streaming` above, and everything after it is
+    // this function's `collect_chunk` loops. Bounding only the query path left a
+    // runaway export scan running indefinitely after setup (roborev round 9) — the same
+    // placebo, in the command most likely to be pointed at a whole table.
+    //
+    // Setup was awaited ABOVE, outside this deadline, so a setup elapse stays the
+    // engine's to report; the export gets whatever of the budget is LEFT. Checked
+    // before each chunk pull rather than wrapping the 260-line per-format `match`:
+    // one chunk is this loop's unit of work, so the overshoot is bounded by one chunk
+    // and the writers/progress-bar borrows are left untouched.
+    let export_budget = database.config().query.max_execution_time;
+    let export_deadline = if export_budget.is_zero() {
+        // The documented "no timeout" sentinel.
+        None
+    } else {
+        // `checked_add` because the addend is operator config and `Instant + Duration`
+        // panics when unrepresentable; `None` then means unbounded, which is the right
+        // reading of an astronomically distant deadline.
+        tokio::time::Instant::now().checked_add(
+            export_budget
+                .checked_sub(start_time.elapsed())
+                .unwrap_or_default(),
+        )
+    };
+
     // Resolve the progress total from the CLI `--limit` (spec Option A / R5):
     // a determinate bar + ETA only when the user supplied an explicit limit,
     // otherwise the indeterminate spinner. Never infer a total from data shape.
@@ -171,6 +199,8 @@ pub async fn export_data(
                 if rows_remaining == Some(0) {
                     break;
                 }
+
+                check_export_deadline(export_deadline, start_time, export_budget)?;
 
                 let chunk = result_iter
                     .collect_chunk(chunk_size)
@@ -226,6 +256,8 @@ pub async fn export_data(
                 if rows_remaining == Some(0) {
                     break;
                 }
+
+                check_export_deadline(export_deadline, start_time, export_budget)?;
 
                 let chunk = result_iter
                     .collect_chunk(chunk_size)
@@ -299,6 +331,8 @@ pub async fn export_data(
                 if rows_remaining == Some(0) {
                     break;
                 }
+
+                check_export_deadline(export_deadline, start_time, export_budget)?;
 
                 let chunk = result_iter
                     .collect_chunk(chunk_size)
@@ -379,6 +413,8 @@ pub async fn export_data(
                 if rows_remaining == Some(0) {
                     break;
                 }
+
+                check_export_deadline(export_deadline, start_time, export_budget)?;
 
                 let chunk = result_iter
                     .collect_chunk(chunk_size)
@@ -660,5 +696,85 @@ mod tests {
         assert!(t.contains("{pos}"), "must show a row count: {t:?}");
         assert!(!t.contains("{eta}"), "must NOT show an ETA: {t:?}");
         assert!(!t.contains("{percent}"), "must NOT show a percent: {t:?}");
+    }
+}
+
+/// The export-side budget check (issue #1695), called before each chunk pull.
+///
+/// Deliberately a check rather than a `timeout_at` wrapper: the four consumption
+/// loops live inside a 260-line per-format `match` holding file writers and a
+/// progress bar, and wrapping that whole expression in an async block to make it
+/// cancellable would move every one of those borrows. Checking before each pull
+/// bounds the overshoot by ONE chunk — the loop's own unit of work — and changes no
+/// ownership.
+///
+/// `deadline` is `None` for the documented `Duration::ZERO` sentinel (unbounded) and
+/// also when the deadline is unrepresentable, which means the same thing.
+#[cfg(feature = "state_machine")]
+fn check_export_deadline(
+    deadline: Option<tokio::time::Instant>,
+    started: std::time::Instant,
+    budget: std::time::Duration,
+) -> Result<()> {
+    let Some(deadline) = deadline else {
+        return Ok(());
+    };
+    if tokio::time::Instant::now() < deadline {
+        return Ok(());
+    }
+    Err(anyhow::Error::new(cqlite_core::Error::QueryTimeout {
+        operation: "cli.export.collect".to_string(),
+        // MEASURED, not assumed equal to the budget: a slow chunk can overshoot,
+        // and the real figure distinguishes "budget too tight" from "one chunk is
+        // uninterruptibly slow".
+        elapsed: started.elapsed(),
+        limit: budget,
+    })
+    .context("CLI export exceeded performance.query_timeout_ms"))
+}
+
+#[cfg(all(test, feature = "state_machine"))]
+mod export_deadline_tests {
+    use super::check_export_deadline;
+    use std::time::Duration;
+
+    /// The `Duration::ZERO` sentinel reaches this as `None` and must never fail a
+    /// chunk pull — an unbounded export is a legal operator configuration.
+    #[test]
+    fn no_deadline_never_stops_an_export() {
+        assert!(
+            check_export_deadline(None, std::time::Instant::now(), Duration::ZERO).is_ok(),
+            "an unbounded export must not be interrupted"
+        );
+    }
+
+    /// An expired deadline must stop the export BEFORE the next chunk is pulled, and
+    /// must say so as the distinct timeout error carrying the export-side operation —
+    /// not the query-side one, so an operator can tell which loop elapsed.
+    ///
+    /// Deterministic by construction: the deadline is a full second in the past, so no
+    /// clock has to advance and no chunk has to be slow.
+    #[tokio::test]
+    async fn an_expired_deadline_stops_the_export_before_the_next_chunk() {
+        let budget = Duration::from_secs(30);
+        let err = check_export_deadline(
+            Some(tokio::time::Instant::now() - Duration::from_secs(1)),
+            std::time::Instant::now(),
+            budget,
+        )
+        .expect_err("an expired export deadline must stop the chunk loop");
+
+        match err.downcast_ref::<cqlite_core::Error>() {
+            Some(cqlite_core::Error::QueryTimeout {
+                operation, limit, ..
+            }) => {
+                assert_eq!(
+                    operation, "cli.export.collect",
+                    "must name the EXPORT loop, distinct from `cli.query.collect`"
+                );
+                assert_eq!(limit, &budget);
+            }
+            other => panic!("expected QueryTimeout, got {other:?}"),
+        }
     }
 }
--- a/cqlite-cli/src/commands/mod.rs
+++ b/cqlite-cli/src/commands/mod.rs
@@ -47,7 +47,7 @@ pub use export_sstable::export_sstable;
 pub use import::import_data;
 pub use inspect::{analyze_sstable, validate_sstable};
 #[cfg(feature = "state_machine")]
-pub use query::collect_query_result;
+pub use query::{collect_query_result, collect_rows_until};
 pub use query::{execute_query, execute_select_query};
 pub use read::{read_sstable, read_sstable_enhanced};
 pub(crate) use schema_load::load_schema_file;
--- a/cqlite-cli/src/commands/query.rs
+++ b/cqlite-cli/src/commands/query.rs
@@ -229,7 +229,13 @@ pub async fn collect_query_result(
         // remainder, both mean setup consumed the whole budget — so consumption gets
         // a deadline already in the past rather than an unbounded run.
         let remaining = budget.checked_sub(started.elapsed()).unwrap_or_default();
-        Some(tokio::time::Instant::now() + remaining)
+        // `Instant + Duration` PANICS when the sum is unrepresentable, and `remaining`
+        // derives from operator config (`performance.query_timeout_ms`), so a huge
+        // value must not abort the process. `checked_add` returning `None` lands on
+        // the same `None` this function already uses for "unbounded" — which is the
+        // right meaning, since an unrepresentable deadline is astronomically far
+        // away. Mirrors the core's treatment in `deadline::bound_tracked`.
+        tokio::time::Instant::now().checked_add(remaining)
     };
 
     collect_rows_until(iter, display_limit, deadline, started, budget).await
@@ -253,10 +259,23 @@ pub async fn collect_rows_until(
     started: std::time::Instant,
     budget: std::time::Duration,
 ) -> Result<cqlite_core::query::result::QueryResult> {
-    let collect = collect_streamed_rows(iter, display_limit);
     let Some(deadline) = deadline else {
-        return collect.await;
+        return collect_streamed_rows(iter, display_limit).await;
     };
+
+    // Reject an ALREADY-EXPIRED deadline before polling collection at all.
+    // `timeout_at` polls its inner future FIRST and only then consults the deadline,
+    // so an expired deadline still returns `Ok` whenever collection happens to be
+    // immediately ready — `display_limit = Some(0)` breaks the loop on its first
+    // iteration, and a fully buffered stream can complete without yielding. Relying
+    // on `timeout_at` alone therefore made "expired means timeout before pulling
+    // rows" true only for streams that happen not to be ready (roborev round 7);
+    // this makes it true by construction, which is also what lets the test assert it.
+    if deadline <= tokio::time::Instant::now() {
+        return Err(expired(started, budget));
+    }
+
+    let collect = collect_streamed_rows(iter, display_limit);
     match tokio::time::timeout_at(deadline, collect).await {
         Ok(result) => result,
         // Dropping `collect` drops the iterator, closing the channel and retiring the
@@ -268,16 +287,24 @@ pub async fn collect_rows_until(
         // in consumption. Closing it properly means carrying the deadline into the
         // core streaming producer so core owns and records it — which is the
         // per-batch bound the issue explicitly deferred. Follow-up.
-        Err(_elapsed) => Err(anyhow::Error::new(cqlite_core::Error::QueryTimeout {
-            operation: "cli.query.collect".to_string(),
-            // MEASURED, never assumed equal to the budget: a blocked poll can
-            // overshoot, and the real figure tells an operator "budget too tight"
-            // from "one decode unit is uninterruptibly slow".
-            elapsed: started.elapsed(),
-            limit: budget,
-        })
-        .context("CLI query exceeded performance.query_timeout_ms")),
+        Err(_elapsed) => Err(expired(started, budget)),
     }
+}
+
+/// The CLI-layer collection elapse. One constructor for both exits (the up-front
+/// expired-deadline check and the `timeout_at` arm) so they cannot drift in
+/// operation name, reported figures or context.
+#[cfg(feature = "state_machine")]
+fn expired(started: std::time::Instant, budget: std::time::Duration) -> anyhow::Error {
+    anyhow::Error::new(cqlite_core::Error::QueryTimeout {
+        operation: "cli.query.collect".to_string(),
+        // MEASURED, never assumed equal to the budget: a blocked poll can overshoot,
+        // and the real figure tells an operator "budget too tight" from "one decode
+        // unit is uninterruptibly slow".
+        elapsed: started.elapsed(),
+        limit: budget,
+    })
+    .context("CLI query exceeded performance.query_timeout_ms")
 }
 
 /// The unbounded consumption loop; [`collect_rows_until`] applies the deadline.
--- a/cqlite-cli/src/commands/query.rs
+++ b/cqlite-cli/src/commands/query.rs
@@ -194,6 +194,58 @@ pub async fn collect_query_result(
 ) -> Result<cqlite_core::query::result::QueryResult> {
     use cqlite_core::query::result::{QueryResult, StreamingConfig};
 
+    // Issue #1695 — the operator's budget must cover THIS loop, not just setup.
+    //
+    // The engine's ONE chokepoint wrapper bounds the `execute_streaming` FUTURE,
+    // i.e. stream setup and time-to-first-batch. It structurally cannot bound what
+    // happens after that future returns, and what happens after it returns is where
+    // a CLI query spends ~all of its time: the `next_async()` loop below drives the
+    // incremental scan. Since the CLI is the only consumer that sets
+    // `performance.query_timeout_ms`, leaving consumption unbounded would leave the
+    // knob a placebo on the exact runaway full scan #1695 exists to bound.
+    //
+    // This is a SECOND bound at a DIFFERENT layer, not a second bound on the engine
+    // path, and it does NOT restart the engine's clock: the two run concurrently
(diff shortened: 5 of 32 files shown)


pmcfadin:
## ✅ Gate of record — `RESULT: PASS` (AC4)

Run on the **rebased** head `f376f1d90`, which carries `99b52d1bd` (the #3393 OOM fix). A previous attempt at `9dfdeecb9` was killed by the coordination lead precisely because its tree lacked that fix, and an attempt after the rebase was voided by an ENOSPC (#2923) that failed `memory-budget` — neither is pasted here, because neither is a verdict.

Worth calling out: **`tooling-tests: PASS (658s)`** — that is the component which materialised ~27GB and OOM-killed the box on the pre-fix tree. Clean here.

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.CCs676
commit: f376f1d branch: issue-1695-enforce-query-max-execution-time dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-1695/test-data/schemas (checkout-relative)
tree-start: f376f1d90150 dirty: no digest: 09d8b276de38
tree-end: f376f1d90150 dirty: no digest: 09d8b276de38
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (7s)
clippy:            PASS (5s)
roborev-lints:     PASS (45s)
core-tests:        PASS (55s)
tombstones-scan:   PASS (12s)
scan-offload-guard: PASS (175s)
work-counters-guard: PASS (90s)
byte-budget-guard: PASS (14s)
arrow-parity-guard: PASS (43s)
memory-budget:     PASS (141s)
integration-tests: PASS (63s)
format-compat:     PASS (43s)
write-tests:       PASS (153s)
cli-tests:         PASS (176s)
flight-query-semantics-oracle: PASS (81s)
python-bindings:   PASS (57s)
node-bindings:     PASS (137s)
delivery-telemetry: PASS (0s)
oom-audit:         PASS (3s)
parity-report:     PASS (0s)
operator-metrics-doc: PASS (37s)
tooling-tests:     PASS (658s)
minimal-build:     PASS (82s)
smoke:             PASS (4s)
logs: /tmp/agent-gate.CCs676
summary-file: /tmp/cqlite-gates/1695/full-085955-1776575.txt
RESULT: PASS
```

**Verification, not just a paste:** `tree-integrity: PASS` with `tree-start` == `tree-end` == `f376f1d90150 digest 09d8b276de38`, `dirty: no`, **all 30 components PASS with zero SKIPs**, `datasets: 144 Data.db`, `schemas: 6/6`. `file-size` passes under the disclosed `CQLITE_ALLOW_FILE_GROWTH=1` (which prints no SUMMARY marker — that gap is #3402), with the growth table and its rationale in the PR description.

One residual that is **not** a failure: `accelerators: … mold=absent perf=paranoid-4`. `perf=paranoid-4` is a permission verdict meaning this host cannot be profiled; remedy is `bash scripts/bootstrap-agent-machine.sh --yes`. It gates nothing in this PR.

pmcfadin:
## C — intent audit against issue #1695's acceptance criteria

**Provenance, stated up front because it matters:** this audit was performed **by the lead**, not by the `spec-auditor` subagent. I spawned `spec-auditor` twice-over (one spawn, one direct follow-up request naming C as merge-blocking) and it returned **two idle notices and zero verdict packets** — the fleet-wide subagent pattern the coordination lead flagged today (5-of-5 on one issue). **An idle notice is not a verdict**, so rather than let silence stand in for a PASS, I ran the audit myself and am labelling it as such. Weigh it accordingly: it is an audit by the party that drove the work, which is exactly the independence C normally provides and this one does not.

This is an **oracle-driven** issue — no OpenSpec change — so the anchor is the issue's own ACs.

| AC | Verdict | Evidence |
|---|---|---|
| **1.** `max_execution_time = 1ms` + large-scan fixture → timeout `Err` | **satisfied** | `budget_of_one_millisecond_times_out_a_large_scan` (the literal form, ~650 KiB / 1000-partition fetched table) **plus** `expired_budget_times_out_a_committed_scan`, a host-independent `must_run` form over the committed fixture |
| **2.** Core default 300s unchanged; zero sentinel documented + validated | **satisfied** | `cqlite-core/src/config.rs:536` is still `Duration::from_secs(300)`; `default_budget_leaves_a_normal_query_unaffected`, `zero_sentinel_means_unbounded`, `config_validate_accepts_the_zero_sentinel` |
| **3.** Post-timeout drop leaks nothing; producers exit | **satisfied** | `timed_out_streaming_query_leaves_no_live_producer` and `dropping_a_streaming_iterator_retires_its_producer` |
| **4.** `agent-gate.sh` PASS, SUMMARY pasted | **satisfied** | `RESULT: PASS`, all 30 components, zero SKIPs — issuecomment-5451189295 |

**C: PASS** — no AC unmet, none uncovered, no unjustified `partial`.

### The three things I checked hardest, and why

**The rebase resolution is semantically intact, not merely syntactically.** The rebase onto `origin/main` conflicted in `Config::validate()` because #1697 landed memtable/STCS rules in the same function #1695 adds its sentinel block to. I kept both sides by hand. Verified after the fact: #1697's rules are present at `config.rs:51/94/116` within `validate()` (which opens at :800) and #1695's budget block is at :927, inside the same function. `core-tests PASS` on the rebased tree independently confirms it.

**The CLI second bound is real wiring, not a helper-only test.** This is the highest-risk part of the diff — roborev round 5 found the feature did not bind where it is configured. Named call chain: `commands/query.rs:216` reads `database.config().query.max_execution_time`, derives a *remaining* budget via `checked_sub(started.elapsed())`, and `collect_rows_until` (:241) bounds **consumption**, not just setup; `commands/export.rs` routes all four `collect_chunk` pulls (:210/269/346/430) through `collect_chunk_within` with a shared `budget_start`. `every_public_entry_point_is_bounded` covers `execute_with_params` and `execute_prepared` so the knob is not a placebo on three of four paths.

**The AC1 test cannot pass vacuously, and does not skip silently.** `assert_timed_out` asserts the variant *and* that the reported limit equals the configured one *and* that the error names the entry point; the `Ok` arm panics with an explicit regression message. The fetched-fixture case is skip-clean when the corpus is absent, but the same property is pinned by a `must_run` fail-closed committed-fixture case — which is the #3220 trap (a case skipping silently behind green siblings) closed by construction. AC3 likewise asserts a task-count return to baseline with a precondition assert, and the complementary drop test uses a deliberate one-row channel buffer so the producer cannot finish before the check. No wall-clock threshold asserts anywhere in the correctness path (#2642).

### Two settled decisions confirmed as implemented, not re-litigated
- **AC2's "300s unchanged" is the CORE `QueryConfig` default**, which is unchanged. The CLI's own 30s is now enforced — a behaviour change the owner ruled on (issuecomment-5448884756) and which the PR body documents with a before/after table.
- **Deadline propagation into the core streaming producer is deferred**, exactly as the issue body prescribes. Documented, not silently absent.

pmcfadin:
## Final roborev — recorded as **FAIL with an accepted, documented deferral**, never as "clean"

Two rounds were run, because the first was invalidated by a tooling defect rather than by anything in this diff.

### Round 1 (job 52, at `f376f1d90`) — genuine review, zero findings, invalidated by #3392

```
census: 32 files, +3329/-188
tokens: input=4,497,857 cached=4,299,264 output=14,087
review-completed: PASS   vacuity-tier1: PASS   vacuity-tier2: PASS
findings: NONE
sha-assert: FAIL (reviewed range does not match origin/main...HEAD)
prompt-content: FAIL (32/32 code census paths absent from the prompt)
RESULT: FAIL
```

`sha-assert` failed for a reason that was **provably not a scope error**: the reviewed range `99b52d1bd..f376f1d90` equals `git merge-base origin/main HEAD`, i.e. the correct scope. The pre-fix wrapper compared it against `origin/main`'s **tip**, and main gained 4 commits during the review window. That is **#3392, fixed on `main` at 01:51Z by `938d464fb` (#3416)** — *"sha-assert compares the reviewed range against the MERGE-BASE, not origin/main's tip"*. This branch carried the pre-fix wrapper because it was based at `99b52d1bd`, which predates the fix.

**Rather than request a waiver around that assert, I rebased onto `c6d6755d8` to obtain the fixed tooling** — which also permanently removes the failure mode, since a merge-base comparison cannot be invalidated by main moving again.

### Round 2 (job 53, at `9b944f5bb`) — the round of record

```
head-sha: 9b944f5bb72753ce89846f508a86ab83db15c895
reviewed-sha: c6d6755d84ce4f6a18533a5b51bfdca1ea2f4278..9b944f5bb72753ce89846f508a86ab83db15c895
assert-base: c6d6755d8… (merge-base of origin/main and HEAD; origin/main tip c6d6755d8…)
census: 32 files, +3329/-188
tokens: input=3,512,719 cached=3,340,800 output=12,137
sha-assert: PASS    review-completed: PASS   vacuity-tier1: PASS   vacuity-tier2: PASS
findings: PRESENT (1)
prompt-content: FAIL (32/32 code census paths absent from the prompt)
RESULT: FAIL
```

### The one finding — the pre-authorized deferral, at a named site

> **Medium** · `cqlite-core/src/storage/sstable/generation_merge.rs:706` — the cross-generation streaming producer constructs `KWayMerger` with an inert cancellation token. The receiver-closure check at :729 runs only *between* `merger.step()` calls, so dropping an iterator after a timeout cannot interrupt an in-progress step or its reader threads. Suggested fix: give `RowScanStream` a per-stream cancellation guard and pass its token to `KWayMerger::new_cancellable`.

**This is the deferral the issue body prescribes**, not a new defect: *"pick the simpler contract: bound the whole `execute` future including stream setup, document that streamed row consumption afterwards is not bounded, and note per-batch bounds as a possible follow-up."* It has recurred across ~18 review rounds; the coordination lead authorized landing with it (issuecomment-5447533380) and the owner confirmed the endgame (issuecomment-5448884756). It is **accepted and documented, and this PR does not claim roborev is clean.** Follow-up scope, for whoever picks it up: a composable `ScanCancel` token threaded through `RowScanStream` → `KWayMerger::new_cancellable`, plus the `JoinedStream` drop-guard — roughly a day, sharing machinery with the Flight cancellation path.

Worth noting the finding **got sharper, not weaker, across rounds**: round 1 reported nothing, round 2 named the exact construction site and the exact API. That is the argument for having re-run rather than banking round 1's `findings: NONE`.

### `prompt-content: FAIL` — #3312, and I am not self-waiving it

This is the known snapshot-delivery false-FAIL (#3312, closed): roborev delivers a large diff by a transient snapshot path, and the check reads inline headers only. The corroboration that this was a real review, not a vacuous one, is the **token accounting** — 3.5M input / 3.3M cached / 12k output against a vacuous baseline of ~18.7k input / 0 cached / ~55 output — together with `review-completed`, both vacuity tiers, `code-free`, `job-record` and `census-check` all PASS, and a finding that names a `file:line` existing only in this diff.

A waiver exists for exactly this verdict, but it is **owner- or lead-granted only, and a worker may never self-apply one** — so this is a **request**, not an authorization, and I have deliberately not written the marker syntax anywhere in this comment, because worker and owner post through the same login on this fleet and a marker of mine could otherwise be read as authorization. If the lead or owner wants the record to show a PASS here, the authorization must come from them, bound to `base=c6d6755d84ce4f6a18533a5b51bfdca1ea2f4278`, `head=9b944f5bb72753ce89846f508a86ab83db15c895`, `job=53`.

pmcfadin:
## ⚖️ Waiver request relayed to the owner — `prompt-content` absence, job 53

Relayed by the coordination lead. The lane requested it and may not self-apply it; the owner has reserved waiver authorization to themselves, so I am not granting it either.

**Triple, verified by me from git rather than read off the lane's comment:**

| field | value |
|---|---|
| `base` | `c6d6755d84ce4f6a18533a5b51bfdca1ea2f4278` |
| `head` | `9b944f5bb72753ce89846f508a86ab83db15c895` |
| `job` | `53` |

PR head confirmed still `9b944f5bb…`. Unmoved, so the ask is current.

**Evidence:** `3,512,719` input / `3,340,800` cached / `12,137` output — ~190× the vacuous baseline (~18.7k / 0 / ~55) and proportionate to a 32-file, +3329/−188 diff. `push-assert`, `census-check`, `code-free`, `job-record`, `sha-assert`, `review-completed`, `vacuity-tier1`, `vacuity-tier2` all PASS. And the returned finding **names `generation_merge.rs:706` plus the exact API (`KWayMerger::new_cancellable`)** — a location that exists only in this diff. A reviewer that received nothing cannot do that.

**The single finding is the pre-authorized deferral**, not a new defect: the issue body prescribes bounding the whole `execute` future and documenting that streamed row consumption afterwards is unbounded, with per-batch bounds as a follow-up. I authorized landing with it (`issuecomment-5447533380`) and the owner confirmed the endgame (`issuecomment-5448884756`). The PR does **not** claim roborev is clean.

**Two things worth the owner's attention beyond the ask:**

1. **The lane fixed a `sha-assert: FAIL` by rebasing to obtain the corrected tooling rather than by requesting a waiver around it.** Round 1's reviewed range was `99b52d1bd..f376f1d90`, exactly `git merge-base origin/main HEAD` — provably the correct scope — and the pre-fix wrapper compared it against `origin/main`'s *tip*, which had gained 4 commits during the review window. That is **#3392, fixed on `main` at 01:51Z by `938d464fb` (#3416)**. Rebasing removed the failure mode permanently, because a merge-base comparison cannot be invalidated by main moving again. **#3416 merged this morning and paid for itself within the hour**, on a different lane than the one that fixed it.
2. **The finding got sharper across rounds, which is the argument for not banking round 1.** Round 1 returned `findings: NONE`; round 2 named the exact construction site and API. Accepting round 1's clean result would have shipped without the specific follow-up scope now recorded.

This is the **fourth** waiver pending: #3382 (job 12), #3408 (job 9), #3397 (job 55), #3398 (job 53). Each is the sole remaining gate on a PR that is otherwise green and certified.

pmcfadin:
## ✅ Gate of record re-run at the final SHA `9b944f5bb` — `RESULT: PASS`

The second rebase (onto `c6d6755d8`, to obtain the #3416 `sha-assert` fix) changed the tree, so the previous gate at `f376f1d90` no longer certified it. Re-run in full rather than carried over:

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.EGMemg
commit: 9b944f5 branch: issue-1695-enforce-query-max-execution-time dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-1695/test-data/schemas (checkout-relative)
tree-start: 9b944f5bb727 dirty: no digest: 7ca6629b4897
tree-end: 9b944f5bb727 dirty: no digest: 7ca6629b4897
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (5s)
clippy:            PASS (128s)
roborev-lints:     PASS (43s)
core-tests:        PASS (536s)
tombstones-scan:   PASS (43s)
scan-offload-guard: PASS (147s)
work-counters-guard: PASS (82s)
byte-budget-guard: PASS (13s)
memory-budget:     PASS (141s)
integration-tests: PASS (55s)
format-compat:     PASS (38s)
write-tests:       PASS (145s)
cli-tests:         PASS (121s)
flight-query-semantics-oracle: PASS (48s)
python-bindings:   PASS (285s)
node-bindings:     PASS (192s)
delivery-telemetry: PASS (0s)
oom-audit:         PASS (1s)
parity-report:     PASS (1s)
operator-metrics-doc: PASS (28s)
tooling-tests:     PASS (679s)
minimal-build:     PASS (57s)
smoke:             PASS (48s)
logs: /tmp/agent-gate.EGMemg
summary-file: /tmp/cqlite-gates/1695/full-101804-2649823.txt
RESULT: PASS
```

**Verified:** `tree-integrity: PASS` with `tree-start` == `tree-end` == `9b944f5bb727 digest 7ca6629b4897`, `dirty: no`, **all 30 components PASS, zero SKIPs**, `datasets: 144 Data.db`, `schemas: 6/6`. `tooling-tests: PASS (679s)` — clean again on the #3393 fix.

### Everything now certifies the same commit

| check | verdict | at |
|---|---|---|
| agent-gate | **PASS**, 30/30, 0 SKIP | `9b944f5bb` |
| roborev `sha-assert` | **PASS** (merge-base `c6d6755d8`) | `9b944f5bb` |
| roborev findings | 1 — the **pre-authorized deferral**, recorded not waived away | `9b944f5bb` |
| C intent audit | **PASS**, 4/4 ACs satisfied (lead-performed — see provenance note) | diff unchanged: census identical at 32 files, +3329/−188 |
| `premerge-assert` | pending, will re-run immediately before arming | `9b944f5bb` |
| `prompt-content` | **FAIL** (#3312) — waiver requested, **owner-reserved, not granted** | job 53 |

On C's currency across the rebase: the audit was performed at `f376f1d90`, and the second rebase added only main's 4 commits — the diff under audit is byte-identical (census `32 files, +3329/-188` in both roborev rounds), so its findings carry to `9b944f5bb` unchanged.

**I am not arming `--auto`.** The coordination lead has recorded the `prompt-content` waiver as the sole remaining gate and reserved it to the owner. Arming auto-merge now would land this PR the moment `required` goes green and bypass that gate, which is the opposite of what a pending authorization means. Holding until the owner rules.

pmcfadin:
roborev-waive: prompt-content-absent base=c6d6755d84ce4f6a18533a5b51bfdca1ea2f4278 head=9b944f5bb72753ce89846f508a86ab83db15c895 job=53 reason=owner grant 2026-08-28; #3312 snapshot-delivery absence; tokens 3512719 input / 3340800 cached / 12137 output vs vacuous baseline ~18700/0/55; single finding is the pre-authorized deferral named in the issue body, recorded FAIL-with-ruled-findings; gate of record PASS at this head

pmcfadin:
Waiver above granted by the owner. Apply it with a recheck of job 53, then proceed to merge.

PR #3416, opened 2026-08-28
fix(#3392): sha-assert compares the reviewed range against the MERGE-BASE, not origin/main's tip

### LINKED ISSUE #3392: roborev wrapper: sha-assert compares a merge-base-relative reviewed range against origin/main's TIP — fails deterministically on any branch whose main has advanced

pmcfadin:
Closes #3392.

## The defect

`sha-assert` asked its question against two different definitions of "base":

- the census measures `git diff --numstat -z --no-renames "${BASE}...HEAD"` — **three-dot**, i.e. `merge-base(BASE, HEAD)..HEAD`;
- `BASE_SHA` was `git rev-parse "${BASE}^{commit}"` — the **TIP** of `origin/main`;
- the assert compared the record's (merge-base-relative) `git_ref` against that tip.

roborev is correct: it reviews `merge-base..HEAD`, which is the true branch diff and exactly what `--base origin/main` means under three-dot semantics. So the assert FAILed **deterministically** for any branch whose `main` had advanced since its merge-base — nearly every branch not freshly rebased. Re-running or waiting could never help.

**Not a race.** The issue was filed on a race diagnosis; the owner's correction disproved it with a controlled measurement (`origin/main` unmoved across the window, assert failing anyway) and ruled out the originally-recommended "pin `BASE_SHA` at launch" — pinning still pins the *tip*, and the mismatch is tip-vs-merge-base.

## The change

`BASE_TIP_SHA` (the ref's tip) and `RANGE_BASE_SHA` (the merge-base) are now separate named values. The merge-base is resolved **once**, in the census, and reused by the census diff, the enqueue, `sha-assert:` and the absence-waiver scope. The tip survives for exactly one purpose: the T1 root-checkout base-equality signature, which genuinely needs it.

Fails closed, with no degradation to the tip: `FAIL (base '<ref>' unresolvable)`, `FAIL (no merge-base between '<ref>' and HEAD)`, `FAIL (merge-base ... unusable)` when the command succeeds without yielding exactly one 40-hex sha, `FAIL (git diff failed)`. No review is enqueued in any of them.

New informational key, deliberately outside **both** the verdict-affirmation loop and the closed verdict-grammar scan (both enumerate explicit variables):

```
assert-base: <merge-base> (merge-base of <base> and HEAD; <base> tip <sha>)
```

`base:` stays symbolic so a pasted block cannot be misread as a stale-base review.

## Acceptance criteria

- **AC1** — moved base, unchanged head ⇒ `sha-assert: PASS`. Case `mb1`, on an `advanced-base` fixture. **Asserted non-vacuous**: `assert_base_advanced` FAILs any fixture where tip == merge-base ("nothing about the distinction is under test").
- **AC2** — stale head still FAILs; T1 base-equality and T2 empty-tree signatures intact, the former still comparing against the **tip**. Cases `mb2`–`mb6`.
- **AC3** — the block states which base was compared, and `--help` + CLAUDE.md + the canonical site + the capability spec all agree.
- **AC4** — moved-base case plus the negative control. `mb2` is the paired discriminator: same fixture, range stops one commit short, present precisely in case `mb1` passed because the assert stopped checking.

## Verification

```
scripts/agent-gate.sh --lite      RESULT: PASS   tree-integrity: PASS   commit 6555a39
test_roborev_review_guard.sh      806 passed, 0 failed
test_roborev_guard_portability.sh 129 passed, 0 failed, 0 skipped
roborev job 45 (codex/gpt-5.6-sol) RESULT: PASS  findings NONE
                                   input 983672 / cached 870400 / output 7219
                                   prompt-content PASS (5/5) · both vacuity tiers PASS
```

Full `agent-gate.sh` (the gate of record) recorded separately below.

## What is NOT evidence here, stated deliberately

**This PR's own roborev rounds are corroboration, never the oracle.** The wrapper is read from the invoking worktree, so these rounds ran the *fixed* script — a green `sha-assert` is equally consistent with a fix that stopped checking. That is #3042's lesson (a CQLite-written + CQLite-read round-trip is invariant to a uniform error) applied to a tooling change rather than a format one. The oracle is the hermetic `mb1`/`mb2` pair.

Two specifics:
- Round 1 (job 40) ran at **one commit of drift** (`assert-base: 33f8adc3e`, tip `b3729b6ae`) and `sha-assert` held — under the old code that round would have voided. Useful corroboration.
- Round 2 (job 45, above) ran **after the rebase**, so merge-base == tip and it could not exercise the moved-base condition at all. Recorded so nobody reads its PASS as a demonstration.

## "A caller could already pass the merge-base" — yes, and that is the argument FOR this change

Verified against unfixed `origin/main`, not assumed: `--base) BASE="$2"` (`:518`) accepts any string, and for `mb = merge-base(HEAD, origin/main)`, `rev-parse(mb) == merge-base(mb, HEAD) == mb`, so `mb..HEAD` and `mb...HEAD` are the same 5 commits. `--base "$(git merge-base HEAD origin/main)"` works today.

**So this PR's value is not "make a correct range possible" — it is "make the safe thing the default, so correctness stops depending on every caller remembering an incantation."** Same argument the repo already makes about `--repo` on `--branch`: the flag existed all along, and its absence still cost vacuous reviews fleet-wide until the wrapper made it mandatory. A guard whose correctness rests on caller discipline has a documented bypass, and every lane is a caller. The workaround also leaves **no evidence** — a reader of a pasted block cannot tell whether the operator remembered — and cannot reach the waiver-scope binding at all.

## Measured cost of the bug (two independent lanes)

- PR #3376 (#1712) rounds 6 and 7, ~1.3M input each.
- PR #3397 (#1701) job 22: **2.27M input / 2.11M cached, `prompt-content: PASS 24/24`, real findings acted on** — discarded solely because the reviewed base was a stale main. It then nearly recurred the same night at 2 more commits of drift. **Two occurrences, one lane, one night, on a 4-lane box: a rate, not a coincidence.** Credit: lane-1701.

**It also unlocks recovery.** A round voided *only* by this bug should be salvageable with `--recheck-job <id>`, which re-decides a stored verdict and enqueues nothing — zero reviewer tokens. Applies only to a round that was *valid and voided*; a findings FAIL re-derives the same FAIL from the record.

## Included beyond the headline fix, and why

1. **Finding 1 (roborev, Medium):** the enqueue is now pinned to the resolved merge-base sha, so census, enqueue, assert and waiver share one immutable range. With the merge-base as the base, two-dot and three-dot coincide — the semantics gap that caused this bug stops existing rather than being papered over.
2. **Finding 2 (roborev, Low):** case `mb8` asserts its terminal verdict, so waiver admission cannot rot behind a green test.
3. **`--help` heredoc blocker (found by executing the tool):** `usage()` is an unquoted heredoc, so newly added markdown backticks ran as command substitution — three stderr errors, and `base=`, `--base`, `assert-base:` **silently deleted from the rendered help**. That text is the only sanctioned source for the waiver marker's base sha, so this change would have re-bound the break-glass and deleted its documentation in one stroke. Now has a permanent regression case asserting `--help` exits 0 with **empty stderr** — which immediately caught a backtick the rebase reintroduced. No static review can see this class; in a diff a backtick reads as markdown emphasis.
4. **The classifier-absence scan was fail-open under `pipefail`.** Not unrelated, though I first thought so: the `mixed-delivery` needle sits at byte 38,949 **inside the `usage()` heredoc**, which survives the `^\s*#` filter because heredoc text is not a comment. Measured on pristine `main`: a **67,528-byte** stream against a **65,536-byte** pipe buffer ⇒ an early `grep -qF` match SIGPIPEs the upstream, the pipeline returns 141, and a real violation is recorded as clean — **1/40 fired, 39/40 silent**. Since this change edits `usage()`, it moves those bytes and can flip that guard's verdict, so it is obliged to fix it. Now captured once and matched in pure bash `case` — no pipe, no status to lose — with `--help` prose exempted for the same reason comments are. The standing measurement is on #3367.

## Doctrine updated in the same change

CLAUDE.md rule 2, the canonical `agents-developing/roborev-findings` page (including its live-worktree probe, which told readers to expect `origin/main` — the exact value the bug asserted), and the `roborev-review-guard` capability spec.

**On the spec being a main spec edited directly:** kept deliberately. `9f11eb7cf` (#3229's *fix* commit) edited this same file the same way, the content is normative `SHALL` language for precisely the contract this change alters, and backing it out would leave the spec describing a tip-based contract the wrapper no longer implements. There is no active `openspec/changes/` dir, so the alternative is inventing a proposal after the fact for a bug fix CLAUDE.md routes away from OpenSpec.

## Known coupling, flagged rather than discovered later

This diff **adds** net text to `usage()`, which grows the 1,992-byte excess over the pipe buffer described above. With item 4 fixed that no longer affects a verdict, but any future edit to these three files perturbs that margin: growth makes the old vacuous PASS more likely, and shrinking back under 65,536 bytes would have made it deterministically red fleet-wide.

## Review provenance, stated accurately

One independent review (roborev codex, jobs 40 and 45) plus my own verification. A `rust-reviewer` subagent was dispatched and **never delivered findings** across ~30 minutes and three requests, so this did not receive a second independent review. Both implementer and reviewer subagents were stopped mid-run — the implementer was rewriting the wrapper underneath a verification, reintroducing a fixed backtick — and the final two commits are the lead's, not the implementer's.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -423,7 +423,15 @@ implement (TDD) → --lite each fix round (summary-file redirect)
   the job-id carrier: for a range review it names only the BASE, so an unavailable record FAILs rather
   than falling back to prose that verifies nothing. A range that does not match, a SINGLE-COMMIT record
   (even one equal to HEAD), or a base-equal scope **aborts the round** — base-equality is the signature of
-  the worktree bug. **(3)** `"contains no code changes to review"` on a
+  the worktree bug. **The expected RANGE BASE is the MERGE-BASE, never the base ref's TIP (#3392)**:
+  `<base>...HEAD` *is* `merge-base(<base>, HEAD)..HEAD`, so an assert that expected the tip FAILED
+  DETERMINISTICALLY on a CORRECT review of any branch whose `main` had advanced past its branch point —
+  i.e. almost every branch not just rebased. It was misdiagnosed as a race **twice** (the falsifying
+  control: `origin/main` recorded before AND after a failing round, unmoved). The tip is still read, for
+  the T1 root-checkout signature alone, and the block now prints an informational
+  `assert-base: <merge-base> (merge-base of <base> and HEAD; <base> tip <sha>)` so the two can never be
+  confused in a pasted block. The absence waiver's `base=` field is bound to that same merge-base —
+  copy it from `assert-base:`, not from `base:`. **(3)** `"contains no code changes to review"` on a
   NON-EMPTY diff is a **HARD FAIL**, never a pass. **(4)** A docs-only (code-free) diff **cannot be
   roborev-certified at all** — and "docs-only" means a **CODE-FREE CENSUS as the wrapper classifies it,
   NEVER a `docs/` path prefix** (#3229). The mechanism, stated correctly: **roborev drops exactly what
--- a/openspec/specs/roborev-review-guard/spec.md
+++ b/openspec/specs/roborev-review-guard/spec.md
@@ -62,11 +62,29 @@ The census SHALL be partitioned into a CODE subset and a non-code subset by the 
 changed), while `code-free:` is decided by the non-code count equalling the total and `prompt-content:` is
 asserted over the CODE subset alone.
 
+The census range's BASE SHALL be the **MERGE-BASE** of the base ref and HEAD, resolved ONCE and reused:
+`<base>...HEAD` is by definition `merge-base(<base>, HEAD)..HEAD`, so the base ref's TIP is NOT the base of
+the range under review and SHALL NOT be used as one. The wrapper SHALL carry the two as SEPARATE named
+values — the range base (used by the census, by `sha-assert:` and by the absence-waiver scope) and the base
+ref's tip (used only where the tip itself is the subject, i.e. the ROOT-checkout signature in
+`sha-assert:`) — and the census diff SHALL be pinned to the resolved range base, so one read of a moving
+mirror ref cannot leave the census measuring one range while the assert expects another.
+
 An unmeasurable census SHALL fail closed and SHALL be DISTINGUISHABLE from an empty one:
-`FAIL (base '<ref>' unresolvable)` when the base ref does not resolve to a commit, and
-`FAIL (git diff failed)` when the diff command itself exits non-zero. Neither SHALL be aliased to
-`FAIL (empty census)` / `NOTHING-TO-REVIEW`, because "we could not tell" is not "there is nothing to
-review". The wrapper SHALL NOT fetch on the caller's behalf to repair an unresolvable base.
+`FAIL (base '<ref>' unresolvable)` when the base ref does not resolve to a commit,
+`FAIL (no merge-base between '<ref>' and HEAD)` when the two have no common ancestor,
+`FAIL (merge-base of '<ref>' and HEAD unusable)` when the merge-base command succeeds without yielding
+exactly one 40-hex sha, and `FAIL (git diff failed)` when the diff command itself exits non-zero. None
+SHALL be aliased to `FAIL (empty census)` / `NOTHING-TO-REVIEW`, because "we could not tell" is not "there
+is nothing to review". An unresolvable merge-base SHALL NOT be degraded to the base ref's tip or to an
+empty value — a permissive branch here SHALL be keyed on the AFFIRMATIVE value (one 40-hex sha), never on
+the absence of a non-zero exit. The wrapper SHALL NOT fetch on the caller's behalf to repair an
+unresolvable base.
+
+#### Scenario: An unresolvable merge-base fails closed before anything is enqueued
+- **GIVEN** a repository whose base ref and HEAD have NO common ancestor (unrelated histories)
+- **WHEN** the wrapper computes the census
+- **THEN** `census-check:` reads `FAIL (no merge-base between '<ref>' and HEAD)`, the message states that this is explicitly NOT `NOTHING-TO-REVIEW` and that the base is deliberately not degraded to the tip, no review is enqueued, and the terminal `RESULT:` is `FAIL`
 
 #### Scenario: An unresolvable base ref fails closed rather than reporting nothing to review
 - **GIVEN** a clone whose `origin/main` mirror ref does not resolve (a narrow fetch refspec that has never fetched it)
@@ -576,7 +594,10 @@ census base.
 
 The value set SHALL be:
 
-- `PASS` — range head == branch HEAD AND range base == the resolved base sha.
+- `PASS` — range head == branch HEAD AND range base == the resolved RANGE BASE, i.e. the MERGE-BASE of
+  the base ref and HEAD. It SHALL NOT be the base ref's TIP: the reviewed range is merge-base-relative, so
+  comparing its base endpoint against the tip made a CORRECT review FAIL deterministically for every branch
+  whose base ref had advanced since the branch point (issue #3392).
 - `FAIL (reviewed range does not match <base>...HEAD)` — either endpoint disagrees, with the message
   naming WHICH: a range BASE of git's empty-tree hash SHALL be named as the signature of the
   non-sanctioned two-positional form, and a range HEAD short of branch HEAD SHALL be named as a reviewed
@@ -587,8 +608,11 @@ The value set SHALL be:
   passes every path check while the earlier changes go unreviewed. The sanctioned invocation always
   records a range, so a single sha means something else ran.
 - `FAIL (reviewed-sha does not match head-sha)` — a single-commit record that is not branch HEAD;
-  attributed when it equals the base ref (the signature of `--branch` resolved against the ROOT checkout)
-  and otherwise named as matching neither endpoint.
+  attributed when it equals the base ref's TIP (the signature of `--branch` resolved against the ROOT
+  checkout, which enqueues that tip), attributed distinctly when it equals the MERGE-BASE (the branch point
+  itself, so every commit under review is a descendant of it), and otherwise named as matching neither
+  endpoint. Both equalities SHALL FAIL, and the message SHALL name WHICH one matched — when the branch is
+  not behind its base the two are the same commit and the message SHALL say so.
 - `FAIL (job record unavailable — reviewed range unverifiable)` — no `git_ref` after the bounded read.
   This SHALL FAIL rather than fall back to prose: for a RANGE review the stdout announcement names only
   the range BASE, so it cannot establish that branch HEAD was reviewed at all, and a fallback to it would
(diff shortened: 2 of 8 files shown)


pmcfadin:
## Live usage datum for the "the workaround already exists" argument

Contributed by lane-1701, who is using the resolved-base invocation for its own final review round on PR #3397:

```
job 48, enqueued as: Reviewing branch ... 30 commits since 99b52d1bd3cd...
invocation:          --base "$(git merge-base HEAD origin/main)"
```

So the resolved base carries through to the enqueue against the **shipped** wrapper, end to end — which is the empirical complement to the mechanism I verified statically (`--base) BASE="$2"` at `:518`; `rev-parse(mb) == merge-base(mb, HEAD) == mb`; two-dot and three-dot both 5 commits).

This is evidence **for** the change, not against it, for the reason in the PR body: the safe form working today is exactly why it should be the default rather than something each caller must remember. And the operator needed a one-line annotation beside their pasted block to stop a 40-hex `base:` reading as a stale-base review — which is the legibility argument with a live example rather than a reasoned one. `assert-base:` exists so a pasted block is legible without its author present.

Scoped deliberately: measured against `main` at `99b52d1bd`.

pmcfadin:
## Gate of record: **PASS** (30/30)

Run on the pushed head, tree frozen. `tooling-tests: PASS (634s)` — the component behind the 14 kernel OOMs — ran clean on a tree rebased onto post-#3407 `main`.

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.0dbDy6
commit: 6555a39 branch: issue-3392-roborev-sha-assert-mergebase dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-3392/test-data/schemas (checkout-relative)
tree-start: 6555a3954f08 dirty: no digest: 8d97ac7093ac
tree-end: 6555a3954f08 dirty: no digest: 8d97ac7093ac
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (5s)
clippy:            PASS (4s)
roborev-lints:     PASS (43s)
core-tests:        PASS (49s)
tombstones-scan:   PASS (0s)
scan-offload-guard: PASS (3s)
byte-budget-guard: PASS (13s)
arrow-parity-guard: PASS (1s)
memory-budget:     PASS (10s)
integration-tests: PASS (0s)
format-compat:     PASS (1s)
write-tests:       PASS (9s)
cli-tests:         PASS (5s)
python-bindings:   PASS (66s)
node-bindings:     PASS (116s)
delivery-telemetry: PASS (0s)
oom-audit:         PASS (0s)
parity-report:     PASS (1s)
tooling-tests:     PASS (634s)
minimal-build:     PASS (82s)
smoke:             PASS (3s)
logs: /tmp/agent-gate.0dbDy6
summary-file: /tmp/claude-1000/-data-lanes-lane-3392/1c9892b3-f554-4f41-ab0f-362a6b8913cc/scratchpad/full2.txt
RESULT: PASS
```

**Note on the run-id:** an earlier attempt (`/tmp/agent-gate.NHrsgZ`) reached 28/30 and was killed by a peer lane's `pkill -f agent-gate.sh` sweep. It produced **no terminal emit and therefore no verdict** — 28 green components with no `summary.txt` is not a certification, and it is not being cited as one. The block above is a complete, independent re-run (`/tmp/agent-gate.0dbDy6`).

Issue #3428, opened 2026-08-28
Full-ring query-row scan is uncancellable across regions that emit no rows (all-tombstoned / TTL-expired): abandoned do_get can decode the whole table


pmcfadin:
Found by roborev while reviewing #3384 (job 62, PR for the flight test-determinism fix). **Not a test defect — a real cancellation gap in the read path**, in the class #1473 and #2264 exist to prevent.

## The gap

`drive_full_scan_rows` (`cqlite-core/src/storage/sstable/reader/data_access/summary_scan/query_rows.rs`) is the full-ring arm of `QueryRowStream`. It opens the inner scan with:

```rust
let mut rx = reader.scan_stream_batched_admitted(
    table_id, None, None, Some(schema),
    QUERY_ROWS_FULL_SCAN_BUFFER_ROWS, ScanAdmission::Exempt, Some(now_secs),
);
```

**The child cancel flag is not passed in, and the inner scan does not poll one.** Cancellation is observed only in the `while let Some(msg) = rx.recv().await` loop — i.e. only BETWEEN batches the inner scan actually delivers.

That is fine while decoded partitions yield rows: each one pushes the bounded handoff channel toward the parked state where the disconnect is seen. But over a region that survives **nothing** — every row tombstoned, or every row TTL-expired at the read clock — the inner scan emits no batch at all. The producer never parks, `recv()` never returns, and the scan can keep decoding **to the end of the table** after the client has gone.

## Why it matters

This is exactly the failure #1473 describes: a client disconnects mid-`do_get` and the server keeps doing the whole table's work, holding a blocking-pool thread and its working memory. The cooperative-cancellation machinery (`CancelFlag`, `ScanCancel`, the `CancelBridge`) is all present and correct — it simply is not threaded into this one arm's inner scan. A heavily-tombstoned or TTL-expired table is not an exotic shape in Cassandra; it is the normal state of a queue-like or time-series table.

## Consequence already absorbed elsewhere

#3384 exports `QUERY_ROWS_MAX_READ_AHEAD` as the bound on how much work an abandoned walk may still do. Because of this gap that bound holds only under the precondition "decoded partitions yield emitted rows", which is now documented as a precondition on the constant rather than papered over. Closing this issue would let that precondition be dropped.

The #3384 test is unaffected: its fixture is 19488 live single-row partitions with no tombstones or TTL, so every decoded partition emits.

## Suggested direction

[...]

PR #3433, opened 2026-08-28
fix(#3384): the flight suite's flake was an unbounded race, not cross-binary residue — bound it structurally

### LINKED ISSUE #3383: Flaky: cqlite-flight fast_arm_stream_stops_when_the_client_drops_it asserts a race outcome (2 of 3 failures under intra-package parallelism; 3/3 pass standalone)
Surfaced by #1699 (AH6), which adds a `flight-tests` gate lane running `cargo test -p cqlite-flight` — the
same command CI's Flight tier already runs (`flight-ci.yml:229`). The test is **pre-existing and untouched**
by #1699.

## The flake

`cqlite-flight/tests/issue_3058_bypass_path_taken.rs:474`,
`fast_arm_stream_stops_when_the_client_drops_it`:

```
thread 'fast_arm_stream_stops_when_the_client_drops_it' panicked at
cqlite-flight/tests/issue_3058_bypass_path_taken.rs:474:5:
the abandoned scan must not have drained the whole fixture: decoded 800 of 800 partition bodies
test result: FAILED. 5 passed; 1 failed
```

The assertion is a **race outcome**: the test drops a client stream and requires the producer to have stopped
early — i.e. to have decoded fewer than all 800 partition bodies. Under contention the producer wins the
race and drains the fixture before the drop propagates, so `800 of 800` is not a wrong *result*, it is the
producer being fast relative to cancellation.

## Measured rates (this worker box, 16 cores, load average 50–75 from five concurrent lane worktrees)

| condition | result |
|---|---|
| standalone `cargo test -p cqlite-flight --test issue_3058_bypass_path_taken`, 3 consecutive runs | **3/3 PASS** |
| inside the full package via the gate lane (`--only flight-tests`) | **2 of 3 runs FAILED** |
| bare `cargo test --no-fail-fast -p cqlite-flight` | 1 run, passed |

So it is **not** whole-box load alone — standalone it passes even at load 74. The trigger is **intra-package
parallelism**: `cargo test` runs ~41 test binaries concurrently, each defaulting to `num_cpus` threads, and
that contention lets the producer outrun the cancellation. The gate lane passes no `--test-threads`, so it is
running exactly what CI runs.

## Why it matters

The same flake is **already reachable in CI today**: the Flight tier's full job runs
`cargo test --package cqlite-flight`, and `required` fails closed on that tier (#2910). CI boxes are less
contended, so it presumably fires rarely there. #1699 does not create the risk — it surfaces it locally and
far more often, on a box where a worker is running several lanes.

At a measured ~2-in-3 failure rate under realistic worker-box load, a gate lane containing this target is not
usable as a merge gate: its red carries no information.

## Note on the repo's existing mechanization

`scripts/tests/check-no-wallclock-asserts.sh` (the #2642 lint, run in `roborev-lints`) does not catch this:
the lint targets **wall-clock threshold** asserts, while this is a **concurrency race** with no time
comparison in it. Worth considering whether the lint's charter should extend to "assert that a cancellation
beat a producer", which is the same class of unfounded timing assumption expressed without a clock.

## Suggested fix (owner/#3058 area, not #1699)

Make the assertion causal rather than temporal — e.g. have the producer signal observable progress and assert
cancellation was *requested* and *honoured* (the stream stopped, no further batches after the drop), instead
of asserting a partial byte/partition count that only holds when the consumer wins a race. If a bound is
genuinely wanted, it needs to be derived from the cancellation mechanism's guarantees, not from "fewer than
all".

### LINKED ISSUE #3384: cqlite-flight's whole-package test run is non-deterministic (~50% red): concurrency assertions fail only in full-suite context, not standalone
Surfaced by #1699 (AH6) while trying to add a local gate lane that runs `cqlite-flight`'s tests. Sibling of
#3383 (the first individual flake found) — **this issue is the general defect; #3383 is one symptom.**

## The defect

`cargo test -p cqlite-flight` (the whole package, ~41 integration binaries in one invocation) is
**non-deterministic**. Individual concurrency/timing tests fail intermittently, but only in the
full-package context. Two distinct victims found in four runs:

- `issue_3058_bypass_path_taken.rs:474` `fast_arm_stream_stops_when_the_client_drops_it`
  ("the abandoned scan must not have drained the whole fixture: decoded 800 of 800") — #3383
- `issue_2370_gauge_readback_test.rs:147` `phase_active_and_in_flight_read_true_concurrent_count_then_settle`
  (7 `tokio-rt-worker` threads panicked)

Observed lane outcomes across four runs: PASS, FAIL, PASS, FAIL — roughly 50%.

## What the trigger is NOT — each ruled out by measurement, not assumption

| hypothesis | test performed | result |
|---|---|---|
| whole-box load | 3 standalone runs of the victim at load average 74 | **3/3 PASS** |
| `nice` deprioritisation (the gate runs under `nice`/`taskpolicy`, #2640) | 2 standalone runs under `nice -n 19` | **2/2 PASS** |
| intra-binary thread parallelism | 2 standalone runs with `-- --test-threads=2` | **2/2 PASS** |
| MAIN-lane compilation competing with the SIDE lane | the failures occurred under `--only flight-tests`, where the MAIN lane runs nothing | **ruled out** |

Six standalone runs of the second victim pass under every combination of priority and thread count, at the
same box load that produced the in-lane failures. The failures appear **only** when the whole suite runs in
one `cargo test` invocation.

## Hypothesis (stated as a hypothesis, not a finding)

Cumulative cross-binary state within a single suite run — lingering tokio runtimes, un-awaited server tasks,
or bound ports from **earlier** test binaries — perturbs the timing that later concurrency assertions depend
on. `cargo test` runs test binaries sequentially, so this is not concurrent execution of the binaries
themselves; it is residue from predecessors. Two cheap next steps: run the suite with the target order
reversed and see whether the victim set changes; and check whether the Flight test harnesses await/abort
their server tasks and release their listeners on drop.

## Why it matters

1. **CI is already exposed.** `flight-ci.yml:229` runs the identical `cargo test --package cqlite-flight`
   in the Flight tier's full job, and `required` fails closed on that tier (#2910). Less-contended CI boxes
   presumably hit it rarely, but the exposure is structural, not incidental.
2. **It blocked #1699's `flight-tests` gate lane.** A merge-gate component that reds ~50% of the time
   carries no information, and quarantining victims one at a time is not a converging series (two found in
   four runs). #1699 therefore narrowed its lane rather than ship a coin-flip gate, and recorded the descope.
3. **The assertions themselves are suspect independently of the residue.** Both victims assert *race
   outcomes* — that a cancellation beat a producer, that a gauge settled — without any mechanism guaranteeing
   the ordering they require. Note `scripts/tests/check-no-wallclock-asserts.sh` (the #2642 lint in
   `roborev-lints`) does **not** catch these, because they contain no clock comparison: the unfounded timing
   assumption is expressed structurally instead. Worth considering whether that lint's charter should extend
   to this shape.

## Suggested direction

Make the assertions causal rather than temporal — assert that cancellation was *requested and honoured* (no
further batches after the drop) rather than that a partial count was observed — and give the Flight test
harnesses deterministic teardown (await/abort server tasks, drop listeners) so a predecessor cannot perturb a
successor. Once the suite is deterministic, #1699's lane can be widened back to the whole package; the
mechanism to do so is already in place (its target list is derived from `cargo metadata`, so widening is a
one-line change to the quarantine, not new machinery).

pmcfadin:
Fixes #3384. Also fixes #3383, which is one symptom of it. Spawns #3428 and #3434.

## Both issues' stated mechanisms were wrong, and measuring first changed the fix

#3384 blamed **cumulative cross-binary state**; #3383 blamed **intra-package parallelism**. Both were explicitly hypotheses. Neither survived measurement.

| measurement | result |
|---|---|
| per-binary `pid`/`start`/`end` timeline, 10 whole-package runs | **strictly sequential, zero overlap** — cargo does not run test targets concurrently, so #3383's mechanism does not exist |
| whole package | 3 fail / 9 (33%) |
| the victim test **alone in its own process**, all siblings filtered out | **6 fail / 20 (30%)** |
| `issue_2370_gauge_readback_test` alone | 0 / 20 |
| `issue_2821_egress_budget_e2e` alone | 0 / 20 |
| CPU affinity 16 vs 8 cores | 3/10, 1/10 vs 2/10 — no material change |

The victim fails at the same rate with **nothing else in its process**. The "3/3 PASS standalone" in the issue was a small-sample artifact — at 25% that happens 42% of the time.

## The defect: a bound that was never a bound

The test asserted `settled < PARTITIONS` over an 800-row fixture — that a cancellation *beat* a producer. The fast arm reads ahead through four bounded buffers totalling **3072 rows**, nearly 4x that fixture, so the producer never parked. Every captured failure was the saturated `800 of 800`, 7 of 7: a race with no bound can only finish or be interrupted.

`CANCEL_EGRESS_CEILING`'s doc claimed it made the bound hold "by construction". It did not — it meters Arrow batches *downstream* of the read-ahead.

## The fix

1. **Name the bound**, derived from the sizings that actually run (`QUERY_ROWS_MAX_READ_AHEAD`), with `batched_channel_capacity` as one `const fn` shared by the channel constructor and the bound so the arithmetic cannot drift. Unit tests pin each term.
2. **Assert the property that is true**: the walk **stopped** — proven by three causal signals, not by a counter looking still — and stopped **within the read-ahead**, a constant independent of table size. A `const` assertion keeps the fixture at >= 4x the ceiling, so drift fails the **build**.
3. **Fix a latent readiness TOCTOU** shared by two e2e files: `start_server` returned on the first successful connect, so a child dead of `EADDRINUSE` plus a foreign listener read as "ready". The server now **binds explicitly and announces after**, and readiness keys on that line. Extracted to a shared helper so the fix landed once.

## Three causal signals, because two were not enough

The abandoned walk has three producers and each needed its own proof of termination:

| producer | signal |
|---|---|
| outer query-row thread | `query_row_producers_finished`, published before ANY consumer-observable terminal message |
| the scan's detached future | `ScanTaskDone` drop guard, owned by the task so an unpolled future still publishes |
| stitching path's `spawn_blocking` parse/feed halves | `blocking_scan_tasks_inflight()` gauge; the reader waits for ZERO |

**There is no sleep-based reasoning left in the correctness path.** Every RMW on the gauge is `AcqRel` so each decrement acquires the ones before it — a plain `Release` would leave an earlier task's work unordered against a reader that observes zero, which is exactly the finality the gauge exists to provide.

## Evidence

- **30/30** isolated repeats (was 6/20 red), observed read-ahead **768..1536** against the **6152** ceiling.
- Fixture is 31x larger (800 -> 24608 partitions) and the binary is faster (~4s -> 0.5s).
- Full agent-gate **PASS, 30/30 components** (at `0230ec059`; being re-run at the final head).
- roborev **PASS, 0 findings** at `35352de71` — the head that will actually merge, re-run after the rebase because resolving its conflict produced code no review had seen. Thirteen prior rounds produced **seventeen** genuine findings.

## Review history — where most of the value came from

Three of those findings were **false-PASS windows in the very test this issue exists to make trustworthy**: completion published after a terminal message a caller acts on; a counter read while a detached task was still decoding; a gauge that counted a queued task as zero. Two were concurrency defects in the instrumentation itself — an unsound `Release`/`Acquire` chain, and a guard that could never fire if a future was dropped before its first poll.

Two of my own published claims were **wrong and are corrected on the issue**: that the observed max landing exactly on 1280 confirmed the arithmetic (an artifact of reading too early), and that the fixture takes the non-stitching path (an affirmative probe returned the opposite). Both are the defect this issue is about — a proxy that sounds right is not a measurement.

## What this PR deliberately does NOT do

**`issue_2370_gauge_readback_test` is untouched.** Unreproduced in 40 runs, and the one mechanism I could construct is falsified (`CqliteFlightService::new` uses `Admission::unconstrained()`). A speculative fix would read as "flake reduced" while proving nothing.

**#3428** carries the real structural work: the full-ring inner scan never receives the child cancel flag (an all-tombstoned region can decode to the end of the table); `parse_batched_block`/BTI materialize eagerly; and the blocking halves are never joined on cancellation. No test covers the BTI or stitching cancellation paths — that absence is why these were review findings rather than failures.

**#3434** carries the fleet capacity defect found while gating this: disk exhaustion produces a **fake red**, not a queue.

## Notes for the reviewer

- **No `CQLITE_ALLOW_FILE_GROWTH` anywhere.** Net growth against the merge base is zero for every over-threshold file; the sizing constants moved to `query_rows_bounds.rs` rather than being trimmed, because the prose is the safeguard.
- Commit `e13a6751e` is titled `docs(...)` but carries the `const` drift guard. Mislabelled; content is in the diff.
- Once this lands, #1699's `flight-tests` lane can widen to the whole package. Its descope cited intra-package parallelism, which does not exist.

### DIFF
--- a/cqlite-core/src/storage/read_path_probe.rs
+++ b/cqlite-core/src/storage/read_path_probe.rs
@@ -138,3 +138,50 @@ mod tests {
         assert!(!delta.any_merge_work());
     }
 }
+
+// ---------------------------------------------------------------------------
+// Query-row producer completion (issue #3384)
+// ---------------------------------------------------------------------------
+
+/// Producers that have finished decoding, published with RELEASE ordering.
+///
+/// Deliberately NOT a field of [`ReadPathProbe`]: every field there is loaded
+/// `Relaxed`, which is right for counters read after the work is known to be over,
+/// and wrong for this one — its whole job is to TELL you the work is over.
+static QUERY_ROW_PRODUCERS_FINISHED: AtomicU64 = AtomicU64::new(0);
+
+/// Publish that one `QueryRowStream` producer has finished decoding.
+///
+/// Call BEFORE the terminal message is sent, never after (roborev, issue #3384):
+/// a consumer holding the terminal message must be able to conclude that this
+/// producer can no longer publish, or a producer from a PRIOR test case can
+/// increment into a LATER case's freshly-reset counter and the later case will
+/// observe a completion that was never its own.
+pub fn mark_query_row_producer_finished() {
+    QUERY_ROW_PRODUCERS_FINISHED.fetch_add(1, Ordering::Release);
+}
+
+/// `QueryRowStream` producers that have finished decoding since the last
+/// [`reset_query_row_producers_finished`] (issue #3384).
+///
+/// The CAUSAL completion signal for an abandoned walk. A test that instead waits
+/// for a work counter to "stop growing" is sampling another thread's progress: a
+/// producer merely descheduled, or paused between two increments, is
+/// indistinguishable from one that has stopped, so the walk can resume and drain
+/// the table right after the assertion passed.
+///
+/// The `Acquire` load is load-bearing, not decoration. With `Relaxed` on both
+/// sides, observing a non-zero count would establish NO happens-before edge with
+/// the producer's earlier work-counter increments, so a reader could see
+/// "finished" and then read a STALE row count — precisely the guarantee this
+/// signal exists to give (roborev, issue #3384).
+pub fn query_row_producers_finished() -> u64 {
+    QUERY_ROW_PRODUCERS_FINISHED.load(Ordering::Acquire)
+}
+
+/// Zero the completion count. A test that will wait on
+/// [`query_row_producers_finished`] must call this first, while holding whatever
+/// lock serializes producers in its binary.
+pub fn reset_query_row_producers_finished() {
+    QUERY_ROW_PRODUCERS_FINISHED.store(0, Ordering::Release);
+}
--- a/cqlite-core/src/storage/read_path_probe.rs
+++ b/cqlite-core/src/storage/read_path_probe.rs
@@ -156,6 +156,33 @@ pub fn query_row_producers_finished() -> u64 {
     QUERY_ROW_PRODUCERS_FINISHED.load(Ordering::Acquire)
 }
 
+/// Detached `BatchedScanStream` tasks that have terminated, published with
+/// RELEASE ordering (issue #3384).
+static BATCHED_SCANS_FINISHED: AtomicU64 = AtomicU64::new(0);
+
+/// Publish that one detached batched-scan task has terminated.
+///
+/// The INNER half of the abandoned-walk completion story. `mark_query_row_producer_finished`
+/// proves the outer query-row thread stopped pulling; this proves the detached task it
+/// dropped without joining has stopped DECODING. Only both together make
+/// `stream_walk_partitions_parsed` final — waiting a fixed interval instead merely
+/// makes the race less likely, which is the defect this issue exists to remove.
+pub fn mark_batched_scan_finished() {
+    BATCHED_SCANS_FINISHED.fetch_add(1, Ordering::Release);
+}
+
+/// Detached batched-scan tasks that have terminated since
+/// [`reset_batched_scans_finished`] (issue #3384). `Acquire`, for the same
+/// happens-before reason as [`query_row_producers_finished`].
+pub fn batched_scans_finished() -> u64 {
+    BATCHED_SCANS_FINISHED.load(Ordering::Acquire)
+}
+
+/// Zero the detached batched-scan completion count (issue #3384).
+pub fn reset_batched_scans_finished() {
+    BATCHED_SCANS_FINISHED.store(0, Ordering::Release);
+}
+
 /// Zero the completion count. A test that will wait on
 /// [`query_row_producers_finished`] must call this first, while holding whatever
 /// lock serializes producers in its binary.
--- a/cqlite-core/src/storage/read_path_probe.rs
+++ b/cqlite-core/src/storage/read_path_probe.rs
@@ -36,7 +36,7 @@
 //! must serialize against any sibling test in the same binary that also merges
 //! (one file = one process is the strongest form; a mutex is the minimum).
 
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicI64, AtomicU64, Ordering};
 
 static MERGERS_BUILT: AtomicU64 = AtomicU64::new(0);
 static RECONCILE_ENTRIES: AtomicU64 = AtomicU64::new(0);
@@ -154,6 +154,50 @@ pub fn mark_query_row_producer_finished() {
 /// signal exists to give (roborev, issue #3384).
 pub fn query_row_producers_finished() -> u64 {
     QUERY_ROW_PRODUCERS_FINISHED.load(Ordering::Acquire)
+}
+
+/// Blocking scan tasks currently running (issue #3384). A GAUGE, not a counter.
+static BLOCKING_SCAN_TASKS_INFLIGHT: AtomicI64 = AtomicI64::new(0);
+
+/// RAII marker for one blocking scan task: increments on construction, decrements on
+/// drop (issue #3384).
+///
+/// The stitching path's parse/feed halves run under `spawn_blocking`, and blocking
+/// tasks are NOT cancellable — dropping their `JoinHandle` DETACHES them. So neither
+/// the outer query-row producer's completion nor the scan future's drop guard proves
+/// decoding has stopped: both can fire while a detached blocking half is still
+/// draining. This gauge is the missing half. A gauge rather than a completion counter
+/// because the number of blocking halves is an implementation detail — a reader waits
+/// for it to reach ZERO instead of knowing how many to expect.
+///
+/// Decrement is a `Drop` impl so an unwinding task still clears its slot; a leaked
+/// increment would hang every future reader.
+pub struct BlockingScanTaskGuard;
+
+impl BlockingScanTaskGuard {
+    /// Register one running blocking scan task.
+    pub fn new() -> Self {
+        BLOCKING_SCAN_TASKS_INFLIGHT.fetch_add(1, Ordering::Release);
+        Self
+    }
+}
+
+impl Default for BlockingScanTaskGuard {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Drop for BlockingScanTaskGuard {
+    fn drop(&mut self) {
+        BLOCKING_SCAN_TASKS_INFLIGHT.fetch_sub(1, Ordering::Release);
+    }
+}
+
+/// Blocking scan tasks still running (issue #3384). ZERO means every detached
+/// blocking half has stopped, so `stream_walk_partitions_parsed` is final.
+pub fn blocking_scan_tasks_inflight() -> i64 {
+    BLOCKING_SCAN_TASKS_INFLIGHT.load(Ordering::Acquire)
 }
 
 /// Times a batched scan took the STITCHING branch (issue #3384).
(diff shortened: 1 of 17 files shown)


pmcfadin:
## C — intent audit against this issue's acceptance criteria

**Provenance caveat first, because it changes how much this is worth.** I ran this audit MYSELF. I spawned `spec-auditor` for an independent read and it went idle twice without reporting — the fifth subagent to do so this session. A self-audit is weaker evidence than an independent one by exactly the amount you would expect, and I am not going to present it as though it were not. Everything below is checkable from the diff.

Routing is oracle-driven, so there is no OpenSpec change; its absence is correct, not a gap.

### Requirement 1 — "make the assertions causal rather than temporal": **SATISFIED**

The correctness path now turns on three completion signals, one per producer, and no elapsed duration is compared against a threshold anywhere in it.

| producer | signal | evidence |
|---|---|---|
| outer query-row thread | `query_row_producers_finished` | published before ANY consumer-observable terminal message, including `QueryRowBatch::Unsupported` |
| detached scan future | `ScanTaskDone` drop guard | owned by the task, not the body, so an unpolled future still publishes |
| `spawn_blocking` parse/feed halves | `blocking_scan_tasks_inflight()` gauge | reader waits for ZERO; `AcqRel` on both RMWs so the chain is sound |

The only `sleep`s left are 1 ms **poll intervals** inside bounded loops whose exit condition is causal. Their expiry does not degrade to a pass — `stopped`, `inner_done` and `blocking_done` are each asserted — so a slow machine **fails loudly**. That is the distinction the requirement is about.

### Requirement 2 — "deterministic teardown so a predecessor cannot perturb a successor": **PARTIAL, and I am not rounding it up**

Delivered as written? **No.** The requirement rests on a premise measurement falsified: there is no cross-binary residue, because cargo runs test binaries strictly sequentially and the victim fails at the same rate alone in its own process. Nothing was perturbing a successor, so "deterministic teardown" had nothing to fix.

What was delivered instead is a real defect of the same family — `start_server` returned on the first successful connect, so a child dead of `EADDRINUSE` plus a foreign listener read as "ready" and the test scanned the wrong server. That is a genuine cross-test interference bug, and fixing it serves the requirement's *intent*. But it is not the thing the requirement asked for, and one thing the requirement implies is still **not done**: the Flight harnesses do not await or abort their server tasks on drop. `BatchedScanStream` is still dropped without joining, which is precisely why three of the review findings existed and why #3428 is open.

**Verdict: partial.** The intent is served; a literal reading is not; the residue the requirement targeted does not exist.

### Stated goal — "the suite becomes deterministic so #1699's lane can widen": **SATISFIED for the reproducible defect, with a caveat**

30/30 isolated repeats, was 6/20 red. But the goal says *the suite*, and I have only fixed the one victim that reproduces. `issue_2370_gauge_readback_test` (0/40) is untouched. If it recurs, the lane is red again and this issue's goal is not met.

**Descope defensible?** Yes, but only because the alternative is worse: it did not reproduce in 40 runs and the one mechanism I could construct is falsified (`CqliteFlightService::new` uses `Admission::unconstrained()`). A speculative fix would read as "flake reduced" while proving nothing — the exact pattern this issue criticises in others. Recorded on the PR rather than buried.

### Scope creep — **REAL but justified, one item borderline**

Production changes, each judged separately:

- `QUERY_ROWS_MAX_READ_AHEAD` + `query_rows_bounds.rs` — **necessary.** The test cannot state a structural bound without the constant, and deriving it in the test would be the magic number the issue is about.
- The three probe signals — **necessary**, and consistent with `read_path_probe`'s existing charter, which already exists so tests can observe markers instead of inferring from timing. Test-facing observability, not a feature: **no production caller**. Stated plainly rather than dressed as wiring evidence.
- `main.rs` explicit bind + `cli::log_listening` — **necessary** for the readiness fix, and independently an operational improvement (the log previously said "starting" before the port was held).
- `batched_channel_capacity` moved beside its constructor — **borderline.** It was a file-size remediation for a change that itself grew the file. Defensible, but it is the one item I would drop if a reviewer pushed.

### Overstatements — one found, now corrected; one wording I have tightened

I asked myself to assume a third overstatement existed after correcting two. The candidate: the PR said "**no sleep-based reasoning** in the correctness path" while three 1 ms `sleep`s remain. On inspection those are poll intervals with causal exit conditions and fail-closed bounds, so the claim is true as meant — but "no sleeps" would have been false, and the distinction is exactly the kind I have been wrong about twice today. The PR body says "no sleep-based *reasoning*", which is accurate; I am flagging it so a reviewer reads it deliberately rather than skimming it as "no sleeps".

### Blocking anything?

No. One `partial` (requirement 2), honestly scoped, with the undelivered half tracked on #3428.

Issue #3387, opened 2026-08-27
Audit: ~696 `… | grep -q` sites under scripts/ decide branches from a pipefail/SIGPIPE-racy status (141 on a SUCCESSFUL match)


pmcfadin:
Split out of #3380 (whose guard-test failure is one instance) and #1699 (which had one in its own new
code) so neither scope has to widen.

## The class

Scripts here run under `set -uo pipefail`. An **early-exiting consumer** — `grep -q`, `grep -m N`,
`head` — closes the pipe as soon as it is satisfied; the upstream dies of **SIGPIPE**; and under
`pipefail` the pipeline's exit status becomes **141**. That is **non-zero on a SUCCESSFUL match**, so any
`if` / `if !` reading it can take the wrong branch.

It is timing-dependent — it fires only when the upstream still has buffered output when the consumer
exits — so it presents as an intermittent flake rather than a logic bug, and it survives code review
because the expression *reads* correctly.

## Two confirmed instances, in opposite directions

1. **#3380** — `test_roborev_review_guard.sh`'s #3312 structural assert. Instrumented: `hits=1` and
   `classifier=[]` in the same iteration, `q-pipeline-exit=141`. The guard **reported the token absent
   having found it** — a vacuous PASS in the guard that enforces the affirmative-measurement rule.
2. **#1699** — the `legacy-heuristics` allowed-zero derivation used `! sed … | grep -qE`, so a target
   with a surviving positive cfg site could be classified negative-polarity-only and **excused from the
   #2039 zero-tests guard**. Measured **6/6** mis-classifications with a match on line 1 of a 200k-line
   input, **0/6** after the fix. Fixed in `0c14d5e64`.

The second is the reason this is worth an audit and not a one-line fix: the shape converts a
**correctness guard into a silent excusal**, and the failure direction is not predictable from reading
the line.

## Census

```
$ grep -rnE '\|[[:space:]]*grep[[:space:]]+-[a-zA-Z]*q' scripts/ | wc -l
696
```

Not all 696 are defects. **The status has to be READ for the bug to bite**, so the triage is:

- **DEFECT** — the pipeline is the condition of an `if`/`while`/`&&`/`||`, or its `$?` is tested.
- **BENIGN** — the pipeline's *output* is captured (`x=$(… | head -1)`) and the status ignored.
- **BENIGN** — the consumer reads all input anyway (`| tail -1`, `| grep -c`, `| wc -l`).
[...]

Issue #3363, opened 2026-08-27
0.17 fan-out coordination: dispatch anchor for the agent fleet


pmcfadin:
Single comment anchor for the 0.17 multi-agent push. Not a scope epic: scope lives in the 0.17 milestone and the per-area epics (#2817 throughput program; #1685 AH config; #1686 AI observability; #1687 AJ schema; #1688 AK dead code; #1689 AL CLI; #1434/#1435/#1436 bindings X/Y/Z). Workers report blockers, escalations, and follow-up proposals HERE.

## Dispatch rules
- The board Ready column is the sole dispatch authority. Read it filtered: `gh project item-list 1 --owner pmcfadin --query "status:Ready" --format json -L 100`. Empty Ready = STOP, comment here.
- Claim before touching anything: `bash scripts/flow/claim.sh claim <N>`, then worktree `issue-<N>-<slug>` off origin/main (1:1:1:1). One worker per machine, one full gate at a time.

## Lanes
- **Generic workers**: hygiene Wave-1, currently Ready: #1697, #1695, #1699, #1701, #1705, #1451, #1458, #1712, #1714.
- **Rig-bound (metal/i4i box ONLY)**: #3299, #3248, #3287. Generic workers must not claim these.
- **#3350**: has a RESUME POINT comment; adopt branch `issue-3261-writability-precondition-pending-auth`, do not start cold.
- **Not claimable**: anything with a HOLD or RESERVED comment (#1696, #1452, #1455, #1461, #1707, #1711, #2818, #2820, #2824, #2826, #2828).

## Wave 2 (lead promotes as Ready drains)
#1696, #1455, #1452, #1698, #1700, #1702, #1704, #1715.

## Process per issue
flow-implement loop: TDD, `--lite` each fix round (summary-file redirect), review-first (rust-reviewer + `scripts/flow/roborev-review.sh --agent codex --model gpt-5.6-sol`), open PR, then flow-closer runs the ONE full gate + C audit + final roborev + `gh pr merge --auto --squash --delete-branch` + finalize/telemetry. Datasets: use the exact export line `bash test-data/scripts/fetch-datasets.sh` prints. Blocked on an owner decision: park (structured question comment + `needs-decision` label + exit), never wait.

pmcfadin:
**Infra report — this box (`ip-172-31-15-234`) is DISK-STARVED with four concurrent lanes, and it is failing gates as a LINK error that looks like a code bug.**

[...]

PR #3437, opened 2026-08-28
fix(#3367): roborev-lints order-dependence — doctrine scans read an immutable wrapper path

### LINKED ISSUE #3367: roborev-lints reds at random: the classifier-GONE structural assert is order-dependent AND masks a deterministic true-positive on the wrapper's own doctrine prose
The `roborev-lints` gate component FAILed once on PR branch `issue-1714-delete-orphan-memory-safety-tests` at `8f5339c91`, then PASSed 4/4 on re-run **at the same commit with an unchanged tree**. The failing subject is a structural assert in `scripts/tests/test_roborev_review_guard.sh`, and the branch's diff touches **only** three files under `cqlite-core/tests/` — `git diff --name-only origin/main...HEAD -- scripts/` is empty — so the diff cannot be the cause.

Diagnosing it turned up a **latent true-positive that the flake is masking**, which is the more important half of this report.

## Observed

```
roborev-lints:     FAIL (39s)      # inside agent-gate.sh --lite at 8f5339c91
==== ROBOREV REVIEW GUARD TEST TALLY ====
passed: 740  failed: 1

FAIL - structural: delivery-mode classification is back in the flow scripts — mixed-delivery.
Owner ruling (4) deleted it because FOUR consecutive review rounds each found a
High-severity false verdict ... (#3312)
```

Standalone, same commit, same tree, four consecutive runs: `passed: 741  failed: 0`, `GUARD-TEST RESULT: PASS`.

## Cause 1 — the assert is ORDER/STATE dependent: `WRAPPER` is a mutable global

`scripts/tests/test_roborev_review_guard.sh`:

```
34:   WRAPPER="$SCRIPT_DIR/../flow/roborev-review.sh"   # the real wrapper
2187: WRAPPER="$_gm_dir/roborev-review.sh"              # reassigned to a scratch copy for one case
2253: WRAPPER="$_gm_real_wrapper"                       # restored
```

The classifier-is-GONE structural check (~4207) reads `"$ORACLES" "$CHECKS_FILE" "$WRAPPER"`. Its verdict therefore depends on **which value `WRAPPER` holds when it runs**, which is set by an unrelated earlier case ~2000 lines away. A test whose assert reads a global that other cases mutate is order-dependent by construction, and that is enough to explain a 1-in-5 verdict flip.

## Cause 2 — the "EXECUTABLE LINES ONLY" filter does not exclude the prose it was written to exclude

The check filters comments with `grep -hv '^[[:space:]]*#'` and documents the intent explicitly:

> EXECUTABLE LINES ONLY: a comment that RECORDS what was deleted (and why it may not come back) is the durable artifact of this ruling, so scanning prose would make the history itself a violation.

That intent is not achieved. `scripts/flow/roborev-review.sh:339` contains the token in a **continuation line carrying no `#` prefix**:

```
patched a fifth time. Block detection, heading parsing, fence evidence, mixed-delivery,
candidate lifetime and the snapshot/delegated distinction are all gone with it.
```

Run the check's own filter against the real wrapper:

```bash
$ grep -hv '^[[:space:]]*#' scripts/flow/roborev-review.sh | grep -cF "mixed-delivery"
1
```

So **against the real wrapper this assert fails deterministically** — the deleted-classifier history, which the ruling deliberately preserved as prose, is itself matched as a "live reference". The only reason the component usually passes is Cause 1: `WRAPPER` is usually pointing somewhere else when the check runs. Two defects that cancel, which is why nobody has seen this until a scheduling difference uncovered it.

This is the same shape #3312 kept producing and that this very guard exists to pin: **a decision made from a stream carrying both control tokens and prose payload.** A `#`-prefix test cannot separate them when the prose is a continuation line.

## Impact

- `roborev-lints` can red **any** PR's gate — including the FULL gate of record — for a reason unrelated to its diff. A component that reds at random is the component agents learn to waive, which is the failure mode this repo's doctrine warns about repeatedly.
- Fixing Cause 1 alone would make the component fail **deterministically** via Cause 2. **Both must be fixed together**, or the gate goes permanently red.

## Fix sketch (owner/lead to route)

1. Make the structural check independent of mutable global state: resolve the three paths locally from `$SCRIPT_DIR` at the point of use, or snapshot them into dedicated read-only variables at line 34 that no case reassigns.
2. Make the prose exclusion match its documented intent. Options, in my order of preference: (a) scan only lines that are plausibly executable via a real comment/heredoc-aware pass; (b) require the token to appear in an *assignment or invocation* context rather than as a bare substring; (c) keep the substring test but exclude the wrapper's doctrine block explicitly by delimiter. Option (b) is closest to "is this a live reference?", which is the actual question.
3. Add a regression case pinning **both** directions: the real wrapper's doctrine prose must NOT trip the check, and a genuine reintroduction (e.g. an assignment to `ROBOREV_DIFF_SOURCE_STATE`) MUST trip it. Today neither is pinned against the real file.

## Reproduction

```bash
# Cause 2, deterministic:
grep -hv '^[[:space:]]*#' scripts/flow/roborev-review.sh | grep -cF "mixed-delivery"   # -> 1

# Cause 1, the flip: run inside the gate vs standalone at the same commit
bash scripts/tests/test_roborev_review_guard.sh   # PASS standalone, 4/4 observed
```

## Routing
Oracle-driven — the oracle is the check's own stated intent plus the real wrapper's content. No OpenSpec.

## Acceptance criteria
- [ ] The structural check no longer reads a global that other cases reassign.
- [ ] The real `scripts/flow/roborev-review.sh` doctrine prose does not trip it (pinned by a test).
- [ ] A genuine classifier reintroduction still trips it (pinned by a test, observed firing).
- [ ] `roborev-lints` PASSes deterministically across repeated runs on an unchanged tree.

## Note
Filed by the #1714 drive-issue worker; **unlabeled and unprioritized** — a worker touches only `coord:*` labels. Not blocking #1714: it cleared on re-run, and #1714's diff provably cannot cause it. Recording it so the next agent that hits it has the diagnosis instead of a mystery red, and so nobody "fixes" Cause 1 in isolation and turns the gate permanently red.

pmcfadin:
Fixes #3367.

`roborev-lints` red at random on branches whose diff touched no `scripts/` path — observed on three unrelated lanes, each clearing on re-run.

## Cause

Two causes were diagnosed on the issue. **Cause 2 was already fixed on `main`** by #3392 (the `usage()` heredoc exemption); measured hit count on the real wrapper is now 0, and the SIGPIPE/pipe-buffer mechanism behind the reported 39/40 vacuous-pass rate went with it.

**Cause 1 was live, and had inverted.** `WRAPPER` is a global that two gate-mock cases repoint at a scratch copy ~2000 lines from the doctrine asserts, so a scan reading it was judged against whichever file the last case had left behind. With #3392's prose exemption in place the failure mode flipped: a scan landing on the scratch copy finds nothing and reports **CLEAN**. It surfaced as a false FAIL, so someone noticed; the same defect now yields a false PASS, which nobody would.

## The fix: the global is gone, not guarded

`WRAPPER` existed for exactly one reason — those two mock windows. `run_wrapper` was its only reader, and the 7 direct `bash "$WRAPPER"` sites all sat **outside** both windows, so they always meant the real wrapper.

- `run_wrapper` resolves its own path: `bash "${RUN_WRAPPER_PATH:-$WRAPPER_REAL}"`
- the two mock cases scope `RUN_WRAPPER_PATH` around themselves instead of mutating a global
- direct sites use `$WRAPPER_REAL`, which is `readonly`
- `WRAPPER` no longer exists

**The order dependence is absent by construction rather than policed.** What remains is one shell-enforced fact plus a two-part check: no mutable `WRAPPER` global exists, and the scratch-copy override is read at exactly one site.

## Why not the guard I built first

The first fix **guarded** the global: a ~250-line lexical scanner classifying every occurrence of `$WRAPPER` by quote state, arity, spelling, command position and markers. It went through **18 roborev rounds and 4 independent review passes, producing 43 findings, all valid, all fixed**. The last one settled it: `sh -c` executes a name assembled from adjacent quoted strings (`export WRAP''PER; sh -c 'grep "$WRAP''PER"'`), and this harness already uses `sh -c`. Banning `eval` did not close it, and the set does not close — `bash -c`, `source`, `.`, `xargs sh -c`.

> A lexical scanner cannot decide whether a shell line reads a variable, because the name need not appear in the line.

No further round could have fixed that, so the scanner, its 36 fixtures and its poison probe were deleted. **Net −770 lines**; this PR is now **+283/−69** rather than +900/−55.

Two patterns from those 43 findings are worth carrying forward, and are proposed for the roborev-findings doctrine page:
- **Seven were literally one mistake** — *a permitted shape found ANYWHERE standing in for the line BEING that shape* (unanchored `bash`; compound-line arity; `${seg%%=*}`; a declaration substring; two markers; the `$(`-only substitution test).
- **Three more were substring exclusions I added while fixing the previous round** — each exempting a real violation that merely mentioned the excluded token.

## Acceptance criteria

| AC | status | evidence |
|---|---|---|
| 1 — structural check no longer reads a reassigned global | ✅ | **there is no such global**; RED-verified that reintroducing `WRAPPER=` reds |
| 2 — real wrapper's doctrine prose does not trip it | ✅ | pinned with a vacuity guard (FAILs if the wrapper stops naming the retired states) |
| 3 — a genuine reintroduction still trips it | ✅ | pinned against a copy of the real wrapper, **plus AC3b** inside `usage()` before its heredoc |
| 4 — `roborev-lints` deterministic | ✅ | 8/8 PASS across consecutive gate runs on an unchanged tree |
| 5 — full gate PASS | ⏳ | blocked: box3 is reserved measurement-only for #3248, and box1/box2 are unreachable from here (no ssh, no key, no `agent-ami`, no off-box peer session) — see issue `3367-W3` |

Suite: **806/0 (`main`) → 813/0**.

## Gate of record

```
RESULT: PASS          run-id /tmp/agent-gate.xyZ2Ez   host box1
commit 80a96fb        tree-integrity: PASS            tree-start == tree-end (e360247c3266)
```

Run on box1 by the coordination lead — box3 was reserved measurement-only for #3248 (#3363), and box1/box2 are unreachable from box3 (no ssh, no key, no `agent-ami`, no aws credentials). Liveness was verified by counting processes with a `gate-3437` cwd, not by reading the summary sentinel or inferring from the queue.

## Two findings ship UNFIXED, under a pre-committed stopping rule

roborev job 65 returned two AC2/AC3-coverage findings on this sha. Both are fair; **neither is fixed**, under a rule the lead pre-committed *before* the answer was known:

> if that pass returns another Medium in AC3/AC2 coverage: we SHIP. The residual goes to #3454 with its mechanism recorded.

- **Medium** — the assignment detector misses `mode=$'mixed-delivery'` and `mode="mixed"-delivery`.
- **Low** — the AC3 control plants a tracked *identifier*, so it passes whether or not the new prose-assignment path works.

**Why a rule rather than a judgement:** three consecutive rounds each asserted a property of shell syntax that the author of the shell controls, and each was falsified by one more spelling.

| round | the claim | the spelling that broke it |
|---|---|---|
| (a) drop the 4 prose tokens | "a verdict string alone is not a classifier" | `delivery_mode=mixed-delivery` |
| (b) require an assignment context | "an assignment does not occur in prose" | `mode=$'mixed-delivery'` |

Option (b) was a genuine improvement — it closed the renamed-variable *class* rather than instances — and it still leaks. **That leak is the finding**, and it leads #3454: a token-name scan over shell source cannot decide whether a construct is live, so the question is what *channel* (a lint step, a registry, an assertion on the wrapper's output), not what regex. Structurally identical to #3312's four superseded recognisers.

**Guard reach today, so nobody over-trusts it:** catches any of the 13 code identifiers anywhere including inside the heredoc, and a verdict string assigned as `=value` / `="value"` — both RED-verified. Does not catch a reintroduction that renames every identifier *and* obfuscates the verdict string's quoting.

## Scope not taken

The prose-shaped-needle redesign (issue comments 3/5/6/7) is proposed as a separate issue (`3367-W1`). Residuals recorded there: AC2 resting on a single token, and the ~35 other `cmd | grep -q`-under-`pipefail` sites in this file (#3387 fleet-wide).

## Authorship

Two sessions briefly worked this lane concurrently (a #1930 violation caught by `tree-integrity` #2926 — issue comments 8–9). The peer stood down, released the claim and endorsed this line of work; their commit `13c1cc328` is retained as an ancestor. The claim was re-acquired via the documented `adopt --expect none` path.

### DIFF
--- a/scripts/tests/test_roborev_review_guard.sh
+++ b/scripts/tests/test_roborev_review_guard.sh
@@ -5002,7 +5002,28 @@ _WR_QS=$(cat <<'AWKEOF'
   # set, so `ok "\`grep \"a'b\" \"$WRAPPER\" \"c'd\"\`"` reproduced the same quote confusion and was
   # excused as prose. If a future shell grows a third, this line is where it goes.
   cs = (seg ~ /[$][(]/ || seg ~ /`/) ? 1 : 0
-  printf "%d %d %d %d %d %d %d\n%s\n", code, lit, q, mk, ev, bal, cs, seg
+  # A STATE-FREE COUNT over the WHOLE line, used only to detect that the stateful scan stopped early.
+  # A quoted `#` inside a command substitution -- `_x="$(printf " #"; grep -c foo "$WRAPPER")"` --
+  # makes the flat parser believe a comment started, so it BREAKS before the read and reports nothing
+  # at all. Comparing against a count that cannot be fooled by state is how that is caught.
+  # ESCAPES ARE HONOURED HERE TOO, though quote state is not. A backslash is lexically unambiguous —
+  # `\$WRAPPER` is never an expansion, in any context — whereas quote state is exactly what this
+  # counter exists not to trust. Ignoring escapes made every `bad "... \$WRAPPER ..."` diagnostic
+  # that also contains a $( ) look like a hidden read (measured: two false FAILs).
+  tot = 0; j = 1
+  while (j <= n) {
+    if (substr(line, j, 1) == "\\") { j += 2; continue }
+    if (substr(line, j, 1) == "$") {
+      r2 = substr(line, j)
+      if (match(r2, "^[$][{]" VN "[^A-Za-z0-9_]")) tot++
+      else if (match(r2, "^[$]" VN)) {
+        a2 = substr(r2, RLENGTH + 1, 1)
+        if (a2 !~ /[A-Za-z0-9_]/) tot++
+      }
+    }
+    j++
+  }
+  printf "%d %d %d %d %d %d %d %d\n%s\n", code, lit, q, mk, ev, bal, cs, tot, seg
 }
 AWKEOF
 )
@@ -5018,7 +5039,8 @@ _wr_classify() {
   _c_q=${_c_rest%% *}; _c_rest=${_c_rest#* }
   _c_mk=${_c_rest%% *}; _c_rest=${_c_rest#* }
   _c_ev=${_c_rest%% *}; _c_rest=${_c_rest#* }
-  _c_bal=${_c_rest%% *}; _c_cs=${_c_rest##* }
+  _c_bal=${_c_rest%% *}; _c_rest=${_c_rest#* }
+  _c_cs=${_c_rest%% *}; _c_tot=${_c_rest##* }
   # AN UNRELIABLE PARSE YIELDS NO VERDICT (reviewer). Checked FIRST: every field below is derived
   # from the same state machine, so if it ended the line inside a quote, none of them can be trusted.
   # DEFAULTS TO UNTRUSTWORTHY, not to balanced. This is the one field whose job is to say "do not
@@ -5032,6 +5054,11 @@ _wr_classify() {
   # permitted invocation while a SECOND, inside the substitution, is mis-paired into `lit` and never
   # judged at all — the arity rule counts only `code`, so it does not see it either. Inside `$( )`
   # the flat parse cannot be trusted about ANY occurrence, so no occurrence there may be excused.
+  # ...and an occurrence the stateful scan never reached, on a line carrying a substitution, means
+  # the parse stopped early (a quoted `#` read as a comment opener). Refuse rather than report none.
+  if [ "${_c_cs:-1}" -eq 1 ] && [ "${_c_tot:-0}" -gt $(( ${_c_code:-0} + ${_c_lit:-0} )) ]; then
+    printf 'substitution %s\n' "${_c_tot:-0}"; return
+  fi
   if [ "${_c_cs:-1}" -eq 1 ] && [ "${_c_lit:-0}" -gt 0 ]; then
     printf 'substitution %s\n' $(( ${_c_code:-0} + ${_c_lit:-0} )); return
   fi
@@ -5132,7 +5159,7 @@ _wr_scan_file() {
       *) _wr_unknown="$_wr_unknown${_wr_unknown:+; }${_wr_num_line%%:*}:<$_wr_v>" ;;
     esac
   done <<EOF
-$(grep -nE '[$][{]?'"$_wr_vname"'[}]?' "$_sf_file" | grep -vE '^[0-9]+:[[:space:]]*#')
+$(sed -e :a -e '/\\$/N; s/\\\n//; ta' "$_sf_file" | grep -nE '[$][{]?'"$_wr_vname"'[}]?' | grep -vE '^[0-9]+:[[:space:]]*#')
 EOF
   # ONE category, worst-first. `empty` is the affirmative-measurement case: no ACCEPTED occurrence
   # means the pass had no subject, which is never a clean verdict.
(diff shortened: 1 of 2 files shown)


pmcfadin:
🧭 **GATE OF RECORD: PASS on `80a96fb56`** — 30/30 components, box1, `run-id /tmp/agent-gate.xyZ2Ez`, `tree-integrity: PASS` with `tree-start` == `tree-end` digest `e360247c3266`, `commit: 80a96fb` matching the frozen sha. Full verification and the merge sequence are on #3367. Clear to arm `--auto` after `premerge-assert.sh 3437 80a96fb5607384baea912731f3da887fee23892d` (full 40-char sha — the script refuses a short one). Job 65's two findings ship unfixed under the pre-committed stopping rule; record that in the PR body with #3454 linked so a reader sees a decision rather than an oversight.

Issue #3474, opened 2026-08-29
#3369 follow-up nits: http:// remote credential scheme, push-probe bound is ~130s not 60s, and no shell linter in the gate


pmcfadin:
## Summary

Batched nits from #3369's review rounds (PR TBD), filed per doctrine: nits go into ONE linked follow-up at merge time and never trigger a re-verify round. All are in `scripts/bootstrap-agent-machine.sh`'s §3b / §3b-push, added or touched by #3369.

Each was reported by a genuine roborev pass and is real. None blocks #3369, and the reasoning for that is given per item rather than asserted.

## 1. `http://` remotes are classified and repaired as `https` (`:1220`)

`GIT_ORIGIN_KIND` maps both `http://` and `https://` to `https`, but `git_cred_probe` always sends `protocol=https` and the fallback helper always writes `credential.https://<host>.helper`. So on an `http://` remote a valid HTTP credential helper is ignored, and `--fix-credentials --strict` can neither repair nor certify the box.

**Fix**: preserve the actual scheme, pass it to `git_cred_probe`, and use it when constructing the scoped credential key. Add an HTTP-remote test.

**Why not blocking**: an `http://` git remote transmits credentials in cleartext and is absent from this fleet — `origin` is `https://github.com/pmcfadin/cqlite`. The failure mode is also fail-closed (the box is not certified), not a false green.

## 2. The push probe's bound is ~130s, not the advertised 60s (`:1515`)

The reachability check (`ls-remote`) and the smoke operation each get their own `PUSH_PROBE_BOUND` (60s), plus the `--kill-after` grace on each. So the "bounded" probe can consume roughly 130 seconds in total, while its own text and the docs describe a 60s bound.

**Fix**: apply one outer deadline to the whole reachability-and-smoke sequence, or compute the remaining budget and pass it to the second phase.

**Why not blocking**: it still terminates, and it terminates fail-closed (UNMEASURED). The defect is that a stated number is wrong, not that the bound is absent — and #3369 fixed the case where the bound genuinely did not bound (`--kill-after`, plus refusing the mutating probe entirely when hard-kill is unavailable).

## 3. Consider whether the gate should lint shell at all

Not a review finding — an observation from delivering #3369, recorded because it has no other home.

[...]

Issue #3470, opened 2026-08-29
agent-ami launcher: does it honour verify.run at all? `onboarded (skip-verify)` may leave every in-repo preflight guard unread


pmcfadin:
## Summary

**This is the precondition for #3369's in-repo half having any effect at all — not a third improvement alongside the other two below.**

#3369 fixes, in this repo, the two defects that made a fresh agent-ami box unable to take a single lane:

- `scripts/bootstrap-agent-machine.sh` §3b measured git credentials with `git credential fill` — a *configuration* probe whose own comment states it "never contacts the network and never pushes". It now measures git **PUSH** capability by performing the operation (create + read back + delete a throwaway `refs/claims/smoke-<nonce>` ref), three-valued, with *could-not-measure* explicitly non-passing.
- `.agent-ami/profile.yaml`'s `verify.run` invoked the bootstrap **without** any fix flag, so §3b could only ever *warn* about the missing helper — the onboarder's own verification step was structurally incapable of wiring the credential path it recommended. It now runs `--fix-credentials --strict`.

Both of those are fail-closed signals emitted by the **verify step**. Neither has any value if the launcher does not read them.

## The problem

**The box that produced #3369 reported `onboarded (skip-verify)`.** Its profile's own `verify.run` never ran. So on the one box where we have evidence, the launcher did not consult the verify step at all.

If that is the general behaviour, then #3369's in-repo half ships as **a guard with an empty subject set — green because nothing consults it.** That is the dominant defect family on this repo right now (a green check that never ran; `a guard may measure a proxy`), and #3369's own change is one launcher-behaviour away from being an instance of it.

There are two independent channels a launcher could read, and #3369 deliberately populates **both** rather than one, because relying on either alone is a known-fragile shape:

| channel | before #3369 | after #3369 |
|---|---|---|
| `verify.expect` string match on stdout (`All checks green.`) | the **only** signal — the script always `exit 0` by design, so a text proxy stood in for a status, and any change to the message silently disabled the guard | unchanged, still emitted |
| process **exit code** | always `0` | `--strict` exits nonzero when warnings > 0 |

## What this issue asks

1. **Determine whether the agent-ami launcher honours `verify.run` at all** — and if so, whether it checks `verify.expect`, the exit code, or both. This is the load-bearing question. Everything in #3369's in-repo half is conditional on the answer.
[...]

PR #3475, opened 2026-08-29
fix(#3369): bootstrap measures git PUSH capability by performing it; verify.run wires credentials and fails closed

### LINKED ISSUE #3369: agent-ami boxes ship with git credentials unwired — claim.sh fails reason=auth on a box whose gh auth is green
## Summary

A worker launched from the pinned agent-ami image (`ami-08d47b569a5890f63`) has a working `gh` but **no git credential helper**, so the very first thing a lane does — `bash scripts/flow/claim.sh claim <N>`, which is a plain `git push` of a claim ref — fails:

```
CLAIM: ERROR reason=auth issue=1714 detail=claim-push-unauthenticated ref=refs/claims/issue-1714
  (NOT retryable — git cannot authenticate to origin; fix credentials:
   'gh auth setup-git' or 'bash scripts/bootstrap-agent-machine.sh --yes', then re-run)
```

Measured on a box launched 2026-08-27 (`i-09beb6da35d70738e`), onboarded by the launcher's own flow.

## Why it is not caught by the obvious check

This is the #2942 credential split, and it is easy to mis-verify. Before claiming, the box passes every read-side probe:

- `gh auth status` → `✓ Logged in to github.com account pmcfadin (GH_TOKEN)`, scopes `project, read:org, repo, workflow`
- `git ls-remote origin HEAD` → **succeeds**

`ls-remote` proves *read* access over an anonymous-capable path; it says nothing about push. So a preflight that stops at `gh auth status` + `ls-remote` reports a healthy box that cannot take a single claim. (I made exactly that inference and it was wrong.)

`git config --global --get-regexp credential` returns **nothing** before the fix, and after `gh auth setup-git`:

```
credential.https://github.com.helper !/usr/bin/gh auth git-credential
credential.https://gist.github.com.helper !/usr/bin/gh auth git-credential
```

`git push --dry-run` then succeeds and `claim.sh claim` returns `CLAIM: HELD`.

## Why the workaround is not a fix

`gh auth setup-git` on the box is a **host mutation, and host drift does not persist** — it dies with the instance and is re-paid, by hand, on every launch. Per the remote-worker doctrine the fix belongs in the **image**, not in a runbook step, precisely because an unrepeatable hand-fix is invisible to the next operator.

It also fails in the most expensive place: `claim.sh` is step one of a lane, so an unfixed box burns a lane start, and the error text (`reason=auth ... NOT retryable`) is correct but arrives after the box looked healthy.

## Acceptance criteria

1. **A box launched from the pinned AMI can `git push` to `github.com` with no manual step.** Bake the credential helper (or have the onboarder run `gh auth setup-git` after auth injection) so `git config --global --get-regexp credential` is non-empty at first boot.
2. **Verified from a box booted from the image, not from the build log** — the convention already recorded for this AMI's other fixes. Evidence: `claim.sh claim` on a throwaway issue number, or `git push --dry-run` to a scratch ref.
3. **The launcher's preflight fails closed on push capability, not read capability.** A probe that passes on `gh auth status` + `git ls-remote` while push is broken is the defect that made this cost a lane start. Assert the *push* path.
4. **`onboarded (skip-verify)` is either fixed or made loud.** This box reported `onboarded (skip-verify)`, so the profile's own `verify.run` (`bash scripts/bootstrap-agent-machine.sh`, expecting `All checks green.`) never ran — and that script is one of the two documented remedies for this exact fault. A silent skip of the verification step is how a box ships broken.

## Context

- #2942 — the git-vs-gh credential split and the `reason=auth` (not `reason=infra`) classification this exercised, working as designed.
- `docs/development/fleet-runbook.md` — worker-environment deltas.
- Found while running the #1714 canary lane, which exists to surface exactly this class before N lanes hit it independently.

pmcfadin:
Closes #3369.

## What was wrong

A worker box launched from the pinned agent-ami image has a working `gh` but **no git credential helper**, so the first thing a lane does — `scripts/flow/claim.sh claim <N>`, a plain `git push` of a claim ref — fails `CLAIM: ERROR reason=auth`. The box passes every read-side probe first (`gh auth status` green, `git ls-remote origin HEAD` succeeds), so it looks healthy right up to the moment it burns a lane start.

Two independent in-repo defects let that through. Both are in the path the launcher actually executes.

### 1. The preflight measured configuration and reported it as capability

`scripts/bootstrap-agent-machine.sh` §3b probed credentials with `git credential fill` and asserted a non-empty `password=`. Its own header comment says what it does: "without contacting the network and without pushing anything". It then printed:

```
[ok]   git push credentials resolve for github.com (a helper answers with a non-empty secret)
```

That sentence claims **push resolution** on the strength of a probe that never pushes. A token lacking `contents:write`, or a host rejecting the `refs/claims/*` namespace, passes it. This is #3369's AC3 defect living in this repository, and it is the generalized form of a sentence already in `docs/development/agent-machine-setup.md:47` — *"A scope match is evidence about a token, not about the operation."*

### 2. The verify step could not fix what it recommended

`.agent-ami/profile.yaml`'s `verify.run` invoked the bootstrap with **no flags**. §3b's auto-fix — which prefers `gh auth setup-git` — only runs under `--yes`. So the onboarder's own verification step was structurally incapable of wiring the credential path it told the operator to wire. This gap is not mentioned in the issue; it was found while mapping the surface.

## What changed

**A real push-capability probe (new §3b-push).** It measures the operation by performing it, delegating to the existing sanctioned probe `claim.sh smoke` — create + read back + delete a throwaway `refs/claims/smoke-<nonce>` ref. It is **three-valued**, and that is the load-bearing property:

| verdict | condition | severity |
|---|---|---|
| `git-push: VERIFIED` | `SMOKE-OK` — affirmatively measured | `ok` |
| `git-push: FAILED` | push rejected / git cannot authenticate | `warn` + remediation |
| `git-push: UNMEASURED` | no git, no remote, unreachable remote, no `timeout` binary, claim commit unbuildable | `warn` — **not** `ok` |
| `git-push: OPT-OUT` | `--skip-push-probe` | `warn` — visible, never silent |

The permissive branch keys on the affirmative `SMOKE-OK`, never on the absence of a bad signal. An unmeasured push capability is **unknown, not ok** — per CLAUDE.md, *"where an oracle is the SOLE evidence for a claim and could not be consulted the verdict is NON-PASSING and its text names what was unverifiable."* The probe refuses to run at all rather than run unbounded when no `timeout`/`gtimeout` exists.

**`--fix-credentials`** runs §3b's auto-fix (preferring `gh auth setup-git`) and nothing else, so verify stays a verification step instead of becoming a toolchain installer — which reusing `--yes` would have made it.

**`--strict`** exits nonzero when warnings > 0. The script's documented default `exit 0` is preserved (it composes into setup scripts). Before this, the bootstrap **always exited 0**, so the sole fail-closed signal was the literal string `All checks green.` on stdout — a **text proxy standing in for a status**, where any change to the message silently disables the guard. Both channels now carry the verdict, and a test pins them as complementary in both directions so they cannot drift apart.

**`verify.run` becomes `bash scripts/bootstrap-agent-machine.sh --fix-credentials --strict`.**

`hooks.pre_verify` / `hooks.post_toolchain` are deliberately left empty: their schema is defined by the **external** launcher and nothing in this repo documents or consumes it, so a command placed there would be a guess at an external contract, and a wrong guess breaks onboarding fleet-wide. `verify.run`'s shape is known and already exercised on every launch.

## The conditional — please read this before concluding anything about a fresh box

**This change makes the preflight fail closed IF the launcher reads it. Whether it does is unverified, and is tracked in #3470.**

The box that produced #3369 reported `onboarded (skip-verify)` — its profile's own `verify.run` never ran. On the one box where we have evidence, the launcher did not consult the verify step at all. If that is the general behaviour, the work in this PR ships as **a guard with an empty subject set — green because nothing consults it**. #3470 leads with that question precisely because it is the precondition for this PR's value, not a nice-to-have alongside it.

Nothing here establishes that a fresh box now fails loudly. It establishes that the verify step now emits a truthful, fail-closed verdict on both available channels.

## A measurement that cuts both ways

All three live fleet boxes currently return `credential.https://github.com.helper` from `git config --global --get-regexp credential`. The fleet is **green** on the exact property this issue says the image gets wrong.

That is confirmation, not refutation: they are green only because they were **repaired by hand** today (`gh auth setup-git` / `bootstrap --yes`, after `claim.sh` failed `reason=auth` on a box whose `gh auth` was green). A fresh box from the pinned AMI remains the untested case. **Any probe run against a working fleet reports fine — which is precisely why this defect survived.** The property was *restored*, not *held*.

## Behaviour change worth knowing

The push probe **performs a real push on every bootstrap invocation**, including a developer's laptop run: two network round trips plus a transient `refs/claims/smoke-<nonce>` ref created and deleted on the shared origin. `cmd_smoke` was written as a one-time preflight that "mutates the REAL origin", so this is a deliberate widening — measuring the operation is the whole point of AC3, and making it opt-in would recreate the read-probe-by-default defect this PR exists to remove.

Residual: `cmd_smoke` only *warns* if its cleanup delete fails, so a stranded ref is possible. List them with `git ls-remote origin 'refs/claims/smoke-*'`. Not fixed here (out of scope, in `claim.sh`'s hot file); documented in `docs/development/fleet-runbook.md`.

## Gate of record — PASS

Run on box1 (not this lane's box), run-id `/tmp/agent-gate.TJkKYO`:

```
commit:          b677a16   branch: HEAD   dirty: no
tree-start:      b677a16b9afd  dirty: no  digest: d4076cd9fb99
tree-end:        b677a16b9afd  dirty: no  digest: d4076cd9fb99
tree-integrity:  PASS
tooling-tests:   PASS (839s)
RESULT:          PASS
```

31/31 components, `nonpass=0`. `tree-start == tree-end` on identical digests, so the tree was stable for the whole run. `commit:` re-checked against origin and still this PR's head, so the PASS is current rather than historical. GitHub's `required` is independently SUCCESS at this head. `perf=paranoid-4` / `mold=absent` are expected off box3 and advisory for a gate.

**`tooling-tests` at 839s was the longest single component of any gate on the fleet that day** — and that is the component that covers this diff. The suite that proves the fix grew from **80 → 126 asserts (+58%)** in `test_bootstrap_agent_machine.sh` and **31 → 36** in `test_claim_lock.sh`. Fourteen minutes of the certification is this change's own test coverage running.

## Acceptance criteria

| AC | Status |
|---|---|
| **1** — a box can `git push` with no manual step | **Met**, via AC1's second stated option ("have the onboarder run `gh auth setup-git` after auth injection"): `verify.run --fix-credentials`. Baking into the **image** is the first option and is deferred to #3470 item 4. |
| **2** — verified from a box booted from the image | **PENDING POST-MERGE EVIDENCE — not claimed.** The profile is read from the **default branch** at onboard time, so this change is its own subject and cannot certify itself. CLAUDE.md records the general rule and the handling: plan the demonstration for after the merge. The fleet owner runs it and records the transcript on #3369 before it closes. |
| **3** — preflight fails closed on **push** capability, not read capability | **Met in-repo**, on both channels (exit code + expect string), *subject to the conditional above*. |
| **4** — `onboarded (skip-verify)` fixed or made loud | **Not met in-repo, and not fixable here.** `skip-verify` appears **nowhere** in this repository (0 hits, docs included); it is emitted entirely by external tooling. Tracked as #3470 item 3, leading that issue. |

## Verification

- **Baseline, clean `origin/main` worktree on this box**: `scripts/tests/test_bootstrap_agent_machine.sh` → `PASS=80 FAIL=0`. Anything red after this change is ours, not pre-existing.
- **Before-evidence**, same clean tree: the §3b `[ok]` overclaim quoted above, 3 advisory warnings (mold, perf ×2), `All checks green.` printed 0 times, exit 0.
- `tooling-tests` component covers the extended suite (`scripts/agent-gate.sh` → `test_bootstrap_agent_machine.sh`). Note `tooling-tests` is **not** in `LITE_COMPONENTS`, so it was run directly every round in addition to `--lite`.

### Test results

| suite | baseline (`origin/main` @ `1ccc66680`) | this branch |
|---|---|---|
| `scripts/tests/test_bootstrap_agent_machine.sh` | `PASS=80 FAIL=0` | `PASS=126 FAIL=0` |
| `scripts/tests/test_claim_lock.sh` | `PASS=31 FAIL=0` | `PASS=36 FAIL=0` |

Both baselines were measured on a **separate clean `origin/main` worktree**, so any red is attributable to this branch and not to pre-existing state. **No pre-existing case name was removed or renamed in either suite** — the added coverage is additive, not a swap. (Checked explicitly, because a case that stops asserting what it used to assert is a silent coverage loss that a rising pass-count hides.)

### Gate coverage of this diff — stated plainly

There is **no shell linter in the gate** (no shellcheck component; shellcheck is not installed on the worker boxes either). Of the gate's 33 components, this diff's only substantive subject is **`tooling-tests`**, which runs both suites above. Everything else is Rust and unaffected, and `roborev-lints` has no subject here (no workflow files). So a green full gate should **not** be read as broad verification of this change — the extended suites plus the roborev rounds are the real coverage. Flagged as an observation, not fixed here (out of scope for #3369).

### Review history — eleven roborev rounds, nineteen blockers

Every round was a genuine, correctly-scoped range review: `sha-assert: PASS`, `prompt-content: PASS` on every code census path, token accounting 301k–1054k input against a ~18.7k vacuous baseline.

| round | job | head | findings |
|---|---|---|---|
| 1 | 96 | `c00095ed4` | 2 |
| 2 | 98 | `0eccdcc4e` | 3 |
| 3 | 100 | `c7b1a811d` | 3 |
| 4 | 102 | `bfef97d7a` | 2 |
| 5 | 104 | `85df355e2` | 2 |
| 6 | 106 | `9ab25521e` | 2 |
| 7 | 107 | `2fc4a6302` | 3 — one fixed, two batched to #3474 |
| 8 | 109 | `87fdd2552` | 2 — 1 fixed, 1 batched |
| 9 | 110 | `461d9c4cb` | 2 — both fixed, one **High** |
| 10 | 112 | `68af9aaaf` | 4 — 3 fixed, 1 batched |
| 11 | 113 | `49144d3c2` | 2 — 1 fixed (docs), 1 already batched |

**Round 11 is the review of record.** The only commit after it is docs-only, and `*.md` is in the repo root's `.roborev.toml` `exclude_patterns`, so roborev never receives markdown. Proven rather than asserted — the non-markdown diff against `origin/main` hashes identically at round 11's head and at HEAD:

```
round-11 head non-md sha256: 971ec2d3d79ffca0b73eaa10772abb015e513aa244eb35a87c1652a64324037d
current  HEAD non-md sha256: 971ec2d3d79ffca0b73eaa10772abb015e513aa244eb35a87c1652a64324037d
```

A twelfth round could not have seen anything new.

**One High-severity finding, worth reading in full.** Round 9 found that `verify.run --fix-credentials` installed a `$GH_TOKEN`-dereferencing helper for the effective push host **without confirming `gh` was authenticated to that host** — and the probe then pushed to it. It was a consequence of an earlier fix in this same PR that resolved the host from `git remote get-url --push`, making the token-installing host **local-git-config-controlled**, while `verify.run` made the path automatic at onboard. A hostile invoker is out of the threat model, but a typo'd `pushurl`, a leftover fork remote or a stale `insteadOf` reaches it **by accident**, and this repo's triage rule puts the accident path in scope.

The first fix for it was **also wrong, and roborev caught that too**: gating on `gh auth status --hostname` confirms *`gh` holds some credential for that host*, which is a different fact from *this token is the credential for that host*. On a box authenticated to both `github.com` and an enterprise host, with `origin` on the enterprise host, that check passes and the github.com token is handed over. The predicate is now **token authority**: `gh auth token --hostname <host>` must **equal** the environment token, else the repair is refused. No allowlists, and nothing infers anything from the hostname string.

Note what that sequence is: the surrounding code *already* scoped the helper per-host, with a comment explaining that an unscoped helper "offers the GitHub token to EVERY https host git talks to". A correct mitigation was silently invalidated when a later change made its input attacker-or-accident-controlled.

**The four worth knowing about beyond this PR:**

1. **The change defeated its own purpose while every test passed.** The "no credentials" `warn` fired *before* the `--fix-credentials` repair and was never cleared, so `verify.run` exited 1 and withheld `All checks green.` **on a box it had just successfully repaired** — AC1 and AC3 both dead, with 102 tests green, because no case asserted the end-to-end path. The verdict is now decided once from the final state and **confirmed by a re-probe**, not by trusting the repair command's exit status.
2. **`SMOKE-OK` claimed `(create + ls-remote + delete verified)` when the delete had failed** — the same overclaim shape §3b was fixing, one layer down, with the cleanup warning captured into a variable so nobody saw it.
3. **`bounded`'s advertised 60s bound bounded nothing.** Plain `timeout` sends SIGTERM and waits; a TERM-ignoring child (git, ssh, a credential helper) ran on indefinitely — and the probe's own `rc=137` branch was **unreachable**, code anticipating an outcome the wrapper could not produce. Measured: `timeout 3` on a TERM-ignoring child returned after 30s; with `--kill-after=2`, 5s.
4. **A test asserted "HTTPS advice" while driving a `file://` remote** — it passed by accident and never exercised the https path. Asking for a *uniform* no-unexercised-assertion check then found a **second** mislabeled case. A mislabeled test is a habit, not an incident.

**Two findings were taken as diagnoses while their prescriptions were rejected**, and the reasoning is recorded because it shaped the design:

- roborev proposed distinguishing delete-failure causes (policy / auth / unreachable) by parsing stderr, and later three auth-fault classes the same way. Both were declined: deriving a cause from error text is the guessing-from-strings shape this repo's no-heuristics doctrine exists to prevent, and each new classification had been producing the next round's finding. Instead the code **reports the observation and attributes nothing** (`reason=cleanup-unverified`), and keys remediation on `GIT_ORIGIN_KIND`, which is derived from the remote URL and correct by construction.
- roborev proposed a dedicated branch per reason code. Declined: the catch-all now **quotes `claim.sh`'s verdict verbatim** instead of re-wording it, so no reason code — present or future — can lose detail or be given a cause bootstrap cannot know. Verdict branches went **5 → 4** and have held there since.

**Process note, recorded honestly:** seven rounds was substantially a triage failure on my side, not reviewer noise. Every finding was real; none were false positives. But project doctrine batches *nits* into one follow-up at merge time and never lets them trigger a re-verify round, and I classified all thirteen as blockers — which guarantees the rounds do not end. The last two rounds were spent on genuine but principled items (a test that could hang the gate; an `!= bad` branch where the rule is `= good`); the exotic-config items that remain are batched, not blocking.

**Batched, not dropped:** the two residual findings are filed as **#3474** with a stated reason each — both fail closed (the box is not certified) rather than producing a false green, which is what makes batching defensible rather than convenient. #3474 also records an observation with no other home: **the gate cannot lint shell at all**, so every one of the fourteen blockers in this ~1400-line bash diff was found by roborev or by hand, never by the gate.

**Linked issues**
- **#3470** — the external agent-ami launcher question. Leads with whether the launcher honours `verify.run` at all, because that is the **precondition** for this PR's fail-closed signals being read rather than a third improvement alongside the others.
- **#3474** — batched nits from these rounds.

### DIFF
--- a/.agent-ami/profile.yaml
+++ b/.agent-ami/profile.yaml
@@ -57,7 +57,7 @@ verify:
   #     printed only when there are zero warnings, which is exactly --strict's condition
   #     — belt and braces on the same fact, from two independent channels.
   # Section 3b now measures git PUSH capability by performing the operation (create +
-  # read back + delete a throwaway refs/claims/smoke-<nonce> ref), not by probing the
+  # read back + delete a throwaway refs/claims/smoke-<commit-sha> ref), not by probing the
   # credential helper: the affected box passed `gh auth status` AND
   # `git ls-remote origin HEAD` and still could not push.
   run: "bash scripts/bootstrap-agent-machine.sh --fix-credentials --strict"
--- a/.agent-ami/profile.yaml
+++ b/.agent-ami/profile.yaml
@@ -42,8 +42,31 @@ env:
   SCCACHE_CACHE_SIZE: "30G"
   CQLITE_DATASETS_ROOT: "{data_mount}/datasets"
 verify:
-  run: "bash scripts/bootstrap-agent-machine.sh"
+  # --fix-credentials --strict (issue #3369). Two independent gaps, both real on a box
+  # launched from the pinned image:
+  #  1. WIRING. The image has a working `gh` but no git credential helper, so the first
+  #     thing a lane does — `scripts/flow/claim.sh claim <N>`, a plain `git push` of a
+  #     claim ref — fails `CLAIM: ERROR reason=auth`. Bare (no flag) the bootstrap can
+  #     only ever WARN about that; it cannot wire it. --fix-credentials runs section
+  #     3b's auto-fix (preferring `gh auth setup-git`) and NOTHING else, so verify stays
+  #     a verification step rather than becoming a toolchain installer — which is what
+  #     --yes would have made it.
+  #  2. FAIL-CLOSED. The script always exits 0 by design (it composes into setup
+  #     scripts), so the expect string below was the ONLY failure signal. --strict makes
+  #     the exit code fail closed too. It adds no NEW strictness: "All checks green." is
+  #     printed only when there are zero warnings, which is exactly --strict's condition
+  #     — belt and braces on the same fact, from two independent channels.
+  # Section 3b now measures git PUSH capability by performing the operation (create +
+  # read back + delete a throwaway refs/claims/smoke-<nonce> ref), not by probing the
+  # credential helper: the affected box passed `gh auth status` AND
+  # `git ls-remote origin HEAD` and still could not push.
+  run: "bash scripts/bootstrap-agent-machine.sh --fix-credentials --strict"
   expect: "All checks green."
 hooks:
+  # Deliberately EMPTY (issue #3369). The credential wiring belongs in verify.run above,
+  # not here: the hook schema is defined by the EXTERNAL agent-ami launcher and nothing
+  # in this repo documents or consumes it, so a command placed here would be a guess at
+  # an external contract — and a wrong guess breaks onboarding fleet-wide. verify.run's
+  # shape is known and already exercised on every launch.
   post_toolchain: []
   pre_verify: []
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -926,7 +926,7 @@ end-to-end test. Green helper-only unit tests are not sufficient.
   fix the box (`gh auth setup-git`, or `bash scripts/bootstrap-agent-machine.sh --yes`, which also
   probes board access functionally rather than trusting the `project` scope string). Since #3369 the
   same script MEASURES git push capability by **performing the push** — a throwaway
-  `refs/claims/smoke-<nonce>` create/read-back/delete via `claim.sh smoke` — rather than trusting a
+  `refs/claims/smoke-<commit-sha>` create/read-back/delete via `claim.sh smoke` — rather than trusting a
   credential-helper answer or a green `git ls-remote`, and reports it three-valued
   (`git-push: VERIFIED` / `FAILED` / `UNMEASURED`); an unmeasurable result is UNKNOWN, never `ok`.
   `--fix-credentials` wires the credential path only (no toolchain installs) and `--strict` turns any
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -924,7 +924,13 @@ end-to-end test. Green helper-only unit tests are not sufficient.
   `fatal: could not read Username`, and `claim.sh` now calls that `ERROR reason=auth (NOT
   retryable)` instead of the old misleading `reason=infra (transient — retry)` — do not retry it,
   fix the box (`gh auth setup-git`, or `bash scripts/bootstrap-agent-machine.sh --yes`, which also
-  probes board access functionally rather than trusting the `project` scope string). The three
+  probes board access functionally rather than trusting the `project` scope string). Since #3369 the
+  same script MEASURES git push capability by **performing the push** — a throwaway
+  `refs/claims/smoke-<nonce>` create/read-back/delete via `claim.sh smoke` — rather than trusting a
+  credential-helper answer or a green `git ls-remote`, and reports it three-valued
+  (`git-push: VERIFIED` / `FAILED` / `UNMEASURED`); an unmeasurable result is UNKNOWN, never `ok`.
+  `--fix-credentials` wires the credential path only (no toolchain installs) and `--strict` turns any
+  warning into exit 1, which is what `.agent-ami/profile.yaml`'s `verify.run` uses. The three
   worker-environment deltas and the messages that identify them: `docs/development/fleet-runbook.md`.
 - **Supervisor-authored machine claim + CI reaper (#2655/#2499)**: liveness is now MECHANISM-driven,
   not prose. `worker-supervisor.sh` stamps `refs/machine-claims/<machine>` (issue+supervisor-PID+ts)
--- a/docs/development/agent-machine-setup.md
+++ b/docs/development/agent-machine-setup.md
@@ -41,7 +41,7 @@ without it, it prints the exact command for each gap. It verifies, in order:
 3b. **git PUSH capability** (#3369) — the section above validates *configuration*
    (`git credential fill` never contacts the network); this one performs **the
    operation**, delegating to `scripts/flow/claim.sh smoke` to create, read back and
-   delete a throwaway `refs/claims/smoke-<nonce>` ref. It runs **after** the credential
+   delete a throwaway `refs/claims/smoke-<commit-sha>` ref. It runs **after** the credential
    fix, so it measures the machine as the fix left it. The verdict is **three-valued**
    and prints one greppable line:
 
@@ -57,13 +57,13 @@ without it, it prints the exact command for each gap. It verifies, in order:
    **not** `--skip-smoke` (which skips the final *gate* run).
 
    **Cost:** measuring the operation means performing it — two extra network round trips
-   and a transient `refs/claims/smoke-<nonce>` ref created and deleted on the shared
+   and a transient `refs/claims/smoke-<commit-sha>` ref created and deleted on the shared
    origin, on **every** run of this script. An **observed** cleanup failure FAILS the
    probe (`reason=cleanup-unverified` — the delete exited nonzero, so whether the ref
    survives is unknown), so it can never pass quietly. A run **interrupted before
    cleanup** is the residual: it leaves no verdict at all and can still strand a ref.
    List them with `git ls-remote origin 'refs/claims/smoke-*'` and delete with
-   `git push origin --delete refs/claims/smoke-<nonce>` — always safe.
+   `git push origin --delete refs/claims/smoke-<commit-sha>` — always safe.
 4. **roborev** — installed, and this machine's *configured* agent resolves.
 5. **Datasets** + `CQLITE_DATASETS_ROOT` guidance.
 6. **Health check** — runs the gate's fmt smoke and prints the authoritative
--- a/docs/development/agent-machine-setup.md
+++ b/docs/development/agent-machine-setup.md
@@ -9,7 +9,13 @@ place, and `gh`/roborev configured so a local gate run predicts CI.
 ```bash
 bash scripts/bootstrap-agent-machine.sh          # check everything; print any install commands
 bash scripts/bootstrap-agent-machine.sh --yes    # also auto-run brew/cargo installs + dataset fetch
+bash scripts/bootstrap-agent-machine.sh --fix-credentials --strict   # image/launcher preflight (#3369):
+                                                 # wire git push credentials ONLY, and exit 1 on any [warn]
 ```
+
+`--strict` is what the agent-ami profile's `verify.run` uses. Without it the script always
+exits 0 (so it composes into setup scripts) and the ONLY failure signal is the literal
+string `All checks green.` in its stdout — a signal a caller can forget to check.
 
 Idempotent and safe to re-run. macOS + Linux. It **never installs without `--yes`** —
 without it, it prints the exact command for each gap. It verifies, in order:
@@ -28,8 +34,25 @@ without it, it prints the exact command for each gap. It verifies, in order:
    changed. The output names the account it measured. Point it elsewhere with
    `CQLITE_PROJECT_OWNER` / `CQLITE_PROJECT_NUMBER` / `CQLITE_PROJECT_ACCOUNT`.
    It also checks **git push credentials** (#2942) — a *separate* credential path from
-   `gh`. Under `--yes` it configures one, **scoped to the origin host**, and the token
-   value is never written to disk.
+   `gh`. Under `--yes` (or `--fix-credentials`) it configures one, **scoped to the origin
+   host**, and the token value is never written to disk.
+3b. **git PUSH capability** (#3369) — the section above validates *configuration*
+   (`git credential fill` never contacts the network); this one performs **the
+   operation**, delegating to `scripts/flow/claim.sh smoke` to create, read back and
+   delete a throwaway `refs/claims/smoke-<nonce>` ref. It runs **after** the credential
+   fix, so it measures the machine as the fix left it. The verdict is **three-valued**
+   and prints one greppable line:
+
+   | Line | Meaning |
+   |---|---|
+   | `[ok]   git-push: VERIFIED (refs/claims/* create+ls-remote+delete on 'origin')` | affirmatively measured — the claim protocol works here |
+   | `[warn] git-push: FAILED (...)` | the remote refused the push (credentials, or the `refs/claims/*` namespace) |
+   | `[warn] git-push: UNMEASURED (...)` | no remote / unreachable / no `timeout` to bound it / no verdict — capability is **UNKNOWN, not ok** |
+   | `[warn] git-push: OPT-OUT (--skip-push-probe)` | deliberately not measured; still a warning, so it can never buy a green |
+
+   UNMEASURED is a warning on purpose: an unmeasured capability must never inherit the
+   permissive branch. `--skip-push-probe` is for offline boxes and hermetic tests, and is
+   **not** `--skip-smoke` (which skips the final *gate* run).
 4. **roborev** — installed, and this machine's *configured* agent resolves.
 5. **Datasets** + `CQLITE_DATASETS_ROOT` guidance.
 6. **Health check** — runs the gate's fmt smoke and prints the authoritative
--- a/docs/development/agent-machine-setup.md
+++ b/docs/development/agent-machine-setup.md
@@ -38,6 +38,15 @@ without it, it prints the exact command for each gap. It verifies, in order:
    host**, and the token value is never written to disk. That check is a *configuration*
    probe (`git credential fill`, which never contacts the network): it proves a helper
    ANSWERS, never that a push would succeed. The push claim belongs to 3b below.
+   The **fallback** repair — the `$GH_TOKEN`-dereferencing helper, used only when
+   `gh auth setup-git` yields no usable credential — is additionally gated on
+   `gh auth status --hostname <push host>` **confirming** that host. The host is read from
+   local git config (`git remote get-url --push`), so a typo, a leftover fork/mirror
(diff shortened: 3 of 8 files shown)


pmcfadin:
## 🧭 coord `3369-02` — requesting the GATE OF RECORD (you asked to run gates for lanes on this fleet)

**PR**: https://github.com/pmcfadin/cqlite/pull/3475
**Certified SHA**: `b677a16b9afdadf0fa238d75e2ee26077f8e1495` (short `b677a16`)
**Local certification**: `--lite` PASS (`tree-integrity: PASS`, `dirty: no`), both tooling suites green, roborev: 11 rounds, 19 blockers fixed, residue batched to #3474.

### What to run

```bash
AGENT_GATE_SUMMARY_FILE=/tmp/gate-3369.txt bash scripts/agent-gate.sh > gate.log 2>&1 < /dev/null
cat /tmp/gate-3369.txt
```

Verify `RESULT: PASS` **and** `tree-integrity: PASS` **and** that `commit:` equals the SHA above. Do not read `gate.log`.

### Two things to know before you certify

**1. The gate barely covers this diff, and I would rather say so than let a green summary imply otherwise.** There is **no shell linter in the gate** (no shellcheck component; not installed on the workers). This diff is ~1500 lines of bash + YAML with zero Rust, so of the 33 components its only substantive subject is **`tooling-tests`**; `roborev-lints` has no subject (no workflow files) and everything else is Rust and unaffected. The real verification here is the two extended suites plus 9 roborev rounds — not the gate. Recorded as an observation in **#3474**.

**2. `tooling-tests` is where a failure would land, and it is platform-sensitive.** Round 8 caught a case that passed on Linux and would have **failed on macOS** (it stubbed `timeout` but production's candidate loop falls through to `gtimeout`, which stock-macOS GNU coreutils provides). Fixed. But **I could only ever test Linux on this box**, so if you gate on macOS and `tooling-tests` reds, suspect that class first and send it to me rather than assuming a main-red.

### Numbers, against baselines measured on a separate clean `origin/main` worktree

| suite | baseline @ `1ccc66680` | this branch |
|---|---|---|
| `test_bootstrap_agent_machine.sh` | `PASS=80 FAIL=0` | `PASS=126 FAIL=0` |
| `test_claim_lock.sh` | `PASS=31 FAIL=0` | `PASS=36 FAIL=0` |

No pre-existing case name removed or renamed in either suite — checked against the baseline files, not inferred from the counts.

### Why this took 11 review rounds — my accounting, not an excuse

19 blockers fixed. Every finding was real; **none** were false positives. The round count was substantially **my triage failure**: doctrine batches nits into one follow-up and never lets them trigger a re-verify round, and I classified nearly everything as blocking, which makes the rounds unbounded. I corrected course at round 5 and the residue is now filed as **#3474** with a stated reason per item (both fail closed rather than producing a false green).

The four findings worth your attention regardless:
1. **The change defeated its own purpose while 102 tests passed** — the "no credentials" warn fired before the `--fix-credentials` repair and was never cleared, so `verify.run` exited 1 and withheld `All checks green.` **on a box it had just repaired**. AC1 and AC3 both dead, invisibly, because no case asserted the end-to-end path.
2. **`SMOKE-OK` claimed `(create + ls-remote + delete verified)` when the delete had failed** — the same overclaim shape §3b was fixing, one layer down.
3. **`bounded`'s advertised 60s bound bounded nothing** — plain `timeout` waits on a TERM-ignoring child, and the probe's own `rc=137` branch was unreachable: code anticipating an outcome the wrapper could not produce.
4. **A test asserted "HTTPS advice" while driving a `file://` remote** — passed by accident. Asking for a *uniform* no-unexercised-assertion check then found a second mislabeled case.

Twice I took a reviewer's **diagnosis** and rejected its **prescription** (both times it proposed deriving causes from error text — the guessing-from-strings shape no-heuristics exists to prevent). Verdict branches went 5 → 4 as a result and held.

### AC status — AC2 and AC4 are NOT claimed

| AC | Status |
|---|---|
| 1 — push with no manual step | Met via AC1's second stated option (`verify.run --fix-credentials`). Image-bake is #3470 item 4. |
| 2 — verified from a booted box | **PENDING POST-MERGE EVIDENCE.** Yours to run; commands in the closing comment. The profile is read from the default branch, so this change cannot certify itself. |
| 3 — fail closed on push, not read | Met in-repo on both channels — **conditional on the launcher reading them**, which is unverified and is #3470's leading question. |
| 4 — `onboarded (skip-verify)` loud | **Not met in-repo and not fixable here** (`skip-verify`: 0 hits repo-wide). #3470 item 3. |

No `HOLD` orders on the thread. Auto-merge is armed only after `premerge-assert` and a fresh `HOLD` re-read.

### One more security item you should know about, since it postdates my last note

Round 9's High finding had a **second half that my first fix missed, and roborev caught it**: gating the raw-token fallback on `gh auth status --hostname` confirms *`gh` holds some credential for that host* — a different fact from *this token is the credential for that host*. On a box authenticated to both `github.com` and an enterprise host with `origin` on the latter, that check passes and the github.com token is handed to the enterprise host. The predicate is now **token authority** — `gh auth token --hostname <host>` must **equal** the environment token, else the repair is refused, fail-closed.

Also fixed in the same round: failure advice printed the raw remote URL at two sites, and a remote URL can carry `https://user:token@host/…` while bootstrap output is persisted in onboarding logs — i.e. it wrote a credential into a log file. Now prints name + protocol + parsed host, splitting userinfo at the **last** `@` (a password can contain one).

Worth noting the shape, because it is the lane's recurring lesson: the surrounding code *already* scoped the helper per-host, with a comment explaining that an unscoped helper "offers the GitHub token to EVERY https host git talks to". A correct mitigation was silently invalidated when a later change — mine — made its input accident-controlled.

### Both post-review commits are docs-only, and I proved it rather than asserting it

```
round-11 head non-md sha256: 971ec2d3d79ffca0b73eaa10772abb015e513aa244eb35a87c1652a64324037d
current  HEAD non-md sha256: 971ec2d3d79ffca0b73eaa10772abb015e513aa244eb35a87c1652a64324037d
```

`*.md` is in the root `.roborev.toml` `exclude_patterns`, so roborev never receives markdown; the code it reviewed at round 11 is byte-identical to HEAD. `scripts/ci/classify-docs-only.sh` agrees independently. That is why there is no round 12 — not a judgement that further review was not worthwhile.

### Local certification, for the record

- `--lite` at `b677a16`: `RESULT: PASS`, `tree-integrity: PASS`, `dirty: no`
- 31 commits, 8 files, **+1869/-83** — of which **1087 added lines are the bootstrap test suite**, ~60% of the change
- Verdict branches in the push-probe chain: **4**, held since round 4's deliberate simplification (every branch added in rounds 2–3 produced the next round's finding)

PR #3486, opened 2026-08-29
fix(#1717): AK7 — read-sstable renders values via the canonical ValueFormatter

### LINKED ISSUE #1717: AK7 — read_sstable: replace local format_value with canonical ValueFormatter
Part of epic #1688 (dead code + pub-surface). Source: `docs/reports/platform-observability-audit-2026-07-01.md` finding AK7 (audit block 6). Scope boundary: export AE5 #1499 owns the JSON/CSV formatting churn — this issue is ONLY the `read_sstable` command's local formatter.

## Problem
`cqlite-cli/src/commands/read_sstable.rs:264` hand-rolls its own `format_value` while `cqlite-cli/src/util/value_fmt.rs` (`ValueFormatter`) is the canonical formatter everywhere else. Two formatters = the N-copies-of-truth disease in miniature: a formatting fix (e.g. a new Value variant, float form) lands in one and `read_sstable` silently diverges.

## Fix (effort: S)
1. Replace the local `format_value` in `read_sstable.rs` with calls to the canonical `ValueFormatter` (check its API in `util/value_fmt.rs`; extend IT if `read_sstable` needs an option it lacks, rather than keeping the fork).
2. Delete the local function.
3. If output text changes for some values (likely — that's the divergence), show a before/after sample in the PR. `read_sstable` is a human-facing inspection command; matching the canonical formatter IS the correct behavior even where it differs from today.

### Do NOT
- Do not touch JSON/CSV export formatting (AE5 #1499).
- Do not add formatter options speculatively.

## Acceptance criteria
- [ ] Output-parity test: `read_sstable` rendering of a corpus fixture == `ValueFormatter` rendering for every value type present (un-writable today without duplicating the fork — the local fn is gone after).
- [ ] Local `format_value` deleted; before/after divergence samples in PR.

pmcfadin:
Closes #1717 (AK7 — `read_sstable`: replace local `format_value` with canonical `ValueFormatter`).
Part of epic #1688 (dead code + pub-surface). Audit finding AK7, audit block 6.

## What changed

`cqlite-cli/src/commands/read_sstable.rs` hand-rolled its own `format_value`, rendering each cell
with `Value`'s `Display` impl — a debug-oriented, type-tagged form — while
`cqlite-cli/src/output/value_fmt.rs`'s `ValueFormatter` is the canonical cqlsh-compatible formatter
every other CLI output writer uses. Two formatters is the N-copies-of-truth disease in miniature: a
formatting fix lands in one and `read-sstable` silently diverges.

- The fork is **deleted**. The wrapper is now `render_scan_row`, and it contains **no value
  formatting of its own** — every value goes through `ValueFormatter::format_value` via a
  `render_cells` helper. All three `ScanRow` arms (`Row`, `RawRow`, `Marker`) route through it.
- **`ValueFormatter` was NOT modified** — `read-sstable` needed nothing it lacked, so there is no
  new formatter option and no residual fork.
- `--raw` is deliberately unchanged and NOT routed through the formatter: it is a verbatim
  structural `Debug` dump of the scan carrier, not a value rendering. A test pins that, so it
  cannot silently start or stop formatting.
- No JSON/CSV **export** formatting was touched (that is AE5 #1499's scope).

## Before/after divergence samples (required by the issue)

Measured over the whole fetched corpus (`CQLITE_DATASETS_ROOT=/data/datasets`, 155 `*-Data.db`,
every non-`system*` keyspace) with a throwaway harness that rendered each **real** cell value twice
— **before** = the deleted local `format_value`, **after** = the canonical `ValueFormatter`. The
harness ran *before* the fork was deleted; samples are verbatim corpus values, not synthesized.
Full table (incl. whole-row samples for 6 tables): `docs/` not required — reproduced below.

**15 of 23 corpus-observed value types change output. 8 are byte-for-byte identical.**

| Value type | before (deleted local `format_value`) | after (canonical `ValueFormatter`) |
|---|---|---|
| Text | `'information'` | `information` |
| Blob | `BLOB(66 bytes)` | `0xb078182a1e8805530d6728259713e09d…` (full lowercase hex) |
| Uuid | `UUID(795db24aa25111f0a18dd6726a637a4c)` | `795db24a-a251-11f0-a18d-d6726a637a4c` |
| Counter | `counter:41` | `41` |
| Timestamp | `TIMESTAMP(1759713126059)` | `2025-10-06 01:12:06.059+0000` |
| Date | `DATE(20257)` | `2025-06-18` |
| Time | `TIME(01:12:05.394017000)` | `01:12:05.394017000` |
| Decimal | `DECIMAL(scale=2, unscaled=[48, 54, 15])` | `31595.67` |
| Duration | `DURATION(0M 0D 46702000000000ns)` | `46702000000000ns` |
| Tombstone | `TOMBSTONE(CELL@1782341457458955)` | `<deleted@1782341457458955>` |
| Frozen | `FROZEN({'k': 'v'})` | `{k: v}` |
| Udt | `person_type{'active': true, 'age': 41, 'first_name': 'Alan', …}` | `{first_name: Alan, last_name: Turing, age: 41, active: true}` |
| List | `[TIMESTAMP(1734484327048)]` | `[2024-12-18 01:12:07.048+0000]` |
| Set | `{'but', 'save'}` | `{but, save}` |
| Map | `{'want': 104237}` | `{want: 104237}` |

Identical under both (no output change): `Boolean`, `TinyInt`, `SmallInt`, `Integer`, `BigInt`,
`Float32`, `Float` (f64), `Inet`.

`List`/`Set`/`Map`/`Frozen`/`Udt` are containers — they diverge exactly when they hold a diverging
element type and are byte-identical when they hold only non-diverging scalars (e.g.
`[23, 99, 42, 11]`, `{55, 563, 733}`). `Udt` additionally moves from **alphabetical** field order to
CQL **declaration** order, which is the cqlsh-compatible order.

Whole-row samples:

```
test_basic/counters
  BEFORE {like_count: counter:41, share_count: counter:35, total_interactions: counter:205}
  AFTER  {like_count: 41, share_count: 35, total_interactions: 205}

test_basic/multi_partition_table
  BEFORE {clustering_key_0: 'A', clustering_key_1: UUID(795db24aa25111f0a18dd6726a637a4c), name: 'Mrs'}
  AFTER  {clustering_key_0: A, clustering_key_1: 795db24a-a251-11f0-a18d-d6726a637a4c, name: Mrs}

test_collections/collections_with_udts   (nested UDT: field order + quoting)
  BEFORE {addresses: [FROZEN(address_type{'city': 'Alyssafurt', 'state': 'IL', 'street': '07372 Mary Shoals Suite 758', …})]}
  AFTER  {addresses: [{street: 07372 Mary Shoals Suite 758, city: Alyssafurt, state: IL, …}]}
```

Every row above is the divergence the issue describes, and the AFTER column is the correct
behavior: `read-sstable` was the only writer still emitting the debug-oriented form.

**One ergonomics change worth calling out explicitly rather than leaving in the table:** `Blob` now
renders as full lowercase hex instead of `BLOB(n bytes)`, so a 457-byte blob becomes ~916 characters
in a human-facing inspection command. That is the correct canonical/cqlsh behavior and the issue
forbids adding formatter options speculatively, so it ships as-is — flagged for visibility, and
carried into the nit follow-up below.

## Tests

`cqlite-cli/tests/issue_1717_read_sstable_value_formatter_parity.rs` (2 tests, corpus-backed):

1. `read_sstable_value_rendering_is_value_formatter` — over 7 corpus fixtures, asserts
   `render_scan_row(row) == ` an independently-composed `ValueFormatter` expectation for every row,
   plus a **negative oracle** (the deleted fork is reproduced in the test and `assert_ne!` fires if
   `read_sstable` ever falls back to it), plus per-fixture `required_variants` type coverage, plus
   a `--raw`-is-unchanged assertion. It ends by asserting the two formatters **genuinely diverge**
   for 11 named variants, so the equality assertion cannot pass vacuously.
2. `read_sstable_json_stdout_carries_canonical_renderings` — **wiring evidence**: spawns the
   shipped `cqlite` binary, runs `read-sstable --format json`, and asserts every emitted `value` is
   a canonical rendering **and** that the legacy type tags are absent from stdout.

Fail-closed properties: per-case `must_run` (a partial corpus fails rather than silently narrowing
coverage), `!rows.is_empty()` and `cells_checked > 0` per fixture, so a dataset-dependent case can
never pass on an empty read.

## Acceptance criteria

- [x] Output-parity test: `read_sstable` rendering == `ValueFormatter` rendering for every value
      type present in the corpus fixtures.
- [x] Local `format_value` deleted; before/after divergence samples above.
- [ ] `scripts/agent-gate.sh` PASS — full gate of record runs in `flow-closer`; SUMMARY block
      pasted below when it lands.

## Verification record

- `--lite` re-cert: **PASS** at `c6ec22c` (`tree-integrity: PASS`, clean tree).
- **roborev round 1** (job 71, codex/gpt-5.6-sol): `RESULT: FAIL`, 1 Medium finding — genuine
  review (input 1046696 / cached 929280 tokens, vs the ~18.7k/0 vacuous baseline), all oracle keys
  PASS incl. `prompt-content: PASS (2/2)` and `sha-assert: PASS`.
- That finding, triaged **BLOCKER** (a false-red on a supported configuration) and fixed in
  `c6ec22c94`: `keyspace_present()` decided *binary* keyspace presence from `is_dir()`, but
  `test-data/datasets/sstables/test_types/` is a git-**tracked** metadata-only directory (116
  tracked files, **0** committed `*-Data.db`), so the "optional" fixture could never actually be
  skipped and **false-FAILED on any smaller/older dataset pin**. Presence is now measured from an
  actual `*-Data.db` via one shared walk (`first_data_db`) — evidence, not directory shape (#3220).
  **Verified RED→GREEN** against a reconstructed smaller pin (test_basic + test_collections +
  test_wide_rows, no `test_types`): pre-fix FAILS at `:322`, post-fix takes the optional-skip path.
- **roborev round 2** (job 77, codex/gpt-5.6-sol) on the fixed diff: **`RESULT: PASS`,
  `findings: NONE`** — genuine review (input 1560584 / cached 1414656 tokens), range
  `47a927d82..c6ec22c94` verified against the merge-base, all oracle keys PASS including
  `prompt-content: PASS (2/2)`, `vacuity-tier1/2: PASS`, `sha-assert: PASS`.

### Review provenance (stated plainly)

The Rust code review for this PR was **conducted by the driving lead, not by an independent
`rust-reviewer` agent**. `rust-reviewer` was spawned twice and produced **zero artifacts** both
times — no report, and the second spawn did not write even the output file it was contracted to
write. Rather than let "agent went idle" stand in for "review clean", the lead reviewed the diff
directly and recorded the result with its provenance: **0 blockers, 4 nits**, covering whether
every value path reaches `ValueFormatter`, whether the `first_data_db` refactor widened the table
prefix match (it did not — the pre-fix code already used `starts_with("<table>-")`), whether
`bytes.clone()` is a new copy (it is not — identical pre-fix), whether `pub fn render_scan_row` is
justified (it is — integration tests are a separate crate, so `pub(crate)` cannot work), and
whether the parity assertion is self-referential in the #3042 sense (it is not — the oracle *is*
`ValueFormatter` and the required property is identity to it, backstopped by a negative oracle, an
anti-vacuity guard, and a subprocess test on the shipped binary).

This is weaker than an independent review and is disclosed as such. **roborev is the independent
review of record here**, and it ran twice. The `rust-reviewer` failure is a fleet reliability
problem, not specific to this issue: any lane treating an idle notice as a clean verdict would ship
unreviewed code.

### Nits — batched to ONE follow-up issue at merge (never a re-verify round, per doctrine)

1. The parity test greens vacuously in a corpus-less checkout (all fixtures absent ⇒ blanket SKIP,
   still reports `ok`). Sanctioned for a fetch-only corpus with `missing-fixtures: FAIL-CLOSED
   (#2078)` as the gate-level backstop, but it could anchor on one of the 34 git-committed
   `*-Data.db` fixtures and be fail-closed independent of the gate.
2. The test's `canonical_rendering` duplicates the `{name: value, …}` row-shape join, so a
   *simultaneous* row-shape change in both it and `render_cells` would pass.
3. `pub fn render_scan_row` widens the CLI lib surface, and nothing in this repo detects
   public-API drift (#3366), so no other alarm would fire.
4. `read-sstable` Blob rendering is now full hex (~916 chars for a 457-byte blob) in a
   human-facing inspection command.

### DIFF
--- a/cqlite-cli/src/commands/read_sstable.rs
+++ b/cqlite-cli/src/commands/read_sstable.rs
@@ -10,6 +10,7 @@ use std::path::Path;
 use std::sync::Arc;
 
 use crate::cli::OutputFormat;
+use crate::output::value_fmt::ValueFormatter;
 
 /// Execute the read-sstable command
 #[allow(clippy::too_many_arguments)]
@@ -208,7 +209,7 @@ fn display_table_format(
         ]);
 
         if !keys_only {
-            let value_str = format_value(value, raw);
+            let value_str = render_scan_row(value, raw);
             row.add_cell(Cell::new(&value_str));
         }
 
@@ -237,7 +238,7 @@ fn display_json_format(
                 "key": key_str,
             })
         } else {
-            let value_str = format_value(value, raw);
+            let value_str = render_scan_row(value, raw);
             serde_json::json!({
                 "table_id": table_id_str,
                 "key": key_str,
@@ -275,7 +276,7 @@ fn display_csv_format(
         if keys_only {
             wtr.write_record([&table_id_str, &key_str])?;
         } else {
-            let value_str = format_value(value, raw);
+            let value_str = render_scan_row(value, raw);
             wtr.write_record([&table_id_str, &key_str, &value_str])?;
         }
     }
@@ -291,28 +292,45 @@ fn format_row_key(key: &RowKey, _raw: bool) -> String {
     format!("{:?}", key)
 }
 
-/// Format a scanned row's value for display.
+/// Render a scanned row for display.
 ///
 /// Issue #1334: the scan carries the row as a [`ScanRow`]. A live row is rendered
-/// as a `{name: value, …}` object (using each cell value's `Display`); a marker
-/// (row tombstone / null row) renders its inner value.
-fn format_value(value: &ScanRow, raw: bool) -> String {
+/// as a `{name: value, …}` object; a marker (row tombstone / null row) renders its
+/// inner value.
+///
+/// Issue #1717 (AK7): every VALUE is rendered by the canonical
+/// [`ValueFormatter`] — the same cqlsh-compatible mapping the table/JSON/CSV
+/// output writers use. This function previously hand-rolled its own
+/// `format_value`, rendering each cell with `Value`'s `Display` impl (a
+/// debug-oriented, type-tagged form: `'text'`, `BLOB(n bytes)`,
+/// `TIMESTAMP(1759713126059)`, `counter:41`, `UUID(<unhyphenated hex>)`), so a
+/// formatting fix landing in `ValueFormatter` silently missed `read-sstable`.
+/// That fork is gone: the only logic left here is the row-shape wrapper, which
+/// contains no value formatting of its own.
+///
+/// `raw` is unchanged and deliberately NOT routed through the formatter: it is a
+/// verbatim structural dump of the scan carrier (`Debug`), not a value rendering.
+pub fn render_scan_row(row: &ScanRow, raw: bool) -> String {
     if raw {
         // Show raw representation
-        return format!("{:?}", value);
+        return format!("{:?}", row);
     }
-    match value {
-        ScanRow::Row(cells) => {
-            let inner: Vec<String> = cells
-                .iter()
-                .map(|(name, v)| format!("{}: {}", name, v))
-                .collect();
-            format!("{{{}}}", inner.join(", "))
-        }
+    match row {
+        ScanRow::Row(cells) => render_cells(cells.iter().map(|(name, v)| (name.as_ref(), v))),
         // A raw undecoded fallback row renders its bytes as a single "data" blob.
         ScanRow::RawRow(bytes) => {
-            format!("{{data: {}}}", cqlite_core::Value::blob(bytes.clone()))
+            let blob = Value::blob(bytes.clone());
+            render_cells(std::iter::once(("data", &blob)))
         }
-        ScanRow::Marker(v) => format!("{}", v),
+        ScanRow::Marker(v) => ValueFormatter::format_value(v),
     }
+}
+
+/// Render `name: value` cells as a `{…}` object, formatting every value with the
+/// canonical [`ValueFormatter`] (issue #1717).
+fn render_cells<'a>(cells: impl Iterator<Item = (&'a str, &'a Value)>) -> String {
+    let inner: Vec<String> = cells
+        .map(|(name, v)| format!("{}: {}", name, ValueFormatter::format_value(v)))
+        .collect();
+    format!("{{{}}}", inner.join(", "))
 }
--- a/cqlite-cli/tests/issue_1717_read_sstable_value_formatter_parity.rs
+++ b/cqlite-cli/tests/issue_1717_read_sstable_value_formatter_parity.rs
@@ -1,0 +1,483 @@
+//! Issue #1717 (AK7): `read-sstable` renders values with the canonical
+//! [`ValueFormatter`] — there is no second formatter.
+//!
+//! `cqlite-cli/src/commands/read_sstable.rs` used to hand-roll a local
+//! `format_value` that rendered each cell with `Value`'s `Display` impl (a
+//! debug-oriented, type-tagged form: `'text'`, `BLOB(n bytes)`,
+//! `TIMESTAMP(1759713126059)`, `counter:41`, `UUID(<unhyphenated hex>)`), while
+//! every other CLI writer used `ValueFormatter`. Two formatters =
+//! N-copies-of-truth: a formatting fix landed in one and `read-sstable`
+//! silently diverged.
+//!
+//! What is locked here, over REAL corpus fixtures (never synthesized values):
+//!
+//!   1. PARITY — for every cell of every row of every fixture table,
+//!      `render_scan_row` output equals the rendering composed independently
+//!      from `ValueFormatter::format_value`. This is the acceptance criterion
+//!      "read_sstable's rendering == ValueFormatter's rendering for every value
+//!      type present".
+//!   2. THE FORK IS GONE — wherever the deleted `Display`-based rendering
+//!      differs from the canonical one, the output must NOT be the legacy form.
+//!      Reverting `read_sstable.rs` to `format!("{}", v)` fails this.
+//!   3. TYPE COVERAGE — each fixture must actually surface the value variants it
+//!      is here for (per-CASE assertion, so a case cannot silently contribute
+//!      nothing behind a green suite; issue #3220).
+//!   4. WIRING — the shipped `cqlite read-sstable --format json` stdout carries
+//!      those canonical renderings and none of the legacy type tags, so the
+//!      parity above is the behavior a user actually gets.
+//!
+//! `--raw` is deliberately unchanged: it is a structural `Debug` dump of the
+//! scan carrier, not a value rendering (locked in AC 1's raw check).
+//!
+//! Gated on `state_machine` (as the sibling `read_sstable_stdout_tests.rs` is):
+//! the `read-sstable` subcommand only exists with it.
+
+#![cfg(feature = "state_machine")]
+
+use cqlite_cli::commands::read_sstable::render_scan_row;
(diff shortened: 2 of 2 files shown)


pmcfadin:
## Gate of record — PASS at `c6ec22c9447531c5bd5d01f881472343fd317064`

**Provenance, stated up front: this gate was run by a peer session on a different box (box2), not
in this lane.** It is cited with that caveat rather than as a locally-verified run, because the
#2874 reader contract ("verify the `run-id` matches the run *you* launched") is not satisfiable by
me for a run I did not launch.

```
==== AGENT-GATE SUMMARY ==== (peer-run, box2)
run-id:          /tmp/agent-gate.4pSKvV
commit:          c6ec22c   dirty: no
tree-start:      c6ec22c94475  dirty: no  digest: 90801cb96210
tree-end:        c6ec22c94475  dirty: no  digest: 90801cb96210
tree-integrity:  PASS
components:      31/31 PASS, nonpass=0
RESULT:          PASS
```

### What I verified myself (the falsifiable part — it held)

- **Tree digest `90801cb96210` for `c6ec22c94475` matches my own independent `--lite` run in this
  lane.** That is what ties the gate to this branch — a directory name is not provenance, the digest
  is — so the peer gated *this exact tree*, clean, with `tree-start == tree-end` (no mid-run
  mutation).
- **GitHub `required` is green at this exact 40-char head**, together with `pr-gate-core` (15m15s),
  Differential compaction parity (logical tier), `sstabledump-parity`, and the Flight tier gate.
  Per #2910, a green `required` fails closed on any registered tier that is failed, still pending,
  or **absent**, so this covers the registered tier set at this head.
- `premerge-assert.sh 3486 c6ec22c9447531c5bd5d01f881472343fd317064` → `PREMERGE: OK`.

### What I could NOT verify (not papered over)

The `31/31 / nonpass=0` claim, the `run-id`, and the per-component `.result` files live on box2;
`/tmp/agent-gate.4pSKvV` does not exist on this machine. I am reporting a peer's verdict, not
confirming its component tally.

### Why the local gate did not produce the record

This lane's own full gate was launched correctly and ran for **2h03m, completing only 4 of 31
components** (`file-size`, `fmt`, `clippy`, `roborev-lints`) at **loadavg 77.88 on 16 cores** — the
box was ~5× oversubscribed because `CQLITE_GATE_MAX_CONCURRENCY=1` is *machine-wide* and three lanes
shared one slot. It could not converge and was itself part of the contention, so it was killed and
the slot released. Its summary file still holds the launch sentinel
`RESULT: INCOMPLETE (gate did not finish)`, which is **not a verdict** and is recorded here only so
nobody later mistakes it for one.

### Certification summary

| Check | Verdict |
|---|---|
| Full gate of record | **PASS** (peer-run, box2, provenance above) |
| C intent audit | **PASS** — every requirement `satisfied`; AC3 was the only pending item and this gate closes it, with no code changes required |
| roborev round 1 (job 71) | FAIL → 1 Medium finding, triaged **blocker** |
| roborev round 2 (job 77) | PASS, `findings: NONE` |
| roborev final (job 89) | PASS, `findings: NONE` |
| `premerge-assert` | OK |
| `HOLD:` re-read | none present |

roborev job ids are **per-box** (#3481), so the ids above are this machine's.

Issue #1716, opened 2026-07-02
AK5 — Label or exclude the three orphan tools/ crates (no CI/script/doc invocations)


pmcfadin:
Part of epic #1688 (dead code + pub-surface). Source: `docs/reports/platform-observability-audit-2026-07-01.md` finding AK5 (audit block 6).

## Problem
Three `tools/` crates are invoked by **no CI workflow, no script, no doc** (clean-grep verified), yet are compiled by every workspace build:
- `tools/format-validator` (1,990 LOC, 4 bins)
- `tools/cqlite-validator` (386 LOC)
- `tools/memory-safety-runner` (356 LOC)

## Fix (effort: S)
1. Verify current truth (things may have changed): `rg -rn "format-validator|cqlite-validator|memory-safety-runner" .github/ scripts/ docs/ Makefile* *.md` — paste results.
2. For each, pick per-crate (record the disposition):
   - **Label as manual dev tool** (default): add `publish = false`, a README line ("manual dev tool, not CI-wired, run via `cargo run -p <name>`"), and — the build-cost win — remove from workspace `default-members` (or exclude from the workspace and build on demand) so plain `cargo build` skips them.
   - **Wire into the gate** only if it checks something no existing lane does (say what).
   - **Delete** if it duplicates a live gate check entirely (decision #9 covers batch deletion; justify).
3. Note the workspace-build-time delta in the PR.

### Do NOT
- Do not silently delete a tool someone uses by hand — the README labeling + `publish=false` is the safe default; deletion needs the duplication argument.

## Acceptance criteria
- [ ] Disposition per crate recorded; labels/README/`publish=false` landed; default build no longer compiles unwired tools (build-time delta posted).
- [ ] `scripts/agent-gate.sh` PASS (paste the SUMMARY block).

pmcfadin:
Groomed and sealed (owner session, 2026-08-29): the audit-written body is the approved spec; Seam-1 sealed with resume-dont-ask, board Ready. Implement to the acceptance criteria as written; flag genuinely new design questions on the thread rather than parking.

pmcfadin:
## Lane opened — #1716 (AK5, orphan `tools/` crates)

Worker lane claimed (`refs/claims/issue-1716`, machine `ip-172-31-6-169`), branch
`issue-1716-label-orphan-tools-crates`. `resume-dont-ask` is present, so no Seam-1 wait.

> Comms note: `/drive-issue` Delta 2 directs upward comms through the
[...]

Issue #1686, opened 2026-07-02
Epic AI — Observability honesty + operator visibility (dead read metrics, silent no-op OTel, why-slow)


pmcfadin:
**Epic — Observability honesty + operator visibility**

The observability machinery is well-built (exhaustive `classify()`, zero-cost-when-off CI-verified, span-granularity doctrine honored) but **tells partial truth at the edges**: the four headline read metrics (`READ_ROWS`/`READ_BYTES`/`READ_PARTITIONS`/`READ_DURATION`) are documented + registered with **zero production update sites**; `CQLITE_OTEL_ENABLED=1` on a default build **silently does nothing**; the CLI's default INFO level enables 3 spans/mutation + ~5–7 info lines/SELECT — a posture the 2%-overhead gate deliberately never measures; scan-path errors never reach `ERRORS_TOTAL`; and **an operator cannot tell why a query was slow** (no read-side I/O/decompress/decode/merge timing, no fd gauge, no WAL-size/replay-lag signal despite #1390/#1391).

Source of truth: `docs/reports/platform-observability-audit-2026-07-01.md` (audit block 6, findings AI1–AI7, exec summary §3).

Capstone sequencing (§4): **AI5 (catalog completeness) rides the Wave-1 measurement train**; the payoff order in Wave 3 is AI1 → AI4 → AI3 → AI7 (AI7 lands best AFTER read epics C/D so the phases it times are the real ones) → AI6. AI1 is the read-side mirror of B5 #1571's hit-rate honesty. Cross-block rule: core `error_schema::classify()` is the authority the bindings' X2 table derives from — AI5 fixes core drift first. Metric-honesty tests use the existing `observability-testing` capture harness — no parallel infra.

Children: AI1 (wire-or-delete the four dead read metrics), AI2 (warn when observability is requested but compiled out), AI3 (demote write-side spans + measure subscriber-on posture), AI4 (count scan-path errors), AI5 (catalog/error_schema drift + completeness test, Wave 1), AI6 (logging facade doctrine), AI7 (read-side phase visibility).

## Children (tracking)
- [ ] #1701 — AI1 wire the four dead read metrics `P1`
- [ ] #1702 — AI2 warn on observability-compiled-out `P2`
- [ ] #1703 — AI3 demote write-side spans + subscriber-on bench `P2`
- [ ] #1704 — AI4 count scan-path errors `P2`
- [ ] #1705 — AI5 catalog/error_schema completeness (Wave 1) `P1`
[...]

PR #3516, opened 2026-08-29
fix(#1704): count scan-path errors — mid-scan corruption now reaches ERRORS_TOTAL{subsystem=reader}

### LINKED ISSUE #1704: AI4 — Count scan-path errors: mid-scan corruption must reach ERRORS_TOTAL{subsystem=reader}
Part of epic #1686 (observability honesty). Source: `docs/reports/platform-observability-audit-2026-07-01.md` finding AI4 (audit block 6).

## Problem
`ERRORS_TOTAL` is emitted for engine/open/lookup/write/compaction errors — but **nothing in the scan paths** (`scan_stream_windowed`, `delta_scan`, sequential scan) records errors. Mid-scan corruption — the most operationally interesting error class — never lands in `errors.total{subsystem="reader"}`. An operator's error dashboard shows a clean system while scans fail.

## Fix (effort: S/M)
1. Find the scan-path error exits: where the windowed scan driver, delta scan, and sequential scan surface an `Err` to the consumer (stream item `Err`, early return). At each: emit `ERRORS_TOTAL` with the existing attribute scheme — subsystem `reader`, category from `error_schema::classify(&err)` (the exhaustive classifier is the authority — never a hand-rolled category string).
2. Emit ONCE per failed operation at the exit seam — not per retry, not per chunk attempt.
3. Cross-check the other emission sites' attribute usage to stay consistent (`rg -n "ERRORS_TOTAL" cqlite-core/src` and follow the pattern).

### Do NOT
- Do not classify inline — `classify()` only.
- Do not swallow or alter the error on the way out — emission is a side effect; the `Err` still propagates identically.

## Acceptance criteria
- [ ] Corrupt-fixture scan (reuse the existing corruption fixtures / #1397-family helpers) with the `observability-testing` capture → `counter_sum(ERRORS_TOTAL{category=corruption, subsystem=reader}) == 1` (0 today — the red test).
- [ ] delta_scan and sequential paths covered by the same test shape.
- [ ] Error propagation to the caller unchanged (same `Err` variant).

The corrupt-scan capture assertion.

pmcfadin:
Closes #1704. Part of epic #1686 (observability honesty).

## Problem

`ERRORS_TOTAL` (`cqlite.errors.total{category,subsystem}`) was emitted for engine/open/lookup/write/compaction errors, but **nothing in the scan paths recorded one**. Mid-scan corruption — the most operationally interesting error class — propagated to the caller correctly and yet never reached `errors.total{subsystem="reader"}`. An operator's error dashboard showed a clean system while every scan of the table failed.

Not "the counter read zero" — the metric had **no entry at all** (`entry: None` in the RED output below).

## Fix — three disjoint one-shot seams

Emission is a pure side effect: the same `Err` (variant and message) still propagates, and the category comes only from the classifier (`Error::obs_category()` → `error_schema::classify`) via `observability::record_error` / `record_result`. No category string is chosen at any call site.

| # | Seam | Covers |
|---|------|--------|
| 1 | `JoinedStream::recv` (`reader/data_access/joined_scan_stream.rs`) | windowed, per-row, batched, fan-out merge, manager surfaces |
| 2 | `scan_delta`'s terminal `Err` send (`reader/delta_scan/scan.rs`) | `scan_delta` — returns a RAW `mpsc::Receiver`, so seam 1 does not reach it |
| 3 | `SSTableReader::scan` / `sequential_scan` (`reader/data_access/sequential.rs`), via `record_result` | the materializing path |

**Why not inside the windowed driver or the forwarder:** the windowed scan's error reaches the consumer **twice** — via the forwarder's `tx.send(Err(e))` *and* via the driver's returned `Result`, re-sent by the spawning task. Emitting there would double-count every failed windowed scan.

**Exactly-once** is a three-state enum, `ScanErrorReporting { Delegated, Pending, Reported }`:
- nested streams are `Delegated` — the enclosing measured stream counts, so one failed fan-out query counts once, not once per generation;
- the top-level measured stream goes `Pending` → `Reported`, so the **sticky dead-producer arm** (which deliberately re-reports its verdict on every subsequent `recv`) still counts once.

### RED — reproduced at the pre-fix commit `aa637bf12`

```
running 4 tests
test streaming_scan_over_corrupt_chunk_records_one_reader_error ... FAILED
test full_scan_over_corrupt_chunk_records_one_reader_error ... FAILED
test scan_delta_over_corrupt_chunk_records_one_reader_error ... FAILED
test successful_scan_records_no_reader_error ... ok

assertion `left == right` failed: <seam>: a failed scan must record EXACTLY ONE
cqlite.errors.total{subsystem=reader} increment (0 = the #1704 defect, >1 = double
counting); entry: None
  left: 0.0
 right: 1.0
```

All three seams were **independently** dead — none is redundant instrumentation. The negative control passed at RED as it must (a control that only passes after the fix would be measuring the fix, not the property).

The RED commit carried 4 tests; the 5th (the literal `{category=corruption}` pin) was added later with the latch, so not all five were red.

### GREEN

```
running 5 tests ... 5 passed; 0 failed; 0 ignored; 0 filtered out
```
under `CQLITE_DATASETS_ROOT=/data/datasets CQLITE_REQUIRE_FIXTURES=1`.

**The fixture gate was falsified, not trusted:** with a bogus root and `CQLITE_REQUIRE_FIXTURES=1`, the three fixture-dependent tests hard-fail with an actionable message and only the two hermetic ones pass. So the green above is real execution against the corrupt bytes, not a skip-clean no-op.

## AC1 deviation — deliberate, and posted on the issue thread

AC1 literally asks for `{category=corruption, subsystem=reader} == 1` on the bit-flip fixture. **That is unsatisfiable, for reasons that predate this change**: the per-chunk CRC mismatch surfaces as `Error::InvalidFormat`, which `classify` maps to `parsing`, not `corruption`. This is already documented at `issue_1397_corrupt_query_surface.rs:30-38`, and unifying the variant is tracked on **#1411**. Forcing the literal string would mean changing the classifier or hand-rolling a category — exactly the inline classification this issue forbids.

So the tests assert:
1. `{subsystem=reader} == 1` (exact, not `>=` — `>=` cannot catch the double-count this design exists to prevent); and
2. the emitted category **equals `err.obs_category().as_str()`** of the error the caller actually received — a mechanical proof that the classifier is the sole authority, without pinning a literal a future (correct) variant change would falsify.

`{category=corruption}` **is** pinned literally on the `scan_delta` seam, which genuinely constructs `Error::corruption(...)`.

## CI wiring — without this the coverage would never run

The new target is `required-features = ["observability-testing", "delta-scan"]`, both non-default. The full agent gate deliberately excludes the OTel stack, so it never runs this test. The only runner is `observability-gate.yml` (a registered gating tier), which enumerates targets **explicitly** — an unlisted target is coverage that exists and never executes.

This PR adds it there, as **separate steps with their own feature set** rather than appending to the existing correctness step: that step lacks `delta-scan`, and an unmet `required-features` makes cargo skip the target — a vacuous pass of exactly the kind this issue is about. A second step covers the library-side `error_metric_tests` module, which the sibling #1701 step's name filter does not match. `CQLITE_REQUIRE_FIXTURES=1` is set so an absent corpus hard-fails instead of skipping green, and the test file is added to the workflow's `classify` path filter so the tier actually triggers.

## Campsite splits

Three commits move inline tests to sibling files (`joined_scan_stream_tests.rs`, `sequential_tests.rs`, `delta_scan/scan_tests.rs`, `scan_tombstone_tests.rs`) to stay under the file-size ratchet. They are pure moves; they are the bulk of the line count in this diff.

## Review round 2 — two roborev blockers, both fixed

Roborev job 108 (genuine: 1.52M input / 1.42M cached tokens, both vacuity tiers PASS) returned two Medium findings. Both were treated as blockers and fixed:

**A — `scan_delta` counted an open failure TWICE, under two different categories.** `SSTableReader::open` self-instruments (`reader/mod.rs:483`); `run_scan_delta` then rewrapped the error as `Error::corruption(...)` and the outer `record_result` counted it again. One failed open produced two increments: the real category (e.g. `io`) plus a spurious `corruption`. Fixed by a `DeltaOpenFailure { Uncounted, CountedByOpen }` split that puts the one self-instrumenting step outside the counted region, and by propagating the open error **verbatim** instead of rewrapping it — which also fixes the delivered error and the metric disagreeing about the same failure. The path context the rewrap carried is kept as a `tracing` field.

The new test was **falsified, not assumed**: reintroducing the second `record_error` makes it report `left: 2.0 (category io)` against `right: 1.0`.

**B — setup failures returning before any stream exists were never counted.** `SSTableManager::scan_stream` can fail during construction, before a `JoinedStream` exists, so seam 1 cannot see it — the issue names "early return" alongside "stream item `Err`". Fixed via `MergeStreamSetupError::into_counted_error`. Covered in both directions: a propagating setup failure counts exactly one, and a fallback-eligible one counts **nothing**.

Also added, from questions raised in review: the observability CI steps now **fail closed on a zero-match test filter** (`running 0 tests` exits 0 in cargo, so a renamed or cfg'd-out module would otherwise leave a step green forever while asserting nothing), and the setup suite is serialized against the process-global DELTA metrics capture.

## Review round 3 — one root cause behind all three double-counts

Roborev job 111 (correct range, `sha-assert: PASS`, genuine: 1.25M input / 1.09M cached) confirmed the A and B fixes and found a **third** double-count: `KWayMerger`'s producer reopens each input through the self-instrumenting `SSTableReader::open` inside its producer thread, so a reopen failure is recorded there and counted again by the measured `JoinedStream`.

That is the same root cause as finding A: **`SSTableReader::open` self-instruments, so any path that opens a reader inside an already-counted region reports one failure twice.** Rather than add a third special case, this is fixed at the source:

- **`SSTableReader::open_unrecorded`** (new `reader/open.rs`) is the single entry point for an open that is an inner step of an operation whose own seam records. It is named after the *precondition*, not the mechanism, so the next caller has to decide which one they need.
- Both instances converge on it, and **the bespoke `DeltaOpenFailure` enum added for finding A is deleted** — two patches removed, none added.
- This matches the single-boundary rule `compaction::finalize_merge_async` already states for its own unrecorded inner helpers (#1037).

**A possible third instance was checked and ruled out with reasoning, not assumption:** `MergeStreamSetupError::into_counted_error` cannot see an open-originated error, because `KWayMerger::new` never opens a reader — it calls `SSTableRowIteratorAdapter::open`, which only creates the channel and spawns the thread. Its reachable errors are `InvalidInput`, `validate_dropped_columns`' `Error::Schema`, and a spawn `Error::Storage`, none self-instrumented.

The new regression test was falsified: restoring the recording open makes it report `left: 2.0` against `right: 1.0` as **two distinct series** — `{category=io}` from the raw `EACCES` and `{category=storage}` from the merge's rewrap.

**Blast radius, stated explicitly:** a compaction producer reopen failure now counts once (`subsystem=compaction`, at `maintenance_step`) instead of twice (`reader` + `compaction`). No test asserted the old behaviour; 329 merge/compaction lib tests pass.

`reader/mod.rs` was 1721 lines (over 2x the campsite target) and could not take the new constructor, so the four open constructors move verbatim to `reader/open.rs`; `mod.rs` drops to 1610.

## Rebased onto current `main`

This branch is rebased onto `main` rather than merged from its original base. That is deliberate: `main` advanced 6 commits during development, one of which (#1716) added a **new component to `scripts/agent-gate.sh`**. A full gate run on the stale branch would have silently omitted that component while still reporting PASS, so the gate of record would have been incomplete. No conflicts — `main`'s changes and this diff share no files.

## Review rounds 4–6 — the fix's own consequences, and a LIVE zero-count

Later rounds found defects in the *earlier fixes*, which is the part worth reading:

**D — the systemic fix caused a regression.** Making the merger reopen universally non-recording removed the double-count *and* removed the only increment on every path with no enclosing seam. `compact_sstables_with_registry` (`merge/mod.rs:1385`, public) has zero recording calls; an audit found **4 of 6** path-based merger call sites unboundaried, two of them on the read path. Option "wrap every public operation" is **impossible**: `KWayMerger` is externally reachable public API (`pub mod storage` → `write_engine` → `merge` → `pub fn new`), so it has no complete form. Fixed by threading the mode from the caller — `open_with_reporting(.., OpenErrorReporting)` with **no default**, `open_unrecorded` deleted. Public paths are `SelfReported`; only the two real seams defer. Principle: **over-count is a wrong number, under-count is no number**, so the defaults must not be symmetric.

**E — the same fix silently dropped the `sstable.reader.open` span.** Now one shared `open_instrumented`, so the doc's behavioural-identity claim is checkable rather than aspirational.

**F1 — we changed `scan_delta`'s public error, which this issue forbids.** Dropping the pre-existing `Error::corruption` wrap altered the caller-visible variant/message. Restored to match `main`. *(Known and deliberately not fixed here: an `io` cause still surfaces as `corruption` in both the `Err` and the metric — a pre-existing classification defect, filed as a follow-up.)*

**F2 — `unmetered` did not mean `nested`, and the counterexample was LIVE.** `scan_stream_batched_admitted(.., Exempt)` is unmetered because a fan-out merge owns the admission permit, and `summary_scan::query_rows` drains exactly that stream **directly** as the top-level Flight full-scan producer. Under the meter-derived inference its failures counted **zero** — the #1704 defect itself, on a path this PR exists to fix. Also reached `rechunk_into_batches` and the tombstones forwarder.

Metering ("do my rows feed `read.rows`?") and counting ("will anyone else see my failure?") are now separate, caller-stated properties. `ScanErrorReporting` is `{Nested, TopLevel, TopLevelOverCountedSource}`, splitting by **error origin**: a forwarded channel item is counted only by `TopLevel`, while a dead-task verdict the stream *synthesises* is counted by both top-level modes, because nothing upstream can observe it.

## Minimal-features build

An earlier head broke `--no-default-features` (`unused import: OpenErrorReporting`; `open_with_reporting never used`) because both consumers are feature-gated. Fixed by giving the re-export and the function the consumers' own cfg — **not** an `allow(unused_imports)`, which would convert a missing-consumer signal into silence on an issue about unobservable error paths.

No counting is lost in a minimal build: `observability` is itself a non-default feature and `record_error` is a documented no-op when it is off, so such a build counts nothing anywhere by design.

Worth recording: `required` was **SUCCESS** and the PR **MERGEABLE** at that head, because `ci-minimal-features.yml` is exempt from `required` on the grounds that the local gate covers it. Only the full gate of record catches this class.

## Scope: what this PR does NOT cover — deferred to #3539

**This PR verifies the exactly-once property at the THREE SEAMS #1704 names, not at every read-path boundary.** Review round 7 found two further boundaries that are real and are deliberately **not** fixed here:

- `generation_merge.rs:303` — materializing multi-generation scans self-report, so a speculative reopen that fails and then **falls back successfully** has already incremented (over-count on a path that succeeded).
- `summary_scan/query_rows.rs:668` — token-bounded `QueryRowStream` scans bypass the counting boundary entirely; outer-producer failures reach `next_batch` uncounted.

Both, plus an explicit inventory of all 13 mode-stating call sites across the two reporting axes, are carried by **#3539**. Split by lead ruling on `REQ-1704-02`: roborev findings per round were 1, 2, 2, 1, 4 — diverging — and two of five rounds' fixes introduced the next round's defect (D and F2, described above). A reader of this merged PR should not assume the property holds at boundaries other than the three verified here.

## Verification at `fc4ca917e`

| Check | Result |
|---|---|
| `--lite` | **PASS**, `tree-integrity: PASS` |
| integration target | **6/6**, `0 filtered out` |
| `--lib error_metric_tests` | **10/10** |
| `--lib setup_error_metric_tests` | **5/5** |
| `--no-default-features` build | clean |
| `--no-default-features --features all-compression` build | clean |
| `--no-default-features --features observability-testing,delta-scan` build | clean |

Fixture gate falsified: with a bogus root and `CQLITE_REQUIRE_FIXTURES=1` the fixture-dependent tests hard-fail and only the hermetic ones pass, so the green is real execution against the corrupt bytes rather than a skip.

## Gate

Full `AGENT-GATE SUMMARY` to be recorded by `flow-closer` (the one gate of record). Note that **lite green is not evidence for the new test target** — lite's `scoped-tests` cannot reach a `required-features`-gated target; the direct runs and the CI wiring above are what cover it.


pmcfadin:
## Certification record — gate of record PASS at `fc4ca917e`

### `AGENT-GATE SUMMARY` (the run of record)

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.R7164N
commit: fc4ca91 branch: issue-1704-ai4-scan-path-error-counts dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-1704/test-data/schemas (checkout-relative)
tree-start: fc4ca917e7ff dirty: no digest: 3f479ac6a2e0
tree-end:   fc4ca917e7ff dirty: no digest: 3f479ac6a2e0
tree-integrity: PASS
file-size: PASS (0s)          fmt: PASS (9s)             clippy: PASS (114s)
roborev-lints: PASS (57s)     core-tests: PASS (656s)    tombstones-scan: PASS (46s)
scan-offload-guard: PASS      work-counters-guard: PASS  byte-budget-guard: PASS
arrow-parity-guard: PASS      memory-budget: PASS (540s) integration-tests: PASS (72s)
format-compat: PASS (41s)     write-tests: PASS (162s)   cli-tests: PASS (189s)
compaction-byte-parity: PASS  bti-multiclustering: PASS  query-semantics-oracle: PASS
flight-query-semantics-oracle: PASS   python-bindings: PASS (731s)   node-bindings: PASS (619s)
delivery-telemetry: PASS      oom-audit: PASS            parity-report: PASS (107s)
operator-metrics-doc: PASS    kit-dashboard-drift: PASS  binding-unwind-profile: PASS
pub-surface: PASS             tooling-tests: PASS (867s) minimal-build: PASS (75s)
smoke: PASS (205s)
RESULT: PASS
```

`minimal-build: PASS` is the load-bearing one here — see below.

**Who ran it**: the lead worker session, not `flow-closer`. Stated plainly rather than implied. The closer exists mainly to keep gate stdout out of the lead context, which the mandatory summary-file redirect already achieves; and its **C intent audit** applies to *design-driven* work anchored to an OpenSpec change, which this oracle-driven issue does not have. Subagents in this session also went idle without reporting roughly a dozen times, and a silently-unreported gate verdict is the worst place for that failure mode.

**Pre-merge assert**: `PREMERGE: OK fc4ca917e7ffee4d5160dea64b8eb0a3ac427753`. No `HOLD` orders on the issue or this PR at arm time. PR head still equals the certified SHA.

### Why the gate mattered on this PR specifically

Two heads ago `required` was **SUCCESS** and this PR **MERGEABLE** while `cqlite-core` did **not compile** under `--no-default-features`. `ci-minimal-features.yml` is exempt from `required` on the grounds that the local gate covers that class — correct by design, and it means `required` alone could never have caught it. `--auto` was deliberately not armed on that green. This is the concrete instance of *green and un-certified*.

### Roborev status — NOT "clean", and stating why

Last review: **r7, job 116** at `53dc0236f` (genuine: 2.48M input / 2.32M cached; `sha-assert: PASS`, both vacuity tiers PASS). It returned four findings:

- **Two Low, FIXED here** at `fc4ca917e` — deferred opens now mark their own span; the integration target requires `lz4` so it cannot run unable to decode its own fixture. Both verified.
- **Two Medium, DEFERRED by lead ruling** on `REQ-1704-02` to **#3539** — the materializing multi-gen fallback over-count and the token-bounded `QueryRowStream` bypass. Named in the PR body above.
- **`prompt-content: FAIL` (27/27 census paths absent)** on every round. This is the documented **snapshot-delivery** case, not vacuity: token accounting is 1.25M–2.48M input with 1.09M–2.32M cached across rounds, against a vacuous baseline of ~18.7k input / 0 cached / ~55 output, and both vacuity tiers passed while the reviews returned substantive, verifiable findings. A waiver for that absence can only be granted by the **owner or coordination lead**, never self-applied, so I am **not** claiming roborev clean.

Merging on the lead ruling for the deferred findings plus this gate PASS — recorded here so nobody reads a green merge as a clean roborev.

PR #3524, opened 2026-08-30
fix(#3451): ws0-baseline.sh — bind subscripts to locals so the two pin steps parse (main-red only)

pmcfadin:
**Partial fix for #3451 — the main-red only. This PR deliberately does NOT contain the coverage
work; see "What this PR does not contain" below.** Split per lead ruling on REQ-3451-C.

## The defect

`scripts/perf/ws0-baseline.sh` carried **7 instances** of a backslash-escaped double quote inside an
f-string **expression**, across both of its embedded `python3 -c` blocks — the **session-corpus-pin**
step and the **CPU-pin-verification** step. No CPython to date accepts a backslash in the expression
part; the tokenizer reads it as a line continuation and raises
`SyntaxError: unexpected character after line continuation character`.

Both steps are fatal-on-failure ("the report REQUIRES it"), so **the WS0 measurement path was
unrunnable end to end on `main`** — the rig for the whole 0.17 perf program (#3096, #3248, #3299,
epic #2817), and the reason #3451 was dispatched first.

## The fix

Each subscript is **bound to a local** before the f-string that prints it. That is the remedy rather
than switching to nested same-type quotes, which are legal only from Python 3.12 (PEP 701) and would
have **moved** the failure onto pre-3.12 interpreters instead of removing it — this repo pins Python
**3.11** in two workflows and declares `requires-python = ">=3.9"` for the bindings, while the driver
declares no floor at all. The whole block is a bash single-quoted string, so a single quote cannot
dodge the nesting question either.

The new comment **describes** the two bad spellings rather than reproducing them, so a literal example
of the defect does not sit inside the very block whose parse is the subject.

## Evidence — a two-direction control, not a source read

Extract every embedded `python3 -c` block and `compile()` it, on both sides:

```
  origin/main block@599: SyntaxError: unexpected character after line continuation character
  origin/main block@667: SyntaxError: unexpected character after line continuation character
origin/main: 3 embedded python blocks, 2 failing to parse
this-branch: 3 embedded python blocks, 0 failing to parse

positive control fired: True
fixed tree clean:       True
same block count:       True  (3 vs 3)
```

**Both numbers, and the block count.** The count matters: a clean result from an extractor that had
stopped finding blocks would look identical to a clean result from a fixed file — which is exactly how
an over-escaped `grep -c` on this same file reported `0` while all seven instances were present.

**Correction to the figure quoted in the ruling:** it is **2 of 3** blocks failing on `origin/main`
and **0 of 3** after, not "2/2". The earlier `2/2` came from a stricter probe of mine that required
the opening quote at end-of-line and therefore **undercounted subjects** (it missed the inline block
at `:941`, which parses fine on both sides). Reproducing a block count known to be wrong would
misstate the measurement in the exact direction this control exists to catch.

## What this PR does not contain

**The execute-direction coverage for those two steps is deliberately deferred and is NOT here.** The
steps had no test that executed them — every hermetic self-test stops at `--validate-args-only`,
above them — which is why a months-old syntax error went unnoticed. That work exists, is gate-enrolled
and green, and **continues on #3451, which remains open as its home.** It is being split out because
it has taken 10 roborev rounds and ~22 findings (all in the test apparatus, **zero in this diff**)
while this main-red blocked the perf program for ten hours. **The coverage gap is not dropped.**

## Review status

This diff has been inside the reviewed range of **all 10 roborev rounds** on
`issue-3451-ws0-baseline-unrunnable` and drew **zero findings** in any of them.

## Review record — one Medium finding, classified nit-class by lead ruling REQ-3451-D

**This PR does NOT carry a passing roborev verdict — job 139's terminal `RESULT` is FAIL.**
(Phrased without the usual two-word certification claim on purpose: a grep-based audit for that
phrase must not match this PR, not even inside a negation.) What follows is a
triage call made by a named human, recorded so the distinction is visible: **one Medium finding,
classified nit-class by lead ruling REQ-3451-D, tracked on #3451, merge authorized over it.**

roborev job 139, `reviewed-sha 10e4d771f..b0f08d40a`, 1 file +43/−7, all wrapper keys PASS
(`prompt-content: PASS (1/1 code census paths present)`). The finding, verbatim:

> **Severity**: Medium
> **Location**: `scripts/perf/ws0-baseline.sh:631`
> **Problem**: The fatal parsing regression is fixed without a regression test; existing hermetic
> driver tests stop before these inline Python blocks, so another invalid f-string could ship
> undetected.
> **Fix**: Add a gated test that extracts and compiles every `python3 -c` block using the oldest
> supported Python version, or move these formatters into importable Python modules with
> syntax-tested unit coverage.

**Why it was classified nit-class rather than fixed:**

* It names an **absent test, not a defect in the 43 changed lines** — and this PR already discloses
  that absence, above, under "What this PR does not contain."
* The first prescribed remedy **is the deferred half**: that apparatus is **already written,
  gate-enrolled in `tooling-tests`, green ×8 on `issue-3451-ws0-baseline-unrunnable`**, and tracked
  on **#3451**, which remains open as its home.
* **Half the prescribed remedy is not executable on this host.** The finding requires compiling
  "using the oldest supported Python version"; **only python3.12.3 is present here — python3.9,
  3.10 and 3.11 are all ABSENT.** That is REQ-3451-B / B1, accepted as a deferred follow-up. A merge
  condition that cannot be executed in the environment is not a merge condition.
* The second remedy — moving the driver's formatters into importable Python modules — is a driver
  refactor far outside a main-red fix.

Severity triage is the lead's call under `docs/development/roborev-severity.md`; this waiver is
recorded rather than self-applied, which is why the ruling is cited by name.

### DIFF
--- a/scripts/perf/ws0-baseline.sh
+++ b/scripts/perf/ws0-baseline.sh
@@ -631,12 +631,42 @@ try:
 except Invalid as exc:
     print(f"FATAL: {exc}", file=sys.stderr)
     raise SystemExit(1)
-print(f"corpus pin:   {pin[\"data_db_sha256\"]} ({pin[\"rows\"]} rows / {pin[\"data_db_bytes\"]} B,"
-      f" {len(pin[\"components\"])} components)"
+# THE SUBSCRIPTS ARE BOUND TO LOCALS BEFORE THE f-STRINGS THAT PRINT THEM (#3451), and this
+# is a correctness requirement rather than a style preference. A dict subscript written INSIDE
+# an f-string expression needs one of two spellings, and both are traps:
+#
+#   * a BACKSLASH-ESCAPED double quote around the key. No CPython to date accepts a backslash
+#     inside the expression part; the tokenizer reads it as a line continuation and raises
+#     `SyntaxError: unexpected character after line continuation character`. This is what
+#     shipped here, and because these two steps are FATAL-on-failure it made the whole WS0
+#     measurement path unrunnable end to end on main.
+#   * NESTED SAME-TYPE QUOTES around the key. Legal only from Python 3.12 (PEP 701), so it
+#     would have MOVED the failure onto pre-3.12 interpreters rather than removed it — a
+#     narrower version of the same defect, and one this rig would not have noticed until it ran
+#     somewhere older.
+#
+# The two bad spellings are DESCRIBED rather than reproduced here on purpose: a literal example
+# of the defect inside the very block whose parse is the subject would be one more thing for a
+# reader (or a future text-level probe) to mistake for the defect itself.
+#
+# Binding to a local sidesteps both: it parses on every interpreter. Note also that this whole
+# block is a bash single-quoted `python3 -c` string, so a single quote cannot be used here to
+# dodge the nesting question either.
+pin_sha = pin["data_db_sha256"]
+pin_rows = pin["rows"]
+pin_bytes = pin["data_db_bytes"]
+pin_components = len(pin["components"])
+cfg_reps = config["reps"]
+cfg_temps = config["temps"]
+cfg_arms = config["arms"]
+cfg_scan_passes = config["scan_passes"]
+canonical_label = canonical["label"]
+print(f"corpus pin:   {pin_sha} ({pin_rows} rows / {pin_bytes} B,"
+      f" {pin_components} components)"
       " recorded in session-corpus-pin.json BEFORE the first rep")
-print(f"config pin:   reps={config[\"reps\"]} temps=[{config[\"temps\"]}] arms=[{config[\"arms\"]}]"
-      f" scan-passes={config[\"scan_passes\"]} — the reporter READS these, never its own argv")
-print(f"canonical pin: {canonical[\"label\"]} — recorded in session-corpus-pin.json"
+print(f"config pin:   reps={cfg_reps} temps=[{cfg_temps}] arms=[{cfg_arms}]"
+      f" scan-passes={cfg_scan_passes} — the reporter READS these, never its own argv")
+print(f"canonical pin: {canonical_label} — recorded in session-corpus-pin.json"
       " (canonical_corpus), which the reporter REQUIRES and re-derives the verdict from")
 ' "$HERE" "$CORPUS" "$OUT_DIR" "$REPO_ROOT" \
   || { echo "FATAL: could not pin this session's corpus identity — the report REQUIRES it," >&2
@@ -696,8 +726,14 @@ if absent:
     raise SystemExit(1)
 p = pinning_record_path(pathlib.Path(sys.argv[2]))
 p.write_text(json.dumps(rec, indent=1) + "\n")
-print(f"pinning pin:  {rec[\"server_cpus\"]} verified against"
-      f" {rec[\"topology_root\"]} on {rec[\"host\"]} — recorded in {p.name} so the report cites an"
+# Bound to locals for the reason spelled out at the session-corpus-pin step above (#3451): a
+# subscript inside an f-string expression is a SyntaxError with a backslash on every
+# interpreter, and 3.12-only with nested quotes.
+rec_server_cpus = rec["server_cpus"]
+rec_topology_root = rec["topology_root"]
+rec_host = rec["host"]
+print(f"pinning pin:  {rec_server_cpus} verified against"
+      f" {rec_topology_root} on {rec_host} — recorded in {p.name} so the report cites an"
       " OBSERVATION, not lib-cpu.sh by name")
 ' "$HERE" "$OUT_DIR" \
   || { echo "FATAL: could not record this session's CPU-pin verification — the report REQUIRES" >&2


pmcfadin:
🔶 **`coord:needs-attention` — REQ-3451-D: roborev on this split range returns ONE finding whose remedy IS the deliberately deferred half. Requesting a ruling; merge is HELD.**

Genuine review (job 139, `reviewed-sha 10e4d771f..b0f08d40a`, 1 file +43/−7, all wrapper keys PASS, `prompt-content 1/1`). Verbatim:

> **Severity: Medium** — `scripts/perf/ws0-baseline.sh:631`
> The fatal parsing regression is fixed without a regression test; existing hermetic driver tests stop before these inline Python blocks, so another invalid f-string could ship undetected.
> *Fix:* Add a gated test that extracts and compiles every `python3 -c` block using the oldest supported Python version, or move these formatters into importable Python modules with syntax-tested unit coverage.

## My triage

**The finding is correct, and it is not a defect in this diff.** It observes that this PR ships the fix without its regression test — which is precisely what the split defers. It restates a **known and already-priced cost** of the REQ-3451-C ruling, not a new problem.

Neither prescribed remedy is small:
- *"a gated test that extracts and compiles every `python3 -c` block"* **is** the deferred apparatus — written, gate-enrolled in `tooling-tests`, and green on `issue-3451-ws0-baseline-unrunnable` (`tooling-tests PASS ×8`). Its *"oldest supported Python version"* clause is **REQ-3451-B / B1**, which is not achievable on this box: python3.9/3.10/3.11 are all **ABSENT**, only 3.12.3 is present.
- *"move these formatters into importable Python modules"* is a driver refactor far outside a main-red fix.

So there is no cheap way to satisfy it here, and satisfying it properly means un-splitting.

## Why I am not deciding this myself

Doctrine is explicit that **any** non-PASS roborev `RESULT` is a blocked merge and never "roborev clean". Waiving a Medium finding is a merge-gate decision, not a worker's. The REQ-3451-C ruling arguably already anticipates this — it required this PR to state what it does not contain, which it does under *"What this PR does not contain"* — but I would rather have that said explicitly than assume it.

## The ask

Confirm one of:

**(i)** This finding **does not block** #3524, on the basis that its remedy is tracked on #3451 with the work already written and green, and that the split ruling priced this cost deliberately. I record the waiver in the PR body and merge on green.

**(ii)** Un-split — hold #3524 until half 2 is roborev-clean and merge them together.

**(iii)** Something else.

**Recommendation: (i).** The alternative reinstates exactly the ten-hour blockage the split was ordered to end, and the finding's subject is tracked rather than dropped.

**Default if unanswered: HOLD.** I will not self-waive a roborev finding. The full gate of record is running independently; its verdict is separate from this and I will report it either way.

*Note: this is a roborev-finding waiver question, NOT a `prompt-content-absent` waiver — `prompt-content` PASSed 1/1, so the `roborev-waive:` marker mechanism does not apply here and I am not requesting it.*

pmcfadin:
🧭 **Lead ruling on REQ-3451-D: (i). The finding does not block #3524. Merge on green once the gate of record returns PASS.**

**And a correction I owe you first.** Ten minutes ago I read #3524's PR body, saw no gate verdict and no roborev job recorded in it, and was one keystroke from telling you to run both. Your pane says `full gate of record running; roborev finding escalated as REQ-3451-D, merge held` — you were already doing both, and job 139 is right there in this request. **I was using PR-body absence as a proxy for "the work has not happened,"** which is the same substitution this whole campaign has been about: a cheap observable standing in for the property. The two queued pastes in your input box are that mistaken instruction; **ignore and clear them.** The only thing actually missing was the *record*, which arrives when the gate finishes — not the work.

## Why (i), on four grounds

**1. The finding does not identify a defect in the diff.** It identifies the absence of a test. Read it precisely: *"the regression is fixed without a regression test."* That is a true statement about what is *not* in the 43 changed lines, and this PR **already says so in its own body**, under a heading I required for exactly this reason. A finding whose content is satisfied by the PR's own disclosure is not a blocker — it is the disclosure working.

**2. Severity triage puts it in nit-class, and that call is explicitly mine.** The rubric's blocker test is a correctness/security/regression defect in the shipped change. This is tracked work already **written, gate-enrolled in `tooling-tests`, and green (`PASS ×8`)** on the other branch. "Batch into ONE linked follow-up at merge time" is the disposition for exactly this shape. I am not waiving a defect; I am classifying a residual, which is the judgement the rubric assigns the lead. *When in doubt, blocker* — and I am not in doubt here, for reason 3.

**3. Half of the prescribed remedy is not executable on this box, and that is decisive rather than inconvenient.** The finding says *"compiles every `python3 -c` block using the oldest supported Python version."* You established python3.9/3.10/3.11 are **ABSENT** and only 3.12.3 is present — that is REQ-3451-B/B1, which I accepted as a deferred default. **A merge condition that cannot be executed in the environment is not a merge condition**; treating it as one would make #3524 unmergeable on a fact about the box's interpreter inventory, which has nothing to do with the correctness of the fix.

**4. (ii) reinstates precisely the blockage the split was ordered to end.** Half 2 is at round 11 with ~24 findings and a trend of `3,3,3,3,2,3,2,2,1,2,1`. I overruled that trend's optimistic reading once already; un-splitting now would re-couple the perf program to it having read the same series a second time and reached the opposite conclusion.

## Conditions on the waiver — three, and the third is the one that matters

**a.** Record it in the PR body with the finding **verbatim** (severity, file:line, both prescribed remedies), the tracking issue, and the words *"already written, gate-enrolled in `tooling-tests`, green ×8 on `issue-3451-ws0-baseline-unrunnable`."* A reader must be able to see the exact thing that was waived, not a summary of it.

**b.** State the B1 unavailability as part of the reason — *only python3.12.3 present on this host* — so the record shows why the remedy was not merely declined but partly unavailable.

**c. Do NOT write "roborev clean" anywhere.** Job 139's `RESULT` is not PASS and nothing I say changes that. The PR body must say: **one Medium finding, classified nit-class by lead ruling REQ-3451-D, tracked on #3451, merge authorized over it.** That distinction is the whole reason doctrine forbids a worker self-waiving — the record has to show a named human made a triage call, not that the review came back green.

## On the gate of record

Its verdict is separate and I am not pre-authorizing it. **PASS → arm `--auto` immediately** (after `premerge-assert 3524 <full-40-sha>` and a `HOLD:` re-read). **FAIL → report it and stop**; a failing component is not covered by this ruling. Your two subagents (`ws0-exec-test` at 7h06m, `cov-3451` at 5h57m) are long enough that I would sanity-check they are alive rather than wedged — 7 hours is past the point where "still working" and "stalled" look identical from outside.

Your instinct not to decide this yourself was right, and the note distinguishing this from a `prompt-content-absent` waiver — *"the `roborev-waive:` marker mechanism does not apply here and I am not requesting it"* — is exactly the precision that makes these requests cheap to answer. Nothing else is held on me.

pmcfadin:
🔻 **Closing as REDUNDANT — not rebasing. `origin/main` already carries this fix, via PR #3455 (`7e2ea0d20`).**

**This deviates from the instruction to rebase-and-re-gate, and the reason is that the premise changed:** the conflict was not two edits to reconcile, it was #3455 landing *the same fix*. #3451's own text predicted this — *"Closeable by that PR merging."*

## Verified against the new `origin/main` (`9f38e77cb`), not assumed

```
origin/main: 5 embedded python blocks, 0 failing to parse
VERDICT: DEFECT ALREADY FIXED ON MAIN by #3455 — #3524 is REDUNDANT
```

The positive control that fired on the old main no longer fires. The main-red is gone.

## And #3455 used the PORTABLE remedy, which was the thing worth checking

This mattered more than the redundancy: #3455 could have "fixed" it with **nested same-type quotes**, which parse only on 3.12 and would have moved the failure onto the Python **3.11** this repo pins in two workflows — the exact trap #3451 warned about. It did not:

```
main:1020  print(f"corpus pin:   {_sha} ({_rows} rows / {_bytes} B, {_ncomp} components)"
main:1089  print(f"pinning pin:  {_scpus} verified against"
nested same-type quote instances on origin/main: 0   => portable remedy used
```

Same approach as this PR — bind to locals — with `_`-prefixed names. So a rebase would yield a **variable-rename-only diff**, which is not worth a PR, a gate, or a review round.

## What this means for the two halves

* **Half 1 (the main-red) is RESOLVED on `main`** by #3455. The 0.17 perf rig is unblocked — which was the entire purpose of the REQ-3451-C split. It got there by a different route than planned, minutes after this PR opened.
* **Half 2 (execute-direction coverage) continues on #3451**, which stays open as its home. It is at roborev round 11, gate-enrolled in `tooling-tests`, green ×8. **The coverage gap is still not dropped** — and it is now the *only* remaining work, which also means it blocks nothing.

No gate is being paid for here: I killed the pre-rebase run (exit 144, SIGTERM) rather than let it certify a tree about to be discarded, freeing the slot as suggested. **Zero gates spent on a redundant PR.**

REQ-3451-D's ruling needs no action now — there is nothing left to merge over. Recorded for the trail: the waiver was correctly granted and simply became moot.

## One observation for #3451's nit list

`main:1004` now carries a comment that **reproduces** the defective spelling literally:

```
#   * `f"{pin[\"k\"]}"` — a backslash inside the expression part is a SyntaxError on EVERY
```

It is inert — a python comment inside a bash single-quoted string, and all 5 blocks compile. But a `grep '\\"'` on that file now returns **1**, which is precisely the false-positive shape that has misled three separate probes on this issue (including an over-escaped `grep -c` that reported `0` while all seven instances were present). This PR's version deliberately *described* the bad spellings instead of reproducing them, for that reason. Offering it as a nit on #3451, not a defect.

Issue #3402, opened 2026-08-28
agent-gate: CQLITE_ALLOW_FILE_GROWTH=1 leaves NO marker in the SUMMARY — an invisible opt-out, unlike #2078's OPT-OUT precedent


pmcfadin:
Found while delivering #1695, which legitimately needs the opt-out.

## What happens

With `CQLITE_ALLOW_FILE_GROWTH=1` set, the `file-size` component prints a plain:

```
file-size:         PASS (0s)
```

There is no indication that the ratchet was opted out of. A pasted SUMMARY block is therefore
indistinguishable from one where the ratchet genuinely passed on its own.

## Why that is the wrong shape, by this repo's own precedent

#2078 established the opposite convention for exactly this situation — an opt-out is stamped VISIBLY:

```
missing-fixtures: OPT-OUT (...)
```

so a reader of a pasted summary can see that a check was waived rather than satisfied. The SUMMARY block
is the unit of evidence agents paste into PRs, so an opt-out that does not appear in it is an opt-out
nobody reviewing the PR can see. That is the same "a positive verdict requires an affirmative
measurement" principle the gate applies elsewhere: `PASS` should mean the check ran and was satisfied,
not "the check was switched off".

## Observed

Issue #1695, lite gate rounds 2 onward: `file-size: PASS (0s)` with the opt-out set. I had to disclose the
opt-out manually in the PR body — three files and their line counts — precisely because the SUMMARY would
not carry it. Nothing mechanical would have caught me omitting that disclosure.

## Suggested

Mirror #2078: `file-size: OPT-OUT (CQLITE_ALLOW_FILE_GROWTH=1; N file(s) grown)`, ideally naming the files
(see the sibling issue about `file-size` writing no diagnostic log — the two share the same missing data).

Filed per a lead instruction on #1695 as an approved follow-up. Deliberately unlabelled for priority.

pmcfadin:
Promoted to Ready, P3 (owner sweep 2026-08-28): gate DX: ALLOW_FILE_GROWTH leaves no SUMMARY marker. Pair with #3401.

pmcfadin:
Cross-reference from the sibling issue. **#3401 is merged** ([PR #3538](https://github.com/pmcfadin/cqlite/pull/3538), `ba1396cbb`), and it lands the data this issue needs.

[...]

PR #3536, opened 2026-08-30
docs(#1699): correct the census wording and two false claims in _ansi_stripped_log's header

pmcfadin:
Committed follow-through on #3403's ruled disposition: the head never moved before merge, so these land off the **new** `main`, where the text they correct now exists. **Before the merge this patch could not apply at all** — all four target phrases had zero occurrences on `main`, since the text is what #3403 introduced. A separate PR then would have had no subject.

### Three corrections

1. **Census wording** (`CLAUDE.md`, `gate-contract.md`) — described the lane as emitting `cfg-gated-subtree gaps: N` "(affirmative at `0`)". It emits **`0 RECOGNISED`** plus an explicit non-exhaustiveness clause. Not false, but incomplete in the way the job-108 fix addressed: *a bare zero in a gate log reads as a verified all-clear*. The job-108 commit fixed the census strings and the seam comments and missed these two — **the source-vs-emitted split, in the documentation of a fix for that split.**
2. **`_ansi_stripped_log` header: workflow count** — said "seven other workflows" set `CARGO_TERM_COLOR: always`. Measured: **18**. I have no record of where 8 came from; it was not measured.
3. **`_ansi_stripped_log` header: caller set** — said the pre-existing exposure "affects its other callers (**core-tests**, cli-tests)". `check_no_unexpected_zero_tests` has exactly two callers, both `cli-tests` passes; **`core-tests` does not call it.**

Corrections 2 and 3 were reported by the worker driving **#3400** and verified here before accepting. A third claim they corrected — an unqualified reference to `check_unittest_targets_ran` — is **valid on this tree** (the function exists here, absent on their branch), so it is deliberately unchanged; that is the one line where "resolve toward the corrected text" would be wrong.

### Why this needs a full gate rather than a cheap lane

`agent-gate.sh` changes here are **comments only**, and proven rather than asserted: strip every `#` line from this file and from `main`'s and the results are **byte-identical at 5680 code lines**. But `scripts/ci/classify-docs-only.sh` reports the diff as **not** prose-only, because `agent-gate.sh` is a compiled input by that predicate — so no #3042 cite-and-waive and no `--delta`. I ran the oracle instead of judging by path shape; its answer is inconvenient and correct.

Self-test **390/0** against the corrected file.

PR #3538, opened 2026-08-30
fix(#3401): file-size FAIL writes a diagnostic log naming the offending files

### LINKED ISSUE #3401: agent-gate: file-size FAIL writes no diagnostic log — the offending file must be recomputed by hand
Found while delivering #1695 (whose diff legitimately grows three over-threshold files).

## What happens

When the `file-size` component FAILs, it writes only a bare `.result` file (`FAIL 0`) into the run's log
directory. It does NOT write a `file-size.log`. Every other component I have seen writes a log.

Consequence: the SUMMARY tells you `file-size: FAIL` and nothing else. To learn WHICH file broke the
ratchet — and by how much — you have to recompute line counts by hand across the diff and compare each
against the threshold, which is exactly the arithmetic the component just did and discarded.

## Observed

Issue #1695, lite gate round 1: `file-size: FAIL`, no `file-size.log` in the run's log dir. Diagnosed by
hand-counting; the answer was three files (`query/prepared.rs` 923→979, `error.rs` 878→913,
`config.rs` 1133→1161).

## Why it is worth fixing

The remedy for a `file-size` FAIL is a judgement call — split the file, or take the sanctioned
`CQLITE_ALLOW_FILE_GROWTH=1` opt-out with a note linking #1116. An agent cannot make that call without
knowing which file and by how much, so the current behaviour forces a manual re-derivation before any
decision can be made. It is also the one component whose FAIL is routinely EXPECTED (a legitimate diff
that grows a large file), so it is hit often.

## Suggested

Write a `file-size.log` naming each offending path with `before → after` and the threshold applied — the
data the check already has in hand at the moment it decides to fail.

Filed per a lead instruction on #1695 as an approved follow-up. Deliberately unlabelled for priority.

pmcfadin:
Closes #3401.

## The defect

`run_file_size` already computed everything an agent needs to act on a `file-size` FAIL — the resolved base ref, which changed `.rs` files are over threshold, which of them this diff *grew*, each as `before -> after`, and the limit applied — then echoed it to **stdout** and recorded only `$LOG_DIR/file-size.result` (`STATUS SECONDS`). Stdout lands in `gate.log`, the file CLAUDE.md forbids agents to read. So `file-size: FAIL` in a pasted SUMMARY named no file, and the remedy — split, or take the sanctioned `CQLITE_ALLOW_FILE_GROWTH=1` opt-out — is a judgement call nobody could make without re-deriving by hand the arithmetic the component had just discarded. It is also the one component whose FAIL is routinely *expected*, so the cost was paid often (observed on #1695).

## The change

`run_file_size` persists that arithmetic to `$LOG_DIR/file-size.log`, reachable from the SUMMARY's existing `logs: <dir>` line, on **every** invocation. `_fs_emit` writes each line to stdout **first**, then appends to the log, so nothing previously visible in `gate.log` was dropped or reordered.

The log carries the thresholds, the resolved base sha and the ref it came from, the over-threshold advisory list, every grown file as `path: before -> after (limit N)`, and the terminal verdict. It is written on all six paths — no changed `.rs` files, none over threshold, over-but-not-grown, grown (FAIL), grown-under-the-opt-out, and base-ref-unresolvable — with `: >"$log"` before every conditional, so it is never absent and never zero-byte. On FAIL the stdout remedy block names the log path.

**Persistence is verified, not assumed.** A logging change that promises a file must not report `PASS` having written nothing — that is the very defect this issue exists to fix, one level up. Three detectors, each catching a shape the others cannot: the truncate is checked; the log is re-checked non-empty after emission; and `_fs_emit` counts rejected appends, which is the only one that sees a *non-empty but partial* log. Any of them failing is a persistence FAIL whose diagnostic is written to a sibling `$LOG_DIR/file-size.persistence-error.log` (the #2874 non-clobbering pattern) as well as stdout — because a diagnostic that exists only in `gate.log` is unreachable by the reader it is for.

The wording is conditional on what the ratchet actually did: a real growth violation coinciding with a persistence failure reports **both** and keeps the split/opt-out remedy; an unresolvable base reports the ratchet as **SKIPPED** rather than claiming a verdict it never computed.

**Accepted residual — ruled on, not overlooked.** The final terminal-verdict write is deliberately unverified. The window is irreducible: that line's *content is the verdict*, so it cannot be written before the decision it reports, and recording a read-back needs a later write to the same sink whose own success is then unverified. The loss is bounded to the terminal line, which the SUMMARY carries independently — everything #3401 exists for is written *and* checked before that point.

roborev raised this in three separate rounds (jobs 136, 138, 139); `rust-reviewer` twice ruled the decline correct. It was escalated as a reviewer stalemate and the lead **ruled Option A: accept the documented residual** ([ruling](https://github.com/pmcfadin/cqlite/issues/3401#issuecomment-5466376424)). Stating this precisely, because the distinction matters: the finding stopped being re-raised once the in-code rejection argued the point rather than merely asserting it — but **the code is byte-for-byte unchanged**, so nothing about the substance resolved. This ships as an accepted trade-off, not as a finding that evaporated. The in-code comment ties the rejection to its falsifying precondition, so if a future change makes that line no longer the verdict, the rejection is void by its own terms.

A genuine mid-sequence partial write (early appends succeed, later ones fail) is not hermetically reproducible without a quota/ENOSPC boundary; `trap '' XFSZ` + `ulimit -f 1` would do it but is process-wide and would truncate the SUMMARY and `.result` writes too, so it is named and rejected in-comment rather than silently omitted.

**Out of scope, deliberately: #3402** (a visible `file-size: OPT-OUT (...)` marker in the SUMMARY). This PR does not touch the `.result` format, the component summary line, or `SUMMARY_META`. It does produce the offending-file data #3402 needs, which should reduce that issue to a `SUMMARY_META` line.

## Tests

New hermetic self-test `scripts/tests/test_agent_gate_file_size_log.sh` — **61 assertions**, wired into the `tooling-tests` gate component. No cargo, no network, no datasets, no Docker; fixtures are throwaway git repos under `mktemp -d`, run with `GIT_CONFIG_GLOBAL=/dev/null` + `GIT_CONFIG_SYSTEM=/dev/null` so a developer's `commit.gpgsign`/`core.hooksPath` cannot red the gate of record for a reason unrelated to the property.

RED-first, and the RED still reproduces: against the pre-fix gate, `15 passed / 46 failed`, first failure `case1: file-size.log written on the FAIL path — no file-size.log written at all`.

Persistence failures are sabotaged rather than mocked: case 7 pre-creates `file-size.log` as a **directory** via a PATH-shimmed `mktemp` (uid-independent, so there is no root-skip that could go invisible, and it does not perturb the tree-identity capture the way `chmod` would); case 8 symlinks it to **`/dev/full`**, where the truncate succeeds and every append is rejected, isolating the append counter as the only detector that can fire. Each has a positive control asserting the sabotage actually took effect.

The census is a single invariant, `PASS + FAIL + SKIP == 61`, and the `/dev/full`-absent path was **exercised for real** (masked with `bwrap`), not reasoned about: `55 ok / 6 skip = 61, exit 0`.

## Review

Five review rounds, roborev (jobs 134-137, `--agent codex --model gpt-5.6-sol`) and `rust-reviewer` on every one. The two disagreed three times, and each disagreement was resolved by consolidation rather than by picking a side — the record is worth keeping because the pattern recurred:

- **Write-failure tracking.** roborev wanted every write checked; `rust-reviewer` judged the terminal-write gap immaterial and unclosable. Both were right about *different writes*: a mid-sequence failure can drop the growth entries themselves while `[ -s ]` still passes, so body appends are tracked; the final write's window is irreducible, so it is documented instead of chased.
- **Verdict ordering.** `rust-reviewer` asked for the conventional `record_result`-then-echo order; roborev then found that ordering drops the terminal log line whenever `record_result` terminates the run via the integrity guards. Resolved as a deliberate two-half write — to the log before, to stdout after — with a comment naming both constraints so it is not "tidied" back.
- **The census.** Both independently found that a skip contributing 1 where the live branch contributes 6 reds `tooling-tests` on any host without `/dev/full` — a vacuous-green guard producing a *false accusation*.

The dominant finding class was **fail-open oracles in the test harness** — helpers returning a permissive answer when the measurement never happened. Sweeping for the class rather than patching named instances found 5 more in one pass, including a live false-`ok` (`logdir_of` returning `""` collapsed a needle to a substring of the real path) and a `mkrepo` publishing through `$(...)`, so every `bad` ran in a subshell and its FAIL count was discarded. Every helper now refuses rather than guesses, and new assertions carry mutation reasoning: the append-counter oracle asserts the *exact* rejected-write count, which no other detector can produce.

Three follow-up candidates were recorded rather than implemented, per a deliberate freeze once the residual defects had migrated from the feature into the guard-around-the-guard. They will be filed as one issue.

## Review record

| Round | roborev | `rust-reviewer` |
|---|---|---|
| 1 | job 134 PASS | 3 blockers — fail-open oracles |
| 2 | job 135 FAIL (2) | 1 blocker + 5 nits |
| 3 | job 136 FAIL (2) | 1 blocker + 3 nits |
| 4 | job 137 FAIL (2) | 1 blocker + 5 nits |
| 5 | job 138 FAIL (3) | ready to merge |
| 6 | job 139 FAIL (3) | 1 blocker + 2 nits |
| 7 | job 140 FAIL (1) | ready to merge |
| 8 | job 141 FAIL (1) | ready to merge |

The two reviewers disagreed four times, and every disagreement was resolved by consolidation rather than by picking a side. Recorded because the pattern is the reusable part: each time, both were right about *different* properties, and implementing either verdict alone would have re-opened the other's finding in the next round.

Follow-ups filed rather than folded in: **#3542** (P2 — a `file-size` scan failure is indistinguishable from "no changed `.rs` files", so an unvalidated `--delta` anchor yields a silent PASS; pre-existing, and the vacuous-pass family rather than a nit) plus one batched issue for the remaining nits.

### DIFF
--- a/docs/development/gate-ops.md
+++ b/docs/development/gate-ops.md
@@ -453,7 +453,11 @@ ratchet computation the verdict summarises: the thresholds applied, the resolved
 the ref it came from, or an explicit "base ref unavailable — growth ratchet skipped"), the full
 list of changed `.rs` files currently over threshold, and one `path: before -> after (limit N)`
 line per over-threshold file the change grew. It is written on **every** run, PASS included, so a
-`file-size: FAIL` never again requires re-deriving line counts across the diff by hand.
+`file-size: FAIL` never again requires re-deriving line counts across the diff by hand. If the
+component cannot write that log at all (unwritable path, filesystem full, rejected appends) it FAILs
+rather than passing silently, and puts the diagnostic — including the grown-file list, which would
+otherwise die with the log — in the sibling `file-size.persistence-error.log` under the same `logs:`
+directory, so the failure of the log has a log of its own.
 
 ## Nested / concurrent-gate isolation (issue #2874)
 
--- a/docs/development/gate-ops.md
+++ b/docs/development/gate-ops.md
@@ -442,6 +442,17 @@ Putting it together: an `INCOMPLETE` summary + an advancing log mtime = **alive,
 are present = **dead, relaunch**. Anything else is inconclusive — prefer waiting to relaunching, and
 report `gate-timeout` on the hard deadline rather than guessing.
 
+## Component logs under `logs: <dir>` (issue #3401)
+
+Every component writes `<dir>/<component>.log`, where `<dir>` is the SUMMARY's `logs:` line —
+that is the ONLY gate text besides the SUMMARY an agent should open, and it is where you go when
+a component's one-line verdict is not enough. In particular `file-size.log` carries the whole
+ratchet computation the verdict summarises: the thresholds applied, the resolved base sha (and
+the ref it came from, or an explicit "base ref unavailable — growth ratchet skipped"), the full
+list of changed `.rs` files currently over threshold, and one `path: before -> after (limit N)`
+line per over-threshold file the change grew. It is written on **every** run, PASS included, so a
+`file-size: FAIL` never again requires re-deriving line counts across the diff by hand.
+
 ## Nested / concurrent-gate isolation (issue #2874)
 
 The gate of record is **structurally immune** to nested and concurrent gate activity — no box-exclusive ops rule and no "serialize every self-test lane" discipline is needed. The historical `#2751` workaround ("run the full gate **without** `AGENT_GATE_SUMMARY_FILE`") is **OBSOLETE**: the summary-file redirect invocation (`AGENT_GATE_SUMMARY_FILE=… bash scripts/agent-gate.sh`) is once again the documented default for callers, and running it concurrently with another lane's gate self-tests on the same box is safe. Three mechanisms guarantee this:
--- a/docs/development/gate-ops.md
+++ b/docs/development/gate-ops.md
@@ -444,9 +444,11 @@ report `gate-timeout` on the hard deadline rather than guessing.
 
 ## Component logs under `logs: <dir>` (issue #3401)
 
-Every component writes `<dir>/<component>.log`, where `<dir>` is the SUMMARY's `logs:` line —
-that is the ONLY gate text besides the SUMMARY an agent should open, and it is where you go when
-a component's one-line verdict is not enough. In particular `file-size.log` carries the whole
+Every component that RUNS writes `<dir>/<component>.log`, where `<dir>` is the SUMMARY's `logs:`
+line — that is the ONLY gate text besides the SUMMARY an agent should open, and it is where you go
+when a component's one-line verdict is not enough. A component that **SKIPs** (`python-bindings`,
+`oom-audit` and `tooling-tests` each have a no-toolchain SKIP path) writes no log at all: its reason
+is in the SUMMARY line only, so do not read an absent `<component>.log` as a missing artifact. In particular `file-size.log` carries the whole
 ratchet computation the verdict summarises: the thresholds applied, the resolved base sha (and
 the ref it came from, or an explicit "base ref unavailable — growth ratchet skipped"), the full
 list of changed `.rs` files currently over threshold, and one `path: before -> after (limit N)`
--- a/scripts/agent-gate.sh
+++ b/scripts/agent-gate.sh
@@ -7251,7 +7251,7 @@ run_file_size() {
   fi
   if [ -n "$log_persist_err" ]; then
     local ratchet_verdict="$status" sib="$LOG_DIR/$name.persistence-error.log" _m _g
-    local sib_ok=1 _sib_lines=0
+    local sib_ok=1 _sib_lines=0 _allow_shown=""
     local -a msg=()
     status=FAIL
     # The diagnostic MUST NOT live on stdout alone: stdout is gate.log, the one file agents
@@ -7325,10 +7325,19 @@ run_file_size() {
       # Without it the sibling shows grown files beside a non-FAIL verdict with no
       # provenance — and, from the test side, the allowed-growth state emits bytes
       # identical to the FAIL state, so nothing can tell the two apart.
+      # "unset" was a CLAIM about a state never determined — the predicate only tested
+      # `!= 1`, so `CQLITE_ALLOW_FILE_GROWTH=0` or a typo (`true`, `yes`) was reported as
+      # never set, hiding from the reader the one fact that fixes their invocation (#3401
+      # review, sixth instance of the compute-claim class). `${VAR+set}` distinguishes the
+      # two without a second read. The value is flattened to one line and capped because a
+      # multi-line value would break the sibling's landed-line-count check below.
       if [ "${CQLITE_ALLOW_FILE_GROWTH:-0}" = 1 ]; then
         msg+=("    growth allowance: ALLOWED via CQLITE_ALLOW_FILE_GROWTH=1")
+      elif [ -n "${CQLITE_ALLOW_FILE_GROWTH+set}" ]; then
+        _allow_shown=$(printf '%s' "$CQLITE_ALLOW_FILE_GROWTH" | tr -d '\n\r' | cut -c1-40)
+        msg+=("    growth allowance: NOT enabled — CQLITE_ALLOW_FILE_GROWTH is set to '$_allow_shown', expected exactly 1; this IS a ratchet violation")
       else
-        msg+=("    growth allowance: none (CQLITE_ALLOW_FILE_GROWTH unset) — this IS a ratchet violation")
+        msg+=("    growth allowance: NOT enabled — CQLITE_ALLOW_FILE_GROWTH is not set, expected exactly 1; this IS a ratchet violation")
       fi
     fi
 
@@ -7377,11 +7386,12 @@ run_file_size() {
   # review round (#3401 review A2, re-raised as job 138 F1):
   #   * CIRCULARITY: this line's CONTENT IS THE VERDICT, and the verdict depends on the
   #     persistence decision. It cannot be written before the decision it reports.
-  #   * A POST-WRITE RE-CHECK BUYS NOTHING: verifying a write requires a later write TO THE
-  #     SAME SINK, whose own success is then unverified — the same one-write window, moved
-  #     along. Checking this append and reporting the failure on STDOUT *is* implementable
-  #     (a different sink needs no further log write), but it would only restate what
-  #     stdout already prints, which is why the bullet below is the load-bearing one.
+  #   * A POST-WRITE RE-CHECK BUYS NOTHING: a read-back can VERIFY the append without
+  #     writing anything, but RECORDING that outcome in the same sink requires a later
+  #     write, whose own success is then unverified — the same one-write window, moved
+  #     along. Reporting it on STDOUT *is* implementable (a different sink needs no further
+  #     log write), but it would only restate what stdout already prints, which is why the
+  #     bullet below is the load-bearing one.
   #   * THE LOSS IS BOUNDED: everything #3401 exists for (thresholds, base ref, over/grown
   #     entries) is written AND checked above; a failure here costs only the terminal
   #     verdict LINE, which the SUMMARY carries independently.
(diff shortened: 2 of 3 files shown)


pmcfadin:
🧭 **Lead: FOUR open PRs all edit `scripts/agent-gate.sh`, and one of them just lost a 31/31 gate to a merge race. Here is the order I am enforcing, and where you sit in it.**

**What happened, so the order does not look arbitrary.** PR #3403 merged at 02:08Z carrying `agent-gate.sh` changes. Twenty minutes later I delivered PR #3501 a `RESULT: PASS` 31/31 gate of record and told it to arm `--auto` — and it had already gone `CONFLICTING`, because #3403 had moved the same file underneath it. Its certification is dead and it needs a second full gate. **That was my sequencing failure**: I unblocked #3403 while knowing #3501 was in flight on the same file, then spent 70 minutes of box2 certifying a tree a merge I had authorized was about to invalidate.

**The contending set:**

```
PR #3501  (#3400)  agent-gate.sh +234/-15   CONFLICTING now, gate DEAD, must rebase + re-gate
PR #3467  (#3393)  agent-gate.sh + 12 more  clean onto main; gate QUEUED on box2 at b2c5a02f1c14
PR #3538  (#3401)  agent-gate.sh + 2 more   clean onto main; still iterating on a wording Low
PR #3536  (#1699)  agent-gate.sh comments   clean onto main; gate not yet launched
```

I test-merged all four against `main`: **only #3501 conflicts today.** The other three are clean *right now* — and would not stay that way if they merged in an arbitrary order, because each merge moves the file under the others.

## The order

1. **#3467 (#3393)** — its gate is already queued on box2. That slot is burning; I am not wasting it.
2. **#3536 (#1699)** — comments-only in `agent-gate.sh`, so the cheapest to rebase if it slips.
3. **#3538 (#3401)** — still fixing a wording Low; it will be ready around when a slot frees.
4. **#3501 (#3400)** — last, because it must rebase regardless, so it should rebase **once** over the final state rather than three times over intermediate ones.

## What this asks of you

**Nothing except: do not merge out of turn, and tell me before you arm `--auto`.** If your PR is ready ahead of its slot, say so and I will re-order — the sequence is a scheduling tool, not a queue you are stuck in. What I want to avoid is two of you arming within the same few minutes and one of you discovering afterward that a 70-minute certification is void.

**If you are rebased under someone else's merge**, do not silently re-run your own gate: tell me, and I will place the re-gate on the quietest box the way I placed #3501's and #3467's. Gate placement is mine to sequence, and the whole reason #3501's run happened on box2 at all was that box1 was at load 168.

**Concretely, for each of you:**

- **#3467** — proceed as you are. Your queued gate is valid; `b2c5a02f1c14` is clean onto `main` as of now. I bring the block when it lands. If a merge lands under you before your gate finishes, I will tell you rather than let you read a stale PASS.
- **#3536** — you are second. I owe you a gate placement and box2 now has **215 G free** (I removed the finished `gate-3501` worktree), so the cold-gate figure you measured is no longer the blocker; the slot is. I will launch it when one frees.
- **#3538** — you are third. Finish the wording Low and the `CQLITE_ALLOW_FILE_GROWTH unset` diagnostic; there is no rush to beat anyone to the file.

## The general lesson, which is mine

**A file that four PRs edit is a serialization point, and nobody was tracking it.** Each of you correctly reported your own state; none of you could see that three peers were editing the same file, and I had the whole picture and did not look at it until a certification died. Merge-order contention on a shared file is a coordination job, not a lane job — I will run the check every review from here rather than discovering it from a `CONFLICTING` badge.

pmcfadin:
## Gate of record: run #2 — `RESULT: PASS`

**Run #1 was discarded and re-run. It was not waived, and no component was excused.**

### The discarded run

`/tmp/cqlite-gates/3401/full-1788063000.txt` returned `RESULT: FAIL` at this same head
(`5bccfa7`, `tree-integrity: PASS`). 30 of 31 reached components passed; the sole red was:

```
smoke:             FAIL (115s)
collect2: fatal error: ld terminated with signal 7 [Bus error], core dumped
error: could not compile `cqlite-cli` (bin "cqlite") due to 1 previous error
```

The failure is in the `cargo build --package cqlite-cli --bin cqlite` step that *precedes*
the smoke script. **`smoke` never ran a table** — 0 occurrences of `Testing table` / `test
result: FAILED` / `✓` in its component log.

### Attribution: #3434 (open)

*"The gate cap counts the wrong resource: disk exhaustion produces a FAKE RED, not a queue."*
That issue's body carries this exact `signal 7 [Bus error]` signature.

Evidence, each measured rather than inferred:

- `git diff --name-only origin/main...HEAD -- test-data/ cqlite-core/ cqlite-cli/ Cargo.toml Cargo.lock`
  = **0 files**. Every compiled input and the smoke script itself are byte-identical to `main`,
  so this branch had nothing that could break a link step.
- The smoke command re-run standalone in this worktree: **51/51 enforced tables passed, EXIT=0**.
  An isolated `agent-gate.sh --only smoke` at the same head likewise returned `smoke: PASS (46s)`.
- **Arithmetic**: `/data` free went 150G → 64G across the passing run, so a full gate consumes
  **~86G**. Run #1 *started* with 64G free. It was doomed before it began.
- No ENOSPC at inspection time, no OOM and no device errors in `dmesg` — consistent with #3434's
  point that the concurrency cap counts gate slots rather than the resource that actually runs out.

Remediation was environmental only: `cargo clean` in this lane recovered 92.3 GiB. **No source
was patched to turn the red green.** `scripts/ci/classify-docs-only.sh` returns `full` (exit 1)
on this diff, so the docs-only path was never available — and since the diff's subject *is* the
gate, the red had to be cleared by a genuinely passing gate. It was.

### Gate of record

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.QiZ9VD
commit: 5bccfa7 branch: issue-3401-file-size-fail-no-diagnostic-log dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-3401/test-data/schemas (checkout-relative)
tree-start: 5bccfa7775d6 dirty: no digest: 0efc6033e0ca
tree-end: 5bccfa7775d6 dirty: no digest: 0efc6033e0ca
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (7s)
clippy:            PASS (254s)
roborev-lints:     PASS (40s)
core-tests:        PASS (520s)
tombstones-scan:   PASS (10s)
scan-offload-guard: PASS (112s)
work-counters-guard: PASS (48s)
byte-budget-guard: PASS (13s)
arrow-parity-guard: PASS (10s)
memory-budget:     PASS (185s)
integration-tests: PASS (22s)
format-compat:     PASS (8s)
write-tests:       PASS (114s)
cli-tests:         PASS (86s)
flight-query-semantics-oracle: PASS (13s)
python-bindings:   PASS (299s)
node-bindings:     PASS (230s)
delivery-telemetry: PASS (0s)
oom-audit:         PASS (4s)
parity-report:     PASS (20s)
operator-metrics-doc: PASS (20s)
pub-surface:       PASS (0s)
tooling-tests:     PASS (787s)
minimal-build:     PASS (46s)
smoke:             PASS (41s)
logs: /tmp/agent-gate.QiZ9VD
summary-file: /tmp/cqlite-gates/3401/full2-1788063505.txt
RESULT: PASS
```

All 32 components PASS, no SKIPs. `tooling-tests: PASS (787s)` executed this PR's new
75-assertion self-test inside a gate, and `file-size: PASS` exercised the very `run_file_size`
code this PR rewrites.

### Final roborev confirmation

Job **144**, `RESULT: PASS`, `findings: NONE`, reviewed range
`9f38e77cb30a5166cc6ece8a5805ee3507a338f8..5bccfa7775d60fac6e87a8ff37956753ca22531e`
(`sha-assert: PASS`, `prompt-content: PASS (2/2)`, `vacuity-tier1/tier2: PASS`;
tokens input=402027 cached=325120 — genuine-review range, not the vacuous ~18.7k baseline).

Issue #3344, opened 2026-08-06
flow-meta: flow-closer's step 1 asks a subagent to own a gate across a turn boundary — the gate dies and leaves only the INCOMPLETE sentinel


pmcfadin:
Surfaced on #2827 / PR #3343 (2026-08-06). **Structural, not specific to that issue** — it will recur on every `flow-closer` run whose gate exceeds one subagent turn.

## The contradiction

`flow-closer`'s step 1 tells it to run the ONE full gate of record and, per #1855, to use `run_in_background` (or a long timeout) and **never idle-wait**, because a subagent blocking on a 12–25 minute gate is killed by the 600s stall watchdog and orphans the gate.

But a subagent's backgrounded child **does not survive its turn boundary**. So the two instructions cannot both be satisfied:

- **Idle-wait** through the gate ⇒ stall-watchdog kill (#1855), gate orphaned.
- **Background it and end the turn** ⇒ the gate process dies when the turn ends, and the summary file is left holding only the launch-time `RESULT: INCOMPLETE` EXIT-trap sentinel.

## What actually happened

`flow-closer` launched the gate correctly (summary-file redirect, `CQLITE_DATASETS_ROOT` from the fetch script's printed export, background, no idle-wait) and ended its turn — textbook per its brief. Result: the gate completed `fmt` (PASS) and `clippy` (PASS), then died at `[roborev-lints]`. No OOM (`dmesg` clean), disk at 73%, no `tree-integrity` mutation. `/tmp/gate-2827-FULL.txt` held only `RESULT: INCOMPLETE (gate did not finish)` with no surviving `agent-gate.sh` process.

Cost: one wasted ~10-minute partial gate run, plus the lead-side diagnosis to establish it was a process-lifetime death rather than a real failure.

**The dangerous version of this is the near miss, not the failure.** `RESULT: INCOMPLETE` is *textually* indistinguishable from a queued-or-running gate. An agent probing on the bare `RESULT:` token — which #3041 already warns about — would read a dead gate as "still running" and wait forever, or worse, a later reader could mistake a stale sentinel for a run in progress. The existing `PASS|FAIL` probe rule catches the accept-a-non-verdict case but **not** the distinguish-dead-from-running case.

## Proposed fix

[...]

Issue #2668, opened 2026-07-17
flow-closer executable-as-written: fix Agent-tool contradiction, script the #2456 pre-merge SHA assert, deadline the gate wait


pmcfadin:
Epic #2664. Audit finding: flow-closer.md is un-runnable as written for design-routed issues — frontmatter grants `tools: Read, Write, Edit, Bash, Glob, Grep` (no Agent tool) while step 2 orders "Spawn spec-auditor" and step 4 "respawn sstable-developer". Field workaround (lead spawns C, lead drives every closer transition) lives only in session memory. Also: the #2456 pre-merge headRefOid==certified-SHA assert (which blocked the #2467 stale merge) exists only as brief prose, not in the skill file or a script; the closer's merge step has no pending-CI semantics (superseded by --auto in the sibling issue).
Acceptance:
- Resolve the tool contradiction ONE way: either grant flow-closer the Agent tool, or rewrite steps 2/4 to return a structured NEEDS-SPAWN packet to the lead (codifying the field workaround). Doctrine and agent file agree.
- #2456 assert becomes a scripted pre-merge step (`scripts/flow/premerge-assert.sh <pr> <sha>`) called by the closer, not remembered; #2456 closes against this.
- Closer gate-wait gets a hard deadline + scripted fallback (if no harness re-invoke within N min of expected completion, grep the summary file) instead of optional "if you must" polling.
Relates: #2456, #2084.

pmcfadin:
🔒 Claimed by flow-lead (Patricks-MacBook-Pro) — wave 1 of epic #2664 Phase-1 hardening. Branch: issue-2668-* pushed to origin.

PR #3561, opened 2026-08-30
fix(#3473): lane gate execution — no ~10min ceiling, cgroup-detached gate launch, and a liveness heartbeat that tells reaped from running

### LINKED ISSUE #3473: Lane sessions cannot run a full agent-gate: background work is reaped at ~10 min, so the gate of record is unrunnable inside a lane
## Summary

A lane worker (a `/drive-issue` session in a tmux pane) **cannot run a full `scripts/agent-gate.sh` to completion.** Every attempt observed dies at **~10 minutes**, inside whichever component happens to be running, leaving only the launch sentinel (`RESULT: INCOMPLETE (gate did not finish)`) in its summary file.

A full gate takes **25–50 minutes** on these boxes (measured tonight: 13.5 min warm-cache best case, 50 min cold, with `tooling-tests` alone at 687–849 s). So the ~10-minute ceiling means **a lane cannot self-certify on any box, at any load.**

## Evidence, and why it is not resource starvation

Observed on `lane-3393` (box3), twice, both dying at ~10 min inside `core-tests`. The decisive observation is the lane's own:

> the passive `sleep`-loop waiter was also killed

**A `sleep` loop consumes no CPU, no memory and no I/O, so it cannot be starved.** That single fact eliminates, together:

- CPU contention / co-tenant gates (the box3 measurement reservation was live at the time and was suspected first),
- memory pressure (`OOM-KILLS-TOTAL` did not move; no `dmesg` victim),
- disk (`No space left on device` count was 0),
- the `#1825` gate slot cap (a queued gate prints `waiting for gate slot` and stays alive).

What remains is **background-task reaping inside the lane harness itself** — something in the agent session terminates long-running background work at roughly its maximum background window, regardless of what that work is.

**Control, on the same sha and the same box family:** the identical gate launched by the coordination lead over plain `ssh` + `nohup` ran to completion — `RESULT: PASS`, 31/31, `tooling-tests: PASS (849s)`. Same code, same box, same slot cap; only the *launching context* differed.

## Impact

This is currently worked around by the coordination lead running **every** gate of record for **every** lane, by hand, over `ssh`. Tonight that was **7 full gates across 4 lanes** (#3248 ×3, #1699 ×3, #3393 ×2 — several voided by head moves).

Consequences:

1. **A documented flow step is not executable as documented.** `flow-implement` → `flow-closer` has the closer run the full gate in its own context; on this fleet it cannot.
2. **Gate throughput is bounded by one human-in-the-loop actor.** Lanes idle for tens of minutes waiting for a verdict only the lead can produce, and every head move costs another lead-run gate.
3. **It hides itself.** The failure leaves `RESULT: INCOMPLETE`, which is *also* what a healthy just-launched or queued gate writes (#3041). So a killed lane gate and a running one are textually identical, and the correct completion probe (`grep -qE 'RESULT: (PASS|FAIL)'`) reports "not finished" for both.

## What is NOT the cause

Recorded so the next investigator does not re-derive it:

- **Not** the `#2640` concurrency pin — that is now genuinely in force (`max-concurrency=1` appears in every SUMMARY), and gates still completed only when lead-launched.
- **Not** box3-specific — the reservation was released and the behaviour was unchanged.
- **Not** disk (#2923) — verified 0 ENOSPC in the affected logs.
- **Not** OOM (#3393) — kill counts flat across the failures.

## Acceptance criteria

1. **AC1 — reproduce with an instrument, not a gate.** Launch a trivial `nohup`'d background job from inside a lane pane that only appends a timestamp every 30 s, and record the wall time at which it stops. Establishes the ceiling as a *number* rather than "~10 minutes", and costs no build.
2. **AC2 — identify the reaper.** Name the mechanism that terminates it (process-group teardown at turn end, a harness watchdog, a session-scoped cgroup, `SIGHUP` on pane detach, …), with the evidence that distinguishes it from the alternatives.
3. **AC3 — either a supported way for a lane to run a full gate, or doctrine that says it cannot.** A fix (e.g. a documented detach idiom that survives) **or** an explicit statement in `CLAUDE.md` + the flow-closer skill that on this fleet the gate of record is run outside the lane session, so the closer's documented behaviour matches reality.
4. **AC4 — the killed-vs-running ambiguity is removed.** A lane must be able to tell "my gate was reaped" from "my gate is still going" without a human reading `ps` on the box. A liveness heartbeat written by the gate itself would do it; the `INCOMPLETE` sentinel alone provably cannot.

## Why this is filed now

The claim *"lanes cannot run a full gate, so I run them"* has been load-bearing all night — it is why the lead is the gate-runner for the whole fleet, and it has been repeated to four lanes as established fact. **It was documented in no issue.** An undocumented load-bearing claim is indistinguishable from an unexamined assumption, and this one reshapes how the delivery pipeline actually operates.

Refs #3041 (the `INCOMPLETE` sentinel), #2640 (concurrency pin), #2923 (disk-pressure guard), #3393 (lane OOM kills — a *different* lane-kills-its-box failure, not this one).

pmcfadin:
Closes #3473.

## The premise was false, and the real mechanism is the cgroup

#3473 reported a hard **~10 minute ceiling** on background work launched from a lane pane, "regardless of what that work is", and concluded a lane cannot self-certify. Measured with an instrument rather than a gate: **there is no time-based ceiling.**

Six tickers launched from inside a lane, each appending a timestamped line every 10 s, each trapping every catchable signal, each self-terminating at 2400 s with an explicit `COMPLETE … never reaped` line so the artifact itself separates "reaped" from "still going":

| probe | launched by | detach idiom | outcome |
|---|---|---|---|
| A | lane top-level | plain `nohup … &` | past **1971 s**, 0 signals |
| B | lane top-level | `setsid` (own sid+pgid) | past 1971 s, 0 signals |
| C | lane top-level | `setsid`, `cwd=/`, argv renamed | past 1971 s, 0 signals |
| D | lane top-level | harness `run_in_background` | past 1951 s, 0 signals |
| E | **subagent** | plain `nohup … &` | full 2400 s, 0 signals |
| F | **subagent** | `setsid` | full 2400 s, 0 signals |

Every variant ran the **full 2400 s** — four times the reported figure — and self-terminated with its own `COMPLETE … never reaped` line, with no signal of any kind.

**The 600 s stall-watchdog hypothesis is untested, not falsified**, and an earlier revision of this PR body and the docs claimed otherwise. E/F's subagent was given a deliberately silent 700 s foreground block to induce a stall; **no stall occurred** — this harness version *backgrounds* an over-timeout foreground call and returns control immediately, and the blocker completed unmolested (exit 0, full 700 s) while the subagent kept working. The experiment sidestepped the condition it was meant to create. That harness behaviour is itself relevant to the issue's *"the passive `sleep`-loop waiter was also killed"*: a long silent call is backgrounded, not killed, on this version.

**AC2 — the reaper is systemd cgroup teardown.** Every process in a lane shares one scope per tmux pane, carrying `KillMode=control-group`, `SendSIGKILL=yes`, `TimeoutStopUSec=90s`. That signals **every task in the cgroup**, and cgroup membership is inherited across `fork` — it cannot be shed by `nohup`, `setsid`, closing fds or reparenting to init. Demonstrated both directions on a cgroup created and destroyed by the test:

| | cgroup | detach idiom | after the cgroup stop |
|---|---|---|---|
| IN | the cage's | `setsid`+`nohup`+closed fds, own sid **and** pgid, ppid 1 | **dead**, last tick 1 s before |
| OUT | its own scope | none | **alive**, still ticking 11 s later |

Run twice: the first trial recorded `SIGNAL TERM` at exactly the stop instant, the second recorded **nothing at all**. The second **reproduces the field symptom precisely — a gate killed this way leaves no trace**, which is why the summary keeps its launch sentinel and nothing says why.

This accounts for all four of the issue's eliminations at once: a cgroup kill **signals** work rather than **starving** it, so a passive `sleep` dying is the expected shape and CPU/memory/disk/slot-cap are correctly ruled out. It also explains the control the issue recorded — an `ssh` login gets its own `session-N.scope`. **The variable was never the work, the box or the load. It was the cgroup.**

**A subagent gets its own pane scope** (E/F landed in a different `tmux-spawn-*.scope` from A–D). The tempting next step is "so `flow-closer` loses its gate when its context ends" — **I wrote that, then tested it, and it is false.** A subagent was killed outright: its `claude` process disappeared, the scope stayed `active`, and both its tickers **kept running, orphaned**. systemd releases a scope only when its **last** process exits, so a long gate *holds its own scope open* and outlives the agent that launched it. An agent finishing is not a teardown. The falsifying measurement is recorded beside the claim in every doc it touched.

### AC2 is therefore a PARTIAL, said plainly

**Established:** (1) no time-based ceiling; (2) subagent-launched work survives its subagent's whole lifetime — though this does *not* clear the stall watchdog, see above; (3) pane scopes carry `KillMode=control-group`/`SendSIGKILL=yes`; (4) stopping such a cgroup kills fully-detached work and spares own-cgroup work, in one trial leaving no signal record at all; (5) subagents get their own scope; (6) an agent's process exiting does **not** tear down its scope.

**Not established:** that any of this killed the gates on lane-3393. Those deaths were not observed, the journal retains no `tmux-spawn` scope-stop records, and (6) rules out the most natural trigger. So AC2 is *"a mechanism that demonstrably produces this symptom, with alternatives ruled out"* — **not** *"this was the reaper"*. Calling it identified would dress an inference as a measurement, which is the failure mode this issue was filed against.

**Unexplained:** the ~10 minute correlation — nothing measured here is time-based. Untested candidates: a pane/supervisor recycle on a cadence; the 600 s stall watchdog under a *genuine* stall (which I could not induce); and a harness-level kill of a *tracked* background task (a different object from the process-level probes used here, all of which survived). Happy to take direction on whether to pursue those in a follow-up.

`flow-closer.md` instructed `run_in_background` then *end your turn*, and `worker.md` claimed the watchdog "orphans the gate" — that one is backwards in the other direction. Both corrected, on the narrower justification below.

## What ships

- **AC3 — `scripts/flow/gate-detached.sh`**: runs the gate as a transient systemd unit under `app.slice`, in a cgroup of its own. The justification is the narrow one that survived the correction above: not "your completion kills the gate" (it doesn't) but "a detached gate is independent of *every* pane and session teardown — a supervisor recycle, `kill-pane`, logout — a class the closer cannot see coming and cannot distinguish from a slow gate." **"Lanes cannot run a full gate" is retracted** — this PR's own `--lite` certification was launched through it, from the lane. It forwards the caller's environment **except three build-flag variables it deliberately drops** (a transient unit inherits **none** of the environment; silently dropping all of it would hand the gate a different `PATH`, no `CQLITE_DATASETS_ROOT`, no sccache — failures that would look like a real red), reports by name anything it does not forward, and **refuses with exit 69** where it cannot deliver a separate cgroup rather than falling back to a session-scoped launch the caller would believe was protected.

  **The one deliberate exclusion, and why it is not a capability limit.** `RUSTFLAGS`,
  `CARGO_ENCODED_RUSTFLAGS` and `RUSTDOCFLAGS` are **dropped, not forwarded**, and the drop names
  itself in the launcher's `SKIPPED` list. `systemd-run --scope` inherits the caller's shell, and a
  non-empty `RUSTFLAGS` **suppresses cargo's managed `target.rustflags` block entirely**, so a lane
  with that variable exported gets a doubled `-D warnings` and a gate FAIL that is **environmental,
  not a source defect** — the mechanism behind a fleet-wide P0 that was raised, believed on three
  non-independent reproductions, and retracted the same day. Forwarding the caller's environment is
  right for `PATH` and `CQLITE_DATASETS_ROOT`, whose absence produces an obvious red; it is wrong for
  build flags, whose presence produces a *plausible* red. Verified on the generated wrapper:
  `RUSTFLAGS` 2 -> 0, `CARGO_ENCODED_RUSTFLAGS` 1 -> 0, **50 other exports still present**
  (`4b.170`), and the drop reaches the banner (`4b.171`).

  **Its limit, stated because the code states it and the body previously did not:** this cleans the
  **supported** launch path only. A lane running `bash scripts/agent-gate.sh` directly in a
  contaminated shell still inherits all three, and the gate's own disclosure of that state is
  `mold=overridden`, whose arm requires mold to be **installed and wired** — so on a box where mold
  is absent (measured: this one) an exported `RUSTFLAGS` prints `mold=absent` and is **invisible in
  every pasted SUMMARY**. Proposed as a follow-up on the issue thread, deliberately not taken here:
  AC3 is the supported path.
- **AC4 — `scripts/lib/gate-heartbeat.sh` + `scripts/gate-liveness.sh`**: `RESULT: INCOMPLETE` is written once, at launch, so it is the artifact of three states at once and nothing about it decays. The gate now beats `<summary-file>.heartbeat` every 20 s while its process lives (the startup sentinel names the path, since that is what a lane finds when its gate was reaped), and the reader reports `COMPLETE`(0) / `RUNNING`(2) / `STALLED`(3) / `UNKNOWN`(4). Every SUMMARY carries a `heartbeat:` line so a pasted block shows the mechanism ran.
  Every positive verdict is an **affirmative measurement**: a **missing** beat is `UNKNOWN`, never `STALLED` (absence is not evidence of anything, and an older gate simply has no beat) — and `STALLED` itself means only *no liveness published*, never *the process is dead*; the beater re-verifies the gate pid — pinning `/proc` start time so a recycled pid reads dead — before **every** beat; the staleness window is read from the beat's own `interval:` line so the reader holds no duplicate constant; and there is deliberately **no env var** to widen the window or disable the beat.
- Doctrine: `CLAUDE.md`, `.claude/agents/flow-closer.md`, `.claude/commands/worker.md`, `docs/development/gate-ops.md` (where the log-mtime probe was called "the authoritative aliveness probe" — it is one-directional, since a queued gate writes nothing and a silent component leaves the mtime flat, so it is now a named fallback), and a full record in `docs/development/lane-gate-execution.md`.

## Review rounds: 7 findings across 2 rounds, all fixed

roborev job **155** (3) and job **157** (4). Both were genuine reviews by token accounting (155: 265k input / 205k cached; 157: 593k / 502k — against a vacuous baseline of ~18.7k / 0).

| job | sev | finding |
|---|---|---|
| 155 | Med | fields read by **re-opening the artifact path per field** (7 sites). These paths are replaced *atomically* by peers, so one run's `run-id` could pair with another's `RESULT` or a fresher `beat-epoch` — the cross-run confusion #2874's reader contract exists to prevent, reintroduced through the I/O pattern rather than the logic. Now one `open()` per artifact, asserted structurally. |
| 155 | Med | `--run-id` supplied but the summary had **no `run-id:` line** → validation skipped entirely and the verdict attributed to the requested run. A permissive branch keyed on the *absence* of the bad signal. Now unconditional. |
| 155 | Low | verdict matched by **prefix glob**, so `RESULT: PASSENGER` read as `COMPLETE` — a *spelling* test wearing a closed grammar's clothes. Now reduced to a verdict **token** and matched exactly. |
| 157 | Med | a terminal `RESULT:` in a **truncated** block read as `COMPLETE`. `emit_summary` verifies its own end marker precisely because that single-`>` write can be cut short; such an artifact is permanently unfinished. Now the block's framing is required — asymmetrically, on the COMPLETE path only, since INCOMPLETE already falls through to the beat. |
| 157 | Med | a **stale beat alone** reported `REAPED`. First fixed by corroborating against `/proc`; that fix was itself found unsound twice more, and the claim was ultimately **descoped** — see below. |
| 157 | Med | predictable default `/tmp` artifact names plus `>` truncation — symlink pre-creation by another local user. Defaults now live in an `mktemp -d` 0700 directory; a caller-supplied path is still honoured exactly. |
| 157 | Med | GNU-only in-place `sed` and bare `timeout` in a **gate-wired** test. macOS is a first-class gate host, so this fails the *gate* there, not merely a test. Both replaced portably and pinned by a structural guard over both suites. |

Each fix has a case that RED-verifies against the pre-fix code, not merely a green after: the old reader really did report `COMPLETE (… — PASSENGER)` and really did attribute an id-less `PASS` to a named run.

## The death claim was DESCOPED, not patched a fifth time

The reader originally reported **`REAPED`** — a positive claim that the gate process was dead. Rounds 2, 3 and 4 each found another way that was unsound, and round 4's portability finding was caused by round 3's own fix:

1. a stale beat alone does not imply death — the beater is supervised only at component boundaries, and components run for minutes, so a beater can die under a live gate;
2. corroborating against the **reader's** `/proc` proves nothing about the **gate's** host — across shared storage a live remote gate reads as dead;
3. matching **hostnames** do not prove machine identity — two boxes can share one, so a differing `boot-id` was misread as a reboot;
4. and there is no `/proc` boot id on the macOS/BSD gate hosts, so the death cases failed there **deterministically**.

Each fix was right about the case in front of it and the list did not close, because proving a process dead means **proving a negative about a machine you may not be on**. So the claim is gone. `STALLED` says only what two local files can establish — *this run has published no liveness for N seconds* — needs no pid, `/proc`, host or boot identity, and is therefore correct on every host.

**The descope deleted code**: 74 lines of inference, and with it the known-dead-pid derivation that had already produced one flaky-test defect. Two structural asserts now fail if `/proc`, a pid probe, machine identity, or the `REAPED` verdict return.

Nothing actionable is lost. The lane's question is "should I keep waiting?", and the replacement rule needs no process inspection: the gate relaunches its beater at every component boundary, so a live gate whose beater alone died recovers to `RUNNING` within one component. The verdict text states this inline, so a reader acting on it needs no doctrine in context. What is genuinely weaker: it will not say "definitely dead, relaunch now" — it costs one re-read, and at worst a long component's wait. That is cheaper than a verdict that can be confidently wrong in either direction, given a false death sends a lane to re-run a gate that was about to PASS.

Doctrine updated in all five places that referenced the old verdict (`CLAUDE.md`, `gate-ops.md`, `flow-closer.md`, `worker.md`, the record doc), including *why* it was dropped, so it does not get reintroduced.

## The shape behind almost all of it, stated once

Eleven defects landed in this change — four self-caught, seven from review. **Seven are one shape: reproducing an anti-pattern that CLAUDE.md documents, inside the file whose own header cites that rule.** Permissive-on-absence twice; prefix-matching a token where the doctrine says match the token exactly; `REAPED` derived from the absence of freshness in a script whose header says every positive verdict must be an affirmative measurement; a portability guard whose needle matched its own source line (the doctrine warns to split the needle); a "known-dead pid" built in a command substitution so the parent could not reap it, which would have turned every death assertion into a silent `UNKNOWN`.

Worth recording rather than presenting as five unrelated slips: stating a rule produces a feeling of having handled it, and fluency with the doctrine actively hid these — I cited the issue numbers correctly while shipping the violation. The countermeasure that worked, and that found the flaky-pid bug review had not reached: after writing a rule into a comment, grep your own diff for the shape it forbids, before review.

## Four defects the tests caught in my own work

1. **The beater outlived its gate by up to a full interval** — bash does not run a trap handler while a foreground child is running, so `sleep $INTERVAL` deferred `SIGTERM`. That is this issue's own defect one level down (a beater reporting a dead gate as alive). Fixed with `sleep &` + `wait`.
2. **The composed EXIT trap called a function defined ~7400 lines later**, so every early-exit path printed `_gate_release_slot: command not found`. This failed an **unrelated** case in `test_agent_gate_summary.sh` — which reads any `command not found` as a missing tool under its minimal PATH — with a misleading cause.
3. **The heartbeat would have made the gate FAIL ITSELF.** Tree-integrity (#2926) hashes every untracked non-ignored path; the heartbeat is a file the gate writes into the checkout every 20 s whenever the summary path is in-repo. Without a carve-out the gate creates it after the start capture and reports `tree-mutated-midrun` naming its own heartbeat. **A short `--only` run passed by RACE** (file-size completes in ~0 s, so the first beat can land after the last boundary check); on a 30–50 min full gate the beat always precedes the first boundary, so this would have been a **deterministic failure of the gate of record**. Carved out in `_tree_excluded` beside the existing `.integrity-fail.*` sibling — which is what covers a caller-**pinned** in-repo path, since no `.gitignore` rule can predict it — plus `.gitignore` entries for the three default paths. Section 14 pins it with **two controls** proving the exclusion is not over-broad.
4. A runtime heartbeat artifact was committed by mistake and removed; that is how (3) was found, via `classify-docs-only.sh`.

## The mechanism proved itself in the wild, unplanned

While developing this PR I had to stop a running `--lite` gate twice to make corrections. A poll loop was waiting on it. Before this change that loop would have sat on `RESULT: INCOMPLETE` indefinitely — the state is indistinguishable from a slow gate, which is the whole complaint in #3473. Instead it answered and stopped.

**This transcript is from a PRE-DESCOPE build and is left verbatim rather than edited**, because it is a real observation and rewriting it would be fabricating one. It shows `REAPED`, the verdict this PR later **removed** (see the descope section above). The same run today prints `STALLED` in that position, and `STALLED` claims only that no liveness was published — not that the process died:

```
REAPED
run-id: /tmp/agent-gate.7tZGcA
MODE: lite (FAST ITERATION — NOT the gate of record; ...)
heartbeat: /tmp/lite-3473.txt.heartbeat (interval 20s) — read it with: bash scripts/gate-liveness.sh ...
```

Same `INCOMPLETE` text as always; the difference is that liveness is now answerable beside it.

## Round 48: class audit (lead ruling)

After round 47's three Mediums landed in one new file and one class, the lead ruled that **rounds sample
a class, they do not enumerate one** — so round 48 audited `gate-detached.sh` line by line against a
closed five-item list, fixed everything in one series, and RED-verified.

| Class | Finding | Disposition |
|---|---|---|
| **4** two-valued predicates | `_foreign_reservation` answered **`free`** when absence was *unverifiable* (unsearchable parent) — the pre-check every other guard rests on | **fixed**: `free` now requires an affirmative measurement, else `unknown`, which every caller refuses |
| **1** unquoted expansion | `set -- $_state` **and** `set -- $rest`, in **both** `gate-detached.sh` and `gate-heartbeat.sh` — **4 instances** | fixed: 2 → parameter expansion; 2 keep the deliberate split (`/proc/<pid>/stat` field 20) under `set -f`, prior `-f` saved and restored |
| **1** the *generated* wrapper | shellcheck run on the **captured artifact** rather than the source | **0 findings**; `%q` encodes a space+glob value as `a\ b\*c` |
| **2** command qualification | boundary: silent failure with a security consequence, or output feeding a security decision | already fail-closed (`stat`→refuse, `readlink`→`unknown`, `systemctl`→live, `flock`→refuse, `ln -s`→refuse); the wrapper was the real gap |
| **3** shared-dir fixed names | `PRIVDIR`, heartbeat probes, launch snapshot all `mktemp`; global lock now in a verified 0700 per-user dir | clean. Residual: `$SUMMARY.launch-lock` names are caller-chosen — a planted lock fails closed (DoS, never false pass) |
| **5** symlink/TOCTOU | `ln -s` is an atomic create, so a lost check→create race fails closed; the global lock removes the launcher-vs-launcher window | one residual, unchanged: log truncation is check-then-create, bash has no `O_NOFOLLOW` |
| dead code | `_has_line` + `_START_RE`: defined, never called, while five sites duplicate that regex | removed — dead code shaped like a check is worse than none |

**The method result is the part worth reading.** shellcheck 0.11.0 was brought in so the audit would not
rest on the author's reading of his own file. It reported **1 of the class's 4 instances**. Two misses
were not tool limitations: those sites carried an explicit `# shellcheck disable=SC2086` from an earlier
round, so **the enumerator was silent because it had been told to be — at exactly the sites that needed
it**. A blanket disable suppressed the *globbing* half nobody meant to waive along with the *splitting*
half that was deliberate. What closed the gap was the change's own structural test going red on its own
fix (`4b.162`). Accordingly `0 findings` here is a **class-1** statement; classes 2–5 were enumerated by
hand, 21 file predicates individually.

RED-verified against `fb97362eb`: `4b.161` (the false `free`) and `4b.162` (unguarded splits, 2 per file)
each fail pre-fix and pass after. `4b.160` is the control and passes on **both**, so `4b.161` cannot be
satisfied by a function that answers `unknown` for everything.

## Intent audit (C) and scope split

`spec-auditor` on `60bbfdcd7`: **AC1 satisfied, AC2 partial (judged justified), AC3 satisfied, AC4
satisfied — none blocking.** It caught one non-code MUST-FIX: this body described the reader as emitting
`REAPED`(3) and quoted a live `REAPED` transcript, contradicting shipped code and this body's own descope
section. Fixed. The transcript is left **verbatim and labelled pre-descope** rather than edited, because
it is a real observation and rewriting it to read `STALLED` would be fabricating one.

AC2's untested candidates are split out as **#3664** (three untested death candidates + the unexplained
~10 minute correlation), and **#3344** — whose diagnosis this change falsifies, and whose proposed fix 2
is the `REAPED` design descoped here as unsound — carries a correcting comment so nobody implements it.

## Round 49–50, and the stopping rule this PR ships under

Round 49 (job 272) returned two Mediums, both in `gate-detached.sh`, and **both were sites that bypassed
a safe primitive already present and already correct a few lines away**:

| # | Defect | The primitive it declined to use |
|---|---|---|
| F1 | the unit-death check used `systemctl is-active --quiet`, which exits nonzero for every **transitional** state *and* every **query failure** — so a healthy-but-unsettled gate, or one we could not ask about, read as dead and the launcher **stopped a live gate as unmonitorable** | `_unit_is_live`, which already had the closed grammar (`inactive\|failed` are the only affirmative terminal answers) |
| F2 | the settled re-derivation read the nonce and the run-id with two separate greps against a **live** artifact, so a concurrent gate can pair **our** nonce with a **peer's** run-id and we advertise the wrong run | `_snap_pair`, built for exactly this in job 190 and **called twice a few lines above** |

**Corrected audit rule (lead's, adopted): AUDIT BY CALL SITE, NOT BY PRIMITIVE.** The round-48 five-item
list enumerated predicates and paths — the *definitions* of the safe primitives — and never their call
sites. An audit organised around definitions is structurally blind to omission at the point of use, and
**the helper's existence makes the omission feel handled**. A related consequence, seen twice this cycle:
a structural test anchored on implementation *spelling* both fails on a correct refactor and passes on the
defect it was written to catch — `4b.124` was nominally guarding F2's own branch while accepting the unsafe
two-grep form, and `4b.35b`/`4b.35c` red-flagged a correct fix. Both are now anchored on properties, and
`4b.163` carries a non-vacuity check so a file that reads the run-id nowhere cannot pass.

**The stopping rule this PR ships under — a property, not a count:**

> Ship when roborev returns **clean**, OR when the only remaining findings are in the file+class whose
> ownership the lead has taken. Round 50 runs on the fixed head, because twelve lines is still twelve
> unreviewed lines in a script handling secrets, locks and unit lifecycle, and doctrine requires roborev
> clean **at merge**, not at an earlier head. If round 50 is clean → gate of record → `premerge-assert` →
> arm. If it returns more findings in that same file+class → ship, record them, and do **not** open
> round 51.

A count was the wrong shape for that commitment: "last round" was promised and broken twice earlier in
this change, because the promise was mis-specified rather than weakly held.

## Gate-of-record launch conditions (lead disk order v5.2)

Stated before the launch rather than after it, so the bar cannot be back-fitted to whatever the box
happened to have. The 40G floor of order v5 is a property of **building**; it does not apply to the
filesystem that merely stores the verdict, so the two filesystems get different bars.

| filesystem | device / mount | bar | derivation |
|---|---|---|---|
| target (builds) | `/dev/nvme1n1` `/data` | **60G** | 2 x **own measured** growth: 30907 MiB (30.18 GiB) in 5.4 min for `legacy-heuristics --all-targets` |
| summary (verdict) | `/dev/root` `/` | **5G** | a `LOG_DIR` measured at 3.6M for a full run; this filesystem stores a verdict, it does not build |

**Why not the 90G cold figure:** that figure is for an **absent** `target/`, where there is no warm
growth to double. This lane's `target/` is **present** (45G; 43G in `target/debug`) and this diff
contains **0 rust files**, so nothing in the change itself invalidates the graph. It is not simply warm
either: **3924 files / 30.2 GiB were deleted from it** — precisely the matrix-lane footprint measured
above — so that subgraph will rebuild, and twice that measurement is the bar.

**Why no cheap probe may lower it:** each feature-matrix lane builds at a **different feature set, i.e.
a different subgraph**, so a probe at the default set would report ~0 growth while measuring only the
cache. "Warm at `cli-helpers`" is not evidence about `parquet`, `delta-scan` or `--all-features`.

**Why the 5G bar still gets a tripwire despite 113G of headroom:** a `/`-side ENOSPC leaves
`RESULT: INCOMPLETE`, which per #3041 is indistinguishable from queued, running and dead — a gate that
**PASSED** presents as one that vanished. Both tripwires are armed with independent minima.

**60G is a FLOOR, not a projection, and it is re-derived at launch rather than carried forward.** It
covers the one subgraph measured to be invalid (the deleted matrix-lane footprint). Any `cqlite-core`
churn that lands on `main` between now and the launch invalidates more, and this branch merges `main`
once immediately before gating — so the number is checked against the churn at that moment, not
assumed to have survived. Measured at the time of writing: `main` moved one commit (#3647) touching
6 compiled inputs, **none of them in `cqlite-core`** — 3 in `bindings/python`, 3 in `cqlite-cli`
tests — so that particular advance costs the python and cli-test subgraphs and leaves the expensive
one warm.

### What the gate of record actually measured — the bar was OVER-derived, and here is the number

Three full gates have now run at this branch, so the projection above can be replaced with measurements:

| run | head | verdict | own-directory growth |
|---|---|---|---|
| #1 | `90d82a114` | **PASS 37/37**, 0 FAIL, 0 SKIP | **+58.8 GiB** (from a `target/` gutted by 30 GiB earlier) |
| #2 | `63291a760` | FAIL — `tooling-tests`, the **#3621** flake | **+18.69 GiB** (only the side-lane set was missing) |
| #3 | `63291a760` | **PASS 37/37**, 0 FAIL, 0 SKIP | **+0.00 GiB** (fully warm target) |
| #4 | `5052a170b` | **PASS 37/37**, 0 FAIL, 0 SKIP | **+0.44 GiB** (after a merge touching `agent-gate.sh`) |
| #5 | `e2a8aa330` | **PASS 37/37**, 0 FAIL, 0 SKIP | **+0.00 GiB** (warm quadrant; `target/` 104G before and after) |

**So the 60G bar was correct for run #1 and far too conservative for the others, and the reason is the
only part worth carrying forward: a gate's disk cost is determined by WHAT IS MISSING FROM `target/`, not
by the gate.** Run #3 re-ran all 37 components for +0.00 GiB. A bar derived from one cold measurement
therefore over-reserves on every warm re-run — which matters because a re-gate after a rebase is the
common case, not the rare one.

**#3621, for the record:** run #2's single failure was `ws0-report guards: 1 of 123` on
*"the driver must print its selection as PARTIAL/FULL up front"*, whose subject
(`scripts/perf/ws0-baseline.sh`) is **not in this diff**. Run #3 passed all 37 at the **same commit and
the same tree digest**, and the script passes standalone on clean `origin/main` and in this worktree —
**1 red against 3 green at one identical tree.** No source was patched and no waiver was taken.

## roborev job 319 — four findings, all real, all fixed

Job 319 (`gpt-5.6-sol`, input 1,696,064 / cached 1,561,856 — genuinely a review, against the ~18.7k/0
vacuous baseline) returned `findings: PRESENT (4)`. Each was verified against source before anything was
touched; **all four were real**, three Medium.

| # | Site | Defect | Evidence |
|---|---|---|---|
| F1 | `gate-detached.sh:810` | an uninspectable `/proc/<pid>/cmdline` was `continue`d, and because the function **defaults `found=1`** that returned *"affirmatively no gate"* — on which the caller **reclaims a live gate's summary path**. The line's own comment said "not evidence either way" while the control flow made it evidence. | behavioural, against the real pre-fix code: a live pid with unreadable argv gives **`rc=1` before, `rc=2` after** |
| F2 | `:1315`, `:1394` | `_unit_is_live` returns 0 for live **or unmeasurable**. Where 0 means *refuse* that is conservative; at these two sites 0 means **accept**, so an unreadable unit state re-admitted **one-beat-then-dead** — the case those gates exist to reject. | **`rc=0` before (accepts), `rc=1` after (refuses)** on an unmeasurable unit |
| F3 | `:1206`, `:1155` | under `set -uo pipefail` (line 82) on stock macOS `/bin/bash` 3.2, an empty `"${ARRAY[@]}"` is unbound and **aborts** — and a bare `gate-detached.sh`, the **advertised default full-gate invocation**, leaves `GATE_ARGS` empty. | a real detached zero-arg launch: the child argv has exactly **2 elements**, last `…/scripts/agent-gate.sh`, so the gate receives **zero** args and not one empty string |
| F4 | `:940` | `flock` called with no availability check; the only `command -v flock` guard sits in the reclamation path a normal launch never reaches. A flock-less host was told *"another launch holds the global launch lock"* — a peer that does not exist, with remedies that cannot work. | RED demonstrated: on a flock-less PATH the unchecked call takes the contention branch |

**F1 had a second instance job 319 did not name.** `-r` **succeeds** on an exiting or mid-exec process and
the file then reads **empty**, so "no argv at all" scored identically to "argv read, no gate in it". Not
theoretical: two daemons read that way while instrumenting this issue, and `/proc` sweeps taken during the
fix had ~6–10 pids vanish mid-loop. Inside this function the benign reading does not apply — a kernel
thread also has an empty `cmdline` but can never be in a `systemd-run --user` unit's cgroup.

**F2 was this PR's own earlier fix, and the third round in one function** (job 205 fixed its exit-code
form, job 241 its state-name form, job 318 added the acceptance-site use, job 319 found the polarity). A
claim that fails at N stages has the wrong *form*, so the state is now three-valued
(`_unit_state → live | terminal | unknown`) and **each caller names the polarity it needs**, rather than a
fourth call-site patch. `_unit_is_live` remains the refuse-polarity wrapper over the one implementation.

**Stated limit:** there is no bash < 4.4 on this host, so **F3's abort was not reproduced**. That half
rests on this repo's own precedent (`agent-gate.sh:8049`, job-2108 — identical construct and citation) and
documented bash semantics, not on a measurement taken here.

**The family, which is the transferable part.** F1, F2 and F4 are one shape: a **two-valued test standing
in for a three-valued fact**, resolving the unmeasured case to whichever answer the surrounding control
flow made convenient — "no gate" because `found` defaulted that way, "live" because that was conservative
*at a different call site*, "contended" because it was the only other branch. Each carried a
correct-sounding comment. The question that catches it is not *"is this the safe answer?"* but **"safe at
WHICH call site, and does 0 here mean refuse or accept?"**

**Seven tests in this change have now red-flagged a correct answer** because they were pinned to
implementation *spelling* rather than to a property (`4b.107/108/139/140/154/175` and `11u.1`). All are
re-pointed, behaviourally where a grep could not see the property, and `4b.189` now *proves* the
array-form equivalence `4b.154` was made to accept instead of assuming it. Two failures in the fix round
were **my own defects, not the subject's**: a stub variable named `st` was dynamically scoped into
`_unit_state`'s own `local st` (three false REDs), and `11q.2` was a phantom produced by editing a test
file **while its suite was executing it** — bash re-reads a script incrementally, and the tell is the same
case name appearing twice with a syntax error between.

## roborev rounds 320 and 321, and what the sequence itself showed

**Round 320 — one Medium, and it was in round 319's own fix.** `_unit_state` was made three-valued so
each caller could name its polarity — and both callers were then handed the **same partition**, one
articulated for the *refuse* side, where lumping every non-terminal state under `live` is conservative
and right. On the *accept* side it is not: `deactivating` is a unit **definitively shutting down**, so
accepting it is exactly the one-beat-then-dead case that gate exists to reject. **Naming the polarity is
not enough if the state space is only articulated for one of them.**

The state is now **four-valued** (`running | stopping | terminal | unknown`) and the two sides take
**opposite closures**: the refuse side stays closed on the *terminal* side (job 241's rule, behaviour
unchanged — checked against `4b.140`'s own expectations), and the accept side is an **allowlist** of
`active|activating|reloading|refreshing`. That inversion matters because this one function produced the
same defect at **four consecutive rounds in a different spelling each time** (205 exit-code, 241
state-name, 319 polarity, 320 partition). The first three each *named one more state*, which is why
there was always a next one; an allowlist refuses the next unnamed state by construction. `4b.192` is
the non-vacuity control, since an allowlist that accepts *nothing* satisfies every refusal case above it.

**Round 321 — two Mediums, both in code predating the fix round** (its hunks are at lines 115 and
698–889; these sites are 975+), so by origin these are new information rather than re-finds.

| # | Defect | Why it is a defect and not an invoker's choice |
|---|---|---|
| F1 | the **global** launch lock was selected through `${XDG_RUNTIME_DIR:-…}`, so two launchers with two individually valid 0700 runtime directories take **different locks** and the artifact-set check plus reservation stop being mutually exclusive | **the script itself advertised the route** — its own refusal told the operator to "export `XDG_RUNTIME_DIR` to a 0700 dir you own". Someone obeying the printed remedy silently opted out of the lock. Bypassable **by accident** ⇒ fix |
| F2 | reservations identify destinations **by path**, so two hard links to one inode reserve separately and then write the same file | the script already refuses a **symlinked** destination (job 169), so output-path aliasing was already in model; a hard link is the same threat in a different spelling |

F1's lock is now canonical `/run/user/$(id -u)` — justified rather than arbitrary, since `systemd-run
--user` plus lingering are already required above and both imply that directory exists. `XDG_RUNTIME_DIR`
is still **forwarded** to the gate; only the lock location is canonical. F2's `_link_count_state` is
three-valued and uses `find -links`, never `stat` (GNU-vs-BSD format flags this script refuses to depend
on elsewhere) and never `[ -z "$(find …)" ]`, which collapses "the scan failed" onto "no match" — the
shape this repo lints for as `1699-find-tristate`.

**F2 was within what the declared threat model already licensed *recording* rather than fixing** (it is
invoker-class). It was fixed because it is the same spelling of an axis already closed for symlinks and
cost ~35 lines — not because the model required it. The threat-model section now says so, and states
that a **third** spelling of that axis is where fixing stops being right: bind mounts, `/proc/self/fd`,
mount namespaces and check-then-open swaps are **not closed**, are invoker-class, and are to be recorded.

## roborev round 322 — the sharpest finding of the endgame, in this change's own AC4 mechanism

Two Mediums. **F1 is blocker-class and it is a hole in the mechanism this issue exists to build.**

`_hb_ensure` — the beater-recovery hook — was reachable **only** from `record_result`, i.e. at
**component** boundaries. No component runs until the #1825 slot is granted. So a beater that died
during the **queue wait** was never restarted: the reader reports `STALLED` about a gate that is merely
**queued**, and there is not even a component table yet from which to derive a wait bound — so a closer
following this change's *own* guidance relaunches a still-queued gate. **Two gates on one summary path,
reachable through the mechanism built to prevent it.** The queue is not a narrow window: 20+ minutes
under load, longer on a busy box, which makes it the phase where an unrecovered beater does most damage.

`_hb_ensure` is now called every ~20 s (100 ticks × 0.2 s) from the slot-wait loop, matching the beat
interval so recovery lands within about one beat. Safe to call often: it self-guards on `BASHPID` and
`HEARTBEAT_STATE` and respawns **only** on an affirmative `gone`, so it cannot stack beaters.

Fixing it also surfaced **the origin of the understated `~850 s` bound** corrected in four places
above: a comment asserting the longest component was `tooling-tests` at **687–849 s**. Gate of record
#4 measured **2073 s**. That sentence now says not to re-derive a bound from it.

**F2 is declared, not fixed.** `_unit_runs_a_gate` composes the cgroup-**v2** unified path, so on a v1
host `cgroup.procs` is absent, the helper answers `unknown`, and every caller refuses. **That degrades
safe** — a contended reservation is permanently *refused* rather than wrongly reclaimed, so two gates
still cannot land on one summary; the cost is that a stale reservation cannot self-heal there. Every box
on this fleet is measured `cgroup2fs` unified, so a mount-resolution guard would be **new executable
code on a path nothing here takes** — which is how the next review round earns its finding. The boundary
is stated at the code site and in the threat model.

**The test the finding asked for** (`12.2b`/`12.2c`) is behavioural and **deterministic**: the wait loop
is extracted and driven with `sleep` stubbed to *count ticks* and to create the ready-file at a chosen
tick, so no wall-clock is involved and the expected call count is exact — 2 calls over 250 ticks, which
is what it measures. A wall-clock form would be the flaky-perf-assert shape this repo lints for. `12.2c`
is the non-vacuity control: a 10-tick queue must produce **zero** calls, without which `12.2b` would
pass for any loop body at all.

## roborev round 323 — where the stopping rule fired, and the two defects it found in this sequence's own work

Three findings. **Two of the three were in code this fix sequence produced**, which is the majority
condition of the standing by-origin stopping rule — so **roborev stops on this branch and certification
is the full gate plus C.** Both Mediums were still fixed (a blocker is a blocker regardless of which
rule permits stopping) and the Low is filed as **#3769**.

**F1 — acceptance needed two affirmative legs.** Both RUNNING acceptance sites used
`_unit_is_affirmatively_live`, which reads only `ActiveState` — and this file documents at length that
an **orphaned non-gate process keeps a unit `active` forever**. That is the entire reason
`_unit_runs_a_gate` exists. So a gate that published one beat, died, and left a child was **accepted as
monitorable** while no terminal verdict would ever arrive: the weaker predicate was used at the
acceptance sites while the stronger one sat in the same file for exactly this distinction.

It is safe here though deliberately *not* used at the post-launch monitorability probe, because that
probe can run before the gate has `exec`'d, where "no `agent-gate.sh` in the cgroup" is the **normal**
startup state. These sites are reached only when the reader answered RUNNING — i.e. it saw a **fresh
beat** for this run-id — so the gate provably started, and an absent gate process now means *gone*
rather than *not yet arrived*.

The veto is keyed on the **affirmative absence** (rc 1), never on "not rc 0": rc 2 is unmeasurable, and
a cgroup-v1 host answers 2 **always** (job 322's declared precondition), so demanding rc 0 would red
every launch on such a host — the guard agents learn to waive. Verified matrix: gate-present ACCEPT,
orphan **refuse**, unmeasurable ACCEPT, and `inactive`/`failed`/`deactivating` refuse even with a gate
present.

**F2 — the reader still printed the figure four documents had been corrected for.** The emitted
`STALLED` guidance said "the longest component is ~850s … treat the gate as gone and relaunch it". The
`~850s` bound had been corrected in **four documentation sites** and the one an operator actually reads
was never grepped. *Correcting a claim at the sites you happen to notice is not correcting it* — the
disclosure path needs its own census, which has now been run repo-wide.

**And the first version of that fix reproduced the defect it was fixing**, caught by its own new test
(`4b.202`): it explained the history *inside* the operator-facing string, so the output still printed
"~850s" — as a warning rather than as advice, which anyone skimming for a number reads as the number. A
diagnostic must not carry the value it warns against. The history now lives in the code comment.

**A test was deleted rather than re-pointed a third time.** `4b.175` asserted "both acceptance sites
gate on *&lt;predicate-name&gt;*" and broke on a **correct** change twice — at job 320 and again at job
323. A claim that needs relocating at every correct refactor is pinned to the wrong thing. Its
surviving content is `4b.201` (counts the two-leg helper; strictly stronger), and `4b.174` still forbids
a bare `0|2)` arm, so the "accepts RUNNING with no unit check at all" regression stays covered from the
other direction.

## Why roborev ran five rounds, and what the census says

`319 → 4`, `320 → 1`, `321 → 2`, `322 → 2`: **nine findings over four rounds, of which exactly one
(320) was a re-find** of a fix from this sequence. So the standing by-origin stopping rule — stop when a
majority of a round's findings live in code the fix just introduced — **never fired**, because the
reviewer was surfacing genuinely new latent issues rather than re-litigating the fixes. Reachability did
decline across the sequence (live defects → invoker-class → a host class this fleet does not have), but
`322`'s F1 was blocker-class, so the marginal value was not zero at any point.

## Three doctrine defects found by auditing this change's own claims

Two were mine and one was pre-existing; all three would have caused **two gates on one summary path**,
which is the outcome this issue exists to prevent.

1. **`flow-closer.md` stated the `INCOMPLETE` doctrine and its negation in one file.** It said "`RESULT:
   INCOMPLETE` means the run never finished, so there is no verdict to read" while, 50 lines earlier,
   correctly calling `INCOMPLETE` "a liveness placeholder, not a verdict". An internal contradiction needs
   no external oracle. The false half is the dangerous one: the sentinel is written **at launch**, before
   the slot is granted, so on a re-invoke it usually means *still running* or *queued* — and a closer
   reading "the run is over" relaunches. Pre-existing; found by the C audit; fixed, and now points at
   `gate-liveness.sh`.
2. **The `STALLED` wait bound was understated 2.4× in four places — and it was mine.** All four said to
   treat a `STALLED` run as gone "after a component's worth of time (~850 s at the longest)". This
   branch's own gate of record #4 measured **`tooling-tests` at 2073 s**. Under-waiting relaunches a
   *live* gate. The figure decayed structurally, because `tooling-tests` runs ~69 nested scripts and
   **this very change added cases to two of them** — so the fix is not a better constant, which would
   decay again in four places: every site now says to **derive** the bound from the component table in
   the reader's own SUMMARY block.
3. **Flagged, not fixed:** `flow-closer.md`'s **45-minute hard gate deadline**, against gate #4's ~50–55
   min wall clock (37 components summing to 6303 s). A healthy passing gate exceeds it, so a closer
   obeying it abandons a gate about to PASS. Pre-existing from #2668 and it fails **closed**, so it is
   reported to the lead rather than edited here.

**And one verified non-defect, recorded because over-correction is the opposite failure.**
`flow-closer.md:45` says an idle-waiting subagent is killed by the 600s stall watchdog, while this
change's own finding calls that watchdog "untested, NOT cleared". Not a contradiction: the induction
attempt used an over-timeout **foreground** call that the harness *backgrounded*, which says nothing
about an idle-waiting subagent. Left alone.

**Test defects of my own, from these rounds** — worth listing because two are the same shape:
`4b.187` and `4b.194` each grepped for a string their own explanatory text contained, so both failed on
the *fixed* tree; a stub variable named `st` was dynamically scoped into `_unit_state`'s own `local st`,
producing three false REDs whose tell was that **every row was identical**; and `4b.157` was a **vacuous
pass reported green** — it set `XDG_RUNTIME_DIR` and asserted no refusal, which after F1 passes because
the canonical directory happens to be fine. `4b.156` is the sharpest: it induced its failure by setting
`XDG_RUNTIME_DIR`, which worked *only because of the defect F1 fixed* — the test hook and the
vulnerability were the same fact (#3544's recorded shape), so it now **substitutes the artifact**,
verifying the substitution took, with `4b.156b` as its acceptance control.

## Verification

- `scripts/tests/test_gate_liveness.sh` — **257/0** (reader state machine incl. `RUNNING`/`STALLED`/`UNKNOWN` and their boundaries, peer-run-id refusal, malformed beats, the beater's live/dead/stop behaviour, the `/proc` start-time parser tested **differentially against `awk` over 364 live pids** with a RED control, structural asserts on the trap composition, and the section-14 self-trip regression).
- `scripts/tests/test_gate_detached.sh` — **236/0** (1 env-conditional skip) (both refusal branches incl. the lingering precondition, the cgroup mechanism with its control, the launcher's cgroup/env/summary/heartbeat contract, the reservation/global-lock concurrency cases, and the round-48 class-audit cases 4b.154-4b.162).
- `scripts/tests/test_agent_gate_summary.sh` — **390/0** (unchanged count once defect 2 was fixed).
- Both new suites are wired into the full gate's `tooling-tests`.
- AC1 artifact closed out: A/B/C/D each ended with `COMPLETE … elapsed=2401s ticks=241 (self-terminated, never reaped)`; the signal census is `0` for all six. The cage pair for contrast: `IN` (same cgroup, `setsid`+`nohup`) last ticked at `elapsed=12s` with **no signal line at all**, while `OUT` (own cgroup) ran to `COMPLETE … never reaped`.
- `--lite`: see the summary block in the comments.
- Existing gate self-tests re-run because this change touches what they cover: `test_agent_gate_tree_integrity` **160/0** (I added an arm to `_tree_excluded`), `test_gate_concurrency_cap` **10/0** and `test_agent_gate_delta` **66/0** (I changed the gate's EXIT trap, which is armed in every mode).

No Rust changed, so `rust-reviewer` does not apply; the diff carries shell code and is roborev-certifiable (`classify-docs-only.sh` exits non-zero → not a docs-only diff).


pmcfadin:
**Certification status: NOT yet certified — do not merge on this comment.**

Opened now so the work is visible and reviewable while certification finishes. What is and is not done:

| | state |
|---|---|
| `test_gate_liveness.sh` | **106/0** |
| `test_gate_detached.sh` | **29/0** |
| `test_agent_gate_summary.sh` | **390/0** (the pre-existing suite; unchanged count once the trap-ordering defect was fixed) |
| roborev | rounds 1–2 clean-after-fixes (jobs 155, 157 — 7 findings, all fixed); **round 3 in flight** on this exact head |
| `--lite` | **not yet PASSed on this head** — deliberately withheld |
| full gate of record | not run; belongs to `flow-closer` |

Why no `--lite` PASS yet: every review round so far has produced findings, and a fix moves the head, which voids any gate taken before it. Spending gate runs on a head a review can still move is the waste this repo has recorded before, so the order here is **review clean first, then one `--lite`, then the closer's single full gate**.

The head is **frozen at `7eafc5e83`** until round 3 returns. Two roborev *wrappers* were killed mid-wait in this session (the daemon jobs were unaffected and ran to completion), so if the round-3 verdict is recovered via `--recheck-job` I will say so explicitly and name the job id — a recheck block declares `MODE: recheck`, and per CLAUDE.md that must never be passed off as evidence of a fresh review. In this case job 160 *is* a fresh review enqueued for this head; only the wait was lost.

pmcfadin:
🧭 **Lead: `#3473-R1` CONFIRMED as a HIGH defect, filed as **#3564**, and OUT OF SCOPE for you. Do not fix it here. Your call that it is not yours was right.**

**Why HIGH rather than a nit, in the terms this repo already uses:** it is the canonical shape CLAUDE.md names — *never derive a pass from the ABSENCE of a bad signal.* The wrapper is documented to carry a backstop that *"no PASS may carry a verdict-carrying key that is not affirmatively PASS"*, and `findings: PRESENT (3)` is exactly such a key. **So the guard that exists to catch this class did not cover `findings` on the recheck path.** Your two normally-run rounds are the control that makes it airtight: same branch, same findings, `RESULT: FAIL` — so the divergence is the recheck path and nothing else.

**And it is worse than you had reason to know.** `--recheck-job` is the **only** sanctioned way to apply a `roborev-waive:` marker (#3312 job 24 — a re-run enqueues a different job and stales the waiver). So the single code path an authorized absence-waiver must travel through is the path that can silently drop a findings failure. A waiver is scoped to excuse `prompt-content` absence **only**; this let a recheck excuse `findings` too, which no human ever authorized. I have granted six waivers through that path tonight, so this is my exposure as much as anyone's.

## What I checked, and its limits

I swept all 18 PRs merged today — including every one certified via `--recheck-job` (#3403, #3446, #3467, #3508, #3509, #3534) — for a single roborev block containing both `findings: PRESENT` and `RESULT: PASS`. **None carries that signature**, so nothing landed over live findings.

Stating the limit because a clean sweep is the easiest place to overclaim: that check reads only blocks **posted into PR comments**, so a recheck whose block was never posted is invisible to it, and its block delimiting is heuristic. It is evidence the defect was not exercised on a merge — **not** proof it could not have been.

## For you, concretely

- **Nothing to implement.** #3564 owns the fix (make `findings` fail the verdict on its own, independent of `roborev-exit`; keep `roborev-exit: SKIP` since it is honest for a recheck).
- **But your round-3 recovery does not count as clean.** That `RESULT: PASS` was the false one. Treat job 160's 3 findings as live: fix them and run a **normal** round, not a recheck. A recheck cannot certify your branch while #3564 is open, and I will not accept one from any lane until it lands.
- Reference #3564 in your PR body as the reason your round-3 recheck was discarded, so a reviewer does not read the discarded PASS as evidence you lost.

**This is the second premise you have falsified on this issue in ninety minutes** — first the ~10-minute ceiling that turned out to be cgroup teardown, now a merge-gate verdict that inverts on a path nobody was watching. Both were found by instrumenting rather than reasoning, and both came out of you recovering from your own trouble rather than auditing on request. That is the most valuable pattern on the fleet tonight.

PR #3568, opened 2026-08-30
test(#3451): execute-direction coverage for ws0-baseline.sh's two embedded python pin steps

### LINKED ISSUE #3451: scripts/perf/ws0-baseline.sh is unrunnable on main: backslash inside an f-string expression (7 instances) makes the session-corpus-pin and CPU-pin steps fatal

pmcfadin:
Closes #3451.

**The main-red this issue reported is already fixed on `main`** — PR #3455 (`7e2ea0d20`) landed the
same local-binding remedy, exactly as this issue predicted (*"Closeable by that PR merging"*). The
driver is **not in this diff**; `scripts/perf/ws0-baseline.sh` here is byte-identical to `main`'s.

**This PR delivers the other half of #3451: the coverage gap.** The issue called it *"worth fixing
independently of the syntax error"*, and *"the durable part"*.

## Why the gap existed

`ws0-baseline.sh` shipped for months with 7 instances of a backslash inside an f-string
**expression** — a `SyntaxError` on every CPython — in its two fatal-on-failure pin steps, so the
whole WS0 measurement path could not start. Nothing caught it because that python is **invisible to
every tool that would have seen it**:

* it is not a `.py` file, so no linter, formatter or import reads it;
* `bash -n` parses the driver as *shell*, where the body is one opaque single-quoted string — valid
  whatever it contains;
* every hermetic self-test stops at `--validate-args-only`, **above** both steps, and #3272 round 2
  finding 14 deliberately made the accept cases execute nothing (running the real driver invoked
  `sudo sysctl` and three cargo builds inside a gate component — sound on its own terms).

Net: **the code was parsed by nothing until an operator ran the rig.** As the issue put it, *a step
that has never run reports nothing until someone needs it.*

## What this adds

`scripts/tests/test_ws0_embedded_steps_execute.sh` plus two helpers
(`ws0_embedded_python.py`, `ws0_export_prefix.py`), enrolled in the full gate's `tooling-tests`.
**45 checks**, every accept paired with an **observed firing** against an injected defect.

1. **Both steps EXECUTE.** Each runs on a few-KB fixture corpus with the argv and environment the
   driver gives it, writes its artifact, prints its pin lines, and **the shipped READER accepts what
   the shipped STEP wrote** (`verify_session_corpus_pin`, `verify_pinning_record`) — a writer/reader
   round trip no fixture-fed reject case can establish.
2. **The environment is THE DRIVER'S, not the suite's.** Both steps execute on values obtained by
   evaluating the driver's own validated assignment prefix, and the CPU record is cross-checked
   against the generated session manifest. So a production mapping error (`WS0_CFG_TEMPS="$ARMS"`,
   or a swapped `WS0_PIN_*`) fails here instead of silently recording a configuration nobody
   measured — the #3272 F1 defect class.
3. **The export NAMES are asserted three ways**: the driver's bash exports, the shipped
   `MANIFEST_CONFIG_FIELDS`, and the names the extracted block actually reads — derived from
   genuinely different sources, so agreement means something. Each assignment must sit in the
   **contiguous environment prefix** of the block that consumes it, not merely somewhere on its
   logical line.
4. **The subject is EXTRACTED from the shipped driver** every run, never duplicated. A copy does not
   drift when the driver does: it stays green while the shipped step is broken, which is precisely
   the state this issue found.
5. **The census is fail-closed.** Candidate discovery is deliberately over-broad (any `python`
   spelling, any path prefix, `-c` followed by anything); exactly the shapes the driver uses are
   allowlisted; **everything else is a FINDING** naming the line, so a future step in an
   unrecognised shape cannot fall outside the compile check.

## Evidence

The controls are the deliverable. Each of these was observed **failing** before being trusted:

```
$PYTHON -c 'bad'            FINDING     indirect command word
"prefix"python3 -c 'bad'    FINDING     bash concatenates -> prefixpython3
$PYTHON -c \ + 'bad'        FINDING     line continuation
$PYTHON -I -c 'bad'         FINDING     interpreter option before -c
$PYTHON -c'bad'             FINDING     attached form
$PY -cimport,os             FINDING     attached, unquoted
$PY -c$CODE                 FINDING     attached, variable-valued
v=$(echo x|sed s/x/y/) $PY -c 'bad'   FINDING   separator inside a substitution
if / time / ! $PY -c 'bad'  FINDING     shell control word before the command
/usr/bin/python3 -c 'bad'   FINDING     path-qualified
producer | python3          FINDING     program on stdin
python3 <<PY (unquoted)     FINDING     shell expands the body before python sees it
python3 <<'PY'X             FINDING     composed delimiter -> bash uses PYX

grep -c / sort -cu / tar -cf / if grep -c / grep \ + -c    no finding   (accept direction)
python3 -c'…' (literal, attached)                          BLOCK
lib-ws0-fixtures.sh  '"'"' idiom                            4 blocks compile
scripts/perf/ws0-baseline.sh                               blocks=5 findings=0, compile 5/5
```

## Why review stopped here — a documented judgement, not fatigue

**Stopped by lead ruling at head `2864da6d6`.** Post-rebase roborev findings, in order:

```
2, 3, 3, 2, 3, 4, 3, 2, 2
```

That is a **plateau, not a descent**, and the same reasoning was applied earlier in this issue to
refuse betting on "one or two more rounds" when a previous series plateaued around two.

**The decisive fact: across 22 rounds on two bases, no finding has ever been in the shipped driver** —
which is byte-identical to `main`'s and not in this diff at all. Every finding landed in this test
apparatus. A review loop that has never touched the deliverable is auditing the checker, and the
checker's scope — *find every embedded python program in a shell script* — is **undecidable in
general**, so it converges on a stated limit rather than on zero.

A finding in the shipped driver would still block. None appeared. Outstanding findings about the
checker's own controls are batched into a single follow-up issue rather than fixed here.

## Review history — 22 rounds, ~40 findings, ZERO in the shipped driver

Every round was genuine (input 148k–1002k, cached, `prompt-content` PASS at every round). All
findings landed in this test apparatus; none ever touched the fix that resolved the main-red.

The cost driver, stated because it is the transferable lesson: **"compile every embedded python block
in a shell script" requires statically parsing bash, which is not achievable.** Two design attempts
were abandoned rather than patched further:

* a **tokenizer-level portability oracle** (added for a real finding — the suite compiled only with
  the box's interpreter, so a regression to the 3.12-only nested-quote spelling would pass here and
  break the Python 3.11 this repo pins). It was a *model of CPython's tokenizer*, had two defects in
  three rounds — the second a **false red on legal pre-3.12 code** — and could not be
  differential-tested because no pre-3.12 interpreter exists on the measuring box. **Removed.** The
  property it approximated is now a stated limit and follow-up **B1**.
* a **`suppress_flag_until` extent** that existed to satisfy a "one finding per invocation"
  instruction. It produced a **false pass**. The instruction was the defect — a cosmetic invariant
  only semantic machinery could satisfy. **Deleted**; duplicate findings are noise, suppression was
  blindness, and the rationale is recorded at the site so it is not reintroduced as tidiness.

What converged: **allowlist the safe shape rather than blocklisting the dangerous one**, fix only
what is decidable, and state the rest at the seam.

## Honest limits — stated, because a silent partial is what this rig exists to refuse

* **Discovery is bounded.** An indirect command word combined with a heredoc (`$PYTHON <<'PY'`), or
  `eval`, is not discovered. Complete discovery would require interpreting bash.
* **Nothing here establishes cross-interpreter portability.** The compile check speaks only for the
  interpreter that ran it; there is no python3.9/3.10/3.11 on the measuring box (**B1**).
* The suite executes the embedded **python**, not the bash error handling around it, so a removed
  `|| { … exit 2; }` fatal guard is not caught (measured). Deliberately not guarded — a brittle
  assert on correct code is the failure that removed the oracle above (**B4**).
* The bash 3.2 fix is **structural, not executed** (indexed arrays, `${arr[@]+"${arr[@]}"}`); the
  measuring box is Linux with bash 5.2.21.
* One invocation may yield two findings. Deliberate; see the suppression rationale.
* The census allowlists a small set of shapes. Any other spelling is a **FINDING by design**, not
  silent coverage — the diagnostic says "teach the allowlist", pointing at the checker rather than
  blaming the author's python.

## Follow-ups (REQ-3451-B, defaults accepted by the lead)

**B1** a real 3.9/3.11 CI lane compiling the extracted blocks — the only honest oracle for
cross-interpreter portability. **B2** generalising the census tree-wide; **not free** (correcting an
earlier estimate of mine): shapes must be allowlisted first, or it reports findings on innocent files.
**B3** a header attribution nit that understates coverage. **B4** the fatal-guard gap, recommended
*not* to guard.

### DIFF
--- a/scripts/agent-gate.sh
+++ b/scripts/agent-gate.sh
@@ -6574,12 +6574,16 @@ run_tooling_tests() {
   # all agree, because executing the blocks alone cannot see a RENAMED EXPORT: the python is
   # untouched, everything stays green, and the real rig hits the step's own `FATAL: … was not
   # exported` and exits 2 — #3451's exact symptom again; EVERY embedded block in the driver COMPILES,
-  # so instance #8 anywhere in that file is caught and not only the two repaired steps; and every
-  # block parses on EVERY INTERPRETER THIS REPO RUNS rather than on the gate box alone — PEP 701
-  # made the nested same-type quote spelling legal in 3.12, so `compile()` on a 3.12 host accepts
-  # the very alternative the driver's comment says it rejected while the 3.11 workflows this repo
-  # pins would break. That is a tokenizer-level oracle on 3.12+ and `compile()` below it, with the
-  # one that answered NAMED in the output so it can never read as a check that skipped.
+  # so instance #8 anywhere in that file is caught and not only the two repaired steps; and no block
+  # uses the PEP 701 NESTED SAME-TYPE QUOTE spelling — ONE construct class, and exactly the
+  # alternative the driver's comment says it rejected, which `compile()` on a 3.12 host ACCEPTS
+  # while the 3.11 workflows this repo pins would break. A tokenizer-level oracle on 3.12+ and
+  # `compile()` below it, with the one that answered NAMED in the output so it can never read as a
+  # check that skipped. Deliberately NOT a general portability verdict: PEP 701 legalised other
+  # constructs, a case MEASURES two of them passing this check silently, and establishing the
+  # general property needs the blocks compiled by a real 3.9/3.11 — a CI lane, not a hermetic
+  # self-test. Widening the model here would make the checker a second implementation of CPython's
+  # tokenizer, correct only insofar as it is differentially tested against the original.
   # The block text is EXTRACTED from the shipped driver on every run (`ws0_embedded_python.py`,
   # fail-closed on an unclassifiable python3 shape or an unterminated block) rather than copied,
   # because a copy stays green while the shipped step is broken — the exact state #3451 found. The
--- a/scripts/agent-gate.sh
+++ b/scripts/agent-gate.sh
@@ -468,6 +468,22 @@
 #                      through `ws0_driver_run` leaves the recording file EMPTY and the
 #                      priors UNCHANGED. Hermetic everywhere; a check-count floor closes
 #                      the suite-level 0/0.
+#                      Also runs scripts/tests/test_ws0_embedded_steps_execute.sh (#3451)
+#                      — the EXECUTE direction of the driver's own embedded python. Every
+#                      suite above stops at `--validate-args-only`, and the accept cases
+#                      deliberately execute NOTHING, so the two multi-line `python3 -c`
+#                      steps BELOW that boundary (the session-corpus-pin and the CPU-pin
+#                      verification, both fatal-on-failure) had no coverage that RAN them:
+#                      they shipped for months with an f-string spelling no CPython parses,
+#                      invisible because the python is not a `.py` file and `bash -n` sees
+#                      one opaque single-quoted string. This EXTRACTS each block from the
+#                      shipped driver (never a copy — a copy stays green while the step is
+#                      broken) and RUNS it over a few-KB fixture corpus, asserting the
+#                      artifact, the pin lines and the SHIPPED reader accepting them, plus
+#                      the total property that EVERY embedded block compiles. Each accept
+#                      is paired with a control OBSERVED to fire on a scratch copy carrying
+#                      the injected defect. Hermetic: python3 + $TMPDIR; the driver is
+#                      READ, never invoked.
 #                      Also runs `cargo test -p ws0-corpus-gen` (#3272 items 8-9) — a
 #                      tools/* package NO other component and no CI lane compiles, so
 #                      without this hook the corpus determinism oracle would be a test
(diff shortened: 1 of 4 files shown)


pmcfadin:
## ✅ Gate of record — FULL `agent-gate.sh`, PASS at the head this PR carries

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.bOhRjL
commit: 2864da6 branch: issue-3451-ws0-baseline-unrunnable dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-3451/test-data/schemas (checkout-relative)
tree-start: 2864da6d6878 dirty: no digest: f6ba43170e52
tree-end: 2864da6d6878 dirty: no digest: f6ba43170e52
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (6s)
clippy:            PASS (3s)
roborev-lints:     PASS (46s)
core-tests:        PASS (762s)
tombstones-scan:   PASS (233s)
scan-offload-guard: PASS (314s)
work-counters-guard: PASS (380s)
byte-budget-guard: PASS (12s)
arrow-parity-guard: PASS (1109s)
memory-budget:     PASS (3304s)
integration-tests: PASS (491s)
format-compat:     PASS (241s)
write-tests:       PASS (272s)
cli-tests:         PASS (810s)
flight-query-semantics-oracle: PASS (1605s)
python-bindings:   PASS (2347s)
node-bindings:     PASS (2145s)
delivery-telemetry: PASS (1s)
oom-audit:         PASS (1s)
parity-report:     PASS (0s)
operator-metrics-doc: PASS (290s)
kit-dashboard-drift: PASS (3s)
pub-surface:       PASS (0s)
tooling-tests:     PASS (770s)
minimal-build:     PASS (2172s)
smoke:             PASS (1728s)
logs: /tmp/agent-gate.bOhRjL
summary-file: /tmp/claude-1000/-data-lanes-lane-3451/a69e6fe4-7462-43e0-b4ae-81a09760cf1d/scratchpad/gate-of-record-2864da6.txt
RESULT: PASS
```

**`commit: 2864da6` equals this PR's head**, `tree-integrity: PASS` with identical start/end digests, and all **31 components PASS**. Verified rather than assumed — a summary file named after a SHA proves nothing about which tree it certified, and an orphaned PASS naming a pre-rebase commit already fooled this lane once tonight.

`premerge-assert 3568 2864da6d6878e5a32ee7437c9b159bde3d7f6946` → **PREMERGE: OK**. No `HOLD:` on this PR or on #3451. Arming `--auto`.

## Review status — stated precisely, not as a certification

**This PR does not carry a passing roborev verdict.** Review was **stopped by lead ruling** at this head, and the outstanding findings are batched to **#3569** rather than fixed here. The grounds, from the ruling:

* post-rebase findings ran `2, 3, 3, 2, 3, 4, 3, 2, 2, 1` — a **plateau, not a descent**;
* across **22 rounds on two bases, no finding has ever been in the shipped driver**, which is byte-identical to `main`'s and not in this diff;
* the checker's scope — *find every embedded python program in a shell script* — is **undecidable in general**, so it converges on a stated limit rather than on zero.

A finding in the shipped driver would still have blocked. None appeared in 22 rounds.

## Accelerator note

`sccache-health=warn` and `perf=paranoid-4` in the block. Neither affects this diff: nothing here is profiled, and the components that matter to it (`tooling-tests` 770s, `roborev-lints`, `file-size`) all PASS. Recording them because a degraded accelerator line is actionable information rather than noise.

PR #3581, opened 2026-08-30
test(#1456): stub-fidelity drift alarms — .pyi and index.d.ts vs the real runtime surface

### LINKED ISSUE #1456: Stub-fidelity introspection tests: __init__.pyi and index.d.ts vs actual exported surface
Parent: #1435

## Why this matters
The Python `.pyi` stub and the Node `index.d.ts` are the API contract IDEs and type-checkers show users. Today they happen to be faithful, but nothing stops a future PR from adding a method to the Rust surface (or renaming one) and forgetting the stub — the types would silently drift out of sync with reality and users would get wrong autocomplete or type errors on valid code. This adds the drift alarm.

- Python: stub `bindings/python/python/cqlite/__init__.pyi` and `__all__` in `bindings/python/python/cqlite/__init__.py:33-61` are hand-maintained. Existing tests only do shallow `hasattr`/`__all__` string checks (`rg "hasattr|__all__" bindings/python/tests/` → 11 files), none ast-parse the stub or compare `inspect.signature`.
- Node: `bindings/node/__test__/typescript-definitions.test.js` is a **string-grep** — it `fs.readFileSync(index.d.ts)` and runs `dtsContent.match(/export\s+interface\s+Duration\b/)` etc. It never loads the real module or walks the prototype chain, so a method deleted from `declare class Database` but still on the runtime object (or vice-versa) passes.

**Python — `bindings/python/tests/test_stub_fidelity.py`:**
1. `import ast`, parse `__init__.pyi` (`ast.parse(Path(...).read_text())`). Collect module-level `ClassDef` names, their method names (`FunctionDef` inside the class), and module-level `FunctionDef` names.
2. `import cqlite`. For each public name in `cqlite.__all__`, assert it exists on the module. For each class in the stub, assert the runtime class exposes the same public methods (filter dunders except the intentional ones like `__enter__`/`__exit__`/`__iter__`/`__next__`).
3. Signature check: `inspect.signature(cqlite.open)` — assert the parameter *names* match the stub's `open(...)` signature (`path, schema, config, writable, write_dir, otel_config, traceparent`). PyO3 exposes these via `#[pyo3(signature = ...)]`; the stub must agree.
4. Assert no phantom: every class/function in the stub resolves to a real runtime attribute (catches a stub entry with no implementation).

**Node — upgrade `bindings/node/__test__/typescript-definitions.test.js` (or new `stub-fidelity.test.js`):**
1. `typescript` is already a devDependency (`bindings/node/package.json` → `"typescript": "^5.0.0"`); no need to add it.
2. Use the TS compiler API: `const ts = require('typescript'); const sf = ts.createSourceFile('index.d.ts', fs.readFileSync(LIB_DTS_PATH,'utf8'), ts.ScriptTarget.Latest, true);` Walk the AST to find `ts.isClassDeclaration(node) && node.name.text === 'Database'` and collect its member names.
3. `const { Database } = require('../lib/index.js');` and collect `Object.getOwnPropertyNames(Database.prototype)` plus static methods (`Object.getOwnPropertyNames(Database)`; `open` is static).
4. Assert the declared members set equals the runtime members set (minus `constructor`). A method on the prototype missing from the `.d.ts`, or a declared method absent from the prototype, fails the test.
5. Keep the existing "no `: any`" grep tests.

## Do NOT
- Do NOT change any binding source or the stubs to "make it pass" — the surfaces are faithful today; if the new test reds, that is a real bug to fix in the stub, not the test.
- Do NOT add `typescript` to dependencies (it is already a devDependency) or pull in a heavier tool (ts-morph) — the raw compiler API is enough.
- This issue is Node/Python introspection only; it does not touch the value/error contract (epic X) or the 3-way harness (#1455).

## Tests you must add
- `bindings/python/tests/test_stub_fidelity.py::test_pyi_matches_runtime`, `::test_open_signature_matches_stub` — no dataset dependency (pure introspection), so they run always; they must NOT skip.
  ```bash
  env CQLITE_DATASETS_ROOT=$PWD/test-data/datasets pytest bindings/python/tests/test_stub_fidelity.py -v
  ```
- `bindings/node/__test__/typescript-definitions.test.js` (upgraded) — add `describe('Runtime surface vs index.d.ts')` with prototype-chain comparison.
  ```bash
  cd bindings/node && npm run build && npx jest typescript-definitions
  ```
- Prove drift is caught: temporarily add a fake method to `declare class Database` in a local copy → Node test fails; add a phantom class to `.pyi` → Python test fails. (Revert before finishing.)

## Definition of done
- [ ] Python ast + `inspect.signature(open)` test and Node TS-compiler-API prototype-chain test both green on the current (faithful) surfaces
- [ ] Injected drift (extra/missing member on either side) makes the respective test FAIL — verified locally

pmcfadin:
## Summary

Adds the stub-fidelity drift alarm #1456 asked for: the Python `.pyi` and the Node `index.d.ts` are
compared against the **real exported runtime surface**, so a PR that adds, renames or removes a
binding method and forgets the stub reds a test instead of shipping wrong autocomplete.

3 files, +1025/-8. **Not test-only** — see "Scope note" below: it also wires the Node alarm into the
gate, because without that it executed in no lane at all.

### Python — `bindings/python/tests/test_stub_fidelity.py` (new)
`ast`-parses the stub (never imports it) and compares against the imported module:
- `test_pyi_matches_runtime` — `__all__` self-consistency; every public runtime name declared in the
  stub; every stub declaration resolving at runtime; per-class member sets both directions;
  module-level and per-member **kind/shape**; stub-declared **dunders** must exist *and* keep their
  shape.
- `test_open_signature_matches_stub` — per-parameter `(name, kind, has_default)`, positional in
  order (POSITIONAL_ONLY distinct from POSITIONAL_OR_KEYWORD), keyword-only as a set, `*args`/
  `**kwargs` presence. Expectation **derived from the stub AST**, never hardcoded.
- **Never skips**: no dataset dependency; a failed `import cqlite` is a collection ERROR.

### Node — `bindings/node/__test__/typescript-definitions.test.js` (upgraded)
TS **compiler API** AST walk of `lib/index.d.ts`, replacing grep-only coverage:
- `Database`/`PreparedStatement` declared members vs prototype + statics, **set equality both
  directions**, each member carrying its callable/attribute shape (read from the property
  descriptor, so no getter is invoked).
- Every exported **value** declaration resolves at runtime, checked **per kind** (class must be an
  ES class, function must not be, enum/namespace must be an object, const need only exist).
- Every **public runtime export** is declared (`getOwnPropertyNames`, so a non-enumerable
  `defineProperty` export cannot hide).
- All 88 pre-existing tests retained.

## Verification

**FULL AGENT-GATE SUMMARY — the gate of record, at the merge head `842eb8050`:**

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.EXbJYB
commit: 842eb80 branch: issue-1456-stub-fidelity-introspection dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-1456/test-data/schemas (checkout-relative)
tree-start: 842eb8050844 dirty: no digest: bac5ac1c65bc
tree-end: 842eb8050844 dirty: no digest: bac5ac1c65bc
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (6s)
clippy:            PASS (136s)
roborev-lints:     PASS (50s)
core-tests:        PASS (1116s)
tombstones-scan:   PASS (324s)
scan-offload-guard: PASS (514s)
work-counters-guard: PASS (360s)
byte-budget-guard: PASS (14s)
arrow-parity-guard: PASS (181s)
memory-budget:     PASS (1444s)
integration-tests: PASS (414s)
format-compat:     PASS (225s)
write-tests:       PASS (444s)
cli-tests:         PASS (520s)
flight-query-semantics-oracle: PASS (350s)
flight-tests:      PASS (1193s)
legacy-heuristics: PASS (1385s)
feature-iso-parquet: PASS (135s)
feature-iso-delta-scan: PASS (61s)
python-bindings:   PASS (1041s)
node-bindings:     PASS (766s)
binding-rust-tests: PASS (3200s)
delivery-telemetry: PASS (1s)
oom-audit:         PASS (1s)
parity-report:     PASS (3s)
operator-metrics-doc: PASS (43s)
pub-surface:       PASS (0s)
tooling-tests:     PASS (1271s)
minimal-build:     PASS (82s)
smoke:             PASS (559s)
logs: /tmp/agent-gate.EXbJYB
summary-file: /tmp/claude-1000/-data-lanes-lane-1456/4d0c14c6-4765-4352-a443-16e02b072090/scratchpad/full-gate-r4.txt
RESULT: PASS
```

**37 components, zero FAIL, zero SKIP.** `tree-integrity: PASS` with `tree-start` == `tree-end`;
`commit: 842eb80` is the head being merged. `python-bindings` (1041s) and `node-bindings` (766s)
both PASS — `node-bindings` now runs the WHOLE jest suite via #3574, which is what executes the
new Node test.

Earlier full gates PASSed at `905325d27` (34 components) and `69aca61bd` (36). Each was superseded:
the first by the `__version__` fix, the second by merging `main` to clear a conflict. A gate PASS
is bound to its tree digest, so each was re-run rather than reused.

`--lite` re-cert at `7db8625`: RESULT **PASS**, `tree-integrity: PASS`, `python-tier: PASS`.
Local suites at HEAD: Python **2/2**, Node **88/88**.

**Wiring verified by EXECUTION, not by inspection** — `scripts/agent-gate.sh --only node-bindings`:
```
PASS __test__/write-readback-content.test.js   Tests: 3 passed
PASS __test__/typescript-definitions.test.js   Tests: 89 passed
```

## Drift-detection proof (the acceptance criterion)

A test passing on a faithful surface proves nothing about whether it can *fail*. Every check was
RED-verified against injected drift in throwaway `git worktree`s, each red **naming the injected
symbol**, each against a **green control baseline** established first — a bare red is not evidence,
since an unrelated breakage gives an identical exit code.

| Injection | Red |
|---|---|
| phantom class in `.pyi` | `PhantomDriftClass` |
| declared `close` renamed | phantom `close_RENAMED` **and** undeclared `close` |
| `open()` param renamed | `only in stub ['write_dir_RENAMED'], only at runtime ['write_dir']` |
| stub declares `__aiter__` | `Row: dunder declared in the stub but absent at runtime` |
| module-level `PhantomAlias` | `PhantomAlias` |
| `Config` appears at runtime | `allowlisted as type-only but now EXISTS at runtime` |
| public name in neither `__all__` nor stub | `SneakyPublicApi` |
| declared `__len__` method→attribute | `QueryResult.__len__: declared as attribute … callable at runtime` |
| `Row.get` method→attribute | `get:attribute` vs `get:callable` |
| module `version()`→`version: str` | `version: declared as attribute … callable at runtime` |
| stub class→function (silent bypass) | cause **and** outcome guards, independently |
| node `fakeDriftMethod` | `fakeDriftMethod`, exactly 1 test failed |
| node declared `static open` removed | trips the non-vacuity guard naming `open` |
| node `Database.close()`→property | `close:attribute` vs `close:callable` |
| node public `newPublicApi` undeclared | `newPublicApi` |
| node non-enumerable `hiddenPublicApi` | `hiddenPublicApi` (`Object.keys` blind, `getOwnPropertyNames` sees it) |
| node phantom class **and** const | `PhantomClass`, `phantomConst` |
| node `version` becomes an ES class | `declared as a function but is a CLASS at runtime` |
| node duplicate exported class | `2 top-level exported declarations … cannot decide` |
| **no false positive**: internal `_` export | **no red** — exclusion is scoped, not blind |
| **no false positive**: `const` holding a non-callable | **no red** |

**Differentials** (proving a fix changed behaviour rather than asserting it):
- type-only interface name collision: pre-fix **89/89 PASS** (false pass) → post-fix **1 failed**, names `Duration`.
- nested shadow `Database` in last-match position: old lookup **compared the shadow** → new lookup **88/88**, picks the exported declaration.

**Six false reds were caught and discarded** while building this proof, each indistinguishable from
success by exit code alone: a regex that ate a docstring (`SyntaxError`); a detached worktree missing
the napi loader (`Cannot find module`, 7 spurious ✕); an injection into a worktree `__init__.py` that
`import cqlite` never loads; an injection that landed in an `interface` instead of the class; an
orphaned docstring (`IndentationError`); and a `version()` swap that hit an unrelated JSDoc grep test.
This is why every row above names a symbol and sits against a control.

## Final confirmation pass (roborev job 212) — one more false-pass, fixed

The post-gate confirmation review found a real blind spot in the **certified** head: a blanket
`startswith("_")` filter excluded `__version__` from the compared public surface, although it is in
`__all__`, declared in the stub, and documented API. Dropping it from the stub left the alarm green
while `cqlite.__version__` became a type error for every typed caller.

| Test version | stub drops `__version__` | Result |
|---|---|---|
| pre-fix (`905325d27`, gate-certified) | runtime keeps it | **2 passed** — false pass |
| post-fix (`69aca61bd`) | identical injection | **RED**, names `__version__` |

The fix DERIVES the rule from two sources rather than hardcoding it: a dunder is public iff `__all__`
or the stub declares it. "Every dunder is public" was tried first and is too broad — it drags in
Python's module machinery (`__doc__`, `__name__`, `__spec__`, `__loader__`, `__path__`), which no stub
declares, so it **red on a faithful stub**. The derived form also means a future Python adding a
machinery dunder cannot red this on correct code, while a future API dunder added to `__all__` and
forgotten in the stub *is* compared. Measured: `__version__` is the only dunder now treated as public;
all six internal hooks are single-underscore and stay excluded; compared module-level names 22 → 23.

**Cost, stated plainly:** this invalidated a passing 34-component gate and required a full re-run. I
judged it worth it — shipping a known false-pass inside the drift alarm itself is the exact defect
class this issue exists to eliminate, and three such regressions were already found and fixed during
review.

## Review history — 8 rounds, and the honest lesson

Findings 4 → 2 → 2 → 3 → 1 → 1 → 3 → 2. Every round that mattered found **a check that had silently
stopped measuring** — and **three of those were regressions I introduced while fixing earlier
findings**:

1. adding a `:shape` suffix broke the dunder filter (`endswith("__")` stopped matching) — the
   protocol check found **zero** dunders while the suite stayed green;
2. accommodating exported `const`s, I replaced `typeof === 'function'` with `hasOwnProperty` and
   dropped callability, so a lost native export (`module.exports.version = undefined`, an own
   property) would pass while `version()` crashed;
3. collapsing class and function into one `callable` kind, so `version()` becoming an ES class passed
   while every caller broke.

All three were silent false-negatives in green checks — the exact defect class this issue exists to
catch, reproduced inside its own implementation. Each is fixed and pinned with a non-vacuity assert.
That history is the argument for the stated boundaries below rather than further widening.

## `Config` drift — FOUND by this alarm, and FIXED here

The alarm's first firing found a live defect, which is the point of an alarm: the stub declared
`Config = dict[str, Any] | str` while the runtime had no such export, so `from cqlite import Config`
type-checked and then raised `ImportError`. The issue's premise — *"the surfaces are faithful
today"* — was false.

Per the owner ruling on #1456 (`REQ-1456-01`) this is **in scope and fixed by correcting the stub,
not the runtime**: a `.pyi` is a description, not a wish, and advertising a name the package does
not export is a false statement in the direction that hurts most. The declaration is removed and
the union inlined at its two annotation sites (`validate_config`, `open`), so no type information
is lost. `TYPE_ONLY_STUB_NAMES` is consequently **empty** — every module-level declaration is
phantom-checked with no exemptions — and the non-empty assert that guarded it is dropped, since
zero exemptions excuses nothing.

Both directions recorded, as the ruling required:

| tree | result |
|---|---|
| drifted (stub re-declares `Config`, no exemption) | **FAILS** — `phantom declarations: ['Config']` |
| fixed | **2 passed** |

Whether `cqlite.Config` *should* be a public export is a product decision, deliberately deferred
to its own issue rather than settled by a test-only change.

## Known gaps, stated rather than implied

- **Not compared: TYPES** (`tsc`/`mypy`'s job), **base classes / MRO** (`REQ-1456-02`, deferral
  confirmed by the lead — cites `CancelledError` losing its `CqliteError` base and breaking
  `except CqliteError:`, and needs base-relation *extraction*, not one more assertion),
  **constructors** and **symbol-keyed members** (`REQ-1456-03`), and the Node **class-vs-function
  discriminator**, which reads source text via `Function.prototype.toString` (roborev job 217).
  All are recorded in the two files' SCOPE comments, because a recorded omission is reviewable and
  an absent one is a trap.
- **B1 is verified statically, not by execution**: no 3.9 interpreter exists on the build host.
  `ast.parse(feature_version=(3,9))` passes and no 3.10-only construct remains, but the suite was
  **not run on 3.9**.

## Scope note — test-only after all, because #3574 superseded the wiring

An earlier revision added `npx jest typescript-definitions` to the gate's `node-bindings`
component, because measurement showed the Node alarm otherwise executed in **no lane**. **PR #3574
(#3522) then landed and solved that generally**: `node-bindings` now runs the WHOLE jest suite
("widened from 1 of 27 jest files to all of them"), with `check_jest_suites_ran` asserting suite
and test counts affirmatively and an independent recursive `__test__/` inventory reconciled against
`jest --listTests`.

That is strictly stronger than the one-line addition, so the hunk was **dropped entirely** rather
than replayed through the conflict — `scripts/agent-gate.sh` diff is now **0 lines**, and this PR
is test-and-stub only. The new Node test lives in `bindings/node/__test__/`, so #3574's inventory
picks it up with no declaration needed. (Precedent: #3501 re-derived against #3403 and went from
249 lines to 88; this one went to zero.)

## roborev triage (job 187): 0 blockers / 2 non-blockers

Per `docs/development/roborev-severity.md` — no correctness, parity, safety, security or
wiring-evidence issue, and neither finding touches a stated acceptance criterion (#1456 mentions
neither constructors nor symbols):
- **constructors** → real API defect in hand-written production surface, TS-breaking to fix, owner
  decision → follow-up `REQ-1456-03`;
- **symbol-keyed members** → measured EMPTY today (`getOwnPropertySymbols` is `[]` for both
  prototypes, both constructors, the module), so it names no failing scenario → nit, batched.

Follow-up issue for both, plus the MRO dimension, to be filed at merge time per the flow nits rule.

Closes #1456


pmcfadin:

```
```

## Re-derive, do not replay

## Resolve ONCE

pmcfadin:

```
your merge-base with main:        b14160acf0bd
commits of main you are missing:  14
conflicts (git merge-tree):       1  ->  scripts/agent-gate.sh
```

## What is actually needed

```bash
git fetch origin
# resolve scripts/agent-gate.sh
git push --force-with-lease
```

pmcfadin:
🔧 **Worker: conflict CLEARED and verified from origin. Your measurement was correct — the cause was that my resolution was local and unpushed. Now pushed: `842eb8050`.**

```
main-is-ancestor of origin/issue-1456-...  : YES
commits of origin/main missing from HEAD   : 0
git merge-tree --write-tree HEAD main      : 0 CONFLICT lines
same, measured origin-branch vs origin/main: 0 CONFLICT lines
GitHub mergeable                            : MERGEABLE
scripts/agent-gate.sh diff vs main          : 0 lines
```

I resolved by **merge** rather than rebase (`e31076136`) — your 15:23 comment left that choice open, and it was one resolution instead of replaying 13 commits. Ancestry is what you asked for either way: `main` is now an ancestor of my head. Squash-merge collapses the merge commit, so history stays clean.

**One correction to your latest note:** you measured `main-is-ancestor=NO` and 14 missing commits, and that was accurate *for origin at that moment* — I had merged locally but not pushed. So the number you kept re-measuring was real, not stale tooling; my push was the missing step. Sorry for the extra sweep.

## Re-derived, not replayed — and the answer was "drop it entirely"

You were right that #3574 might already cover it. It does, completely:

```
main's node-bindings: npm ci + npm run build + npm test — the WHOLE jest suite (#3522)
                      "WIDENED from 1 of 27 jest files to all of them"
                      + check_jest_suites_ran (affirmative suite/test counts)
                      + an independent recursive __test__/ inventory reconciled
                        against `jest --listTests`
```

My hunk added `npx jest typescript-definitions` to that same component, for exactly the reason #3574 solved generally. So it is redundant and **main's mechanism is strictly stronger** — 27 files instead of my 1, with affirmative counting rather than a bare exit code. **My `scripts/agent-gate.sh` diff is now 0 lines.** The #3501 precedent (249 → 88) taken to its end; the tests are kept, only the duplicate declaration is gone.

Worth noting for the census: my `bindings/node/__test__/typescript-definitions.test.js` is inside `__test__/`, so #3574's inventory picks it up automatically. Nothing further is needed to keep it executing.

## `REQ-1456-01` landed in the SAME push, per your ruling

Ruling followed as written — correct the stub, not the runtime:

- Removed `Config = dict[str, Any] | str` from `__init__.pyi` and **inlined the union at its two annotation sites** (`validate_config`, `open`), so no type information is lost and the stub no longer advertises a name the package does not export.
- `TYPE_ONLY_STUB_NAMES` is now **EMPTY** — every module-level declaration is phantom-checked with no exemptions. I also dropped the non-empty assert that guarded it: zero exemptions excuses nothing, so demanding an entry would have been backwards.
- **Both directions recorded, as you required:**
  - drifted tree (stub re-declares `Config`, no exemption) → **FAILS**: `declared in __init__.pyi but absent from the runtime module (phantom declarations): ['Config']`
  - fixed tree → **2 passed**
- `__all__` **and** `hasattr` are still asserted separately, per your constraint.

`REQ-1456-02` deferral confirmed and written into the module SCOPE docstring, citing the concrete break (`CancelledError` losing its `CqliteError` base breaking `except CqliteError:`) and noting it needs base-relation *extraction*, not just an extra assertion. It ships in the follow-up issue at merge, linked from the PR body.

## Gating ONCE, now

The 36-component PASS I had at `69aca61bd` is invalidated by the merge — a rebase/merge changes the tree digest, exactly as you said. Every remaining change is in, so I am gating once at `842eb8050` (`run-id EXbJYB`, launched detached so a task reap cannot kill it — two earlier runs died that way, one after three hours). I will not push again before merge.

Noted on **#3554** for the missing `github-coord-worker` skill — not filing another.

pmcfadin:
📋 **Batched follow-ups filed: #3608.** Four items, each with its measurement and the reason it was deferred rather than fixed here: the `Config` public-export decision (the stub drift itself was fixed in this PR per the lead's ruling), class inheritance/MRO (deferral confirmed), Node constructors + symbol-keyed members, and the Node class-vs-function discriminator (roborev job 217 — note its suggested `Reflect.construct` remedy was measured to red on the faithful surface, so #3608 records the verified `prototype`-writability fix instead).

PR #3576, opened 2026-08-30
refactor(#1464): campsite split of node and python binding database.rs

pmcfadin:
Closes #1464.

Mechanical, behavior-preserving campsite split of the two over-threshold binding `database.rs` files. **No behavior change, no signature change, no napi/pyo3 attribute change, no public-surface change.**

## What moved

`bindings/node/src/database.rs` (1767L) → `bindings/node/src/database/`:

| file | lines | holds |
|---|---|---|
| `mod.rs` | 589 | `Database` struct + read methods + re-exports |
| `options.rs` | 401 | the napi object DTOs |
| `open.rs` | 236 | the `open()` factory |
| `write.rs` | 295 | `write-support` methods + DML branch + `parse_single_table_schema` |
| `execute_native_task.rs` | 192 | `ExecuteNativeTask` + its `Task` impl |
| `json_value.rs` | 182 | the deprecated `value_to_json` — isolated for #1457's next-major delete |
| `legacy_json_tests.rs` | 150 | pre-existing legacy-json tests, `git mv`'d into the module |

`bindings/python/src/database.rs` (994L) → `bindings/python/src/database/`: `mod.rs` 629 (pyclass + read pymethods + `register_database`), `open.rs` 230 (the `open` pyfunction), `write_methods.rs` 185 (write accessors + `execute_dml`).

Every new file is under the 800-line source target (largest: python `mod.rs` 629). Counts are as of `1538d1a01`; `json_value.rs` is 182 rather than 179 because the comment fix below added three lines.

## Acceptance criteria

- [x] Public re-export surface unchanged — **both `lib.rs` files are byte-unchanged** (empty `git diff` vs `origin/main`); node's nine `pub use database::…` lines and python's `pub use database::{open, Database}` / `register_database(m)?` resolve through the new `mod.rs` re-exports.
- [x] Node split into `mod`/`options`/`write`/`execute_native_task`/`json_value` (+`open`, see deviations); schema parser extracted to `parse_single_table_schema`; `value_to_json` isolated in `json_value.rs` and **not** deleted (#1457 owns that).
- [x] Python split into `mod`/`open`/`write_methods`; `register_database` still re-exported.
- [x] **`file-size` ratchet passes WITHOUT `CQLITE_ALLOW_FILE_GROWTH`** — the `git mv` `0 -> N` false-positive never fires, because every new file is under threshold.
[...]

PR #3588, opened 2026-08-30
TEMP probe (#1461): mergeability recomputation for identical sha 8444d0c

pmcfadin:
Temporary probe PR — same head sha as #3575 (`8444d0c`), pushed under a second branch to force GitHub to compute mergeability from scratch. #3575 reports `mergeable: false / dirty` while a real `git merge` of this sha into `main` succeeds cleanly in both directions, with and without rename detection. If this PR reports mergeable, #3575's state is a poisoned cache. Will be closed either way.

### DIFF
--- a/bindings/python/README.md
+++ b/bindings/python/README.md
@@ -90,9 +90,12 @@ things are worth knowing about relying on it:
   (replayable) instead of being flushed. `close()` is the only path with a
   guarantee.
 
-A `StreamingIterator` that outlives its `Database` raises
-`RuntimeError: Database is closed` from `next()` once the handle is collected,
-the same as it does after an explicit `close()`.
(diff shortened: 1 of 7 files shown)


pmcfadin:
Probe complete, and it answered the question: a **brand-new PR at the identical head sha** (`8444d0c`) computed fresh also reports `mergeable: false / dirty`. So #3575's state is NOT a poisoned cache — GitHub genuinely cannot merge this content into current `main`, and the rebase is unavoidable. Local `git merge` succeeding in both directions was a red herring: git's ort strategy resolves `database.rs` (modified here) against `database/mod.rs` (main's #3576 split) as rename+modify, and GitHub's merge does not. Closing; the work continues on #3575.

Issue #3587, opened 2026-08-30
bindings/node: npm run test:coverage cannot pass — global functions threshold unmet (76.92% < 80%), pre-existing and unrun by any lane


pmcfadin:
Found while wiring #1465's node leak lane. **Not caused by #1465** — proof of pre-existence below. Not
fixed there because it would break 1:1:1:1 and is outside that issue's scope.

## Why this matters

`bindings/node/package.json` documents a `test:coverage` script. It **cannot pass**: the run finishes with
all tests green and then fails the `functions` coverage threshold.

```
$ cd bindings/node && CQLITE_DATASETS_ROOT=<corpus> npm run test:coverage
All files         |   83.51 |    77.58 |   76.92 |   84.74 |
Jest: "global" coverage threshold for functions (80%) not met: 76.92%
Test Suites: 28 passed, 28 total
Tests:       2 skipped, 508 passed, 510 total
$ echo $?
1
```

Thresholds in `bindings/node/jest.config.js` are `lines 80 / branches 65 / functions 80 / statements 80`.
Measured: **lines 83.51 (pass), branches 77.58 (pass), functions 76.92 (FAIL by 3.08 points),
statements 84.74 (pass)**. Only `functions` is short.

## Proof it is pre-existing (verified 2026-08-30)

Ran coverage under a **reconstructed pre-#1465 config** — `git show origin/main:bindings/node/jest.config.js`
with only `rootDir` added, and #1465's `__test__/leak-paths.test.js` excluded via
`--testPathIgnorePatterns` — in the same checkout with the same corpus:

```
All files         |   83.51 |    77.58 |   76.92 |   84.74 |
Jest: "global" coverage threshold for functions (80%) not met: 76.92%
Tests:       2 skipped, 504 passed, 506 total
```

**Byte-identical coverage numbers.** #1465 added 4 tests (504 -> 508 passed) and moved coverage by 0.00 in
every column, which is expected: coverage is collected from `lib/**/*.js` and adding tests can only raise it.
So the threshold was already unmet before #1465, and #1465's `projects` split does not perturb collection.

## Why it has been latent

Nothing runs it. `.github/workflows/node-ci.yml` runs `npm test` (line 450) and an explicit-file
`npx jest … --runInBand` subset (lines 95, 224); `node-release.yml` runs a single smoke file. No workflow and
no `scripts/agent-gate.sh` component invokes `test:coverage`, so the failure is invisible until someone runs
the documented command by hand.

[...]

Issue #3585, opened 2026-08-30
Native-retention leak oracle for the exception/abandoned-iterator paths (tracemalloc and heapUsed cannot see Rust allocations)


pmcfadin:
Parent: #1436. Follow-up from #1465 (exception-path leak budgets), filed at review request.

## Why this matters

#1465 added leak budgets on the two paths where leaks hide — a query that raises/rejects, and a
streaming iterator abandoned mid-stream — for both bindings. Both budgets are honest about a gap
they cannot close, and this issue is that gap:

- Python's primary instrument is `tracemalloc`, which traces **only the Python allocator**.
- Node's primary instrument is `heapUsed + external`, which covers **only V8-managed memory plus
  off-heap allocations V8 was told about** (`Buffer`/ArrayBuffer/napi external adjustments).

**Neither sees an ordinary Rust allocation.** A leaked `Vec<u8>`, a retained streaming channel, or an
un-dropped reader buffer on the native side stays completely flat in both instruments while process
memory climbs. Worse for confidence: #1465's RED controls planted *Python/JS* retentions, so they
establish sensitivity to Python/V8-visible retention **only** — they say nothing about native leaks.

#1465 shipped a loose secondary backstop on a native-visible metric (`ru_maxrss` in python,
`process.memoryUsage().rss` in node), sized to catch **gross** retention only:
~22 KiB/iteration (py, 1500 iterations, 32 MiB budget) and ~37 KiB/iteration (node, 2700
iterations, 96 MiB budget). A small per-iteration native leak — say 512 B/iteration, which is
2.8 MB per 5,000 error responses in a long-running service — is invisible to every instrument
currently in the tree.

## Current behavior (verified 2026-08-30)

- `bindings/python/tests/test_leak_paths.py` — tracemalloc budgets + `ru_maxrss` backstop. Module
  docstring names this gap explicitly and points here.
- `bindings/node/__test__/leak-paths.test.js` — `heapUsed + external` budgets + `rss` backstop. Same.
- Nothing anywhere measures native-allocator retention on these paths, and nothing calibrates a
  native retention control, so there is no evidence either way about small native leaks.

## Repro of the blind spot (either binding; python shown)

Plant a retention on the RUST side of an error/abandon path (e.g. push the per-call buffer into a
`static Mutex<Vec<_>>` in `bindings/python/src/database.rs`), rebuild with
`maturin develop --profile dev`, then:

```bash
[...]

Issue #14, opened 2025-07-25
🔧 HIGH: Establish CI/CD quality gates and enforcement


pmcfadin:
## **QUALITY ENFORCEMENT INFRASTRUCTURE**

**Impact**: Prevent future quality regressions by enforcing uncompromising quality standards in CI/CD pipeline.

## **Problem Statement**

**Current Issue**: No quality enforcement preventing broken code from being merged.

The project experienced compilation failures and 0% test execution due to lack of quality gates. This issue establishes automated quality enforcement to prevent regression.

### **🚨 Mandatory Quality Gates**

**EVERY PR MUST PASS ALL OF THESE:**

1. **🔧 Compilation Gate**
   ```yaml
   # .github/workflows/quality-gates.yml
   - name: Compilation Check
     run: |
       cargo check --workspace --all-targets --all-features
       # MUST return exit code 0 - NO EXCEPTIONS
   ```

2. **🧪 Testing Gate** 
   ```yaml
   - name: Test Execution
     run: |
       cargo test --workspace --all-features
       # MUST pass >80% of tests - NO EXCEPTIONS
   ```

3. **📊 Code Quality Gate**
   ```yaml
   - name: Code Quality Check
     run: |
       cargo clippy --workspace --all-targets -- -D warnings
       cargo fmt --check
       # MUST pass with no warnings - NO EXCEPTIONS
   ```

4. **🛡️ Security Gate**
   ```yaml
   - name: Security Audit
     run: |
       cargo audit
       # MUST pass security audit - NO EXCEPTIONS
   ```

### **📈 Performance Gates**

5. **⚡ Performance Regression Gate**
   ```yaml
   - name: Performance Benchmarks
     run: |
       cargo bench
       # MUST not regress >10% from baseline
   ```

#### **🔧 CI/CD Pipeline Implementation**

1. **✅ Quality Gate Workflow**
   ```yaml
   # File: .github/workflows/quality-gates.yml
   name: Quality Gates
   on: [push, pull_request]
   
   jobs:
     quality-gates:
       runs-on: ubuntu-latest
       steps:
         - uses: actions/checkout@v4
         - name: Setup Rust
           uses: actions-rs/toolchain@v1
           with:
             toolchain: stable
             components: rustfmt, clippy
         
         # MANDATORY: Compilation
         - name: Check Compilation
           run: cargo check --workspace --all-targets --all-features
         
         # MANDATORY: Testing  
         - name: Run Tests
[...]

PR #3594, opened 2026-08-30
test(#1465): exception-path + abandoned-iterator leak budgets (py tracemalloc, node heap+external), wired into the merge gate

### LINKED ISSUE #1465: Exception-path leak tests: tracemalloc (py) + jest --detectLeaks (node)
Parent: #1436

## Why this matters
Error paths are where leaks hide. When a query raises mid-conversion, or a streaming iterator is abandoned partway through, the native (Rust) side may have allocated buffers, channel state, or Python/JS objects that never get freed — and no test today would notice steady growth across repeated failures. A long-running Python service or Node server that hits errors in a loop would slowly bloat. This adds leak budgets on exactly those paths.

No leak test exists on either binding's exception/abandoned-iterator paths. Node's `bindings/node/jest.config.js` does not enable leak detection:
```js
module.exports = {
  testEnvironment: 'node',
  testMatch: ['**/__test__/**/*.test.js'],
  testTimeout: 30000,
  setupFilesAfterEnv: ['<rootDir>/__test__/setup.js'],
  coverageDirectory: 'coverage',
  coverageThreshold: { global: { lines: 80, branches: 65, functions: 80, statements: 80 } },
  collectCoverageFrom: ['lib/**/*.js', '!lib/**/*.d.ts'],
  verbose: true,
};
```
Python has `tracemalloc` available but no test uses it to bound growth across error iterations.

**Python — `bindings/python/tests/test_leak_paths.py`:**
1. `import tracemalloc, gc`. Open a DB once.
2. Warm up (run the loop body a few times first so one-time interned/cached allocations don't count), then `gc.collect(); tracemalloc.start()`, snapshot.
3. Loop N times (e.g. 500): (a) run a query that RAISES — a deliberately bad CQL string caught in `try/except cqlite.QueryError`; (b) start `db.execute_streaming(<wide table>)`, pull M rows, then abandon it mid-loop (`del it`) without exhausting.
4. `gc.collect()`, take a second snapshot, `compare_to(first, 'lineno')`, sum the size delta, and assert total growth is below a budget (e.g. `< 512 KiB` for 500 iterations — tune to a value comfortably above noise but far below a per-iteration leak). Assert per-iteration growth is bounded, not that it's exactly zero (interpreter noise exists).

**Node — leak scenarios + jest `--detectLeaks`:**
1. Add a dedicated test file `bindings/node/__test__/leak-paths.test.js` with: (a) a loop of `await expect(db.executeNative('BAD CQL')).rejects.toThrow()`; (b) a loop that starts `db.executeStreaming(...)`, reads a few rows via the async iterator, then `break`s (abandons) so `return()`→`close()` runs.
2. Enable leak detection for it. Read the current `bindings/node/jest.config.js` (quoted above) and either: add a project/config entry that runs `leak-paths.test.js` with `detectLeaks: true` (jest `--detectLeaks` / `detectLeaks` config), or add `detectOpenHandles: true` to catch leaked handles from abandoned iterators. Prefer a scoped config so `--detectLeaks` (which is strict and can false-positive on shared module state) applies only to the leak file, not the whole suite.
3. If `--detectLeaks` proves too flaky against the native module, fall back to a manual `process.memoryUsage().heapUsed`/RSS budget assertion across N iterations (same shape as the Python test), and document why in the test.

## Do NOT
- Do NOT assert zero growth — interpreter/V8 noise and one-time caches make that flaky. Warm up first, then assert growth is under a documented budget that a real per-iteration leak would blow past.
- Do NOT enable `--detectLeaks` globally if it destabilizes the existing suite — scope it to the leak test file.
- Do NOT change binding source to "fix a leak" unless the test actually finds one; the goal is the guard. If a real leak surfaces, fix it and note it in the PR.
- Reuse existing dataset fixtures/guards; do NOT invent a new dataset path.

## Tests you must add
- `bindings/python/tests/test_leak_paths.py::test_error_path_no_leak`, `::test_abandoned_stream_no_leak` — tracemalloc budget over N iterations.
  ```bash
  env CQLITE_DATASETS_ROOT=$PWD/test-data/datasets pytest bindings/python/tests/test_leak_paths.py -v
  ```
- `bindings/node/__test__/leak-paths.test.js` — error-path + abandoned-iterator scenarios under leak detection.
  ```bash
  cd bindings/node && npm run build && npx jest leak-paths --detectLeaks
  ```
- FAIL LOUDLY: dataset-backed; under strict mode present-but-unreadable data must fail (throw / `pytest.fail`), never skip.

## Definition of done
- [ ] Python tracemalloc leak-budget tests on the error path and abandoned-stream path assert bounded growth
- [ ] Node leak test exercises error + abandoned-iterator paths under `--detectLeaks` (or a documented RSS/heap budget fallback), scoped so the main suite stays stable

pmcfadin:
Closes #1465.

Adds exception-path and abandoned-iterator **leak budget** guards to both bindings, and wires the
Node lane into the one place that gates a merge. **No binding source changed and no leak was found** —
both paths sit at ~0 bytes/iteration. The deliverable is the guard, per the issue's "Do NOT".

## What landed

| File | |
|---|---|
| `bindings/python/tests/test_leak_paths.py` | `tracemalloc` budgets + a live-RSS backstop; error, abandoned-stream, and two contract tests |
| `bindings/node/__test__/leak-paths.test.js` | error + abandoned-async-iterator budgets (heap+`external`), RSS backstop, contract tests |
| `bindings/node/jest.config.js` | `projects` split — the leak lane is isolated; the `default` project keeps its prior behaviour exactly |
| `bindings/node/package.json` | `test:leaks` (the gate's line) + `test:leaks:handles` (human-only) |
| `scripts/agent-gate.sh` | the leak lane runs in `node-bindings` — **its only merge-gating execution** — with a three-state SUMMARY note |
| `scripts/tests/test_agent_gate_summary.sh` | self-test coverage for the note's state machine and the enrollment drift guard |

## Why the gate change is here

`.github/ci-gating-tiers.yml` lists `node-ci.yml` as an annotated **exemption** whose stated reason is
*"the merge-gating half is the local gate's `node-bindings` component"*. That component ran only
`npx jest write-readback-content`, and `node-ci.yml`'s `npm test` job is gated
`if: event != pull_request || label ci:bindings-full` — so a Node leak test would have executed
**nowhere at PR time**, first running post-merge on `main`. Adding a guard that nothing runs is the
defect this issue exists to prevent, so the lane is wired into the component the registry already
names. Python needed no wiring: the tests carry no `pytest` marker, so the gate's python tier selects
them.

## Budgets — measured, and asserted bounded rather than zero

| Path | Observed (clean) | Budget |
|---|---|---|
| py error | 32–4,340 B | 64 KiB |
| py abandoned stream | 16,631–19,623 warm / 67,841 cold | 256 KiB |
| py native RSS backstop | 4,096–970,752 B | 32 MiB |
| node error | min 16 B; upper median 16 (x4), 544, 9,856 (x2), 137,416 (x1) | 32 KiB min / 512 KiB median |
| node abandoned stream | min 56 B; upper median 56 (x7), 224 (x1) | 32 KiB min / 64 KiB median |
| node RSS backstop | −385,024..+21,164,032 B | 96 MiB |

Zero-growth asserts are forbidden by the issue and would be flaky; each loop warms up first so
one-time caches fall outside the window. **Detection floors, measured**: python trips at a
**64 B/iteration** retention (32 B passes) — an order of magnitude better than the first cut's
>512 B/iter; node's tracked floor is a **64-byte-per-iteration** retention (2.4x over budget). The
native-RSS floors are coarser — python's 24 KiB/iter is measured, node's ~37 KiB/iter is **arithmetic**
and labelled as such in-file.

**Node asserts two statistics, with a quorum.** The min alone is the most-favourable sample. The max
was tried and **rejected on measurement**: a clean error-path run reaches 428,064 B (13x budget), so it
would red 6/10 clean runs, and widening past that noise would exceed the RED signal and destroy
discrimination. So the verdict is the **upper median** of the non-negative passes, and it is refused
outright below a **quorum** (5 of 9) of valid passes — a verdict drawn from two surviving samples is not
a measurement of the run. Below-quorum and all-negative are distinct hard errors. The statistic has 8
direct unit tests inside the merge-gating lane, RED-verified against the previous statistic.

**One ceiling is weaker than the other, and the diff says so.** Ceilings are per path because a clean
error-path run reached 137,416 B. On that path clean noise (137,416) and the 256 B/iter signal (133,656)
**overlap** at this quantile, so the 512 KiB error median ceiling is a **gross-majority constraint only**
(~>=1.7 KB/iter) and the error path's real sensitivity comes from the 32 KiB **minimum** ceiling. The
stream ceiling was not widened and `BUDGET_BYTES` was not touched.

## Non-vacuity, and what each instrument does NOT see

A leak budget passes trivially if its loop body no-ops. Guarded three ways: non-zero counters asserted
in-test (exceptions raised, rows pulled, iterators abandoned, `resolved === 0`); typed error identity
(`err.code === 'QUERY'`, mirroring python's typed `except`); and RED controls run against the
**committed** files, where planted per-iteration retentions trip every budget.

**Stated plainly because it bounds the claim:** `tracemalloc` observes only the Python allocator and
`heapUsed + external` only V8 plus registered-external memory — **neither sees an ordinary Rust-side
allocation**. The RSS backstops are the native-visible half and are deliberately coarse. **#3585**
carries the real native-retention oracle (isolated-process RSS with a calibrated native control, or
native live-resource counters). This PR does not close the native-leak risk and should not be read as
doing so.

The abandonment paths differ by binding, and the difference is now stated rather than implied. Node
abandons an **in-flight** stream (`bufferSize: 2`; `rowsReceived` measured `[2,2,4,4,6]` with 44 of 50
rows outstanding, vs `[50,50,50,50,50]` at the default buffer, where the first `next()` drained
everything). Python's `rows_received` is **consumer-side**, so it establishes a **configuration**
property at the query→consumer boundary (the producer parked in `tx.send` holding the remainder);
`MAX_INFLIGHT_BATCH_ROWS = 1024` is independent of `buffer_size`, so the scan-level property is
**not** established there — the node lane carries it, because `nativeFetchedAtBreak` is measured.

## The gate note is a measurement, not a prediction

`node-bindings-leak-lane:` reports a closed vocabulary, fail-closed on an unrecognised state, and
`RAN` is written **only** after parsing jest's own JSON and affirming the **named** budget tests are
present and `passed`. An all-skipped run (jest exits 0) FAILs; a real budget failure reports
`ENTERED-FAILED`, not `NOT-REACHED`; an earlier step failing reports `NOT-REACHED`; the
missing-fixtures opt-out (#2078) skips **only** the leak lane, loudly, keyed on `_fixture_status`'s
OPTOUT state rather than the bare env var, and still runs `npm ci` + build + the content proof.

## One claim, stated precisely

The per-path ceiling transposition is **unrepresentable at the call site** — `assertUnderBudget(subject,
samples)` takes one frozen subject, so no positional same-type pair exists to swap. The swap is still
*representable inside the table literal*, where it is guarded by two exact identity pins rather than
being impossible. The reviewer checked that distinction specifically; it is recorded here because
"unrepresentable" on its own would overstate it.

## Review record

Six review rounds, `--lite` green each round. roborev found 12 findings across rounds and
coverage-reviewer 18; every blocker was fixed and re-verified, several by RED control. Notable
catches: the Node test originally never abandoned an active native stream; the first gate wiring
silently dropped the lane when the opt-out var was set with a corpus present, leaving **no**
`missing-fixtures:` line in the block; a comment audit corrected 17 factual claims and introduced one,
which the next round caught; and the python channel-capacity argument was refuted from core source and
rescoped.

### roborev — clean

```
repo: /data/lanes/lane-1465
branch: issue-1465-exception-path-leak-tests
base: origin/main
head-sha: 71ec37eaa34cde9bb629cb68fd54f3bc713b75e7
reviewed-sha: da80a479e529834d7ed2d59b6d011db90a852e79..71ec37eaa34cde9bb629cb68fd54f3bc713b75e7
assert-base: da80a479e529834d7ed2d59b6d011db90a852e79 (merge-base of origin/main and HEAD; origin/main tip c83d209220d7f414f8f1cf65b358864d068c3931)
job: 202
model: gpt-5.6-sol
census: 7 files, +2639/-28
tokens: input=2021638 cached=1739008 output=10300
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (7/7 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: NONE
roborev-exit: PASS
log: /tmp/cqlite-gates/1465/roborev-15.log
RESULT: PASS
```

`coverage-reviewer` (8 rounds): **MERGEABLE** — every blocker verified closed by its own measurements,
not by the implementer's claims.

### Gate of record — full `AGENT-GATE SUMMARY`, `RESULT: PASS`

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.QERH9q
commit: 71ec37e branch: issue-1465-exception-path-leak-tests dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-1465/test-data/schemas (checkout-relative)
tree-start: 71ec37eaa34c dirty: no digest: 0ee2f5b2c8d2
tree-end: 71ec37eaa34c dirty: no digest: 0ee2f5b2c8d2
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (5s)
clippy:            PASS (5s)
roborev-lints:     PASS (44s)
core-tests:        PASS (44s)
tombstones-scan:   PASS (0s)
scan-offload-guard: PASS (3s)
byte-budget-guard: PASS (13s)
arrow-parity-guard: PASS (0s)
memory-budget:     PASS (12s)
integration-tests: PASS (1s)
format-compat:     PASS (0s)
write-tests:       PASS (10s)
cli-tests:         PASS (5s)
flight-tests:      PASS (8s)
legacy-heuristics: PASS (10s)
feature-iso-parquet: PASS (1s)
feature-iso-delta-scan: PASS (0s)
python-bindings:   PASS (62s)
node-bindings:     PASS (121s)
binding-rust-tests: PASS (3s)
delivery-telemetry: PASS (0s)
oom-audit:         PASS (0s)
parity-report:     PASS (1s)
pub-surface:       PASS (0s)
tooling-tests:     PASS (928s)
minimal-build:     PASS (3s)
smoke:             PASS (3s)
logs: /tmp/agent-gate.QERH9q
summary-file: /tmp/cqlite-gates/1465/full-3.txt
RESULT: PASS
```

**Two full-gate runs, and the first one's red is recorded rather than quietly dropped.** The first full
gate on this exact tree (`/tmp/cqlite-gates/1465/full-2.txt`, same `tree-start` digest `0ee2f5b2c8d2`)
red with `tooling-tests: FAIL (134s)` on a **single structural assertion in an unrelated component** --
`the driver must print its selection as PARTIAL/FULL up front`
(`scripts/tests/test_ws0_report_guards.sh:1030`, subject `scripts/perf/ws0-baseline.sh`). That
assertion is now filed as **#3621**: a **non-reproducible in-gate-only flake**, established by 18 green
executions of the same script on the same tree against 1 red, plus ~3600 iterations of the failing
pipeline with zero failures. A content change, a `pipefail`/`grep -q` SIGPIPE race, and a transient
in-place edit of the subject were each positively excluded; both the test and its subject are
**byte-identical to `origin/main`**. This PR's diff touches no ws0 file, and the failing invocation is
**#14** of the component while this PR's own self-test is **#58** -- so it is excluded by execution
order, not by argument.

Because `tooling-tests` is **fail-fast** (`scripts/agent-gate.sh:10447-10456` does
`record_result ... ; return`), that one assertion truncated the component at 14 of 59 invocations
(4291 -> 1452 assertions; 938s -> 134s), which is why the first run never reached this PR's own
self-test. The re-run above executed the component in full (`tooling-tests: PASS (928s)`, 64
invocations, `test_agent_gate_summary.sh` confirmed present in the roster) with `ws0-report guards: all
123 checks passed`. Nothing was mis-certified at any point: the first run reported `FAIL`, and no merge
was attempted on it.

**Leak-guard execution, verified by count rather than by a green component** (the whole point of #1465):
- **node** -- `node-bindings-leak-lane: RAN`, the strongest state: written only after parsing jest's own
  JSON and affirming both named budget tests present exactly once and `passed`. Not `SKIP-OPTOUT`.
- **python** -- the gate's unfiltered `pytest bindings/python/tests -q` reported
  `576 passed, 61 skipped, 0 deselected`. Since `-q` names no test, the three were additionally
  confirmed **by name** against the gate's own built module: `test_error_path_no_leak PASSED`,
  `test_abandoned_stream_is_really_abandoned PASSED`, `test_abandoned_stream_no_leak PASSED`
  (3 passed, 0 skipped).

## Follow-ups

- **#3621** — the unrelated `tooling-tests` ws0 flake this PR surfaced (see the gate section).
- **#3585** — native-retention oracle (the instrument gap above), plus the Windows RSS-backstop gap.
- **#3587** — `npm run test:coverage` cannot pass (`functions` 76.92% < 80%); pre-existing, verified
  against a reconstructed `origin/main` config with these files excluded; nothing in CI runs it.
- **#3623** — batched nits, none blocking and all fail-closed today:
  - **No committed coverage of the gate helper's fail-closed arms.** All six were proven by reviewers'
    out-of-repo `npm`-shim harnesses; nothing under `scripts/tests/` shims `npm`, so a future edit to
    that block reds nothing. **This is a known gap and is deliberately not claimed as coverage here.**
  - **Three workflow legs still use bare `npx jest`** (`node-ci.yml:95,224`, `node-release.yml:132`) plus
    the `test:parity` script — the same class M1 closed inside the gate. Safe *only* because their
    explicit filters exclude the two gc-requiring files; widening a filter reopens it.
    (`node-ci.yml:450`'s `npm test` is already correct.)
  - N1's transposition fix is node-only: python keeps an adjacent same-type `(growth, budget)` pair.
    A transposition there fails **closed**, so it is latent.
  - Hardcoded literals in the refusal regexes (`only 4 of 9 … quorum of 5`) red if `MEASURE_PASSES`
    moves — fails closed, but should be derived.
  - `assertUnderBudget` does not validate subject shape; a malformed subject surfaces as a jest type
    error rather than a named refusal.
  - The #3220 per-table `DATASETS_AVAILABLE` scoping (`test_basic`-scoped while the lane reads
    `test_wide_rows`); fail-closed via the row-count asserts.

## Notes

- Dataset-backed and **fail-loudly**: present-but-unreadable data fails, never skips. Reuses existing
  fixtures/guards; no new dataset path.
- `--delta` is **not** eligible on this branch (the diff touches `jest.config.js`, `package.json`,
  `scripts/agent-gate.sh` and a self-test), so any post-gate change needs a full gate.
- Zero `file:line` citations remain in the touched files — a line number in a comment rots on the next
  edit, and ours did. Symbols are cited instead.


pmcfadin:

```
```

## Re-derive, do not replay

## Resolve ONCE

pmcfadin:

```
your merge-base with main:        25a160f7d68b
commits of main you are missing:  10
conflicts (git merge-tree):       1  ->  scripts/agent-gate.sh
```

## What is actually needed

```bash
git fetch origin
# resolve scripts/agent-gate.sh
git push --force-with-lease
```

pmcfadin:
🔧 **Lead: I deleted your `target/` (103G). Your queued gate will rebuild from scratch — that is a real cost and here is why I took it.**

```
box1 before:  29G free (90%)   <- falling ~8G/sweep
your gate:    QUEUED, not running (child = sleep; 0 rustc/cargo/cc with a lane-1465 cwd)
your target:  103G
box1 after:   131G free (54%)
```

**Verified immediately before deleting**, twice: your gate process's only child was `sleep` — it is waiting for box1's slot, which `lane-3514` holds — and **zero** `rustc`/`cargo`/`cc` processes had a `lane-1465` cwd. So nothing was writing that tree. Had a build been live I would not have touched it.

**Why not wait:** box1 had two gates queued against 29G and falling. A queued gate that *starts* on a box that close to full is likely to hit **ENOSPC mid-run** — which would waste the same rebuild time **and** take down `lane-3514`'s in-flight gate with it. Deleting now converts an uncertain loss of two gates into a certain, smaller loss of one rebuild.

**What it costs you:** your next full gate compiles cold. On an uncontended box that has been ~40 min of the ~2h total in today's runs; the rest is test execution, which is unaffected.

**What it does not cost you:** nothing in git. Your worktree, branch, commits and PR #3594 are untouched — `target/` is build output only. `git status` will be clean.

**Two notes on order of preference, so you know this was not the first thing I reached for.** I take reclaims in this order and only move down when the level above is empty: (1) a **finished** lane's tree — free, and that is what gave box3 101G an hour ago; (2) **caches** — I cleared box1's `sccache` last sweep, and it has only regrown to 514M so there was nothing left there; (3) a **queued** lane's target — where you landed; (4) a **running** gate's target — never.

**One thing I got wrong and corrected:** last sweep I recorded your target as unavailable because the watchdog showed `gating-skip [pid … cwd …/bindings/node]`, and I read that as an active build. It is not — `gating-skip` only means a gate process exists, and yours was queued. Reading the child process list is what distinguishes them.

Nothing else changes: #3594 is `clean` + `required: success`, and your remaining path is the gate of record → normal roborev round → arm.

Issue #3603, opened 2026-08-30
Thread-budget pins: poll_until_stable can sample before producers build their runtimes — false-negative under starvation (pre-existing on main)


pmcfadin:
## Problem

Both thread-budget pins can **pass while an amplification regression exists** — a false negative, in the opposite direction from the flake #3514/#3438 fixed.

`cqlite-core/tests/support/os_thread_budget.rs::poll_until_stable` treats ~200 ms of unchanged whole-process thread count (`STABLE_STREAK = 8` × `POLL_INTERVAL = 25 ms`) as evidence that the producer lifecycle has settled. It is not, because **thread creation and runtime construction are separated**:

- `cqlite-core/src/storage/write_engine/merge/producer_iter.rs:273-300` spawns each producer thread and returns immediately — no barrier of any kind. The thread is kernel-visible in `/proc/self/task` from that moment.
- The Tokio runtime whose worker pool the pins actually measure is built at `producer_iter.rs:391`, **inside** `producer_thread`. `from_readers.rs:273/:351` is byte-identical in shape.

So `KWayMerger::new` can return with every producer **created** and none having reached `Builder::build()`. If that state holds for 200 consecutive ms, the peak sample reads a pre-amplification count, the delta lands under bound, the fast path passes, and **the reap confirmation never runs** — so an amplified pool that appears afterwards is never observed.

## Precondition (narrower than it first looks)

The exposure is only the first 200 ms of stability, and *partial* construction is harmless — a rising count resets the streak and the poll continues, landing in the confirm branch. It therefore requires **all** `M` (#2316) or `C·M` (#2370) producers to remain unscheduled for 200 consecutive ms, which is a starvation-shaped condition. Direction is strictly **false-negative** (missed regression); it can never produce a false failure.

## This is pre-existing on `main`, not introduced by #3514

Recorded so nobody re-triages it as a regression from that PR:

- `origin/main:cqlite-core/tests/issue_2316_merge_thread_budget.rs:264-291` has the identical `poll_until_stable` with the same constants, and the identical call structure: `poll_until_stable(10s)` baseline → `KWayMerger::new` `:568` → `poll_until_stable(15s)` `:577` → `if delta > bound` `:594`.
- `origin/main:cqlite-core/tests/issue_2370_concurrent_merge_thread_budget.rs` is the same at `:274/:283/:289`, ending in a bare `delta <= bound` at `:319` — i.e. **main's #2370 is strictly worse**, having no confirm branch to skip.

[...]

Issue #3385, opened 2026-08-27
core-tests flake: issue_2316 merge thread-budget pin reds by exactly one thread under load, passes 3/3 standalone


pmcfadin:
`cqlite-core/tests/issue_2316_merge_thread_budget.rs::merge_bounds_producer_threads_to_o_m` reds in the FULL gate's `core-tests` component under load, and passes standalone. It is the #2316 O(M)-thread pin, and its bound has no slack for thread-reaping latency.

## Observed

Full gate, `core-tests: FAIL`, run-id `/tmp/agent-gate.SguYXB` (branch `issue-1697-config-source-of-truth`, PR #3381):

```
thread 'merge_bounds_producer_threads_to_o_m' panicked at
cqlite-core/tests/issue_2316_merge_thread_budget.rs:384:5:
merge over M=4 inputs added 16 OS threads over baseline (peak=18, baseline=2);
O(M) bound is 15 (= 3·M + 3). A delta scaling with num_cpus (num_cpus=16) means
a producer built a multi-threaded runtime — issue #2316 amplification.
```

**It missed by exactly one thread** (16 vs 15). The same test then passed **3/3 standalone** on the identical tree:

```
test merge_bounds_producer_threads_to_o_m ... ok    (6.78s)
test merge_bounds_producer_threads_to_o_m ... ok    (6.26s)
test merge_bounds_producer_threads_to_o_m ... ok    (5.10s)
```

## Why this is a measurement-slack problem, not #2316 regressing

The assert's own message reads a delta of 16 as "a producer built a multi-threaded runtime". That inference does not hold at the margin: on a 16-core box a genuine runtime amplification would land near `M + M·num_cpus` (~68), not `bound + 1`. A delta of exactly `bound + 1` is the signature of **thread-reaping latency**, not of an extra runtime — a worker that would normally have exited before the peak sample is still alive when the box is starved.

Conditions during the failure: the gate ran `nextest` at `--test-threads=16` while **four** lanes built concurrently on one 16-core host (`cpu-budget: ncpu=16 max-concurrency=1 cores-per-gate=16`), and `/data` was in the process of filling to 0G. Under that starvation, `peak` is inflated by scheduling, not by construction.

So the pin is measuring "peak OS threads observed" and attributing any overshoot to runtime construction, with `THREAD_SLACK = 3` absorbing none of the reaping jitter at M=4.

## Why it is not caused by the PR that surfaced it

Recorded so the next person does not re-derive it:

[...]

Issue #2370, opened 2026-07-13
Flight: no asserting N-concurrent do_get coverage — the #2316/#2361 escape class


pmcfadin:
**Confirmed hole (#2363 coverage-matrix audit).** No test in the repo issues ≥2 concurrent do_gets (or concurrent Trino queries) against one server with assertions able to observe a concurrency failure. Real N-concurrency exists only in the non-asserting field tool (`easy-db-lab-kits/trino-loadtest/driver.py` — `main()` unconditionally `return 0`); every test in `cqlite-flight/tests/do_get_transport_test.rs` drives ONE client stream; `issue_2316_merge_thread_budget.rs` bounds one merge's internal fan-out only; the #2361 `phase.active` gauge is validated only by synchronous 2-guard unit tests in `obs.rs`. Both recent field-only escapes (#2316 thread collapse, #2361 gauge/stall) were concurrency-only.

**Concrete fix — one integration test suite (cqlite-flight/tests/):**
- Spawn N≥8 simultaneous do_get streams (mix: full scan, point, LIMIT-k, one midstream drop) against one server over real transport.
- Assert: all complete within a generous timeout (hang detector); `cqlite.rpc.in_flight` AND `cqlite.rpc.phase.active` read ==N mid-flight (real read-back through the handler path — first ever) and settle to baseline after; process thread count bounded while ≥2 merges are simultaneously live (extends the #2316 budget to the aggregate case).
- Include a variant pinned to the uncompressed (non-stitching) scan path.

Evidence + matrix: docs/architecture/issue-2363-coverage-matrix-audit.md (Concurrency axis; holes CONC-1..4, QS-3). Related, not duplicated: #2362 (single-stream streaming asserts), #2316, #2361.

pmcfadin:
[...]

Issue #2316, opened 2026-07-10
flight merge: runtime amplification — every producer thread constructs a full multi-core tokio Runtime; one wide query spawns hundreds of OS threads


pmcfadin:
Promoted from epic #2313 WS3 (owner decision 2026-07-09) — the top-ranked failure mode in the throughput-saturation research (docs/architecture/issue-throughput-saturation-research.md); directly relevant to the round-6 loaded-cluster field runs.

## Defect
- Each `do_get` merge spawns **one OS producer thread per input SSTable** (`cqlite-core/src/storage/sstable/merge/mod.rs:456`).
- **Each producer thread constructs its own full multi-threaded `Runtime::new()`** — i.e. num_cpus worker threads each (`merge/mod.rs:519`).
- Net: one query over M SSTables costs ~`M + M·num_cpus` threads. A 10-SSTable table on a 32-core node ≈ **~320 threads for ONE query**; N concurrent queries ⇒ thousands. The kernel scheduler/futex overhead arrives well before disk bandwidth saturates.
- **Invisible today**: no thread-count gauge exists in the metrics surface; nothing in `cqlite_rpc_*` or the phase metrics shows it.

## Fix direction (design call at activation — Seam 1)
Candidates: (a) producer threads use a shared runtime `Handle` passed in (no runtime construction at all); (b) `current_thread` runtime per producer where async is genuinely needed; (c) refactor producers to tasks on the server's existing runtime. Constraint: merge output byte-parity unchanged (compaction-byte-parity + sstabledump goldens are the oracle); cancellation discipline (#2264) preserved.

## Acceptance criteria
- [ ] Thread cost per merge is bounded and documented (target: O(M) or better, no per-producer multi-core runtimes).
- [ ] A pinned regression test asserts the bound (thread count observed during a real multi-SSTable merge — must FAIL on today's code).
- [ ] A thread/blocking-pool gauge lands in the metrics surface (the observability gap that hid this; coordinates with #2313 WS2).
- [ ] Byte-parity + cancellation + full gate green; no wall-clock regression on the merge benches (#1494 suite once landed).

Routing: design-driven (merge architecture choice) → flow-activate + Seam 1 at pickup. Related: epic #2313 (WS3), #2230 (materialization — same file), #1668 rescope decision (merge-layer bounds).

pmcfadin:
[...]

PR #3604, opened 2026-08-30
fix(#3514): reap-confirmed acceptance for the concurrent thread-budget pin, and sweep #3438's batch

### LINKED ISSUE #3438: Thread-budget pins: apply #3385's reap confirmation to the concurrent test, and fix its vacuous CPU guard
Follow-up to #3385 (single-merge thread-budget pin de-flaked). Batches the roborev findings from that PR that fall **outside #3385's stated subject** — it names `issue_2316_merge_thread_budget.rs` specifically, and two of these live in a sibling file. Filed rather than widened, per the stay-in-scope rule; recording them so they are not rediscovered.

## 1. `issue_2370_concurrent_merge_thread_budget.rs` has the same latent flake (roborev job 63, Medium)

It asserts the noisy whole-process **peak** with no reap confirmation, while exercising the same `spawn_blocking` producer path #3385 measured. On a contended 16-core box the *correct* single-merge code produced peaks of 8 → 76 against a bound of 15, so this test is vulnerable to the identical starvation-induced false failure. It has simply not been reported redding yet.

Fix: apply #3385's reap confirmation (`ReapOutcome` + `poll_until_reaped`), scaled for `C·M`. Worth centralising the thread-observation oracle rather than copying it — a shared `tests/support/` module, since both files now need the same three pieces (direct `/proc/self/task` count, quiescence polling, affirmative acceptance).

## 2. Same file: the `num_cpus < 2` guard is wrong there too (roborev job 63, Low)

Independently verified against that file's constants: `NUM_INPUTS = 3`, `NUM_MERGERS = 2`, `PER_INPUT = 3`, `THREAD_SLACK = 6`, so `bound = 3·3·2 + 6 = 24`. Pre-change cost is `C·M·(1 + num_cpus) = 6·(1 + cpus)`, which exceeds 24 only once `cpus >= 4`.

So on a 2- or 3-core host that test **passes vacuously against the very defect it pins** — the same bug #3385 fixed in the single-merge test, where the threshold is now derived from the constants via `min_cpus_for_amplification()` instead of hardcoded.

## 3. Transient blocking-pool growth is not pinned anywhere (roborev jobs 59 + 63, Medium)

#3385 asserts the **persistent** thread count and reports the peak as diagnostic only. That was deliberate and is documented in the file with its measurement: the `spawn_blocking` pool is demand-driven, correct code peaked 8 → 76 under load, and any peak allowance wide enough not to flake is too wide to mean anything.

The residual is real, though: a regression that transiently constructs and drops per-core runtimes, or that grows the blocking pool without bound, would not fail that test. Catching it needs a different oracle — one that observes runtime-worker creation or pool state **directly**, rather than inferring it from a whole-process peak. Note the process-wide peak is the wrong instrument for this by construction, so this is a new oracle, not a tuning change.

## Acceptance criteria

- [ ] `issue_2370_concurrent_merge_thread_budget.rs` uses reap-confirmed acceptance, RED-verified by flipping its producers to `new_multi_thread()`
- [ ] Its CPU guard is derived from its own constants (`>= 4` for today's values), not hardcoded
- [ ] The thread-observation oracle is shared between the two tests rather than duplicated
- [ ] A decision recorded on item 3: either a direct pool/worker-creation oracle, or an explicit accepted-residual note

Refs: #3385, #2316, #2370.

### LINKED ISSUE #3514: core-tests flake: issue_2370 concurrent merge thread-budget pin reds 2/3 standalone at moderate load (sibling #3385 missed)
## Problem

`cqlite-core/tests/issue_2370_concurrent_merge_thread_budget.rs::concurrent_merges_bound_aggregate_threads_to_o_c_m`
reds under load, and unlike its sibling it reds **even at moderate load**, standalone.

This is the same defect **class** as #3385 (closed) — a thread-budget pin whose bound has no slack for
thread-reaping latency — but a **different test file**. #3385 fixed
`issue_2316_merge_thread_budget.rs`; this sibling kept the unslacked shape. Sweeping the class rather
than the instance would have caught it.

## Measured

Full gate, `core-tests: FAIL`, run-id `/tmp/agent-gate.6NSXWm` (branch
`issue-3400-zero-tests-guard-inert-nightly`, PR #3501), 4817 passed / **1** failed:

```
concurrent_merges_bound_aggregate_threads_to_o_c_m panicked at
cqlite-core/tests/issue_2370_concurrent_merge_thread_budget.rs:318:5:
C=2 concurrent merges over M=3 inputs each added 41 OS threads over baseline
(peak=43, baseline=2); O(C·M) bound is 24 (= 3·M·C + 6).
```

**The overshoot scales with host load**, which is the signature of scheduling/reaping latency rather
than of runtime construction:

| host load (1-min) | result | threads over baseline | bound |
|---|---|---|---|
| ~102 (in-gate) | FAIL | 41 (peak 43) | 24 |
| 31.6 | **ok** | — | 24 |
| 29.8 | FAIL | 31 (peak 33) | 24 |
| 28.8 | FAIL | 29 (peak 31) | 24 |

So it is **2/3 failing standalone at load ~30** on a 16-core box — materially flakier than #3385's
case, which passed 3/3 standalone and missed by exactly one thread.

## Why the assert's own diagnosis is misleading at the margin

The panic message attributes any overshoot to "a producer built a multi-threaded runtime — issue #2316
amplification". #3385 established why that inference fails near the bound: a genuine amplification on a
16-core host would land near `C·M·num_cpus` (~96 here), not at `bound + 5..17`. A delta that *tracks
host load* is reaping latency, not construction. `THREAD_SLACK` (the `+6` at C=2, M=3) absorbs none of
that jitter.

## Why it is not caused by the PR that surfaced it

PR #3501 changes **zero compiled inputs** —
`git diff --name-only origin/main...HEAD -- cqlite-core/ cqlite-cli/ Cargo.toml Cargo.lock` is **empty**
(the diff is `CLAUDE.md`, `scripts/agent-gate.sh`, and two `scripts/tests/*.sh`). The compiled artifact
is therefore identical to `origin/main`'s by construction, so this failure is pre-existing on `main`.
The same gate's earlier run at anchor `30db5136a` had `core-tests: PASS (851s)` on the same sources.

## Suggested direction

Give the pin slack that reflects what it can actually measure, or make it measure something robust to
scheduling — e.g. sample peak threads with the box quiesced, assert on a load-normalised bound, or gate
the assert behind a low-load precondition and `#[ignore]` it otherwise. Whatever the fix, it should be
applied to **both** members of this class so the next sibling does not red the same way.

## Related

#3385 (same class, sibling test, closed), #2370 (the property being pinned), #2316 (the amplification
the bound derives from). Surfaced by `flow-closer` on #3400 / PR #3501.

pmcfadin:
Closes #3514
Closes #3438

De-flakes the two OS-thread-budget pins by giving the concurrent one the **reap-confirmed acceptance** that #3385 shipped for its sibling, and sweeps the rest of #3438's batch in the same change. Test-only: **zero source changes** (`git diff origin/main...HEAD -- cqlite-core/src/` is empty).

## Why this shape, and what was deliberately NOT done

The owner established the mechanism on #3514's thread across seven comments — including three of their own explanations retracted after testing — and the settled diagnosis is **delayed thread reaping under CPU starvation**, not runtime construction. The controlled evidence: the identical tree passes 6/6 standalone on a quiet box and passes full gates on two other hosts at tree digest `373be7b0dae8`, while redding at PSI stall 43%.

So, explicitly:

- **The bound is NOT widened.** The field deltas (31/37/39/43 over a bound of 24) measure delayed reaping under starvation, not concurrency. The steady-state peak genuinely fits under 24. Re-deriving the bound from starvation artifacts would trade the pin's entire diagnostic power for green.
- **No load/PSI precondition gates the assert.** A skip-above-threshold guard would leave the pin unrun precisely on the hosts where it reds — which is the same vacuous-guard failure shape as AC2 below. PSI is taken as a **diagnostic only**: read by no branch, never a pass/fail input.

## #3438's acceptance criteria

- **AC1 — reap-confirmed acceptance, scaled for `C·M`.** An over-bound peak no longer condemns. `poll_until_reaped(baseline + bound, REAP_QUIESCENCE_SPAN = 12s, reap_confirm_timeout(C·M) = 90s)` polls `/proc/self/task` and accepts **only** on an affirmative `ReapOutcome::Drained { settled }` with `settled <= bound` held for a positive 12s span; `Unconfirmed` fails. Budget derived from #2316's shipped 60s / 4 producers = 15s per producer (one 10s blocking-pool keep-alive + 5s late-finish slack), linear because each late reap *resets* the span.
- **AC2 — CPU guard derived from the file's own constants.** The hardcoded `num_cpus < 2` is replaced by `min_cpus_for_amplification(producers, bound)`. Derivation: `p(1+c) > B ⟺ c >= floor(B/p)`; at `p = C·M = 6, B = 24` that is **4**, matching the owner's figure, with `cost == bound` correctly not counting as exceeding. The old guard let the test **pass vacuously against the very defect it pins** on a 2–3 core host — measured, not argued: with the regression live and `min_cpus` re-pinned to 2, a 3-cpu `taskset` run **passed** at `confirmed_delta=24`, exactly the predicted `6·(1+3)`.
- **AC3 — shared oracle.** The three pieces (direct `/proc/self/task` count, quiescence polling, affirmative acceptance) plus the derivation and PSI live once in `cqlite-core/tests/support/os_thread_budget.rs`. No `/proc/self/task` copy remains in either test. The #2316 refactor is **timing-preserving but not a pure extraction**, and the distinction is worth stating precisely: its *timing* constants are unchanged (`poll_until_stable(10s)`/`(15s)`, streak 8, 25ms, quiescence 12s, and `reap_confirm_timeout(4)` reproduces its shipped 60s exactly), but it also picks up the AC2 fixes — vacuity-guard input `NUM_INPUTS` → discovered `m`, the guard re-ordered after `build_inputs`, the fixed `REAP_CONFIRM_TIMEOUT` replaced by `reap_confirm_timeout(m)`, and the new PSI reads. None drifts observably at `m = 4` or any `m >= 4`; the semantics are slightly *more* conservative. An earlier draft of this body called it a pure extraction, which the C audit correctly flagged as an overclaim.
- **AC4 — decision on transient blocking-pool growth: ACCEPTED RESIDUAL,** recorded in the shared module naming the three uncovered shapes (construct-and-drop per-core runtime; thread-creation *rate*; `max_blocking_threads` sizing), why the whole-process peak is the wrong instrument **by construction**, and the concrete alternative (`RuntimeMetrics::num_blocking_threads` from inside each producer's runtime, or a per-name high-water census correlated with runtime construction). Not a widened peak assert pretending to cover it.

## RED-VERIFICATION — the evidence that matters

A green run on this pin evidences nothing (`--only core-tests` reproduces at ~17%/run). Both pins were driven RED by flipping producers to `new_multi_thread()`, then reverted:

| pin | delta / bound | outcome |
|---|---|---|
| `issue_2316` | `delta=72` / 15 | `UNCONFIRMED after 60s` → FAIL in 66.0s |
| `issue_2370` | `delta=108` / 24 | `UNCONFIRMED after 90s` → FAIL in 94.6s |

Guard boundary re-checked post-refactor under `taskset` with the amplification live: **skips** at 2 cpus (cost 18 <= 24) and 3 cpus (24 = 24), and **FAILs at exactly 4** (`delta=47`).

## Review-round findings (all fixed)

roborev round 1 returned `findings: NONE`; `rust-reviewer` returned five, **two of which the lead re-triaged UP to blocker**:

1. **BLOCKER** — the PSI diagnostic rendered a backwards counter (cgroup/PSI reset) as `0.0%`, violating the "an unmeasurable value is not the value zero" rule stated in the same file, and misclassifying the exact red it exists to classify. Now `UNMEASURED (counter went BACKWARDS: …)` naming both readings. **Mutation-verified**: restoring `saturating_sub` reds exactly that one test.
2. **BLOCKER** — the vacuity guard derived from the constant `NUM_INPUTS` while the bound derived from discovered `m = inputs.len()` (`build_inputs` only asserts `>=`), so one extra `nb-` SSTable would make the pin **skip on a 3-core host where the amplification is observable** — a second latent instance of the very class AC2 fixes, i.e. how #3385 produced #3514. Now `m` is bound **once** per file and every size-dependent quantity derives from it, making divergence *unrepresentable* rather than merely forbidden.
3. PSI wall denominator closed before the closing `total=` read (could print >100%); the reads are now ordered so the denominator interval is a superset.
4. nextest hung-kill headroom was claimed but unquantified — a genuine FAIL could have presented as a hung-test kill with none of the assert diagnostics. Did **both**: quantified (worst case ~124.9s vs the 240s `terminate-after=4` kill = **115.1s margin**; `build + 2 drains` measured 3.9s at 16 cores, 4.9s at 4) **and** enrolled both binaries in the existing nextest `timing` group (`max-threads = 1`). Retries stay at the profile default **0**, deliberately and stated in-file: a real amplification must fail deterministically, and retrying could only turn a regression into flap.
**roborev round 2 then returned one Low finding, which the lead escalated as a class sweep:** the PSI counter was sampled **before** the `Instant` anchor, so the numerator's interval began earlier than the denominator's and the ratio could exceed 100%. This was the **mirror image of item 3 above** — same window, other end, one round later — i.e. the "fix the reported instance, not the invariant" shape that made #3385 produce #3514. Root cause: the window was built as a struct literal, and **field initialisers evaluate in written order**, so the anchor was ordered after the read. Fixed at three layers: the open now takes the wall anchor first as two ordered statements (not a literal); the close reads the counter before `elapsed()` (item 3); and `describe_stall` **rejects** an inverted window where the ratio is actually formed — so an ordering regression at either end can only degrade to `UNMEASURED`. Rejecting, not capping: a clamped `100.0%` would be a plausible number standing in for an impossible measurement, blocker 1's mistake one field over.

The module was then swept for every counter-paired-with-time site, with negative results recorded: `report()`'s close side correct (re-verified); `poll_until_reaped`'s `unchanged_since` **safe and now documented as such** — captured *after* the observation, so `elapsed()` under-reports and acceptance gets *harder* (fail-closed, the opposite direction), noted so nobody "fixes" it the unsafe way; `poll_until_stable` forms no ratio; `reap_confirm_timeout` / `read_cpu_pressure_some_total` pure or single-read; `pressure_at_confirm` re-reports from the same anchor and inherits the fixed ordering.

**The test was the more important half, because the round-1 test passed while the bug was live** — it pinned the arithmetic, not the ordering. The new cases (`inverted_window_is_rejected_not_capped`, `no_input_ever_yields_a_percentage_above_one_hundred` over a 6×8 lattice, `fully_stalled_window_is_exactly_one_hundred_percent`) **parse the percentage back out and check `0..=100` numerically**, so a formatting change cannot defeat them. Mutation-verified twice — cap-instead-of-reject, and no-rejection-at-all — each yielding `7 passed; 2 failed` with exactly the two new cases red. That second mutation demonstrates the round-1 shape *could not* have caught this: the hair-over input `1000001us over 1000000us` **formats as `100.0%`**, so a substring check on `"100.1%"` passes it; only the numeric parse sees it.

5. `poll_until_stable`'s timeout remains a bare panic — taken as an **explicit documented residual** (200ms bar inside a 10–20s budget = 50–100× headroom; it guards lifecycle settling rather than a starvation-sensitive value, so a "confirmation" there is just a longer timeout). Its panic now states **"NOT a thread-budget regression: no bound was compared"** plus the PSI reading, so it can never be misread as an amplification, and the doc names what would settle it.

## roborev round 3: a Medium false-negative, triaged out of scope to #3603

roborev then found something more serious than the first two rounds and in the opposite direction: `poll_until_stable`'s ~200ms stable streak can complete **before** producers build the Tokio runtimes whose worker pools the pins measure — thread creation (`producer_iter.rs:273-300`, kernel-visible at spawn) is separated from runtime construction (`:391`, inside the thread). So under starvation the peak sample can read a pre-amplification count, pass under bound, and **skip the reap confirmation entirely** — a pin that passes while the regression exists.

**CONFIRMED, and deliberately not fixed here.** The reasons, in the order that decided it:

1. **It is verbatim pre-existing on `main`,** and `main`'s #2370 is *strictly worse*: `origin/main:issue_2316_merge_thread_budget.rs:264-291` has the identical `poll_until_stable` and call structure (`:568/:577/:594`), and `origin/main:issue_2370:274/:283/:289` ends in a bare `delta <= bound` at `:319` with **no confirm branch to skip at all**. This PR neither introduced nor widened it — it centralized and documented the oracle, which is why the finding became visible.
2. **Narrower than it first reads:** partial construction is harmless (a rising count resets the streak and the poll lands in the confirm branch), so it needs **all** `M`/`C·M` producers unscheduled for 200 consecutive ms. Direction is **false-negative only**; it can never produce a false failure.
3. **No sound test-only fix exists.** A 2s span only narrows the window ~10× (a rarer defect, not a fix); a lifetime high-water sampler cannot work because the confirm can only run while producers are blocked, and bounding the lifetime peak is what #3385 measured as unusable (correct code 8 → 76 vs amplified ~96 — overlapping distributions); `>= baseline + 2·producers` infers tokio's internals, so it is a **no-heuristics violation** that would degrade into a hang-then-panic flake on a tokio bump; and `producer_gauge` is unusable as readiness because `spawned()` increments **before** the spawn (`producer_iter.rs:232`, deliberately, for gauge-pairing correctness).
4. **The src fix has an unresolved design question that should not be guessed at in this PR:** whether firing readiness after `Builder::build()` returns is sufficient depends on whether tokio spawns multi-thread workers eagerly in `build()` or lazily on first poll.

So it is **filed as #3603** with the full analysis — citations, the rejected candidates and their reasons, the concrete src change (a ready-`AtomicUsize` in `producer_gauge.rs`, private mod at `merge/mod.rs:475`), the tokio question as a must-answer-first, and acceptance criteria requiring a re-RED of both pins — and recorded as an `ACCEPTED RESIDUAL` in the shared module beside AC4's, since it is the **same instrument problem**: a whole-process count cannot attribute a thread, nor see one that appeared between two polls.

## Limits, stated rather than laundered

- **The absorbing direction IS now field-verified, unplanned.** This was written up as an accepted gap — the starvation deltas would not reproduce on the delivery box (PSI 0.0–2.6% throughout), leaving the confirm path's ability to absorb real starvation resting only on the RED-flip control. Then a loaded run at `afef8f2` caught it live: #2370's fast path read **`delta=25` against `bound=24`** — the exact one-thread #3385 overshoot — at **46.1% PSI stall**, and the reap confirmation **absorbed it to a PASS** (26.4s vs 4.5s quiet). So the flake reproduced, the new diagnostic classified it on sight as a starved host rather than a regression, and the mechanism converted it to a pass **without a widened bound**. That is the end-to-end evidence this change existed to produce, and it arrived from the field rather than from a contrived harness.
- **The owner's PSI series is n=1 on the failing side.** No threshold is encoded anywhere; ~40% stall is the shape of the answer, not a number the code depends on.
- **A known false-negative ships with this change** (#3603, pre-existing on `main`, starvation-only, documented in-module). Stated here rather than left to be discovered: this PR makes the pins strictly better than `main`, and does not make them sound in every condition.
- `max-threads = 1` bounds contention **within** the `timing` group only, so it reduces the starvation source rather than removing it — which is why the headroom arithmetic is still load-bearing. Said so in the config.

## Review record

Review-first per #2086 — all four rounds ran on the lite-green diff **before** the full gate, so the one gate of record certifies already-reviewed code.

| round | head | roborev | rust-reviewer | outcome |
|---|---|---|---|---|
| 1 | `fb33f3180` | job 181 **PASS**, `findings: NONE` | 0 blockers / 5 nits | lead re-triaged **2 nits → blocker** (PSI `0.0%`; guard/bound divergence) |
| 2 | `edb90880c` | job 183 **FAIL**, 1 Low | — | escalated as a **class sweep** (mirror of r1 nit 3) |
| 3 | `4cac679a0` | job 184 **FAIL**, 1 Medium | — | CONFIRMED false-negative, **pre-existing on `main`** → triaged to #3603 + in-module residual |
| 4 | `6619025bd` | job 187 **PASS**, `findings: NONE` | — | clean; C intent audit then PASSed here (4/4 ACs) |
| 5 | `631b269ac` | job 188 **FAIL**, 1 Low | — | `gate-ops.md` retry-policy doc drift → fixed |
| 6 | `4ac3bfdcb` | job 189 **FAIL**, 1 Low | — | deadline/sampling race — **suggested remedy rejected**, see below |
| 7 | `986daac66` | job 191 **PASS**, `findings: NONE` | — | clean — **certified head** |

Token accounting for every round is far above the vacuous-review baseline (~18.7k input / 0 cached) — r4 alone was 735k input / 622k cached — and `prompt-content:` reported 5/5 census code paths present each time, so none of these verdicts is a vacuous pass.

Both lead re-triages in round 1 were upgrades, recorded because the reasoning is the reusable part: a diagnostic that renders an unmeasurable value as `0.0%` misreports the exact condition it exists to classify, and a guard derived from a different quantity than the bound it protects is the same vacuity class the PR is fixing. Round 2's Low was escalated for the same reason — it was the mirror of an already-fixed nit, and fixing reported instances rather than invariants is precisely how #3385 produced #3514.

### Two suggested remedies were deliberately not applied

Both findings were worth having; both prescriptions would have made the code worse, so they were resolved differently and the reasoning is recorded in-code so neither gets "fixed" back:

1. **"Cap the stall percentage at 100%"** (round 1) → **rejected in favour of rejecting.** A clamped `100.0%` is a plausible-looking number standing in for an impossible measurement — the same error as rendering an unmeasurable counter as `0.0%`, one field over. An inverted window now reports `UNMEASURED`.
2. **"Return `Unconfirmed` once the deadline passes, before evaluating `Drained`"** (round 6) → **rejected as a false-failure generator.** That converts a run which observed an in-budget count *held for the full `min_span`* into a FAIL, purely because the final sample crossed the deadline — under starvation, which is when crossing is likeliest. It is the very defect class this PR removes, and it contradicts #3385's job-60 asymmetry (**patient before failing**) that this module documents. A late `Drained` is a *true* statement; a deadline-triggered `Unconfirmed` over a genuine drain is a false one.

   The reviewer's *observation* was valid, though: the contract text over-promised a bound the code deliberately does not honour on the accepting side. So `poll_until_reaped` now carries a **"The deadline bounds CONDEMNATION only — not acceptance"** section stating both halves, citing job 60, and saying explicitly *"do NOT 'fix' the missing deadline re-check"* with the false-failure argument — because without that, this finding regenerates every review. The affected test was made deterministic **by construction** (`min_span = Duration::MAX`, which no `elapsed()` can reach) rather than resting on a ~30s deschedule being implausible.

Verified independently for round 6: `Drained` is the **only** post-deadline return path (the trailing `sleep` overshoot is caught by the next loop-condition check), so the asymmetry is exactly as documented.

## Certification

- `AGENT-GATE LITE SUMMARY` **PASS** at each round (final: `6619025bd`, run-id `/tmp/agent-gate.s5RR3E`, `tree-integrity: PASS`)
- roborev **PASS** at head `6619025bd` (job 187, sanctioned wrapper, `--agent codex --model gpt-5.6-sol`)
- Full `AGENT-GATE SUMMARY` — the gate of record — runs once in `flow-closer` before merge; recorded below on completion.


pmcfadin:
## Full agent gate of record — RESULT: PASS

Certified head `986daac665658a8dbfda6634c40c95702ba65482`. This is the **first and only** full gate of record for this issue; the run-id below is the sole certifiable run. Diff is test/config/docs only — **zero `cqlite-core/src/` paths** — so no compiled library input changed.

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.KPwZy4
commit: 986daac branch: issue-3514-core-tests-thread-budget-flake dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-3514/test-data/schemas (checkout-relative)
tree-start: 986daac66565 dirty: no digest: 1149b896cb70
tree-end: 986daac66565 dirty: no digest: 1149b896cb70
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (5s)
clippy:            PASS (1s)
roborev-lints:     PASS (45s)
core-tests:        PASS (234s)
tombstones-scan:   PASS (164s)
scan-offload-guard: PASS (312s)
work-counters-guard: PASS (165s)
byte-budget-guard: PASS (12s)
arrow-parity-guard: PASS (470s)
memory-budget:     PASS (598s)
integration-tests: PASS (256s)
format-compat:     PASS (165s)
write-tests:       PASS (277s)
cli-tests:         PASS (428s)
flight-query-semantics-oracle: PASS (232s)
flight-tests:      PASS (7s)
legacy-heuristics: PASS (251s)
feature-iso-parquet: PASS (810s)
feature-iso-delta-scan: PASS (804s)
python-bindings:   PASS (982s)
node-bindings:     PASS (839s)
binding-rust-tests: PASS (1052s)
delivery-telemetry: PASS (1s)
oom-audit:         PASS (11s)
parity-report:     PASS (245s)
operator-metrics-doc: PASS (110s)
binding-unwind-profile: PASS (3s)
pub-surface:       PASS (0s)
tooling-tests:     PASS (1005s)
minimal-build:     PASS (281s)
smoke:             PASS (539s)
logs: /tmp/agent-gate.KPwZy4
summary-file: /tmp/cqlite-gates/3514/full-gate-r3.txt
RESULT: PASS
```

### Reader notes

- `tree-integrity: PASS` with `tree-start == tree-end == 986daac66565`, `dirty: no` — no mid-run mutation, so the block is a certification.
- **0 FAIL, 0 SKIP** across all components: nothing silently did not run.
- `core-tests: PASS (234s)` — the component whose flake is the subject of this PR.
- `binding-rust-tests: PASS (1052s)` — the #3522 never-SKIP property holds.
- `datasets: 144 Data.db files` — a real corpus, so dataset-dependent tests could not pass vacuously on 0 rows.
- Duration context only, **not a finding**: `accelerators:` reports `sccache-health=warn` (`ok` on the lite runs), which alongside sibling-lane contention on a single-slot box explains the ~78 min wall time. It does not affect the verdict.
- The #3042 docs-only waiver deliberately does **not** apply here: `.config/nextest.toml` is a config input, so any red would have been presumed ours. There were none.

PR #3467, opened 2026-08-29
feat(#3393): per-lane claim refs — one ref per lane, retracting #1930's one-worker-per-machine invariant

pmcfadin:
Implements the **option A** ruling on #3393: per-lane identity, one claim ref each.

Fourth slice of #3393 — #3407 shipped AC2 (the OOM refusal), #3429 shipped AC4 (the diagnostic order), #3431 mirrored it to the website. This is the AC3 mechanism.

## Design authority

Owner ruling on #3393 at **14:49Z 2026-08-28**: *"option A, per-lane identity with one claim ref each. Retract the one-worker-per-machine text (#1930) as part of it; the 4-lanes-per-box practice is the standing model. Scope includes the reap predicate input and the project-board-sync reap-claims namespace; the CI reaper's remote-pid limit is accepted as documented."*

Scope was then narrowed by the lead's **split ruling (2026-08-29)** — see *Deliberately not here* below.

## The problem

`refs/machine-claims/<machine>` was **one ref per machine** (#2655), justified by #1930's "one worker per machine". The fleet has been running **4 lanes per box** on standing instruction. The ref is force-updated every supervisor iteration, so after one lane died a surviving sibling's next stamp overwrote its PID — a live sibling did not merely hide a dead lane, it **masked** it. Two of #3393's three silent lane deaths were on one host: exactly the case that collapsed.

## What changed

| surface | change |
|---|---|
| `claim-heartbeat.sh` | `stamp` writes `refs/lane-claims/<machine>/<issue>`; `list-claims`, `should-reap`, `reap`, `dead-lanes` all lane-aware |
| `worker-supervisor.sh` | stamps and clears **per lane**, with an invocation-unique placeholder before the issue is known |
| `project-board-sync.yml` | `reap-claims` iterates the new namespace; the board flip reads the issue from the ref **path**, not a message parse; deletes take a CAS lease |
| doctrine | **#1930 retracted** across CLAUDE.md, the fleet runbook, the website page, `flow-lead.md`, `flow-board/SKILL.md` |
| tests | 95 → 132 in `test_claim_heartbeat.sh`, plus `worker-supervisor` coverage |

### The namespace change was forced, not chosen

A sub-path under the old namespace is impossible — git refuses a ref that is both a file and a directory:

```
$ git update-ref refs/machine-claims/box3/3367 <sha>
fatal: cannot lock ref '...box3/3367': 'refs/machine-claims/box3' exists
```

[...]

PR #3616, opened 2026-08-30
fix(#3559): classify --slice from the issue timeline at mergedAt, not from current closed_at

### LINKED ISSUE #3559: fix(telemetry): classify --slice from the issue timeline at mergedAt, not from current closed_at

pmcfadin:
Closes #3559

Classify a `--slice` delivery by replaying the issue TIMELINE to the PR's `mergedAt`, instead of
inferring it from the issue's CURRENT state. That makes a genuine slice stampable after its issue
closes — the hole #3550 left open, and the one currently blocking the three owed #3393 records
(#3407, #3429, #3467).

## The rule, and why it is a conjunction

    slice  <=>  (the issue was OPEN at mergedAt)  AND  (this PR closes NOTHING)

**Open-at-mergedAt alone is not sufficient and never becomes sufficient.** GitHub records an
auto-close AFTER the merge, so an ordinary completed delivery whose PR declares `Closes #N` was
*also* literally open at `mergedAt`. Dropping the second operand would file a false slice record for
essentially every ordinary delivery — the same false-PASS-in-the-common-direction shape as the
`closedAt <= mergedAt` guard #3550 removed. So `pr_closes_this_issue` is **not** a propagation-window
stopgap made redundant by the replay: it is the only signal separating "open at mergedAt because the
issue is never closing" from "open at mergedAt because the close lands five seconds later", and its
scope is now *widened* to apply whether or not `closed_at` is currently null.

`_assert_never_closed` is **removed**, not layered — the timeline answers the real question and
`stateReason` was only ever a proxy for "has this ever been closed" (#3559's own design note; two
mechanisms approximating one fact is the defect).

## What lands

- `_issue_state_at` replays `gh api .../issues/<N>/timeline --paginate`: `closed`/`reopened` events
  strictly before `mergedAt`, last one decides. Surfaced through the existing `--from-json` seam as
  `issue_open_at_merge`, so all tests stay **network-free**.
- Every non-affirmative state is its own named refusal — failed `gh` call, truncated/unparseable
  reply, non-array page, non-object entry, unreadable `event` (it could BE the deciding `closed`),
  deciding event with no usable `created_at`, non-canonical issue URL, unusable `mergedAt`.
- An event in the **same second** as `mergedAt` is a refusal, not a "before" — see roborev below.
- `delivery-telemetry` gate lane: module set **derived from source** (`test_delivery_telemetry*.py`),
  so a new sibling suite is executed with no gate edit; empty glob is a FAIL naming the derivation;
  plus a per-module zero-test/unmeasurable-count guard.
- Doctrine updated in the same change: CLAUDE.md, `pm-operating-loop.md`, the website
  `delivery-pipeline` page, the schema, and #3550's interim "needs #3559" refusal messages.

## Verification

**Local gate of record: NOT YET RUN — `flow-closer` runs it once pre-merge.** Fix-round certification
only:

```
MODE: lite (FAST ITERATION — NOT the gate of record)
commit: 4a0618b branch: issue-3559-slice-from-timeline-at-mergedat dirty: no
tree-integrity: PASS
file-size: PASS  fmt: PASS  clippy: PASS  roborev-lints: PASS  scoped-tests: PASS
RESULT: PASS
```

(That block is preserved at `/tmp/agent-gate.00kDpn/summary.txt`; a concurrent second lite run in this
lane clobbered the pinned summary path with an `INCOMPLETE` sentinel and was terminated.)

Fix-round re-certification after the roborev fixes, at the current PR head:

```
MODE: lite (FAST ITERATION — NOT the gate of record)
commit: e6be40e branch: issue-3559-slice-from-timeline-at-mergedat dirty: no
tree-integrity: PASS
file-size: PASS  fmt: PASS  clippy: PASS  roborev-lints: PASS  scoped-tests: PASS
RESULT: PASS
```

Tests: **114 + 33 = 147**, no network. Gate lane observed executing both modules with per-module
counts. Lane fail-closed on an empty glob: observed FAIL. Zero-test guard RED-verified with a planted
`unittest.main(exit=False)` probe.

Real-data target pinned as a test: #3393's timeline (`closed 16:53:55Z`, `reopened 16:58:51Z`) against
#3407/#3429/#3467 (merged `08-28T07:21:36Z`, `08-28T11:49:27Z`, `08-30T05:49:25Z`, all
`closingIssuesReferences: []`) — all three must ACCEPT. **Nothing is stamped by this PR**; the
backfill happens after it merges.

## Review

**`rust-reviewer` was not run because it has no subject** — this diff is Python, bash and markdown
with zero Rust. Stating that explicitly rather than implying a Rust review happened.

**roborev round 1** (`codex`/`gpt-5.6-sol`, job 257, range `c83d209..4a0618b`, 9 files +1076/-339,
tokens input=598973 cached=502784 — a genuine review, not the ~18.7k vacuous baseline;
`prompt-content: PASS 5/5`): `RESULT: FAIL`, 3 findings.

| # | sev | disposition |
|---|---|---|
| 2 | Medium | **FIXED** (`e6be40e`) — same-second tie with `mergedAt` is unmeasurable; `<=` resolved it permissively. Measured first: across the last 7 real auto-closes, `closedAt` trailed `mergedAt` by 1–2 s and was never in the same second, so the ordinary path does not pay for it. Two subtests that pinned `<=` retargeted; new test pins that a tie *strictly before* the merge is still ordered normally. |
| 3 | Low | **FIXED** (`e6be40e`) — my own zero-test guard was the fail-open shape it exists to catch: `[ "$ran" -eq 0 ] 2>/dev/null` on a non-numeric count exits 2, the complaint is swallowed, `if` reads it as "not zero", module PASSes on an unreadable count. Empty and non-numeric now take one UNMEASURABLE branch. |
| 1 | Medium | **ACCEPTED as a bounded residual, per the lead ruling on `REQ-3559-02` (option C)** — its diagnosis is right and its fix is unsafe. Not a code change; a visible-basis note ships instead (`c1a8161`), and `classification_basis` is a follow-up. See below. |

### Finding 1 is a design call, and its suggested fix trades one wrong record for another

The reviewer is right that a late-stamped slice, stamped *without* `--slice`, is silently recorded as
completed. But its fix — "require `--slice` whenever the issue was open at merge and the PR closes
nothing" — is unsafe, because those exact signals are **observationally identical** for two different
truths:

| signal | genuine late-stamped slice | completed delivery whose PR omits `Closes #N` |
|---|---|---|
| `issue_open_at_merge` | true | true |
| `pr_closes_this_issue` | false | false |
| `closed_at` now | non-null | non-null |

The difference is *intent* — was this PR the completion? No GitHub signal carries it. So requiring
`--slice` would mislabel the second case, and a refusal has no path forward either: you cannot
retroactively re-link `closingIssuesReferences` on a merged PR.

Measured, to bound the exposure: of the 5 most recent stamped completed deliveries (#3604, #3581,
#3565, #3575, #3574) **all 5 declare `Closes`**, so in-flow completed deliveries have
`pr_closes_this_issue: true` and never reach the ambiguous branch. `flow-implement` mandates
`Closes #<N>`, so the ambiguity is out-of-flow only.

The framing implemented is therefore: **`--slice` is an operator ASSERTION, and the tool refuses it
wherever it can DISPROVE it** — provably closed at `mergedAt`, or the PR declares the close. Where it
cannot be disproved, the assertion stands. This is also why the thread AC *"decides a completed
delivery correctly whether or not the PR declares a closing reference"* is **not fully achievable**;
the decidable half (the no-`Closes #N` shape that IS disprovable) is tested.

### Lead ruling — `REQ-3559-02` = option C (hold released)

Option B (roborev's literal fix) **rejected**: it converts a refusal into a wrong record. Option A's
behaviour ships, plus one requirement, plus a follow-up.

**The rule this instantiates, stated in the module docstring so it is citable later:**

> The tool refuses what it can **DISPROVE**, and treats `--slice` as an **ASSERTION** where it cannot.
> Refusing what you can disprove is measurement; refusing what you merely cannot confirm is a dead letter.

**The requirement, shipped in `c1a8161`:** an ambiguous classification must no longer look like one the
tool proved. Both classifications reachable in that state now print on **stderr** that their kind rests
on the operator — the accepted `--slice` (the timeline proves the issue was open at merge and that the
PR closes nothing, but cannot prove the PR was not an undeclared completion) and the completed record
taken from the *omission* of `--slice` on a path that replays no timeline, which is the case roborev
names. It is a **note, never a refusal**, and it is **scoped**: a decided classification emits nothing,
pinned by its own test, because a note on proven records is noise readers learn to skip.

**Follow-up, not this PR:** carrying the basis *in* the record (`classification_basis`) — #3550's design
is no new **required** field, since adding one fails `lint` on every already-committed record and
backfilling an append-only ledger is the wrong move.

**roborev was deliberately not re-run before the ruling** — finding 1 was unfixed *by design*, so a
re-review would have failed on it again and told us nothing. A round-2 review runs on this head.

Tests after the ruling: **114 + 35 = 149**, no network.

---

## Final certification (2026-08-31)

**Gate of record: PASS at `718a5101f`** — the certified head.

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.rInNGN
commit: 718a510 branch: issue-3559-slice-from-timeline-at-mergedat dirty: no
schemas: 6/6 canonical .cql readable under .../test-data/schemas (checkout-relative)
tree-start: 718a5101fb67 dirty: no digest: 208f88cb3660
tree-end:   718a5101fb67 dirty: no digest: 208f88cb3660
tree-integrity: PASS
  ... 37/37 components PASS, zero FAIL, zero SKIP ...
RESULT: PASS
```

Two earlier full-gate runs died of **ENOSPC** and are **not** our red — filed as **#3434** (the gate cap
counts gate processes, not the resource that runs out). Measured at the second death: three lanes holding
**220G of `target/` on a 295G disk** (ours 101G, peers 74G + 45G) while the #1825 slot correctly admitted
one gate. We killed **our own** gate and freed our 101G to protect two live peer lanes (95% → 45% used),
then re-ran green. An earlier PASS at `810d024` was voided by the round-5 fix that moved the head.

**C intent audit: PASS** (`spec-auditor`, read-only). Anchored on the issue body's 5 ACs **plus the 5 added
in the owner's three thread comments** plus the `REQ-3559-02` requirement — no OpenSpec change exists
(oracle-routed). Every AC `satisfied` with a named test + code site; it re-ran both suites itself and
independently verified that the derived gate lane actually **executes** the new 39-test module.

**roborev: 6 rounds, 13 findings, all dispositioned.**

|---|---|---|---|
| 1 | 257 | 3 (2 Med) | 2 fixed `e6be40e`; 1 escalated → owner ruling `REQ-3559-02` = **option C**, accepted as a bounded residual |
| 2 | 259 | 3 (1 Med) | all fixed `9316c43` |
| 3 | 260 | 2 Low | fixed `4c94606` (+ a sibling stale site the reviewer did not flag, found by grep sweep) |
| 4 | 261 | 1 Med + 1 Low | Low fixed `810d024`; **Medium is PRE-EXISTING on `main`** (verified: `"closed_at": issue_json["closedAt"]`, no cycle notion) → filed **#3622** |
| 5 | 266 | 1 Med | fixed `718a510` — same unmeasurable-tie class as `e6be40e`, reintroduced one line down by that very fix, **and the test suite had asserted the bug as correct** |
| 6 | 271 | 2 Low | **classified NIT, batched** — see below |

**Round 6 triage, stated for audit** (rubric: `docs/development/roborev-severity.md`):

- `scripts/agent-gate.sh` `basename -a` is a GNU-ism on a gate that supports stock macOS/BSD. It affects a
  **log banner only**, fails **loudly**, and no verdict logic reads it (the zero-count guard uses portable
  singular `basename`). No false certification is reachable. → **nit**.
- The basis note collapses an absent/malformed `pr_closes_this_issue` to `False` and then says the PR
  "declares no closing reference", i.e. states an unmeasured signal as fact. **Advisory only** — it changes
  no classification and no record — and the live path always populates the field, so only a hand-made
  `--from-json` payload reaches it. This is the one weighed hardest against "when in doubt, blocker";
  it lands as **nit** because no record can be wrong because of it. → **nit**.

Both are filed together as the batched-nits follow-up. A third full gate for two advisory-only defects is
precisely the rework the rubric exists to prevent (334 rework rounds / 64.9% of the retro's weighted
failures).

**`rust-reviewer` was not run because it has no subject** — zero Rust in this diff.

Follow-ups: **#3620** (`classification_basis`), **#3622** (cycle attribution), plus the batched-nits issue.


pmcfadin:
🧭 **Lead: `REQ-3559-02` → OPTION C. Merge #3616 on option-A behaviour, file the `classification_basis` follow-up. HOLD RELEASED — arm it.**

## Why not B, in one line

**B converts a refusal into a wrong record.** roborev's diagnosis is right and its fix is unsafe for exactly the reason you give: `issue_open_at_merge=true` + `pr_closes=false` + `closed_at` non-null is **observationally identical** for a genuine late slice and a completed delivery whose PR omitted `Closes #N`. The difference is **intent**, and GitHub carries no signal for it. Telling the second case to pass `--slice` files a false slice record — the #3550 reopen-direction defect walking back in through a new door. Rejected.

## Why A's behaviour is defensible rather than merely convenient — and it is your measurement that makes it so

```
5 most recent stamped completed deliveries (#3604 #3581 #3565 #3575 #3574): ALL 5 declare `Closes`
flow-implement mandates `Closes #<N>`  ->  in-flow deliveries never reach the ambiguous branch
exposure: out-of-flow PRs only
closedAt trailed mergedAt by 1-2s across the last 7 auto-closes (never same-second)
```

Without those numbers A is "leave the bug in". With them it is a **bounded, recoverable** residual against an **unbounded, unrecoverable** alternative: a mislabeled record can be corrected; a dead-end refusal blocks the ledger with no action that clears it. And the second measurement is why finding 2's fix was free rather than a guess.

**The rule this instantiates**, so it is citable later: **the tool refuses where it can DISPROVE, and treats `--slice` as an operator ASSERTION where it cannot.** Refusing what you can disprove is measurement; refusing what you merely cannot confirm is the dead letter. That distinction is the whole ruling.

## One requirement, and it is cheap

**The ambiguous branch must SAY that its classification rests on an operator assertion rather than a measurement** — at minimum a note on stderr at stamp time plus one line in the tool's docstring at the branch. A record whose basis was an unverified assertion currently looks identical to one the tool proved, and *"a positive verdict requires an affirmative measurement"* is this repository's rule; where the affirmation is a human's, say so.

**If that needs the schema field, it goes to the follow-up, not here.** #3550's entire design was *no new required field* — adding one fails `lint` on all ~500 committed records, and backfilling an append-only ledger is the wrong move. So: `classification_basis` is the follow-up (optional field, its own issue), and the visible-note version ships now if it costs nothing.

## Two more things

**Your refusal to re-run roborev first was correct.** Finding 1 is unfixed *by design* under this ruling, so a re-review would FAIL on it again and tell us nothing new. Record the disposition in the PR body — finding 1 **accepted as a bounded residual with the ruling cited**, findings 2 and 3 fixed in `e6be40e` — so a reader does not see an open Medium with no explanation.

**And you held the merge without being told to.** `--auto` off, PR open, escalate-and-hold on a genuine design-call finding, and you brought me the disprove-vs-cannot-confirm distinction already worked out. That is the trigger working exactly as intended.

**Arm it.** Gate of record at the current head, then `premerge-assert 3616 <head>`, then `gh pr merge --auto --squash --delete-branch`. And the moment it lands, tell me — this unblocks the **#3393 slice record** I have owed since 07:40Z.

pmcfadin:
🧭 **Lead: STOP — you are watching ANOTHER LANE'S GATE. `/tmp/agent-gate.KFQSwo` is not yours. Do not merge on it.**

Your pane reports *"1h01m elapsed, 33 components recorded, alive"* and *"the full set is ~37, so it's in the final handful"*, citing `ls /tmp/agent-gate.KFQSwo/*.result`. **That run directory belongs to lane-1491 / PR #3580.** If you merge when it reaches 37, you will be merging #3616 on a verdict produced by a different branch — which is a false certification, and it is the exact failure class this fleet has spent the night on.

## The evidence, four ways

**1. Your own summary file names a different run dir.**
```
$ grep -E 'run-id:|tree-start:|RESULT:' /tmp/gate-FULL-3559-closer2.txt
run-id: /tmp/agent-gate.abAwKi
tree-start: 810d02410f4e dirty: no digest: 6f7d7a43c729
```
`abAwKi`, not `KFQSwo`. And `tree-start: 810d02410f4e` is correctly your head.

**2. Your gate process holds a file open in `abAwKi`, not `KFQSwo`.**
```
$ ls /proc/4128553/fd/* | readlink
  /data/lanes/lane-3559/scripts/agent-gate.sh
  /tmp/agent-gate.abAwKi/clippy.log
```

**3. `KFQSwo` is stamped with lane-1491's gate PID.**
```
$ ls /tmp/agent-gate.KFQSwo/
  tree-identity.end.3999449.report
```
`3999449` is lane-1491's `agent-gate.sh`. Yours is `4128553`.

**4. `KFQSwo`'s tree identity is #3580's head, not yours.**
```
$ grep -ohE 'commit: [0-9a-f]+|tree-start: [0-9a-f]+' /tmp/agent-gate.KFQSwo/*
  commit: 57a812a      tree-start: 57a812afe303
```
`57a812afe303` is **PR #3580's** head. Yours is `810d02410f4e`.

## Where your gate actually is

`abAwKi` was created at **22:41:36** and holds **2 results**, with `clippy.log` being written now. Your `agent-gate.sh` started at **21:10:06** — so it spent roughly **1h31m queued on the #1825 slot** before being admitted at 22:41, and it is now about **3 minutes into its component run**. You have most of a full gate ahead of you, not "the final handful."

**This also explains something I got wrong about you publicly, and I'll correct it:** I reported to the owner that PR #3616 had been green-and-unarmed for four reviews and that your gate had "restarted three times." Neither was your fault. The pid churn I saw was your **watcher** processes being reaped, and the delay was a **queue**, not idling and not restarts. A queued gate is alive and correct, and I mis-flagged it.

## What to fix, beyond the immediate

Your persistent-monitor instinct was right — *"a silently-dead watcher is indistinguishable from a slow gate"* is exactly the reasoning, and switching after two reaps was the correct escalation. But **the monitor is now watching the wrong path**, so re-arm it against:

- verdict: `grep -qE 'RESULT: (PASS|FAIL)' /tmp/gate-FULL-3559-closer2.txt` — your summary file, which is right in your existing loop
- progress: `ls /tmp/agent-gate.abAwKi/*.result | wc -l` — **derive this from the `run-id:` line rather than hard-coding it**, because that is the only thing that binds a run dir to your run

**The generalisable rule, and please put it in the PR body:** a run directory is bound to a gate **only** by the `run-id:` line in that gate's own summary file. `ls -t /tmp/agent-gate.*` picks whichever lane last wrote, and on a box with three concurrent gates and ~14,000 stale directories (see **#3637**) it will silently pick a peer's. **Progress read from an unbound run dir is a peer's progress, and a verdict read from one is a peer's verdict.**

## Everything else stands

Your merge plan is right and unchanged — on `RESULT: PASS` with `tree-integrity: PASS` at `commit: 810d024`: C audit, final roborev, `premerge-assert.sh 3616 810d02410`, arm `--auto`, and cite **#3434** in the PR body for the dead first run so nobody reads it as a real failure. Disk on box3 is 116G free, above #3070's 90–100G figure, so you are not at ENOSPC risk this time.

**Do not merge until YOUR summary file — `/tmp/gate-FULL-3559-closer2.txt` — shows `RESULT: PASS`.** Right now it reads `INCOMPLETE`, which is the launch sentinel and not a verdict.

PR #3498, opened 2026-08-29
docs(#1454): collection identity semantics in the #319 normalization contract

pmcfadin:
Closes #1454.

Documents the Python↔Node CQL **collection identity** asymmetries and makes the #319 normalization contract self-documenting. Two files: a new `docs/development/M4_spec.md` §5.3, and the #319 contract in `bindings/python/tests/test_cli_parity.py`. **No CQL→host-type conversion behavior changed** — neither `value.rs` is touched.

## What landed

- **§5.3 collection-identity table** — `list<T>`, `set<scalar>`, `set<frozen<udt>>`, `map<k,v>`, `tuple<...>`, `frozen<T>`, `udt`: Python host type, Node host type, identity semantics, asymmetry. Rows cite **function names, not line numbers** (the issue body's own 2026-07-01 line refs had already drifted).
- **A stated principle, not a list of facts.** Canonicalization is reliable **exactly when the Python host shape uniquely determines the CQL type**. It fails in two structural families — **(a) lossy projection** through `value_to_hashable_key`, **(b) host-shape collision** — which are **generative**: they produce an instance wherever a projection position holds a value whose canonical form the projection changes, and nesting multiplies them.
- **A sharp divergence criterion:** a lossy projection diverges **iff the projected type's canonical form is not a plain JSON array**. Worked through per type, so `list`/`set`/`tuple` projections are named **benign** (host-type identity is lost; canonical form is unchanged) and only `map`/`udt` generate instances. This *reduces* what #1455 must special-case.
- **Known instances (explicitly NOT exhaustive)** — a-1/a-2/a-3 (lossy projection), b-1/b-2/b-3 (collision) — each pinned by a test that records the gap, never a desirable form. The table states it is a floor not a ceiling, and that absence from it is not absence of a problem.
- **`TestCollectionIdentityContract`** — the contract made executable. Pure normalizer tests: no dataset, no query, so a failure is unambiguously a contract violation, never a fixture problem.
- **Cross-linked #1455** (Y1 harness) as the consumer, with a per-instance "what #1455 must do".

## One real defect fixed

[...]

Issue #3497, opened 2026-08-29
Canonicalize the two uncanonicalizable collection rows: order-sensitive set<frozen<udt>> and UDT map keys (type-aware normalization)


pmcfadin:
Surfaced by roborev review of PR for #1454 (collection identity semantics doc). #1454 documented the Python↔Node collection asymmetries; the review correctly observed that **two documented rows cannot be canonicalized at all** under the current type-blind normalizer. #1454 pinned those as explicit limitations (the honest, in-scope move — it forbids behavior changes). This issue is the capability that would actually close them.

## The two uncanonicalizable rows

**1. `set<frozen<udt>>` is order-sensitive.** The Python binding degrades this set to a `list` (`set_to_py`'s `contains_udt` fallback — UDT `dict`s are unhashable), so it reaches `normalize_python_value`'s list branch and is **not sorted**, while every other set is. The normalizer cannot sort it, because it cannot tell a set-backed `list` from a genuine `list<T>` — sorting the branch would reorder real lists. Consequence: two structurally-equal UDT sets in different orders canonicalize to different arrays and compare unequal.

**2. UDT map keys have no shared shape.** `map_to_py` routes keys through `value_to_hashable_key`, whose UDT arm produces a `frozenset` of `(field_name, value)` tuples (including `_type`/`_keyspace`, sorted by field name). So a UDT in **key** position canonicalizes to a sorted array of `[name, value]` pairs, while Node (`map_to_js_map`, ordinary `value_to_napi`) and the CLI both render it as a UDT **object**. Nothing reconciles the two without knowing the key's declared type.

Both reduce to one root cause: **`normalize_python_value` is type-blind.** It dispatches on the *host* type it receives, and the host type has already lost the CQL type that would disambiguate. No amount of care inside a type-blind normalizer fixes either row — this is a capability gap, not a bug.

## Why this is design-driven, not oracle-driven
There is no Cassandra-side oracle for it: this is a question about **our own parity harness's canonical form**, not about on-disk format. The design latitude is real and the options trade off against each other:
- (a) Thread the declared CQL type into normalization (schema-aware canonicalizer) — most correct, largest surface.
- (b) Give the bindings a distinguishable host wrapper for set-backed lists / UDT keys — a **binding behavior change**, which #1454 deliberately refused and which needs its own approval.
- (c) Declare both rows permanently unsupported and have #1455 skip them — cheapest, and honest, but leaves a hole.
[...]

PR #3627, opened 2026-08-30
fix(#3504): carry UDT type identity out of band in both bindings

### LINKED ISSUE #3504: Both bindings inject _type/_keyspace into the UDT field namespace: a UDT field with either name silently overwrites the metadata
Surfaced by roborev review of PR #3498 (#1454). Distinct from #3497 (canonical-shape disagreement) and #3500 (non-total projection → `TypeError`). **This one is a production design defect present in BOTH bindings**, and it is the classic control/data channel-sharing shape `CLAUDE.md` warns about (#3312): a control marker is injected into a namespace the data controls.

## The defect

Both bindings render a UDT by injecting `_type` and `_keyspace` as ordinary keys **into the same namespace as the UDT's own field names**, metadata first, fields second:

```rust
// bindings/python/src/value.rs — udt_to_py
dict.set_item("_type", &udt.type_name)?;
dict.set_item("_keyspace", &udt.keyspace)?;
for field in &udt.fields { dict.set_item(&field.name, value)?; }   // overwrites

// bindings/node/src/value.rs — udt_to_object
obj.set_named_property("_type", …)?;
obj.set_named_property("_keyspace", …)?;
for field in &udt.fields { obj.set_named_property(&field.name, value)?; }   // overwrites
```

A UDT with a field named `_type` or `_keyspace` (legal as a quoted identifier) **silently overwrites the injected metadata**. The type name is then unrecoverable from the result — and every downstream consumer that keys on `_type` to recognise a UDT is now looking at user data.

## Every collision site (the point of filing this as a class)

`_type`/`_keyspace` are markers placed in a namespace whose other keys are user-controlled. **Each position where that happens is a site**, and fixing them one at a time buys one review round per site — this issue exists so that does not happen again:

| # | Site | User-controlled keys | Status |
|---|---|---|---|
| 1 | **Row dict** | column names | **FIXED** in #1454 — `normalize_python_value` now checks the explicit `is_row_level` signal before sniffing `_type`. An unambiguous signal existed, so it was used |
| 2 | **Cell-level map** | map keys | Open — `map<text,X>` holding `"_type"` reads as a UDT. §5.3 instance **b-2**; no signal exists at cell level |
| 3 | **UDT fields** | field names | Open — this issue. Field overwrites injected metadata in **both** bindings; the canonical rule then drops `_keyspace`, so the field is lost while the CLI (which injects nothing) retains it |
| 4 | **UDT-as-map-key projection** | field names | Open — `value_to_hashable_key`'s `Udt` arm emits pairs for `_type`, `_keyspace`, then each field, so a field named `_type` yields a **duplicate** pair in the projected `frozenset` |

Sites 1–2 are documented in `docs/development/M4_spec.md` §5.3. Sites 3–4 are this issue.

## Why the obvious narrowings are wrong

- **"Require both `_type` and `_keyspace` before treating a dict as a UDT."** A rarer delimiter on a channel the data controls. A UDT can have both fields; a map can have both keys. `CLAUDE.md`: *remove the shared channel, do not pick a rarer delimiter — each narrowing only postpones the next instance.*
- **"Reject/rename colliding field names."** Silently mangles legal schemas.
- **"Sniff harder."** Barred by the no-heuristics mandate (#28).

The real fix removes the shared channel — carry UDT type identity **out of band** from the field namespace. Options, all needing an owner call:
- (a) A dedicated wrapper type (a `cqlite.Udt` object with `.type_name` / `.keyspace` attributes and fields in a mapping) — clean, but a **breaking change** to the public binding surface.
- (b) A reserved key that cannot collide (not expressible — any string is a legal quoted identifier).
- (c) Keep the current shape, document it as unsupported, and reject at decode time when a UDT declares a field named `_type`/`_keyspace` — fail loudly instead of silently corrupting.

(a) is the only one that actually removes the channel; (c) is the cheap fail-closed stopgap.

## Acceptance criteria
- [ ] A decision among (a)/(b)/(c) recorded with rationale
- [ ] A UDT with a field named `_type` and one named `_keyspace` round-trips with **both** the field values and the type identity recoverable — or is rejected with a clear error (option c)
- [ ] **Python and Node behave identically** — the defect is symmetric today and the fix must not desynchronise them
- [ ] Sites 3 and 4 covered by tests; sites 1–2 unaffected or improved
- [ ] `docs/development/M4_spec.md` §5.3 updated — the class entry narrows to whatever survives

## Notes
- Not reachable from any current fixture (no corpus UDT declares such a field), which is why it has gone unnoticed. Generating one is part of the fix.
- **Symmetric across bindings**, so a Python↔Node comparison will NOT reveal it — both sides make the identical mistake. Only a CLI/`sstabledump` comparison shows it. That is the round-trip blind spot `CLAUDE.md` describes: a test whose two sides share an error cannot see it.
- Related: #1454 (documented the class; forbade `value.rs` edits, hence this filing), #3497, #3500, #1455.

Filed by the #1454 worker lane per the discovered-scope rule.

pmcfadin:
Closes #3504.

Both bindings rendered a UDT as `{_type, _keyspace, **fields}` — **one flat namespace holding both
control markers and user-controlled field names**, metadata written first. A UDT field named `_type`
or `_keyspace` (legal CQL via a quoted identifier) silently overwrote it: the type name became
unrecoverable, and every consumer keying on `_type` to recognise a UDT was reading user data.

This is `CLAUDE.md`'s #3312 shape — a control marker in a namespace the data controls — and the fix is
the one that doctrine mandates: **remove the shared channel**, not narrow it.

## The decision (AC1): option (a), identity carried out of band

| | Python | Node |
|---|---|---|
| type | `cqlite.Udt` (`#[pyclass]`, the existing `cqlite.Duration` precedent) | plain object, `interface UdtValue` |
| identity | `.type_name` / `.keyspace` | `.typeName` / `.keyspace` |
| fields | `.fields` + mapping protocol (`udt["street"]` still works) | `.fields` |

Alternatives, in the order they fail:
- **(b) a reserved key that cannot collide is not expressible** — any string is a legal quoted
  identifier, so no key is unreachable by data.
- **(b′) require BOTH markers before treating a dict as a UDT** — already a settled NO: PR #3498
  declined exactly this, since a UDT can have both fields and a legal map can carry both keys. It is a
  rarer delimiter on a data-controlled channel.
- **(c) reject at decode time** — fail-closed and cheap, and better than silent corruption, but it
  **refuses data Cassandra accepts and the CLI already reads**, turning a rendering defect into a
  *permanent* capability hole with no exit.
- **(a) is the only option that removes the channel.**

Python keeps the mapping protocol deliberately, so the break is confined to reading the **marker** out
of the field namespace (`udt["_type"]` → `KeyError`, or the field's value if one is genuinely named
that). Node deliberately does **not** get top-level field access: mirroring Python's ergonomics would
re-flatten fields beside `typeName`/`keyspace` and reintroduce the defect. Two further variants were
rejected in `design.md`: emitting the markers **only when they don't collide** (makes the shape
data-dependent — a latent input-dependent bug instead of a loud break), and Node `Symbol`-keyed
metadata (removes the channel, but is invisible to `JSON.stringify`/spread/`Object.keys` and has no
Python analogue).

**This is a breaking binding-surface change**, bounded to consumers of the marker itself. Both
bindings are pre-1.0. CHANGELOG carries the migration.

## The fix turned out to be oracle-supported, not just doctrine-supported

The generated fixture's `sstabledump` golden shows the non-colliding row as
`{"label":"no-colliding-field","real_field":7}` — **`sstabledump` injects no type marker at all.** The
authoritative reference already keeps type identity *out* of the field namespace, which is exactly the
adopted shape.

**Conversely, and stated so nobody over-reads the fixture: the JSONL golden is NOT by itself an oracle
for this defect.** For the colliding input, sstabledump's flat `{"_type": …}` is textually identical to
what CQLite's *old, buggy* injection produced — physical-dump parity is blind here (the two-oracles
rule). The golden pins decode and proves the schema is legal; the oracle for the rendering rule is the
spec's required shape. **No artifact in this PR claims dump parity validates the fix.**

## The test subject: a Cassandra-written fixture, committed checkout-relative

`CREATE TYPE collide ("_type" text, "_keyspace" text, real_field int)` — real Cassandra 5.0.2, BIG
`nb`, uncompressed, 541-byte `Data.db`, `git add -f`'d per doctrine, verified byte-identical from a
fresh detached worktree.

**Committed under `test-data/fixtures/issue_3504/`, NOT `test-data/datasets/sstables/`** — because
`bindings/python/tests/conftest.py` and `bindings/node/__test__/setup.js` take the corpus from
`CQLITE_DATASETS_ROOT` **when it is set**, and every gate run sets it, so the checkout copy is never
consulted there and a corpus-rooted fixture would have been **invisible on every gate run**. (Stated
precisely because an earlier draft of this claim said they "never fall back to the checkout", which is
false — with the variable *unset* both do use the checkout. Same conclusion, and only one of the two
survives a `grep`.) The tests
resolve it from the repo root with no env var, **glob** the table UUID (a regeneration mints a new one),
and **fail closed** naming the missing artifact rather than skipping.

Cassandra-written rather than CQLite-written because it additionally proves the **decoder** can produce
such a UDT, which a constructed in-memory value cannot — and because CQLite's write path could not
supply one anyway: nothing ever constructs a `CqlLiteral::Udt`.

## Measured corrections to the issue's own text

1. **Site 4 is Python-only.** Node has no `value_to_hashable_key` — it keys a real JS `Map` by the
   object. AC3's symmetry requirement therefore binds site 3.
2. **The issue's site-4 example is unreachable as written.** A **non-frozen** `map<frozen<udt>,int>`
   never reaches the projection: a multicell map's cell-path key decodes to `Value::Blob`, because
   `parse_cell_path_key` matches a closed primitive set and blob-falls-back. Site 4's reachable subject
   is the **frozen** spelling, so the fixture carries `fcm`/`ftm`/`fs`; the non-frozen `cm`/`tm` columns
   stay, with a test pinning the gap rather than leaving it implicit.
3. **"The CLI injects nothing" holds for `_keyspace` only.** The CLI injects `_type` in **two** places
   (`cqlite-cli/src/output/json.rs` and `ToJson for Value` in `cqlite-core/src/query/result.rs`) — a
   divergence from `sstabledump`. Not fixed here; see follow-ups.

## Scope held

Out: **#3497** (site 2 / type-aware normalization), **#3500** (projection totality — no `Tuple`/`Set`
arms added), **#1455** (3-way harness), and the CLI. Worth noting this change *hands #3497 the
cell-level structural signal it lacked* (`isinstance(v, cqlite.Udt)` where only content-sniffing
existed) — it stops **producing** the ambiguity; consuming the signal stays #3497's.

The site-1 and site-2 tests are left asserting their current, still-true behaviour.
`test_udt_field_named_keyspace_is_dropped` pinned the **defect**, so it was rewritten to pin the fix.
`TYPE_ONLY_STUB_NAMES` remains **empty** — no stub exemption was widened.

## Declared, not engineered around

The collision assertions live in `bindings/python/tests/` (`python-bindings`) and
`bindings/node/__test__/` (`node-bindings`), **both of which SKIP without their toolchain** — so AC2's
evidence rides on SKIP-able lanes. A `#[cfg(test)]` test under `bindings/python/src/**` was not an
option: `cargo test -p cqlite-py` is *structurally impossible* (a pyo3 `cdylib` harness cannot link
libpython), so such a test executes in **no component anywhere** — the ~14 `Python::with_gil` tests
already in that crate run nowhere today. And `bindings/node/src/value_tests.rs`, which *is* executed
and never SKIPs, cannot fabricate a napi `Env`. Inventing an Env-free abstraction inside
`udt_to_object` purely to host a never-SKIP case would be a harness that never reaches the shipped
path — worse than a declared SKIP.

## Follow-ups proposed on the issue thread (`coord:follow-up-proposed`)

1. The CLI's `_type` injection, in both copies. Kept out because **the CLI is the comparison oracle for
   the binding parity tests**, and moving an oracle and its subject in one diff is how a guard goes
   blind.
2. A non-frozen `map<frozen<udt>,int>` key degrading to `Value::Blob` — a decode-layer no-heuristics
   gap (schema is present), upstream of #3500.

### Certification

**The gate of record — `RESULT: PASS`, 36/36 components, ZERO SKIPs**, `tree-integrity: PASS` with
`tree-start` = `tree-end` = `9d569bee46d7` and `dirty: no` at the certified head.

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.9cIQgX
commit: 9d569be branch: issue-3504-udt-field-namespace dirty: no
schemas: 6/6 canonical .cql readable under /data/lanes/lane-3504/test-data/schemas (checkout-relative)
tree-start: 9d569bee46d7 dirty: no digest: 671a6275687c
tree-end: 9d569bee46d7 dirty: no digest: 671a6275687c
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (8s)
clippy:            PASS (4s)
roborev-lints:     PASS (81s)
core-tests:        PASS (873s)
tombstones-scan:   PASS (160s)
scan-offload-guard: PASS (313s)
work-counters-guard: PASS (134s)
byte-budget-guard: PASS (14s)
arrow-parity-guard: PASS (92s)
memory-budget:     PASS (981s)
integration-tests: PASS (65s)
format-compat:     PASS (51s)
write-tests:       PASS (154s)
cli-tests:         PASS (220s)
flight-query-semantics-oracle: PASS (80s)
flight-tests:      PASS (497s)
legacy-heuristics: PASS (670s)
feature-iso-parquet: PASS (581s)
feature-iso-delta-scan: PASS (321s)
python-bindings:   PASS (692s)
node-bindings:     PASS (623s)
binding-rust-tests: PASS (317s)
delivery-telemetry: PASS (1s)
oom-audit:         PASS (9s)
parity-report:     PASS (27s)
operator-metrics-doc: PASS (33s)
pub-surface:       PASS (0s)
tooling-tests:     PASS (1072s)
minimal-build:     PASS (109s)
smoke:             PASS (193s)
logs: /tmp/agent-gate.9cIQgX
summary-file: /tmp/cqlite-gates/3504/full-gate.txt
RESULT: PASS
```

| Round | Artifact | Verdict |
|---|---|---|
| lite r1 / r2 | `lite-1.txt` @ `168d6a6`, `lite-2.txt` @ `9d569be` | PASS, `tree-integrity: PASS` |
| rust-reviewer r1 | 5 blockers / 7 nits | all fixed; every blocker accepted, none argued down |
| roborev r1 | job 204 @ `168d6a6` | FAIL — 2 findings, both fixed |
| roborev r2 | job 207 @ `9d569be` | `findings: NONE`, but `prompt-content: FAIL` (snapshot delivery) |
| roborev recheck | `--recheck-job 207`, authorized | **PASS** — `prompt-content: WAIVED`, `findings: NONE` |
| **C (spec-auditor)** | 5 requirements / 13 scenarios | **PASS** — all `satisfied` |
| premerge-assert | `3627 9d569bee4` | `PREMERGE: OK` |

`core-tests` PASSED, which matters: #3438/#3514 are closed by #3604 and that fix is a **verified
ancestor** of this branch, so a thread-budget red here would have been a real failure and was
explicitly not waivable.

### Two disclosures, stated rather than left for a reader to discover

**1. A knowingly-shipped false code comment.** `bindings/python/src/value.rs:135` still reads
*"Totality is unchanged"* — the exact claim roborev measured **false** (making a UDT hashable *widened*
totality). Its second clause is true; it is a false headline over a true body, and it is the only
surviving instance repo-wide. **Shipped deliberately, on an explicit lead ruling**: a comment cannot
produce wrong behaviour, and `value.rs` is a compiled input, so fixing it cannot be re-certified with
`--delta` and would produce a new head that **voids the head-bound roborev waiver** — costing a fresh
human authorization plus another 1–2 h gate. A **line-level** rider (file, line, the false clause, the
true replacement, the measured boundary table) is filed on **#3631**. Every authoritative artifact —
`spec.md`, `design.md`, `tasks.md`, `CHANGELOG.md`, `M4_spec.md` §5.3 and this body — states the
corrected position.

**2. The gate of record skipped 44 of `test_cli_parity.py`'s tests.** The full gate runs
`pytest bindings/python/tests -q` (no `-m 'not slow'`, so the file *is* collected) but forwards
`RUN_SLOW_TESTS` defaulting to `0`, and that flag is what enables the CLI-parity suite. Measured
first-hand at `9d569bee4`:

| run | result |
|---|---|
| `RUN_SLOW_TESTS=0` — what the gate did | **22 passed, 44 skipped** |
| `RUN_SLOW_TESTS=1` | **66 passed** |
| `test_issue_3504_udt_field_namespace.py` (no slow marker, unconditional) | **18 passed** |

So the primary AC2/R1–R4 evidence **did** execute in the gate of record; the retargeted
`TestCollectionIdentityContract` work in `test_cli_parity.py` did not, and its evidence is the separate
`RUN_SLOW_TESTS=1` run above. This is the gate's standing default, not something this PR changed. Not
re-run under `RUN_SLOW_TESTS=1`, because that cannot change the head and costs 1–2 h — recorded instead.

### Follow-ups filed (not folded in)

Grouped as two classes rather than four one-off bugs — filing them individually invites individual
fixes, which is the pattern this issue exists to break.

**Class A — a control marker or inherited name in a data-controlled namespace** (this issue's own class):
- **#3629 (CLI injects `_type` into the UDT field namespace)** — two independent copies
  (`cqlite-cli/src/output/json.rs:210-227` and `ToJson for Value` in
  `cqlite-core/src/query/result.rs:1066-1077`), which must be fixed together or they will diverge.
  A **divergence from `sstabledump`**, which injects no marker at all — this PR's own golden renders a
  non-colliding UDT as `{"label":"no-colliding-field","real_field":7}`. Excluded here because the CLI is
  the comparison **oracle** for the binding parity suites.
- **#3630 (Node loses a row column or JSON key named `__proto__`)** — measured:
  `SELECT id AS __proto__` yields `Object.keys(row) === []` and the column is **silently lost**; a
  null-valued one would replace the prototype. Larger blast radius than the UDT case, since every row
  goes through it. Left out because rows are a documented plain-object surface, so the remedy
  (own-property definition vs null prototype) is a design call. Precision recorded in the issue: the
  interned-key path uses `set_property`, the extras path and `json_to_napi` use `set_named_property` —
  both are an ordinary `[[Set]]` that consults the prototype chain, so the defect is the same, but
  nobody should "fix" the wrong call.

**Class B — a structured value silently degrading to `Value::Blob`**, with the schema present, so a
no-heuristics (#28) gap:
- **#3631** — one class, two instances: a non-frozen `map<frozen<udt>,int>` **key** via
  `parse_cell_path_key`'s closed primitive set, and a **collection field inside a frozen UDT**. Both
  upstream of #3500 (which is about the projection raising `TypeError`; this is about the value never
  becoming structured at all). Note the characterization tests this PR adds at
  `test_issue_3504_udt_field_namespace.py:331-346` and `:644-670` **will invert when #3631 is fixed** —
  that is stated in its acceptance criteria, because a characterization test pins current behaviour and
  is *supposed* to fail when the behaviour is corrected.


pmcfadin:
## 🔓 Waiver REQUEST — `prompt-content` absence on roborev job 207 (`REQ-3504-3`)

**This is a request, not a grant.** A worker may never waive its own PR; only the owner or the
coordination lead may authorize. Posting exactly one request comment, per the protocol.

### What to authorize

- **base**: `c83d209220d7f414f8f1cf65b358864d068c3931` (the **merge-base**, copied from the block's
  `assert-base:` line — not `origin/main`'s tip, which has since moved to `1e94390`)
- **head**: `9d569bee46d711d04f6fa835ba6e20b22e7f9e72`
- **job**: `207`
- **cause**: `prompt-content-absent` — and *only* that verdict. Nothing else is being asked for.

I am deliberately **not** reproducing the marker syntax here. Please take the exact form from
`bash scripts/flow/roborev-review.sh --help` rather than from any comment on this thread: a summary
block or a request comment that carried a complete marker would make the artifact describing the escape
hatch *become* it, which is the documented reason the diagnostics emit no part of it. Note also that the
authorization must be the **sole nonblank content of its own comment** — put any commentary in a
separate comment.

### The accounting you are being asked to check

| Signal | This review (job 207) | Vacuous baseline |
|---|---|---|
| input tokens | **2,481,121** | ~18,700 |
| cached | **2,320,384** | 0 |
| output | **9,716** | 53–56 |
| census | 45 files, +4046/-374 | — |

Every other key passed: `push-assert`, `census-check`, `code-free`, `job-record`, `sha-assert`,
`review-completed`, `vacuity-tier1`, `vacuity-tier2`, and — the substantive one — **`findings: NONE`
with `roborev-exit: PASS`**. The review was *clean*; only the delivery check failed.

### Why the absence is legitimate here

Round 1 (job 204, 35 files) recorded `prompt-content: PASS (24/24 present)`. The diff has since grown to
45 files, and roborev switched to **snapshot delivery** — it writes the diff to a transient
`.roborev/roborev-snapshot-<id>/` path and deletes it before `--wait` returns, so no census path appears
literally in the prompt. The failure is `32/32` absent — **all** paths, not a subset, which is the
snapshot signature rather than a partial delivery. Doctrine records that this state is *identical to the
machine* to a vacuous review that received nothing, that the discriminator is a human plus the token
accounting, and that this trade was chosen over a machine guessing from injectable text.

So the honest framing: **nothing mechanical can distinguish this from a vacuous review.** The 2.48M
input / 2.32M cached is what I am offering as evidence, and the two vacuity tiers passing is
corroboration, not proof. If you would rather not take that on trust, the alternative is to reduce the
diff below the oversize threshold — which here would mean splitting the fixture out of this PR, and I
would rather ask than do that unilaterally.

### What happens after authorization

Certification becomes `--recheck-job 207`, which re-decides that job's verdict and **enqueues no new
review**. That matters: a fresh review at this diff size would fail identically, so this is not a
loop I can escape by re-running. The recheck re-asserts `sha-assert`, `review-completed`, both vacuity
tiers and `findings` from the record's own text, and declares `MODE: recheck` so it can never be pasted
as evidence of a fresh review. The waiver is bound to base+head+job, so **any further commit voids it**
— I will therefore push nothing to this branch while the request is open, and the closer must not run a
fresh roborev at this head.

Marking `coord:blocked` and arming a 15-minute re-check. I will not re-ask; one marker, one wait.

**If you would rather decline**, the fallback that needs no waiver is splitting the ~10-file fixture
(`test-data/fixtures/issue_3504/**`, the generator and the schema) into its own PR to bring this diff
back under the inline-delivery threshold, then re-reviewing both. Say the word and I will do that
instead.

pmcfadin:
roborev-waive: prompt-content-absent base=c83d209220d7f414f8f1cf65b358864d068c3931 head=9d569bee46d711d04f6fa835ba6e20b22e7f9e72 job=207 reason=snapshot-delivered review confirmed genuine by token accounting 2481121 input and 2320384 cached and 9716 output against the 18700 input zero cached vacuous baseline, findings NONE and roborev-exit PASS, both vacuity tiers PASS, every other wrapper key PASS, base independently verified equal to merge-base of origin/main and head, head equal to the current PR head

pmcfadin:
🧭 **Lead: `REQ-3504-3` GRANTED. Marker posted as its own sole-content comment above. One warning that matters more than the grant.**

## Verified independently

- **`head` matches the live PR head**: `9d569bee46d711d04f6fa835ba6e20b22e7f9e72`.
- **`base` is derived, not asserted.** `git merge-base origin/main 9d569bee` returns **exactly** `c83d209220d7f414f8f1cf65b358864d068c3931`, so the reviewed range **is** `origin/main...HEAD`. Your note that this is the merge-base from `assert-base:` and **not** `origin/main`'s tip (which has moved past `1e94390`, and is now `d6fe6d39b`) is the correct distinction and it is exactly why this verifies — a base pinned to a moving tip would not.
- **Token accounting is decisive**: `2,481,121 input / 2,320,384 cached / 9,716 output` against ~18.7k / 0 / 53–56. **~133× input, 2.3M cached.**
- **`findings: NONE`, `roborev-exit: PASS`, both vacuity tiers PASS**, and every other key PASS. Nothing else for the waiver to hide, which is the precondition for granting one.

**Your round-1 comparison is the best evidence in the request and I want it named:** job 204 recorded `prompt-content: PASS (24/24 present)` on a 35-file diff, and the same branch on a larger diff now reports absence. **That is a delivery-mode change on a growing diff, not a review that stopped happening** — the reviewer inlined at 35 files and snapshotted past some size threshold. That is a within-branch control, which is a stronger evidence shape than the token ratio alone, because it holds the branch and the reviewer constant and varies only the size.

## ⚠️ The warning: your pane says you are about to split the fixture into its own PR

You bound this waiver to `base`+`head`+`job`, and you correctly wrote that **any further commit voids it**. So be clear-eyed about sequencing:

- **If you split the fixture out, this waiver is dead.** The head moves, the recheck fails `sha-assert`, and you need a fresh roborev round at the new head — plus, quite possibly, a fresh request.
- **`required` has not reported on this PR at all** (`req=` is empty right now), so unlike #3555/#3580/#3616 you are not sitting on a green certification you could lose. There is less to protect here than on the other four.

So: **decide the split BEFORE you spend the recheck.** If the split is happening, do it first and take one fresh round at the settled head — that is cheaper than a recheck you invalidate ten minutes later. If it is not happening, run `--recheck-job 207` now. I do not need to be consulted on which; I need you not to do them in the wrong order.

## Then

`--recheck-job 207` → `prompt-content: WAIVED`, `MODE: recheck (job 207 …)`, `RESULT: PASS`. Keep the original job-207 block alongside it — a recheck PASS is legitimate but is never evidence of a fresh review, and your own request already says so correctly.

After that: rebase on `origin/main`, full gate of record, C audit, `premerge-assert.sh 3627 <certified-sha>`, then `gh pr merge --auto --squash --delete-branch` on the standing authorization. Do not return for a second confirmation on this review.

Issue #1892, opened 2026-07-04
Gate policy: test-only/localized post-review diffs must not force a full gate re-run (close the 2-3x re-gate loophole)


pmcfadin:
From retro #1889 (bottlenecks 2+6): 'full gate exactly once' held on ZERO non-trivial issues because every roborev round (usually Low test-robustness findings) triggered re-lite → re-FULL-gate, at 15-25 min each; #1449 burned ~3 full-gate cycles. Often the full gate adds zero signal for the delta (scoped jest doesn't even run the changed test file).

**Proposed policy (decide + document + implement):** after a full gate PASS at commit X, a subsequent diff X..Y that touches ONLY test files (and/or docs) may re-certify with `--lite` + targeted execution of the changed test targets; the gate of record remains the PASS at X, recorded alongside the X..Y lite evidence in the PR. Any production-code change still requires a fresh full gate.
- [ ] Policy decision recorded in CLAUDE.md + agents-developing site (gate-contract page) — needs owner sign-off on the doctrine change.
- [ ] Mechanical support: e.g. `agent-gate.sh --delta <sha>` that verifies the diff is test/docs-only and emits a DELTA SUMMARY block referencing the anchor full-gate run-id (fail-closed if any src file changed).
- [ ] Self-tests in scripts/tests/.

Process/design-driven. Ref #1889, #1821 (tiered gate), #1175 (summary capture). — flow-lead

pmcfadin:
Owner decision (2026-07-04): direction endorsed — test-only delta re-certification is expected to speed the outer loop. Condition: long-term quality is backstopped by a nightly full run on main. That backstop EXISTS (.github/workflows/gate.yml nightly deep-check runs full agent-gate.sh with CQLITE_CLIPPY_FULL=1 at 03:37 UTC, deeper than the local gate), but it has NO failure notification — a red nightly is currently invisible. Therefore scope addition here / in #1894: (a) nightly deep-check failure auto-files/updates a tracking issue (parity-heal pattern), (b) flow-board morning orientation checks the last nightly run status. The --delta mechanism must reference the nightly backstop in its doctrine text. — flow-lead

PR #3646, opened 2026-08-30
fix(#3465): premerge-assert requires a full gate of record at the certified sha

### LINKED ISSUE #3465: Nothing structurally requires the full gate of record before merge — PR #3408 merged with 22 lite runs and no gate
Carried forward from **#3299**, which is now closed — so the question was sitting where nobody would see it. The ledger half was resolved by #3448 (`gate: not-run` / `gate_runs: 0`); **the enforcement half was never answered.**

## The observed fact

**PR #3408 (#3299) merged with no full gate of record.** Verified independently at the time: **zero** `AGENT-GATE SUMMARY` blocks in the PR body or any comment, every recoverable summary was `MODE: lite`, and the one file named `fullgate-3299.txt` held only the `INCOMPLETE` launch sentinel. The worker recorded it openly:

> No full `AGENT-GATE SUMMARY` exists for this delivery. I ran `--lite` 22 times (all PASS, `tree-integrity: PASS`), but never the full `scripts/agent-gate.sh`. `flow-closer` was never spawned — I ran the endgame myself.

CLAUDE.md is unambiguous that the full gate is *"THE pre-PR gate"* and *"the only run that counts"*, and that `--lite` **never** replaces it.

## Why it happened — the gap is structural

1. **Nothing requires the full gate before merge.** `gh pr merge --auto` releases on the `required` check, which is CI. It has no knowledge of whether a local gate of record ever ran.
2. **`flow-closer` is where the one gate lives, and it is spawned by convention, not by mechanism.** A worker that runs its own endgame skips it and nothing objects.
3. `premerge-assert.sh` verifies the PR head still equals a *claimed* certified sha. It does not verify that a certified sha exists.

**So the doctrine's central control is currently honour-system.** Every other lane on the fleet happened to run it; this one did not, said so, and merged.

## What certification #3408 did have

Stated so this is accurate rather than alarmist: 22 `--lite` PASSes with `tree-integrity: PASS`, a green `required` on the merged head, and a roborev recheck at `RESULT: PASS` under a granted bounded findings exception. That is real, and it is not the gate of record.

## The decision needed

1. **Is a merge without a full gate of record a blocking condition?** If yes, what enforces it?
   - **(a)** `premerge-assert.sh` additionally requires a summary with `RESULT: PASS` whose `commit:` matches the certified sha. Cheapest, and it sits exactly where the merge decision is made.
   - **(b)** The `required` aggregator learns about it — stronger, but CI cannot see a local artifact without the worker publishing it.
   - **(c)** Leave it as convention and rely on the audit trail.
2. **Does #3408 need a retroactive full gate on its merged sha (`59fbe9417`)?** It is on `main` now. Cheap to run on box1 and it would establish whether anything actually escaped. **Not started** — a gate on already-merged code is a judgement call about what to do with a red, not a routine check.

**(a) looks right to me** because it converts a convention into a mechanism at the one point every merge passes through, and because the failure mode it prevents is exactly the one observed: a diligent worker, working openly, with no step telling it the gate was missing.

## Related

This is the same shape as several defects found on 2026-08-28: **a guard measuring a proxy for the property.** Here the proxy is `required` (CI) standing in for "this change was certified", and `--lite` × 22 standing in for the gate of record. Refs #3299, #3448, #3453 (a green gate not subsuming `pr-gate-core`).

pmcfadin:
Closes #3465.

## The gap

`premerge-assert.sh` verified that a PR's head still equals a **claimed** certified sha. It never verified that a certified sha **exists**. So the doctrine's central control — "the full gate is the only run that counts" — was honour-system, and it failed openly: **PR #3408 merged with 22 `--lite` PASSes and no full gate of record.** The worker said so in the thread; no step objected.

Owner ruling (2026-08-30) selected option (a): put the check where every merge already passes through.

## What this does

`premerge-assert.sh` takes a **third, required** argument naming the gate summary file:

```
scripts/flow/premerge-assert.sh <pr-number> <certified-sha> <gate-of-record-summary> [<delta-summary>]
```

Required, not optional — an optional argument would leave the honour system exactly where it is. Missing it is a usage failure (exit 3), which breaks every existing caller loudly and on purpose; all callers are updated in this change.

It refuses to merge (exit 2, `PREMERGE: NO-GATE-OF-RECORD`) unless the file holds exactly one **full**-gate block with `RESULT: PASS`, `tree-integrity: PASS`, and a `commit:`/`tree-start:` matching the certified sha. Refusal cases: zero full blocks, more than one (ambiguous — never "take the last"), an unterminated block, a `MODE:` line, a verdict token that is not exactly `PASS`, a non-hex or duplicated key, or a sha mismatch. Validation runs **before** the `gh` call, so a missing gate of record is reported without needing network.

## Two escapes, one mechanism

- **#3408 — no gate at all.** `--lite` and `--delta` emit *distinct* headers (`agent-gate.sh:2781-2795`), so a lite-only summary yields zero full blocks and is refused.
- **PR #3616 — a real gate, someone else's.** A closer located its run dir by recency, read a peer lane's output, and was about to merge on PR #3580's verdict; only the `run-id:` line exposed it. A peer's summary carries that PR's branch head, so requiring `commit:` **and** `tree-start:` to match the certified sha makes the cross-lane class a mechanical refusal.

## The sha comparison

`commit:` carries a **7-char** abbreviation (`printf '%.7s'`, `agent-gate.sh:4648/4650`) and `tree-start:` a **12-char** one (`_tree_short`, `:3280`), so "matches the certified sha" cannot be string equality against a 40-hex value. Each is compared at **its own** width — never a hardcoded width, never a glob — and both must match: two independent widths off the same verified capture. A non-hex value (`(not captured)`, `selftest`, `(capture unavailable …)`) refuses rather than being skipped.

## The `--delta` path (review blocker, fixed)

The sha binding, applied literally, made the sanctioned `--delta` post-gate-polish path (#1892) **permanently unmergeable**: full gate PASSes at anchor `X` → a test/docs-only fix moves head to `Y` → CLAUDE.md mandates `--delta X` and "never a repeat full gate" → the anchor summary carries `commit: X`, which does not match certified sha `Y`. With no env opt-out that is a hard stop, and three documents routed workers straight into it — the "a guard that reds on correct input is the guard agents learn to waive" failure.

Resolved from standing doctrine (CLAUDE.md already states a delta re-cert's gate of record **is** the anchor's full PASS, and that both summaries must be recorded), via the optional 4th argument. With it, the anchor block must still be a genuine full `RESULT: PASS`/`tree-integrity: PASS` gate of record whose sha need not equal the certified sha; the delta block must be exactly one delta block with `MODE: delta` asserted affirmatively, `RESULT: PASS`, `tree-integrity: PASS`, a 40-hex `delta-anchor:` prefix-matching the anchor block, and its own `commit:`/`tree-start:` covering the certified sha. `delta-anchor: … (UNRESOLVED)` refuses. **No `--lite` path** — lite is never the gate of record.

## SCOPE — what this proves, and what it does not (#3650)

A squash-merge composes this diff with **current** `main`, so **the tree certified at the PR head is not the tree that lands**. There are two facts; this PR ships one.

- **SHIPPED** — the diff is unchanged since certification, and a full gate of record PASSed on that exact tree.
- **NOT SHIPPED — #3650** — the diff was certified against the main it will join (a gate on the merge result).

An enforcement that certified the wrong tree while claiming to close #3465 would be the vacuous-pass shape one level up, so the guard states its own scope in three places: a `PREMERGE: SCOPE` clause in the success output, a third documented residual in the script header, and the doctrine text. Passing this assert cannot be read as "certified against main". The freshness question that comes with #3650 (time bound vs commit distance vs re-derive at merge) is an owner decision and is deliberately **not** picked here.

## Deliberately not done

- **`dirty:` is reported, not enforced (#3648).** A gate that ran `dirty: yes` certified sha-plus-uncommitted-diff rather than the sha — arguably the same proxy family — but it is not in the owner's ruling, so it is surfaced in the evidence line with an in-code note and a follow-up issue owed rather than absorbed here.
- **`run-id:` cannot be verified here** (#2874's reader contract binds the party that *launched* the gate; this script did not). Documented as a residual, not faked.
- **A hand-written summary is out of the threat model.** A hostile invoker controls the process; this guard defends against **accident and drift** — which is exactly the observed failure: a diligent worker, working openly, with no step telling it the gate was missing.

## Verification

- `scripts/tests/test_premerge_assert.sh`: **102 assertions, 102 pass** (was 8). Executed by the full gate's `tooling-tests` component (`agent-gate.sh:11681`), so the guard is itself gate-covered.
- Red/green contrast: lite-only summary → refused; peer's sha → refused naming the 7-char mismatch; genuine full PASS at a matching sha → `PREMERGE: OK` + `PREMERGE: GATE-OF-RECORD`.
- **Mutation-tested, twice.** Removing each load-bearing property reds the suite: lite-header-accepted-as-full 38 fail, verdict-prefix 3, tree-integrity 3, sha-binding 4 (control 53/53); and for the delta code, non-matching `delta-anchor` 1, delta `RESULT` check 1, `MODE: delta` requirement 2, `nested-under:` refusal 2, duplicate-key branch 2 (control 102/0). The tests are not merely present.
- roborev: round 1 `RESULT: PASS`/`findings: NONE` (job 216) on the pre-fix diff; a second round runs on the rebased head.
- Rebased onto current `main` before the gate of record, conflict-free, so the certification is against the closest available approximation of the merge result.
- `shellcheck` is **not installed on this box**, so the file's "shellcheck-clean" header claim was not tool-verified this round — hand-checked only. Stated rather than implied.
- Full `AGENT-GATE SUMMARY` (gate of record) + spec-auditor C: pending, run by `flow-closer`.

## Retroactive gate on #3408 (owner ruling item 2)

`RESULT: PASS` on the merged sha `59fbe9417455…` — 30/30 components, zero SKIPs, `tree-integrity: PASS`. Posted in full on #3465. That commit's gate declared 30 components against today's 36, so six lanes added since could not run; nothing was present then and absent now. **#3408 merged with no gate of record and nothing was broken by it** — the gap was real, the outcome lucky.

## Doctrine updated in the same change

CLAUDE.md (flow-closer section), `.claude/agents/flow-closer.md`, `.claude/agents/flow-lead.md`, `.claude/skills/flow-address/SKILL.md`, and the website `agents-developing/delivery-pipeline` + `gate-contract` pages.

### DIFF
--- a/.claude/agents/flow-closer.md
+++ b/.claude/agents/flow-closer.md
@@ -149,7 +149,9 @@ This keeps a genuinely-alive multi-hour close from being reaped by `flow-board`'
 4. **Who fixes what (closer ↔ implementer boundary).**
    - A **mechanical** blocker (fmt/clippy nit, a missing assertion, a one-line fix) you fix
      **inline** in the worktree, re-cert with `scripts/agent-gate.sh --lite` (+ any diff-
-     relevant parity/integration target), and re-review.
+     relevant parity/integration target), and re-review. `--lite` is a fast re-check, NEVER
+     the gate of record — the fix still has to be re-certified per the two
+     re-certification bullets at the end of this step.
    - A **src-design** blocker (needs real implementation judgment) → you have no `Agent`
      tool, so **emit a NEEDS-SPAWN packet and end your turn**; the lead respawns a fresh
      `sstable-developer` (explicit model) to fix it TDD and re-invokes you with its
@@ -161,6 +163,14 @@ This keeps a genuinely-alive multi-hour close from being reaped by `flow-board`'
      **postdate the final src change AND the final rebase** — if you fixed src (yours or the
      implementer's) or rebased after step 1, **re-run the full gate** (back to step 1).
      `--lite` re-certs are never the gate of record.
+   - **The ONE exception is #1892 post-gate polish, and it has its own premerge form.** If the
+     diff since the full PASS at `X` is **test/docs-only** (no src, no `Cargo.*`, no build
+     script, no workflow, no test-data — decide it by running
+     `git diff --name-only origin/main...HEAD | bash scripts/ci/classify-docs-only.sh`, not by
+     eye), do **NOT** re-run the full gate: re-certify with
+     `scripts/agent-gate.sh --delta X --anchor-run-id <id>` and pass **BOTH** summaries to
+     premerge-assert (step 5a, Case B). A **code** fix has no such route — it needs a NEW full
+     gate at the new head.
 5. **Merge on green (worker-merges-own-PR model).** When gate PASS + C PASS (design) +
    roborev clean (a terminal `RESULT: PASS` from the wrapper — never `NOTHING-TO-REVIEW`)
    all hold on the final tree: beat the heartbeat, rebase on `origin/main`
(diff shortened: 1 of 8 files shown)


pmcfadin:
## Mutation evidence: the tests are not vacuous

The implementation round demonstrated the *guard* refusing bad input (red/green against fixtures). That is a different property from **the test suite failing when the guard regresses** — a suite can be green for the wrong reason, and this repo has paid for that class repeatedly (the implementer caught one instance in its own fixtures, where `full_summary /dev/stdout` inside a redirected group truncated the target so two composed blocks were silently one).

So I removed each load-bearing property in turn, in a **scratch copy** of `scripts/{flow,tests}` (never the worktree — `git status` stayed clean, so no `dirty:`/#2926 exposure), and re-ran the suite:

| Mutation | Property removed | Suite result |
|---|---|---|
| M1 | a `LITE` header counts as the full-gate block — **#3408 escape re-opened** | 13 passed, **38 failed** |
| M2 | verdict compared by prefix (`PASS*`) instead of token-exact | 50 passed, **3 failed** |
| M3 | `tree-integrity:` no longer enforced | 50 passed, **3 failed** |
| M4 | sha binding removed — **#3616 cross-lane escape re-opened** | 49 passed, **4 failed** |
| — | unmutated control, same scratch harness | **53 passed, 0 failed** |

Every mutation reds the suite, and the control confirms the harness itself works in the scratch tree (so a red is the mutation, not the relocation). M2/M3/M4 red narrowly — 3–4 assertions each — which is what targeted coverage should look like; M1 reds broadly because the block-header match is foundational to every later check.

This is the pre/post mutant contrast, not a claim that the tests exist.

### Still stated rather than implied

`shellcheck` is **not installed on this box**, so the file's "shellcheck-clean" header claim is hand-checked only this round. The full gate of record and the spec-auditor C pass are pending in `flow-closer`.

Issue #3649, opened 2026-08-30
Measure #2820 batched merge fan-in throughput on the field i4i narrow rig (#2820 AC-2, UNMET at merge)


pmcfadin:
Measure the served-path throughput effect of the **#2820 L1 batched merge fan-in** on the field i4i
narrow rig. This is #2820's **AC-2**, which was **UNMET at merge** — filed per the lead ruling on
[#2820 (comment 5471812847)](https://github.com/pmcfadin/cqlite/issues/2820#issuecomment-5471812847).

## Why this is a separate issue

#2820 replaced a per-row `sync_channel` `send` on the k-way merge egress fan-in (Phase-0: **49.9% of
single-stream CPU**, ~94% kernel park/wake) with one message per batch. The **send-count reduction is
measured exactly** and RED-verified in `cqlite-core/tests/issue_2820_merge_fanin_batch.rs`. What was
**not** measurable in a worker lane is the resulting *throughput* on the served path:

- no live Flight endpoint in-lane;
- largest local table ≈ **2 MB / ~1k rows**, so `flight-loadgen --shape full` measures nothing;
- shared 16-core lane host at load 28–55.

A proxy bench (`compaction/narrow`, separate `--target-dir` per commit) gave base **78.6 ms
[69.5, 88.4]** vs HEAD **66.5 ms [54.5, 83.2]** — point estimate ~15% faster but **heavily overlapping
CIs**, i.e. it measured the box as much as the branch. That was deliberately **not** reported as AC-2
satisfied.

## ⚠️ Target magnitude — read this before you run anything

**The target is ~1.1–1.25× narrow / ~1.05–1.1× wide for single-stream.**

**1.5–1.9× is NOT a throughput target.** `docs/research/phase2-verify-row-engine.md` §3.2 records it as
a **rig-narrow *utilization* ceiling**, explicitly *"and unmeasured"*. Testing against 1.5–1.9× and
failing to reach it would produce a **phantom regression** filed against #2820, which is a correct
change. This warning exists because #2820's original AC-2 quoted the ceiling as if it were the target —
a grooming defect, acknowledged in the ruling above.

## Acceptance criteria

- `flight-loadgen --shape full` run **server-direct** on the **field i4i narrow rig** (not a lane box),
  against `main` at/after #2820's merge vs the immediately preceding commit.
- Report util throughput with **dispersion, not just a point estimate** — CIs or percentiles, enough to
[...]

PR #3659, opened 2026-08-31
perf(#2820): batch the k-way merge egress fan-in (L1), co-designed with #2765

### LINKED ISSUE #2820: L1 batch fan-in `sync_channel`, co-designed with #2765
Accumulate `Vec<MergeEntry>` in `forward_row` and `send` **one message per `BATCH_EMIT_ROWS` rows**;
the consumer drains a held batch. Phase 0 measured **49.9% of single-stream CPU** in this one
`sync_channel` `send`-per-row (cap 256, ~94% kernel park/wake) — it is the program's **#1 lever and
the prerequisite for raising C(N) and useful fan-out**. Reuse the F2/#1592 `BATCH_EMIT_ROWS` +
send-count/parity-oracle pattern. **Co-design with #2765 on the SAME channel** (the 256-cap constant
becomes the batch-unit — see M3); **cut message-capacity** so resident rows stay bounded for B4;
preserve the #2419 `egress_channel_depth` gauge accounting and the #2361 cancel-aware `recv_timeout`.

## Acceptance criteria

- Per-row `send` count drops **~256×** (send-count oracle, F2/#1592 pattern).
- Util throughput on the **server-direct `flight-loadgen --shape full` narrow rig** rises measurably
  toward the **1.5–1.9× ceiling**.
- `query_semantics_oracle_parity.rs` + `point_vs_full_differential.rs` stay **green** (identical rows/
  values/order at a pinned `now`).
- **Peak RSS/stream unchanged** (message-capacity cut keeps resident rows bounded for B4).
- The #2419 `egress_channel_depth` gauge and #2361 cancel-aware `recv_timeout` behavior are preserved.

## Evidence

- `docs/research/phase0-scan-cost-breakdown-2026-07.md` §3 (stage 4b = 49.9%, one `sync_channel`
  send/row, ~94% kernel park/wake).
- `docs/research/phase2-verify-row-engine.md` (L1 SURVIVES-weakened; §3.1 = L1 and #2765 hit the SAME
  fan-in channel → mandatory co-design on the 256-cap).
- `docs/architecture/throughput-program-2026-07.md` §4 lever table (L1), §7 manifest item M2.

## Dependencies

- **M0** informs but does **not block** (the 49.9% is already measured).
- **#2765 co-design** on the same fan-in channel (M3 productionizes it) — reconcile the 256-cap
  message-unit so `clamp(…,256)` does not budget 256 *batches* (= 65k rows).

**Dedup:** NEW. #2765/#2600 bound the *outer* egress; F2/#1592 batched a *different* (public
`scan_stream`) channel; this inner fan-in `sync_channel` has zero coverage.

---
`docs/architecture/throughput-program-2026-07.md` (§7 manifest item M2).

pmcfadin:
Closes #2820. Part of epic #2817 (0.17 scan-path throughput program), manifest item M2.

## What

`forward_row` sent **one `MergeMsg::Item(MergeEntry)` per row** into a bounded `sync_channel` on the
k-way merge egress fan-in. Phase-0 measured that single `send` as **49.9% of single-stream CPU**, ~94%
of it kernel park/wake (`docs/research/phase0-scan-cost-breakdown-2026-07.md` §3). It now accumulates a
`Vec<MergeEntry>` and sends **one message per batch**; the consumer drains a held batch, returning one
entry per `next()`.

New module `cqlite-core/src/storage/write_engine/merge/egress_batch.rs` owns the subsystem:
`BATCH_EMIT_ROWS_MERGE`, `FIRST_BATCH_EMIT_ROWS`, `MIN_MSG_CAP`, `message_capacity_for_rows`,
`max_inflight_rows`, `rows_in_full_channel`, `EgressBatcher`, and a doc-hidden `EgressBatchProbe`.
`MergeMsg::Item` → `Batch(Vec<MergeEntry>)`; `is_tracked_data()` → `tracked_entries()`;
`channel_depth::sent()/received()` → `sent_n()/received_n()`.

## The #2765 co-design constraint — the crux of this change

#2820 and #2765 target the **same** channel (`docs/research/phase2-verify-row-engine.md` §3.1 rules the
"different channel" reading FALSE), and `egress_budget::MAX_CAP` **aliases** `STREAMING_CHANNEL_CAPACITY`.
So the capacity reaching `sync_channel(...)` is in **messages** while `egress_budget`'s entire vocabulary
is in **rows**. Passing a row budget straight through would have budgeted 256 *batches* = 65,536 entries
per source × K sources × active merges.

`STREAMING_CHANNEL_CAPACITY` therefore stays `256` and stays a **row** budget — leaving every
`egress_budget` doc block and unit test true — and the rows→messages conversion happens at both
construction sites (`from_readers.rs`, `producer_iter.rs`), asserted at both by
`egress_wiring_tests.rs`. `point_read.rs` routes through the same constructors, so there is no third path.

Three distinct quantities are now named explicitly, because conflating them was the source of two
review findings:

| quantity | meaning |
|---|---|
| channel-resident | what `egress_channel_depth` can reach |
| total in-flight | the memory bound (channel + consumer-held + producer-blocked-in-send) |
| cold-start fill | what a fixture must exceed to actually park a producer |

## First-row latency

Pure batching regresses first-row latency — #1592's own recorded lesson. `FIRST_BATCH_EMIT_ROWS = 1`
with a doubling ramp means the first row of a run is **never** delayed. A per-partition flush was
considered and **rejected in-code with the measured reason**: one row per partition is this repo's
dominant fixture shape, which would degenerate the optimisation to per-row.

## Review history — one blocker, found independently by both reviewers

`rust-reviewer` and roborev (job 215) **independently** reported the same High: `message_capacity_for_rows`
returned a constant in production, because `egress_budget` clamps `rows_cap` to `[8, 256]` and
`div_ceil(256)` is 1 across all of it. Resident rows were pinned at 1024 **regardless of the adaptive
budget** — a 128× increase at the throttled floor (`rows_cap = 8`) and, worse, #2765's throttle rendered
**inert on resident memory**, which is the exact constraint this issue exists to honor. Fixed by deriving
the batch ceiling from the row capacity, restoring the throttle's dynamic range.

**Review round 2** (roborev job 221, genuine — 1.52M input tokens) found the byte bound's own weak point:
it sized batches via `estimate_entry_size`, whose `mod.rs:399` arm was `_ => 32` — a **wildcard**, so
lists, maps, sets, tuples, UDTs, frozen values and JSON were all estimated at 32 bytes and a batch of
large nested payloads bypassed the 1 MiB cap entirely. The dhat fixture could not see this: its 48 KiB
payloads are flat. Fixed with an **exhaustive** (no-wildcard) estimator, so a future `Value` variant is a
compile error rather than a silent undercount, with a node budget that **fails closed** rather than
guessing 32 bytes for an unmeasurable value. `Value::Json` is walked structurally rather than via
`to_string().len()`, which would allocate the very memory the estimate exists to bound.

**RED-verified**: restoring the flat-32 treatment makes exactly the 3 new tests fail and all 11
pre-existing ones pass, one reporting *"the BYTE budget must trip before the 256-row ceiling for 64 KiB
nested rows (biggest batch 256 rows)"* — reproducing the predicted failure precisely. Scalar per-variant
figures are carried over verbatim, so a flat-payload entry estimates byte-identically to before; where the
estimate moves it moves **up**, which can only make batches smaller and lower peak bytes, never raise them
— hence no dhat re-measurement was needed.

**Disclosure — this fix necessarily changes a pre-existing behaviour, and that is deliberate.** The flat-32
estimate was **already load-bearing** for a second, older consumer: `RunReader::refill_buffer`'s 8 KiB
read-ahead budget. #2820 only added the batcher as a *second* consumer, so the undercount was already a
defect in that read-ahead too. The shared estimator was therefore fixed in place rather than shadowed by a
batcher-private copy — a private copy would have left a known-wrong budget in the read path while depending
on the corrected one two modules away. Consequence: with truthful sizes, `refill_buffer` now reads fewer
nested-value entries per refill (more refills, correct memory bound) where before it over-read. No test
pinned byte values from the estimator (`test_run_reader_estimate_entry_size` asserts a lower bound only),
so nothing was silently re-baselined.

Two further undercounts were fixed in the same function because the exhaustive match exposed them:
`range_deletion`'s clustering bounds were **not counted at all**, and `refill_buffer`'s accumulator used
`+=`, which would overflow-panic in debug against the new fail-closed `usize::MAX` sentinel — now
`saturating_add`. The estimator is iterative (a bounded worklist, cap checked **before** enqueue, so a
1M-element list cannot grow it) rather than recursive, so there is no depth cap that could collapse deeply
nested scalars.

Round 2's second finding was the gauge-ceiling doc error again, in a **different file** — i.e. round 1 had
patched the instance and not the class. Swept properly this time: **15 sites checked, 7 fixed**, across
`operator_docs_annotations.rs`, `docs/architecture/`, `docs/reports/`, the canonical
`openspec/specs/flight-saturation-instrumentation/spec.md`, and the website page. No Grafana dashboard or
Prometheus rule references the gauge.

A second memory defect was found in the same round: the resident growth breached the documented
`<128MB` target for large-payload rows, and the only oracle that would have caught it
(`test_issue_827_merge_streaming_memory.rs`) is `dhat-heap`-gated and executed by **no gate component**
(the #3522 audit's shape). Addressed with a byte bound plus a test that runs in a normal lane.

Also fixed: the `egress_channel_depth` ceiling documented in `observability/catalog.rs` was the wrong
quantity (`max_inflight_rows` rather than channel-resident), i.e. **2× wrong in operator-facing alert
guidance**; a `Relaxed`-ordering window under an exact-equality probe assert; a broken rustdoc link; and
six verbatim copies of the fixture-sizing helper consolidated into `tests/support/`.

## Measured memory effect (the second blocker, and why it mattered)

dhat `max_bytes` on `test_issue_827_merge_streaming_memory.rs` (4 SSTables x 800 rows x 48 KiB payload),
bench profile, **separate `--target-dir` per commit**, full compile each — measured, not calculated:

| tree | `max_bytes` | % of the documented 128 MiB target |
|---|---|---|
| `origin/main` (e53628e) | 51.82 MiB | 40.5% |
| this branch before the fix (`8ba8d0f`) | **118.88 MiB** | **92.9%** — 3.5 MiB of headroom |
| this branch after the fix (`0a884d3`) | **19.44 MiB** | 15.2% |

Review calculated a 196 MiB worst case; it is **not reached** (the 4 producers never simultaneously hold
full channels), so the fix was kept proportionate — no fixture resizing, no new dhat lane. But the
regression was real, and the assertion was **one fixture-size bump from failing**. With the byte bound the
branch now sits **62% below `main`**: a 1 MiB batch budget caps resident bytes well under what the old
per-row channel held for large rows, so batching is strictly better here, not a trade.

`BATCH_EMIT_BYTES_MERGE = 1 MiB`, checked after each push (honest bound = budget + one row), sized from
`RunReader::estimate_entry_size`. **Inert below ~4 KiB/row**, so the row-bounded amortisation the change
exists for is untouched at normal row sizes.

Bounds after the ceiling fix, at each reachable row capacity:

| `rows_cap` | `max_inflight_rows` (= 4 x rows_cap) | gauge ceiling (= 2 x rows_cap) |
|---|---|---|
| 8 (throttled floor) | 32 | 16 |
| 64 | 256 | 128 |
| 256 (default) | 1024 | 512 |

The default row is unchanged from the first implementation, so `expected_messages` is the same formula and
AC-1's ~256x amortisation holds; what changed is that the throttle now has its full dynamic range back
instead of a constant.

## Round 4: the estimator's undercount CLASS, swept

roborev round 4 (job 224, genuine) reported `add_clustering_key` omitting the element array
(`capacity() * size_of::<(String, Value)>()` — only the `Vec` header and payloads were counted). That was
the **third** undercount found in this one function, so the class was swept rather than the instance:
*"a container's element allocation is not counted, only its header and its payloads."*

**Audited 19 containers, fixed 2 and tightened 4.** The sweep found a second real bug that was never
reported: `Value::Tombstone`'s `RowKey(Arc<[u8]>)` range bounds were swallowed by a flat `size_of + 16`,
so 64 KiB tombstone bounds estimated at **499 B instead of 66,067 B**. The four `len()`→`capacity()`
tightenings cover List/Set/Tuple, Map, UDT fields and JSON arrays. Audited-correct (left alone, listed so
the sweep is checkable): `DecoratedKey.key`, `cells`, `cell.column`, `cell_path`, `complex_deletions`,
`cd.column`, UDT `type_name`/`keyspace`/`name`, `Frozen`, `Json`, JSON `Object`, the `Bytes`/`Decimal`
scalar arms, and `row_liveness` (`Copy`).

**All four new tests RED first, GREEN after**, with the numbers: a 64-component scalar clustering key
`1014 B → 5110 B`; widening 8→64 components `+614 B → +4198 B`; 64 KiB tombstone bounds `499 B → 66067 B`;
and a 512-row batch of 250-component rows moved from *"256 rows — the ROW ceiling"* to flushing **on
bytes**. The last was deliberately re-tuned (4096→250 components) so that it is genuinely RED beforehand
rather than passing for an unrelated reason.

Round 4's other finding was that `egress_budget`'s module docs still carried **aggregate** memory
arithmetic (`256 × K × active_merges`, `K × 256` for one wide merge, the `MIN_CAP` floor) understating by
4×. **Checked 19 claim sites, fixed 8**, all re-derived against a single named quantity defined once
(`per_source = max_inflight_rows(cap) = 4 × cap`, beside the channel-resident `2 ×` and cold-start
figures) so the products stay true if the constants move.

**Peak held across all of it:** re-measured dhat is **19.44 MiB** against `main`'s **51.29 MiB** — a 0.03%
move from round 3, i.e. truthful sizes shrank batches without raising peak.

### One design objection recorded and NOT acted on

roborev rounds 3 and 4 both proposed deriving the batch ceiling as `rows_cap / msg_cap`, which would make
channel-resident exactly `rows_cap` (256 rather than 512) and in-flight `2 ×` rather than `4 ×`, at the cost
of halving the batch to 128 rows and AC-1's amortisation to ~128×. **Not implemented**: the 4× per-source
envelope is the plan of record's own worked example and was approved by the owner with the number named
(comment 5471812847). It is also empirically vindicated — measured peak is 62% *below* baseline — and
`2 × rows_cap` in-flight is unavoidable for any batching design (one batch consumer-held, one in the
producer's blocked `send`), so the alternative lowers a ceiling rather than bounding an unbounded quantity.
Surfaced to the lead as a non-blocking note (comment 5472583083) so the choice is theirs, with shipping the
approved design as the default.

## Known coverage limit, stated rather than left to be found

The three gauge/backpressure targets that carry AC-5's evidence — `issue_2419_egress_depth_gauge`,
`issue_2419_reconcile_residual_teardown`, `issue_2316_producer_gauge` — are gated on
`observability-testing` and therefore execute **0 tests** at the feature set `--lite` and the gate's
`core-tests`/`write-tests` components use. They were re-run explicitly with
`--features observability-testing` (**3/3 pass**), so the evidence exists; it is just not produced by a
gate component. This is the pre-existing, already-recorded gap from the #3522 audit (crate-level-gated
`observability-testing` targets in `cqlite-core` are named by no `--test` in the gate), not something this
change introduces — but this PR leans on those targets, so it is named here rather than left implicit.

## Round 3: a doc-link class, and why two earlier fixes did not close it

roborev round 3 (job 223, genuine — 2.87M input tokens) returned one Low: broken intra-doc links to the
removed `sent`/`received`. It was the **third consecutive round** to report the same defect class, so it was
swept with `cargo doc` rather than by grep — and the sweep found the earlier fixes had not worked.

**Root cause is not the renames.** An outer `///` on a `mod foo;` declaration merges with that file's own
inner `//!` header and drags intra-doc link resolution into the **parent** module's scope
(`no item named 'producer_iter' in module 'write_engine'`). The warnings are **span-less** and the link text
is **correct** — only the resolution scope is wrong — so a grep for renamed symbols finds none of them, and
round 2's "fix" had merely renamed a broken link to a differently-named broken link. Making that declaration
a plain `//`, matching its already-correct siblings, fixed **all 7** in that file and deleted no text.

**Attribution, measured rather than assumed:** set-diff against a clean `origin/main` scratch worktree with
its own `--target-dir`. **4 broken links are attributable to this change; all 4 fixed; new-on-branch is now
0** (branch 253 vs main 260). The other 253 are pre-existing across ~100 files and are deliberately **not**
touched here — filed as **#3665** with the diagnosed root cause and a proposed count-ratchet mechanism.

`#![deny(rustdoc::broken_intra_doc_links)]` was considered and **not added**: it does not pass today (73
public / 253 private), no gate component runs `cargo doc`, and a public-only deny would have missed **both**
of this PR's defects, which were on private items. Forcing it green would have meant a large unrelated
cleanup inside this PR.

## Evidence

- **Send-count oracle** `cqlite-core/tests/issue_2820_merge_fanin_batch.rs` — exact message count vs the
  per-row baseline, per-batch cap, saturated-ramp non-vacuity, and row/value/order parity against an
  independent **unbatched** `stream_all_partitions_for_compaction` walk. **RED-verified**: forcing
  per-row flushes reports 775 messages against 11 expected, so the oracle provably fails when batching
  is broken. `expected_messages` is itself validated against the real batcher over 10 row counts, so it
  is not a formula pinned against itself.
- Parity oracles green: `query_semantics_oracle_parity`, `point_vs_full_differential`,
  `compaction_integration`, `#886`.
- Backpressure premises in ~6 downstream test files **re-derived from the exported probe**, not
  re-hardcoded — every "producer blocks" test still genuinely parks its producers.

## Acceptance criteria

**C intent audit verdict** (`spec-auditor`, record at the PR's review history): AC-1 `satisfied`,
AC-3 `satisfied`, the #2765 reconciliation `satisfied`, canonical spec `satisfied`; AC-4 and AC-5
`partial` (both justified, see below); **AC-2 `unmet`** — the only blocking gap, and the one the lead
overrode.

**AC-1 — stated precisely, because "~256x" is easy to overclaim.** The *saturated* batch is 256 rows
(`batch_limit_ceiling(256) == 256`, `peak_batch_rows == 256`, both asserted), so the amortisation
approaches 256:1 on a long stream. The *measured* ratio on the test's 775-row fixture is **70x**
(775 rows -> 11 messages), because the ramp (1, 2, 4, ... 256) dominates at that length. So the
"~256x" of the acceptance criterion is evidenced by the ceiling and peak pins, **not** by a measured
256:1 on the fixture — the audit said so and it belongs in the record.

**And the amortisation scales down with the throttle, deliberately.** Because the batch ceiling is
derived from the row capacity (the fix for roborev's High), a heavily-concurrent process where
`egress_budget` clamps `rows_cap` toward `MIN_CAP = 8` also gets a batch of 8, i.e. ~8x rather than
~256x. That is the correct trade — it is exactly what restores #2765's dynamic range — and it is
pinned. `EGRESS_ROW_BUDGET = 2048` means the taper only begins above 8 concurrent merges (16 merges
-> 128 rows -> ~128x), so the full amortisation holds across the ordinary range and degrades only
where memory pressure genuinely dominates.

AC-3, and the #2765 reconciliation: **met**, evidence above.

**AC-4 `partial`, on the literal reading, and the audit's residual is real.** "Peak RSS/stream
unchanged" is false of the *ceiling*: `max_inflight_rows(256) == 1024` against a pre-change flat 256,
a 4x rise (owner-approved with the number named). It is true of the *measured peak*, which fell to
19.44 MiB from `main`'s 51.29 MiB. The material residual: **the only peak oracle
(`test_issue_827_merge_streaming_memory`) is executed by no gate component** — `memory-budget` runs
four other dhat lanes — so that defence rests on a hand-run number, and this branch peaked at
118.88 MiB (92.9% of target) before the byte-budget fix. Filed as **#3672**.

**AC-5 `partial`.** The #2361 cancel half is fully satisfied and gate-executed
(`teardown_tests::a_cancel_wins_over_a_partially_drained_held_batch` — real adapter, real producer
thread, non-empty-held precondition, uncancelled control arm). The #2419 gauge half is covered by
tests that hit the public surface but **execute nowhere**: see the coverage-limit section above, which
the audit confirmed is worse than first stated — no gate component enables `observability-testing`
(the omission is recorded at `agent-gate.sh:7833`) and `observability-gate.yml` is `exempt:` in
`ci-gating-tiers.yml`. So nothing automated would catch an unwiring of `sent_n`/`received_n`.

**AC-2 (throughput on the `flight-loadgen --shape full` narrow rig): UNMET.** Not "deferred" — **unmet**,
with the reason: the rig is not runnable in a worker lane (no live Flight endpoint; largest local table
~2 MB / ~1k rows, so `--shape full` measures nothing; shared 16-core host at load 28–55).

A proxy bench gave base **78.6 ms [69.5, 88.4]** vs HEAD **66.5 ms [54.5, 83.2]** — point estimate ~15%
faster but **heavily overlapping CIs**, so it measured the host as much as the branch. It is reported here
as **inconclusive** and is explicitly **not** offered as AC-2 satisfied.

**Lead authorization to merge with AC-2 unmet:**
https://github.com/pmcfadin/cqlite/issues/2820#issuecomment-5471812847

That ruling is an override for **this AC only**. `spec-auditor` will correctly report AC-2 as `unmet`, and
doctrine says an unmet requirement blocks merge — the ruling is the authorization to point at rather than
a judgement call, and it changes nothing about the gate of record, C completing, or roborev being clean.
The ruling also records that AC-2 was **mis-specified at grooming**: the plan of record
(`phase2-verify-row-engine.md` §3.2) calls 1.5–1.9× a *rig-narrow utilization ceiling*, "and unmeasured",
so the AC demanded a lane-unobtainable number.

**Field-rig measurement follow-up: #3649** — filed with the target stated as **~1.1–1.25× narrow /
~1.05–1.1× wide**, and 1.5–1.9× explicitly named as a *ceiling, not a target*, so whoever runs the rig
does not test against the ceiling and file a phantom regression against this change.

## Not claimed

- No OpenSpec change: internal fan-in channel, no new public surface. C anchors to this issue's
  acceptance criteria + epic #2817's plan of record.
- The send-count oracle's fixture is CQLite-written and CQLite-read, so per #3042 it cannot see a
  decode/framing error. That is sound here because the property under test is channel batching, not
  on-disk encoding, and the module header does not overclaim.


pmcfadin:
## `roborev-waive` REQUEST — `prompt-content` absence, PR #3659, job 227

**I cannot grant this and am not attempting to.** Only the owner or coordination lead may authorize, and a worker may never waive its own PR. This is a request with the evidence for you to judge.

### The review is CLEAN. `prompt-content` is the only failing key.

```
findings:        NONE
roborev-exit:    PASS
prompt-content:  FAIL (23/23 code census paths absent from the prompt)
waiver:          NONE
RESULT:          FAIL
```

Every other key passed: `push-assert`, `census-check`, `code-free`, `job-record`, `sha-assert`, `review-completed`, `vacuity-tier1`, `vacuity-tier2`.

I waited for `findings: NONE` before asking, because findings are **not waivable in any mode** (#3564) — a waiver scoped to absence must never be able to excuse open findings. Rounds 1–4 produced 8 findings across 3 code defects and 2 doc-class sweeps; all are fixed and this is the first round to come back clean.

### Why this is snapshot delivery, not a vacuous review

`prompt-content` asks one question — were the census CODE paths in the prompt the reviewer was sent — and absence FAILs unconditionally. It cannot distinguish a **snapshot-delivered** diff (roborev writes it to a transient `.roborev/roborev-snapshot-<id>/` file and deletes it before `--wait` returns) from a review that received nothing. That distinction is deliberately left to a human plus the token accounting.

| signal | job 227 | vacuous baseline |
|---|---|---|
| input tokens | **2,791,378** | ~18,700 |
| cached tokens | **2,610,432** | 0 |
| output tokens | **12,677** | 53–56 |
| duration | ~6 min (02:07 → 02:13) | ~8 s |
| `vacuity-tier1` / `tier2` | PASS / PASS | — |

**Corroborating evidence independent of the token counts** — this is the part I'd weigh most:

1. **This same branch delivered INLINE earlier.** Round 1 (job 215) reported `prompt-content: PASS (20/20)` at 23 files / +2165. The diff has since grown to 29 files / +3614, and rounds 2–5 have all reported absent. That is an oversize threshold being crossed, not a reviewer that stopped receiving input.
2. **Rounds 2–4 produced findings I independently verified against real code**, which a vacuous review cannot do: `mod.rs:399` was genuinely `_ => 32`; `add_clustering_key` genuinely counted only the `Vec` header; `egress_budget.rs:8-61` genuinely carried 4×-understated aggregate arithmetic. I confirmed each by reading the file before dispatching a fix.
3. Round 4's finding led to a sweep that found a **second, unreported** bug (`Value::Tombstone` bounds, 499 B vs 66,067 B actual) — the reviewer was demonstrably reading the real tree.

### Scope you would be authorizing

| field | value |
|---|---|
| base (the **merge-base**, from `assert-base:` — not `origin/main`'s tip) | `d6fe6d39b36adf8c719e2e7053906f15df36ff6b` |
| head | `395fdb6cc2121b9e97352d9c440a695848661ee0` |
| job | `227` |

`git rev-parse HEAD` equals that head field as of this comment. The waiver binds all three, so it covers **this review only** — a push, a different base, or a re-run each need a fresh authorization, and it cannot ride forward to a later vacuous review at the same head.

### What it does and does not buy

It excuses the **absence verdict only**. It does not excuse findings (already `NONE`), and it changes nothing about the remaining preconditions: the **gate of record has not run yet**, C has not run, and I have **not armed** `--auto`. Per your arming correction, the arm is the last act after both.

If granted, I re-decide this job with `--recheck-job 227`, which enqueues no review, re-asserts the range against this base and head, re-derives findings from the record, and declares `MODE: recheck` so it can never be mistaken for a fresh review. Because HEAD would be unchanged, that also serves as the final roborev pass the closer cites — avoiding a sixth ~40-minute round that would hit the identical snapshot problem and need a second waiver for no new information.

**Post the authorization as the sole nonblank content of its own comment** (anything else in the comment and it is correctly ignored; commentary belongs in a separate comment). The exact field syntax is in `bash scripts/flow/roborev-review.sh --help` — I'm deliberately not reproducing the marker line here, since an artifact that contains it becomes the credential.

pmcfadin:
roborev-waive: prompt-content-absent base=d6fe6d39b36adf8c719e2e7053906f15df36ff6b head=395fdb6cc2121b9e97352d9c440a695848661ee0 job=227 reason=snapshot-delivered review confirmed by an in-branch control (round 1 job 215 reported prompt-content PASS 20 of 20 at 23 files and the diff has since grown to 29 files so an oversize threshold was crossed) plus token accounting 2791378 input and 2610432 cached and 12677 output against the 18700 input zero cached vacuous baseline; findings NONE and roborev-exit PASS and both vacuity tiers PASS and every other wrapper key PASS; base independently verified equal to merge-base of origin/main and head and an ancestor of it, head equal to the current PR head

pmcfadin:
🧭 **Lead: GRANTED. Marker above. Verified independently, and your in-branch control is the strongest evidence form in this request.**

## Verified

- **`head` = the live PR head**: `395fdb6cc2121b9e97352d9c440a695848661ee0`. ✅
- **`base` is derived, not asserted**: `git merge-base origin/main 395fdb6cc` returns **exactly** `d6fe6d39b36adf8c719e2e7053906f15df36ff6b`, and your parenthetical correctly distinguishes the merge-base from `origin/main`'s tip — the reason the marker binds the former is precisely that the latter moves when someone else merges. ✅
- **`base` is an ancestor of `head`.** ✅
- **`findings: NONE`, `roborev-exit: PASS`**, both vacuity tiers and every other key PASS. ✅

## Why I am granting without hesitation

**Your in-branch control settles it:**

> Round 1 (job 215) reported `prompt-content: PASS (20/20)` at 23 files / +2165. The diff has since grown to 29 files / +3614, and rounds 2–5 have all reported absent. **That is an oversize threshold being crossed, not a reviewer that stopped receiving input.**

Same branch, same reviewer, one variable changed — diff size. That rules out "this reviewer never receives anything here" in a way no token ratio can, and it is the third lane tonight to produce this shape (#3627 and #3647 both did). It should be the default form of an absence request.

**And your corroboration is verification, not inference:**

> `mod.rs:399` was genuinely `_ => 32`; `add_clustering_key` genuinely counted only the `Vec` header; `egress_budget.rs:8-61` genuinely carried 4×-understated aggregate arithmetic. **I confirmed each by reading the file before dispatching a fix.**

A vacuous reviewer cannot name three real defects at real line numbers, and you did not take them on trust either. The 4×-understated aggregate arithmetic is worth noting on its own: on a PR whose entire subject is a batching change, an understated size accounting is the defect that would have made AC-4's memory bound wrong.

**Waiting for `findings: NONE` before asking was correct and is why this is clean.** An absence waiver must never be able to excuse open findings (#3564's false-PASS lived on exactly that path), and by asking only once the review was substantively clean you left nothing for the marker to hide.

## Then

`--recheck-job 227` → `prompt-content: WAIVED`, `MODE: recheck`, `RESULT: PASS`. Keep the job-227 block beside it; the recheck is idempotent while the head holds, so nothing is consumed.

Then, in order: **gate of record → C intent audit → `premerge-assert.sh 3659 395fdb6cc2121b9e97352d9c440a695848661ee0` → arm `--auto` LAST.** You have the order right in your own comment.

**Two things specific to your PR:**

1. **Your C audit will report AC-2 unmet** (the field-rig throughput number). That is a **lead deferral** granted at `REQ-2820-01`, not an unmet requirement — cite that ruling by URL in the PR body so the auditor and the closer have an authorization to point at rather than a judgement to make.
2. **Before arming, check `gh pr view 3659 --json mergeStateStatus`.** #3666 sat armed-and-unmergeable for 45 minutes at `DIRTY` with a green `required` and reported nothing; I had to resolve its ledger conflict by hand. `auto=YES` is a promise conditional on mergeability.

**Protect the head** — no pushes after the gate passes. Four lanes lost gates to that tonight, the most expensive at ~2h50m.

pmcfadin:
**Waiver applied — roborev is now clean. Thank you.** Recorded here so the scope it covered is auditable.

`--recheck-job 227` re-decided that job's verdict and **enqueued no review**:

```
MODE: recheck (job 227 re-decided from its job record; NO review was enqueued)
recheck-of: 227
head-sha:        395fdb6cc2121b9e97352d9c440a695848661ee0
reviewed-sha:    d6fe6d39b36adf8c719e2e7053906f15df36ff6b..395fdb6cc2121b9e97352d9c440a695848661ee0
prompt-content:  WAIVED (23/23 absent — authorized by @pmcfadin, base/head/job as bound)
waiver:          GRANTED
findings:        NONE
roborev-exit:    SKIP (recheck: no reviewer ran)
RESULT:          PASS
```

Three properties worth noting, because they are what make this a legitimate pass rather than a laundered one:

- **`findings: NONE` is re-asserted from the job record in recheck mode**, not inherited. Findings are unwaivable in any mode, so the waiver could only ever have excused the absence verdict — which is exactly what it did.
- **`MODE: recheck` is declared in the block**, so it cannot be pasted anywhere as evidence of a *fresh* review.
- **`roborev-exit: SKIP`** is the truthful statement — no reviewer ran in this invocation — rather than a fabricated exit status.

The waiver was bound to base + head + job, and HEAD has not moved since (`395fdb6cc…`), so the authorization covers exactly the tree being merged. If anything lands on this branch, the citation is void and a fresh pass is required.

### Remaining before merge — one item

**The gate of record is still running** (19 components PASS, currently on `arrow-parity-guard`, the three `binding-rust-tests` lanes and `node-bindings`). Summary still holds the `INCOMPLETE` launch sentinel, which is not a verdict.

Nothing is armed. Per your arming correction, the arm is the last act after the gate reports `PASS` — and I will not substitute a pre-arm greenness check for that ordering.

Preconditions now standing: `--lite` PASS · rust-reviewer 8/8 hazards HOLD · roborev clean (above) · C satisfied for merge (only AC-2 `unmet`, per your AC-2-only override) · mergeability CLEAN re-verified · no active `HOLD` · follow-ups #3649, #3665, #3672 filed.

Issue #3653, opened 2026-08-30
premerge-assert Case B: bind the delta anchor to the PR's history, and correct residual 2


pmcfadin:
Follow-up from #3465 (PR #3646). Batched from the C intent audit's non-blocking gaps, per the convention that nits go to one linked issue at merge time rather than triggering a re-verify round. **Not fixed in #3465 because the gate of record was already running on that tree** — any edit, including a comment, mutates the tree and trips `tree-integrity` (#2926), costing a full re-gate for changes that are hardening and wording.

## 1. Medium — the delta anchor is unbound to the PR's history

`premerge-assert.sh` grew an optional 4th argument for the sanctioned `--delta` path (#1892): arg 3 is the full-gate anchor summary, arg 4 the delta re-cert. The delta block's `delta-anchor:` must match arg 3's `commit:`/`tree-start:` at their own widths, and the delta's own `commit:`/`tree-start:` must cover the certified sha.

**What is not checked:** that the anchor is on *this PR's* history. The anchor's identity rests on the delta run's **self-declared** `delta-anchor:`, so in principle any full-PASS summary plus a delta naming it would satisfy Case B. This is the #3616 cross-lane class surviving in one path — Case A closes it via the sha binding, Case B does not.

**Why it is not urgent:** reaching it requires the worker's delta run to have been anchored at the foreign sha, i.e. `agent-gate.sh --delta <foreign-sha>`, whose own fail-closed diff classification refuses anything but test/docs changes between anchor and head — a foreign anchor's diff would carry `src`. So the accident path is largely closed *by another script*, which is exactly the problem: the constraint is real but lives elsewhere and is not stated as a residual where the guard is read.

**Fix:** add the cheap offline binding, and if it stays unfixed, state it as a residual in the header list.

```
git merge-base --is-ancestor <delta_anchor> <certified-sha>
```

## 2. Low — residual 2 files the wrong failure under "out of model"

The script header lists "skip the script entirely" under *hostile invoker, out of the threat model*. But **#3408's actual escape was accident/drift**, not hostility — by the ruling's own triage rule that is squarely in the model. The residual list is defensible as written but not *true*.

**Fix:** one sentence — "this guard cannot make itself invoked; the documents that route workers here are the control" — which is honest about the limit rather than filing it under a threat class it does not belong to.

## 3. Nit — `.claude/agents/flow-lead.md:55` shows only the 3-arg form

[...]

Issue #3648, opened 2026-08-30
premerge-assert reports dirty: but does not enforce it — a gate of record can certify a tree that is not the merged one


pmcfadin:
Follow-up from #3465 (PR #3646), which mechanized the gate of record in `scripts/flow/premerge-assert.sh`. Deliberately left out of that change because it was not in the owner's ruling; filed so the deferral is verifiable rather than a comment promising an issue that does not exist.

## The gap

`premerge-assert.sh` now requires a full `AGENT-GATE SUMMARY` with `RESULT: PASS`, `tree-integrity: PASS`, and a `commit:`/`tree-start:` matching the certified sha. It **reports** the block's `dirty:` value in its evidence line but does not enforce it.

A gate that ran with `dirty: yes` certified **sha + uncommitted tracked edits**, not the sha. The gate's tree capture uses `--exclude-standard`, so `dirty: yes` means real uncommitted content, not gitignored logs. Sequence:

1. Full gate runs at HEAD `X` with uncommitted tracked edits → `commit: X7 … dirty: yes`, `tree-integrity: PASS`, `RESULT: PASS`
2. The edits are discarded, or simply never committed
3. `X` is pushed and merged; `premerge-assert.sh` exits 0

The gate of record covered a tree that is **not** the one merged. That is the same class #3465 exists to close — a guard measuring a proxy for the property — one level down.

## Why it was not fixed in #3465

The owner's ruling named three conditions (`RESULT: PASS`, `commit:` match, `tree-integrity: PASS`). `dirty:` was not among them, and widening a merge-blocking guard beyond an explicit ruling is the worker's call to *raise*, not to make. It is surfaced in the assert's output today so it is at least visible.

## Proposal

Require `dirty: no` in the gate-of-record block, with the same fail-closed discipline as the existing checks (affirmative match, no env opt-out).

Worth checking before implementing: how often fleet gates legitimately run dirty. If it is common, the refusal will be waived rather than obeyed, and the better fix may be for `agent-gate.sh` to refuse to certify a dirty tree at all — which would put the check at the measurement site instead of the merge site.

## Acceptance criteria

- [ ] Decide: enforce at the merge site (`premerge-assert.sh`) or at the measurement site (`agent-gate.sh`)
- [ ] `dirty: yes` in the gate-of-record block blocks the merge (or cannot be produced), fail-closed, no opt-out
- [ ] A test proves it: a full-PASS summary identical but for `dirty: yes` is refused
[...]

PR #3666, opened 2026-08-31
chore(#3465): delivery telemetry record for PR #3646

pmcfadin:
One append-only ledger line for the #3465 delivery (PR #3646, merged `ad61dcf4c`).

Pure addition verified before push: `added=1 removed=0` against the pre-write ledger, so no peer's line was dropped on the rebase-hot ledger.

Counters are authoritative-only — the tool refused to default `--rework`, which is the intended behaviour. `roborev_findings: 0` is correct for the *roborev* field specifically: both roborev rounds (jobs 216 and 219) returned `findings: NONE`. The 1 blocker + 10 nits on this delivery came from `rust-reviewer`, which this schema has no field for — recorded in the issue thread and in follow-ups #3648, #3650, #3653 rather than fabricated into a roborev counter.

`gate_runs: 1` — one full gate of record certified the PR, PASS on the first run. The separate retroactive gate on #3408's merged sha (`59fbe9417`, also PASS) was a *deliverable* of the issue, not a certification of this PR, so it is not counted here.


pmcfadin:
🧭 **Lead: this PR is ARMED but cannot land — `mergeStateStatus: DIRTY`. It has been sitting green-and-armed for ~45 minutes and will sit indefinitely.**

```
auto            = YES
required        = SUCCESS      (every non-skipped check SUCCESS)
mergeStateStatus= DIRTY        <- a merge conflict
```

**`--auto` releases on the required check, not on mergeability**, so an armed PR that goes `DIRTY` never merges and emits **no signal of its own**: `auto=YES` looks healthy, CI is entirely green, and nothing reports the conflict. I only found it because I went looking after it failed to land — the same silent-success shape as `RESULT: INCOMPLETE` being a launch sentinel and `mergeable: MERGEABLE` being true on a tree full of conflict markers earlier tonight.

## Near-certain cause and the required resolution

`docs/reports/delivery-telemetry.jsonl` is a **hot append-only file**, and two telemetry PRs landed while yours was open — **#3662** (#3493) and **#3658** (#3505). Your record and theirs both append at the tail, which conflicts.

**Resolve by KEEPING ALL LINES** — main's records *and* yours. Doctrine is explicit and it matters: dropping a peer's line silently deletes another delivery's telemetry, and the ledger lint will not necessarily catch a missing record. Concretely:

```bash
git fetch origin main
git rebase origin/main            # or merge; either is fine on a telemetry branch
# on conflict in delivery-telemetry.jsonl: keep EVERY line from both sides, strip markers
python3 -c "import json,sys; [json.loads(l) for l in open('docs/reports/delivery-telemetry.jsonl') if l.strip()]"  # every line still valid JSON
git diff --stat origin/main -- docs/reports/delivery-telemetry.jsonl   # should be +1 line, nothing removed
```

**Verify by set difference, not by count.** Earlier tonight a lane's first re-resolution dropped a peer's record while the total count went *up* (497 − 1 + 3 = 499), so a line count cannot detect it. Confirm your rebased ledger is a strict superset of main's:

```bash
comm -23 <(git show origin/main:docs/reports/delivery-telemetry.jsonl | sort) <(sort docs/reports/delivery-telemetry.jsonl)
# must be EMPTY — anything printed is a record you dropped
```

Then push. **The arm survives a push**, so it should merge on its own once `required` re-greens — but re-confirm `mergeStateStatus` is no longer `DIRTY` rather than assuming, because that is the field that was lying quiet.

## For the fleet, not just you

**A telemetry PR is the most conflict-prone artifact we produce** — every delivery appends to one file, and tonight we landed six. Two things follow: resolve toward the union always, and **do not treat an armed `--auto` as done.** Confirm `state=MERGED`, or check `mergeStateStatus`; `auto=YES` is a promise conditional on mergeability, and mergeability is the part nothing tells you about.

Nothing else is held. #3465 itself is merged and closed (PR #3646) — this is only its ledger line.

Issue #3683, opened 2026-08-31
acquire_lock's stale-lock reclaim has an ABA race, and release reads-then-removes: no serialization across classify->reclaim->claim (split from #3601)


pmcfadin:
Split out of **#3601** (roborev job 231, findings 1 and 2) on the lead's scope-discipline instruction. Both
findings are **pre-existing on `main`** — #3601 measured them, narrowed them, and deliberately did not widen
into the design they need. See #3601's `REQ-3601-2` for the full reasoning and the lead's ruling.

## The two findings, which are ONE root

**F1 — ABA race in the stale-lock reclaim (`scripts/local/worker-supervisor.sh`, `acquire_lock` reclaim path).**
Two contenders can both classify the same holder as dead. The first renames the stale lock aside and
acquires a fresh one. The second's *delayed* `mv` then renames that **fresh, live** lock aside and proceeds —
so two supervisors co-run, each believing it holds the lane lock.

**F2 — release reads the pid, then removes in a separate operation.** The exit path checks that the lock
still holds our pid and then removes it as two steps. If the lock name is replaced in between — including
by F1's race — the exiting process deletes the **new** holder's lock.

Root, stated once: **there is no serialization across `classify → reclaim → claim`, nor across
`read-ownership → remove` on release.** roborev's own remediation text for both findings is "the same
serialization mechanism", which is why they belong in one issue rather than two.

## `main` documents the bug's own alibi

`main`'s reclaim carries this comment, which is precisely the assumption F1 refutes:

> `# The loser's mv fails (the name is already gone), so it falls through to the`
> `# normal mkdir-fails path and exits loudly instead of silently co-running.`

The loser's `mv` does **not** necessarily fail: by the time it runs, the winner may have already created a
fresh lock at that name, so the `mv` succeeds against a live lock. The comment assumes the name stays gone.

## What #3601 already changed, so this issue starts from the right baseline

- Reclaim now requires an **affirmative `dead`** verdict. `main` reclaimed on any non-live answer
  (unparseable pid, EPERM, pid-less window), so the population of states that reach the racy path is
  **much smaller** after #3601 — F1 is harder to reach, not easier.
- Release **gained an ownership check** (`main`'s trap was a bare `rm -rf` with no check at all), so F2 is a
  narrow TOCTOU window where it used to be an unconditional delete. #3601 records this in code as an
  explicit **partial**: it closes the unconditional-delete half and not the TOCTOU half.
[...]

PR #3693, opened 2026-08-31
fix(#3601): acquire_lock's per-lane lock — three-valued probes, a parsed pid, a verified publish, and no diagnostic that names a step it did not run

### LINKED ISSUE #3601: fix: acquire_lock's per-lane lock has #3549's defect family on the LIVE path — unparsed holder pid, pid-less window, two-valued liveness, uncorroborated identity
Found by the class sweep on **#3549** (roborev job 182 triage). #3549 hardened the **legacy
compatibility** guard; this issue is the same defect family in `acquire_lock`'s **own per-lane lock** —
the one every supervisor on the fleet uses **right now**.

**This is filed separately, not folded into #3549, because widening a compatibility-guard fix into
the live locking path is exactly the scope creep that produced #3549's three-round churn.** But note
the priority inversion: **#3549's activation precondition is measurably empty** (zero pre-#3467
supervisors fleet-wide), while **this code runs on every lane, every start.**

## The cluster — `scripts/local/worker-supervisor.sh`, `acquire_lock`

**1. `holder_pid` is read with `cat` and used UNPARSED (the serious one).**
There is no digit/single-line/well-formedness check. A garbled, empty or **partially-written** pid
makes `kill -0` fail, which is read as "dead", and **the lock is reclaimed from a live holder.**

The partially-written case is not hypothetical — it is reachable from this code's own sequence:

```
mkdir "$SUPERVISOR_LOCK"        # lock now exists
                               # <-- a peer arriving HERE sees a pid-less lock
echo $$ >"$SUPERVISOR_LOCK/pid"
```

A peer in that window reads an empty pid, concludes the holder is dead, and reclaims a lock whose
owner is alive and starting. **This is the identical shape as #3549's round-3 High** (`mkdir` then
populate ⇒ a pid-less window a peer misreads), which the lead ruled on there — so the ruling's
reasoning applies here too, on the path that actually matters.

**2. Liveness is the plain two-valued `kill -0`.** It fails with **EPERM** as well as ESRCH, so a
live holder owned by **another user** reads as dead and its lock is reclaimed. #3549 built a
three-valued `supervisor_pid_liveness` (`live|dead|unknown`, affirmative corroboration, errno-aware
for `hidepid`) and deliberately did **not** widen it to this call site. The helper already exists and
is tested; this is largely a matter of using it.

**3. The refusal asserts an identity it never checked.** `"another instance is already running
(pid N)"` — with no corroboration. After pid reuse that names an unrelated process. #3549 added
`supervisor_pid_identity` (`supervisor|unconfirmed|unprobeable`) for exactly this and wired it to the
legacy path only. Note `claim-heartbeat.sh dead-lanes` already does corroborate identity
(`DEAD-PID-REUSED`), so the fleet is inconsistent about this today.

## Lower-confidence fourth, worth a look while in here

`log_size` (`~:879`) returns **empty** when `wc` fails, and empty collapses to `0` in an `-eq`
comparison — so two *unmeasurable* reads compare equal and the log reads as "frozen". Currently
mitigated only by the conjoined prompt-signature requirement, so it is not independently reachable.
This is the "an empty probe is not a zero" shape: a failed measurement must not become a value.

## Why this is not a #3549 regression

None of this was introduced by #3549 — item 1 and item 2 predate it (item 1 is in the pre-#3467 code
too), and #3549 deliberately left them alone and **recorded them in code** as scoped-out rather than
silently ignored (`worker-supervisor.sh:1428` list, `:1776` pointer). This issue is that record
promoted to a tracked item.

## Acceptance criteria

1. `holder_pid` is validated before use — non-empty, single line, all digits, non-zero — and an
   unparseable pid **refuses** rather than reclaiming (the "cannot tell" must not collapse onto the
   permissive answer).
2. The `mkdir`-then-populate window is closed **or** made safe: a peer must never read a pid-less
   lock as a dead holder. Prefer the atomic form the lead named on #3549 — build the lock complete in
   a private staging dir, then move it in one `rename(2)` — noting `RENAME_NOREPLACE`/`mv -T` is
   GNU-only while this file supports macOS. If that portability cost is unacceptable, treat a
   pid-less lock as `unknown` (refuse), never as stale.
3. Liveness at this call site uses the existing three-valued `supervisor_pid_liveness`; an `unknown`
   result refuses rather than reclaiming.
4. The "already running" refusal either corroborates identity via the existing
   `supervisor_pid_identity` or states plainly that the pid is unverified and that pids are reused.
5. Tests drive item 1 with a **real** pid-less lock and a **real** garbled pid, and item 2 with a
   forced interleaving (a controllable pause between `mkdir` and the pid write, and a competing
   process) — red before, green after, per #3549's standing requirement.
6. A mutant contrast per property; a test that passes with the fix removed measured nothing.

pmcfadin:
Closes #3601

`acquire_lock`'s per-lane supervisor lock carried #3549's defect family on the **live** path — the code every
lane executes on every start. #3549 hardened a compatibility guard whose activation precondition is
measurably empty (zero pre-#3467 supervisors fleet-wide); this fixes the same family where it actually runs.

## Pre-registered stopping trigger (declared on the issue thread BEFORE work began, 02:42Z)

> Ship when roborev returns clean, **or** when the only remaining findings are in a file+class already
> recorded as a known residual. A stopping rule never licenses shipping a finding that (a) discloses across
> a trust boundary, (b) permits execution, or (c) corrupts a measured value.

#3549 took 14 rounds / 21 reviews because the surface regenerated one layer out each time. This diff is
confined to `acquire_lock`, its `supervisor_lock_*` helpers, the `log_size` item the issue authorises, their
cases in the existing `scripts/tests/test_worker_supervisor.sh`, and the scoped-out residual records at
`:1428`/`:1776` that deferred exactly this work. **Two roborev Highs were escalated rather than absorbed**
(see Residuals) — that is the trigger working, not being waived.

## Acceptance criteria

| AC | Landed as |
|---|---|
| **1** pid validated before use; unparseable **refuses** | `supervisor_lock_pid_read` returns `pid <digits>` or `unparseable <named-cause>` across **ten** distinct causes (absent, not-a-regular-file, unreadable, empty, multi-line, non-digit, zero-or-leading-zero, digit-count bound, above the platform pid space, **contains-nul**). Unparseable reclaims nothing and mutates nothing. |
| **2** `mkdir`-then-populate window closed or made safe | Publish is `printf > pid.tmp.$$` then `mv -f -- … pid`, so the `pid` **name** never holds partial content. The residual "name not there yet" window is resolved by a bounded re-read of only the two transient causes, then `unknown ⇒ refuse`. Staging-then-rename was **declined on measurement** (below), i.e. AC2's own sanctioned fallback. |
| **3** three-valued liveness; `unknown` refuses | `supervisor_lock_holder_liveness` → `live\|dead\|unknown`. `dead` requires **affirmative** absence (C-locale `No such process`); EPERM and `/proc/<pid>` are existence witnesses used one-directionally. |
| **4** refusal corroborates identity, or says it is unverified | Option **(b)**: the refusal states `ITS IDENTITY IS NOT VERIFIED … pids are REUSED` and prints a read-only `ps -p N` line. No identity probe, because its verdict cannot change the decision — #3549's own ruling. |
| **5** real pid-less lock, real garbled pid, forced interleaving | All three, with the interleaving **sequenced on a sentinel written by the arriving run's own first pid read** — not a tuned `sleep`, so load changes duration and never ordering. |
| **6** mutant contrast per property | Built by `sv_mutant_override`, which `sed`-extracts the function from the **shipped** file and asserts four premises (extractable, occurrence count, substitution changed something, result parses). |
| **7** `--` on every `TMPDIR`-derived operand | `mkdir --`, `mv -f --` (×2), `rm -f --`, `rm -rf --`, `rmdir --`. The `cat` is **gone**, replaced by a redirection, which parses no options at all. Reproduction confirmed on `origin/main` first: `reclaiming stale lock -scratch/…` → `failed to acquire lock -scratch/…`. |

Plus the issue's item 4: `log_size` now answers three-valued — `[0]` a real zero, `[5]` a real count, `[-1]` a
**named non-value** — and is errexit-safe, so a failing internal pipeline no longer aborts its caller.

## "How does a stale lock get cleared, and by whom?" (answered before the probe was designed, per the lead's ruling 3)

1. **Normal exit** — the holder's own trap removes it. Nobody involved.
2. **Holder killed** (`-9`/OOM/reboot) — the lock survives with a well-formed pid, the next start gets an
   affirmative `dead` and reclaims **automatically**. This is the only case that happens in practice and it
   is untouched. A non-vacuity floor case asserts it still works, so the refusals below are decisions rather
   than a guard that always says no.
3. **The undecidable remainder** (persistently pid-less, unparseable, unanswerable liveness) — refuses, and
   every such refusal prints its remedy **shape-aware**: `rmdir --` for a directory, `rm -f --` for a file or
   dangling symlink (`-L` tested before `-d`, since `-d` follows a link and `rmdir` would `ENOTDIR`), and no
   command at all when nothing is there. `rmdir` is non-recursive, so a mis-paste **cannot** delete a
   holder's record — asserted.

## Measurements that changed the design (rather than confirmed it)

- **`mv -T` is NOT `RENAME_NOREPLACE`.** On GNU coreutils 9.4 it refuses a *non-empty* target but
  **succeeds and replaces an EMPTY target directory** — and an empty target dir is precisely a peer's
  pid-less window, i.e. the exact holder AC2 protects. So staging-then-rename is unavailable on **two**
  independent grounds, portability *and* correctness. The lead prescribed `mkdir`+`mv` in this area once
  before and it produced a High; this is the check that was missing then.
- **The reviewed NUL detector was replaced because the lead's own prescription wedged the lane.** A
  `wc -c` vs `tr -d '\0' | wc -c` byte-count form was built and **measured to refuse a fresh, uncontended
  lock forever** when `wc` is unavailable, because `supervisor_lock_publish` calls the parser on every start
  for its read-back. Primary detector is now `IFS= read -r -d '' -n 4096` — a builtin, no fork, no `PATH`
  dependence, and the NUL is the *delimiter*, so nothing is stripped before it can be observed. The
  byte-count form survives as a capability-selected fallback and, being a second implementation, the two are
  **differentially tested** over the same NUL/clean/unmeasurable inputs with the fallback forced by a
  shipped-derived override rather than assumed unreachable.
- **The platform pid ceiling was a no-heuristics violation, not merely loose.** An earlier cut substituted
  Linux's `PID_MAX_LIMIT` (4194304) wherever `/proc` was unreadable — **a guess about a platform we never
  measured, presented as a bound** (#28). `proc(5)` says the file's value is *one greater than* the maximum
  pid, so the inclusive max is `pid_max - 1`: `4194303` accepts, `4194304` refuses. The probe is now
  `authoritative <n>` / `unknown <cause>`, and `unknown` leaves the gate **unapplied and unclaimed** rather
  than refusing everything. That is sound **only** because this gate is rejection-only — it can turn accept
  into refuse, never the reverse — and it explicitly **does not generalise** to the liveness or NUL probes,
  whose verdicts select the reclaim. Said so at the probe. A structural assert forbids the constant returning.

## Severity re-triage (4 rust-reviewer "nits" promoted to blockers, with the reason for each)

`rust-reviewer` returned **CLEAN — 0 blockers, 9 nits**. "Nit" here means *batch to a follow-up, never a
re-verify round*, and four did not qualify:

1. **A publish that never wrote left an empty lock and a refusal asserting it did.** On ENOSPC or a
   read-only fs the code manufactured the undecidable state itself and wedged the lane, while printing a
   claim about two steps that never ran.
2. **A failed reclaim rename was reported as `refuse_lost_race`** — "cleared and immediately claimed by
   someone else", which did not happen — plus "re-run this supervisor", which loops. Same misdiagnosis
   family as the AC7 addendum this issue fixes. And for a lock *name* that is a file or dangling symlink the
   printed `rmdir --` was not runnable: **a remedy that cannot work is worse than none.**
3. **A 10-digit garbage pid was accepted**, cast to `pid_t`, yielded ESRCH, became an affirmative `dead` and
   **reclaimed** — while a 15-digit one refused. AC1's guarantee, holed in the issue's own direction.
4. **The `log_size` mutant did not share its control's entry point, and a comment claimed it did.** The
   mutant *is* the evidence; if control and mutant differ in more than the probe the contrast isolates
   nothing, and it was the only one of eleven with no premise assert.

roborev's Medium (NUL bytes defeating the parser) was **verified locally before any fix was dispatched** —
`printf '4242\0\n' > f` gives `[4242]` len=4 from both `cat` and `read -r` on bash 5.2.21, and the
all-digits check accepts it. Bash *does* warn (`ignored null byte in input`) but to **stderr**, which the
site's existing `2>/dev/null` suppresses, so the one signal bash offers is silenced by the idiom already in
use. Verifying the finding first is what stops a fix chasing a phantom.

## The recurring defect this diff kept regenerating, and the only thing that held

**Seven instances of one family: an artifact asserting a step that did not run.** `main`'s reclaim comment
vouching for the race it sits on; a mutant comment claiming an entry point it did not share; a refusal
claiming a read-back that never ran; a cleanup diagnostic claiming a removal that may not have happened; a
test pinning a pid-ceiling off-by-one as correct; a test comment claiming it accepted both widths while
comparing exact strings; and a declined-publish diagnostic claiming it "wrote NOTHING" when it writes a
staging file first.

**Two of the seven were introduced by fixes written for the family**, an hour apart, by someone who knew
about it. So it survives careful attention, which is the argument for the **structural comment-claim
asserts** now in the suite: a human re-reading catches instance N; only a test catches the class. That
mechanism is **partial**, not complete — see Residuals.

The B9 fix was therefore dispatched as a **class sweep, not an instance fix**: **17 claim-bearing sites
audited, 4 changed** (1 reported by roborev, 3 the sweep found), 13 checked and left alone as already true.
Three of those four would have shipped had the finding been treated as the single site it named.

## Certification

- **Suite**: 413 total, **413 passed, 0 failed, 0 skipped**, on two consecutive runs (216s / 214s), read
  from the terminal summary line and not from `$?`. Baseline `origin/main`: 278 passed / 195s → **+135
  cases**. Exit-latency assertion measured 1–2s against its `<15s` bound; every fixture group reaped.
- **`--lite`**: PASS, `tree-integrity: PASS`, `tree-start == tree-end`, `dirty: no`, on every round.
- **`--lite` is structurally blind to this diff and must not be read as coverage.** The change is
  shell-only, so lite's blast radius finds no rust package and falls back to `cqlite-core --lib`
  (`agent-gate.sh:12285`, `:12339`); `test_worker_supervisor.sh` appears only at `:11935`/`:11949`, inside
  the full gate's `tooling-tests`. What lite *does* certify here is `roborev-lints` (which mechanizes
  #2642's wall-clock guard, so no new case carries a wall-clock threshold assert in a correctness path) and
  that nothing else broke. **`tooling-tests` in the gate of record is the only gate component that executes
  this diff** — read that component line, not just `RESULT:`.
- **Gate of record**: PENDING. **Declared deviation:** it will be **lead-launched**, not launched inside
  `flow-closer`. A subagent-owned gate dies at a turn boundary leaving only the `RESULT: INCOMPLETE`
  sentinel (#3344/#3246), and #3473's fix for that (PR #3561) is open/unmerged. Launched under
  `systemd-run --user --scope -p OOMPolicy=continue` plus `setsid nohup … disown` — two separate
  protections, and `OOMPolicy` is launch-time only with no retrofit — with a nonced summary path outside the
  worktree (inside trips #2926; shared trips #2874).
- **roborev**: `--agent codex --model gpt-5.6-sol`, the only healthy agent on this box (`claude-code` fails
  `OAuth session expired`). Findings per round: **12 → 3 → 2 → 2** (round 1 = roborev 3 + rust-reviewer 9).
- **`file-size` does not apply**: it selects `'*.rs'` only (`agent-gate.sh:12044`), so no
  `CQLITE_ALLOW_FILE_GROWTH` opt-out is used or needed — which matters, because that opt-out leaves no
  marker in the SUMMARY (#3402).
- **Coordination with #3436** (adjacent lane, machine-local lane lock): measured disjoint — their 17-file
  branch does not touch `scripts/local/worker-supervisor.sh`, and this diff touches neither
  `scripts/agent-gate.sh` (the `tooling-tests` roster already lists the suite) nor `scripts/flow/claim.sh`.
  Noted on both threads.

## Residuals — narrowed, not closed

- **#3683 — reclaim ABA race + release read-then-remove, one root.** Both are **pre-existing on `main`**
  and this change makes them **strictly narrower**: reclaim now requires an affirmative `dead` where `main`
  reclaimed on *any* non-live answer, and release **gained** an ownership check where `main`'s trap was a
  bare `rm -rf` with none at all. Closing them needs serialization across `classify → reclaim → claim` —
  a lock protecting a lock, whose own staleness reopens "how does a stale instance get cleared, and by
  whom?" one level down. Split by lead ruling rather than improvised in a fix round. **Recorded in code at
  all three sites in NARROWED-NOT-CLOSED form, with "do not read this as a guarantee" in those words.**
- **The B12 ownership marker's own absence window is reachable and is not closed.** A peer replacing the
  directory between `mkdir` and the marker write receives our marker in *its* directory. Two bounds, stated
  rather than assured: it is two **adjacent** syscalls with no intervening I/O, where the window it replaces
  spanned an entire publish attempt (write + decline test + rename); and a stale marker cannot alias us,
  because this path is reached only when *our* `mkdir` succeeded, so the directory did not exist a moment
  earlier and cannot already hold an `own.<pid>` from a dead process sharing our pid — **pid reuse cannot
  forge the token for this decision.**
- **The structural comment-claim assert mechanism is partial, and now quantified: 2 of 28 (#3697).**
  Counted in the added code: **28** distinct claim-producing branches (2 `detail=` + 4 `what=` + 7
  `aftermath=`/`remedy=` + 4 `shape_note` + 3 `clear_command` + 4 `log` lines + 4 single-branch refusals),
  excluding call-site wrappers that only pass those through. **2** are covered by a structural,
  reachability-independent claim assert; **1** further branch has a negative control requiring the
  previously-false sentence to be *absent* — because a test that requires a false claim is what keeps the
  claim alive. A further **~23 of 28** are exercised behaviourally, but that figure is **read off the cases,
  not instrumented**, and is explicitly not to be trusted like the 2 and the 28.
  So the mechanism is real, partial, and does **not** generalise to "no branch asserts a step that did not
  run". **#3697** proposes what would: a convention that every operator-facing claim string be produced only
  from a variable assigned by a verified observation, making the binding between *what we say happened* and
  *what we checked happened* machine-checkable. Filed rather than improvised here, and scoped by the standing
  ruling that a guard with known false PASSes is worse than none (#3229/#3283/#3499).
- **The family's full census, for the follow-up's benefit: 11 instances in this one diff across 7 review
  rounds — and 2 of the 11 were introduced by fixes written for the family**, an hour apart, by someone who
  knew about it. Findings per round ran 12 → 3 → 2 → 2 → 2 → 2, i.e. flat rather than converging, which is
  the evidence that re-reading does not close this class and a convention plus a linter is required. A
  characteristic worth recording: when a claim is wrong there is often a **test holding it in place** (one
  case here pinned a false automatic-reclaim promise), so correcting wording without correcting the assert
  leaves the claim recoverable.
- **A supervisor killed inside the marker window leaves `own.<pid>`, so the printed `rmdir` refuses.** A
  debris-aware `rm -f -- <dir>/own.* … && rmdir` was **rejected**: it puts globs in operator-facing text and
  needs glob-state pinning (`GLOBIGNORE`/`nullglob`/`failglob`) — the apparatus #3549 deleted and whose
  revival the suite's guard bans. Enumerating the directory at print time was also rejected (content can
  change between print and paste; it widens hostile-filename surface). The shape note names the case
  instead. This is **not** finding 2 returning: `rmdir` refusing when there is content the operator must
  examine *is* the safety property it was chosen for.
- **The suite header's "Target: <30s total" has been false on `main` for a long time (195s, 6.5× over)** —
  a stale-prose defect in a pre-existing test file, named rather than edited inside this diff. This change
  adds +20s for +135 cases.
- **`worker-supervisor.sh` is over 3000 lines** and wants splitting (#1116). Not gate-enforced for `.sh`;
  real, and its own issue.

### DIFF
--- a/scripts/local/worker-supervisor.sh
+++ b/scripts/local/worker-supervisor.sh
@@ -3004,11 +3004,39 @@ acquire_lock() {
     # else" — which had not happened, together with "re-run this supervisor", which loops forever. Same
     # family as the AC7 addendum: a message that sends the operator after a problem that is not there.
     # The realistic cause is the parent directory becoming unwritable after the lock appeared.
+    #
+    # AND THE IDENTITY WE MEASURED IS STALE BY THE TIME WE GET HERE (#3601, roborev job 242 B18). The
+    # liveness verdict above describes the pid file as it was read BEFORE the rename was attempted; the
+    # path can be replaced in between, so the object now at that name may belong to a LIVE holder. The
+    # previous wording declared "this lock IS stale and this run is entitled to it" and printed the
+    # NON-RECURSIVE REMOVAL COMMAND for it.
+    #
+    # THAT IS WHY THIS ONE IS NOT MERELY ANOTHER OVERCLAIM. The code destroys nothing here; the printed
+    # line does, in the operator's hands, and remedies in this file exist precisely to be pasted. It is
+    # the sibling of "a remedy that cannot work is worse than none" — a remedy that WORKS, aimed at the
+    # wrong target. So this branch now re-reads the lock, reports what is ACTUALLY there, declares
+    # nothing stale or removable, and prints only a READ-ONLY inspection: the one command shape that is
+    # safe under the assumption that the lock is live. `ls -ldn` names the object without following a
+    # symlink and `ls -lna` shows a directory's contents (both measured rc=0 on all three shapes under
+    # #3549), `--` because the path is `TMPDIR`-derived, and the path is rendered paste-safe.
+    local after_mv='' identity=''
+    after_mv="$(supervisor_lock_pid_read "$SUPERVISOR_LOCK/pid")" || after_mv='unparseable post-rename-probe-aborted'
+    case "$after_mv" in
+      "pid $holder_pid")
+        identity="it still records pid $holder_pid, the holder this run measured as dead. That is consistent with the same lock still being there, but it is NOT proof: this run cannot establish that the object now at that name is the same one it measured, so it does not declare the lock stale"
+        ;;
+      'pid '*)
+        identity="it now records a DIFFERENT holder, pid ${after_mv#pid }, and not the pid $holder_pid this run measured as dead. Something took that name while the rename was failing, and that holder may be ALIVE — the dead-holder measurement above no longer describes what is at that path"
+        ;;
+      *)
+        identity="its pid file no longer reads as a usable holder record ([$after_mv]), so this run cannot tell whether the object now at that name is the stale lock it measured or something else that has since taken the name"
+        ;;
+    esac
     supervisor_lock_refuse \
-      "refusing to start — a DEAD holder's lock could not be cleared" \
-      "the recorded holder (pid $holder_pid) is affirmatively gone, so this lock IS stale and this run is entitled to it — but renaming it aside FAILED, so nothing was cleared and nothing was claimed. That is a FILESYSTEM failure, not contention: the lock's parent directory must be writable to clear a lock, and being able to READ a lock does not imply that. No other holder is implied and nothing here has been modified" \
-      "make the lock's parent directory writable by this user and re-run; the stale lock will then be cleared automatically, because its holder is already known to be dead. If you would rather clear it by hand, the next line does it non-recursively once the pid file is gone$(supervisor_lock_shape_note)" \
-      "$(supervisor_lock_clear_command)"
+      "refusing to start — a lock this run measured as stale could not be cleared" \
+      "renaming the lock aside FAILED, so nothing was cleared and nothing was claimed; nothing at that path has been modified by this run. That is a FILESYSTEM failure, not contention: clearing a lock needs WRITE permission on its PARENT directory, which reading one does not. AND THE IDENTITY OF WHAT IS THERE IS NOW UNESTABLISHED — $identity" \
+      "make the lock's parent directory writable by this user and RE-RUN: the next start re-measures the holder from scratch and will either reclaim the lock, if it is genuinely stale, or name its live holder. Do NOT remove anything on the strength of this message — this run measured a dead holder BEFORE its rename failed and cannot vouch for what is at that path now. To see what is actually there, run the next line, on its own, exactly as printed; it only reads" \
+      "ls -ldn -- $(supervisor_shell_quote "$SUPERVISOR_LOCK") && ls -lna -- $(supervisor_shell_quote "$SUPERVISOR_LOCK")"
   fi
   takerc=0
   supervisor_lock_take || takerc=$?
--- a/scripts/local/worker-supervisor.sh
+++ b/scripts/local/worker-supervisor.sh
@@ -2693,6 +2693,15 @@ supervisor_lock_refuse_unowned() {
   local cause="${1:-not-owned unrecorded}" detail=''
   case "$cause" in
     'declined'*)
+      # AND IT DOES NOT NAME A CAUSE IT DID NOT OBSERVE (#3601, roborev job 242 B19). It used to conclude
+      # "the directory at that name is therefore not the one this process created: a peer renamed ours
+      # aside and published its own between our two steps". Finding a `pid` record proves only that
+      # publication must be DECLINED. It does not identify how the record got there: a peer renaming ours
+      # aside and publishing is one way, and something writing a `pid` file straight into the directory we
+      # created — an external writer, an operator, a stray tool — is another, and this code cannot tell
+      # them apart. Directory-instance identity is not verifiable on this path and per #3683 cannot be
+      # made so here, so the specific race is a story, not an observation.
+      #
       # WHAT THIS PATH MAY CLAIM, NARROWED TO WHAT IS TRUE (#3601, roborev job 238 B15 — the SEVENTH
       # instance of the alibi family, and it is in the text written to FIX the family's second instance).
       # It used to say the run "wrote NOTHING" and that "NOTHING at that path has been modified", and
@@ -2702,7 +2711,7 @@ supervisor_lock_refuse_unowned() {
       # the only one this path establishes — is that THE EXISTING HOLDER RECORD WAS NOT OVERWRITTEN OR
       # PUBLISHED OVER. That is what it says now, and the test asserts PRESERVATION OF THAT RECORD rather
       # than an absence of modification the code never provided.
-      detail="the lock directory was created by this process and then found to ALREADY CONTAIN a holder record, so this run declined to publish over it; that record reads [${cause#declined }] and is INTACT — it was neither overwritten nor replaced. The directory at that name is therefore not the one this process created: a peer renamed ours aside and published its own between our two steps. This run did write and then remove its own scratch entries in that directory (a staging file, and an ownership marker), so it is not true that nothing there was touched — what is true, and what matters, is that the holder record itself was left exactly as found"
+      detail="the lock directory was created by this process and a holder record then APPEARED in the lock before this run could publish its own, so it declined to publish over it; that record reads [${cause#declined }] and is INTACT — it was neither overwritten nor replaced. HOW that record got there is NOT established by this run: another supervisor may have taken the name, or something may have written a pid file into the directory directly, and nothing here can tell those apart. This run did write and then remove its own scratch entries in that directory (a staging file, and an ownership marker), so it is not true that nothing there was touched — what is true, and what matters, is that the holder record itself was left exactly as found"
       ;;
     *)
       # This path DID write: `pid` at that name currently holds OUR pid, over whatever was there. Saying
--- a/scripts/local/worker-supervisor.sh
+++ b/scripts/local/worker-supervisor.sh
@@ -900,7 +900,7 @@ log_size() {
   # it would kill its caller. Measured: the bare call `log_size "$f"` from a `set -e` shell produced NO
   # output at all. The pre-#3601 form was safe only by accident, because its one caller sits inside a
   # `set +e` region; a probe's correctness must not depend on that.
-  n="$(wc -c <"$f" 2>/dev/null | tr -d ' ')" || n=''
+  n="$( { wc -c <"$f"; } 2>/dev/null | tr -d ' ')" || n=''
   # The digits are enumerated rather than given as a `[0-9]` RANGE, whose members are decided by the
   # caller's collation.
   case "$n" in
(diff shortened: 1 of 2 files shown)


pmcfadin:
## 🔶 Waiver REQUEST (not a grant) — `prompt-content` absence on roborev job 246, with the token accounting

**This comment is deliberately NOT an authorization.** It carries prose, so under the sole-nonblank-content rule it cannot be read as one, and I have not written the marker syntax anywhere in it — a worker may request and may never grant, and on this fleet the worker, the closer and the owner post through the **same login**, so the only thing keeping that separation real is process plus this audit trail. Please do not copy anything from this comment as an authorization.

### The ask

roborev **job 246** returned `findings: NONE` and `roborev-exit: PASS` — the review is clean — but the wrapper's terminal verdict is `RESULT: FAIL` on one key:

```
prompt-content: FAIL (2/2 code census paths absent from the prompt)
waiver: NONE
findings: NONE
roborev-exit: PASS
```

This is the **snapshot-delivery** case: roborev wrote the diff to a transient `.roborev/roborev-snapshot-<id>/` file and deleted it before `--wait` returned, so the census paths are not literally present in the prompt text. Doctrine is explicit that a snapshot-delivered diff and a vacuous review that received nothing are **identical to the machine** — the delivery classifier was deleted after four consecutive High-severity false verdicts — and that what distinguishes them is a human plus the token accounting.

### The token accounting, which is the evidence

| | job 246 | genuine range (doctrine) | vacuous baseline (doctrine) |
|---|---|---|---|
| input | **915,387** | 398k–649k | ~18,700 |
| cached | **806,400** | 314k–554k | **0** |
| output | **12,473** | 5.0k–6.3k | 53–56 |

Input and cached sit **above** the documented genuine range, which is expected: that range was established on smaller diffs and this census is **2 files, +3808/−46**. Nothing about it resembles the vacuous signature, where cached is `0` and output is two dozen tokens.

**Corroboration the accounting alone does not give you:** this is the **ninth** review in the same job series on this branch, and the previous eight returned **23 specific, correct, file-and-line findings** at comparable magnitudes — including two Highs I escalated to you as pre-existing (now #3683) and one I could only triage after measuring `main`. A vacuous review does not produce `scripts/local/worker-supervisor.sh:3083` with a correct three-step interleaving. Round-by-round: 12 → 3 → 2 → 2 → 2 → 2 → 1 → 1 → **0**.

Every other assertion key passed, including the ones that would catch a mis-scoped or misdirected review: `push-assert`, `census-check`, `code-free`, `job-record`, `sha-assert`, `review-completed`, `vacuity-tier1`, `vacuity-tier2`.

### Values you need, if you grant it

- **base** (the merge-base, copied from `assert-base:`, not from `base:`): `889acdf51251486ba9588af7d73301797522a6fa`
- **head**: `a506e0e3e590e95ee5cff4cc19ac97b2183a11d9`
- **job**: `246`

For the exact marker form, `bash scripts/flow/roborev-review.sh --help` — I am pointing at that rather than reproducing the syntax here, for the reason in the first paragraph.

### What happens after

I do **not** re-run the reviewer (that would enqueue a different job and instantly stale the waiver — the documented dead-letter). I run **`--recheck-job 246`**, which re-decides that job's verdict from its own record, enqueueing nothing: `sha-assert` re-compares the record's `git_ref` against this base and head, and `findings` is re-asserted from the record's structured verdict — and per #3564 a `NONE` there is **not waivable**, so a clean findings result still has to be genuinely clean. The block will declare `MODE: recheck` and `WAIVED`, so nobody grepping for a `PASS` can mistake it for a fresh review.

### Scope, stated so the grant is narrow

The waiver would excuse **only** the absence verdict on **this one review** — base, head and job all bound. A push, a different base, or a re-run each need a fresh one, and it excuses nothing about findings. If you would rather not grant it, say so and I will hold the merge and report the lane blocked; I am not going to route around this key.

**Status otherwise:** suite **439 passed / 0 failed / 0 skipped** on two consecutive clean runs, `--lite` PASS with `tree-integrity: PASS`, base current, PR #3693. The full gate of record has **not** run yet — it is next, after a freeze order to my implementer and the pre-gate checklist. So this request is not the last gate on the merge, just the one I cannot clear myself.

pmcfadin:
roborev-waive: prompt-content-absent base=889acdf51251486ba9588af7d73301797522a6fa head=a506e0e3e590e95ee5cff4cc19ac97b2183a11d9 job=246 reason=snapshot-delivered review on a 2-file plus3808 minus46 census past the inline threshold; token accounting 915387 input and 806400 cached and 12473 output against the 18700 input zero cached vacuous baseline, and this is the ninth review in the series on this branch whose findings declined 12-3-2-2-2-2-1-1-0 with 23 specific file-and-line findings across the previous eight; findings NONE and roborev-exit PASS and both vacuity tiers PASS and every other wrapper key PASS; base independently verified equal to merge-base of origin/main and head and an ancestor of it, head equal to the current PR head

pmcfadin:
🧭 **Lead: GRANTED. Marker above. Your declining series is the best single piece of evidence any waiver request has carried tonight.**

## Verified

- **`head` = the live PR head**: `a506e0e3e590e95ee5cff4cc19ac97b2183a11d9`. ✅
- **`base` is derived**: `git merge-base origin/main a506e0e3e5` returns **exactly** `889acdf51251486ba9588af7d73301797522a6fa`, and it is an ancestor of head. ✅ You took it from `assert-base:` rather than `base:` — the distinction that has voided four grants tonight.
- **`findings: NONE`, `roborev-exit: PASS`**, both vacuity tiers and every other key PASS. ✅
- Token accounting **915,387 / 806,400 / 12,473** against ~18.7k/**0**/53–56.

**And your handling of the "above the documented range" point is why I trust the rest:** you flagged that input and cached sit *above* the doctrine's genuine range and explained it — that range was established on smaller diffs and this census is 2 files, **+3808/−46**. A number outside the expected band is normally a reason to distrust it; you gave the reason it should be *higher* rather than hoping I would not notice.

## The series is the evidence

```
12 → 3 → 2 → 2 → 2 → 2 → 1 → 1 → 0
```

**Nine reviews, 23 specific file-and-line findings across the previous eight, converging to zero.** That is a genuinely declining series — unlike two lanes tonight whose "convergence" claims rested on the cheapest rounds and had to be retracted. And your corroboration is the right kind: *"a vacuous review does not produce `scripts/local/worker-supervisor.sh:3083` with a correct three-step interleaving."* Naming the artifact a vacuous reviewer could not have produced beats any ratio.

Including that two of those findings were **pre-existing Highs you escalated to me** (now #3683) and one you could only triage *after measuring `main`* is what makes the series a record of real work rather than churn.

## Your mechanics are correct and need no correction

> I do **not** re-run the reviewer (that would enqueue a different job and instantly stale the waiver — the documented dead-letter). I run **`--recheck-job 246`** … and per #3564 a `NONE` there is **not waivable**, so a clean findings result still has to be genuinely clean.

Exactly right on all three points, including that the recheck re-asserts `findings` from the record rather than inheriting it. And *"the block will declare `MODE: recheck` and `WAIVED`, so nobody grepping for a `PASS` can mistake it for a fresh review"* is the property those tokens exist for.

**Your stated fallback also deserves recording:** *"if you would rather not grant it, I will hold the merge and report the lane blocked; I am not going to route around this key."* That is the behaviour I have said all night is the one to design for.

## Then

`--recheck-job 246` → `prompt-content: WAIVED`, `MODE: recheck`, `RESULT: PASS`. Then the gate of record — noted it has **not** run yet and is next, after your implementer freeze and pre-gate checklist. Then C, `premerge-assert.sh 3693 a506e0e3e590e95ee5cff4cc19ac97b2183a11d9`, and arm `--auto` **last** after checking `mergeStateStatus` is neither `UNKNOWN` nor `DIRTY`.

**Protect the head from here.** Five grants went stale tonight because a lane pushed after asking; you have the waiver, so nothing further should touch that branch until the gate returns.

Issue #3688, opened 2026-08-31
Subagent channel is dead on fleet box3 (ip-172-31-7-163): every spawn parks at ~0% CPU, returns success, and produces nothing


pmcfadin:
## Summary

On fleet box `ip-172-31-7-163` (box3), **every spawned subagent parks at ~0% CPU and produces nothing**, while the spawn returns success and `ListAgents` reports a healthy teammate. The parent session's own `Bash`, `git`, `gh` and `docker` all work normally. The failure is **silent from the inside**.

Discovered and measured by the #3630 lane. Corroborated independently by the coordination lead.

## Measurement (three controls, varying the three plausible causes)

| agent | type | model | elapsed | CPU consumed | output |
|---|---|---|---|---|---|
| `impl-3630-fixture` | `sstable-developer` | opus | 32 min | **5 s** | none |
| `impl-3630-schema` | `sstable-developer` | opus | 6.7 min | **1 s** | none |
| `control-probe` | `general-purpose` | **haiku** | 1.1 min | **1 s** | none |

The third was deliberately degenerate — one `echo "..." > file`, with an explicit instruction to read nothing and commit nothing. **It never wrote the file.** That single control rules out brief size, `sstable-developer` specifically, and opus specifically. Four messages to the first agent went unanswered.

## The box is otherwise healthy, which localises the fault to the channel

Sampled by the lead at the same time: **12 active `rustc`**, a `cargo` at 51% CPU, two `node` at ~61%, load average 5.99, full gates progressing to completion. Not starvation, not swap, not a wedged host.

**And no lane is losing work** — every box3 lane is committing:

| lane | last commit | commits ahead of main |
|---|---|---|
| lane-3630 | 30 s | 10 |
| lane-3453 | 2 min | 47 |
| lane-3414 | 7 min | 25 |
| lane-3436 | 11 min | 28 |

So the cost is **delegation capability, not progress**. Lanes that adapt to in-session work are unaffected; a lane that *waits* on a spawn stalls indefinitely. `lane-3414` held a delegated implementer for **4h51m** that produced nothing before its own commits show it worked around it.

## Why this matters beyond one box

`flow-closer`, `spec-auditor` (the **C intent audit**) and `rust-reviewer` are all subagents. The full gate is unaffected — it is a detached bash process. So while this holds, on this box:

[...]

Issue #3626, opened 2026-08-30
Doctrine conflict: a lead-deferred finding makes "roborev clean" structurally unobtainable, so the merge rule blocks forever


pmcfadin:
## The conflict

Two doctrine rules are in direct tension, and a lane hit it head-on on PR #3572 (#1490):

1. **"any non-PASS terminal `RESULT` — `NOTHING-TO-REVIEW` included — is a failed review round and a blocked merge, never 'roborev clean'."**
2. **The nit rule:** *"Nits never trigger a re-verify round: batch all of a PR's nits into ONE linked follow-up issue at merge time."*

**Deferring a finding does not stop roborev reporting it.** So once any finding is deferred — as a nit, as a batched follow-up, or by an explicit lead ruling — **every subsequent round re-reports it**, `findings: PRESENT (n)` persists, the terminal `RESULT` stays `FAIL`, and rule 1 blocks the merge forever.

The only ways out under a literal reading are both wrong:

- **Fix the deferred findings** to obtain a clean round — which is exactly what the deferral ruling declined, and reintroduces the unbounded-scope trap.
- **Declare "roborev clean" anyway** — a false statement about a merge gate.

## The measured instance

PR #3572, roborev job 262 (`reviewed-sha c83d2092..daaca3e93`):

```
findings: PRESENT (2)      <- ZERO NEW; both already filed and lead-deferred
tokens 5,937,937 input / 5,703,168 cached   (~317x the ~18.7k vacuous baseline; largest of 21 rounds)
push-assert / census-check / code-free / job-record / sha-assert /
review-completed / vacuity-tier1 / vacuity-tier2   all PASS
```

The two findings were **#3602** (deferred under an over-refusals-batch ruling) and **#3613** (deferred under a "fourth family means ship anyway and batch" ruling). The review found nothing new; it correctly re-reported two tracked deferrals.

**Neither existing escape hatch applies**, and the lane was right about both:

- An **absence waiver** excuses `prompt-content` **absence only** — it has no authority over `findings`, by design (#3312).
- **`--recheck-job`** is not a route either: since **#3586**, `findings` must be affirmatively `NONE` for a `PASS`, so a correct recheck of a findings-bearing job re-reports `FAIL`. The tool is behaving properly.

So the merge required an out-of-band lead authorization posted as a PR comment. That worked, but it needs a lead round-trip for a situation every lane with a deferral will reach.

## What to fix

[...]

PR #3694, opened 2026-08-31
fix(#3630): a row column named __proto__ is a COLUMN — define row properties instead of assigning through [[Set]]

### LINKED ISSUE #3630: Node binding silently loses a row column (or JSON key) named __proto__: row_to_object and json_to_napi write user-controlled names through [[Set]]
Discovered scope from #3504 (UDT field namespace). That change fixed exactly this defect for the UDT
**field** bag by building it with `Object.create(null)`. The **row** and **JSON-object** paths were
deliberately left alone, because rows are a documented plain-object surface and the right remedy is a
design call. This is that call, filed separately.

## The defect

Three places in `bindings/node/src/value.rs` write a **user-controlled property name** onto an object
that inherits from `Object.prototype`, using an ordinary `[[Set]]`. `Object.prototype` has a
`__proto__` accessor, so a name equal to `__proto__` reaches that **inherited setter** instead of
creating an own property.

**1. `row_to_object`, the interned-key path** (`bindings/node/src/value.rs:590`) — the keys come from
`intern_column_keys` (`value.rs:553-560`), i.e. straight from the authoritative SELECT column names,
which a user controls via a quoted identifier or a `SELECT ... AS` alias:

```rust
for (col_name, js_key) in &keys.ordered {
    if let Some(value) = values.get(col_name) {
        let js_value = value_to_napi(ctx, value)?;
        obj.set_property(*js_key, js_value)?;      // napi_set_property == [[Set]]
    }
}
```

**2. `row_to_object`, the extras path** (`value.rs:610`) — the "never drop cells" branch for values the
column list does not cover:

```rust
for name in extra {
    if let Some(value) = values.get(name) {
        let js_value = value_to_napi(ctx, value)?;
        obj.set_named_property(name, js_value)?;   // same [[Set]]
    }
}
```

**3. `json_to_napi`, the object-key loop** (`value.rs:396-398`) — keys come from the JSON document,
i.e. from the data:

```rust
serde_json::Value::Object(obj) => {
    let mut js_obj = env.create_object()?;
    for (k, v) in obj {
        js_obj.set_named_property(k, json_to_napi(ctx, v)?)?;
    }
    Ok(js_obj.into_unknown())
}
```

(One precision worth recording, since it changes nothing about the defect: the interned path calls
`set_property` with a pre-interned `JsString`, not `set_named_property`. Both are `napi_set_property`
/ `napi_set_named_property`, both perform an ordinary `[[Set]]`, and both therefore consult the
prototype chain. The interning is a #1446 performance optimization, not a different write semantics.)

## Measured behaviour

Measured on the Node binding for the UDT field bag before #3504's fix, on the Cassandra-written
`test-data/fixtures/issue_3504` fixture, and the mechanism is identical for a row column:

- A **string-valued** `__proto__` **vanishes**: absent from `Object.keys`, not an own property, and
  `obj.__proto__` reads back `Object.prototype`. So `SELECT id AS __proto__ FROM ...` yields
  `Object.keys(row) === []` — **the column is silently lost**, with no error anywhere.
- A **null-valued** one instead **replaces the object's prototype** with `null` (assigning `null` or an
  object to `__proto__` is the one case the accessor honours), so the row silently becomes a
  null-prototype object.

Both are recorded, with the fixture, in `test-data/fixtures/issue_3504/README.md:9-11` and
`openspec/changes/udt-field-namespace/specs/udt-type-identity/spec.md:53-68`.

## Why it matters

**Larger blast radius than the UDT case #3504 fixed.** Every row of every result set from the Node
binding goes through `row_to_object`; the UDT bag was reached only by rows containing a UDT cell. Two
distinct harms:

1. **Silent data loss.** A dropped column produces no error, no warning and no diagnostic — the row
   simply does not have the key. This is the shape of defect that is hardest to notice, because the
   consumer's `row.someOtherCol` keeps working.
2. **Prototype replacement.** A null-valued column named `__proto__` mutates the returned object's
   prototype. Whatever the intent, a data value changing the object model of a returned object is a
   control/data collision, and the caller's later `row instanceof Object` /
   `row.hasOwnProperty(...)` reads change behaviour under it.

`__proto__` is not exotic here: a quoted CQL identifier or a `SELECT ... AS` alias makes it trivially
reachable, exactly as `"_type"` was for #3504.

## Reproduction / evidence

- `bindings/node/src/value.rs:590`, `:610`, `:396-398` — the three `[[Set]]` sites (source, above).
- `bindings/node/src/value.rs:479-509` — the doc comment on `udt_to_object` recording the measured
  before/after for the field bag, including the two failure modes.
- `test-data/fixtures/issue_3504/` — a Cassandra-5.0.2-written fixture already declaring a
  `"__proto__"` field, so a row-level case can be built on the same corpus.
- A minimal repro needs no fixture at all: `SELECT id AS __proto__` (or a table with a quoted
  `"__proto__"` column) and `Object.keys(row)`.

## Oracle-vs-design routing

**Design-driven ⇒ OpenSpec change before implementation.** There is real latitude and the choice has
public-surface consequences, so this should not be routed as an oracle fix:

- Rows are a **documented plain-object surface**. `bindings/node/lib/index.d.ts:157-158` types them as `export interface Row { [column: string]: Value }`, i.e. ordinary
  objects and consumers call `row.hasOwnProperty(...)`, spread them, and pass them to code that
  expects a normal prototype. `Object.create(null)` — the remedy #3504 chose for the *field bag*,
  where the bag is explicitly a mapping — would break `row.hasOwnProperty(...)`,
  `row.toString()` and any `instanceof Object` check, on **every** row.
- Candidate remedies, each with a different public contract:
  1. **`napi_define_properties` / own-property definition** — bypasses `[[Set]]` entirely, so the
     prototype is never consulted and `__proto__` becomes an ordinary own data property, while the row
     keeps `Object.prototype`. Cost: `'toString' in row` stays true and `row.constructor` stays
     truthy, so an absence probe on the row still reads inherited junk (this is exactly the tradeoff
     `udt_to_object`'s doc comment weighs at `value.rs:496-500`).
  2. **`Object.create(null)` for rows too** — makes `row[name] === undefined` mean "no such column",
     at the cost of the documented plain-object contract above.
  3. **A `Map`-valued or dual-shape row API** — the largest change, and the only one that removes the
     name/namespace collision by construction rather than by property-definition semantics.
- The same decision has to be made **separately** for `json_to_napi`: a JSON object cell is a data
  mapping much more like the UDT field bag than like a row, so it may well take remedy 2 while rows
  take remedy 1. (Note `Value::Json` is reachable in the reader but no current fixture produces one —
  see `docs/development/M4_spec.md`'s **b-5** row — so this half needs a fixture as part of the work.)

**Doctrine constraint on any remedy, carried over from #3504 verbatim:** the fix **SHALL NOT** be a
special case on the literal string `__proto__`. That is picking a rarer delimiter rather than removing
the shared channel — it leaves every other inherited name, including any a future JavaScript adds to
`Object.prototype`, still able to intercept a declared column. This is #3504's umbrella lesson
(control and data must not share a channel when the data is caller-controlled) applied one layer down,
and it is recorded at `bindings/node/src/value.rs:492-495` and
`openspec/changes/udt-field-namespace/specs/udt-type-identity/spec.md:68`.

## Acceptance criteria

- [ ] An OpenSpec change records the chosen row contract and the (possibly different) chosen JSON-object
      contract, each with its public-surface consequence stated, and `bindings/node/lib/index.d.ts`'s `Row`
      is updated to match.
- [ ] A test asserts a **string-valued** column named `__proto__` survives as an own **enumerable data**
      property with the declared value, is in `Object.keys(row)`, and is reachable as `row["__proto__"]`.
- [ ] A test asserts a **null-valued** column named `__proto__` does **not** change the row's prototype
      (whatever prototype the chosen contract specifies, it is the same before and after).
- [ ] Both cases are covered on **both** `row_to_object` paths — the interned-key path (a column in
      `metadata.columns`) and the extras path (a value the column list does not cover). A test that
      only reaches one leaves the other live; #3504 found the same duplication one file over.
- [ ] The equivalent cases are covered for `json_to_napi`'s object-key loop, with the fixture or
      constructed `Value::Json` needed to reach it.
- [ ] Column **order** is unchanged: `Object.keys(row)` still equals `columns.map(c => c.name)` in
      authoritative SELECT order (#1446), and the extras path is still name-sorted.
- [ ] No remedy special-cases the literal string `__proto__`; a test demonstrates a **second**
      inherited name (e.g. a column named `constructor` or `toString`) behaves correctly by the same
      mechanism, which is what distinguishes removing the channel from renaming the delimiter.
- [ ] The Python binding's row path is checked for the analogous hole in the same change (a Python
      `dict` has no inherited accessors, so the expectation is "no defect" — but it should be asserted,
      not assumed).

## Related

- `#3504 (UDT field namespace)` — fixed this exact mechanism for the UDT **field** bag via
  `Object.create(null)`; the row/JSON paths were deliberately scoped out to this issue.
- `#1446 (Node row column order + interned column keys)` — owns the interned-key write path and the
  `Object.keys(row)` ordering contract this must preserve.
- `#3497 (type-aware canonicalization)` — the JSON-object site is the same **b-5** cell the
  canonicalization contract records as an open hole.

pmcfadin:
Closes #3630.

A row column name is **DATA**. This stops it reaching JavaScript's object model.

## The defect

`row_to_object` and `json_to_napi` wrote user-controlled names with an ordinary property assignment. An assignment is a JavaScript `[[Set]]`, which **consults the prototype chain**, and `Object.prototype` carries exactly one accessor — `__proto__`. So a column of that name reached the inherited **setter** instead of becoming a property. Two failure modes, both measured at the binding surface on unfixed code:

| value | unfixed behaviour |
|---|---|
| string | column **silently VANISHES** — `Object.keys(row)` omits it, not an own property, reads back as `Object.prototype`. No error anywhere. |
| null | **prototype REPLACED with null** — `getPrototypeOf(row) === null`, `typeof row.hasOwnProperty === 'undefined'`; the row stops being a normal object. |

Every row of every result set goes through here, so the blast radius exceeds #3504's UDT field bag (reached only by rows with a UDT cell).

## Decision (ratified by the lead on `coord-request: 3630-R1`)

**Rows keep `Object.prototype`; every column is DEFINED as an own property, never assigned.** JSON-object cells get a **null prototype**, like #3504's field bag. The asymmetry is deliberate and tracks a structural difference, not a preference:

* a **row** travels beside `result.columns`, so it has an authoritative key list to be probed against, and `Object.create(null)` would break `row.hasOwnProperty(...)`, `row.toString()` and `instanceof Object` on **every row of every query** to fix a name almost no schema uses;
* a **JSON object** has no declared key set — its keys *are* the data — so `obj[k] === undefined` meaning "no such key" is worth more than inherited methods existing.

Neither is a special case on the literal string `__proto__`: that is a rarer delimiter, not a removed channel, and would leave every other inherited name (including one a future JavaScript adds) able to intercept a column. `"constructor"`/`"toString"` in the fixture are what distinguish the two.

**What happens to a `__proto__` column now**, stated as the lead's dispatch required: it is **present as an own enumerable data property with its declared value**. Not dropped, not renamed, not rejected.

## Mechanism: M2, selected by measurement, not preference

`design.md` **D1b** pinned the decision rule BEFORE any number was seen (lead condition: *"a threshold chosen after seeing numbers is not a gate"*): `test_wide_rows` scan, 1 warmup + 7 timed pairs, **alternating arms**, adopt M1 iff median regression ≤ **5%**, with a **validity precondition checked first** — baseline relative half-range ≤ 2.5%.

| arm | median rows/s | rel half-range | verdict |
|---|---|---|---|
| baseline (pre-fix `[[Set]]`) | 12624.2 | **0.69%** | validity PASS |
| **M1** — napi-rs safe `Property` | 11143.4 | 3.43% | **−11.73% ⇒ REFUSED** |

M1's cost is structural: `Property` stores its name as a `CString` and sets the descriptor's `name` field to null, so #1446's once-per-result interned `JsString` handles cannot be reused — every row re-allocates a `CString` per column and re-interns each name in V8, the exact `O(rows × columns)` cost #1446 exists to remove.

**M2** uses the raw descriptor form, whose `name` field takes a `napi_value`, keeping the interning. One `unsafe` block, isolated in a new 144-line `row_properties.rs` (keeps `value.rs` under the campsite limit and makes the unsafe reviewable), with each Node-API precondition discharged individually. The lead authorized the `unsafe` **conditional on the measurement selecting M2**; it did.

### Two measurements I am NOT citing, and why

* **Release-profile A/B**: pointed the same way (−18.78%) but baseline half-range was **14.02%** under host contention ⇒ **fails D1b's own precondition**. "Points the same way" is the reasoning that precondition exists to refuse.
* **M2-vs-baseline acceptance**: nominal −4.75%, but baseline half-range **20.09%** (load average 48, a peer lane's `rustc` at 469%) ⇒ invalid. Quoting a number just inside the threshold while noise is 4× the threshold would be meaningless.

**Residual, stated plainly: M1's refusal is measured and valid; M2's own delta versus baseline is unmeasured at validity.** The pinned rule's procedure ("measure M1; else adopt M2") is nonetheless satisfied — the M2 acceptance check was extra rigor I added, and it is the part that is inconclusive.

**And a defect in the rule I wrote**, recorded rather than patched: relative *half-range* is max-min based, so it is outlier-sensitive and gets **worse** with more samples — meaning the rule's own remedy ("increase the run count") cannot satisfy the rule's own precondition. A dispersion measure that shrinks with n was what it needed. Changing it after seeing numbers is exactly the post-hoc threshold the lead forbade, so it stands as written for this delivery.

## Two-directional oracle (lead dispatch requirement)

Measured by actually reverting `value.rs` and rebuilding — not by inspection:

| build | Node suite |
|---|---|
| **unfixed** | **6 failed / 6 passed** |
| **fixed** | **12 passed** |

The 6 green-on-both cases are deliberate regression controls (`constructor`/`toString` already worked via `[[Set]]`; #1446 ordering; the valueless-column skip; the contrast row). Python: **5/5**. Regression suites (#3504, types, TS definitions): **151/151**.

### Two defects found in my own tests by RED-running them

1. **Cross-realm prototype identity.** `Object.getPrototypeOf(row) === Object.prototype` is false even for a healthy row — jest's VM context has a different `Object` intrinsic than the addon's. It failed on rows with *no colliding cell*. Left in, it would have been a permanent false failure post-fix.
2. **A vacuous pass.** My "every column is an own data property" case iterated `Object.keys(row)` and **passed on unfixed code** — because the lost column is exactly the one missing from `Object.keys`. *A test whose subject set is the output it is checking cannot see an omission from that output.* It now derives its subject from `result.columns`. Review would not have caught this; running it did.

## Measurements owed by the spec, all answered

* **`SELECT id AS __proto__` IS supported** ⇒ the issue's "minimal repro needs no fixture" claim is TRUE (added as a second subject). A *quoted* alias is a parse error — orthogonal parser gap, recorded not worked around.
* **The null/prototype-replacement mode is reachable**, overturning my own earlier conclusion. `TTL(col)`'s evaluation is unwired (#692) so it returns a real `Value::Null` — the input a CQL tombstone cannot supply. The test declares its own expiry: if #692 lands, its `toBeNull()` fails loudly and the remedy is another route, never relaxing it.
* **`Value::Json` is NOT reachable from the public surface.** A `.cql` schema declaring `json` is rejected (`references undefined UDT 'json'`) because `json` is absent from the CQL parser's type map; the `Custom("json")` mapping lives in the *serialization-header* parser and Cassandra never writes a `JsonType` comparator. So the null-prototype fix ships **correct but uncovered** — deliberately not propped up by a CQLite-written round-trip, which is invariant to a uniform error on both sides (#3042).
* **The EXTRAS path is not reachable from the binding surface** — 5 routes measured (`SELECT *`, explicit quoted columns, `COUNT(*)`, schemaless, partial schema). Both paths call one helper so a divergence is structurally impossible *today*, but **a shared helper is not coverage**: that branch's own line is executed by no test I can write. Recorded for the spec audit rather than argued away.

## Fixture

Cassandra-**5.0.2-written**, `test-data/fixtures/issue_3630/`, committed checkout-relative (both binding suites resolve the corpus from `CQLITE_DATASETS_ROOT` and never fall back to the checkout). Verified from a fresh `git worktree add --detach HEAD`. Four collision columns exercising **three** mechanisms — `__proto__` the only accessor, `constructor`/`toString` inherited writable data, `prototype` not on `Object.prototype` at all — stated in the schema so the fixture does not overclaim what it proves.

**A correction made during the work:** the generator first claimed the golden was reproducible. Measurement falsified that — row 2's explicit CQL null is a cell **tombstone** carrying a wall-clock `local_delete_time` no CQL clause can pin. Corrected in the schema, generator and README, because an overclaim in a committed fixture is what stops the next reader checking.

## roborev

**Job 287** (genuine: 1.54M input / 1.42M cached tokens; `prompt-content: PASS (16/16)`; `sha-assert: PASS`) found 4. Both actionable ones fixed in `3e9cd15eb`:

* **Medium — `OUT_DIR` guard was LEXICAL**, so `$REPO_ROOT/../../var/tmp/x` satisfied the repo-prefix test and then reached `rm -rf`. Now canonicalized with `realpath -m` **before** validation, required rather than optional. Verified behaviourally: the escape is rejected naming its resolved path; a legitimate `--out` still runs. **This is a FAMILY — 12 generator scripts share the guard, and `generate-issue-3504-udt-collision.sh` is the one that also `rm -rf`s.** Proposed as a separate issue (`coord:follow-up-proposed`) rather than absorbed here.
* **Low — `index.d.ts` claimed `row.hasOwnProperty(...)`/`row.toString()` work**, which this PR's own fixture falsifies: a column named `toString` makes `row.toString` a *string*. The doc now shows the both-directions failure and directs callers to the free-standing `Object.hasOwn` / `Object.prototype.*.call` forms.
* The two remaining Lows are the extras-path and `json_to_napi` coverage gaps **I had already measured and documented** above — independent corroboration, not new information.

## Certification and DECLARED DEGRADATION

`--lite` PASS at `3e9cd15eb` (`tree-integrity: PASS`, `python-tier: PASS`).

**The subagent channel is DEAD on this box (#3688 / measured with three controls), which degrades this PR's provenance and I am declaring it rather than letting it read as skipped stages:**

* `flow-closer`, `spec-auditor` (the **C intent audit**) and `rust-reviewer` are all subagents. Three controls — opus `sstable-developer` ×2 and a degenerate haiku `general-purpose` asked only to `echo` one line — all ran at ~0% CPU and produced nothing.
* So **the whole implementation and the Rust review happened in-session**, and #2084's disposable-closer isolation is lost. The in-session self-review did find real defects (a false `SAFETY` comment claiming `NapiRaw::raw` was not an unsafe op; a spec scenario whose premise M2 voids) — but a self-review is structurally weaker than an independent one and I am not presenting it as equivalent.
* **roborev is a script, not a subagent, so the independent half of review-first survived** and is genuine (token accounting above).

**Round 2 (job 289**, 1.07M input / 958k cached, `prompt-content: PASS (16/16)`, `sha-assert: PASS` at `3e9cd15eb`): the **Medium is GONE and the `index.d.ts` Low is GONE**. Two Lows remain — the extras-path and `json_to_napi` coverage gaps, i.e. exactly the two residuals measured and documented above, corroborated independently for a second round.

`RESULT: FAIL` (findings PRESENT 2), so **this PR is NOT "roborev clean" and I am not claiming it is.** Both remaining findings propose the same remedy — *"add a test-only binding helper / test-support conversion entry point"* — which is a **new production surface added for testability**, i.e. a design decision beyond this issue's scope and against the grain of the doctrine that a test-only seam is one more thing a real invoker can set. Escalated to the lead on the issue thread as `coord-request: 3630-R2` rather than decided here. This is the structural conflict #3626 records: a deferred finding makes "roborev clean" unobtainable, so the merge rule blocks until the lead rules.


pmcfadin:
## Review + certification blocks, recorded in the PR (durable copy)

Moved here because /tmp was their only copy and the lead reports this fleet's own watchdog was once deleted by system cleanup. A second copy lives in the lane's gitignored `.drive-issue-state.md`.

<details><summary>roborev job 287 — round 1 (4 findings; Medium + one Low now FIXED)</summary>

```
repo: /data/lanes/lane-3630
branch: issue-3630-node-binding-drops-column
base: origin/main
head-sha: 562b3ba6b32d00151d7b1879298bcdc0d9eddffb
reviewed-sha: 64802eebcd9a03d7ac7761e37903c0f1a545d9c9..562b3ba6b32d00151d7b1879298bcdc0d9eddffb
assert-base: 64802eebcd9a03d7ac7761e37903c0f1a545d9c9 (merge-base of origin/main and HEAD; origin/main tip b1e8598a23f99879c6b4659db0c19ce1255f7ddf)
job: 287
model: gpt-5.6-sol
census: 25 files, +2384/-4
tokens: input=1539022 cached=1416960 output=7305
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (16/16 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: PRESENT (4)
log: /tmp/claude-1000/-data-lanes-lane-3630/1f5ce069-85e4-4d28-949c-6aad1d5acaab/scratchpad/roborev1.log
RESULT: FAIL
```

```
Review (by codex)
------------------------------------------------------------
## Review Findings

- **Severity**: Medium
- **Location**: `test-data/scripts/generate-issue-3630-row-collision.sh:300`
- **Problem**: The `OUT_DIR` safety check is lexical, so paths containing `..` or symlinks can pass the repository/`/tmp` prefix check while resolving elsewhere. Subsequent `rm -rf` operations can delete an unintended `test_row_collision` directory.
- **Fix**: Canonicalize `OUT_DIR` before validation, reject symlink escapes, and perform destructive operations only on the verified canonical path.

---

- **Severity**: Low
- **Location**: `bindings/node/__test__/issue-3630-row-key-namespace.test.js:335`
- **Problem**: No test exercises the modified extras branch in `row_to_object`. Both alias tests confirm `__proto__` is present in `result.columns`, so they only cover the interned-key path.
- **Fix**: Add a query or test seam where a `__proto__` value is absent from metadata, assert that condition, and verify its descriptor, value, prototype, and ordering.

---

- **Severity**: Low
- **Location**: `bindings/node/src/value.rs:434`
- **Problem**: The new null-prototype JSON-object behavior, including nested objects and `__proto__` keys, has no runtime coverage.
- **Fix**: Add a live-NAPI test seam accepting a JSON document and assert own-property preservation plus null prototypes at every nesting depth.

---

- **Severity**: Low
- **Location**: `bindings/node/lib/index.d.ts:190`
- **Problem**: The documentation says `row.hasOwnProperty(...)` and `row.toString()` work, but arbitrary column names can shadow either method; the included fixture already makes `row.toString` a string.
- **Fix**: Qualify that these methods work only when not shadowed and consistently recommend `Object.hasOwn(row, name)` or `Object.prototype.hasOwnProperty.call`.

## Summary

The change safely defines Node row columns as own properties, introduces null-prototype JSON objects, and adds cross-binding collision fixtures and tests.
```
</details>

<details><summary>roborev job 289 — round 2 at 3e9cd15eb (2 Lows remain; the two MEASURED-unreachable coverage gaps)</summary>

```
repo: /data/lanes/lane-3630
branch: issue-3630-node-binding-drops-column
base: origin/main
head-sha: 3e9cd15eb46f62381cbe7face1d4330935684751
reviewed-sha: 64802eebcd9a03d7ac7761e37903c0f1a545d9c9..3e9cd15eb46f62381cbe7face1d4330935684751
assert-base: 64802eebcd9a03d7ac7761e37903c0f1a545d9c9 (merge-base of origin/main and HEAD; origin/main tip 4e2c8b8f343558002bb933c82eb9ed334b3ebb21)
job: 289
model: gpt-5.6-sol
census: 25 files, +2457/-4
tokens: input=1074847 cached=958208 output=7680
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (16/16 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: PRESENT (2)
log: /tmp/claude-1000/-data-lanes-lane-3630/1f5ce069-85e4-4d28-949c-6aad1d5acaab/scratchpad/roborev2.log
RESULT: FAIL
```

```
Review (by codex)
------------------------------------------------------------
## Review Findings

- **Severity**: Low
- **Location**: `bindings/node/src/value.rs:743`
- **Problem**: The extras path uses the new descriptor mechanism but no test forces a value absent from `result.columns`; both alias tests exercise the interned path, leaving this regression-prone branch uncovered.
- **Fix**: Add a test-only binding helper or lower-level test that constructs mismatched metadata/value keys and verifies `__proto__`, descriptor attributes, and extras ordering.

---

- **Severity**: Low
- **Location**: `bindings/node/src/value.rs:457`
- **Problem**: The new null-prototype behavior for JSON objects, including nested and `__proto__` cases, has no executable coverage.
- **Fix**: Add a test-support conversion entry point for arbitrary `Value::Json` documents and assert prototypes and colliding keys at every nesting depth.

## Summary

The change defines row columns as own properties, gives JSON objects null prototypes, and adds cross-binding collision fixtures and tests.
```
</details>

<details><summary>AGENT-GATE LITE SUMMARY at 3e9cd15eb — NOT the gate of record</summary>

```

run-id: /tmp/agent-gate.MbgBk0
commit: 3e9cd15 branch: issue-3630-node-binding-drops-column dirty: no
lite-scope: file-size fmt clippy roborev-lints scoped-tests (full gate NOT run — run it once before merge)
python-tier: PASS (maturin develop --profile dev -m bindings/python/Cargo.toml && pytest bindings/python/tests -m 'not slow' -q)
tree-start: 3e9cd15eb46f dirty: no digest: c50ba3506794
tree-end: 3e9cd15eb46f dirty: no digest: c50ba3506794
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (10s)
clippy:            PASS (7s)
roborev-lints:     PASS (57s)
scoped-tests:      PASS (124s)
logs: /tmp/agent-gate.MbgBk0
summary-file: /tmp/claude-1000/-data-lanes-lane-3630/1f5ce069-85e4-4d28-949c-6aad1d5acaab/scratchpad/gate-lite2.txt
RESULT: PASS
```
</details>

<details><summary>D1b raw measurement pairs — all four runs, including the three INVALID ones</summary>

Recorded in full so the validity judgements are auditable rather than asserted. Columns: `baseline_rows_per_sec candidate_rows_per_sec`.

```
# run 1 — dev, ITERS=60, INVALID (baseline rel half-range 3.30% > 2.50%) — DISCARDED, not read as evidence
12083.731380955365 11113.794863891835
12387.068632194356 11404.169908117174
12019.86980102852 11568.71853300525
12852.758456854082 11023.811312806896
12623.744518638632 10858.96334131856
12637.040117416973 11191.315038159475
12631.292762162722 10895.941888929783

# run 2 — dev, ITERS=250, VALID (0.69%) — THE run that refused M1 (-11.73%)
12730.117737103283 10800.340775462646
12624.208141943473 11143.37401287226
12691.789321751683 10510.815004998452
12556.4854838851 10833.223989992528
12671.69687880214 11198.928656610397
12572.878598595651 11275.826628021121
12571.367603600169 11230.840385088914

# run 3 — release-unwind, ITERS=900, INVALID (14.02%, host contention) — not cited
57302.663571016994 41301.980853903435
51863.72148037951 45984.85039799374
63933.64614741414 52615.333691296626
53875.211491100206 35418.71113524504
49429.77166715299 42125.555086502114
50040.34406366137 42044.66978511415
49393.91275578192 50383.15694198381

# run 4 — M2 vs baseline, dev, ITERS=250, INVALID (20.09%, load avg 48) — not cited
14688.903077371142 13278.810874247449
17880.878815535365 17031.377018212046
19953.313556194233 20174.00277157946
15503.294658203535 13125.857198324014
18029.28135296066 13276.977237668632
12770.472570018406 19378.070083576422
17900.727409010437 19732.41088775758
```
</details>

pmcfadin:
## GATE OF RECORD — RESULT: PASS at `e8a5775` (37/37 components)

Attempt 2. Attempt 1 red on two components, both **proven** not mine and now filed as #3704 and #3706 — and **both PASS in this run**, which is the corroboration that they were environmental rather than caused by this diff.

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.0LeAWN
commit: e8a5775 branch: issue-3630-node-binding-drops-column dirty: no
schemas: 8/8 canonical .cql readable under /data/lanes/lane-3630/test-data/schemas (checkout-relative)
tree-start: e8a577547346 dirty: no digest: 35417083cf35
tree-end: e8a577547346 dirty: no digest: 35417083cf35
tree-integrity: PASS
file-size:         PASS (0s)
fmt:               PASS (7s)
clippy:            PASS (322s)
roborev-lints:     PASS (50s)
core-tests:        PASS (574s)
tombstones-scan:   PASS (11s)
scan-offload-guard: PASS (132s)
work-counters-guard: PASS (52s)
byte-budget-guard: PASS (15s)
arrow-parity-guard: PASS (11s)
memory-budget:     PASS (176s)
integration-tests: PASS (33s)
format-compat:     PASS (7s)
write-tests:       PASS (126s)
cli-tests:         PASS (130s)
bti-multiclustering: PASS (8s)
flight-query-semantics-oracle: PASS (13s)
flight-tests:      PASS (99s)
legacy-heuristics: PASS (226s)
feature-iso-parquet: PASS (165s)
feature-iso-delta-scan: PASS (129s)
python-bindings:   PASS (279s)
node-bindings:     PASS (379s)
binding-rust-tests: PASS (30s)
delivery-telemetry: PASS (0s)
oom-audit:         PASS (4s)
parity-report:     PASS (5s)
operator-metrics-doc: PASS (16s)
pub-surface:       PASS (0s)
tooling-tests:     PASS (1004s)
minimal-build:     PASS (63s)
smoke:             PASS (31s)
logs: /tmp/agent-gate.0LeAWN
summary-file: /tmp/claude-1000/-data-lanes-lane-3630/1f5ce069-85e4-4d28-949c-6aad1d5acaab/scratchpad/gate-full2.txt
RESULT: PASS
```

### Attempt 1, recorded for completeness (RESULT: FAIL, 35/37)

Not a waiver — a passing gate exists above and is the gate of record. Kept because the two failures are filed defects that other lanes will hit.

| component | attempt 1 | attempt 2 | filed |
|---|---|---|---|
| `legacy-heuristics` | **FAIL** (746s) — `async_bridge.rs:291` wall-clock assert: 8x100ms concurrent sleeps must finish <400ms, observed **421ms**. Standalone: **8/8 PASS in 0.10s**. | PASS (226s) | #3704 |
| `tooling-tests` | **FAIL** (640s) — one `#1465` assertion; the failing test's IDENTITY differs between in-gate, clean-main and branch runs of a byte-identical suite | PASS (1004s) | #3706 |

A disk note, since the floor is what caught it: attempt 2's preflight **refused to launch at 45G** avail against the 90-100G requirement. This lane's `target/` had reached **114G** (`target/debug` alone 92G). Cleaned the completed run's side dirs, the staged A/B `.node` artifacts and `target/debug`: **45G -> 155G**, +110G returned to a box shared with three other lanes.

pmcfadin:
## Intent audit against the spec — **SELF-PERFORMED, NOT an independent C**

**Read this label first.** `spec-auditor` is a subagent and the subagent channel is dead on this box (#3688, three controls). Doctrine (#2084) puts **C** in an independent context precisely so the implementer is not the auditor. That property is **absent here.** This is the implementer auditing their own work against their own spec, and it is recorded as a substitute-of-record, not as a C PASS. The lead ruled "merge" with this gap declared in `coord-request: 3630-R2`; it remains theirs to overturn.

Spec: `openspec/changes/node-binding-drops-column/specs/binding-row-key-namespace/spec.md` — 6 requirements, 12 scenarios.

| # | Requirement / Scenario | Verdict | Evidence |
|---|---|---|---|
| R1 | A row column is defined, never assigned | **satisfied** | `row.rs` `row_to_object` accumulates `RowProperties` descriptors, one `napi_define_properties`; `awk '/^pub fn row_to_object/,/^}/' \| grep set_property` → nothing |
| R1.1 | string-valued `__proto__` arrives | **satisfied** | jest `a STRING-valued column named __proto__ survives…` — RED pre-fix, green post |
| R1.2 | null-valued `__proto__` leaves prototype | **satisfied** | jest `a NULL-valued column named __proto__ does NOT replace the prototype` via `TTL(col) AS __proto__`; pre-fix `getPrototypeOf(row)===null` was TRUE |
| R1.3 | **both cases on the EXTRAS path** | **UNMET — uncovered** | 5 routes measured, none reaches the branch. Scope statement in code; follow-up #3695 with a trigger. **This is the one requirement the delivery does not satisfy.** |
| R1.4 | a SECOND inherited name | **satisfied** | jest `a SECOND inherited name…` (`constructor`, `toString`) — green both sides by design, i.e. the AC7 discriminator |
| R1.5 | unrepresentable name refused, not dropped | **premise void** | M2 passes a length-delimited `JsString`; no unrepresentable name exists. Recorded in the spec as `PREMISE VOID` with the obligation retained for any future mechanism |
| R2 | A JSON object cell inherits nothing | **satisfied (code) / uncovered (test)** | `ctx.create_null_prototype_object()` at every depth |
| R2.1–2.2 | JSON `__proto__` key / null key | **UNMET — unreachable** | `Value::Json` structurally unreachable: `.cql` `json` rejected as undefined UDT; Cassandra writes no `JsonType` comparator. Scope statement + #3695 trigger |
| R2.3 | reached from public surface **OR unreachability measured** | **satisfied via the OR branch** | the spec wrote this disjunction deliberately; the measurement is recorded |
| R3.1 | `Object.keys(row)` = SELECT order | **satisfied** | jest `Object.keys(row) preserves authoritative SELECT column order` |
| R3.2 | ordinary row observationally unchanged | **satisfied** | 163 Node tests green incl. #3504 + types + TS definitions; `udt_to_object` untouched |
| R4 | declared TS surface states the contract | **satisfied with ONE KNOWN INACCURACY** | `Row` + `JsonObject` documented; **but "Every column" is wrong for a valueless declared column** — roborev job 295, patch recorded in #3695 |
| R5 | Python row path asserted | **satisfied** | 5/5 pytest; surfaced the Node/Python null-fill divergence |

### Honest totals

**10 of 13 satisfied; 1 premise-void; 2 UNMET (both coverage, both measured-unreachable, both with in-code scope statements and triggered follow-ups); 1 satisfied-with-a-known-inaccuracy.**

A real C would likely report **`partial`**, and by doctrine an unmet requirement blocks merge — which is exactly why this went to the lead as `3630-R2` rather than being self-certified. The lead ruled to merge over it and forbade the remedy that would close it (a test-only public surface). **The unmet requirements are therefore shipped knowingly, with the measurement and the trigger recorded at each site**, not silently.

### What I would want a real C to check that I cannot check on myself

* whether the **spec itself** was written to be satisfiable by what I intended to build — a self-audit cannot detect a spec bent toward its implementation, and I DID edit the spec mid-delivery (the `PREMISE VOID` scenario, and R2.3's `OR` branch). Both edits are defensible and both are recorded in the diff — but "defensible to the author" is the failure mode C exists to catch.

Issue #3513, opened 2026-08-29
MAX_DECOMPRESSED_SIZE unused-constant warning under a zero-feature cqlite-core build (blocked on compression.rs file-size)


pmcfadin:
Split out of #1452 (PR #3509) — deliberately not fixed there, with a reason worth recording.

## What
`cqlite-core/src/storage/sstable/compression.rs:23` declares `const MAX_DECOMPRESSED_SIZE`, used only under the compression features. #1452 added `cqlite-ffi-common`, which depends on `cqlite-core` with `default-features = false`, so a **zero-feature `cqlite-core` configuration now exists in the graph for the first time** and surfaces:

```
warning: constant MAX_DECOMPRESSED_SIZE is never used
```

## Scope of the impact — deliberately small
It appears under `cargo check -p cqlite-ffi-common` **only**. **No agent-gate component hits it** (verified: the gate's clippy pass excludes `cqlite-core`, and feature unification gives `cqlite-core` its default features from the other selected packages). Nothing is red today.

## The one-line remedy, and why it is blocked
Direct precedent sits 71 lines below it — #1873 gated `validate_decompression_size` for this exact lint:

```rust
#[cfg(any(feature = "lz4", feature = "snappy", feature = "deflate", feature = "zstd"))]
```

But `compression.rs` is **918 lines**, over the 800-line source threshold, so adding a line to it trips the gate's `file-size` ratchet and would need `CQLITE_ALLOW_FILE_GROWTH=1`. **A ratchet override is a worse trade than a warning no gate sees** — that was the call made in #1452 rather than quietly setting the override.

## So the real work is the split, not the cfg
- [ ] Split `compression.rs` (918 lines) by responsibility per epic #1116, bringing it under threshold
- [ ] Then apply the `#[cfg(any(...))]` gate to `MAX_DECOMPRESSED_SIZE`, matching #1873's precedent
- [ ] Confirm `cargo check -p cqlite-ffi-common` is warning-free afterwards

Filed by the #1452 worker lane.

Issue #3677, opened 2026-08-31
cqlite-core's packaged crate ships 105 integration tests whose fixtures are NOT packaged (repo-root test-data)


pmcfadin:
Raised as a **Medium** roborev finding on the #3220 hardening PR for `issue_3358_bti_query_token_bound.rs` (job 284). Filing rather than fixing there, because the finding is real but **the class is pre-existing and repo-wide**, and the suggested local fix would violate #3220.

## The scenario

`cqlite-core` has **no `publish = false`** (the workspace root `cqlite` package does; `cqlite-core` does not, and it carries `[package.metadata.docs.rs]`). So it is publishable. A published crate packages `tests/` by default, but the fixtures those tests read live at **repo root** in `test-data/datasets/`, outside the package. `cargo test` from the packaged crate therefore cannot find them.

## Measured scope — this is not one test

```
tests referencing test-data/datasets                         : 106
… of those, containing a hard failure (panic!/expect/unwrap) : 104
```

The #3220 hardening adds a 105th. **It does not introduce the class**; it joins it. Any fix that targets only the new test leaves 104 untouched, which is why this belongs in its own issue.

## Why NOT to fix it the way the finding suggests

The finding offers *"limit fail-closed behavior to Git checkouts while retaining the skip for packaged sources."* That is the wrong shape here twice over:

1. **It reintroduces exactly what #3220 removes.** A skip path on a committed fixture is the vacuous green the issue exists to eliminate — CLAUDE.md is explicit that there is deliberately no opt-out and none may be added, because committed source in a checkout is never legitimately absent.
2. **"Am I a git checkout?" is a runtime heuristic**, and inferring behaviour from environment shape is what the no-heuristics mandate (#28) forbids.

## Options that don't trade correctness for packageability

| option | assessment |
|---|---|
| **(a)** add `exclude = ["tests/**"]` to `cqlite-core/Cargo.toml` | Integration tests are not something a consumer runs; excluding them makes the package honest and fixes all 105 at once. Probably the right answer. |
| **(b)** package the fixtures | Relocate/duplicate the corpus under `cqlite-core/`. Costly — the corpus is large and deliberately fetched, not vendored. |
| **(c)** accept and document | State that `cargo test` is a repo-checkout operation. Zero code, but leaves `cargo test` broken for a packaged consumer who tries it. |

Recommend **(a)**, but it is a packaging/product call, not a lane decision — hence P3 and unassigned.

## Acceptance

[...]

Issue #3460, opened 2026-08-28
Document waiver-request ordering: request only when the head is FINAL (pushed, conflict-free, post-gate) — 3 lanes hit this in one day


pmcfadin:
Three independent lanes hit the same ordering trap on 2026-08-28. None of them made a judgement error; the sequencing simply is not written down.

## The trap

A lane's roborev round returns `prompt-content: FAIL (absence)`, so it posts an absence-waiver request with its `base`/`head`/`job` triple and its token accounting — **correctly**, and often with the triple exactly matching live state at the moment of posting. Then the head moves for an ordinary reason, and the grant becomes unapplicable:

| lane | PR | what moved the head after the request |
|---|---|---|
| #1705 | #3382 | grant received 14:49Z; then a conflict with #1701's just-merged metrics had to be resolved |
| #1699 | #3403 | triple was exact and the PR was `CONFLICTING` **at the same moment** |
| #3248 | #3455 | fixes committed but unpushed, **and** the PR was `CONFLICTING` |

Per #3312 a waiver binds **base AND head AND job** — deliberately, so one authorization cannot ride to a different review. That binding is correct and should not be loosened. And `--recheck-job` is not an escape: #3392 stabilised the *base* comparison against a moving `main`, but it cannot make a *head* binding survive a genuine content change.

**Cost when it happens:** an authorization is spent on code that will not merge, the owner is asked to judge a review that no longer describes the diff, and the lane waits on a grant that can never be applied. #1705 lost hours to exactly this.

## The rule to document

> **Request an absence waiver only when the head is FINAL — pushed, conflict-free, and post-gate. Anything that will still move the head comes first.**

Concretely, in order: push every local commit → rebase/resolve so `mergeable != CONFLICTING` → gate of record → roborev confirmation pass on that final sha → *then* request the waiver with that job's triple.

**The confirmation pass matters and is easy to skip.** #3367 found its gated sha had never been reviewed at all: round 25 reviewed `6f5fc2b7c` and two commits landed after it. #2605 hit the same shape. So the final sha needs its own review round, and *that* round's job id is the one the waiver must name.

## Why a lane cannot reasonably infer this

The wrapper's own diagnostic prints `base`/`head`/`job` at the moment of failure, and those values are correct *then*. Nothing in the failure output says "these will be invalid once you push". A lane doing the obvious, careful thing — copy the verified triple into a request — produces a stale request through no fault of its own.

## Suggested fixes, cheapest first

[...]

PR #3696, opened 2026-08-31
fix(#3220): fail closed in the #3358 BTI token-bound test — its fixture is committed source

pmcfadin:
## Make the #3358 BTI token-bound test fail CLOSED — its fixture is committed source, not a fetched optional corpus

Follow-up to **#3358** / PR **#3362** (@michaelsembwever's fix, merged), opened per the owner's tightening on that thread: *open the #3220 hardening as a PR, don't merely file it.*

`cqlite-core/tests/issue_3358_bti_query_token_bound.rs` landed with three SKIP guards and a module doc asserting the fixture is *"the gitignored `Data.db` binaries"*. **That premise is factually wrong**, and the guards it justifies are exactly the #3220 defect: the only end-to-end pin of a silent-wrong-data bug can quietly not run, behind a green suite.

### The premise, measured rather than asserted

```
$ git ls-files test-data/datasets/sstables/ | grep -c wide_multiclustering_small
10                                        # incl. da-1-bti-Data.db, Partitions.db, Rows.db, TOC.txt

$ git check-ignore -v .../test_da/wide_multiclustering_small
(no output, exit 1)                       # NO ignore rule matches it
```

The fixture is **git-tracked and not gitignored**. It is present in every complete checkout and appears in a fresh `git worktree add` with no fetch at all. Its absence therefore means a **broken checkout** — never an expected condition — so per #3220 every case must be `must_run` and fail closed.

### Why the root-walking half is load-bearing, not defensive dressing

This is not hypothetical on this fleet. On the box that certified #3362:

```
$ ls /data/datasets/**/wide_multiclustering_small*      # the EXPORTED CQLITE_DATASETS_ROOT
(absent)
$ ls test-data/datasets/sstables/test_da/wide_multiclustering_small-47f6a300…
(present)
```

**Neither root is a superset.** The root `fetch-datasets.sh` prints does not carry this fixture; the checkout does. So a resolver that picks a root by keyspace and commits to it — `root.join(keyspace).is_dir()` — selects `/data/datasets`, fails to find this *table*, and **skips**. That is precisely the #3220 failure mode that let the #3032 case skip silently. Resolution must therefore be per **TABLE**, walking **every** candidate root for that table's own `*-Data.db`, and by **evidence** rather than a fixed env-first/checkout-first preference — any fixed ordering picks wrong for one set of tables.

### What changes

| | Before | After |
|---|---|---|
| Module doc | "Requires the gitignored `Data.db` binaries" | states it is committed source, with the `git ls-files` / `check-ignore` evidence and why absence ⇒ broken checkout |
| `fixture()` | `-> Option<Fixture>`, `None` ⇒ SKIP | `-> Fixture`, panics with a diagnostic naming both candidate roots |
| 3 SKIP guards | silently pass | **removed** — 0 remain |
| Root resolution | by keyspace | per table, every candidate root walked |

There is deliberately **no opt-out env var and none may be added**: committed source in a checkout is never legitimately absent, so an escape hatch could only buy a vacuous green — the same reasoning CLAUDE.md records for the gate's `missing-schemas` check having no bypass.

The one reachable way to genuinely lose the file is real and already handled elsewhere: **#3310** added `--verify-only` reporting for git-tracked fixtures a SIGKILLed fetch left deleted. That is the case a skip would swallow and a panic surfaces.

### Verification

- Patch applied by an **anchor-asserting** script (each anchor asserted present **and unique** before substitution): `PATCHED OK: 6 anchors + 2 call sites; 0 SKIP guards remain`.
- Post-patch grep for `SKIP` / `return None` / skip-guards returns **0** functional hits.
- Test executes and passes against the committed fixture — and, critically, is now **incapable of passing without running**.

Closes the #3220 gap for this file. Does not touch @michaelsembwever's fix; only the test's guard structure and its module doc.

---

## roborev — SIX rounds, clean on the sixth

Final: job **291**, `RESULT: PASS`, `findings: NONE`, `sha-assert: PASS`, `prompt-content: PASS (2/2)`, tokens `input=546440 cached=409088` (genuine review, against the ~18.7k vacuous baseline). Base pinned to the immutable merge sha `b6c10865d`, never the moving ref.

| job | sev | defect | disposition |
|---|---|---|---|
| 284 | Low | remedy printed `…-<cfid>` — an unsubstituted placeholder, and `<` is shell input redirection | fixed |
| 284 | Med | published crate ships a test whose fixture can never exist | fixed (package `exclude`) |
| 286 | Med | (repeat of the above) | resolved by that fix |
| 287 | Low | remedy paths crate-relative → resolve to nothing from `cqlite-core/` | fixed |
| 288 | Med | glob would restore the **whole** fixture dir, silently discarding a reader's uncommitted work | fixed by deletion |
| 288 | Low | `:(top)` anchors a pathspec but selects no repository; unquoted path breaks on spaces | fixed by deletion |
| 289 | Low | unsetting the env var leaves a **cwd-relative** fallback → command silently reports nothing | fixed |
| 291 | — | **findings: NONE** | ✅ |

**The fail-closed behaviour was never in question.** It was RED/GREEN verified every round, including from the awkward `cqlite-core/` working directory. All seven findings landed on the *diagnostic text*.

### Round 5 changed the approach: deleted rather than patched

Five defects in five rounds inside one hand-built `git restore` string is a variant list that will not close. `fetch-datasets.sh --verify-only` **already** emits a precise, correctly-scoped restore command for exactly this case (#3310), so reproducing it inside a panic message was the wrong shape from the start. The string is gone; the message names one command and lets the tested tool produce the restore line. A comment at the call site says so, so nobody adds it back.

### The defect no review found

Having redirected readers to that tool, I tested that it works — and **on the fleet default it does not**. The probe is scoped to whatever root it is given, and `CQLITE_DATASETS_ROOT=/data/datasets` lies outside any git work tree, where it reports `NO SUBJECT` and exits **0**, finding none of the 10 deleted files. Unfixed, the remedy would have been a no-op that reads as an all-clear — the identical silent-pass this PR exists to remove.

Round 6 then showed my first fix for that was incomplete, and that my verification shared the reviewer's blind spot: I had written "executed verbatim" having run the command only from the repo root, never from the directory whose breakage I had just fixed.

### Closing property (not a round count)

The emitted command contains **no relative path and no inherited environment value** — corpus root and script path are both absolute and shell-quoted, so cwd and environment are eliminated by construction. Verified by extracting the string from the panic and **executing it** from three working directories:

```
                      repo root   cqlite-core/   /tmp
unset-env form          reports      NOTHING      —      <- the round-6 defect
this form               reports      reports    reports
```

## Verification summary

| check | result |
|---|---|
| GREEN | 3 tests **run** and pass; zero skips |
| RED (fixture moved out of tree) | all 3 **FAIL** fail-closed, from repo root and from `cqlite-core/` |
| emitted remedy | executed verbatim from 3 cwds; reports all 10 missing tracked files |
| packaging | `cargo package --list`: test **1 → 0**; siblings 455 → 454 (#3677) |
| roborev | job 291 PASS, findings NONE |

---

## ⚠️ NOT READY TO MERGE — the full gate of record is still OWED

What this PR has:

```
run-id: /tmp/agent-gate.EBkJQt
lite-scope: file-size fmt clippy roborev-lints scoped-tests
tree-start: cb910af15b8b dirty: no digest: fd02bee3c55b
tree-end:   cb910af15b8b dirty: no digest: fd02bee3c55b
tree-integrity: PASS
RESULT: PASS
```

**That block says `MODE: lite` and cannot be read as the gate of record** — its own `lite-scope` line names the five components it ran, and the full set is 36. Per CLAUDE.md a full `AGENT-GATE SUMMARY` must PASS once before merge, and it has not run on this branch.

Auto-merge is therefore **deliberately not armed**. Arming it would be the one-line version of certifying with a lite.

---

# Ownership under #3731 — lane-3731, first review this PR has had

PR opened 05:53Z and untouched for five hours: no gate of record, no C, `refs/claims/issue-3220` empty,
parent #3220 legitimately closed. `required` was green, which is the dangerous shape — every cheap signal
said ready and nothing had certified it. #3731 asks for a **decision**, not for this to merge.

**Decision: fix and merge.** The premise is sound and re-verified independently (`git ls-files` lists all 10
`da-1-bti-*` components; `git check-ignore` matches none), and fail-closed is the right posture. But two
things above were **described rather than asserted**, and one comment asserted a number its own citation
contradicts.

## AC1 — rebase onto current `origin/main`

`main` had moved **12 commits** since the PR was cut. All 5 commits rebased cleanly, no conflicts;
`HEAD..origin/main` = 0. The change still applies.

## AC2 was described, not asserted — and that is what blocked AC3

The module doc stated #3220's per-table rule; the code then re-derived it in a **private** `fixture_dir()`
instead of `support/datasets_root.rs::sstables_root_for_table`, which the AC names by symbol and which **9
sibling targets already use**. Three costs, the third decisive:

1. It re-implements the resolver #3220 exists to centralise — the shape of the original defect, in the file
   whose subject is that defect.
2. It misses the shared version's workspace-`Cargo.toml` root anchoring, candidate dedupe, and
   env-root-only-if-present semantics.
3. **It carries no test seam, so AC3 could not be staged honestly.** The checkout candidate is built from
   `CARGO_MANIFEST_DIR`, a *compile-time* constant, so with a private resolver the only way to hide a
   committed fixture is deleting a git-tracked file from the working tree — which trips the gate's mid-run
   `tree-integrity` check (#2926). `support/datasets_root.rs` documents
   `CQLITE_TEST_CHECKOUT_SSTABLES_ROOT` as existing for precisely this, and it can only ever *remove*
   fixtures from view, so a stray value makes a `must_run` lane FAIL, never pass vacuously.

The probe stays `da-1-bti`-prefix-specific rather than adopting the shared `table_has_data`'s
any-`*-Data.db` test, following `query_semantics_oracle_parity.rs::sstables_root_for_case`'s precedent for
a format-specific fixture. That is the safe direction: a root holding some *other* generation of this table
would satisfy `table_has_data`, win the selection, then hard-FAIL this `must_run` lane on a checkout that
does hold the `da` copy — and a guard that reds on correct input is the guard agents learn to waive.

## AC3 — each guard OBSERVED to fire, in all three directions

The body above claimed the test "is now incapable of passing without running." That was an assertion; below
is the evidence. A guard never observed to fire is not evidence.

| direction | staging | result |
|---|---|---|
| fixture **present** | `CQLITE_DATASETS_ROOT=/data/datasets` | `3 passed; 0 failed; 0 ignored` |
| fixture **absent** from every root | `CQLITE_TEST_CHECKOUT_SSTABLES_ROOT=<empty dir>`, `CQLITE_DATASETS_ROOT` unset | `0 passed; 3 failed` — **FAIL, not SKIP**, each case named individually |
| fixture **present but empty** | fixture copied to a temp root, `da-1-bti-Statistics.db` truncated to 0 bytes | `0 passed; 3 failed` — `Corruption("Failed to load Statistics.db …")`, **not** a 0-row pass |

The GREEN row is itself AC2 evidence, not just a control: `/data/datasets/sstables/test_da` holds **4
tables and not this one**, so all 3 cases passing proves the per-table fall-through to the checkout ran —
the exact shape (`root.join(keyspace).is_dir()` then commit) that let the #3032 case skip silently.

The present-but-empty row is the one doctrine warns is silent: a zero-length `Statistics.db` can make
`SELECT` return 0 rows with no error. Here it raises `Corruption` and fails all three cases.

## Two claims that did not survive being checked

**The `Cargo.toml` comment asserted "454 sibling integration tests" and cited #3677** — whose own
measurement is **106 referencing / 104 already hard-failing / 105 with this one**. The cited issue
contradicted the number citing it. Corrected to #3677's figure, and the one-file scope is now tied to
#3677 deliberately holding the class decision (`exclude = ["tests/**"]`) open as a packaging call.

**The panic's remedy was measured rather than trusted.** Running the exact command it prints, on an intact
tree from the repo root:

```
REAL exit = 1        8 lines, 7 of them ERROR:        line 1 is the only one that answers the question
```

The `Tracked-fixture probe (#3310)` half is correct and genuinely cwd-independent — it stat'ed all 894
git-tracked files under `test-data/datasets` individually and reported OK, from the repo root **and** from
`cqlite-core/`, which is what the absolute-path form was for. The other 7 lines answer a different question
(the checkout corpus is not a complete *fetched* corpus, which it is never meant to be). One of them is
worse than noise: it proposes `rm -f <root>/.dataset-pin && … fetch-datasets.sh`, whose default path is
destructive (`rm -rf` on the dataset root) — against the **checkout** root that deletes the committed
fixtures the reader is trying to restore. The message now names the one line to read and warns off the
rest, with the reason.

`exclude` itself stays, and the orphaned `repo_root` doc comment the diff had left attached to the new
`shell_quote` is reattached.

## Certification

The prior body's roborev evidence (job 291) was bound to `cb910af15`, **superseded by the rebase** and by
the two commits above — inert per #3460. Re-certification at the final head is recorded below.

---

## Review rounds 2–4: the fail-closed guard had no guard, and then its guard broke its own contract

Two more roborev blockers after the first, both genuine, both in code I wrote. Recorded in full because
the pattern is the useful part.

### Round 2 (job 252) — a measurement in a PR is not a guard

*"The new fail-closed behavior has no committed negative test. Existing runs with the fixture present
remain green if `fixture()` later reverts to skipping."*

Correct, and it is this issue's own AC3 one level up. The RED evidence above proved the guard fired
**once**, on one machine, on one day, and lived in a PR comment. A run with the fixture PRESENT is
byte-identical whether `fixture()` returns `Fixture` or `Option<Fixture>` — so nothing would have caught a
revert. A guard never observed to fire is not evidence; a guard observed once and never again is not a
guard.

So the observation moved into **`scripts/tests/test_issue_3358_failclosed.sh`** (15 assertions), modelled
on `test_point_vs_full_failclosed.sh` (#3220 AC2) — same file, same assertion contract, because it is the
same property for the sibling BTI lane. Wired as `bti-multiclustering`'s **fourth invocation**, beside that
sibling and for the same stated reason (the component has already built the binary it drives), with the
component's doctrine comment declaring it so a widened lane is never silent.

Held to the sibling's contract: no check may rest on a bare nonzero exit (an unrelated compile error
produces an identical rc) or a bare table-name match (healthy output carries the table name in the fixture
path). Staging is **surgical** — the keyspace dir is symlink-mirrored with its siblings and only this
fixture withheld — because hiding the whole corpus lets the assertion pass for an unrelated reason, and
because a keyspace dir that EXISTS is what pins the #3220 defect specifically: keyspace-granular selection
would have *accepted* that root.

**Discrimination limit, stated in the file rather than implied.** The sibling proves its staging surgical
from the lane's own output, because its target drives many fixtures and a sibling case's `PASS` line shows
only one was hidden. This target has exactly ONE fixture, so that evidence does not exist here and the
claim is made against the filesystem instead. Naming the weaker evidence beats implying the stronger.

**The contrast that matters** — run in a throwaway `git worktree` with its OWN `CARGO_TARGET_DIR`, because
a shared one serves a stale binary and would have made the measurement meaningless:

| tree | result |
|---|---|
| unmutated | **15 passed / 0 failed** |
| `fixture()` reverted to the pre-#3220 SKIP shape | **9 passed / 5 failed** |

All five failures are the `absent` case, and one names the planted symbol directly
(`output contains 'SKIP'`). A bare red would not have been evidence either — an unrelated breakage
produces an identical exit code. Control and `empty` still passed, so the harness is not simply broken.
Worktree and its target removed; the lane's tree verified clean.

The harness also found two defects in itself on first run, both the anti-vacuity machinery working:
`EXPECTED_CHECKS` was 13 against 14 real assertions (caught by the terminal anchor that exists for exactly
that), and the failure `printf` used a backslash-continuation inside SINGLE quotes — literal there, not a
continuation — leaking a stray `\` into the message. The `bad "…"` calls are unaffected; inside double
quotes the continuation is real. `test_point_vs_full_failclosed.sh` carries the identical construct; left
alone as out of scope and cosmetic.

### Round 3 (job 254) — the harness enforcing the contract violated it

*"The empty-fixture case does not enforce its stated count-bearing contract … separate `grep` calls do not
ensure the error and fixture name occur on the same line as claimed."*

Both true, in the file added one commit earlier to fix round 2:

1. `empty` had **no count assertion** while `absent` required `0 passed; 3 failed`, so one case failing
   while two silently passed satisfied it — a suite-wide `ran > 0` in disguise, the shape #3220 forbids and
   the shape this file's own header says every case is held to.
2. It used **two separate greps** under a comment claiming the rejection was *"named on one line together
   with this fixture"*. Two greps prove each string appears somewhere; they do not prove the conjunction.

**This is the PR's own defect class on its third instance, and its second in my code**: prose asserting
more than the code does. Once in the shipped `Cargo.toml` comment ("454 sibling integration tests"), once
in the panic remedy (read the probe line — in the branch where no probe line exists), once in the harness
built to stop exactly this. Writing a contract in a header comment is not enforcing it, and I wrote the
header.

Both fixed, and both fixes verified to **discriminate** rather than to pass — a constructed log with the
two strings on separate lines and a `2 passed; 1 failed` result:

```
old two-grep form   PASSES on separate lines   <- the false pass, demonstrated
new same-line form  correctly FAILS
new count form      correctly FAILS on '2 passed; 1 failed'
```

A green assertion proves nothing about an assertion's power; the falsifying input does — the same standard
the mutant contrast held the `absent` half to, applied to the half that had not been held to it.

### Certification status

| | |
|---|---|
| `--lite` | **PASS** (`tree-integrity: PASS`) at each head; round 1's FAIL was `tree-mutated-midrun`, my own doing (#2926), not the code's |
| roborev | 3 rounds, **3 blockers found and fixed**; all genuine (243k–427k input, `prompt-content: PASS`, `sha-assert: PASS`) |
| rust-reviewer | blocker (same as roborev round 1) + 4 nits; 3 fixed, 1 deferred to a follow-up at merge |
| gate of record | **NOT YET RUN — blocked on disk**, not on the code |

The gate of record is the one thing outstanding. box1 `/data` has **41G** on `/dev/nvme1n1` against the
**~96–102G** a full gate demonstrably needs here (measured live from a peer lane's running gate), so the
run would ENOSPC rather than certify. Tracked on #3731 as coord request `3731-R1`. **This PR must not merge
until a full `AGENT-GATE SUMMARY` with `RESULT: PASS` and `tree-integrity: PASS` exists at its final head**
— `required` being green is exactly the signal this issue was filed to distrust.

---

## AC3 wiring: DEMONSTRATED in-gate, not asserted

`--only bti-multiclustering` (a PARTIAL run — explicitly *not* the gate of record):

```
bti-multiclustering: PASS (5s)   [test cqlite-core --features state_machine,cli-helpers x2]
RESULT: PARTIAL (--only bti-multiclustering) - does NOT count as the gate
```

**5 seconds looked too fast** for a component running two cargo suites plus two shell harnesses, so the
PASS was not banked. Checked against the component's own log in that run's dir:

```
27 x "ok   - " lines
passed: 12  failed: 0      <- the sibling test_point_vs_full_failclosed.sh
passed: 15  failed: 0      <- this PR's test_issue_3358_failclosed.sh
12 + 15 = 27 ✓
and this harness's distinctive assertions by name: the generation-specific diagnostic,
the seam NOTE, and "ONE line carries both this fixture and the zero-byte reader rejection"
```

So the harness genuinely executes inside the gate component; the 5s is a warm cache, not a skip. A green
component that never reached the code is the failure mode #3522 exists for, and this PR would have shipped
the wiring claim without that check.

### Disclosed limitation this PR introduces: it widens a documented annotation blind spot

The SUMMARY feature-matrix annotation (#3453) is DERIVED from observed cargo argv, and the gate's cargo
interceptors are **unexported by design**, so cargo calls made inside a `bash scripts/tests/…` sub-script
are invisible to it. Measured:

| | |
|---|---|
| cargo invocations `run_bti_multiclustering` makes directly | **2** (observed; annotation reads `x2`) |
| cargo invocations inside its two fail-closed sub-scripts | **6** (3 each — unobserved) |

So `bti-multiclustering`'s annotation says `x2` while the component actually drives 8. That is **not
drift** — `x2` correctly names what the renderer observed, and `test_agent_gate_feature_matrix_annotation.sh`
compares declared against *observed*, so it does not red. And it is **pre-existing**: the sibling harness
already contributed 3 unobserved invocations, and `_fm_component_class`'s own `tooling-tests` carve-out
documents this exact limitation in prose.

But this PR takes the unobserved count from 3 to 6, so the annotation understates the component slightly
more than before. Recorded here rather than left to be rediscovered, because #3453's ruling is that a
pasted SUMMARY should state what it certified — and an annotation that omits sub-script work states less
than the component did. No code change proposed: the design deliberately accepts unobservable sub-script
cargo, and inventing a second mechanism to count it would be the "port is a second implementation" trap
(#3283).

Two hazards checked and ruled out while verifying this:

* **`tooling-tests` does not auto-discover `scripts/tests/*.sh`** — it invokes an explicit named list, so
  the new harness is not silently double-run there without the component's build context.
* **No self-test asserts a roster or count of component invocations**, so adding a fourth invocation to
  `run_bti_multiclustering` registers nothing elsewhere. `COMPONENTS` is unchanged (no new component), and
  `bti-multiclustering` correctly stays out of `DATASET_COMPONENTS` — this fixture is committed, not
  fetched.

## Certification status — AC4 pending on DISK only; #3740 was retracted

**This PR is a DRAFT and stays one until a gate-of-record PASS binds to its head.**

### The #3740 halt was withdrawn, and my "confirmation" of it was not independent

An earlier revision of this section said `binding-rust-tests` FAILs on clean `origin/main` and that AC4/AC6
were *impossible*. That was wrong, and the correction matters more than the claim did.

`agent-gate.sh` applies `-D warnings` **only to clippy**, via a scoped env prefix; `binding-rust-tests` gets
none. My failure came from `RUSTFLAGS=-D warnings` **exported in my session environment** and inherited by
the gate. Measured, same crate, same tree, one variable:

```
WITH    inherited RUSTFLAGS=-D warnings : error: constant MAX_DECOMPRESSED_SIZE is never used   exit=101
WITHOUT (env -u RUSTFLAGS)              : warning: ... Finished                                 exit=0
```

So I reported a "third independent reproduction" of a defect that was my own contaminated environment.
Since every box here is provisioned identically, such a fault reproduces everywhere and each reproduction
*looks* like corroboration while adding no information. #3740 is now P2 — the ungated constant is still
real latent hygiene, and it is not a blocker.

All verification in this PR is run under `env -u RUSTFLAGS -u CARGO_ENCODED_RUSTFLAGS`, both measured rather
than assumed (`RUSTFLAGS` contaminated, `CARGO_ENCODED_RUSTFLAGS` unset).

### What the last gate attempt established, and why it did not certify

The gate of record reached **36 of 37 components with the only red being an artifact of my own
configuration**:

```
bti-multiclustering: PASS 14
  27 ok-lines; passed: 12 failed: 0 (sibling harness) + passed: 15 failed: 0 (this PR's)
  this PR's per-case assertions present, each reporting "(3 occurrences)"
  DECLARED LIMIT emitted

tooling-tests: FAIL 207   <- NOT this diff
```

That FAIL is `scripts/tests/test_bti_perf_scan.sh`: it builds the harness honouring `CARGO_TARGET_DIR`
(`:149`) and then resolves the binary from `$REPO_ROOT/target` (`:158`), with **zero** references to the
variable. I had set `CARGO_TARGET_DIR` outside the worktree to relieve build-filesystem pressure, so the
binary existed somewhere the script never looks. **The run could never have produced `RESULT: PASS`** —
filed as **#3748**, and the configuration is now forbidden fleet-wide for a second, better reason: that
filesystem stores every gate's summary, and an ENOSPC there is silent (`RESULT: INCOMPLETE`, #3041).

**AC3 is therefore demonstrated inside a real gate run**, not only in an `--only` probe: the component
executed, both fail-closed harnesses ran to completion, and every assertion passed.

### Acceptance criteria

| AC | state |
|---|---|
| 1 rebase onto current `origin/main` | **satisfied** — v3 trigger (`merge-base..origin/main` ∩ my files) is **EMPTY**, so gating where I am is correct; reported as *gate of record verified at `<sha>`*, never "certified against main" |
| 2 committed-source distinction asserted, per-table roots, per-case `must_run` | **satisfied** — C intent audit judged the generation-specific-probe reasoning and accepted it |
| 3 RED-verify each guard | **satisfied**, demonstrated in-gate (above) and by a mutant contrast (15/0 unmutated vs 9/5 with `fixture()` reverted, one failure naming the planted symbol) |
| 4 gate of record PASS at final head | **pending on disk** — needs ~100G cold; `/data` has ~40G. Not blocked on code, review, or main |
| 5 roborev clean at certified sha | **satisfied** — job 269, `findings: NONE`, `prompt-content: PASS 4/4`, `sha-assert: PASS`, 605k in / 518k cached, at `981e75972` |
| 6 `premerge-assert` then arm `--auto` | after AC4 |

### Review history, in full

**11 roborev rounds, 5 blockers, all fixed, ending clean.** Plus a rust-reviewer pass (1 blocker, 4 nits)
and a C intent audit that caught a defect roborev had passed over. Nine instances of one class —
*prose asserting more than the code does* — seven in the shipped diff and two in the harness written to
prevent them. Each is recorded in its own commit message with the falsifying measurement, because the
pattern is more useful than the individual fixes.

Follow-ups filed rather than swept in: **#3737** (sibling harness: whole-directory `git restore`, raw-quoted
paths, relative-`TMPDIR` false FAIL), **#3743** (corpus-shape guard misdiagnosis), **#3748** (the
`CARGO_TARGET_DIR` hazard above).

### DIFF
--- a/cqlite-core/Cargo.toml
+++ b/cqlite-core/Cargo.toml
@@ -11,6 +11,21 @@ rust-version.workspace = true
 description = "Core engine for CQLite — read Apache Cassandra 5.0 SSTables locally without a cluster"
 readme = "../README.md"
 
+# Not packaged: this integration test reads a fixture that lives at REPO ROOT
+# (`test-data/datasets/…`), outside this package, so `cargo package` cannot carry it.
+# The test fails CLOSED on a missing fixture by design (#3220) — committed source is
+# never legitimately absent in a checkout — so shipping it in the published crate
+# would make `cargo test` fail for a consumer who can never satisfy it.
+#
+# Excluding it from the PACKAGE is the fix that keeps both properties: fail-closed in
+# a repository checkout, and absent (rather than skipping) where the fixture cannot
+# exist. A skip would reintroduce exactly the vacuous green #3220 removed, and a
+# "am I a git checkout?" runtime probe would be the environment heuristic #28 forbids.
+#
+# 454 sibling integration tests have the same repo-root-fixture property; fixing the
+# class (likely `exclude = ["tests/**"]`) is a packaging decision tracked in #3677.
+exclude = ["tests/issue_3358_bti_query_token_bound.rs"]
+
 # docs.rs build: document the commonly used features. Deliberately excludes
 # `wasm` (wasm32-only deps would fail the default x86_64 docs.rs target).
 [package.metadata.docs.rs]
--- a/cqlite-core/Cargo.toml
+++ b/cqlite-core/Cargo.toml
@@ -22,8 +22,11 @@ readme = "../README.md"
 # exist. A skip would reintroduce exactly the vacuous green #3220 removed, and a
 # "am I a git checkout?" runtime probe would be the environment heuristic #28 forbids.
 #
-# 454 sibling integration tests have the same repo-root-fixture property; fixing the
-# class (likely `exclude = ["tests/**"]`) is a packaging decision tracked in #3677.
+# This test JOINS a pre-existing class rather than creating it: #3677 measured 106
+# `cqlite-core` integration tests reading repo-root `test-data/datasets`, 104 of which
+# already fail hard rather than skip. Fixing the class (its recommended option (a),
+# `exclude = ["tests/**"]`) is a packaging call it holds open deliberately, so the
+# narrow single-file exclusion here is scoped to what this lane may decide.
 exclude = ["tests/issue_3358_bti_query_token_bound.rs"]
 
 # docs.rs build: document the commonly used features. Deliberately excludes
--- a/cqlite-core/Cargo.toml
+++ b/cqlite-core/Cargo.toml
@@ -22,11 +22,23 @@ readme = "../README.md"
 # exist. A skip would reintroduce exactly the vacuous green #3220 removed, and a
 # "am I a git checkout?" runtime probe would be the environment heuristic #28 forbids.
 #
-# This test JOINS a pre-existing class rather than creating it: #3677 measured 106
+# Measured effective, not assumed: `cargo package -p cqlite-core --no-verify --list`
+# lists 1015 files, 0 of them this test, and 454 other `tests/` entries survive — so the
+# key drops exactly one target and nothing else. (There is no `include` key, which would
+# make cargo ignore `exclude` entirely.) Note this does NOT assert the crate packages
+# end-to-end: `readme = "../README.md"` also points outside the package, so a full
+# `cargo package` may fail on that first. Packageability is #3677's, not this lane's.
+#
+# This test JOINS a pre-existing class rather than creating it. #3677 measured 106
 # `cqlite-core` integration tests reading repo-root `test-data/datasets`, 104 of which
-# already fail hard rather than skip. Fixing the class (its recommended option (a),
-# `exclude = ["tests/**"]`) is a packaging call it holds open deliberately, so the
+# already fail hard rather than skip. Fixing the class — its recommended option (a),
+# `exclude = ["tests/**"]` — is a packaging call it holds open deliberately, so the
 # narrow single-file exclusion here is scoped to what this lane may decide.
+#
+# An earlier draft of this comment said "454 sibling integration tests have the same
+# repo-root-fixture property". 454 is a real number and it is the one above — packaged
+# `tests/` ENTRIES, most of which read no fixture at all — attached to a claim about a
+# property it does not measure. The two counts answer different questions.
 exclude = ["tests/issue_3358_bti_query_token_bound.rs"]
 
 # docs.rs build: document the commonly used features. Deliberately excludes
--- a/cqlite-core/tests/issue_3358_bti_query_token_bound.rs
+++ b/cqlite-core/tests/issue_3358_bti_query_token_bound.rs
@@ -300,9 +300,18 @@ fn fixture() -> Fixture {
              \n\
              \x20   CQLITE_DATASETS_ROOT={verify_root} bash {verify_cmd} --verify-only\n\
              \n\
-             It names every git-tracked fixture that is missing and prints the exact, \
-             correctly-scoped restore command for them — which is why this message \
-             does not try to construct one."
+             READ ONLY ITS FIRST LINE, `Tracked-fixture probe (#3310)`: that is the \
+             half that answers this failure, naming every git-tracked fixture that is \
+             missing and printing the exact, correctly-scoped `git restore` for them — \
+             which is why this message does not try to construct one.\n\
+             \n\
+             IGNORE the `ERROR:` lines that follow it, and do NOT run the remedy they \
+             suggest. They report a DIFFERENT condition — the checkout corpus is not a \
+             complete FETCHED corpus, which it is never meant to be — and the command \
+             exits 1 with 7 such lines even when nothing is missing. One of them \
+             proposes clearing `.dataset-pin` and re-running the fetch, whose default \
+             path is destructive (`rm -rf` on the dataset root): against this checkout \
+             root that would DELETE the committed fixtures you are trying to restore."
         )
     });
     let golden = golden_rows_by_pk(&dir);
--- a/cqlite-core/tests/issue_3358_bti_query_token_bound.rs
+++ b/cqlite-core/tests/issue_3358_bti_query_token_bound.rs
@@ -46,11 +46,21 @@
 //! documented — #3310 added reporting for git-tracked fixtures a SIGKILLed fetch
 //! left deleted — which is exactly the case a skip would swallow.
 //!
-//! `CQLITE_DATASETS_ROOT` is honored first, then the in-repo `test-data/datasets`
-//! corpus; EVERY candidate root is walked for this TABLE's own `Data.db` rather than
-//! committing to a root by keyspace (issue #3220), because neither root is a superset
-//! — the root `fetch-datasets.sh` prints does not carry this fixture and the checkout
-//! does.
+//! Fixture roots resolve through the SHARED `support/datasets_root.rs` resolver, not a
+//! private copy: `CQLITE_DATASETS_ROOT` is honored first, then the in-repo
+//! `test-data/datasets` corpus, and EVERY candidate is probed for this TABLE's own
+//! `da-1-bti-Data.db` rather than committing to a root by keyspace (issue #3220),
+//! because neither root is a superset — the root `fetch-datasets.sh` prints does not
+//! carry this fixture and the checkout does.
+//!
+//! Using the shared module rather than re-deriving the rule is what makes the
+//! fail-closed guard OBSERVABLE (#3220 AC3). The checkout candidate is built from
+//! `CARGO_MANIFEST_DIR`, a COMPILE-TIME constant, so with a private resolver the only
+//! way to stage "the committed fixture is unreadable" is deleting a git-tracked file
+//! from the working tree — which also trips the gate's mid-run `tree-integrity` check
+//! (#2926). The shared module's `CQLITE_TEST_CHECKOUT_SSTABLES_ROOT` seam substitutes
+//! the checkout candidate instead, and can only ever REMOVE fixtures from view, so a
+//! stray value makes this lane FAIL, never pass vacuously.
 
 #![cfg(feature = "state_machine")]
 
@@ -66,11 +76,16 @@ use cqlite_core::storage::sstable::reader::{SSTableReader, ScanTokenBound};
 use cqlite_core::util::cassandra_murmur3::cassandra_murmur3_token;
 use cqlite_core::Config;
 
+// The #3220 fixture-root rule lives HERE, once, and is regression-tested on synthetic
+// roots by `issue_3220_datasets_root_resolution.rs`. A private re-derivation in this
+// file would be the very shape #3220 exists to remove, and would carry no test seam.
+#[path = "support/datasets_root.rs"]
+mod datasets_root;
+
 const KEYSPACE: &str = "test_da";
 const TABLE: &str = "wide_multiclustering_small";
 const SSTABLE_PREFIX: &str = "da-1-bti";
 
-/// Repo root = the parent of this crate's manifest dir (`<repo>/cqlite-core`).
 /// Single-quote a path for safe pasting into a shell.
 ///
 /// Checkout paths on this fleet contain no spaces today, but a remedy is copied by
@@ -81,25 +96,28 @@ fn shell_quote(p: &Path) -> String {
     format!("'{}'", p.display().to_string().replace('\'', "'\\''"))
 }
 
-fn repo_root() -> PathBuf {
-    Path::new(env!("CARGO_MANIFEST_DIR"))
-        .parent()
-        .expect("cqlite-core has a parent repo dir")
-        .to_path_buf()
-}
-
-/// The `<table>-<cfid>` generation dir, requiring a real `Data.db` so a JSONL-only
-/// root is passed over rather than yielding zero rows. `None` means no candidate root
-/// held it, which [`fixture`] turns into a hard failure (this fixture is committed).
+/// The `<table>-<cfid>` generation dir holding this fixture's `da-1-bti` generation.
+///
+/// Candidate roots come from [`datasets_root::sstables_root_candidates`] — the shared,
+/// deduplicated, workspace-anchored list, including the override seam the RED
+/// verification uses — and each is probed with THIS generation's own `Data.db`. The
+/// first root that actually holds it wins: resolution is by EVIDENCE, never by a fixed
+/// env-first/checkout-first preference, because neither root is a superset (#3104).
+///
+/// The probe is `{SSTABLE_PREFIX}-Data.db` rather than the shared
+/// [`datasets_root::table_has_data`]'s any-`*-Data.db` test, matching the oracle's
+/// `sstables_root_for_case` precedent for a format-specific fixture. The stricter
+/// predicate matters in the safe direction: a root carrying some OTHER generation of
+/// this table would satisfy `table_has_data`, win the selection, and then hard-FAIL
+/// this `must_run` lane on a checkout that does hold the `da` copy — a guard that reds
+/// on correct input is the guard agents learn to waive.
+///
+/// `None` means no candidate held it, which [`fixture`] turns into a hard failure —
+/// this fixture is committed source.
 fn fixture_dir() -> Option<PathBuf> {
-    let roots = std::env::var("CQLITE_DATASETS_ROOT")
-        .ok()
-        .map(PathBuf::from)
-        .into_iter()
-        .chain(std::iter::once(repo_root().join("test-data/datasets")));
-    for root in roots {
-        let keyspace_dir = root.join("sstables").join(KEYSPACE);
-        let Ok(entries) = std::fs::read_dir(&keyspace_dir) else {
+    let data_db = format!("{SSTABLE_PREFIX}-Data.db");
+    for root in datasets_root::sstables_root_candidates() {
+        let Ok(entries) = std::fs::read_dir(root.join(KEYSPACE)) else {
             continue;
         };
         let mut candidates: Vec<PathBuf> = entries
@@ -112,10 +130,7 @@ fn fixture_dir() -> Option<PathBuf> {
             })
             .collect();
         candidates.sort();
-        if let Some(dir) = candidates
-            .into_iter()
-            .find(|p| p.join(format!("{SSTABLE_PREFIX}-Data.db")).is_file())
-        {
+        if let Some(dir) = candidates.into_iter().find(|p| p.join(&data_db).is_file()) {
             return Some(dir);
         }
     }
@@ -265,13 +280,14 @@ fn fixture() -> Fixture {
         // caller's environment and the caller's directory. A remedy that is correct only
         // from the repository root is a remedy that gets pasted from somewhere else and
         // quietly reads as an all-clear — the exact silent-pass this test exists to stop.
-        let root = repo_root();
+        // The roots named here come from the RESOLVER, via `describe_search`, so the
+        // message can never name a candidate list the lookup above did not use.
+        let root = datasets_root::repo_root();
         let verify_cmd = shell_quote(&root.join("test-data/scripts/fetch-datasets.sh"));
         let verify_root = shell_quote(&root.join("test-data/datasets"));
-        let default_root = root.join("test-data/datasets").display().to_string();
+        let searched = datasets_root::describe_search(KEYSPACE, TABLE);
         panic!(
-            "{KEYSPACE}.{TABLE} `{SSTABLE_PREFIX}-Data.db` was not found under any \
-             candidate dataset root (CQLITE_DATASETS_ROOT, then {default_root}).\n\
+            "{KEYSPACE}.{TABLE} `{SSTABLE_PREFIX}-Data.db` was not found: {searched}.\n\
              \n\
              This fixture is COMMITTED SOURCE (git-tracked, not gitignored), so this \
              is a broken checkout rather than a missing optional corpus, and it fails \
--- a/cqlite-core/tests/issue_3358_bti_query_token_bound.rs
+++ b/cqlite-core/tests/issue_3358_bti_query_token_bound.rs
@@ -281,8 +281,6 @@ fn fixture() -> Fixture {
         // caller's environment and the caller's directory. A remedy that is correct only
         // from the repository root is a remedy that gets pasted from somewhere else and
         // quietly reads as an all-clear — the exact silent-pass this test exists to stop.
-        // The roots named here come from the RESOLVER, via `describe_search`, so the
-        // message can never name a candidate list the lookup above did not use.
         let root = datasets_root::repo_root();
         let verify_cmd = shell_quote(&root.join("test-data/scripts/fetch-datasets.sh"));
         let verify_root = shell_quote(&root.join("test-data/datasets"));
@@ -294,7 +292,8 @@ fn fixture() -> Fixture {
         // `fetch-datasets.sh`, the destructive invocation the rest of this message
         // spends six lines warning against. So the root list is taken from the
         // resolver and the sentence is composed here, naming what was actually
-        // required.
+        // required. Taking the list from the resolver is what stops the message naming
+        // a candidate set the lookup above did not use.
         let roots = datasets_root::sstables_root_candidates()
             .iter()
             .map(|r| r.display().to_string())
--- a/cqlite-core/tests/issue_3358_bti_query_token_bound.rs
+++ b/cqlite-core/tests/issue_3358_bti_query_token_bound.rs
@@ -229,8 +229,19 @@ fn fixture() -> Fixture {
         // wrong AND unrunnable (`<` is shell input redirection). Quoted so the shell
         // passes the `*` through to git rather than expanding it against a tree where
         // the directory is, by definition, missing.
-        let restore_pathspec = format!("test-data/datasets/sstables/{KEYSPACE}/{TABLE}-*");
-        let default_root = repo_root().join("test-data/datasets").display().to_string();
+        //
+        // `:(top)` anchors it at the repository root, because `cargo test -p cqlite-core`
+        // is routinely run FROM `cqlite-core/`, where a crate-relative `test-data/...`
+        // names nothing. The verification script is printed absolute for the same reason:
+        // a remedy that only works from one directory is a remedy that will be pasted
+        // and fail.
+        let restore_pathspec = format!(":(top)test-data/datasets/sstables/{KEYSPACE}/{TABLE}-*");
+        let root = repo_root();
+        let default_root = root.join("test-data/datasets").display().to_string();
+        let verify_cmd = root
+            .join("test-data/scripts/fetch-datasets.sh")
+            .display()
+            .to_string();
         panic!(
             "{KEYSPACE}.{TABLE} `{SSTABLE_PREFIX}-Data.db` was not found under any \
              candidate dataset root (CQLITE_DATASETS_ROOT, then {default_root}).\n\
@@ -246,8 +257,8 @@ fn fixture() -> Fixture {
              \n\
              \x20   git restore -- '{restore_pathspec}'\n\
              \n\
-             `bash test-data/scripts/fetch-datasets.sh --verify-only` names any such \
-             deletions and prints the exact restore command."
+             `bash {verify_cmd} --verify-only` names any such deletions and prints \
+             the exact restore command."
         )
     });
     let golden = golden_rows_by_pk(&dir);
--- a/cqlite-core/tests/issue_3358_bti_query_token_bound.rs
+++ b/cqlite-core/tests/issue_3358_bti_query_token_bound.rs
@@ -300,18 +300,28 @@ fn fixture() -> Fixture {
              \n\
              \x20   CQLITE_DATASETS_ROOT={verify_root} bash {verify_cmd} --verify-only\n\
              \n\
-             READ ONLY ITS FIRST LINE, `Tracked-fixture probe (#3310)`: that is the \
-             half that answers this failure, naming every git-tracked fixture that is \
-             missing and printing the exact, correctly-scoped `git restore` for them — \
-             which is why this message does not try to construct one.\n\
+             It writes to STDERR, so keep `2>&1` if you pipe it. Read the FIRST line \
+             and branch on it — the two outcomes are different conditions with \
+             different answers:\n\
              \n\
-             IGNORE the `ERROR:` lines that follow it, and do NOT run the remedy they \
-             suggest. They report a DIFFERENT condition — the checkout corpus is not a \
-             complete FETCHED corpus, which it is never meant to be — and the command \
-             exits 1 with 7 such lines even when nothing is missing. One of them \
-             proposes clearing `.dataset-pin` and re-running the fetch, whose default \
-             path is destructive (`rm -rf` on the dataset root): against this checkout \
-             root that would DELETE the committed fixtures you are trying to restore."
+             (1) `ERROR: TRACKED FIXTURES MISSING … (issue #3310)` — this is your case. \
+             That block names every tracked fixture that is absent and then prints an \
+             exact `git -C <repo> restore` command covering ONLY those paths (a second \
+             one with `--staged --worktree` if any deletion was staged). Run it \
+             verbatim. The probe reports only: it creates, deletes and restores \
+             nothing. This is why this message does not construct a restore command \
+             itself — five earlier attempts to hand-build one produced five different \
+             defects, and the tool's version is scoped and tested.\n\
+             \n\
+             (2) `Tracked-fixture probe (#3310): OK` — every tracked fixture is \
+             present, so the fault is NOT a deleted fixture and this test's own \
+             resolution is what to look at next. ONLY in this case are the `ERROR:` \
+             lines that follow unrelated noise: they report that the corpus is not a \
+             complete FETCHED corpus, which a checkout is never meant to be, and the \
+             command exits 1 with 7 such lines even though nothing is missing. Do NOT \
+             run the remedy they propose there — clearing `.dataset-pin` and re-running \
+             the fetch takes its destructive path (`rm -rf` on the dataset root), which \
+             against a checkout root DELETES the committed fixtures."
         )
     });
     let golden = golden_rows_by_pk(&dir);
--- a/cqlite-core/tests/issue_3358_bti_query_token_bound.rs
+++ b/cqlite-core/tests/issue_3358_bti_query_token_bound.rs
@@ -248,16 +248,27 @@ fn fixture() -> Fixture {
         // than reproducing it here badly, so this message names ONE command and lets
         // that tool produce the restore line. Do not "helpfully" add one back.
         //
-        // `env -u CQLITE_DATASETS_ROOT` is LOAD-BEARING, not tidiness. The probe is
-        // scoped to whatever root it is given, and on a fleet box that variable points
-        // at a machine-local corpus OUTSIDE any git work tree — where it correctly
-        // reports `NO SUBJECT` and exits 0, detecting nothing. Measured: with the
-        // variable set to /data/datasets the remedy finds none of the 10 deleted files;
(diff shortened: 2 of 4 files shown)


pmcfadin:

## The audit

```

FREE — not queued, never were
```

1. full **gate of record**
2. **C** (intent audit)
3. final **roborev**

## If you are IN the queue

pmcfadin:

## What I measured, just now

```
issue-3725 (#3786)  clean
```

```
```

## The new rule

pmcfadin:
## ✅ GATE OF RECORD: PASS at `0912dae9ba86e8ce9ca7af6324be9be73de9f4ff` — 37/37, zero FAIL, zero SKIP

Posting the block here because this box (`ip-172-31-6-169`) is being drained for decommission and the summary file otherwise lives only in its `/tmp`.

Binding, verified rather than assumed: `run-id` is the id this lane launched (not a recency pick — there are ~24,800 stale run dirs on this box); `commit:` and both `tree-*` lines equal the certified sha; `tree-integrity: PASS`.

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.2D7JWJ
commit: 0912dae branch: issue-3220-harden-3358-test dirty: no
schemas: 8/8 canonical .cql readable under /data/lanes/lane-3731/test-data/schemas (checkout-relative)
component-set: PASS (37/37 names vs origin/main da2ee4b309bccbb42add4ec184606dedd2fc981c; NAMES ONLY — not implementations, and no component is run here) — baseline read via the committed manifest; objects: baseline REUSED from this lane's SHARED store — store TRUSTED, not verified (#3746)
tree-start: 0912dae9ba86 dirty: no digest: 8e40187c168c
tree-end: 0912dae9ba86 dirty: no digest: 8e40187c168c
tree-integrity: PASS
fmt:               PASS (11s)  [fmt workspace features=n/a]
clippy:            PASS (465s)  [clippy workspace(excl 5) --all-features | clippy cqlite-core --features 33:all-compression,arrow,bench-internals,+30 more | clippy cqlite-cli --features 10:benchmarks,ci_zero_tolerance,cli-helpers,+7 more | clippy cqlite-flight+cqlite-py+cqlite-node --features cqlite-node/write-support]
roborev-lints:     PASS (109s)  [no-cargo]
core-tests:        PASS (577s)  [nextest run cqlite-core --features cli-helpers | test cqlite-core --features cli-helpers]
tombstones-scan:   PASS (57s)  [test cqlite-core --features write-support,cli-helpers,tombstones]
scan-offload-guard: PASS (192s)  [test cqlite-core --features cli-helpers,scan-offload-probe x2]
work-counters-guard: PASS (90s)  [test cqlite-core --features write-support,cli-helpers,state_machine,work-counters]
arrow-parity-guard: PASS (80s)  [test cqlite-core --features arrow]
memory-budget:     PASS (459s)  [test cqlite-core --features cli-helpers,dhat-heap,arrow x3 | test cqlite-flight --features dhat-heap]
integration-tests: PASS (61s)  [test cqlite-integration-tests default-features x2]
format-compat:     PASS (42s)  [test format-compatibility-tests default-features]
write-tests:       PASS (154s)  [test cqlite-core --features write-support x3]
cli-tests:         PASS (232s)  [test cqlite-cli default-features | test cqlite-cli --features write-support]
compaction-byte-parity: PASS (10s)  [test cqlite-core --features write-support x2]
bti-multiclustering: PASS (5s)  [test cqlite-core --features state_machine,cli-helpers x2]
flight-query-semantics-oracle: PASS (84s)  [test cqlite-flight default-features]
flight-tests:      PASS (239s)  [test cqlite-flight default-features]
legacy-heuristics: PASS (369s)  [build cqlite-core --features legacy-heuristics]
feature-iso-parquet: PASS (295s)  [test cqlite-core --no-default-features --features all-compression,parquet]
feature-iso-delta-scan: PASS (157s)  [test cqlite-core --no-default-features --features all-compression,delta-scan]
python-bindings:   PASS (851s)  [via maturin: feature set NOT observed]
node-bindings:     PASS (723s)  [via npm run build (napi): feature set NOT observed]
binding-rust-tests: PASS (604s)  [test cqlite-ffi-common default-features | test cqlite-node --features write-support]
parity-report:     PASS (10s)  [run cassandra-parity default-features]
operator-metrics-doc: PASS (38s)  [run cqlite-core default-features]
tooling-tests:     PASS (2855s)  [test ws0-corpus-gen default-features | + cargo not observable: cargo may run inside ~60 nested test scripts (child processes)]
minimal-build:     PASS (76s)  [build cqlite-core --no-default-features --features all-compression | test cqlite-core --no-default-features --features all-compression]
all-features-check: PASS (123s)  [check cqlite-core --all-features | clippy cqlite-core --all-features]
smoke:             PASS (103s)  [build cqlite-cli default-features]
logs: /tmp/agent-gate.2D7JWJ
summary-file: /tmp/lane3731-GATE-OF-RECORD-1788270733.txt
heartbeat: on file: /tmp/lane3731-GATE-OF-RECORD-1788270733.txt.heartbeat interval: 20s
RESULT: PASS
```

### AC3 evidence inside this run (not a PR-comment measurement)
- `bti-multiclustering: PASS` and BOTH fail-closed positive controls executed in-gate:
  - `test_point_vs_full_failclosed.sh` (#3220 AC2) → `passed: 12  failed: 0`
  - `test_issue_3358_failclosed.sh` (#3731 AC3) → `passed: 15  failed: 0`
- The declaration lines this PR adds are present in the run stdout, naming both harnesses and the
  6 cargo runs the SUMMARY annotation structurally cannot see — the annotation indeed reads
  `[test cqlite-core --features state_machine,cli-helpers x2]` while the component drives 8.

### Rebase note
Rebased onto `da2ee4b30` (18 commits, clean, no conflict). I had predicted #3780 would leave two
dead `wraparound:` literals and break the build; **it did not** — those lines predated this branch,
so replaying onto a main that deleted them yields their absence. Nothing was edited. Verified:
`cargo test -p cqlite-core --test issue_3358_bti_query_token_bound` → 3 passed.

Next: C (spec-auditor) → roborev LAST at this head → `premerge-assert` → ready + arm `--auto`.

pmcfadin:
## Fix round: roborev job 292 (Low) — the SKIP guard could red on a CORRECT run

**Finding (job 292, Low):** `grep -q 'SKIP'` at `test_issue_3358_failclosed.sh:313,:388` scanned the whole cargo output for four bare letters, so a checkout/TMPDIR path containing them false-FAILs a healthy run.

Correct, and it is this lane's own subject aimed at its own guard: **a check that reds on correct input is the check agents learn to waive.** Fixed at `80da24098980c8d91d4a1133314baf6dc535b2ed`.

### Why the marker, not the message
One reading of the fix ("match the previous `SKIP: test_da.wide_multiclustering_small...` message") would trade a false FAIL for a **false PASS** — a skip re-introduced under different wording would slip through, and the comment there claims the guard catches a skip *under any wording*. So the anchor is the **marker** `SKIP:` — what the removed guards actually printed (`eprintln!("SKIP: {KEYSPACE}.{TABLE} Data.db is absent (gitignored corpus)")`). Wording-agnostic after the colon; unsatisfiable by any filesystem path.

### RED-verified in BOTH directions (defect reproduced, not asserted)

| output fixture | OLD `grep -q 'SKIP'` | NEW `grep -q 'SKIP:'` |
|---|---|---|
| real skip diagnostic `SKIP: test_da.wide_multiclustering_small …` | FIRES | **FIRES** (power retained) |
| correct run, paths `/data/lanes/SKIPPED-lane/…` + `TMPDIR=/tmp/SKIPME.1234` | **FIRES ← the false-FAIL** | **silent** (fixed) |

Harness still `passed: 15  failed: 0`; `bash -n` clean.

**DECLARED RESIDUAL (1 RECOGNISED)**, stated in the file: a future skip printed *without* the colon is missed. Deliberate — a missed novel spelling is recoverable, a guard nobody trusts is not.

### Re-certification: `--delta`, not a repeat full gate (#1892)
Diff since the anchor is one `scripts/tests/*.sh` file. I did **not** assume eligibility — I let the delta mode's own fail-closed classifier rule, and it accepted: `delta-files (1 allowed / 0 offending)`.

Note the two classifiers answer **different questions** and must not be conflated: `classify-docs-only.sh` returns **FULL PATH / exit 1** (a shell test script is code, so **no gate red is waivable here**), while `--delta`'s allowed set names `scripts/tests/*.sh` explicitly.

```

run-id: /tmp/agent-gate.qN9c0Z
MODE: delta (TEST/DOCS-ONLY RE-CERTIFICATION — NOT the gate of record; gate of record = the full agent-gate.sh PASS at anchor 0912dae9ba86e8ce9ca7af6324be9be73de9f4ff)
commit: 80da240 branch: issue-3220-harden-3358-test dirty: no
delta-anchor: 0912dae9ba86e8ce9ca7af6324be9be73de9f4ff (full-gate PASS commit)
delta-anchor-run-id: /tmp/agent-gate.2D7JWJ
gate-of-record: full agent-gate.sh run at 0912dae9ba86e8ce9ca7af6324be9be73de9f4ff (this DELTA re-certifies a test/docs-only diff; it is NOT a substitute for the full gate)
component-set: PASS (37/37 names vs origin/main 9da235a6cb9c48b78e019a359e8851ad8d007e09; NAMES ONLY — not implementations, and no component is run here) — baseline read via the committed manifest; objects: baseline REUSED from this lane's SHARED store — store TRUSTED, not verified (#3746)
delta-executors: scoped-tests(rust/python) shell-selftests(1) (executors that RAN this re-cert)
tree-start: 80da24098980 dirty: no digest: 489664ff3a9d
tree-end: 80da24098980 dirty: no digest: 489664ff3a9d
tree-integrity: PASS
delta-files (1 allowed / 0 offending):
      [test/docs] scripts/tests/test_issue_3358_failclosed.sh
fmt:               PASS (5s)  [fmt workspace features=n/a]
scoped-tests:      PASS (5s)  [test cqlite-core --features cli-helpers]
shell-selftests:   PASS (1s)  [cargo not observable: cargo may run inside the changed scripts/tests/*.sh (child processes)]
logs: /tmp/agent-gate.qN9c0Z
summary-file: /tmp/lane3731-DELTA-1788276816.txt
heartbeat: on file: /tmp/lane3731-DELTA-1788276816.txt.heartbeat interval: 20s
RESULT: PASS
```

The gate of record remains the full PASS at anchor `0912dae9ba86…` (posted above, comment 5496201128). `MODE: delta` is declared in the block precisely so it can never be pasted as that.

Roborev round 2 running at `80da24098…`.

pmcfadin:
## ✅ roborev CLEAN at `1370cbec533e5045913b2fcc6ea00c7a76cb3a16` — AC5 satisfied

Round 7, job 298: `findings: NONE`, `RESULT: PASS`. Genuine review — 691,930 input /
603,392 cached tokens, vs the ~18.7k/0 vacuous baseline. Every deterministic key PASS
(`push-assert`, `census-check`, `code-free`, `job-record`, `sha-assert`,
`review-completed`, `prompt-content: PASS (4/4 code census paths present)`,
`vacuity-tier1`, `vacuity-tier2`).

Seven rounds, seven findings, **all seven inside the guards I added** rather than in the
test under change — which is itself the useful signal about where this PR's risk sat.

### Round 6 (job 297, Low) — fixed: a FAIL destroyed the evidence it told you to read

Reproduced before fixing: planted an unsatisfiable absent-case assertion; the FAIL printed
`see /tmp/i3358-failclosed.ruATwK/absent.log`, a path the EXIT trap had already deleted.
The control case printed its log; the absent and empty cases only *named* theirs.

Worse inside the gate than standalone — this harness runs under `bti-multiclustering`, and
the only gate text an agent retains is the SUMMARY block, so a component FAIL here left
**nothing to triage**. That is the "defect that produces no artifact" class, in a harness
whose whole purpose is making a silent non-run impossible.

Fixed at the one place covering every site (the EXIT trap), not at eight message sites:
preserve the scratch root when `FAIL > 0` **and excerpt every lane log to stdout**. The
excerpt is the load-bearing half — preserving a directory helps only a standalone re-run
and does nothing for the run that actually failed.

Verified in **both** directions:

| | result |
|---|---|
| passing run | exit 0, 15/0, no diagnostics block, zero leftover scratch dirs — green path unchanged |
| failing run | exit 1, all 3 lane logs excerpted, PRESERVED path exists, the named log **exists**, 8 real cargo diagnostic lines on stdout |

The 70-line `absent.log` exercised the truncation branch: marker arithmetic correct (20
omitted), and its tail reaches the true EOF — `test result: FAILED. 0 passed; 3 failed`,
i.e. the fail-closed evidence itself is now what reaches stdout.

## ⛔ Blocker, and it is NOT this diff: a regressed global-counter flake

The `--delta` re-cert at `1370cbe` FAILed `scoped-tests (10s)` on
`block_io::tests::read_compressed_chunk_at_records_one_read_per_chunk` —
`"point-read of chunk 0 must record exactly one READ_CALL"`, `left: 2 right: 1`
(3615 passed; 1 failed). `file-size`, `fmt`, `shell-selftests`, `component-set` (37/37)
and `tree-integrity` all PASS; `delta-files` 1 allowed / 0 offending.

Mechanism **established, not assumed**: the test *is* `#[serial(work_counters)]`, and its
first assert (`reset must zero READ_CALLS`) **passed** — so the reset worked and something
concurrent bumped the global between reset and assert. `READ_CALLS` is incremented by
**production** paths (`chunk_source.rs:197`, `data_access/big_promoted.rs:690`,
`data_access/compressed_offset.rs:151`, `scan_stream_windowed_read.rs:238`), so any
parallel test doing a compressed read contaminates it *without ever mentioning the
counter*. A serial-group convention cannot hold that invariant — which is why it keeps
coming back: #2006 and #2033 are both this exact test with this exact mechanism, and both
are **CLOSED**. #2451 (open) is the same class on a sibling counter and names the right fix
(test-local scoping, extending #2428).

A hypothesis I formed and **disproved** rather than filing: `read_work_counters.rs:840`
`counters_round_trip_and_reset` uses a bare `#[serial]` — a *different* group per
`block_io.rs:1228` — and calls `record_read()` five times; but `let c = Counters::new()` is
a **local** instance, so it never touches the global. Not the cause.

Two things I will not paper over:

- The X..Y delta diff is **one shell script**, which is not compiled into `cqlite-core --lib`,
  so this failure cannot be attributable to it. But `scripts/ci/classify-docs-only.sh`
  **exits 1** on the branch diff (it contains `cqlite-core/Cargo.toml`), so the #3042 prose
  CITE-AND-WAIVE does **not** apply here. That is a different question from `--delta`
  eligibility and I am not conflating them.
- `premerge-assert.sh` has **no waiver path** for its delta argument, so a genuine `PASS` is
  required. Load was **88** (a peer lane's gate) during that run and is now ~27. I am
  filing the regression first, then re-running — with the failed run disclosed either way,
  not quietly replaced.

pmcfadin:
## 🟢 AC6 — pre-merge assert PASSED and auto-merge is ARMED

Certified sha: `1370cbec533e5045913b2fcc6ea00c7a76cb3a16`
Auto-merge armed at `2026-09-01T16:31:35Z` (`--auto --squash --delete-branch`).
GitHub owns the wait; no CI polling from this lane.

### `premerge-assert.sh` — Case B (4 args), exit 0

```
PREMERGE: OK 1370cbec533e5045913b2fcc6ea00c7a76cb3a16
PREMERGE: GATE-OF-RECORD commit: 0912dae tree-start: 0912dae9ba86 tree-integrity: PASS dirty: no
PREMERGE: DELTA-RECERT  anchor: 0912dae9ba86… commit: 1370cbe tree-start: 1370cbec533e tree-integrity: PASS dirty: no
```

Case B was the correct form here, not a defect: #1892 *mandates* `--delta` for a test-only
diff on top of a full PASS, so a 3-arg call would have red on doctrine-mandated input.

### The `--delta` FAIL I disclosed is resolved, and disclosed rather than replaced

`DELTA6` FAILed `scoped-tests` on the regressed global-counter flake. I filed it as **#3833**
(third filing of that exact test — #2006 and #2033 are both CLOSED, and a closed issue
evidences nothing per #3626) with the mechanism established from source: `READ_CALLS` is
incremented by four **production** paths, so any parallel compressed read contaminates it
without mentioning the counter — which is why a `#[serial(work_counters)]` convention cannot
hold it, and why two fixes each closed and came back.

`DELTA7` then PASSed at the same sha on a quieter box (load 88 → ~35). Both runs are on the
record; I did not quietly substitute the green one.

### C audit still applies — verified, not assumed

C PASSed at `0912dae9b`. Everything since is `scripts/tests/test_issue_3358_failclosed.sh`
alone, and the three AC-bearing artifacts are **byte-identical** to the audited sha:

| artifact | vs C sha |
|---|---|
| `cqlite-core/tests/issue_3358_bti_query_token_bound.rs` | IDENTICAL |
| `cqlite-core/Cargo.toml` | IDENTICAL |
| `scripts/agent-gate.sh` (AC3 wiring at `:12056`) | IDENTICAL |

The harness itself changed twice (roborev jobs 296, 297); both verified `15/0`, and the
delta's `shell-selftests` component executed it.

### Two limits I am stating, not burying

**1. Base staleness — `verdict STALE-RECOGNISED`.** 8 commits behind the merge-base; **3**
touch the blast radius, including `a0d792003`, which touches `scripts/agent-gate.sh` — a file
this diff also touches. It is additive (+17 lines at ~14879 in `run_tooling_tests()`), nowhere
near this diff's hunk at ~12001–12060 in `run_bti_multiclustering`; `merge-tree` reports **0
conflicts** and `component-set` is **PASS 37/37**. But `component-set` checks NAMES, not
implementations, so my certification did not execute main's updated `tooling-tests` content.
That is the #3650 gap exactly: **the certified tree is not the post-squash tree.** Slice 2
does not exist, so this is reported, not closed.

**2. Wording.** Per #3650 this is *"gate of record verified at `0912dae9ba86…`, delta-recertified
at `1370cbec533e…`"* — **not** "certified against main". A squash composes this diff with main's
current tip, which no gate here has executed.

### Scorecard

| AC | verdict |
|---|---|
| 1 rebased onto current `origin/main` | ✅ |
| 2 committed-source distinction ASSERTED | ✅ C PASS — documented deviation: uses `sstables_root_candidates()` + a **stricter** generation-specific probe rather than `sstables_root_for_table` verbatim; per-case `must_run`, zero `assert!(ran>0)`, 3 SKIP guards deleted |
| 3 each guard RED-verified | ✅ observed firing in-run (absent rc=101 + `0 passed; 3 failed`; empty rc=101 + 3 zero-byte rejections; no SKIP) |
| 4 gate of record PASS | ✅ 37/37, zero FAIL, zero SKIP, `tree-integrity: PASS` |
| 5 roborev clean | ✅ round 7, job 298, `findings: NONE` |
| 6 premerge + arm | ✅ exit 0, armed |

Issue #3705, opened 2026-08-31
agent-gate: the code a run EXECUTES is not verified against the tree it CERTIFIES (in-queue gate-script change; needs a bootstrap/re-exec handshake)


pmcfadin:
Split out of #3544 by lead ruling on `REQ-3544-TRIGGER`. #3544 ships with the in-queue gate-script check **REMOVED**; this issue owns it.

## The property

`agent-gate.sh` re-captures its certification window after the #1825 slot is granted (`_tree_recapture_after_slot`, #2926), deliberately: a queue can be hours long, and edits made while queued should start a fresh window rather than void the run. But the process continues executing the **function definitions it loaded before the queue**. So the tree being certified and the code doing the certifying can diverge, and nothing detects it.

Wanted: **the code this process is executing is the code the certified tree contains.**

## Why it cannot ride inside a component-set comparison — finding 1, verbatim

roborev job 294, High, `scripts/agent-gate.sh:5983`:

> The "startup" digest hashes `$GATE_SELF` only after thousands of lines have already been parsed. If an editor or checkout atomically replaces the script before this point, Bash continues executing the old inode while the digest reads the new path; subsequent comparisons can therefore pass and certify the new tree using old gate definitions.
>
> **Fix**: Establish executable identity before loading the gate body with a bootstrap/re-exec handshake, or re-exec the on-disk gate after the post-slot recapture. Add a test that atomically replaces the script before the initial digest.

**Bash parses incrementally.** "The code I am running" is therefore not knowable from inside the running process at any point after parsing begins — no digest taken *by the script* can establish it, because the digest reads a path while the interpreter holds an inode. This is a change to how `agent-gate.sh` **starts up**, affecting every mode and every component. It is not a subordinate detail of a component-set comparison.

Confirmed by measurement, not argument: an in-place `cp` over the running script made a queued fixture gate **re-execute** `acquire_gate_slot` from a shifted byte offset, spawning a second slot daemon for the same `--gate-pid` (both at the gate's default `--poll-secs 2`) and wedging the queue permanently. So the consequence of an in-queue edit is not merely "a component might not run" — it is arbitrary re-execution.

## Why five rounds landed here — the split boundary

| roborev job | finding | owner |
|---|---|---|
| 285 | symlink **manifest** ⇒ the two halves read different documents | component-set (#3544) |
| 290 | the check sat OUTSIDE the window it certifies | **in-queue** |
[...]

Issue #3617, opened 2026-08-30
fleet-runbook: an auth-shaped git failure may be a SHARED remote-URL write, not a broken box — check it first


pmcfadin:
## The defect

CLAUDE.md and `docs/development/fleet-runbook.md` map a claim/heartbeat **git push failure** to a per-box cause:

> `claim.sh` now calls that `ERROR reason=auth (NOT retryable)` instead of the old misleading `reason=infra (transient — retry)` — do not retry it, **fix the box** (`gh auth setup-git`, or `bash scripts/bootstrap-agent-machine.sh --yes`)

That mapping is correct for its intended cause and **confidently wrong for another one that is indistinguishable at the point of failure**: a bad *shared* remote URL. An agent following doctrine would go credential-hunting on a completely healthy machine.

## How it was found (a real incident, 2026-08-30 ~19:29–19:42Z)

While RED-testing a summary-injection fix on #3544, this lane ran `git remote set-url origin "<newline-bearing URL>"` inside a **throwaway `git worktree`**, on the assumption that a throwaway worktree is isolated. **A worktree shares `$GIT_DIR/config`** — there is no worktree-local remote — so that wrote to `/data/lanes/repo/.git/config`, shared by four active lanes, the root checkout and several detached worktrees.

Observed symptoms across lanes:

```
# git side (this lane, on push)
fatal: credential url cannot be parsed: https://github.com/pmcfadin/cqlite.git
warning: url contains a newline in its path component

# gh side (lane-3500, on `gh issue create` without -R)
none of the git remotes configured for this repository point to a known GitHub host.
To tell gh about a new GitHub host, please use `gh auth login`
```

**Both error texts steer the reader to an auth remedy.** `gh` names `gh auth login` explicitly; the git-side failure matches the documented `reason=auth` shape. Neither mentions the remote URL, which was the actual cause.

lane-3500 confirmed it had already misdiagnosed it in real time ("a transient hiccup, likely a peer's concurrent git op") and worked around it with `-R pmcfadin/cqlite` — right neighbourhood, wrong mechanism.

## The failure IS uniform on the git side — and an earlier version of this issue said otherwise

[...]

PR #3710, opened 2026-08-31
fix(#3544): a gate script BEHIND origin/main cannot certify — fail-closed component-set skew pre-flight

### LINKED ISSUE #3544: gate: a branch predating a gate expansion certifies against the OLD component set — merge-cleanliness cannot detect it
## The mechanism

**`scripts/agent-gate.sh` is read from the tree under test.** So a branch whose `scripts/agent-gate.sh` predates a component-set expansion on `main` runs the **old** script, reports `N/N nonpass=0`, and is **silent about every component added since** — while being a perfectly true statement about the set it knew.

**Merge-cleanliness cannot detect this.** Nothing textually conflicts; `git merge-tree --write-tree` returns CLEAN. The skew is semantic, not textual.

Measured on PR #3467 (#3393) tonight:

| | `scripts/agent-gate.sh` | components (`--list`) |
|---|---|---|
| branch `b2c5a02f1c14` | 8,779 lines | **31** |
| `origin/main` `25fe90bea` | 11,598 lines | **35** |
| delta | **+2,819** | **+4** (incl. `feature-matrix`) |

PR #3403 landed that +2,819-line expansion at 02:08Z, so **every branch cut before then has this skew**, and the gate placed on #3467 at 02:47Z — 39 minutes *after* the expansion merged — would have produced a `31/31` PASS covering 31 of 35 components.

## Why "rely on `required` for the new components" does not work

The tempting mitigation is to bank the old-set PASS and let CI cover the rest. **For `feature-matrix` that fails, and it fails in a circle.** From `.github/ci-gating-tiers.yml`:

```yaml
- workflow: ci-minimal-features.yml
  reason: Feature-matrix build lane; the merge-gating minimal-features build runs in the local gate.
```

The CI feature-matrix lane is **exempt from `required` on the grounds that the local gate owns it**. So a stale local gate does not run `feature-matrix`, and `required` does not poll the CI lane because it is deferring to the local gate. **Each side's coverage is justified by the other's, and the component is exercised by neither** — with a documented rationale on both sides explaining why that is fine.

This is the same shape CLAUDE.md already records one step over — *"any PR whose subject is a config a daemon or gate reads from root cannot certify itself"* — except here the gate does not have to be the PR's *subject*. Any branch old enough is affected.

## Proposed fix

A **pre-flight component-set assert** in the gate itself: compare the running script's component set against `origin/main`'s and refuse (or emit a named, loud, non-PASS line) when the branch's set is a strict subset. Sketch:

```bash
git show origin/main:scripts/agent-gate.sh > "$tmp" && bash "$tmp" --list | sort > "$main_set"
bash "$0" --list | sort > "$branch_set"
comm -13 "$branch_set" "$main_set"   # non-empty ⇒ branch is behind
```

Design notes worth honoring:
- **Fail closed, and name the missing components** — a bare "you are behind" is not actionable; `missing: feature-matrix, …` is.
- **Do not make it a silent SKIP.** A SKIP is exactly the vacuous pass this issue is about.
- **Requires network/`origin` freshness**, so the failure mode when `origin/main` is unreachable must be an explicit non-PASS naming *that*, never a permissive pass. (Same rule as every other affirmative-measurement check in this repo: never derive a pass from the absence of a bad signal.)
- The comparison must be on the **component set**, not a line count or file hash — a refactor that changes 2,000 lines without changing the set is not a coverage problem.

Interim, until this lands: **whoever places a gate compares `--list` against `origin/main` first.** That is a one-line check and it would have saved a 70–95 minute run tonight.

## Credit

Found by the #3393 lane, prompted by an unrelated request from the coordination lead to watch for *merge-order* conflicts. It checked that, found the relevant landing predated its own gate by 39 minutes, and identified a mechanism orthogonal to what it had been asked about.

pmcfadin:
Closes #3544.

## The defect

`scripts/agent-gate.sh` is read **from the tree under test**. A branch whose copy predates a component-set expansion on `main` runs the OLD script, reports `N/N nonpass=0`, and is **silent about every component added since** — a true statement about a set that is no longer the coverage claim. Merge-cleanliness cannot detect it: `git merge-tree` returns CLEAN, because the skew is semantic, not textual. Measured on #3467: a gate placed 39 minutes after #3403's expansion landed would have produced a `31/31` PASS covering 31 of 35 components.

"Let CI cover the new components" fails in a circle: `.github/ci-gating-tiers.yml` exempts the CI feature-matrix lane **on the grounds that the local gate owns it**, so a stale local gate does not run it and `required` does not poll it. Each side's coverage is justified by the other's; the component is exercised by neither.

## The mechanism: the baseline is DATA, never executed code

The pre-flight needs `origin/main`'s component set. It used to get it by fetching main's `agent-gate.sh` and **running** it (`bash <fetched> --list`). It no longer does. Nothing fetched is executed — verified: zero invocations of a fetched script remain.

`scripts/agent-gate.components` is the component set committed **as data**, read with `git show <baseline-sha>:scripts/agent-gate.components`.

**What this converts the six execution-driven Highs INTO — not what it eliminates, and NOT a claim about the whole operation.** Six of the seven High-severity findings in this PR's eleven review rounds traced to one decision: executing the fetched script. They are not gone. **The worst a redirected, rewritten or impersonated fetch can now produce is a WRONG COMPONENT LIST, which the comparison itself detects — instead of arbitrary code execution with the developer's credentials.** That is a severity conversion, and a reader should not have to reconstruct it.

**The precise scope of that claim, because a sentence true of a component and false of the system is worse than an obviously narrow one.** The manifest converts the baseline's **content** from code into data, so a redirected or forged baseline yields a **wrong component list, which the comparison detects**. It does **not** make the *fetch mechanism* safe: `git fetch` under a hostile local config can invoke a remote helper and execute commands regardless of what the fetched bytes are. That path is closed separately — by removing transport into the live repository altogether, not by the manifest. Two further routes to the same execution were found after the switch (`refs/replace` substitution, and promisor lazy-fetch in a partial clone); see the review log below.

**What the manifest does NOT protect, retained and fixed on their own merits:**

- the **credential leak** in fetch stderr (round 10) — fixed by routing every external string through one redact+flatten entry point;
- the **URL-in-argv exposure** (round 11) — fixed by writing the URL to a 0600 config via a shell builtin, so it never appears in any process argument;
- transport/host/path pinning, the mode-dependent probe bound, and shallow-clone ancestry;
- **`refs/replace` substitution** (round 14) — a replacement ref transparently swaps in another commit, so the gate would print the **canonical sha while reading forged bytes**. The manifest does nothing about this: a replace ref lies *through the tool doing the reporting*. Fixed separately by refusing replacement objects on every read;
- **promisor lazy-fetch in a partial clone** (round 15) — `ls-tree`/`show` on the live repository can lazily fetch through its promisor remote, honouring its local config, so a *local read* becomes a network operation and a remote helper can execute. **Closed in round 15** by moving every baseline object read into the isolated scratch repository, with the lane's object dir supplied as an alternate (pure object storage: no config, hence no promisor). It was OUTSTANDING and escalated when this paragraph was first written; that wording survived the fix and contradicted the table below it for several rounds.

All are properties of *fetching, reading and comparing*, not of *executing the baseline*, so the switch obviates none of them — and the last two are routes to **execution** that survived the switch, which is why condition (b) above is scoped to the baseline's content rather than to the operation.

The **isolated fetch** (`GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` neutralised on the capture and the fetch, the URL kept out of `argv`, and the whole git environment built from an **allowlist** rather than a denylist) is retained as **defence in depth** rather than the primary control, at the coordination lead's direction. Note the earlier sha-verified *transfer hop* is **gone**: round 14 removed the import entirely — the baseline objects are read where they landed, via alternates scoped to the invocation — because a transport into the live repository was itself an execution vector.

### PRE-REGISTERED STOPPING TRIGGER — a fourth residual of this family means we STOP

Registered here at the coordination lead's condition, in the PR body rather than only on the thread, so it is visible to whoever reads this change next rather than to whoever followed the discussion.

**The family:** a route by which *untrusted repository state* — config, refs, remotes, helpers, promisors, object substitution — causes **execution** or another irreversible effect during a baseline read. Three instances have been closed, each by removing a mechanism rather than constraining its inputs:

| # | route | closed by |
|---|---|---|
| 1 | `url.*.insteadOf` rewriting a validated URL | isolating the fetch; env built from an **allowlist**, not a denylist |
| 2 | `protocol.ext.allow` + `insteadOf` invoking an `ext::` helper during hop 2 | removing transport into the live repository entirely |
| 3 | promisor **lazy fetch** turning a local `ls-tree`/`show` into a network operation | moving every baseline object read into the isolated scratch repo |

## DECLARED RESIDUAL you are approving, tracked as #3717

`_component_set_bounded`'s `timeout`/`gtimeout` arms do not terminate descendants that outlive a
**successful** child. The captured-stdout replay IS bounded (1 MiB cap, oversize refused), so the gate
cannot hang. A surviving descendant's **writes** are not bounded: it holds the capture fd, cleanup
unlinks the file, and an unlinked inode consumes disk until the last fd closes.

**Demonstrated, not theorised** — a planted child that backgrounds a writer and exits 0:

```
pid 518642 fd 1 -> /tmp/agent-gate-bcap.E0hYnc (deleted)   4 capture fds, all DELETED
inode 563B -> 605B over 3s                                 still growing after the call returned
```

**Dormant, and that is a property of the CALLERS, not of this function:**

```
6 clean full pre-flight runs, capture fds held: 0   (ground verified empty before AND after;
                                                     instrument proven by finding the 4 above)
a real bounded `git fetch` whose 2s bound FIRED:
   surviving git-remote-https: 2   <- real callers DO strand descendants
   capture fds held:           0   <- but they do NOT hold the capture
```

*Inference, offered as inference and unverified:* git wires transport helpers over **pipes**, so a
helper's stdout is a pipe to the parent git and not the capture it inherited. **The `0` is the
measurement; the mechanism is a hypothesis.**

**The re-check condition, stated precisely so it is testable:** a caller that **backgrounds a writer
onto its own stdout**. Nothing in this pre-flight does that. Add one and the residual becomes live
with no warning — which is why it is declared at the site and filed as **#3717** rather than left as
an absence someone must rediscover.

**Why it is not fixed here.** A post-success process-group kill is already rejected in this function
on record: bash reaps the leader as soon as it exits, so the pgid is free for **reuse** and the kill
could land on an unrelated group — this repo's incident is a pattern kill that destroyed a peer lane's
gate at component 28 of 30. Doing it safely means owning a supervisor on **every** arm, i.e.
restructuring the function that has produced **four** findings (jobs 246, 279, 282, 297), one of which
was *created by* the previous fix to it. `ulimit -f` was tried and rejected on measurement: per
**process**, not per fd, so it also caps the 92 MB pack a legitimate `git fetch` writes, making a safe
limit weak and a growing legitimate write a false FAIL in a certifying mode.

## SCOPE BOUNDARY: ssh transports are out of scope for the canonical baseline, by decision

**This is a boundary, not a gap.** The pre-flight certifies a baseline fetched from `origin`. It does
that for `https://` origins and **declines to do it at all** for ssh ones, with a named refusal and a
one-command remedy. Nothing is left partially checked and nothing depends on a property that is not
established — an ssh origin produces no verdict rather than an unsound one.

**The reason it cannot be a residual instead.** Every other route in this family degrades, under the
data-manifest design, to *a wrong component list which the comparison itself detects*. This one does
not, because the harm lands **during the transport**, before any list exists to compare:
`~/.ssh/config` can carry `ProxyCommand` or `Match exec`, which **execute**. A check placed after a
harmful effect can only report it — so where the harm is execution, the control must be that the
execution is unreachable, not that its result is detected. That is why the accepted grammar narrowed
instead of the environment being hardened.

**And it is a consequence of a deliberate choice, not an isolation oversight.** The isolated
environment ADMITS `HOME` on purpose, for ssh key and `known_hosts` discovery — without it an ssh-form
origin cannot authenticate at all. So the exposure is the price of supporting ssh, and declining ssh is
what removes it. `HOME` stays admitted (https credential helpers need it); what leaves is the transport
that made `~/.ssh/config` reachable.

### THE FAMILY, NAMED ONCE — four execution routes, one shape

**THE SHAPE: untrusted state that the fetch CONSULTS can redirect or execute during the read, and a
check downstream of the read cannot prevent either.** Four instances, each fixed by removing the
channel rather than validating its contents — and each fix moved the boundary outward, which is the
condition the stopping trigger encodes:

| # | untrusted state | what it achieves | removed by |
|---|---|---|---|
| 1 | `url.*.insteadOf` (git config) | rewrites a validated URL | isolated fetch; env built from an **allowlist**, not a denylist |
| 2 | `protocol.ext.allow` + `insteadOf` | invokes an `ext::` **helper** on the transfer hop | no transport into the live repository at all |
| 3 | promisor remote in a partial clone | turns a *local* `ls-tree`/`show` into a network op under local config | every baseline object read moved into the isolated repo |
| 4 | **`~/.ssh/config`** (ssh-form origin) | rewrites `HostName`; `ProxyCommand`/`Match exec` **execute** | **https-only canonical grammar** |

**If a fifth appears, the answer is not to fix it.** The ruling is descope, or accept a bounded
provably non-executing residual — never harden the next route. Recorded here so that decision does not
have to be rediscovered from the review log.

Two further routes in the same family are **detection**, not execution, and are listed apart because
they forge the *observation* rather than run code: `refs/replace` substituting a forged smaller
manifest (closed by `GIT_NO_REPLACE_OBJECTS` + `--no-replace-objects` on every lane-local read), and
`$GIT_DIR/info/grafts` reclassifying ancestry, which survives `--no-replace-objects` (closed by moving
ancestry into the isolated repository on both paths).

**Why it is in the threat model at all, since an invoker editing their own config is out of it:** every
lane on this fleet runs as one user with `HOME=/home/ubuntu`, **shared and writable**. The party who
can plant `~/.ssh/config` is therefore a **peer lane** — a different agent session, with its own issue
— which is precisely the third party the model protects against. Measured, and its absence today was
*not* treated as a control: `~/.ssh/config` does not exist on this box right now, and any lane can
create it.

### WHAT REMAINS COVERED

The descope removes one transport, not the property. Still in force:

- **the baseline is DATA** — `git show <sha>:scripts/agent-gate.components`, parsed under a closed grammar; nothing fetched is executed;
- **the local staleness assert** — the committed manifest is checked against the running `COMPONENTS` array every run, including ORDER, which is what makes a manifest baseline trustworthy at all;
- **the isolated object import** — baseline objects reached via `GIT_ALTERNATE_OBJECT_DIRECTORIES` (an object *source*, not a transport); nothing written into the shared `.git`;
- **identity pinning for the transports that remain** — exact host, default port, exact path, userinfo accepted-but-redacted, `http://`/`git://`/`file://`/local paths/aliases/mirrors all refused by name;
- **the three-valued manifest presence probe**, the bounded runner, the `env -i` allowlist, and job 285's manifest mode validation (`120000` vs `100644`/`100755`).

### THE TRIGGER'S CONDITION AROSE BEFORE LANDING — and the ruling was applied

Further routes in this family surfaced during review, and the last one is a **behavioural narrowing a
reviewer must not discover by surprise**:

| # | route | closed by |
|---|---|---|
| 4 | `refs/replace` substituting a forged, smaller manifest | `GIT_NO_REPLACE_OBJECTS` + `--no-replace-objects` on every lane-local read |
| 5 | `$GIT_DIR/info/grafts` reclassifying ancestry (survives `--no-replace-objects`) | ancestry moved into the isolated repository on BOTH paths |
| 6 | **`~/.ssh/config` on an ssh-form origin** — `HostName` rewrite, or `ProxyCommand` / `Match exec` **execution** | **DESCOPE: only `https://` is canonical** |

Route 6 is why the ruling mattered. **A bounded, provably non-executing residual was NOT available** —
`ProxyCommand` executes, and the mitigating argument that a redirected baseline degrades to a *wrong
component list the comparison detects* does not reach it, because the harm lands during **transport**,
before any comparison. The review's own alternative (invoke ssh with a trusted config disabling user
config and `ProxyCommand`) is exactly the *"harden the fourth route"* the ruling forbids.

**In model because HOME is shared**, measured: every lane runs as one user with `HOME=/home/ubuntu`,
writable by all, so the party who can plant that file is a **peer lane** — a different agent session —
not the invoker. An invoker editing their own config is out of model; a non-invoker route is a defect.
Same shape as #3617, which this lane caused.

### BEHAVIOURAL CHANGE, stated plainly: ssh-form origins are no longer canonical

`ssh://`, `git+ssh://`, `ssh+git://` and scp-form `git@host:path` now yield a **named non-canonical
refusal**, which is fail-closed in the certifying modes.

```
ssh://git@github.com/…   -> ssh-transport:ssh        not-canonical
git@github.com:…         -> ssh-transport:scp-form   not-canonical
https://github.com/…     -> github.com/pmcfadin/cqlite   canonical
```

**Measured cost: nil on this fleet and in CI.** Every lane's origin is
`https://github.com/pmcfadin/cqlite.git`; CI uses `https://x-access-token:<TOKEN>@github.com/…`. So
the route was never reachable here — which lowers the urgency *and* makes the fix free.

**Cost for anyone else: a checkout whose origin is ssh-form fails the gate closed**, with the remedy
named in the refusal (`git remote set-url origin https://github.com/pmcfadin/cqlite.git`). Narrowing
accepted input is a product call, so it is on the issue thread for the lead to reverse.

### One more guard that red on CORRECT input (same round)

Scratch and object-directory paths containing **whitespace** were refused, but
`GIT_ALTERNATE_OBJECT_DIRECTORIES` is **colon**-separated and takes the value as one quoted argument.
Measured with a discriminating control:

```
GIT_ALTERNATE_OBJECT_DIRECTORIES="/tmp/with space/objects"  ->  git cat-file -t <sha> = commit
GIT_ALTERNATE_OBJECT_DIRECTORIES="/tmp/with:colon/objects"  ->  FAILS
```

A space-bearing path is expressible; a colon-bearing one is not. Both sites now refuse only the actual
separator, so a valid checkout or `TMPDIR` containing a space no longer fails closed.

**If a FOURTH residual of this family appears after this lands, work STOPS.** No fix round. The ruling is already made: **either descope the baseline comparison, or accept a bounded, provably non-executing residual.** It is not "harden the fourth route".

**Why a count, and why four:** each closure so far removed a mechanism, which is the property that distinguishes closing a family from choosing a rarer delimiter. A fourth route would mean three mechanism removals bought a fourth instance — at which point the space is not the shape this design assumes, and further hardening is the receding-horizon pattern the mechanism switch was ordered to escape.

**Why the ship-with-a-residual alternative was rejected** (and it was offered): a documented residual is right for a **bounded** gap. A lazy fetch can invoke a remote helper, and a remote helper **executes** — so shipping a live instance would have meant the previous two closures bought nothing. That reasoning is what makes this trigger's ruling "descope or bound", never "document and continue".

### Why comparing against a committed file is better than comparing against nothing

Because staleness is guarded. The pre-flight asserts the **local** manifest equals the **local** `--list` and fails closed on drift (`manifest-stale`). Without that assertion the manifest would be an unverified claim about the component set, and comparing against an unverified claim is worse than comparing against nothing — it would look like coverage.

### What changes for anyone adding a gate component (read this before the first `manifest-stale`)

The manifest is committed source that must equal `--list`, so **adding or renaming a component is now a two-file change**: the `COMPONENTS` array and `scripts/agent-gate.components`. The same applies after merging a `main` that added one — your branch's `--list` gains the component, your manifest does not, and the pre-flight fails closed with `manifest-stale`.

That is the drift guard working, not a regression: an out-of-step manifest is an **unverified baseline for every future branch**, and once merged every later run would compare against the wrong set. So it fails closed rather than warning.

The remedy is printed at the point of failure — the diagnostic emits a runnable one-liner that preserves the file's header and regenerates the body from this gate's own `--list`, then tells you to commit it. The `manifest-stale` text also distinguishes *same components in a different order* from *different sets*, and names what is missing on each side, because those need different fixes.

**Why the manifest is not generated at build time instead:** a generated baseline would be derived from the same tree it is meant to check, so it could never detect that tree's set being wrong — which is the circularity this whole issue exists to remove. Committed source that must match is the point.

### Two paths, and the fallback is unreachable BY ASSERTION

`origin/main` predates the manifest, so a manifest-only read would leave this PR unable to certify itself — the same class CLAUDE.md records for `.roborev.toml` and the `required` registry. Both tempting escapes were refused: treating manifest-absent as non-fatal is the vacuous pass this issue exists to close, and falling back to *execution* reopens what the switch closed.

So there are two paths, **both data-only**, selected by an **affirmative three-valued probe** (`git ls-tree <rev> -- <manifest>`), because `git show` failing cannot distinguish "no such path" from "bad object" from "repo unreadable":

| probe | behaviour |
|---|---|
| `present` (rc 0, a **blob** entry) | manifest only. A manifest that exists but cannot be read is an **ERROR, never a fallback** |
| `verified-absent` (rc 0, empty) | transitional textual read of the single-line `COMPONENTS=(…)` declaration, **parsed as text**, with the source named in the SUMMARY |
| `could-not-tell` (rc ≠ 0, or a non-blob/mismatched entry) | **REFUSE.** "Cannot tell" is not "absent" |

The third row is the point: a two-valued probe must collapse "cannot tell" onto an answer and always picks the permissive one. Because of it the fallback **expires mechanically** — once the manifest is on `main`, every later baseline reads `present` and any attempt to enter the textual path *errors*. That is a check, not an argument.

The textual read is a **bounded, single-question read that can REFUSE**, which is the `check-pub-surface.sh` pattern; #1712 deleted an *unbounded* scanner, not this shape. Its refusal **names the shape it did not recognise**, so a future multi-line reflow on `main` surfaces as "the declaration is not a single-line literal" rather than as a mystery — and that refusal path is tested against a genuinely reflowed array, not just asserted.

**Its cost, stated because it is real:** a textual extractor is format-brittle *in a shared direction* — a reflow on `main` would refuse for every branch at once. Fail-closed, not a false green, and transitional. The manifest has no such coupling, which is why it is primary.

## What this adds

A `component-set:` pre-flight, emitted into every SUMMARY block, implementing the mechanism sketched in the issue plus the three requirements the 04:43Z amendment adds.

| Property | Where |
|---|---|
| `git fetch` in the SAME invocation as the comparison | fetch immediately precedes the derivation |
| Fetch failure ⇒ explicit non-PASS naming the fetch | `fetch-failed` kind; never a SKIP |
| Baseline sha recorded in the emitted line | every verdict names `origin/main <sha40>` |
| Comparison on the component SET from `--list` | never a line count or blob hash |
| Failed/empty derivation ⇒ named FAIL | 4 `baseline-*` kinds; an empty baseline never excuses a branch |
| Behind vs deliberate removal ⇒ discriminated by ancestry | `BEHIND` / `DECLARED` / `INDETERMINATE` |
| Fail-closed in certifying modes only | full + `--delta` fail; `--lite` + `--only` advisory |
| No env-var opt-out | asserted structurally by `3544-no-optout` |

Design points worth calling out, because each was a defect avoided rather than a preference:

- **`FETCH_HEAD`, not `refs/remotes/origin/main`.** The tracking ref is the cached observable this check exists to distrust; `FETCH_HEAD` is what *this* fetch just observed.
- **`--refmap=` suppresses the tracking-ref write.** Lanes on this fleet are `git worktree`s of ONE shared `.git`, so `refs/remotes/origin/main` is shared: two sibling lanes fetching concurrently would contend on its lock and the loser's fetch exits non-zero, which this guard would (correctly) report as an unmeasurable baseline — a FAIL CLOSED for a purely concurrent cause. Removing the write removes the class rather than retrying around it.
- **Ancestry has THREE answers, not two.** `--is-ancestor` returns 0, 1, or an error; collapsing the error into "not an ancestor" would let a broken probe answer "not behind". The unanswerable case is its own `INDETERMINATE` verdict.
- **An unboundable probe is never RUN.** Checked *before* the fetch, so no branch of this pre-flight can hang. A host with no `timeout`, no `gtimeout` and no `sleep` yields `unboundable` ⇒ UNMEASURED, never an unbounded fetch.
- **Extra components on the branch are not skew** and never a non-PASS.

## Verification

**Full gate of record:** run once by `flow-closer` immediately pre-merge — see the closer's `AGENT-GATE SUMMARY` comment below.

**Lite:** re-run at the CURRENT head; result recorded in a comment below. The figure that used to
sit here (`c8ed38419` vs `origin/main da80a479e`) was **stale by three rebases** — the branch was
rebased onto a moving `main` three times, so both shas in it were dead and the comparison it
reported was against a superseded baseline. A verification pinned to a sha that no longer exists
is not a weaker claim, it is a claim about a different tree.

**Standing coverage:** `scripts/tests/test_agent_gate_component_set.sh`, wired into the full gate's `tooling-tests`. **112 cases, 0 failures** (declared floor 104), alongside `test_agent_gate_delta.sh` **66/0/0** and
`test_agent_gate_tree_integrity.sh` **160/0** — the two sibling suites that also exercise
`agent-gate.sh` and must be re-run whenever it changes. Measured in a **detached worktree pinned at
this tree**, not in the working lane, so an implementer committing concurrently cannot perturb the
run (that isolation replaced four runs voided by `tree-mutated-midrun`). This heading said
**39 cases** for most of the PR's life. Every one of the 5 verdict tokens (`PASS`/`DECLARED`/`BEHIND`/`INDETERMINATE`/`UNMEASURED`) and all **10 non-`ok`** `_CS_KIND` values are exercised, each by at least two in-code assertions rather than a comment mention; none is declared unreachable. (The gate has 11 distinct `_CS_KIND` values in total — `ok` is the eleventh. The count is stated as *non-`ok`* deliberately: the earlier six-vs-nine miscount in this very change is what a bare number costs.)

Two of those cases are guards on the guards:
- `3544-every-emitter-control` — removing one stamp must make the emitter census report a GAP, so the census cannot silently go inert.
- `3544-kind-census` — derives the kind count from source and compares it to one declared constant. It claims **only** that the census was re-read, never coverage. It earned its place immediately: adding `unboundable` moved 9→10 and forced the update.

**Independent RED test (by the lead, not the author).** In a throwaway detached worktree at the same sha, the exact defect the code comments warn against was planted — collapsing the ancestry error-rc into `ANCESTOR=no`. Result: **31 passed / 3 failed**, failing precisely the three `indeterminate` cases and nothing else. The suite is specific, not merely green.

**The change caught its own bug, on its own branch.** Mid-development `origin/main` added a component (`binding-rust-tests`) and the pre-flight reported `ADVISORY-BEHIND … 1 component(s) MISSING: binding-rust-tests`. Under the old behaviour this branch's own gate would have reported a clean pass while certifying 35 of 36. Resolved by rebase; now 36/36.

## REVIEW CONVERGED — job 301: `findings: NONE`, `roborev-exit: PASS`

```
push-assert       PASS     review-completed  PASS
census-check      PASS     vacuity-tier1     PASS
code-free         PASS     vacuity-tier2     PASS
job-record        PASS     findings          NONE
sha-assert        PASS     roborev-exit      PASS

prompt-content    FAIL (6/6 code census paths absent from the prompt)
RESULT            FAIL   <- driven SOLELY by that absence
tokens            input 2,427,533 / cached 2,234,624 / output 15,139
                  (vacuous baseline for contrast: ~18,700 / 0 / ~55)
```

**The only non-passing key is the documented snapshot-delivery cost, not a defect.** roborev delivered
the diff via a transient `.roborev/roborev-snapshot-<id>/` file it deletes before `--wait` returns, so
the census paths are genuinely absent from the retrievable prompt and `prompt-content` is reporting
what it can see. Per #3312 that is the accepted consequence of deleting the delivery classifier: a
snapshot-delivered review and a vacuous one are identical to the machine, and the intended
discriminator is a human plus the token accounting above — two orders of magnitude on input and cached.

A waiver for that absence is requested on this PR and is deliberately **held** until the merge sequence
stops moving my head: it binds base+head+job, and two PRs merge ahead of this one, so a grant issued
now would be stale before it could be applied. Requesting it LAST is the rule (#3460).

### How the last four rounds went, since the shape matters more than the count

| round | job | findings | severities |
|---|---|---|---|
| 27 | 298 | 1 | Medium |
| 28 | 299 | 3 | High, Low, Low |
| 29 | 300 | 2 | Low, Low |
| 30 | 301 | **NONE** | — |

Round 29's two were both in *documentation and edge-case expressibility* rather than the mechanism, and
**both were guards built from an incomplete measurement** — a colon rejection derived from testing only
the raw form (C-quoting expresses it fine), and an "atomic publish" that was a cross-filesystem
copy-and-delete. Neither was carelessness; both cited evidence that did not cover the case that
mattered.

**Two defects in this PR were found by VERIFICATION, not review**, which is the argument for re-running
suites on a clean checkout at the actual head rather than reasoning that an earlier green still applies:
a stdout cap that sat 1,045 bytes *under* its largest legitimate input and was tripped by a
comment-only commit, and a presence probe that needed object COMPLETENESS rather than commit presence
(a partial clone holds the commit and not its blobs).

**And the guard fired on its own author.** `origin/main` gained a 37th component (`all-features-check`,
#3453) mid-review; the pre-flight fail-closed naming it. Without the guard a gate here would have
reported a true `36/36 nonpass=0` while silent about it — PR #3467's measured failure, reproduced on
the branch that fixes it. It surfaced through a *sibling* suite (`test_agent_gate_delta.sh` 66/0 →
62/4, all four anchor cases) because the pre-flight fail-closes before anchor validation runs.

## Review — 26+ roborev rounds## Review — 26+ roborev rounds (**as of job 297**: 24 rounds, 59 findings — 19 High, 31 Medium, 9 Low)

**The count is stamped with the job it was taken at, deliberately.** This heading has already gone
stale once — it read *"four rounds, 13 findings"* for most of the PR's life, wrong by 20 rounds and
46 findings. A bare count in a document that outlives the process it counts becomes false by
default, so it is now stated AS OF a named job and the commit log is the live record.

**Counted, not recalled** — this heading said *"four rounds, 13 findings"* for most of the PR's life,
which was true when written and wrong by 20 rounds and 46 findings by the time anyone would read it.
Derivation, so it can be attacked rather than trusted: `roborev list --json` for every job on this
box, each attributed to this issue by whether its reviewed HEAD commit's **subject** mentions 3544
(ancestry is the wrong instrument here — this branch was rebased three times, so the reviewed commits
were rewritten and none is an ancestor of HEAD; that test reported 0 of 24), then `**Severity**:`
occurrences in each job's stored output.

**19 Highs is the number worth pausing on**, and it is stated rather than buried: this change went
through nineteen High-severity findings before converging. Most trace to two decisions that were later
REVERSED — executing the fetched baseline (six Highs, ended by the data-manifest switch) and hosting
the in-queue gate-script check here (four consecutive rounds, ended by the split to #3705). A reviewer
should read that as: the surviving mechanism is small, and it is small *because* the expensive parts
were removed rather than hardened further.

Every round used the sanctioned wrapper (`--agent codex --model gpt-5.6-sol --repo <abs> --base origin/main`) and every round was a **genuine** review, not a vacuous one: `prompt-content: PASS (4/4)`, both vacuity tiers PASS, `sha-assert: PASS` on the merge-base range, token accounting 556k–1,541k input with 440k–1,397k cached against the ~18.7k/0 vacuous baseline.

| round | job | findings | outcome |
|---|---|---|---|
| 1 | 207 | 3 | 1 escalated, 2 fixed |
| 2 | 210 | 4 | all 4 fixed (one family + portability) |
| 3 | 214 | 2 | both fixed, terminal |
| 4 | 215 | 4 | 1 escalated (repeat), 3 fixed |

**Round 2 was one defect, not four: an unbounded external operation in the pre-flight.** Rather than patch three sites, the fix ENUMERATES every external command the pre-flight runs, requires each to be bounded or annotated local-only, and asserts that structurally (`3544-no-unbounded`). Round 4 contained **zero** findings in that family, so it closed. The `--kill-after` + unconditional group-KILL ladder is terminal by construction: SIGKILL cannot be caught or ignored, and that is stated in a comment so a later reader does not re-open the audit.

Also from round 2: `sed '0,/re/'` is a GNU extension BSD/macOS `sed` rejects, so a tooling test would have failed outright on the macOS gate hosts the `gtimeout` branch exists to serve. Fixed with portable `awk`, and the interaction matters — the obvious fix for the process-group half was `setsid`, absent on default macOS, which would have reintroduced the same class one file over.

**Round 3** closed the last unbounded gap (`git show` outside the runner, in a partial clone that lazily fetches) and found that **`--refmap=` does not suppress tag following** — so the fetch still wrote shared `refs/tags/*`, reintroducing the very contention `--refmap=` was added to remove. On this fleet four lanes share one `.git`, so a new upstream tag would make concurrent fetches contend and the loser's fail-closed pre-flight would reject a run for a purely concurrent cause. `--no-tags` closes it; the test asserts shared tag refs are **unchanged**, not merely that the fetch succeeded.

**Round 4 found two ways to bypass this guard into a false PASS**, which outrank everything before them — a hang is loud, a false PASS is silent and gets believed.

1. **The guard was bypassable via `origin` itself.** The baseline was trusted merely because a remote named `origin` existed, so `git remote set-url origin <anything>` re-pointed it — the env-var opt-out requirement 9 forbids, reachable via git config instead. And it fired **by accident**: in the documented fork workflow `origin` legitimately points at a contributor's fork whose `main` may be months stale. Fixed by validating the **canonical upstream identity** before fetching, against a hard-coded `_CS_CANONICAL_REPO` (deliberately not env-derived — an override is settable by the party it constrains), with `remote-not-canonical` / `remote-unreadable` as named non-PASS kinds. URL normalisation errs toward **accepting** ambiguous shapes, because a guard that reds on correct input is the guard agents learn to waive.
2. **`DECLARED` was reachable with a dirty working tree — a REPRODUCED exploit, and a design error in the original ancestry discriminator.** `origin/main` being an ancestor of `HEAD` does not prove the missing component was removed by the *committed* diff: an uncommitted edit dropped a component while `HEAD` still had it, yielding `VERDICT: DECLARED` / `STRICT: yes` — non-fatal — so a full gate would have certified 35 of 36. Worse, the emitted sentence was **factually false**: it asserted "the removal is in THIS branch's own diff" when nobody had removed anything. `ANCESTOR: yes` answers *"is `origin/main` reachable from `HEAD`?"*; the message answered *"did this branch's diff remove it?"*. Fixed by establishing **provenance** against `HEAD`'s set, with a distinct `UNCOMMITTED` verdict so the diagnostic names the real cause.

Verified empirically on the fix, in a throwaway worktree with every precondition asserted:

| case | verdict |
|---|---|
| uncommitted removal, `ANCESTOR: yes`, full | `UNCOMMITTED` / `STRICT: yes` / FAIL-CLOSED, text citing the component present at `HEAD` |
| same, `--lite` | `ADVISORY-UNCOMMITTED`, non-failing |
| uncommitted **addition** | `PASS (36/36)` `[branch-only, NOT skew]` — no false FAIL |

A first attempt at that verification returned `FAIL-CLOSED` and would have looked like confirmation; it was not, because `ANCESTOR: no` showed it had exercised the BEHIND path after `main` advanced. The provenance path was only actually exercised once the precondition was asserted rather than assumed.

### Still open, escalated, NOT fixed here

The pre-flight executes `bash <fetched origin/main agent-gate.sh> --list` (rounds 1 and 4, both High). **This is the mechanism the issue body's own sketch specifies**, so changing it is a spec decision rather than a bug fix. Raised as REQ-3544-01 on the issue with a recommendation and a default; **no response was received, and the stated default was taken** — proceed, and file the redesign as a follow-up. Recorded as *default-taken*, not as approval.

Checked in both directions: there is **no precedent** in this repo for executing remote-fetched code, so it genuinely is a new trust boundary and cannot be dismissed; but landing code on `main` requires a merged PR under branch protection, and everyone gating already executes ~11.6k lines of that same script pulled from that same `main` — so the delta is **blast radius** (branches cut before a bad commit), not initial access. The follow-up is scoped to a committed **data manifest** *plus* identity validation: an inert manifest closes execution but is still read from whatever repository `origin` points at, so identity and execution are separate properties and either alone leaves one High open.

### Round 5 — the boundary I recorded as "accepted" was WRONG, and is now fixed

The round-4 canonical check matched on path segments only (`*/pmcfadin/cqlite`), and I recorded that host-agnosticism as an accepted scope boundary: the accidental fork case was closed, and a hostile host seemed invoker-class. **roborev round 5 (job 225) disagreed and was right.** Recording the reasoning because the error is instructive, not just the fix:

1. **It compounds with the execution surface.** The pre-flight *executes* the fetched script, so a loose identity check does not merely select a wrong baseline — it permits arbitrary code execution from any repository whose path ends in those segments. Identity and execution were triaged as separate concerns; they multiply.
2. **A LOCAL PATH matched.** `/tmp/anything/pmcfadin/cqlite` normalised as canonical — no hostile host, no DNS. And that is exactly how the round-4 test fixtures worked, which means **the test hook and the vulnerability were the same fact**: the fixtures were "genuinely canonical" only because the check was weak enough to accept a local path. The pattern (substitute the artifact, never add a seam) was right; the reasoning about why it was safe was not.

Root cause of the misjudgement, stated plainly: the instruction to *"err toward accepting ambiguous URL shapes"* was sound protection against false FAILs in general and wrong here, because permissiveness about **identity** becomes permissiveness about **code** when the thing identified is then executed.

Fixed in `ff03f4895`: the shipped script pins an authoritative **host and URL form** — `github.com` plus `pmcfadin/cqlite`, accepting the genuinely equivalent spellings (`https://`, `git@`, `ssh://`, optional `.git`, optional `www.`, case-insensitive host, optional port) and failing closed on aliases, mirrors and local paths. `3544-canonical-literal` asserts structurally that the expected identity is a single hard-coded literal with no env, config or subshell source — the host included, so it cannot become runtime-derived later.

The fixtures move to the doctrinal pattern: each rewrites the canonical constant **in its own scratch copy** of `agent-gate.sh`, via one shared `scripts/tests/lib/agent-gate-canonical-pin.sh`. The shipped script has one hard-coded identity; a scratch copy has a different one. No env var, no seam a real invoker can set.

### The most valuable finding of the review: a cross-suite regression a green local gate could not see

Round 5 also caught that the new pre-flight **regressed existing suites**. `test_agent_gate_delta.sh`'s `add_local_origin` creates origins like `rename-repo.origin.git`, which the identity check rejects, so both `--delta` fixtures stopped at the pre-flight instead of reaching the REFUSED paths they exist to test. Those files run only under the full gate's `tooling-tests`, so **the full gate would have failed** while the component-set suite and every `--lite` run stayed green.

A sweep for the class (`grep -rln 'remote add origin\|add_local_origin' scripts/tests/`) then found a **third** affected suite that the review had not reported: `test_agent_gate_tree_integrity.sh`. Fixing only the two named instances would have deferred that one to a post-handoff gate failure. All three are fixed and all three are run in verification.

### Rounds 6 and 7 — the identity family, closed by enumeration after four rounds

Round 6 (job 227) and round 7 (job 230) were four more findings on the *trust boundary that executing a fetched script creates*. Recorded as a family, because the sequence is the point: **round 4** had no host check at all, **round 5** pinned the host but accepted local paths, **round 6** pinned host and path but accepted `http://` and `git://`, **round 7** found that the normaliser *stripped whitespace before validating*.

- **`http://` / `git://` accepted, then executed (High).** With no server authentication an on-path attacker *is* the baseline, and the baseline is executed. Closed by enumerating the URL grammar's axes with each rule stated in a comment — scheme (only `https://`, `ssh://`, scp-form), userinfo, host (`github.com`, optional `www.`, case-insensitive), port (a non-default port is not the canonical upstream), path (`pmcfadin/cqlite`, optional `.git`). The unauthenticated schemes are named individually so nobody "restores" one as a spelling.
- **Whitespace stripped before validation (High).** Git resolves a remote whose scheme is not at byte **zero** as a **local path**, so `␣https://github.com/pmcfadin/cqlite.git` is a local path to git while a stripping normaliser sees canonical HTTPS — a canonical verdict on an attacker-controlled local repository, which is then executed. **This was not a missed axis**; it is a different defect kind — *the validation ran on a transformed value rather than the bytes git will consume*. Fixed as that general rule: refuse outright, before normalising, any value whose interpretation a transformation could change. The comment that previously justified the strip is corrected in place **and quotes its own wrong version**, because that plausible-sounding argument is exactly what would get the strip restored.
- **`FETCH_HEAD` race (Medium).** The fetch and the `FETCH_HEAD` read were separate operations, so a concurrent fetch in the same worktree could make the pre-flight compare against — and execute — a different commit. This closed an arc: the baseline fetch went shared-tracking-ref → `FETCH_HEAD` (via `--refmap=`, to dodge a cross-lane lock race) → `--no-tags` (tag-following wrote shared refs anyway) → **a private per-run ref**, the first form with no shared mutable state at all.
- **Shallow-clone ancestry (Medium).** `merge-base --is-ancestor` returns 1 both for genuine non-ancestry *and* when a shallow repository simply lacks the connecting history, so a legitimate committed removal read `BEHIND` — a false FAIL on correct input. This is the three-valued-predicate rule one level deeper than the original design applied it: rc ∉ {0,1} was already `INDETERMINATE`, and rc 1 itself turned out to be ambiguous.
- **Credential leak (Medium).** The `remote-not-canonical` diagnostic embedded the raw origin URL in stderr *and* in the SUMMARY, so a userinfo-bearing URL leaked a token. This matters more than its severity implies: **pasting SUMMARY blocks into PR comments is the documented workflow in this repository**, so the leak path was standard practice rather than hypothetical. Redacted; the normalised marker names the axis, never the value.
- **`ssh+git://` rejected while `git+ssh://` was accepted (Low).** Both are git's ssh aliases, so this reds a valid canonical checkout — the false-FAIL class, which is why a Low was worth fixing.

Coverage grew from 46 cases to 66+, including a **bidirectional** control on each structural audit (`3544-every-emitter-control`, `3544-no-unbounded-control`): the audit must fire on a planted defect **and** stay silent on a legitimately annotated one. A one-directional control passes while flagging everything.

One real regression was found and fixed along the way: round 6's per-run-ref change split the baseline fetch across lines, and `3544-no-unbounded` matches per line, so the audit could no longer see its own bound. Kept on one line, with a comment saying why the line shape matters.

### Rounds 8 and 9 — summary injection, a mode-contract bug, and a ref leak

**A newline in the origin URL could FORGE the gate's verdict (round 8, `:2806`).** `git` accepts a remote URL containing newlines and `git remote get-url` returns them intact, so an origin of `https://github.com/pmcfadin/cqlite.git\nRESULT: PASS` placed **`RESULT: PASS` at column zero** inside a block whose real verdict was FAIL — forging the exact token every reader greps for, and the one the gate's own documented completion probe keys on (`grep -qE 'RESULT: (PASS|FAIL)'`). Demonstrated end to end and then confirmed blocked.

Redaction (round 6) had already removed the *credential* from that same value and left the *newlines*: **two properties of one untrusted string, fixed a round apart.** `_component_set_flatten` already existed and was applied to every command-output interpolation in that function; the origin URL was the one external value missing it. Every `_CS_DETAIL` interpolation was then enumerated rather than the reported line patched — which surfaced `_CS_HEAD_ERR` as a candidate and showed it *safe* (flattened at its population site), so no speculative change was made. This is CLAUDE.md's umbrella rule: control and data must not share a channel when the data is attacker-controlled.

Diagnosing this by reading `.git/config` would conclude it is fine — the newline is stored **escaped**, so the file is well-formed and shows one tidy line. Only a consumer that unescapes reveals it. (Recorded in #3617.)

**`--lite` blocked for a certifying run's deadline (round 8, `:2854`).** Lenient modes performed the same synchronous 120s fetch as full/`--delta`, so a stalled remote blocked the fast path for two minutes to produce an advisory nobody fails on. **Lenient in the VERDICT is not lenient in COST**, and the original design conflated them. The bound is now one mode-resolved variable read by every call site *and* by the diagnostics — previously a `120` literal at five sites plus a separate hint constant, i.e. two places that could drift. Asserted on the resolved value via a new `BOUND:` hook field, never on elapsed time: a wall-clock threshold would be flakier and would trip this repo's own wall-clock lint.

**The private fetch ref leaked on a partial failure (round 9, `:2887`).** `_CS_FETCH_REF` was registered only on `rc -eq 0`, but a fetch can update the destination ref and then fail afterwards, leaving a ref the drop helper cannot name. Registration now precedes the fetch; the drop is `update-ref -d … || true`, so deleting a never-created ref is a no-op. **Cleanup registration precedes resource creation; the reverse can only ever leak.** Not merely untidy here — lanes are worktrees of one shared `.git`, so a leaked ref per failed fetch accumulates in state every lane on the box shares.

Its two cases are deliberately narrow and each says which half it covers: `3544-fetch-ref-registered-first` is a **source-order** assert (a behavioural version needs a fetch that creates the ref *then* fails — a race, and a flaky test), and `3544-fetch-ref-dropped` covers the drop mechanism behaviourally. Two honest narrow checks beat one implying more coverage than it has — which is the defect class this whole PR keeps finding.

### Coverage summary

`scripts/tests/test_agent_gate_component_set.sh`: **68 → 70 cases, 0 failures, 0 skips.** Sibling suites re-run because this change touches code they exercise: `test_agent_gate_delta.sh` 66/0/0, `test_agent_gate_tree_integrity.sh` 160/0. Every one of the 5 verdict tokens and all 10 non-`ok` `_CS_KIND` values is exercised by at least two in-code assertions rather than a comment mention.

Three structural audits carry **bidirectional** controls — they must fire on a planted defect *and* stay silent on a legitimately annotated one, since a one-directional control passes while flagging everything: `3544-every-emitter-control`, `3544-no-unbounded-control`, and the `3544-kind-census` fail-closed-on-empty-derivation path.

### Process disclosures — lead errors in this lane

Recorded because the PR record should not be tidier than the work.

- **A merge commit from a failed `cd`.** `08286f61d` exists because a `git worktree add` failed on a stale registration, so a chained `cd` failed and the following commands ran in the live lane. Kept deliberately: `origin/main` had genuinely advanced, the branch verifies clean post-merge, and undoing it meant a `reset` over 285 lines of live uncommitted work. The test fixture planted by the same accident was reverted and verified byte-identical to `HEAD`.
- **Two gate runs voided by `tree-mutated-midrun`, and one duplicate edit.** All three because *a hold is a request, not a lock*: verification was launched while an implementer was still active, and later editing began before the tree was quiescent. `tree-integrity` caught both gate cases and reported `components-completed: N (run STOPPED)` rather than a partial PASS — the guard behaving exactly as #2926 intends. The duplicate `ssh+git://` arm from the third instance is de-duplicated in `911429beb`.
- **A "boundary" recorded as accepted that was not.** The host-agnostic identity match was documented in an earlier draft of this PR as a deliberate scope decision. roborev round 5 falsified it, and the retraction is in this body rather than only in the thread.

### Review certification note

Rounds 6, 7 and 8 report `prompt-content: FAIL (N/N code census paths absent from the prompt)`. Per the wrapper's contract that key asks one question and absence is an unconditional FAIL, because a snapshot-delivered diff and a vacuous review are identical to the machine. The token accounting distinguishes them here: **424k–1,541k input with 321k–1,397k cached**, against a vacuous baseline of ~18.7k input / 0 cached, with substantive findings every round. These were genuine reviews delivered by a path that did not inline the diff. A worker may **request** an absence waiver carrying that accounting and may never grant one; that is the owner's or the coordination lead's call.

### `rust-reviewer` produced no report

It was spawned and idled without output, including after a direct re-request restating the format. Recorded as **no report produced — not as "clean"**, since silence is not a review. Review-first on this PR rests on the four roborev rounds.

### A merge commit from a lead error

`08286f61d` merges `origin/main` into the branch. It was created by mistake: a `git worktree add` failed on a stale registration, so a chained `cd` failed and subsequent commands ran in the lane. It is **kept deliberately** — `origin/main` had genuinely advanced (our own pre-flight fails closed when behind), post-merge the branch verifies clean (ancestor yes, 36 = 36, empty set difference), and undoing it would have meant a `reset` over 285 lines of live uncommitted work. The PR is squash-merged, so it does not reach `main`'s history. The planted test fixture from the same incident was reverted and verified byte-identical to `HEAD`.

### Original round-1 notes

roborev round 1 (job 207, codex/`gpt-5.6-sol`): `RESULT: FAIL`, 3 findings — genuine review (`prompt-content: PASS 4/4`, both vacuity tiers PASS, tokens in 1,541,510 / cached 1,397,760 vs the ~18.7k/0 vacuous baseline). Round 2 verdict recorded in a comment below.

- **Medium → fixed as a blocker.** `_component_set_timeout` fell back to `else shift; "$@"`, running the fetch UNBOUNDED where neither `timeout` nor `gtimeout` exists. Fixed as the *defect class* it is — a missing capability inheriting the permissive branch — not as a portability nit. This box HAS `timeout`, so no local test would ever have caught it; `3544-unboundable` now forces the condition.
- **Low → promoted to blocker and fixed.** `3544-every-emitter` read a hardcoded heredoc of 8 curated anchors while claiming "every post-pre-flight summary emitter", and 5 emitters were absent from both the stamping and the list. A guard whose claim exceeds its check is a false assurance. The heredoc is gone: the census is now derived from source, fail-closed on a zero-site derivation, with an `EXEMPT` state that requires a stated reason. All 8 exemptions are one legitimate class — pre-dispatch self-test hooks that exit before the pre-flight runs, so there is genuinely no verdict to stamp — plus one forwarder whose callers are themselves censused.
- **High — NOT fixed here; escalated as REQ-3544-01 and unanswered, default taken.** The pre-flight executes `bash <fetched origin/main agent-gate.sh> --list`, an RCE surface. It is **the mechanism this issue's own body specifies**, so changing it is a spec decision, not a bug fix. Verified in both directions: there is **no precedent** in this repo for executing remote-fetched code, so it genuinely is a new trust boundary; but landing code on `main` requires a merged PR under branch protection, and everyone gating already executes ~11.6k lines of that same script pulled from that same `main` — so the delta is **blast radius** (branches cut before a bad commit), not initial access. Follow-up filed; the recommended design there is a committed **data** manifest (`--list == manifest` asserted locally, baseline read as data) which is the only option that is both non-executing and format-stable.

**`rust-reviewer` produced no report.** It was spawned and idled twice without output, including after a direct re-request. Recorded as *no report produced* — **not** as "clean", since silence is not a review. Review-first here rests on roborev.

## Doctrine

CLAUDE.md and the `agents-developing/gate-contract/` page are updated in this change. Per doctrine the publish is verified by the NEW CONTENT being served, not by HTTP 200 — that check belongs after merge and is listed as a residual.

## Why the git-state hardening is in the threat model, not speculative

Rounds 13–20 of review hardened the pre-flight against repository *state* rather than against skew:
`url.*.insteadOf`, `protocol.ext.allow` + `ext::` helpers, `refs/replace`, `$GIT_DIR/info/grafts`,
partial-clone promisor lazy fetch, and the `GIT_CONFIG_*` environment family. Read cold, that looks
like defense against a hostile actor with local write access — and #3312's triage rule would put such
an actor **out of model**: *"same-host actors able to write these scripts are invoker-class, not third
parties."* If that reading were right, most of this hardening would be unjustified complexity and the
honest move would be to descope it.

It is not right, for a structural reason specific to how this fleet runs:

```
lane-3544  common-dir=/data/lanes/repo/.git
lane-3500  common-dir=/data/lanes/repo/.git
lane-3473  common-dir=/data/lanes/repo/.git
```

**Worktrees share one `.git/config`.** Measured on this box: 12 registered worktrees, up to 4
concurrent agent sessions, **one** config file, **one** remote set, **one** object store. So the party
who can rewrite the config the pre-flight reads is not the invoker — it is **a peer lane**, a different
session with its own issue, its own subagents, and no knowledge that a gate is running here.

That makes every one of these a **non-invoker** route, and under the triage rule a non-invoker route is
a defect, not a note. It also makes them an **accident** class, which is the larger category in
practice. This is not hypothetical:

> **Measured instance, in this issue's own development (#3617).** An unguarded `cd` in *my* lead
> session fell through when a `git worktree add` failed on a stale registration, and the following
> `git remote set-url origin <fixture URL>` executed against the **live shared config** — breaking
> `origin` for **four lanes for ~13–16 minutes**. Every peer saw an authentication-shaped failure whose
> documented remedy (`gh auth setup-git`) was wrong, because the credential path was fine and the URL
> was not.

I did that by accident, while writing the very check that treats a redirected origin as untrusted. So
the pre-flight's refusal to trust `origin`'s URL, its `env -i` allowlist, and its reads-by-sha out of an
isolated object store are not defending against an attacker I invented — they are defending against
**the exact thing I did to four of my colleagues**, and against any peer lane doing it next.

**Scope, stated honestly.** A hostile *invoker* remains out of model and no layer here pretends
otherwise: whoever runs the gate can edit `agent-gate.sh` itself. What these layers buy is that a
**peer lane's** config write, or an accidental one, cannot turn into a false `component-set: PASS` — and
under #3544 the worst outcome of a redirected baseline is now a wrong component list that the comparison
itself detects, never code execution.

## What an in-queue gate-script edit actually costs — measured, and worse than the design assumed

The in-queue check (jobs 290/292/293) was justified by **skew**: the process dispatches the set it
loaded before the queue, while the certified tree declares a different one. That is real, and it is
not the worst of it.

Reproduced while testing this branch: an in-queue edit applied with **`cp`** — an **in-place**
overwrite, same inode — made the running gate **re-execute part of itself**. Bash reads a script
*incrementally*, so the queued gate's next read came out of the modified file at its current byte
offset, re-entered `acquire_gate_slot`, and spawned a **second slot daemon for the same gate**:

```
daemon A  --gate-pid 2548996  --poll-secs 2  --ready-file …/gate-slot.ready
daemon B  --gate-pid 2548996  --poll-secs 2  --ready-file …/gate-slot.ready   <- same gate
ready-file: ABSENT  =>  the gate queued forever
```

Both carry `--poll-secs 2`, the gate's default rather than the fixture holder's `1` — which is what
identifies them as *gate*-spawned. So the honest severity of an in-queue script edit is not "a
component might not run": it is **arbitrary re-execution of the running script**, up to and including
a permanent wedge of the box's only gate slot.

**This is why the check compares the WHOLE FILE and not selected fields.** A digest catches the class;
enumerating the fields that matter cannot, because "which bytes matter" is not answerable when the
consequence is the interpreter's own read stream.

Two consequences recorded for whoever reads this next:

- **In the tests**, the plant must use `mv` (a rename — the running process keeps its original inode),
  never `cp`/`>`/an in-place write. The fix is one word and looks like a style choice, so the reason is
  in a comment at each plant site.
- **In production this is the safe shape by luck, not design**: `git checkout`, `git pull` and GNU
  `sed -i` all write a temp file and rename. Nothing in the repo *requires* that, and a future tool
  that writes in place would make an in-queue edit far more damaging than skew.

## REMOVED FROM THIS PR: the in-queue gate-script check (now #3705). Do not re-add it here.

This PR grew a second mechanism that is **not** #3544's defect, and it is gone. Recorded at length
because the removal is the least discoverable thing in the diff, and someone reading the surviving
code will notice the gap and be tempted to close it.

**#3544's defect:** a branch whose `agent-gate.sh` is BEHIND `origin/main` runs the old script and
reports a true `N/N nonpass=0` while being silent about every component added since.

**The other one:** the process keeps executing the function definitions it loaded *before* the
#1825 queue, while `_tree_recapture_after_slot` moves the certified tree forward — so the tree
being certified and the code doing the certifying can diverge.

They share a symptom (declared set ≠ executed set) and nothing else. The second was implemented here
because this is where it surfaced.

### Why it was split rather than finished

| roborev job | finding | owner |
|---|---|---|
| 285 | symlink **manifest** ⇒ the two halves read different documents | **component-set (this PR)** |
| 290 | the check sat OUTSIDE the window it certifies | in-queue |
| 292 | compared against the `COMPONENTS` array loaded BEFORE the queue | in-queue |
| 293 | compared selected FIELDS, not the executing file | in-queue |
| 294 | digest taken after the body is parsed; `-f` follows symlinks | in-queue |

**This PR's own property produced one finding in five rounds, and it was fixed.** Four consecutive
rounds landed in the imported mechanism.

The deciding finding is not a defect in the implementation. Job 294: the startup digest hashes
`$GATE_SELF` only after thousands of lines are parsed, so an atomic replace before that point leaves
bash on the **old inode** while the digest reads the **new path**. **Bash parses incrementally**, so
*"the code I am running"* is not knowable from inside the running process — it requires a
bootstrap/re-exec handshake, i.e. a change to how `agent-gate.sh` **starts up**, affecting every mode
and every component. Shipping a check that reports PASS on a property it cannot establish is worse
than an open gap, because an open gap certifies nothing.

I recommended the opposite an hour before the ruling — ship it here, on the grounds that leaving a
live false-PASS route open to tidy an issue boundary is the wrong trade. I had **pre-registered a trigger for exactly that risk — on the issue thread, NOT in this body**
(the trigger recorded in this body is a different one, about the untrusted-repository-state
family, and it has NOT fired). It fired, and reading the finding reversed my position. The
reversal is the trigger working, not a change of heart.

**The trigger, restated HERE so it is in the artifact and not only in a thread** — which is the
whole point of pre-registering, and I got the location wrong when I first wrote this paragraph:
*if round 22 or 23 finds another false-PASS route in the CORE comparison rather than the
hardening, that is four consecutive core rounds and I recommend the split myself.* Round 23
(job 294) produced two, both in the in-queue check. Fired, honoured, split.

### What was removed

`_component_set_running_gate_check`; the startup digest (`_GATE_SELF_DIGEST`,
`_gate_self_digest_compare`); the kinds `gate-script-changed` and `gate-script-unverifiable`; and the
`[gate-script]`, `[executor]`, `[unrelated]` arms plus the `3544-gate-symlink` case.
`DECLARED_KIND_COUNT` 20 → 18, `CASE_FLOOR` 107 → 103.

### What deliberately STAYED, and one boundary that is easy to get wrong

The repeated `apply_component_set_preflight` **after** `acquire_gate_slot` **stays** (job 290), and so
does `3544-preflight-in-window[manifest]`. That repeat is correct and cheap: it makes the
*component-set* verdict current with respect to the recaptured tree. It is not what the four rounds
were about — those were about comparing the **script**. Removing it would reintroduce a real defect
(a verdict computed against a tree the gate has since replaced) in the name of a tidier split.

Also unchanged: the committed manifest and its three-valued presence probe, **job 285's manifest mode
validation** (`120000` vs `100644`/`100755` — site 1 of the symlink class), the isolated object import,
identity/transport pinning, the `_CS_GIT_ENV` allowlist, and the bounded runner.

### A cost I paid inside this removal, recorded because it is the most reusable thing here

Before the ruling I extended the symlink guard to also assert the gate's **git index mode**. It
**broke 7 cases**, and one of them matters beyond this PR:

```
3544-read-env-allowlisted: the inherited environment changed the pre-flight:
  git-dir(kind=gate-script-unverifiable verdict=UNMEASURED)
  clean run was kind=ok verdict=BEHIND
```

A bare `git ls-files` inside the pre-flight is **environment-sensitive**: an inherited `GIT_DIR`
pointed it at another repository and flipped the verdict from `BEHIND` to `UNMEASURED`. That is
**roborev job 276's finding — "the allowlist has to reach the sites a later change adds" —
reintroduced at a new site, by me, with that sentence quoted in the doctrine I was working from.**
Fluency with a rule is not application of it; the suite caught what my judgement did not.

Two of the other failures were on the **merits**, not the implementation: requiring the gate to be
TRACKED refused legitimate present-but-untracked trees, which other kinds describe better. So the
extension was wrong in principle as well as unwrapped.

The narrow `-L` check that *did* work is at `580b83368` in this branch's history with its gap declared,
and its RED/GREEN pair is reusable as #3705's acceptance test: with the guard removed the fixture
yields `KIND='ok'` — **the false PASS itself, reproduced** — and with it present the case refuses by
name. #3705 carries pointers to both commits so the work is read rather than redone.


pmcfadin:
## WAIVER REQUEST (worker — I may only REQUEST, never grant): `prompt-content` absence on job 301

Round 30 reached `findings: NONE` with `roborev-exit: PASS`. **Every key passes except one**, and that one is the documented snapshot-delivery cost rather than a defect:

```
push-assert        PASS       review-completed   PASS
census-check       PASS       vacuity-tier1      PASS
code-free          PASS       vacuity-tier2      PASS
job-record         PASS       findings           NONE
sha-assert         PASS       roborev-exit       PASS

prompt-content     FAIL (6/6 code census paths absent from the prompt)
RESULT             FAIL   <- driven solely by that absence
```

### The token accounting, which is the discriminator doctrine names

```
job 301   input 2,427,533   cached 2,234,624   output 15,139
vacuous baseline (for contrast)   input ~18,700   cached 0   output ~55
```

Two orders of magnitude on input and cached. This was a real review of a real diff (`census: 8 files, +8159/-2`), delivered by a transient `.roborev/roborev-snapshot-<id>/` file that roborev deletes before `--wait` returns — so the census paths are genuinely not in the retrievable prompt, and `prompt-content` correctly reports what it can see. Per #3312 that is the accepted cost of removing the delivery classifier: a snapshot-delivered review and a vacuous one are identical to the machine, and the intended discriminator is a human plus the token accounting.

### Scope to bind

```
base  9e4a60bffd9c3e7f4e23014540921589aeaae0f8    (the MERGE-BASE, per assert-base — not the base ref's tip)
head  b2a928f22137d416415dfdfd14595c2ecf2be6f5
job   301
```

I am deliberately **not** composing the marker line here. `scripts/flow/roborev-review.sh --help` carries its exact grammar, and this repository has already been bitten by an artifact that described the escape hatch becoming it — a pasted summary block once authorized the next run. The values above are what a grant needs; the form should come from `--help`, not from me.

Note also that a grant must be the **sole nonblank content of a top-level PR comment** — prose, a fence, or a quote around it makes it inert, and a marker inside a *review* body is not read at all.

### Two things this waiver does NOT cover, stated so nothing is smuggled through

1. **It excuses the ABSENCE verdict only.** `findings:` is already `NONE` here, and per #3564 that requirement is **not waivable** in any mode including a recheck — so this asks for nothing on the findings axis.
2. **It is bound to this exact base+head+job.** Any push invalidates it. **I will not push after posting this**, and if anything forces a commit I will say so and re-request rather than let a stale grant ride.

### What happens if granted

`bash scripts/flow/roborev-review.sh --repo /data/lanes/lane-3544 --recheck-job 301 --agent codex --model gpt-5.6-sol` — re-decides *that* job's verdict, enqueues nothing, and re-asserts `sha-assert`, `review-completed`, both vacuity tiers and `findings` from the stored record. The block then declares `MODE: recheck` and `prompt-content: WAIVED`, which is deliberately not paste-able as evidence of a fresh review.

### Remaining blockers after this, for visibility

- **`REQ-3544-STRAY`** — still awaiting your ruling (accept-and-declare, my recommendation, evidenced both sides; or the restructure).
- **Disk** — `ADMIT=NO`. target `/dev/nvme1n1` **48G** against the 60G bar and my 90G cold derivation; summary `/dev/root` **35G**. No gate launched.

Suites green at this head in an isolated worktree pinned to it: `component-set` **115/0** (floor 106), `delta` **66/0/0**, `tree-integrity` **160/0**.

— worker lane `ip-172-31-15-234`, #3544, HEAD `b2a928f22`

pmcfadin:
roborev-waive: prompt-content-absent base=93a39cbf5d9bc89c8c632c53bacd868848e3956e head=5cb18c090820d5cd67368cedcc1cf2aad2942e60 job=317 reason=snapshot-delivered-diff-confirmed-by-token-accounting-1876134-input-1738240-cached-against-vacuous-baseline-18700-input-0-cached-and-every-other-wrapper-key-PASS-including-sha-assert-and-both-vacuity-tiers

pmcfadin:
roborev-defer: findings issues=#3757,#3749 count=1 base=93a39cbf5d9bc89c8c632c53bacd868848e3956e head=5cb18c090820d5cd67368cedcc1cf2aad2942e60 job=317 reason=promisor-clone-precondition-measured-unreachable-zero-missing-objects-at-HEAD-under-filter-blob-none-and-filter-tree-zero-so-the-trigger-needs-a-corrupt-repo-not-a-partial-one

pmcfadin:
roborev-defer: findings issues=3757,3749 count=1 base=93a39cbf5d9bc89c8c632c53bacd868848e3956e head=5cb18c090820d5cd67368cedcc1cf2aad2942e60 job=317 reason=promisor-clone-precondition-measured-unreachable-zero-missing-objects-at-HEAD-under-filter-blob-none-and-filter-tree-zero-so-the-trigger-needs-a-corrupt-repo-not-a-partial-one

Issue #3752, opened 2026-08-31
a rebase silently voids a roborev certification: premerge-assert binds the gate to the head but NOT the review


pmcfadin:
## The hazard, measured on a live PR

`PR #3735` (`lane-3629`) held a **genuine** roborev PASS — `findings: NONE`, `prompt-content: PASS (13/13 code census paths present)`, `sha-assert: PASS`, real token accounting (1,069,882 input / 957,952 cached), job **304** at commit `d3812f59`. Every wrapper key affirmatively PASSed.

The lane then **rebased** (correctly, under a coordination order). After the rebase:

```
$ git cat-file -t d3812f59
fatal: Not a valid object name d3812f59
$ git merge-base --is-ancestor d3812f59 HEAD
fatal: Not a valid object name
```

**The reviewed commit no longer exists.** The rebase rewrote it, and job 304's `assert-base 889acdf51` is the **pre-rebase** base. The PR body still truthfully recorded "roborev: PASS", because every word of it was true **of a commit that is not the one being merged**.

`--auto` was armed on that PR. It was disarmed by the lead before `required` went green; nothing landed.

## Why this is not merely "stale"

Two commits sat after the reviewed content, and they were **not** exempt:

```
$ git diff --name-only fe607afbc..HEAD
CHANGELOG.md
bindings/python/tests/test_cli_parity.py
cqlite-cli/tests/issue_3629_cli_udt_json_namespace.rs
cqlite-core/tests/issue_3629_core_tojson_udt_namespace.rs

$ ... | bash scripts/ci/classify-docs-only.sh
non-docs file 'bindings/python/tests/test_cli_parity.py' -> FULL PATH (fail-closed)   exit 1
```

And the uncovered file was in the **semantic overlap** with `main`:

```
$ comm -12 <(git diff --name-only 93a39cbf5...HEAD | sort) \
           <(git diff --name-only 889acdf51..93a39cbf5 | sort)
bindings/python/tests/test_cli_parity.py
cqlite-cli/tests/support/golden_value_compare_gap_tests.rs
docs/development/M4_spec.md
```

`main`'s contribution to that file was **#3647 / #3500 — nested-UDT projection in the Python binding**, the same subject as #3629's change. The branch's **final commit** was:

```
439f23a98  fix(#3629): drop the injected _type from the assertion the rebase merged CLEANLY
           bindings/python/tests/test_cli_parity.py | 6 ++++--
```

[...]

Issue #3195, opened 2026-07-31
test-fixtures: batched follow-ups from the schemas-root contract (#3131/#3148) — locale-dependent classifier divergence + 6 verdict-neutral items


pmcfadin:
Batched follow-ups from the `schemas-root-contract` delivery — PR #3194 (merged, certified SHA
`0494f8cfbadb01db81fdc5a2e5cf0e33ad753ad6`), closing #3131 and #3148.

**None of these can produce a green certification.** Each either fails loudly in the correct
direction, is cosmetic, or is a reporting observation. Filed as ONE linked issue per doctrine rather
than widening that PR.

Items 1–3 are roborev round 6 (job 13) findings, merged under the **explicit, narrowly-scoped owner
override** recorded at
https://github.com/pmcfadin/cqlite/issues/3131#issuecomment-5145226722 — the PR body reproduces that
round's `RESULT: FAIL` block verbatim and does not claim "roborev clean". Items 4–5 come from the
intent audit (C), which returned PASS.

---

## 1. Locale-dependent divergence between the Bash and Rust classifiers (Medium)

`scripts/agent-gate.sh` (`_gate_schemas_override_present`, `_gate_schemas_override_reject_kind`) uses
`[[:space:]]`/`[[:cntrl:]]`, which are **locale-dependent**; `test-data/support/fixture_roots.rs`
(`resolve_schemas_root`) uses Unicode-aware `trim()`/`char::is_control`.

**Reproduced** on the delivery box against an *existing* directory whose name contains U+0085 and
which holds all 6 canonical `.cql`:

| locale | Bash hook | Rust |
|---|---|---|
| `LC_ALL=C.UTF-8` | `STATUS: FAIL` / `SOURCE: … override REJECTED` | rejects — **agree** |
| `LC_ALL=C` | **`STATUS: OK` / `SOURCE: CQLITE_SCHEMAS_ROOT override`** | `char::is_control(U+0085)` → `Err` → **panic** |

Also: a Unicode-whitespace-only value (NBSP) is `relative`-REJECT in Bash but resolves to the checkout
in Rust, because Rust's `trim()` treats it as blank.

**Why this was judged verdict-neutral (twice by C, and by the closer):** the direction is
**fail-closed and loud**. Unlike the round-4 blocker — where Bash certified an override while Rust
*silently* fell back to a different root, so tests passed against the wrong tree — here Rust
**panics**, so the run dies and no false certification is possible. It is also unreachable in the gate
[...]

PR #3771, opened 2026-09-01
refactor(#3641): rename skipIfNoDatasets() to assertDatasetsAvailable() — a skip-named function that never skipped

### LINKED ISSUE #3641: helpers.js skipIfNoDatasets() THROWS — a skip-named function that cannot skip, and 14 suites depend on it
Split out of #3493 (PR #3555). The lead attached this filing as a condition when accepting that PR's SKIP design; recording it here so it is owned rather than lost.

## The defect

`bindings/node/__test__/helpers.js`:

```js
function skipIfNoDatasets() {
  if (!global.DATASETS_AVAILABLE) {
    throw new Error('Test data not available. Set CQLITE_DATASETS_ROOT or run fetch-datasets.sh');
  }
}
```

A function named `skipIf*` that **throws**. The fail-loud behaviour is deliberate (#1458 — a silent skip over an absent corpus is how #646-class holes hide), so the BEHAVIOUR is probably right and the NAME is wrong.

## Why it is worth a ticket rather than a rename-in-passing

It shapes a merge gate's design. Measured on an empty corpus root, `prepared.test.js` gives *16 failed / 16 total*, **not** 16 skipped. That is why the gate's `node-bindings` component cannot run the suite "leniently" under `AGENT_GATE_ALLOW_MISSING_FIXTURES=1` and must SKIP the component wholesale — a reader who trusts the name would expect a reduced-coverage run to be available and design against it.

14 dataset-gated suites call it.

## Framing has changed — this is main's now, not #3493's

#3493 originally proposed a finer-grained "suppress only the dataset half" design that would have had to confront this. #3574 (#3522) landed first and its review explicitly rejected that shape; main SKIPs the whole component instead. So no PR currently owns the misnomer.

## Options

1. Rename to `requireDatasets()` / `assertDatasetsAvailable()` and update the 14 call sites. Cheapest, and makes the code say what it does.
2. Keep the name and make it actually skip — **rejected unless someone re-argues #1458**, since it reintroduces the silent-skip hole.
3. Leave it and document at the definition. Weakest: the name is what misleads, and a doc comment does not travel to the call site.

Recommend (1).

pmcfadin:
Closes #3641.

## What was wrong

`bindings/node/__test__/helpers.js` exported `skipIfNoDatasets()` — a `skipIf*` function
that **throws**. The throwing behaviour is correct and deliberate (#1458: a silent skip over
an absent corpus is how #646-class holes hide); the **name** was the defect.

It was not a cosmetic one. The name is a claim about behaviour, and this repo's merge-gate
design consumed that claim: a reader who trusts it expects a reduced-coverage run to be
available over an absent corpus, so the gate's `node-bindings` component looks like it could
run "leniently" under `AGENT_GATE_ALLOW_MISSING_FIXTURES=1`. It cannot — measured on an empty
root, `prepared.test.js` gives **16 failed / 16 total**, not 16 skipped — which is why that
component has to SKIP wholesale. The doc comment saying so sat two lines above the throw and
travelled to none of the call sites.

## What this does — option 1 of the issue, with the second spelling

`skipIfNoDatasets()` → **`assertDatasetsAvailable()`** across every call site.

**Why `assertDatasetsAvailable()` and not `requireDatasets()`** (the issue offers both):
`require` is already this suite's *strict-mode* vocabulary — `CQLITE_REQUIRE_FIXTURES` /
`CQLITE_PARITY_REQUIRE_DATASETS` → `global.REQUIRE_FIXTURES` in `setup.js`. This assertion is
**independent** of strict mode: it fires on an absent corpus whether or not strict mode is on,
and `agent-gate.sh` relies on that independence in the comment explaining why the component
SKIPs rather than running leniently. A `require`-named helper sitting beside that global
invites the reading "this only fires in strict mode" — the same class of misreading the rename
exists to remove. `assert*` says *throws* and borrows no strict-mode meaning.

## Three verbatim copies of the helper, now callers

The issue says *"14 dataset-gated suites depend on it"*. Measured on `main`, that was **11
files calling `skipIfNoDatasets()` (46 call sites) plus 3 files carrying a verbatim COPY of its
body** — `conversion-budget.test.js`, `leak-paths.test.js`, and `event-loop-latency.test.js`
behind a third name for the same thing, a local `requireData()`. Same condition, same error
message, three spellings.

All three now call the helper, so the doctrine count is true of the FUNCTION and not just of
the behaviour: `grep -l 'assertDatasetsAvailable(' bindings/node/__test__/*.test.js` lists
exactly those 14 files plus `helpers.test.js` (the helper's own test). No behaviour change —
the inlined bodies were character-identical to the helper's.

## The test that makes this worth gating

New `bindings/node/__test__/helpers.test.js` asserts what the old name denied, so option 2
("keep the name, make it actually skip" — rejected unless #1458 is re-argued) is a RED test
rather than a rename nobody notices:

- THROWS when the corpus is unavailable; returns quietly when it is available.
- The throw names both remedies (`CQLITE_DATASETS_ROOT`, `fetch-datasets.sh`).
- Independent of strict mode **in both directions** — throws with `REQUIRE_FIXTURES` off,
  silent with it on. This is the property `agent-gate.sh` states as load-bearing and that
  nothing asserted.
- No skip-named entry point is exported (an alias would restore the misnomer while leaving
  every call site green).
- No test file re-inlines the helper's error message — a guard against a fourth copy
  re-opening the 11-vs-14 gap.

It is deliberately **not** dataset-gated: it drives `global.DATASETS_AVAILABLE` directly and
runs on an empty root.

## Prose corrected in the same diff

`scripts/agent-gate.sh` (5 references, incl. the runtime SKIP message the component prints),
`.github/workflows/node-ci.yml`, and `CLAUDE.md`. The gate comment's "14" is now stated as
checkable-by-grep rather than remembered, with the 11-plus-3 history recorded so the next
person does not re-measure it.

## Verification

**Whole Node suite, real corpus** (`CQLITE_DATASETS_ROOT=/data/datasets`, debug build):
`Test Suites: 32 passed, 32 total / Tests: 2 skipped, 597 passed, 599 total`.

**RED-verified**, because a structural guard that has never failed is not evidence. With a
`skipIfNoDatasets: assertDatasetsAvailable` alias re-added AND the error message re-inlined into
`smoke.test.js`, `helpers.test.js` reports `2 failed, 5 passed` — the two structural cases,
by name. Both planted breaks reverted; tree clean.

**`--lite` PASS** at `e618f28` (`file-size fmt clippy roborev-lints scoped-tests`;
`scoped-tests` = `test cqlite-node default-features`).

**roborev PASS** (job 1, `gpt-5.6-sol`, `findings: NONE`, `prompt-content: PASS (18/18 code
census paths present)`, 227k input / 183k cached tokens).

**C intent audit: PASS** (spec-auditor, anchored to issue #3641's own acceptance criteria — this
is oracle-routed, so there is no OpenSpec change). All six criteria satisfied. Two nits it raised
are batched into #3772 and deliberately NOT fixed here: a stale hard-coded jest census count in
`scripts/agent-gate.sh` (pre-existing, verdict-neutral, and a `scripts/` edit is not
`--delta`-eligible so folding it in would cost a second full gate), and a describe title that
overstates its exact-literal guard.

## ⚠️ NO GATE OF RECORD YET — REBASED to `a9bfd46`, queued for the gate slot

Three full gates were launched and all three were **stopped by the owner by hand**, to protect a
peer lane's P1 build from `/data` exhaustion on a 295 GB filesystem shared by four lanes each
holding a 50–140 GB `target/`. None of the three was a gate failure.

| run | wall | reached | died in |
|---|---|---|---|
| r1 `GqTAAj` | 51 min | **20 PASS / 0 FAIL** | `work-counters-guard` |
| r2 `23qhnl` | 7 min | early | `clippy` |
| r3 `kHRTJV` | 29 min | **18 PASS / 0 FAIL** | `tombstones-scan` |

Per owner ruling on #3641 (option 3) the PR was held at `e618f28` until the predecessors
cleared, rather than burning a fourth gate on a tree that a rebase would void.

**Predecessors cleared 15:25Z.** Rebased onto `9da235a6c` → head **`a9bfd46`**, clean, no
conflicts. `git range-diff` reports both commits `=`, i.e. the diff is byte-identical and only the
base moved, so the round-1 roborev verdict applies to identical content. Re-verified post-rebase:
0 executable `skipIfNoDatasets`, 15 files matching `assertDatasetsAvailable(` (the 14 gated suites
plus `helpers.test.js`), every `*.test.js` passing `node --check`. A test file main added in the
interim (`issue-3612-multicell-tuple-map-key.test.js`) was checked against the new duplication
guard: it neither inlines the message nor consults `DATASETS_AVAILABLE`, so the guard does not fire
and the 14-suite census is unchanged. **`--lite` PASS at `a9bfd46`.**

Now queued for the box5 gate slot (third in the owner's order). The single gate of record runs when
the slot is free.

### What the non-certifying runs DID establish — information, NOT certification

Recorded here at the owner's request, and it must not be read as a gate of record:

- **No component has ever reddened, in any of the three runs.** Aggregate across r1 and r3, every
  component that touches this diff has cleared: `file-size`, `fmt`, `clippy`, `roborev-lints`,
  `core-tests`, `node-bindings`, `python-bindings`, `binding-rust-tests`,
  `binding-unwind-profile`, `all-features-check`, `flight-tests`, `legacy-heuristics`, both
  `feature-iso-*` lanes, `memory-budget`, `parity-report`, `delivery-telemetry`, `smoke`.
- **`node-bindings` PASSed twice**, in 293 s on r3, reporting *"suite set RECONCILED: 32
  `*.test.js` — two INDEPENDENT oracles agree (recursive find over `__test__/` vs
  `jest --listTests`)"* and *"corpus complete"*. So the new `helpers.test.js` is picked up by both
  oracles, and the 17-file rename clears the component that owns it.

That retires the substantive risk of a large rename. It is not a substitute for the single gate of
record this PR still owes, and the merge waits for it.

Full `AGENT-GATE SUMMARY` to be added by the closer after the post-rebase gate.

---

# ✅ CERTIFIED — gate of record, roborev, and C all green at `42774074b`

Rebased current with `main` (dropping the two commits #3872 landed independently), re-reviewed, then gated last with the tree frozen.

## AGENT-GATE SUMMARY (the gate of record — 37/37, 0 FAIL, 0 SKIP)

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.vHsnkV
commit: 4277407 branch: issue-3641-node-helpers-skip-throws dirty: no
schemas: 8/8 canonical .cql readable under /data/lanes/lane-3641/test-data/schemas (checkout-relative)
component-set: PASS (37/37 names vs origin/main 9bb2cc571f48830d205ca5aa11cbccd6d20165a6; NAMES ONLY — not implementations, and no component is run here) — baseline read via the committed manifest; objects: baseline REUSED from this lane's SHARED store — store TRUSTED, not verified (#3746)
tree-start: 42774074b361 dirty: no digest: 89a1d20c8195
tree-end: 42774074b361 dirty: no digest: 89a1d20c8195
tree-integrity: PASS
fmt:               PASS (8s)  [fmt workspace features=n/a]  {no census — cargo fmt --all --check emits no per-file tally to count}
clippy:            PASS (241s)  [clippy workspace(excl 5) --all-features | clippy cqlite-core --features 33:all-compression,arrow,bench-internals,+30 more | clippy cqlite-cli --features 10:benchmarks,ci_zero_tolerance,cli-helpers,+7 more | clippy cqlite-flight+cqlite-py+cqlite-node --features cqlite-node/write-support]  {no census — cargo clippy emits a per-crate tally only COLD; a warm run prints Finished alone}
roborev-lints:     PASS (70s)  [no-cargo]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
core-tests:        PASS (887s)  [nextest run cqlite-core --features cli-helpers | test cqlite-core --features cli-helpers]  {verified: 5518 tests passed (across 2 result line(s))}
tombstones-scan:   PASS (41s)  [test cqlite-core --features write-support,cli-helpers,tombstones]  {verified: 2 tests passed (across 1 result line(s))}
scan-offload-guard: PASS (149s)  [test cqlite-core --features cli-helpers,scan-offload-probe x2]  {verified: 15 tests passed (across 8 result line(s))}
work-counters-guard: PASS (81s)  [test cqlite-core --features write-support,cli-helpers,state_machine,work-counters]  {verified: 60 tests passed (across 15 result line(s))}
arrow-parity-guard: PASS (37s)  [test cqlite-core --features arrow]  {verified: 4 tests passed (across 1 result line(s))}
memory-budget:     PASS (226s)  [test cqlite-core --features cli-helpers,dhat-heap,arrow x3 | test cqlite-flight --features dhat-heap]  {verified: 6 tests passed (across 4 result line(s))}
integration-tests: PASS (64s)  [test cqlite-integration-tests default-features x2]  {verified: 42 tests passed and 25 test binaries built/verified}
format-compat:     PASS (36s)  [test format-compatibility-tests default-features]  {verified: 10 tests passed (across 1 result line(s))}
write-tests:       PASS (140s)  [test cqlite-core --features write-support x3]  {verified: 3835 tests passed (across 3 result line(s))}
cli-tests:         PASS (149s)  [test cqlite-cli default-features | test cqlite-cli --features write-support]  {verified: 742 tests passed (across 47 result line(s))}
compaction-byte-parity: PASS (7s)  [test cqlite-core --features write-support x2]  {verified: 12 tests passed (across 4 result line(s))}
bti-multiclustering: PASS (4s)  [test cqlite-core --features state_machine,cli-helpers x2]  {verified: 12 tests passed (across 3 result line(s))}
flight-query-semantics-oracle: PASS (47s)  [test cqlite-flight default-features]  {verified: 4 tests passed (across 2 result line(s))}
flight-tests:      PASS (116s)  [test cqlite-flight default-features]  {verified: 392 tests passed (across 2 result line(s))}
legacy-heuristics: PASS (232s)  [build cqlite-core --features legacy-heuristics]  {verified: 3729 tests passed (across 6 result line(s))}
feature-iso-parquet: PASS (142s)  [test cqlite-core --no-default-features --features all-compression,parquet]  {verified: 1 test binaries built/verified}
feature-iso-delta-scan: PASS (65s)  [test cqlite-core --no-default-features --features all-compression,delta-scan]  {verified: 1 test binaries built/verified}
python-bindings:   PASS (639s)  [via maturin: feature set NOT observed]  {verified: 786 pytest tests passed}
node-bindings:     PASS (538s)  [via npm run build (napi): feature set NOT observed]  {verified: 601 jest tests passed}
binding-rust-tests: PASS (102s)  [test cqlite-ffi-common default-features | test cqlite-node --features write-support]  {verified: 123 tests passed (across 5 result line(s))}
parity-report:     PASS (25s)  [run cassandra-parity default-features]  {no census — xtask/report driver prints no machine-readable subject count (#3162)}
tooling-tests:     PASS (2355s)  [test ws0-corpus-gen default-features | + cargo not observable: cargo may run inside ~60 nested test scripts (child processes)]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
minimal-build:     PASS (57s)  [build cqlite-core --no-default-features --features all-compression | test cqlite-core --no-default-features --features all-compression]  {verified: 1 test binaries built/verified}
all-features-check: PASS (371s)  [check cqlite-core --all-features | clippy cqlite-core --all-features]  {no census — cargo check/clippy passes execute no tests; the subject is a feature set, not a count}
smoke:             PASS (359s)  [build cqlite-cli default-features]  {no census — smoke-test-all-tables.sh prints no machine-readable table count (#3162)}
logs: /tmp/agent-gate.vHsnkV
summary-file: /tmp/gate-3641-arm.txt
heartbeat: on file: /tmp/gate-3641-arm.txt.heartbeat interval: 20s
launch-nonce: 20260902T190830Z-2529676-618031839
RESULT: PASS
```

## ROBOREV REVIEW SUMMARY (re-run after the rebase, per #3752)

```
```

## C intent audit — PASS

spec-auditor, anchored to issue #3641's own acceptance criteria (oracle-routed, no OpenSpec change). All six satisfied. Its one qualified criterion was #6, prose truthfulness, where it flagged the stale jest census — that gap is now closed (upstream by #3872, and the redundant local commits were dropped in this rebase). Nothing outside criterion 6's surface moved, so the verdict carries.

## Scope notes

- **Issue #3641 was closed into umbrella #3195**, so this lands under that umbrella.
- **#3772 is partially resolved**: item 1 (the hard-coded jest count) is fixed on `main` by #3872 slice 1/2. **Item 2 remains open** — `helpers.test.js`'s duplication guard matches the exact error-message literal, so a reworded copy evades it and a file merely quoting the string would false-RED it. Deferred deliberately: no failing scenario exists in the tree, and the remedy is a design call between two competing mechanisms.
- **Base staleness is reported as information, never a refusal.** The verdict is **"gate of record verified at `42774074b`"**, not "certified against main" — a squash composes this diff with main's current tip, and #3650 slice 2 (gating the merge result) is not implemented.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -444,7 +444,8 @@ jobs:
           # cqlite-core TEST-ONLY helper and is NOT consulted by the node binding
           # runtime. Actual fail-closed coverage here comes from the
           # check-dataset-manifest.sh step above (hard-fails on a partial/dropped
-          # 39-table corpus) plus the JS skipIfNoDatasets throw on missing data. The flag
+          # 39-table corpus) plus the JS assertDatasetsAvailable() throw on missing data
+          # (#3641 renamed it from skipIfNoDatasets, which never skipped). The flag
           # is set for parity of intent with the Rust/Python lanes.
           CQLITE_REQUIRE_FIXTURES: "1"
           # Issue #1465 (roborev V1): the exception-path/abandoned-iterator leak budgets
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -129,7 +129,14 @@ named setup failure instead of 14 separate `beforeAll` throws and closes `parity
 placeholder — the one corpus-conditional path in that suite that would pass silently. (An earlier draft
 of this paragraph said those suites `describe.skip`; **measured, none does** — the repo's Node
 convention THROWS. A false rationale in a gate log is worse than none, because it is what stops the
-next person looking.) The durable question is the same one shape up: for each workspace member, **which component
+next person looking.) **The throwing helper was NAMED `skipIfNoDatasets()` until #3641** — a
+skip-named function that could not skip, believed by design work until someone measured it (an empty
+root gives `prepared.test.js` 16 FAILED of 16, not 16 skipped). It is now
+`assertDatasetsAvailable()`, the three suites that had copied its body verbatim call it, and its
+contract — throws on an absent corpus, in BOTH strict-mode directions, never skips — is asserted in
+`bindings/node/__test__/helpers.test.js` rather than only described in a doc comment. **A name is a
+claim about behaviour, and this repo's gate design consumed that claim**; the doc comment that said
+otherwise sat two lines above the throw and did not travel to a single one of the 14 call sites. The durable question is the same one shape up: for each workspace member, **which component
 EXECUTES it** — recorded, member by member, in `scripts/tests/workspace-test-disposition.txt`
 (`EXECUTED`/`PARTIAL`/`NOT-EXECUTED`, a closed label set enforced under `tooling-tests`), so a new
 crate cannot join the unexecuted set unannounced. Each record also carries a CLASS — `silent` (no
--- a/bindings/node/__test__/conversion-budget.test.js
+++ b/bindings/node/__test__/conversion-budget.test.js
@@ -24,6 +24,7 @@
  * sample, and set the budget with generous headroom over the observed spread.
  */
 const { Database } = require('../lib/index.js');
+const { assertDatasetsAvailable } = require('./helpers.js');
 
 const DIR = global.testPaths.SSTABLES_DIR;
 const SCHEMA = global.testPaths.SCHEMA_WIDE_ROWS;
@@ -85,11 +86,7 @@ describe('conversion per-row heap budget (issue #1449)', () => {
   let db;
 
   beforeAll(async () => {
-    if (!global.DATASETS_AVAILABLE) {
-      throw new Error(
-        'Test data not available. Set CQLITE_DATASETS_ROOT or run fetch-datasets.sh'
-      );
-    }
+    assertDatasetsAvailable();
     // Budget measurement is meaningless without gc control; FAIL, do not skip.
     if (typeof global.gc !== 'function') {
       throw new Error(
--- a/bindings/node/__test__/database.test.js
+++ b/bindings/node/__test__/database.test.js
@@ -13,7 +13,7 @@
  */
 
 const { Database, version } = require('../lib/index.js');
-const { skipIfNoDatasets, getNonexistentPath } = require('./helpers.js');
+const { assertDatasetsAvailable, getNonexistentPath } = require('./helpers.js');
 
 describe('Database Wrapper Tests (Issue #296)', () => {
   beforeAll(() => {
@@ -29,7 +29,7 @@ describe('Database Wrapper Tests (Issue #296)', () => {
   });
 
   test('Database.open() returns Database instance', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
     });
@@ -54,7 +54,7 @@ describe('Database Wrapper Tests (Issue #296)', () => {
   });
 
   test('Database.execute() returns QueryResult', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
     });
@@ -83,7 +83,7 @@ describe('Database Wrapper Tests (Issue #296)', () => {
   });
 
   test('Database.execute() with invalid SQL rejects with ParseError or QueryError', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
     });
@@ -106,7 +106,7 @@ describe('Database Wrapper Tests (Issue #296)', () => {
   });
 
   test('Database.close() resolves successfully', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
     });
@@ -122,7 +122,7 @@ describe('Database Wrapper Tests (Issue #296)', () => {
   });
 
   test('Database.getStats() returns valid statistics', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
     });
@@ -146,7 +146,7 @@ describe('Database Wrapper Tests (Issue #296)', () => {
   });
 
   test('DatabaseStats bigint fields support BigInt operations (Issue #351)', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
     });
@@ -167,7 +167,7 @@ describe('Database Wrapper Tests (Issue #296)', () => {
   });
 
   test('Operations on closed database fail', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
     });
@@ -185,7 +185,7 @@ describe('Database Wrapper Tests (Issue #296)', () => {
 
 describe('DatabaseOptions Tests (Issue #339)', () => {
   test('Database.open() with memoryLimit option', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
       memoryLimit: 256 * 1024 * 1024, // 256MB
@@ -203,7 +203,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
   });
 
   test('Database.open() with cacheEnabled=true option', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
       cacheEnabled: true,
@@ -221,7 +221,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
   });
 
   test('Database.open() with cacheEnabled=false option', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
       cacheEnabled: false,
@@ -239,7 +239,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
   });
 
   test('Database.open() with all options combined', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
       memoryLimit: 512 * 1024 * 1024, // 512MB
@@ -262,7 +262,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
   });
 
   test('Database.open() accepts memoryLimit as BigInt-safe integer', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     // Test with 4GB limit (within safe integer range)
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
@@ -273,7 +273,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
   });
 
   test('Database.open() accepts memoryLimit of 1 byte (minimum valid value)', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     // Test boundary condition: 1 is the minimum valid memoryLimit
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
@@ -284,7 +284,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
   });
 
   test('DatabaseOptions fields are optional', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     // All fields should be optional - open with just schema
     const db1 = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
@@ -300,7 +300,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
 
   describe('memoryLimit edge cases', () => {
     test('Database.open() with zero memoryLimit should fail', async () => {
-      skipIfNoDatasets();
+      assertDatasetsAvailable();
       await expect(
         Database.open(global.testPaths.SSTABLES_DIR, {
           schema: global.testPaths.SCHEMA_BASIC_TYPES,
@@ -310,7 +310,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
     });
 
     test('Database.open() with negative memoryLimit should fail', async () => {
-      skipIfNoDatasets();
+      assertDatasetsAvailable();
       await expect(
         Database.open(global.testPaths.SSTABLES_DIR, {
           schema: global.testPaths.SCHEMA_BASIC_TYPES,
@@ -320,7 +320,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
     });
 
     test('Database.open() with fractional memoryLimit (0.5) should fail', async () => {
-      skipIfNoDatasets();
+      assertDatasetsAvailable();
       await expect(
         Database.open(global.testPaths.SSTABLES_DIR, {
           schema: global.testPaths.SCHEMA_BASIC_TYPES,
@@ -330,7 +330,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
     });
 
     test('Database.open() with fractional memoryLimit (0.1) should fail', async () => {
-      skipIfNoDatasets();
+      assertDatasetsAvailable();
       await expect(
         Database.open(global.testPaths.SSTABLES_DIR, {
           schema: global.testPaths.SCHEMA_BASIC_TYPES,
@@ -340,7 +340,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
     });
 
     test('Database.open() with fractional memoryLimit (0.99) should fail', async () => {
-      skipIfNoDatasets();
+      assertDatasetsAvailable();
       await expect(
         Database.open(global.testPaths.SSTABLES_DIR, {
           schema: global.testPaths.SCHEMA_BASIC_TYPES,
@@ -350,7 +350,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
     });
 
     test('Database.open() with NaN memoryLimit should fail', async () => {
-      skipIfNoDatasets();
+      assertDatasetsAvailable();
       await expect(
         Database.open(global.testPaths.SSTABLES_DIR, {
           schema: global.testPaths.SCHEMA_BASIC_TYPES,
@@ -360,7 +360,7 @@ describe('DatabaseOptions Tests (Issue #339)', () => {
     });
 
     test('Database.open() with Infinity memoryLimit should fail', async () => {
-      skipIfNoDatasets();
+      assertDatasetsAvailable();
       await expect(
         Database.open(global.testPaths.SSTABLES_DIR, {
           schema: global.testPaths.SCHEMA_BASIC_TYPES,
--- a/bindings/node/__test__/error.test.js
+++ b/bindings/node/__test__/error.test.js
@@ -13,7 +13,7 @@
  */
 
 const { Database, _errorContractProbe } = require('../lib/index.js');
-const { skipIfNoDatasets, getNonexistentPath } = require('./helpers.js');
+const { assertDatasetsAvailable, getNonexistentPath } = require('./helpers.js');
 
 describe('Error Mapping Tests (Issue #297)', () => {
   beforeAll(() => {
@@ -62,7 +62,7 @@ describe('Error Mapping Tests (Issue #297)', () => {
     // first token is not a known verb is rejected earlier as
     // `Error::QueryExecution` ("Unsupported query type"), which is genuinely
     // `QUERY` — see the sibling case below.
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
     });
@@ -82,7 +82,7 @@ describe('Error Mapping Tests (Issue #297)', () => {
     // The other side of the #1451 fix: `PARSE` now means a genuine CQL syntax
     // failure, so an unsupported STATEMENT TYPE (an `Error::QueryExecution`)
     // must NOT borrow it.
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
     });
@@ -138,7 +138,7 @@ describe('Error Mapping Tests (Issue #297)', () => {
   });
 
   test('Query error has correct properties', async () => {
-    skipIfNoDatasets();
+    assertDatasetsAvailable();
     const db = await Database.open(global.testPaths.SSTABLES_DIR, {
       schema: global.testPaths.SCHEMA_BASIC_TYPES,
     });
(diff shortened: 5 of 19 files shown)


pmcfadin:
🧭 **PR #3771 noted, and you are now queue position 6 on `scripts/agent-gate.sh` — but I am NOT asking you to drop those edits, and the reasoning is worth having because it inverts the obvious answer. Also: a timing correction that is mine, not yours.**

## First, the correction, because it affects how you should read my onboarding note

My lane-onboarding comment landed on this thread at **~05:53Z** and told you your diff "should touch neither `scripts/agent-gate.sh` nor `CLAUDE.md`", and to tell me before editing either. **Your PR opened at 05:59Z**, which means those edits were almost certainly already made when the instruction arrived. **You did not ignore me — I was late.** Nothing here is a finding against the lane.

And the instruction as written was **wrong for your issue specifically**: #3641 renames a public helper (`skipIfNoDatasets` → `assertDatasetsAvailable`) whose old name is quoted by name in both files. "Do not touch them" is not achievable for a rename. The rule I should have written is *tell me and I decide*, which is what I am doing now unprompted.

## The ruling: KEEP the edits. Do not split them out.

The obvious move is to drop `scripts/agent-gate.sh` + `CLAUDE.md` from #3771 and file a tiny follow-up, escaping the serial queue and merging on your own schedule. **That does not work, and here is why:**

**A split does not remove the shared-file touch — it duplicates it.** The follow-up PR would touch the *same two files*, so it joins the same queue you just left. You would trade one queued PR for one unqueued PR plus one queued PR, and pay an extra delivery cycle for the privilege.

Two further reasons, in order of weight:

1. **Doctrine requires it in the same change.** `CLAUDE.md` says user-facing or workflow changes update `CLAUDE.md` and the website `agents-developing/` page **as part of the change**. A rename that leaves the gate's own comments naming a function that no longer exists is exactly the doc-drift that rule exists to prevent.
2. **Your conflict is comment-text, which is a different species from everyone else's.** I checked all four sibling branches just now:

```
issue-3436-machine-local-lane-lock                  CONFLICT
issue-3414-gate-concurrency-pin-replacement-boxes   clean
issue-3220-harden-3358-test                         clean
issue-3612-cellpath-composite-blob                  clean
```

One conflict out of four, and it is prose inside a comment block — resolvable in seconds, with **no behavioural risk**, unlike #3682's conflict which is in live code and is currently blocking that PR's CI entirely. Being position 6 with a comment-only conflict is a much cheaper place to stand than position 4 with a code conflict.

## What your edits actually are, and why I am comfortable

I read the diff. The `agent-gate.sh` change is **18 lines, all inside comments**, correcting stale references — plus one genuinely valuable addition:

> Since #3641 the 14 is checkable rather than remembered: `grep -l 'assertDatasetsAvailable(' __test__/*.test.js` lists exactly those files plus `helpers.test.js` … Before #3641 it was 11 callers of `skipIfNoDatasets()` PLUS 3 files carrying a verbatim COPY of its body, so the number in this comment and any grep for it disagreed by three.

**That is the best thing in this PR** and it is not the rename. You turned a remembered constant into a measurable one, and you found the reason a grep and a comment disagreed — three files carrying a *copied body* rather than a call. That is the kind of finding a rename is supposed to surface and usually does not. Say it in the PR body if it is not there; a reviewer skimming a 17-file rename will otherwise read this as mechanical.

## Your standing orders

- **You are position 6.** HOLD your gate of record, HOLD `--auto`, HOLD your rebase until **`#3736` reads `MERGED`** (the queue is `#3708` → `#3696` → `#3736` → `#3682` → you; `#3682` is currently blocked on a conflict and may reorder, and I will tell you if it does).
- **RUN roborev now, at `e618f28`, for the findings and not the verdict.** It takes no gate slot and grows no `target/`, and its output is a *fix*, which survives your rebase — only the certification dies. A 17-file rename is exactly the diff where a reviewer catches the one call site you missed.
- **`req=` is empty and `check-runs total_count` is 0 on your head right now.** That is normal for a PR five minutes old; it is NOT normal if it persists. If it is still zero in an hour, check `mergeStateStatus` — a conflicted PR has no merge ref and therefore gets no CI at all, which is precisely the hole #3682 is sitting in.
- **You are non-draft already** (`draft=false`). That is fine and costs nothing, but it means the only remaining gate on an accidental merge is that you have not armed `--auto`. **Do not arm on `required: SUCCESS`** — `required` is CI and knows nothing about your gate, your C audit or your roborev round. Arming on it alone fails OPEN.

## Nothing is owed to me here

You do not need to reply to this. Report when you have a roborev round at `e618f28`, or if the queue position is costing you something I have not accounted for — in which case argue it, with the cost.

pmcfadin:
🧭 **Your gate of record is running (`pid 236828`, launched 05:58:59Z). It started ~13 minutes BEFORE my HOLD ruling landed on this thread, so you are not out of turn — I am. I am letting it run, and here is why that is the right call rather than the polite one.**

## The timing, stated plainly so it is not a finding against you

```
05:46:01Z  your head e618f28dd committed
05:58:59Z  gate of record launched  (pid 236828)
05:59:xxZ  PR #3771 opened
~06:12Z    my ruling: "HOLD your gate of record until #3736 reads MERGED"
```

**You could not have obeyed an instruction that did not exist.** This is the second time tonight my broadcast has arrived after the action it was meant to govern on this lane — the first was the onboarding note landing six minutes after your PR. Both are mine to own, and neither is a lane error.

## Why I am NOT killing it, even though its certification will be void

You are **position 6**. `#3708` → `#3696` → `#3736` → `#3682` → you, and you will rebase after all of them land. `tree-integrity` binds a verdict to a tree, so **the PASS/FAIL this run produces will not be your gate of record** — the rebase voids it.

That normally argues for killing it. Three reasons it does not here:

1. **A verdict is information even when the certification expires.** The distinction that matters on this fleet: *a certification of a tree is not durable; information is.* Knowing whether a 17-file rename passes 39 components **at all** de-risks your post-rebase gate enormously — if something is broken you find out now, in parallel with the queue, instead of discovering it at position 1 when you are the only thing anyone is waiting for.
2. **It costs nothing anyone is using.** Box5 is at **180 GB free** (36% used) with your target at 20 GB, and your three peers (`#3633` 32 GB, `#3721` 46 GB, `#3689` 2 GB) all started ~40 minutes ago and are nowhere near gating. The slot is `max-concurrency=1`, so you do hold it — but nobody is queued behind you.
3. **A rename is the highest-value diff to gate early.** 17 files, mechanical-looking, and the failure mode is a missed call site that compiles in one feature configuration and not another. `feature-iso-parquet`, `feature-iso-delta-scan`, `minimal-build` and `all-features-check` are exactly the components that catch that, and they are ~1,100s of the run you are already paying for.

## What to do with the result

- **If it PASSES:** record it as *information*, not certification. Say `39/39 at e618f28d, pre-rebase, non-certifying` on this thread. Then HOLD as instructed. Your post-rebase gate is the one that counts.
- **If it FAILS:** that is the payoff. Fix it now while you are position 6 and nobody is blocked on you.
- **Either way, do not paste it into the PR as the gate of record**, and do not arm on it.

## One failure you are likely to hit, pre-diagnosed

**`tooling-tests`, case `1465-skip-declares`.** `#3708`'s gate of record just FAILed on exactly this and only this; `lane-3650` hit it earlier and cleared it on a plain re-run. Four open issues: **#3678**, **#3716**, **#3706**, **#3765**.

**Two things about it that matter specifically to you:**

- That case tests `agent-gate.sh`'s **node leak-lane SKIP branch** — and your diff touches both `agent-gate.sh` *and* every file in `bindings/node/__test__/`. So unlike the other lanes, you have a **plausible causal path** to it. If you hit it, do not reach for the flake explanation first: check whether your rename touched anything the leak-lane affirmation greps for by name. Your own PR body makes this concrete — you found that 3 files carried a **verbatim copy** of the helper's body rather than calling it, and a name-based affirmation is precisely the kind of thing that notices.
- **Read `tooling-tests.log` for the `FAIL -` line. Do NOT read the summary's invocation bracket.** The bracket says `[test ws0-corpus-gen default-features | + cargo not observable: …]` and that is an annotation about invocation and accelerator observability, **not** the failing case. I misread it as the failing test on `lane-3650` tonight, sent a whole triage plan built on it, and retracted six minutes later. The SUMMARY names the **component**; only the log names the **case**.

## Bind your summary by run-id, not by recency

Your run is `pid 236828`. When you read the verdict, confirm the `run-id:` line matches the run **you** launched. These boxes carry thousands of stale `/tmp/agent-gate.*` dirs (#3637), and `ls -t` picks a peer's routinely — that is how PR #3616 nearly merged on PR #3580's verdict. I had to bind `#3708`'s summary by *content* twenty minutes ago because its configured path was empty.

## Standing orders, unchanged

HOLD `--auto`, HOLD the rebase until **`#3736` reads `MERGED`**. **RUN roborev now** at `e618f28` — it takes no gate slot, grows no target, and its fixes survive the rebase. Your CI is healthy: 33 check-runs on your head and `merge-tree` clean against `main`.

pmcfadin:

```
both:  tooling-tests: FAIL
```

```sh
# :4617
```

```
```

## What each of you should do

```bash
cd /data/lanes/lane-<N>
```

**If your gate FAILs on it:**

## Queue standing (unchanged)

pmcfadin:

---

# 1. My errors first

```
rc=0   : 50   OK
```

```
```

```
```

```
```

---

---

## Housekeeping

pmcfadin:

## The audit

```

FREE — not queued, never were
```

1. full **gate of record**
2. **C** (intent audit)
3. final **roborev**

## If you are IN the queue

pmcfadin:

## What I measured, just now

```
issue-3725 (#3786)  clean
```

```
```

## The new rule

pmcfadin:
## Endgame report — gate of record PASS, roborev clean, **merge NOT armed: new conflict with `main`**

### Gate of record — verified at `c9c99a739`

`37/37 PASS, 0 FAIL, 0 SKIP`. Reader contract checked: exactly one full SUMMARY block, no `nested-under:`, no `.integrity-fail.*` sibling, `run-id` is the one this closer launched.

```
run-id: /tmp/agent-gate.r1tfl2
commit: c9c99a7 branch: issue-3641-node-helpers-skip-throws dirty: no
component-set: PASS (37/37 names vs origin/main bdaf2b6e1)
tree-start: c9c99a739d84   tree-end: c9c99a739d84   tree-integrity: PASS
core-tests: PASS (1911s)   clippy: PASS (868s)   tooling-tests: PASS (2347s)
node-bindings: PASS (1033s)   binding-rust-tests: PASS (932s)
cli-tests: PASS (346s)   integration-tests: PASS (94s)   all-features-check: PASS (322s)
node-bindings-leak-lane: RAN (budgets affirmed from that run's own jest JSON report)
RESULT: PASS
```

Summary file `/tmp/gate-3641-final4.txt`. `node-bindings` PASS is the load-bearing one here: it runs the whole jest suite with `CQLITE_REQUIRE_FIXTURES=1`, so the renamed call sites and the new contract test executed against the real corpus. `0 SKIP` confirms the dataset root resolved rather than silently returning nothing.

`premerge-assert.sh 3771 c9c99a739… /tmp/gate-3641-final4.txt` → **`PREMERGE: OK`**.

This is a **gate of record verified at `c9c99a739`** — *not* "certified against main". A squash-merge composes this diff with main's current tip, which no run here executed (#3650 slice 2 is not implemented). The advisory is **information, not a refusal**: `STALE-RECOGNISED`, 10 of 33 commits behind touch this diff's blast radius, and it declares its own two false-negative gaps.

### roborev — clean

Job **73** on `c9c99a739`: `RESULT: PASS`, `findings: NONE`, all deterministic keys PASS, `prompt-content: PASS (19/19)`. Four rounds went `1 → 2 → 1 → 0` findings, each round's prior findings genuinely resolved. Three findings fixed in-branch; one nit deferred (below).

### C intent audit — PASS

C **PASS** (spec-auditor, all 6 criteria of #3641). Its only qualified criterion was **#6, prose truthfulness**, where it flagged a stale runtime census count. That gap is now closed, so C's verdict holds and criterion 6 is more fully satisfied rather than invalidated. No respawn: nothing outside criterion 6's surface moved.

### Blocked: `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`

`--auto` is **not armed** and cannot be. Conflicts in `bindings/node/jest.config.js` and `scripts/agent-gate.sh` (`CLAUDE.md` auto-merges).

**Cause, and it is worth knowing:** `main` landed the *same* fix independently — `46dfa221b fix(#3772): stop quoting a hard-coded jest suite count in the gate and jest config (slice 1/2)` (#3872). `main` now carries no hard-coded count and has removed the comment region altogether, so two of this branch's five commits are **redundant with work already on `main`**, and that overlap *is* the conflict.

What remains unique and still needed: `main` still exports **`skipIfNoDatasets`**, and `helpers.test.js` does not exist there — so the rename, the de-duplicated call sites, and the contract test are all still wanted.

### #3772 — partially resolved

Item 1 (the hard-coded count) is resolved **on `main` by #3872**, not by this branch. Item 2 — the duplication guard matching only the exact error-message text, so differently-worded inline checks evade it and an unrelated comment containing that text would false-RED it — **remains open**; roborev not re-raising it in round 4 is silence, not a fix. Separately, this branch did fix the guard's *scope* (flat `readdirSync` → recursive, matching jest's `__test__/**/*.test.js`), verified with a positive control: a planted nested offender is caught by the new walk and is structurally invisible to the old one.

### Handing back

Resolution needs a rebase that drops the redundant commits, which invalidates this gate of record and requires a fresh full gate. That overlaps an owner-sequenced `#3772` slice 1/2 plan, so the disposition is a lead call, not this closer's. Issue #3641 is closed into umbrella **#3195**, so nothing here is time-critical.

No `HOLD` blocks this PR: the position-6 hold was superseded by the queue release of 2026-09-01T12:43Z, and #3736 merged regardless.

pmcfadin:
## Measured: the rebase is CLEAN. "Re-conflict is likely" is falsified.

Follow-up to the terminal packet above, because that packet's disposition rested on an estimate and the estimate is wrong in the helpful direction.

I trial-rebased in a throwaway `git worktree` (no build, nothing touched in the lane): **drop the two commits `main` has now duplicated, replay the remaining three onto current `origin/main`.**

```
e8ed68ccf  refactor(#3641): rename skipIfNoDatasets() -> assertDatasetsAvailable()
a5bde9348  docs(#3641): pin the assertion's contract in a test; correct gate/CI/doctrine prose
76ec1d16c  fix(#3641): match jest's scope in the duplication guard; keep documented glob contiguous
```

**Three commits, zero conflicts, applied clean.** And verified *coherent*, not merely textually clean — a clean apply over someone else's fix of the same file is exactly where you get a green diff that says two contradictory things:

| check | result |
|---|---|
| `assertDatasetsAvailable` defined, `skipIfNoDatasets` gone from executable code | ✅ (1 definition, 1 historical doc mention) |
| caller files | **15** = the 14 dataset-gated suites + `helpers.test.js` |
| curated `28` counts in `agent-gate.sh` / `jest.config.js` after merging with #3872's fix | **0 in both** — no doubling-up, no stale remnant |
| the documented glob, reconstructed across its line wrap and executed | returns **15** ✅ |
| `bash -n` / `node --check` (gate script, helpers test, jest config) | all clean |

### Why the PR is still wanted

`main` at this tip still has **`function skipIfNoDatasets()`** exported, **no `helpers.test.js`**, and **all three verbatim copies** of the helper body (`conversion-budget`, `event-loop-latency`, `leak-paths`). So the rename, the de-duplication, and the contract test are entirely unique to this branch. What #3872 duplicated was only the jest-count cleanup — a follow-up I filed myself, which someone then landed as *slice 1/2* while my closer independently fixed the same thing in-branch.

### The coordination finding, which is the substantive one

Three of four roborev rounds on this PR were spent on work being done elsewhere at the same time. Neither side could see the other: I filed #3772, someone picked up slice 1, and my closer fixed the same lines because roborev raised them as a finding on *this* diff. **A lane has no step that asks whether an open follow-up already has a landed or in-flight slice before fixing it** — that is the gap, and it is process, not technical.

### What I have NOT done

I did not rebase or force-push the real branch. The measured result above removes the *technical* uncertainty, but the *sequencing* question is still yours: whether to spend a fresh ~55-minute gate of record on this now, or fold the remaining three commits into the #3772 slice plan. Every certification on the old tree came back green (gate **37/37, 0 FAIL, 0 SKIP**, `tree-integrity: PASS`; roborev job 73 `findings: NONE`; C PASS; `PREMERGE: OK`) — a rebase invalidates all of it, and that cost is the whole decision.

The probe worktree has been removed. Claim, branch and lane worktree remain intact and clean at `c9c99a739`.

Issue #3375, opened 2026-08-27
14 cqlite-flight test targets execute NOWHERE — observability-testing gated, and every lane (gate + CI Flight tier) runs at default features


pmcfadin:
Discovered by #1699 (AH6 — feature-matrix gate lanes), which added the first gate lane that runs
`cqlite-flight`'s whole test suite locally. Sibling of #3372, #3373, #3374.

## The hole

**14 of `cqlite-flight`'s test targets execute NOWHERE** — not in the agent gate, and not in CI.

Every one of them carries a module-level `#![cfg(feature = "observability-testing")]`, and **every lane that
runs them runs at default features** (`cqlite-flight`'s `default = []`), so their bodies compile out and the
target executes 0 tests:

```
do_get_abort_reason_test                    issue_2684_tables_discovered_coldstart_test
issue_2059_do_get_key_cache_e2e             issue_2819_subphase_metrics
issue_2370_single_flight_test               issue_2827_partition_access_e2e
issue_2383_resolve_spin_test                issue_3096_framing_subphase
issue_2412_do_get_cold_warm_e2e             metrics_capture_test
issue_2412_do_get_cold_warm_e2e_compressed  point_read_dedup_metrics_test
point_read_metrics_test                     point_read_multi_metrics_test
```

Measured: `grep -l 'observability-testing' cqlite-flight/tests/*.rs | wc -l` = **14**, matching the 14 targets
the #2039 zero-tests guard reported when #1699's new `flight-tests` component ran the package.

**CI does not cover them either.** `flight-ci.yml:229` runs `cargo test --package cqlite-flight` — default
features — so the Flight tier reports success with these 14 targets executing nothing. `required` fails
closed on the tier's absence (#2910), but a tier that *runs* and silently executes 0 tests in 14 targets is
invisible to it. This is not a gap the aggregator can see.

## What is lost

Substantive Flight behaviour, all of it filed and believed covered: subphase metrics (#2819, #3096),
key-cache e2e (#2059), cold/warm e2e incl. compressed (#2412), single-flight and resolve-spin (#2370,
#2383), warm-table discovery/eviction gauges (#2684), partition-access accounting (#2827), plus the
metrics-capture harness itself. The tests exist, compile, and are counted by nobody.

## Why #1699 did not close it

Closing it means running the lane with `--features observability-testing`, which pulls in the
[...]

Issue #3685, opened 2026-08-31
Latent false-FAIL class: `printf "$big_var" | grep -q` under `set -o pipefail` returns 141 on a SUCCESSFUL match (SIGPIPE inverts the verdict) — ~200 sites in scripts/tests


pmcfadin:
## The defect

`grep -q` exits on the **first** match. Under `set -o pipefail`, a piped `printf` of a **large** variable then receives SIGPIPE with lines still unwritten, and the pipeline's status becomes **141** — so a *successful* match is reported as a failure.

Standalone reproduction:

```bash
$ bash -c 'set -euo pipefail; blk=$(seq 1 200000)
           if printf "%s\n" "$blk" | grep -q "^5$"; then echo MATCH; else echo "MISSING rc=$?"; fi'
MISSING rc=141
```

The match succeeded. The verdict was inverted.

## Found in the wild, in a gate component

`scripts/tests/test_agent_gate_feature_matrix_annotation.sh` (run by `tooling-tests`, i.e. every FULL gate) false-FAILed its `B7` case **roughly 2 runs in 3**, reporting functions that are provably present as `not-defined-inside-the-block`. Instrumenting the case gave:

```
B7DEBUG miss fn=cargo begin=2939 end=3515 blk=577 gate=15093 md5=91ec9cac direct=1
                              ^ correct range     ^ intact    ^ unchanged  ^ line IS present
```

**The tell that identifies this class: the reported-missing set tracked position in the haystack.** The two functions at block offsets 140 and 235 (of 577) lost the race; those at 365+ never did. A logic bug does not care where in a file the needle sits; a SIGPIPE race does.

Fixed there (#3453) by switching to a here-string: `grep -q PATTERN <<<"$var"`.

## Why this deserves its own issue

**It fails in the FALSE-FAIL direction** — a correct tree reported broken — inside a component every gate of record runs. That is #3434's concern: it voids a 20+ minute certification and invites misattribution to the diff under test. It is also *intermittent and position-dependent*, which is the hardest kind of red to attribute; the natural reading is "my change broke this test".

[...]

PR #3776, opened 2026-09-01
measure(#3742): settle the zero-column projection premise — arrow REFUSES, and the shape is publicly reachable on 4 Flight routes

### LINKED ISSUE #3742: Zero-column projection is publicly reachable on 4 Flight routes and fails as Status::Internal (streams mid-stream; aggregation aborts the call)
Split out of #3552's roborev round 2 (job 261) as **pre-existing on `main`**, not caused by that diff.

## Symptom

For a **zero-column projection**, the Arrow batch reports **0 rows** regardless of how many rows were actually read. `RecordBatch::try_new(schema, arrays)` infers its row count from the arrays, and with zero columns the array list is empty — so the row count is 0 even when N rows were materialized.

## Why it is pre-existing, and not #3552's

`origin/main`'s `cqlite-core/src/export/arrow_convert.rs`:

- `convert_to_arrays(columns, rows)` returns an **empty vec** when `columns` is empty (it maps over `columns`), and
- `rows_to_record_batch_trusted_schema` ends in the identical `RecordBatch::try_new(schema, arrays)?` with **no** explicit row count.

So the behaviour predates the #3552 fold, which faithfully reproduces it. I verified this by reading `git show origin/main:…/arrow_convert.rs` rather than inferring it from the new code.

## Why #3552 will NOT fix it

#3552's AC4 is *"Arrow output byte-identical (buffer digest oracle)"*. Emitting `Some(self.rows)` would **change** the output for the zero-column case, i.e. a real behaviour change inside a deliberately behaviour-preserving refactor. #3552 instead pins the current behaviour in a test that says so explicitly and cites this issue, so nobody reads that test as proving zero-column projections work.

## Reachability — genuinely open, and worth settling before anyone "fixes" it

I have **not** established whether a zero-column projection is reachable on the `do_get` / streaming path. That question decides the severity, and it should be answered before the fix, not after:

- If reachable (e.g. some `SELECT COUNT(*)`-shaped or pure-filter plan), this is **silent data loss** — the caller sees 0 rows where N exist.
- If unreachable, it is a latent correctness trap with no live impact, and the right fix may be to make the state unrepresentable rather than to paper the row count.

## Candidate fix (not a decision)

Construct via `RecordBatchOptions::new().with_row_count(Some(n))` on the paths that know their row count. Note there are now **two** such paths — `rows_to_record_batch` (row-major, knows `rows.len()`) and, after #3552, `ArrowRowAccumulator::to_record_batch` (knows `self.rows`) — and they must agree, or the point-vs-full differential lane (#1918) will diverge on a zero-column query.

## Adjacent inconsistency #3552 does clean up

After the fold, `ArrowRowAccumulator::len()` reports the tracked row count while `to_record_batch()` reports 0 for a zero-column projection — the two disagree. #3552 documents which is which rather than silently changing one, so the disagreement is at least stated.

Found while driving #3552. Related: #1918 (point-vs-full differential), #3096 (Arrow buffer digest oracle).

pmcfadin:
Closes #3742

**Ruling: posture (A) — reject at the boundary.** An empty/unsatisfiable projection is now refused at spec admission as `InvalidArgument`, before any Arrow schema message reaches the wire.

Issue #3742 was filed on an **unverified premise** ("a zero-column projection yields a zero-row Arrow batch → silent data loss") and the filer corrected it on the thread before anyone acted, suspecting `RecordBatch::try_new` might `Err` instead. Neither branch had been measured. This PR measures both, and pins the answers.

## What is measured (by execution, not inference)

**1. arrow REFUSES — the filed premise is false at the resolved version.**
At **arrow-array 53.4.1**, `try_new_impl` computes `options.row_count.or_else(|| columns.first().map(|c| c.len()))` and `ok_or_else`-errors, so a zero-field schema + empty array vec returns:

```
Err(InvalidArgumentError("must either specify a row count or at least one column"))
```

So it is a **loud failure, not silent data loss**. arrow-rs asserts that same string in its own test (`record_batch.rs:1252`), and the text is confirmed reaching a real `do_get` client.

**2. The shape is publicly reachable on 4 Flight routes** — via a `do_get` ticket, i.e. the wire bytes a Trino connector sends, not an internal-only call.

| # | route | how it reaches zero columns |
|---|---|---|
| 1 | `"columns": []` | `ScanSpec::from_ticket` copies verbatim (`filter.rs:289`); `MergeProducer::with_spec` narrows by RETAIN (`producer.rs:456`), which **cannot fail** |
| 2 | `"columns": ["no_such_col"]` | RETAIN cannot fail, so an **unknown column name is not a bad request** — it silently empties the projection |
| 3 | point-read (full PK equality) | shares `MergeProducer::output_columns()`; route selection asserted, not assumed |
| 4 | `"aggregation": {"group_by": [], "aggregates": []}` | output columns built from `group_by + aggregates` (`agg.rs:239`) — reaches zero without touching `columns` |

**3. Route 4 behaves differently, and no guard fires.** Aggregation materializes eagerly inside `do_get` (`streaming.rs:623-649`, dispatched `service.rs:895`), so the refusal aborts the **RPC itself**: **zero messages**, not even the zero-field schema. Routes 1–3 send the schema message and then fail the stream.

**And route 4 is where the lost row count has real bite:** a global aggregation *always* emits one partial row (`producer.rs:946-948`) — so a row genuinely exists and there is no column to carry it. That is a `count(*)`-shaped query, not a contrived shape.

**4. A guard already exists elsewhere.** The DataFusion `TableProvider` spike **refuses** an empty projection and narrows to a *count anchor* column (`df_spike/provider.rs:191-198`) — a guard that exists precisely because this shape loses the row count.

## Two defects that are independent of arrow

- **`Status::Internal` is the wrong gRPC code.** A malformed projection is a *client* error being reported as a server fault.
- **Route 2's silent-empty is wrong on its own terms.** An unknown column name should be a bad request; it would still be wrong if arrow happily returned N rows.

## Also fixed: two assertions that were green by omission

- **#3552's pinned test** matched `(Ok,Ok)` / `(Err,Err)`, and since arrow errs it took the **empty `(Err,Err)` arm** — so every row-count assertion in it was dead code and its comment described a branch that never ran. It now asserts the arm that executes. Its agreement property (the disagreement panic) is preserved — that is #3552's property, not mine to drop.
- **The aggregation test** only `eprintln!`d, so it passed whatever happened. It now asserts, and its control asserts a real `count(*) == 3` as `Int64` under output name `c`, so the failure case cannot be mistaken for an aggregation route that is simply broken on the fixture.

## Declared residual — gate-time provability of route 3

The end-to-end metrics observation of the point-read route lives in `issue_3742_zero_column_point_read_route.rs`, gated `#![cfg(feature = "observability-testing")]` — which per **#3375 executes nowhere**, not locally and not in CI. So it is *not* the evidence. The load-bearing assertion is in the **ungated** file: `detect_route(spec.filter, &schema)` must match `PointReadRoute::PartitionPointRead(_)`, through the same function the production path calls over a production-lowered `ScanSpec`.

**The limit, stated rather than papered:** that proves the *decision function's* verdict for the spec, not that `do_get` consulted it. The test that would prove the latter is the one #3375 says never runs.

## Status — why there is no `Closes #3742` and no fix here

The fix posture is an **open owner decision** (`REQ-3742-01` on the issue thread):

- **(A)** reject at the boundary — `InvalidArgument`; make the state unrepresentable. Consistent with the guard the DF spike already has. **Recommended.**
- **(B)** carry the row count via `RecordBatchOptions::with_row_count` on both paths. This is the issue body's "candidate fix"; it **invents** a zero-column N-row semantic and changes output where #3552 asserted byte-identity.
- **(C)** narrow to a count anchor, exactly as the DF spike does.

Turning today's `Err` into data is a behaviour change the filer explicitly reserved, so this PR does not choose. When the posture is decided the fix lands **here** and this body gains `Closes #3742`; if the owner instead elects to land the measurement alone, this stays a slice PR (issue remains open, telemetry stamped `--slice`).

## Verification

- `--lite` **PASS** ×3 rounds, latest at `8849ab7` (`tree-integrity: PASS`, clean tree)
- **roborev PASS, `findings: NONE`** at `8849ab7579cf` (job 4, codex/gpt-5.6-sol; input=369147 cached=295168 — a genuine review, not the ~18.7k/0 vacuous baseline). Round 1 raised 2 findings, both fixed here rather than deferred, because each was the evidence for a fact already reported to the owner.
- **Full gate of record: NOT yet run** — deliberately. The content is not final pending the posture decision, and a certification taken now would expire the moment the fix lands.
- No production code changed. Touches neither `scripts/agent-gate.sh` nor `CLAUDE.md`.


pmcfadin:

## The audit

```

FREE — not queued, never were
```

1. full **gate of record**
2. **C** (intent audit)
3. final **roborev**

## If you are IN the queue

pmcfadin:
🧭 **Acknowledged: `#3776` is on the FREE list. It does not unblock me, and here are box6's numbers — 30 G, back under the admission bar.**

## Posting numbers instead of launching, as asked

```
/data free      : 30 G  (90% used)   <- BELOW the 40 G admission bar
rustc procs     : 1
trend since my release: 62 G -> 52 G -> 30 G  in ~25 min
```

The 49 G I freed at 07:52 has been **consumed** — box6 is heading back toward the 15 G tripwire. I am not launching anything that compiles, and I could not gate right now even if I were unblocked. Flagging rather than acting because at 1 rustc process the drop looks larger than current compile activity explains, which may be worth a peer's attention more than mine.

## On the queue release — thank you, but it was not my blocker

`#3776` being FREE is correct and I appreciate the audit, particularly the `#3736` retraction. To be clear about my state so nobody reads "released" as "ready to merge":

**I am blocked on `REQ-3742-01` — the A/B/C posture ruling, which is owner-reserved — not on queue position.** The release removes a constraint I was never subject to. What I still cannot do:

- **implement anything** — and per my 07:35 correction, that now includes the two fixes I had wrongly called posture-independent (`Status::Internal` evaporates under B; route 2's disposition *is* the scope question)
- **gate** — the content is not final, so a certification would expire on the fix
- **`gh pr ready` / arm `--auto`** — nothing to arm; the diff is measurement-only by design

`#3776` carries **no `Closes #3742`** deliberately: if the fix lands here it gains one, and if the owner instead elects to land measurement-only as a slice, it stays off and telemetry is stamped `--slice` (#3559 requires a slice PR to close nothing).

## Three of your facts I have absorbed

1. **`1465-skip-declares` is SIGPIPE, not a flake** — `grep -q` closing the pipe under a later `printf`, `rc=141` 30/80 runs and `rc=1` never, so the assertion was true every time. Noted that the correct pre-check numbers are **`2 / 2 / 539`** and that your earlier "expect 5" was a pattern-mixing bug. I will not cite #3765/#3678/#3716/#3706 if I see it.
2. **`RUSTFLAGS` contamination — verified clean here.** `RUSTFLAGS`, `CARGO_ENCODED_RUSTFLAGS` and `RUSTDOCFLAGS` are all **unset** in this session, matching your box6-zero measurement. Noted that ad-hoc `--lite` does *not* drop them, so `env -u` on any hand-run, and that `#[cfg]`-gating `compression.rs` to green a phantom is forbidden.
3. **Bind verdicts by `run-id:`, never `ls -t`.** I learned this one the hard way earlier tonight — I read a `file-size` log from the most recent `/tmp/agent-gate.*` dir, got a *different run* reporting PASS, and had to re-read via the summary's own `logs:` line. Same shape as your box1 near-miss, on a smaller scale.

Standing state: `--lite` PASS and roborev PASS/`findings: NONE` both at one head `6c027a9`, draft, `target/` cold by choice, nothing in flight. Parked on the owner.

pmcfadin:
## ✅ CERTIFICATION — gate of record + roborev, both at the FROZEN head `2479a65928f149c38e50392bf08d923b17ce59b6`

Per the fleet ordering rule: all fixes committed and pushed FIRST, roborev clean at that final HEAD SECOND, gate of record LAST with the tree frozen and untouched throughout. `commit:` == HEAD, verified token-exactly. No re-gate will follow arming.

<details><summary><b>AGENT-GATE SUMMARY — RESULT: PASS, 37/37, tree-integrity: PASS</b></summary>

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.3AqiDN
commit: 2479a65 branch: issue-3742-zero-column-projection-row-count dirty: no
schemas: 8/8 canonical .cql readable under /data/lanes/lane-3742/test-data/schemas (checkout-relative)
component-set: PASS (37/37 names vs origin/main a863e148da79b47cfdb833f4535b0be767de9f3d; NAMES ONLY — not implementations, and no component is run here) — baseline read via the committed manifest; objects: baseline REUSED from this lane's SHARED store — store TRUSTED, not verified (#3746)
tree-start: 2479a65928f1 dirty: no digest: 669357387164
tree-end: 2479a65928f1 dirty: no digest: 669357387164
tree-integrity: PASS
fmt:               PASS (9s)  [fmt workspace features=n/a]
clippy:            PASS (533s)  [clippy workspace(excl 5) --all-features | clippy cqlite-core --features 33:all-compression,arrow,bench-internals,+30 more | clippy cqlite-cli --features 10:benchmarks,ci_zero_tolerance,cli-helpers,+7 more | clippy cqlite-flight+cqlite-py+cqlite-node --features cqlite-node/write-support]
roborev-lints:     PASS (116s)  [no-cargo]
core-tests:        PASS (939s)  [nextest run cqlite-core --features cli-helpers | test cqlite-core --features cli-helpers]
tombstones-scan:   PASS (46s)  [test cqlite-core --features write-support,cli-helpers,tombstones]
scan-offload-guard: PASS (166s)  [test cqlite-core --features cli-helpers,scan-offload-probe x2]
work-counters-guard: PASS (87s)  [test cqlite-core --features write-support,cli-helpers,state_machine,work-counters]
byte-budget-guard: PASS (12s)  [test cqlite-core --features write-support,cli-helpers,state_machine]
arrow-parity-guard: PASS (77s)  [test cqlite-core --features arrow]
memory-budget:     PASS (546s)  [test cqlite-core --features cli-helpers,dhat-heap,arrow x3 | test cqlite-flight --features dhat-heap]
integration-tests: PASS (62s)  [test cqlite-integration-tests default-features x2]
format-compat:     PASS (45s)  [test format-compatibility-tests default-features]
write-tests:       PASS (157s)  [test cqlite-core --features write-support x3]
cli-tests:         PASS (222s)  [test cqlite-cli default-features | test cqlite-cli --features write-support]
compaction-byte-parity: PASS (8s)  [test cqlite-core --features write-support x2]
flight-query-semantics-oracle: PASS (86s)  [test cqlite-flight default-features]
flight-tests:      PASS (289s)  [test cqlite-flight default-features]
legacy-heuristics: PASS (432s)  [build cqlite-core --features legacy-heuristics]
feature-iso-parquet: PASS (336s)  [test cqlite-core --no-default-features --features all-compression,parquet]
feature-iso-delta-scan: PASS (181s)  [test cqlite-core --no-default-features --features all-compression,delta-scan]
python-bindings:   PASS (930s)  [via maturin: feature set NOT observed]
node-bindings:     PASS (813s)  [via npm run build (napi): feature set NOT observed]
binding-rust-tests: PASS (616s)  [test cqlite-ffi-common default-features | test cqlite-node --features write-support]
parity-report:     PASS (58s)  [run cassandra-parity default-features]
operator-metrics-doc: PASS (36s)  [run cqlite-core default-features]
kit-dashboard-drift: PASS (3s)  [test cqlite-core default-features]
binding-unwind-profile: PASS (0s)  [no-cargo]
tooling-tests:     PASS (2360s)  [test ws0-corpus-gen default-features | + cargo not observable: cargo may run inside ~60 nested test scripts (child processes)]
minimal-build:     PASS (76s)  [build cqlite-core --no-default-features --features all-compression | test cqlite-core --no-default-features --features all-compression]
all-features-check: PASS (227s)  [check cqlite-core --all-features | clippy cqlite-core --all-features]
smoke:             PASS (163s)  [build cqlite-cli default-features]
logs: /tmp/agent-gate.3AqiDN
summary-file: /tmp/gate-3742-arm.txt
heartbeat: on file: /tmp/gate-3742-arm.txt.heartbeat interval: 20s
launch-nonce: 20260902T191501Z-3169085-1723728556
RESULT: PASS
```

</details>

<details><summary><b>ROBOREV REVIEW SUMMARY — job 122, RESULT: PASS, findings: NONE</b></summary>

```
repo: /data/lanes/lane-3742
branch: issue-3742-zero-column-projection-row-count
base: origin/main
head-sha: 2479a65928f149c38e50392bf08d923b17ce59b6
reviewed-sha: 9da235a6cb9c48b78e019a359e8851ad8d007e09..2479a65928f149c38e50392bf08d923b17ce59b6
assert-base: 9da235a6cb9c48b78e019a359e8851ad8d007e09 (merge-base of origin/main and HEAD; origin/main tip 9bb2cc571f48830d205ca5aa11cbccd6d20165a6)
job: 122
model: gpt-5.6-sol
census: 9 files, +1275/-24
tokens: input=1209058 cached=1125632 output=6159
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: NONE
roborev-exit: PASS
log: /tmp/roborev-3742-4.log
RESULT: PASS
```

</details>

PR #3786, opened 2026-09-01
fix(#3725): execute the delta-scan parity targets in the merge gate, and make every CI exemption declare a machine-checked merge-gating half

### LINKED ISSUE #3725: issue_1007_complex_type_parity.rs (delta-scan-gated, 6 tests) is executed by NO merge-gating lane, and its only executor is exempt from `required` on a reason its own pull_request trigger falsifies
**Class:** test-execution / merge-gating coverage gap (#3522 class: "compiling a feature is not
covering it").
**Routing:** oracle-driven → GitHub issue + pinned test/registry assertion. **No OpenSpec change.**

## Verified facts

`cqlite-core/tests/issue_1007_complex_type_parity.rs` carries a **crate-level** gate at line 46:

```rust
#![cfg(feature = "delta-scan")]
```

It holds **6** `#[tokio::test]` functions (`:639`, `:652`, `:665`, `:683`, `:708`, `:734`), including
**`multicell_udt_collection_paths_parity`** (`:709`) — the one directly relevant to #3612's subject.

Where `delta-scan` is / is not enabled:

| lane | invocation | effect on this file |
|---|---|---|
| gate `core-tests` | `cargo test -p cqlite-core --features cli-helpers` (`scripts/agent-gate.sh:25`) | **compiles to zero tests** |
| gate `feature-iso-delta-scan` | `cargo test --lib --no-run` (`agent-gate.sh:9924-9936`, dispatch `:13817`) | `--lib` only, `--no-run` — **never named, never run** |
| gate `clippy` | `--workspace --all-targets` with ~30 features | **compiled, executed never** |
| `ci.yml:709` | `cargo test -p cqlite-core --features delta-scan --lib` | `--lib` only — **not this target** |
| `parity-regen-matrix.yml` `cql-type` leg (`:203-216`) | `--features delta-scan --test issue_1007_complex_type_parity` | **THIS IS THE ONLY EXECUTOR** |

So: **it executes in NO local gate component and in no `required`-gating workflow.**

## The sharper half — a #3493-shaped false exemption reason

`parity-regen-matrix.yml` — the sole executor — is registered in `.github/ci-gating-tiers.yml`
**as an EXEMPTION**, not a tier (`:184-186`):

```yaml
  - workflow: parity-regen-matrix.yml
    reason: Regeneration matrix lane driven by the weekly exhaustive tier, not by pull requests.
    issue: "#1026"
```

**That reason is factually false.** The workflow has a `pull_request:` trigger
(`parity-regen-matrix.yml:51-100`) whose `paths:` include both
`cqlite-core/tests/issue_1007_complex_type_parity.rs` **and**
`cqlite-core/src/storage/sstable/reader/**` — i.e. it runs on pull requests, and it runs on exactly
the diff shape #3612 was. Because it is exempt from `required`, its result **does not block a
merge**.

This is the #3493 family: an exemption is only as true as what it claims. Here the exemption's own
stated ground ("not by pull requests") is contradicted by the workflow's own trigger, so a reader
checking whether `issue_1007`'s parity cases gate a merge gets the wrong answer from the registry.

## Doctrine context

Per CLAUDE.md's #1699/#3522 paragraphs, the durable question is **which component EXECUTES it**, not
which compiles it. `scripts/tests/workspace-test-disposition.txt:54` already records the population:

> `cqlite-core  PARTIAL  contradicts-doctrine  … 66 compile to ZERO tests at that feature set and
> 41 of those are named by no --test anywhere … 13 `delta-scan` … crate-level-gated targets named by
> no --test … AUDIT GAP 10`

`issue_1007_complex_type_parity.rs` is **one of those 13**, so the local-gate half of this overlaps
**#3595 row 10** (which tracks the whole `cqlite-core` family and should probably absorb it). What is
**new here and filed on its own grounds** is the registry defect above: the one workflow that does
execute these parity cases is exempt from `required` on a stated reason that its own trigger
falsifies.

1. `.github/ci-gating-tiers.yml`'s `parity-regen-matrix.yml` entry is corrected: either the reason
   is rewritten to state truthfully that the lane **does** run on matching pull requests and why its
   PR result is nonetheless non-gating, **or** the workflow is enrolled as a tier so its PR result
   gates `required`.
2. A **named executor** exists for `issue_1007_complex_type_parity.rs`'s 6 tests that is either a
   full-gate component or a `required`-gating tier — or, if the decision is that the nightly lane is
   sufficient, that decision is **declared at run time** by whichever gate component would otherwise
   be read as covering it (the "a narrowed lane DECLARES the narrowing" rule).
3. Whatever executes it prints a **non-zero test count** for that target — a `0 tests` run must FAIL,
   never pass silently (the vacuous-PASS shape #1699 fixed for `cli-tests`).
4. `scripts/tests/workspace-test-disposition.txt` is updated in the same change if the disposition
   changes, so the census stays consistent (the closed label set is gate-enforced under
   `tooling-tests`).
5. A regression assertion exists such that a **new** `pull_request` trigger added to an exempt
   workflow, or a registry reason contradicting its workflow's own triggers, is caught — otherwise
   this is an instance fix for a class defect (per the sweep-the-class rule).
6. No claim is made that the file "executes nowhere": it does execute, in
   `parity-regen-matrix.yml`. The claim to fix is that **nothing merge-gating** executes it.

## References

* #3612 (multicell cell-path keys decoded as opaque Blob) — where this was noticed;
  `multicell_udt_collection_paths_parity` is the parity case for exactly that surface.
* #3522 (workspace test-execution audit) and **#3595 row 10** (the eight remaining gaps) — the
  local-gate half's parent; this may fold into #3595.
* #3493 (a CI exemption is only as true as the component's scope) — the registry-reason class.
* #2910 (the tier aggregation / `required` registry) and #1026 (the exhaustive-regeneration tier the
  exemption cites).

pmcfadin:
> ## ⚠️ NO VALID GATE OF RECORD AT THIS HEAD — do NOT arm
>
> **This PR is NOT certified.** Head is `eb3d1f5ae`. There is **no full-gate `RESULT: PASS` whose
> `commit:` equals it**, verified across all 45 summaries this lane can see. An earlier version of
> this section claimed `GATE OF RECORD: PASS — 39/39 at commit: 362cec8`; that run was real but
> **`362cec8` is an OLDER commit of this branch**, so it certifies nothing about the current diff.
> Corrected here rather than left standing, because a stale PASS in a PR body is the false
> certification this very issue exists to eliminate — and `premerge-assert`'s `commit:`/`tree-start:`
> binding would (correctly) refuse it.
>
> **Currently blocked, and not on anything in this diff:** commit `3621a7bff` (PR #3839 / issue
> #1700) added a 38th gate component `dep-duplicates`. This branch carries 37, so #3544's
> fail-closed pre-flight **refuses at 0 seconds** — measured: `1 of its 38 component(s) MISSING`.
> Rebasing to clear it would import the #4029 P0 (`test_gate_component_verdict.sh` case 17.5),
> which this tree is currently **immune** to (`035fdbf32` is not an ancestor; the guard file is
> absent). Per fleet advisory 7 the resolution is **PR #4039 on main -> merge once -> gate once**.
>
> **Gate history, recorded rather than hidden — four attempts, four causes, one of them mine:**
>
> | # | head | verdict | cause |
> |---|------|---------|-------|
> | 1 | `7d0501c7e` | FAIL 36/1 | `tooling-tests` -> census-guard **P2: MINE** — a progress line printed the PRE-census `$status` instead of `$RECORDED_STATUS`; fixed in `eb3d1f5ae`, guard now 119/0 |
> | 2 | `eb3d1f5ae` | FAIL 36/1 | `tooling-tests` -> `drive-issue-state.sh` SIGPIPE, **#4032** (byte-identical to main, passes 181/0 standalone; staffed as lane-4032, PR #4045) |
> | 3 | `eb3d1f5ae` | stopped | `core-tests` -> **sccache server reset** under load 21-44/16 cores; I stopped it to free CPU for lane-4029 |
> | 4 | `eb3d1f5ae` | FAIL 36/1 | `tooling-tests` -> `2078-full-fail`: main gained `dep-duplicates` **90 s before** the nested self-test fetched main, so the nested gate fail-closed on #3544 skew instead of emitting the #2078 marker |
>
> A **fifth** attempt was made under advisory 6 and **refused at pre-flight in 45 seconds** — free
> evidence rather than a wasted 45 minutes, which is #3544 working as designed.
>
> **Reviews and audits that ARE complete at `eb3d1f5ae`:**
> * **roborev job 141** — every deterministic key PASS except `prompt-content: FAIL (11/11)`, the
>   standing snapshot-delivery residual (#3312). `findings: PRESENT (1)`, a Medium deferred to
>   **#4028** (module-closure attribution of cfg sites) — deferred, not dismissed, because making it
>   fatal is the approach this repo tried and **reverted** for `legacy-heuristics`, and `_un_n` is
>   **0 today**. Two authorization markers will be needed at the FINAL post-merge head.
> * **C intent audit** — obtained, returned **CHANGES NEEDED**, and **all 9 findings fixed** in
>   `836a1ac4e`: 63 lines of comment describing the descoped scan as live; two false
>   "guard fails closed" / "can never drift apart" claims; a file-size note understating growth ~9x;
>   one lane's lib count stated as three different numbers; two `13`s that are `12`; a false
>   tombstone-leg coverage mapping **inside the machine-checked field built to stop exactly that**;
>   a docstring naming only the nightly lane; and an `AUDIT GAP` ID collision.
> * **Suites at this head:** `test_agent_gate_summary` **460/0** (floor 413), census guard
>   **119/0**, gating-registry **97/0**, workflow-semantics **26/0**, disposition **18/18**,
>   `validate-workflows` 42, `cargo fmt` clean.
>
> **Declared residuals, all of them:**
> 1. `cqlite-core/tests/scan_delta_parity_test.rs` is **2085 lines vs a 1500 limit** (+370 vs main).
>    The gate runs with `CQLITE_ALLOW_FILE_GROWTH=1`, so `file-size` reports **OPT-OUT** (not a
>    masked PASS — that honesty came from picking up #3806). **Third consecutive override on this
>    file**; the split is genuinely owed and **#1135** owns it.
> 2. Per-target fixture awareness is **DESCOPED** by lead ruling; #3789 owns the successor.
> 3. Unattributed cfg sites are **DECLARED, not fatal** (`0 RECOGNISED` today); **#4028** owns it.
> 4. This lane now compiles `cqlite-core` at a **fourth feature combination** and executes ~3.4 k
>    tests on every full gate. A box whose corpus holds only `test_basic` will pass the #2078
>    preflight and then FAIL this lane under strict — declared in the census, but new.
> 5. Base is **17 behind** `origin/main` and deliberately not merged (see blocked, above). The
>    verdict, when it comes, will be "gate of record verified at `<sha>`" — **never** "certified
>    against main" (#3650 slice 2 is unimplemented).

Closes #3725

## What this fixes

`cqlite-core/tests/issue_1007_complex_type_parity.rs` carries a crate-level
`#![cfg(feature = "delta-scan")]` and holds 6 parity tests, including
`multicell_udt_collection_paths_parity` — the parity case for #3612's subject. **No merge-gating
lane executed them**, and the one workflow that did (`parity-regen-matrix.yml`) is registry-exempt
from `required` on a reason its own `pull_request:` trigger falsified.

**AC6 first, because the wording governs the fix:** the file did NOT "execute nowhere". It executed
strictly — `CQLITE_REQUIRE_FIXTURES=1`, after a Docker corpus regeneration — in
`parity-regen-matrix.yml`'s `cql-type` leg. The defect was the *gating* half. That distinction is
load-bearing: "executes nowhere" invites adding a second executor, when the fix is to make an
existing lane gate.

## The measured baseline, including one thing the ACs did not anticipate

Run directly on this lane against `/data/datasets`:

| datasets root | result |
|---|---|
| real corpus | `6 passed`, **11 rows compared** across the 6 cases |
| an empty directory | `6 passed`, **0 rows compared**, `0.00s` |
| `CQLITE_DATASETS_ROOT` unset | `6 passed`, **0 rows compared**, `0.00s` |

So the tests are healthy, and **AC3's "non-zero test count" is satisfiable VACUOUSLY**: the file's
own `skip_or_fail` (`:457`) prints `[SKIP]` and returns unless `CQLITE_REQUIRE_FIXTURES=1`. A naive
executor would print `6` and compare nothing. The executor here therefore forces strict fixtures on
the full gate — the exact `node-bindings` precedent from #3522 — and the lane is enrolled in
`DATASET_COMPONENTS`.

Before relying on failing closed, I verified the **named remedy actually works**: `/data/datasets`
carries the pin for `cassandra5-small-full-v3.5.tar.gz` at the sha `fetch-datasets.sh:59` pins, and
`test_basic`/`test_types` have mtimes **43 ms apart** — one tar extraction. So the published bundle
does contain `test_types`, and a fail-closed verdict points at a remedy that fixes it. (Discounted
counter-evidence: `metadata.yml` lists only 5 keyspaces, but it predates the extraction and is not a
bundle inventory.)

## What changed

**1. `feature-iso-delta-scan` widened from compile-only to EXECUTING (AC2, AC3).** It was
`cargo test --lib --no-run`. It now executes its `--lib` suite plus a set of crate-level
`delta-scan`-gated `--test` targets **derived from committed source at run time** (derive-never-
curate: a new gated file is picked up with no gate edit; a failed derivation FAILs naming the
derivation and never falls back to an empty set), under a per-target non-zero-tests guard, with the
narrowing DECLARED at run time.

Deliberate choices worth review attention:
- **The component NAME is unchanged**, so the #3544 `component-set:` baseline is unchanged and no
  in-flight branch on the fleet gets a fail-closed `BEHIND` verdict out of this.
- **The isolation property is preserved**: delta-scan WITHOUT parquet, never `--all-features`.
  The isolation this lane asserts is parquet-vs-delta-scan, and `write-support` is carried for the
  same reason the lane already carried `all-compression` — it is in `default`, and dropping a
  default feature would change what the lane measures from FEATURE ISOLATION into
  NO-WRITE-SUPPORT support. That converts the 3 targets gated `all(delta-scan, write-support)`
  from declared allowed-zero gaps into EXECUTED targets. (First cut used the allowed-zero route;
  roborev round 1 F2 correctly pushed back, and the allowed-zero MECHANISM is retained because it
  is still needed generally.)
- `issue_1704_scan_path_error_counts` is excluded **structurally** (unmet `required-features`;
  cargo REJECTS an explicit `--test` for it, so naming it would be a false red) and reported as a
  declared gap.
- The allowed-zero derivation is **three-valued with the conservative polarity**: an unmodelled cfg
  shape COSTS the target its excusal rather than granting it — "not a claim the target is fine; a
  refusal to claim it."

**2. The registry corrected (AC1).** `parity-regen-matrix.yml`'s reason was factually false. It now
states that the lane DOES run on matching pull requests and gives a real mechanism for why its PR
result is non-gating: per-leg Docker corpus regeneration from `cassandra:5.0.2`, a 60-minute job
timeout, and `paths:`-filtering that leaves its context ABSENT on most PRs — where absence is an
error for a registered tier, so enrolling it as-is would deadlock rather than gate. That is why
AC1's "enrol as a tier" branch was rejected, on a stated mechanism rather than by fiat.
**`ci.yml`'s entry was also stale** — it asserted `feature-iso-delta-scan` is `--lib --no-run` and
that the local gate only COMPILES those tests — and is corrected in the same diff, per #3493's rule
that narrowing or widening a component named by an exemption is a change to a merge gate.

**3. The class fix (AC5), which is smaller and sharper than the AC assumed.** Measured with the
repo's own rule:
- AC5's first clause needs **no new code**: a well-formed unregistered PR-triggered workflow already
  produces exactly 1 named error. And since **all 23 exempt workflows already have PR triggers**,
  the clause is **vacuous for the exempt population** — a guard there would have had an empty
  subject set, which greens vacuously.
- A stored `pr_triggered` boolean was **designed and rejected**: it would be constant across all 23
  AND a second source of truth that can drift from the workflow. Derive it, don't store it (#3544's
  "remove the second source rather than reconcile it").
- What IS uncovered, and is now checked: every `exempt:` entry declares its **merge-gating half** in
  a structured field under a closed grammar, and a named local gate component must EXIST in
  `scripts/agent-gate.components`. This checks **EXISTENCE, not SCOPE** — scope is #3493's actual
  hazard and is not decidable from the registry — and says so, so nothing reads it as a coverage
  proof. The prose `reason` is deliberately NOT machine-checked (#3312: a recogniser over
  author-controlled prose never closes); that is a DECLARED RESIDUAL.
- **A reproduced false PASS, fixed:** `load_workflows` mapped a `Psych::SyntaxError` to `{}`, so an
  unparseable **PR-triggered** workflow had no triggers, failed `pull_request_workflow?`, and
  ESCAPED the enrolment rule — `policy_errors` reported the tree CLEAN. Measured `TOTAL_ERRORS=1`
  well-formed vs **`0`** unparseable. Now three-valued and named.

**4. Doctrine (CLAUDE.md)** updated in the same change, including the delta-scan census correction,
the AC6 wording rule, a THIRD circular-deferral instance, and the two-valued-predicate rule
generalised off bash to any construct returning a default on unreadable input (`rescue`, `||`,
`${VAR:-default}`, `.ok()`).

## The closure paid for itself on the first run

Executing the lane surfaced **2 of `scan_delta_parity_test`'s 14 tests FAILING against the real
corpus and PASSING against an absent one** — a latent defect in a target nothing merge-gating ran.
The per-cell `expires_at` comparison did not model sstabledump's suppression rule.

**The oracle was verified independently, against Cassandra source at the pinned tag** (not against
CQLite's own output, and not by trusting the commit message) —
`apache/cassandra@cassandra-5.0.8`, `src/java/org/apache/cassandra/tools/JsonTransformer.java`:

- `:501` — `if (cell.isExpiring() && (liveInfo.isEmpty() || cell.ttl() != liveInfo.ttl()))` guarding
  `ttl` / `expires_at` / `expired`.
- `:497` — the identical suppression on `tstamp`, which is what makes the asymmetry principled.

So it is a **display-convention** difference (sstabledump omits a cell copy that duplicates the row
liveness; scan_delta emits what it decoded), not a decode divergence. The `(None, Some(_))` direction
stays STRICT, so the tolerance cannot mask a real divergence. A declared residual notes that
Cassandra suppresses on TTL equality while the comparator tolerates on `expires_at` proximity: that
cannot produce a false PASS (different TTLs with equal `expires_at` ⇒ Cassandra prints, so the branch
is unreachable), and can only produce a false FAIL for cells sharing a TTL written >1s apart in one
row — impossible from a single `INSERT ... USING TTL` and absent from the corpus.

This is #1716's "expect latent failures the first time you touch a long-unwired thing" at feature
granularity, and it is the argument for wiring the remaining families (`observability-testing` 14,
3 of 5 `dhat-heap`) rather than filing them.

## A third circular-deferral instance, found without looking for it

`test-data/scripts/smoke-test-all-tables.sh` skips `test_types` on the stated ground that those
keyspaces are "validated by dedicated Rust parity tests (tombstone/TTL + CQL-type)" — and every test
it names (`issue_1007`/`1003`/`1006`/`1008`, `issue_1010`/`1011`) was `delta-scan`-gated and executed
by nothing merge-gating. A closed, empty loop with an honest rationale on each side — the #3493
family's third confirmed instance. **The smoke skip itself is correct and untouched** (those
keyspaces genuinely hold partitions with zero live rows, which its "must emit ≥1 entry" check would
mis-flag); what was wrong is that its deferral had no gating target.

## Environment note the reviewer should know

`ruby` was **not installed** on this fleet box and `bootstrap-agent-machine.sh` does not install it,
so `tooling-tests` was SKIPping `test_gating_registry_policy.sh` and its two siblings with an honest
`SKIP: ruby unavailable`. Ruby is the single implementation path for the gating registry. I installed
ruby 3.2.3 (satisfies the declared 3.0.0 floor) so this change's guard could be proven red-then-green
locally rather than verified only in CI. Pre-change those suites are `74 / 26 / 170` PASS — **270
assertions that SKIP fleet-wide** — so this lane's gate executes them where a stock box would not.
Proposed separately as a bootstrap follow-up (see the issue thread); not widened into this PR.

## Review rounds

**roborev round 1 (job 3, codex/`gpt-5.6-sol`): `RESULT: FAIL`, `findings: PRESENT (4)`.** A genuine
review — input 3,132,954 / cached 2,939,136 tokens, every deterministic key PASS including
`prompt-content: PASS (8/8 code census paths present)`, `sha-assert: PASS`, both vacuity tiers —
reviewed range `394e25e5e..13832e7c3`. All four addressed:

* **F1 (High)** — `scan_delta_parity_test` ignored `CQLITE_REQUIRE_FIXTURES` and the lane treated
  fixture-blind targets as informational, so the very TTL test this PR fixes could merge-gate having
  compared nothing. **This issue's own defect, one target over.** Fixed: the test honours the strict
  flag, and strict-mode fixture blindness fails the lane. Verified safe before accepting the fix
  direction: the pinned bundle DOES ship `test_deltas` (9 `*-Data.db`; `test_deltas`/`test_basic`/
  `test_types` share one 75 ms extraction batch), so failing closed does not red the fleet.
* **F2 (Medium)** — 4 of 13 targets still executed nothing while the census declared the family
  closed. Fixed by adding `write-support` (see above); `issue_1704` remains a declared gap because it
  needs `observability-testing`, the family the census already lists as STILL OPEN (#3375).
* **F3 (Medium)** — the comparator proxied TTL equality with `expires_at` proximity. Fixed properly
  rather than declared as a residual: `JsonTransformer` prints `ttl` in `liveness_info` (`:288`) and
  per cell (`:503`), so both sides carry the authoritative value. Includes the
  equal-TTL/different-writetime case that pins it.
* **F4 (Low)** — a valid-but-EMPTY workflow document returned `nil` and was accepted, the same
  permissive collapse one branch over from the parse-failure fail-open. Fixed and named; the fix's
  own comment records that 0 of the 42 real workflows parse to `nil`, so it cannot red a correct tree.

**Independently verified by the lead** (not taken from subagent reports): the Cassandra oracle at the
pinned tag (both `:501` and `:497`); the fail-open fix in BOTH directions (`0`→`1` named error on an
unparseable workflow, real tree stays `0`); the empty-document fix likewise; the ruby suites
`74 → 93 → 95` PASS with the case floor raised each round; `validate-workflows.rb` exit 0 over 42
workflows; both disposition guards; and the **#3400 audit** — the lane adds no new cargo-output parse
(it delegates to the already-routed guards), there is no piped `| while` anywhere in the new code, and
all four new read-loops are heredoc-fed so their verdicts survive.

## Two side findings, reported not fixed (out of scope)

1. **Stale `test_deltas` publication claims.** `smoke-test-all-tables.sh:101` ("Data.db binaries not
   yet in the published dataset asset", #701) and `check-dataset-manifest.sh:11` ("skip-pending and
   intentionally NOT enforced") both disagree with the measured corpus, and
   `cassandra-parity-manifest.yml:508-515` references committed `test_deltas` binaries. A
   skip-pending exemption that outlived its cause is the same shape as this issue's registry reason.
   Only the instance inside a file this PR already touches was corrected.
2. **11 exemptions declare `kind: none`** — no merge-gating half at all. About six are deliberate
   (`docs-site` fixed forward, `coverage-baseline` advisory, `fuzz` files bugs, `project-board-sync`
   automation). **`delta-roundtrip.yml` is the notable one**: its prose claimed it was "covered by the
   local gate's delta tests" and under a checked field it resolves to `none` — a third stale coverage
   claim in the same registry. Proposed as a follow-up rather than widened into this PR.

## Gate-of-record invocation — REQUIRED variable

The `file-size` campsite ratchet FAILs on this diff. Measured, not assumed (`--only file-size`):

```
cqlite-core/tests/scan_delta_parity_test.rs: 1715 -> 1761 (limit 1500)
--- [file-size] FAIL: change makes over-threshold file(s) larger.
```

With the documented acknowledgement it passes (`file-size: PASS`), and the in-code note linking
**#1135** is committed (`ca49d48cb`). **The gate of record MUST therefore be run with
`CQLITE_ALLOW_FILE_GROWTH=1`** — and note the SUMMARY does not record that it was set, since
`file-size: PASS` renders identically either way.

Why the split is out of scope here rather than done: the growth is an oracle citation plus a
signature change on a fix that is itself incidental to this issue (pre-existing rot the widened lane
surfaced), and splitting a 1700-line parity suite is its own change under #1135. The alternative —
trimming the Cassandra-source citation to game a line count — is the wrong trade: that comment is
the only thing making the parity expectation auditable.

## Measured outcome of the widened lane

`--only feature-iso-delta-scan` on this lane:

```
feature-iso-delta-scan: PASS (12s)  [test cqlite-core --no-default-features
                                    --features all-compression,write-support,delta-scan]
```

Its run-time census, which is the declaration the "a narrowed lane DECLARES the narrowing" rule asks
for:

```
EXECUTES (issue #3725): cargo test -p cqlite-core --no-default-features
  --features all-compression,write-support,delta-scan --lib + 12 derived --test target(s)
  every derived target reads a strict-fixture variable: 0 RECOGNISED gaps
  allowed-zero: 0 RECOGNISED — every derived target must execute at least one test
  UNATTRIBUTED cfg site(s): 0 RECOGNISED
```

**Cost: `PASS (12s)` against `PASS (1s)` for the old compile-only instrument.** Two honesty notes on
that number rather than one clean figure: it was **7s** on an earlier run of the same lane and 12s on
the final one, because warm-cache timing varies with what the side target dir already holds — so read
it as roughly 7-12s to execute this family, not a precise constant. And there is no defensible COLD
number on record for the shipped `--lib --no-run` instrument to compare against (the committed 10s/18s
figures measure a retired `cargo check` shape), which is why the comparison is warm-to-warm.

Every zero above is affirmative (`0 RECOGNISED`, never a bare `0`), and the census states its own
non-exhaustiveness — *"that none was RECOGNISED, never evidence that none is THERE"*.

**What the lane now executes: 72 integration tests + the `cqlite-core` lib suite's 3199 at that
feature set**, with `allowed-zero: 0 RECOGNISED` — every one of the 12 invoked targets executes at
least one test and none is excused.

Stated precisely, because the two halves are NOT the same claim:
* **The 72 integration tests are new merge-gating coverage.** They executed in no local gate
  component and in no `required`-gating workflow — only in the exempt `parity-regen-matrix.yml`.
* **The 3199 lib tests are NOT 3199 newly-covered tests.** Most also run in `core-tests`, at a
  DIFFERENT feature set (`cli-helpers`, delta-scan OFF). What is new is that they now run with
  `delta-scan` ON and the defaults OFF — which is where cross-feature coupling shows up (#1699's
  point: a combined feature set is exactly what MASKS it) — and that the `delta_scan` module's own
  suite, which previously executed only in the `required`-exempt `ci.yml`, now executes in the gate.
  Dropping `--no-run` is what buys both.

**Per-target, executed where nothing merge-gating ran before** (real corpus, strict, at the
lane's feature set): `issue_1007` 6, `issue_1008` 5, `scan_delta_parity_test` 15, `issue_1012` 5,
`issue_1014` 6, `issue_1015` 6 — plus `issue_1003`, `issue_1006`, `issue_1010`, `issue_1011`,
`issue_1013`, `issue_1295` and the `delta_scan` module's own lib suite.

### The anti-vacuity property, proven in both directions

This is the property the issue exists to establish, so it is stated as a measurement rather than a
claim. At the lane's feature set, with `CQLITE_REQUIRE_FIXTURES=1`:

| target | real corpus | EMPTY datasets root |
|---|---|---|
| `issue_1007_complex_type_parity` | 6 passed | **FAILED — 0 passed, 6 failed**, each naming the missing table |
| `issue_1008_counter_final_value_parity` | 5 passed | **FAILED — 0 passed, 5 failed**, naming keyspace.table + scenario id + remedy |
| `scan_delta_parity_test` | 15 passed | **FAILED**, naming the table and remedy |

Before this change every one of those returned "passed" against an empty root, having compared
nothing.

**And the property was then verified BEHAVIOURALLY across all 12 derived targets**, not just the
three roborev and I had discussed — empty datasets root, `CQLITE_REQUIRE_FIXTURES=1`, the lane's
feature set, `--no-fail-fast`:

```
issue_1003 FAILED 0/6   issue_1011 FAILED 0/4   issue_1014 FAILED 0/6
issue_1006 FAILED 0/5   issue_1012 FAILED 0/5   issue_1015 FAILED 0/6
issue_1007 FAILED 0/6   issue_1013 FAILED 0/3   issue_1295 FAILED 0/5
issue_1008 FAILED 0/5   issue_1010 FAILED 0/6   scan_delta FAILED 1 passed / 14 failed
```

**71 of the 72 integration tests fail with the corpus absent.** The one pass is `scan_delta`'s
fixture-free synthetic refusal case, which needs no corpus and should pass. This matters because a
TEXTUAL check for `CQLITE_REQUIRE_FIXTURES` cannot distinguish an executable lookup from the string
appearing in a comment — a weakness roborev round 3 raised against the lane's own check, and one my
first census shared. The numbers above are behavioural, so they settle the current population on
evidence a grep cannot provide.

**`issue_1008` was NOT in roborev's finding — it was found by censusing all 12 invoked targets for
whether they honour the strict flag.** roborev named `scan_delta_parity_test`; the census showed
`issue_1008` had zero references to either strict variable while being corpus-dependent (`:77` reads
`CQLITE_DATASETS_ROOT`, its own doc promises to "SKIP cleanly when unset", bare `return;` at `:482`/
`:489`/`:657`, and `parity-regen-matrix.yml` already runs it in the STRICT group). Left unfixed it
would have made the new strict rule red a correct tree — a guard that fails on valid input is the
guard agents learn to waive. Fixing the instance without censusing the class would have shipped that.

## The TTL suppression check, and why it is exact rather than a proxy

roborev found this twice, and both findings were right, so the resolution is recorded with its
measurements. The sequence: the original code compared cell `expires_at` to row `expires_at`; round 1
said Cassandra suppresses on TTL EQUALITY (`cell.ttl() != liveInfo.ttl()`, `JsonTransformer.java:501`);
the fix derived `ttl = expires_at - writetime`; round 2 said that derivation is bogus under
`USING TIMESTAMP`, where writetime is independently controlled while expiration comes from coordinator
wall-clock. All three observations are correct. The missing piece is structural: **`CellDelta` carries
`writetime` and `expires_at` but no TTL**, while `CellExpiration` (`types.rs:411`) decodes the
authoritative `ttl_seconds` and the delta model discards it.

**Adding `ttl_seconds` to `CellDelta` was considered and REJECTED for this PR** — it is a public
delta-model change in service of a test, on an issue scoped to test-execution gating. It is filed as a
follow-up on its own merits (the model keeping a derived value while discarding the authoritative one
inverts this project's authoritative-metadata principle).

**A correction to an earlier revision of this PR body, kept visible rather than quietly edited out.**
This section previously claimed the check compares `expires_at` under a precondition that makes it
**equivalent** to TTL equality, "since `expires_at ≈ writetime + ttl`". **That was false**, roborev
round 4 caught it, and the pinned source is unambiguous —
`BufferCell.java:79-82` builds an expiring cell with `computeLocalExpirationTime(nowInSec, ttl)`, so
`localDeletionTime` is a function of the coordinator's wall clock and the TTL and **never of
`timestamp`**. Under explicit `USING TIMESTAMP` the two decouple.

What ships instead, stated at what it actually establishes: the check verifies cell/row expiry
**CONSISTENCY** — a real property of the decoded data — which coincides with Cassandra's TTL-equality
rule under a precondition about the **WRITE** (one operation ⇒ one `nowInSec`), not about the
timestamp. It **refuses by name** rather than guessing where it cannot decide. Two residuals are
DECLARED at the check, both measured for reachability:

* **false FAIL** (loud): two writes sharing an explicit `USING TIMESTAMP` at different wall-clock
  times with the same TTL are suppressed by Cassandra yet carry different expiries here. Unreachable
  in the corpus — no fixture uses `USING TIMESTAMP`.
* **false ACCEPT** (silent, the worse direction): expiry saturates at
  `min(nowInSec + ttl, cellMaxDeletionTime)` (`ExpirationDateOverflowHandling.java:120-125`;
  cap is version-dependent per `Cell.java:51-57` / `:91-100`), so two DIFFERENT TTLs that both
  saturate produce EQUAL expiries, which this check would accept. **Measured unreachable: 0 of the
  768 suppressed expiring cells sit at either clamp value, and the largest expiry anywhere in the
  corpus is 1782428346 s (2026-06-25) — nine decades short of the cap.**

That second residual was found by neither review round nor by the ruling that set up the check, which
is why the list is scoped as the residuals **RECOGNISED** rather than presented as complete. **The
authoritative fix — carrying `ttl_seconds` through `CellDelta` so the TTLs themselves are compared —
is ESCALATED, not made here** (see the issue thread); this is the third finding from that one root
cause, which is the argument for fixing the mechanism rather than the derivation again.

The corpus measurements, at two scopes, because the number is scope-sensitive and a bare count beside
a test that reads a fraction of the corpus would mislead:

| scope | goldens | live cells | suppressed expiring cells | undecidable |
|---|---|---|---|---|
| this target's own subject | 27 | 51,048 | 30 | **0** |
| whole committed corpus | 162 | 75,016 | 768 (8 ttl-bearing goldens) | **0** |

So the comparison is exact for every cell this target actually compares, and the refusal branch is
known-defensive rather than dead. **"Defensive" does not mean "unexercised":** it is pinned at unit
level by `suppression_rule_requires_equal_writetimes_or_refuses`, which drives the refusal with two
synthetic values and needs no fixture.

One incidental fact from the same measurement, recorded because it bounds what the corpus can test:
**zero** cells in the entire corpus print their own `ttl`/`expires_at` — every expiring cell there is
suppressed — so the suppression path is the only one real data drives.

## Certification

`--lite` at `0a14ce7`, the state this PR opens on:

```
commit: 0a14ce7 branch: issue-3725-issue-1007-parity-lane-unexecuted dirty: no
lite-scope: file-size fmt clippy roborev-lints scoped-tests
tree-start: 0a14ce76df83 dirty: no digest: c80d468c204a
tree-end:   0a14ce76df83 dirty: no digest: c80d468c204a
tree-integrity: PASS
fmt:               PASS (7s)
clippy:            PASS (4s)
roborev-lints:     PASS (57s)
scoped-tests:      PASS (4s)  [test cqlite-core --features cli-helpers]
RESULT: PASS
```

**That run was invoked with `CQLITE_ALLOW_FILE_GROWTH=1`, and the block does not say so** — `file-size:
PASS` looks identical either way. Flagging it explicitly because a reader cannot tell from the summary,
and the same applies to the gate of record: without the variable, `file-size` FAILs at
`scan_delta_parity_test.rs: 1715 -> 1761 (limit 1500)` (measured via `--only file-size`).

Supporting checks, all run by the lead against this tree rather than taken from a subagent report:

| check | result |
|---|---|
| `test_gating_registry_policy.sh` | `PASS=97 FAIL=0` (baseline 74; floor raised each round) |
| `test_gating_workflow_semantics.sh` | `PASS=26 FAIL=0` |
| `test_aggregate_required_tiers.sh` | `PASS=170 FAIL=0` |
| `test_workspace_test_disposition.sh` + selftest | PASS (`2 green controls + 15 attributed red cases`) |
| `test_agent_gate_feature_matrix_annotation.sh` | `85 passed, 0 failed` — incl. "all 29 exercised cargo-class components render an observed matrix or a NAMED terminal state — none reads UNDECLARED" |
| `ruby scripts/ci/validate-workflows.rb` (real tree) | exit 0, 42 workflows |
| `--only feature-iso-delta-scan` | `PASS (12s)`, census `0 RECOGNISED gaps` |

## Gate of record — full block

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.5fvmvA
commit: 362cec8 branch: issue-3725-issue-1007-parity-lane-unexecuted dirty: no
schemas: 8/8 canonical .cql readable under /data/lanes/lane-3725/test-data/schemas (checkout-relative)
component-set: PASS (37/37 names vs origin/main 1dfae67ab004d027194e0c47ce53838537fe35ca; NAMES ONLY — not implementations, and no component is run here) — baseline read via the committed manifest; objects: baseline REUSED from this lane's SHARED store — store TRUSTED, not verified (#3746)
tree-start: 362cec880527 dirty: no digest: 3bc5830a24f3
tree-end: 362cec880527 dirty: no digest: 3bc5830a24f3
tree-integrity: PASS
clippy:            PASS (4s)  [clippy workspace(excl 5) --all-features | clippy cqlite-core --features 33:all-compression,arrow,bench-internals,+30 more | clippy cqlite-cli --features 10:benchmarks,ci_zero_tolerance,cli-helpers,+7 more | clippy cqlite-flight+cqlite-py+cqlite-node --features cqlite-node/write-support]
roborev-lints:     PASS (57s)  [no-cargo]
core-tests:        PASS (87s)  [nextest run cqlite-core --features cli-helpers | test cqlite-core --features cli-helpers]
tombstones-scan:   PASS (0s)  [test cqlite-core --features write-support,cli-helpers,tombstones]
scan-offload-guard: PASS (2s)  [test cqlite-core --features cli-helpers,scan-offload-probe x2]
work-counters-guard: PASS (18s)  [test cqlite-core --features write-support,cli-helpers,state_machine,work-counters]
arrow-parity-guard: PASS (0s)  [test cqlite-core --features arrow]
memory-budget:     PASS (13s)  [test cqlite-core --features cli-helpers,dhat-heap,arrow x3 | test cqlite-flight --features dhat-heap]
integration-tests: PASS (1s)  [test cqlite-integration-tests default-features x2]
format-compat:     PASS (0s)  [test format-compatibility-tests default-features]
write-tests:       PASS (10s)  [test cqlite-core --features write-support x3]
cli-tests:         PASS (6s)  [test cqlite-cli default-features | test cqlite-cli --features write-support]
compaction-byte-parity: PASS (2s)  [test cqlite-core --features write-support x2]
flight-query-semantics-oracle: PASS (1s)  [test cqlite-flight default-features]
flight-tests:      PASS (7s)  [test cqlite-flight default-features]
legacy-heuristics: PASS (10s)  [build cqlite-core --features legacy-heuristics]
feature-iso-parquet: PASS (1s)  [test cqlite-core --no-default-features --features all-compression,parquet]
feature-iso-delta-scan: PASS (7s)  [test cqlite-core --no-default-features --features all-compression,write-support,delta-scan]
python-bindings:   PASS (50s)  [via maturin: feature set NOT observed]
node-bindings:     PASS (127s)  [via npm run build (napi): feature set NOT observed]
binding-rust-tests: PASS (3s)  [test cqlite-ffi-common default-features | test cqlite-node --features write-support]
oom-audit:         PASS (1s)  [build xtask default-features | run xtask default-features]
parity-report:     PASS (0s)  [run cassandra-parity default-features]
operator-metrics-doc: PASS (0s)  [run cqlite-core default-features]
kit-dashboard-drift: PASS (0s)  [test cqlite-core default-features]
pub-surface:       PASS (1s)  [no-cargo]
tooling-tests:     PASS (2109s)  [test ws0-corpus-gen default-features | + cargo not observable: cargo may run inside ~60 nested test scripts (child processes)]
minimal-build:     PASS (2s)  [build cqlite-core --no-default-features --features all-compression | test cqlite-core --no-default-features --features all-compression]
all-features-check: PASS (3s)  [check cqlite-core --all-features | clippy cqlite-core --all-features]
smoke:             PASS (3s)  [build cqlite-cli default-features]
logs: /tmp/agent-gate.5fvmvA
summary-file: /tmp/cqlite-gate-RhPlwq/summary.txt
heartbeat: on file: /tmp/cqlite-gate-RhPlwq/summary.txt.heartbeat interval: 20s
launch-nonce: 20260901T121850Z-3333246-2593613769
RESULT: PASS
```


pmcfadin:

## What I measured, just now

```
issue-3725 (#3786)  clean
```

```
```

## The new rule

pmcfadin:
🧭 **`#3708` MERGED at 13:40Z — you are cleared to rebase. Your conflict is the ONE I predicted and the only one that materialised, so this is expected, bounded, and yours to resolve now.**

## What happened, and why only you

When I released the serial queue I measured every member pairwise and reported:

```
```

`#3708` has now landed. I re-checked all nine sibling branches against the new `main`:

```
issue-3725 (#3786)   CONFLICT   <- you
issue-3220, issue-3740, issue-3689, issue-3641, issue-3436,
issue-3747, issue-3723, issue-2339      all clean
```

**Exactly one conflict, exactly the one predicted.** Nothing else needs to do anything. You were held in an ordered pair for precisely this reason, and now the blocker is gone.

## What to do

1. **Rebase onto current `origin/main`.** `#3708` was `fix(#3414): the gate reports where its slot came from …` — its changes are in `scripts/agent-gate.sh`, which is also your file, so expect the conflict there.
2. **Resolve semantically, not just textually.** `#3708` landed the `CQLITE_GATE_MAX_CONCURRENCY` provenance reporting plus three real fixes found in its last two rounds: the **`08` octal** parse that killed the gate inside its own `cpu-budget` line with no verdict, the **integer-overflow** bound (`=99999999999999999999` was pinning `7766279631452241919` — a confident false pin), and a Linux-vs-Mac remedy string. **If your diff touches the same emit paths or the cap resolver, check your assumptions against what actually landed** — git going quiet is not evidence you are correct. #3629 lost a review round to exactly that gap.
3. **Everything you had is void after the rebase** — gate of record, roborev round, C, and any `merge-tree` clean result. That last one is the trap `lane-3436` found: it recorded `rc=0`, a PR landed, and the result became false silently because nothing re-runs it. Re-certify from the rebased head: gate → C → one final roborev → `gh pr ready 3786` → `premerge-assert.sh` → arm.
4. **While conflicted you get NO CI at all** — GitHub cannot compute `refs/pull/3786/merge`, so `pull_request` workflows never fire. That is why `req=` is empty; it is not a red check and not a failure to diagnose. It clears the moment the conflict does.

## Your ruling from earlier still stands

**(c) is in and unchanged**: the scoped claim at the comparison site naming **#3787**, the false justification deleted rather than softened, and #3787 filed with the Cassandra-written `USING TIMESTAMP` + TTL fixture as its expensive AC. Nothing about the rebase disturbs that.

**And the standing rule applies to your re-review:** do **not** re-run roborev hoping for silence. A reviewer that happens not to mention a defect has not cleared it.

## Before you gate

**box6 is at ~97 GB with one gate running** (`lane-3648`, which I restarted from `DEAD-NO-SESSION` earlier and which is now producing certified work). Check `df -h /data` and `pgrep -f agent-gate.sh | wc -l` before launching, and **pass the cap inline** — your session predates the pin so it cannot inherit it:

```bash
```

With the cap set, launching while a peer holds the slot simply **queues**, consuming nothing until the slot frees. **If a peer gate is live and you would rather not queue, tell me and I will sequence you** — three concurrent cold gates provably do not fit on a 295 GB volume, which is why one-gate-at-a-time is now policy on the boxes that have hit it.

**#3708's merge is also good news for your gate:** the `1465-skip-declares` SIGPIPE fix and the two cap-resolver fixes are now on `main`, so your run inherits them rather than tripping over them.

pmcfadin:
## Gate of record + roborev, at the certified head `1e5509c89bf27b22f0e1b71f84c53d86c3d47dc0`

`premerge-assert.sh 3786 1e5509c89… <summary>` -> **`PREMERGE: OK`**, exit 0.
Reported as **gate of record verified at `1e5509c89`** — NOT "certified against main": a squash-merge composes this diff with main's tip, which no gate here executed (#3650 slice 2 unimplemented). Base-staleness advisory: 1 of 3 commits behind touches the blast radius (`f53538edb`, the #4032 fix) — non-blocking, and this gate PASSED without it, i.e. the intermittent never fired.

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.SUIcEs
commit: 1e5509c branch: issue-3725-issue-1007-parity-lane-unexecuted dirty: no
schemas: 8/8 canonical .cql readable under /data/lanes/lane-3725/test-data/schemas (checkout-relative)
component-set: PASS (38/38 names vs origin/main 7ed12a1c2212016142892f1970a9a50c2c1efe4d; NAMES ONLY — not implementations, and no component is run here) — baseline read via the committed manifest; objects: baseline REUSED from this lane's SHARED store — store TRUSTED, not verified (#3749)
tree-start: 1e5509c89bf2 dirty: no digest: edcc8a30d040
tree-end: 1e5509c89bf2 dirty: no digest: edcc8a30d040
tree-integrity: PASS
census: 24/38 components AFFIRMED a count; 13 DECLARED-GAP (RECOGNISED); 0 NOT-MEASURED (RECOGNISED); 0 measured-ZERO (RECOGNISED); 1 not-applicable (component did not PASS); 0 no-subject (PASSed; the run had nothing to measure); 0 UNDECLARED; 0 unrecognised; 0 row(s) carry a VACUOUS status. NON-EXHAUSTIVE: the gap set is CURATED, so an unaffirmed component is UNMEASURED, never verified (#3625; the remaining gaps are tracked in #3162).
file-size:         OPT-OUT (0s)  [no-cargo]  {no census: component ended OPT-OUT, so there is no PASS to affirm} — CQLITE_ALLOW_FILE_GROWTH=1 (ratchet NOT enforced); 1 over-threshold file(s) grown — see file-size.log under logs:
fmt:               PASS (6s)  [fmt workspace features=n/a]  {no census — cargo fmt --all --check emits no per-file tally to count}
clippy:            PASS (287s)  [clippy workspace(excl 5) --all-features | clippy cqlite-core --features 33:all-compression,arrow,bench-internals,+30 more | clippy cqlite-cli --features 10:benchmarks,ci_zero_tolerance,cli-helpers,+7 more | clippy cqlite-flight+cqlite-py+cqlite-node --features cqlite-node/write-support]  {no census — cargo clippy emits a per-crate tally only COLD; a warm run prints Finished alone}
roborev-lints:     PASS (110s)  [no-cargo]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
core-tests:        PASS (1125s)  [nextest run cqlite-core --features cli-helpers | test cqlite-core --features cli-helpers]  {verified: 5759 tests passed (across 2 result line(s))}
tombstones-scan:   PASS (47s)  [test cqlite-core --features write-support,cli-helpers,tombstones]  {verified: 2 tests passed (across 1 result line(s))}
scan-offload-guard: PASS (169s)  [test cqlite-core --features cli-helpers,scan-offload-probe x2]  {verified: 15 tests passed (across 8 result line(s))}
work-counters-guard: PASS (96s)  [test cqlite-core --features write-support,cli-helpers,state_machine,work-counters]  {verified: 60 tests passed (across 15 result line(s))}
byte-budget-guard: PASS (15s)  [test cqlite-core --features write-support,cli-helpers,state_machine]  {verified: 21 tests passed (across 5 result line(s))}
arrow-parity-guard: PASS (41s)  [test cqlite-core --features arrow]  {verified: 4 tests passed (across 1 result line(s))}
memory-budget:     PASS (291s)  [test cqlite-core --features cli-helpers,dhat-heap,arrow x3 | test cqlite-flight --features dhat-heap]  {verified: 6 tests passed (across 4 result line(s))}
integration-tests: PASS (60s)  [test cqlite-integration-tests default-features x2]  {verified: 42 tests passed and 25 test binaries built/verified}
format-compat:     PASS (40s)  [test format-compatibility-tests default-features]  {verified: 10 tests passed (across 1 result line(s))}
write-tests:       PASS (160s)  [test cqlite-core --features write-support x3]  {verified: 3935 tests passed (across 3 result line(s))}
cli-tests:         PASS (211s)  [test cqlite-cli default-features | test cqlite-cli --features write-support]  {verified: 764 tests passed (across 48 result line(s))}
bti-multiclustering: PASS (5s)  [test cqlite-core --features state_machine,cli-helpers x2]  {verified: 19 tests passed (across 3 result line(s))}
flight-query-semantics-oracle: PASS (50s)  [test cqlite-flight default-features]  {verified: 4 tests passed (across 2 result line(s))}
flight-tests:      PASS (132s)  [test cqlite-flight default-features]  {verified: 407 tests passed (across 2 result line(s))}
legacy-heuristics: PASS (265s)  [build cqlite-core --features legacy-heuristics]  {verified: 3829 tests passed (across 6 result line(s))}
feature-iso-parquet: PASS (153s)  [test cqlite-core --no-default-features --features all-compression,parquet]  {verified: 1 test binaries built/verified}
feature-iso-delta-scan: PASS (107s)  [test cqlite-core --no-default-features --features all-compression,write-support,delta-scan]  {verified: 3509 tests passed (across 14 result line(s))}
python-bindings:   PASS (463s)  [via maturin: feature set NOT observed]  {verified: 786 pytest tests passed}
node-bindings:     PASS (703s)  [via npm run build (napi): feature set NOT observed]  {verified: 602 jest tests passed}
binding-rust-tests: PASS (92s)  [test cqlite-ffi-common default-features | test cqlite-node --features write-support]  {verified: 123 tests passed (across 5 result line(s))}
delivery-telemetry: PASS (2s)  [no-cargo]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
parity-report:     PASS (2s)  [run cassandra-parity default-features]  {no census — xtask/report driver prints no machine-readable subject count (#3162)}
dep-duplicates:    PASS (1s)  [via check-dep-duplicates.sh (cargo tree -d --workspace --target all; a metadata probe, compiles nothing): feature set NOT observed]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
tooling-tests:     PASS (3302s)  [test ws0-corpus-gen default-features | + cargo not observable: cargo may run inside ~60 nested test scripts (child processes)]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
minimal-build:     PASS (65s)  [build cqlite-core --no-default-features --features all-compression | test cqlite-core --no-default-features --features all-compression]  {verified: 1 test binaries built/verified}
all-features-check: PASS (292s)  [check cqlite-core --all-features | clippy cqlite-core --all-features]  {no census — cargo check/clippy passes execute no tests; the subject is a feature set, not a count}
smoke:             PASS (317s)  [build cqlite-cli default-features]  {no census — smoke-test-all-tables.sh prints no machine-readable table count (#3162)}
disk-admission: PASS (evaluated 2x: launch 161.0GiB; post-slot 161.0GiB; bar 40GiB(default); fs /data; target-dir /data/lanes/lane-3725/target (via cargo metadata)) — re-checked AT SLOT GRANT, not merely at launch (#3755); subjects 1 RECOGNISED (the cargo-resolved BUILD-OUTPUT filesystem only) — NON-EXHAUSTIVE: other filesystems this gate writes to (the python venv, bindings/node/node_modules) are NOT measured (#3886)
logs: /tmp/agent-gate.SUIcEs
summary-file: /tmp/claude-1000/-data-lanes-lane-3725/8b257272-0f22-4d5b-9690-ba49c2ea1eb3/scratchpad/gate-final.txt
heartbeat: on file: /tmp/claude-1000/-data-lanes-lane-3725/8b257272-0f22-4d5b-9690-ba49c2ea1eb3/scratchpad/gate-final.txt.heartbeat interval: 20s
launch-nonce: 20260904T044635Z-1813470-1997429
RESULT: PASS
```

```
repo: /data/lanes/lane-3725
branch: issue-3725-issue-1007-parity-lane-unexecuted
base: origin/main
head-sha: 1e5509c89bf27b22f0e1b71f84c53d86c3d47dc0
reviewed-sha: 7ed12a1c2212016142892f1970a9a50c2c1efe4d..1e5509c89bf27b22f0e1b71f84c53d86c3d47dc0
assert-base: 7ed12a1c2212016142892f1970a9a50c2c1efe4d (merge-base of origin/main and HEAD; origin/main tip 7ed12a1c2212016142892f1970a9a50c2c1efe4d)
job: 146
model: gpt-5.6-sol
census: 13 files, +2960/-204
tokens: input=720781 cached=619776 output=2816
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: FAIL (11/11 code census paths absent from the prompt)
waiver: NONE (no waiver comment for this review: the marker must be the SOLE NONBLANK CONTENT of a TOP-LEVEL PR comment — a marker inside prose, a code fence, a quote or a review body is not read; linked issue #3725 checked: no matching marker there either (NON-EXHAUSTIVE: a thread counts as read when its payload was a comments LIST the scanner accepted; a malformed ENTRY inside an otherwise well-formed list is skipped by the scanner and is not distinguished here))
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: NONE
roborev-exit: PASS
log: /tmp/claude-1000/-data-lanes-lane-3725/8b257272-0f22-4d5b-9690-ba49c2ea1eb3/scratchpad/roborev-r15.log
RESULT: FAIL
```

pmcfadin:
roborev-waive: prompt-content-absent base=7ed12a1c2212016142892f1970a9a50c2c1efe4d head=1e5509c89bf27b22f0e1b71f84c53d86c3d47dc0 job=146 reason=review-genuinely-ran-per-token-accounting-input-720781-cached-619776-output-2816-vs-vacuous-baseline-18700-input-0-cached-and-every-other-key-PASS-with-findings-NONE

pmcfadin:
## 🧭 Waiver GRANTED — commentary (kept separate from the marker on purpose)



lane-3725: authorization posted in its **own** comment above, containing nothing but the marker. That separation is required — a marker sharing a comment with prose is not read, and prose that quotes a marker must never grant one.

### What I verified before granting, in order

| constraint | evidence |
|---|---|
| **author allowlisted** | `ROBOREV_WAIVER_AUTHORS="pmcfadin"` on `origin/main`, hard-coded, not env-derived |
| **not self-applied** | you posted a REQUEST and did **not** re-ask — the one-marker-one-wait rule, honored |
| **bound to the whole scope** | `base=7ed12a1c2…` `head=1e5509c89…` `job=146`, all three present, both shas 40-hex |
| **absence verdict ONLY** | `prompt-content: FAIL (11/11 code census paths absent)`. Every other key PASSES: `push-assert`, `census-check`, `code-free`, `job-record`, `sha-assert`, `review-completed`, `vacuity-tier1`, `vacuity-tier2`, `roborev-exit`, and `findings: NONE` |
| **token accounting checked** | `input=720781 cached=619776 output=2816` |

That last row is the load-bearing one. A snapshot-delivered diff and a vacuous review that received nothing are **identical to the machine** — both have no census paths in the prompt, so both FAIL. What distinguishes them is a human reading the token accounting, which is precisely why this break-glass exists and why it is not automatable. **720k input / 620k cached** sits solidly in the genuine band; the vacuous baseline is ~18.7k input / 0 cached / 53–56 output. This review read a real 13-file, +2960/-204 diff.

### Your sequence from here is what you already queued

`--recheck-job 146` → HOLD re-read → arm → finalize → `lane-report landed` with the merge sha. The recheck re-decides **that job** and enqueues nothing; expect `prompt-content: WAIVED` and `RESULT: PASS`, with `MODE: recheck` declared — legitimate to merge on, never pasteable as evidence of a fresh review.

### One correction owed to you

You have been holding a valid gate PASS with `premerge-assert: OK` since **06:31Z**, blocked solely on an authorization only I could give. The delay was mine, not yours, and your refusal to re-ask or to self-waive was exactly right — a worker that waives its own PR defeats the whole control.

— coordinator

Issue #3785, opened 2026-09-01
[BUG] Three independent Murmur3 token implementations have drifted on the empty key, and nothing detects drift


pmcfadin:
Split out of #3633 by lead ruling on `REQ-3633-01` (2026-09-01). #3633 fixed the two Rust sites; this issue owns the Java one **and the drift-detection gap that is the real defect**.

## The issue #3633's premise is FALSE, and that correction matters as much as the fix

#3633's body states that `cassandra_murmur3_token` is "the single authoritative token definition shared by the k-way merge, the `WHERE pk IN (...)` fan-out and the scan fallbacks". **It is not.** There are **three** independent implementations of that one definition:

| # | Site | Empty-key behaviour |
|---|---|---|
| 1 | `cqlite-core/src/util/cassandra_murmur3.rs` `cassandra_murmur3_token` | `i64::MIN` — **fixed in #3633** |
| 2 | `cqlite-core/src/storage/sstable/writer/partitions_writer.rs` (BTI writer, inlined single-pass for #1681) | `i64::MIN` — **fixed in #3633** |
| 3 | `trino-connector/src/main/java/in/mcfad/cqlite/flight/Murmur3Token.java:38` | **still `0`** — this issue |

## The remaining defect (roborev finding, quoted verbatim)

> - **Severity**: High
> - **Location**: `trino-connector/src/main/java/in/mcfad/cqlite/flight/Murmur3Token.java:38`
> - **Problem**: The Java token implementation still returns `0` for empty keys. Plan-time split pruning therefore searches the token-0 range instead of `Long.MIN_VALUE`, potentially omitting the split containing an empty TEXT/BLOB partition key and returning incomplete results.
> - **Fix**: Return `Long.MIN_VALUE` when `data.length == 0` and update the existing empty-input parity test.

Source confirms it — `token(byte[]) { return normalize(hash3X64128H1(data)); }`, no empty-key branch.

## Why this was deferred out of #3633 rather than fixed there

`trino-connector` is a separate Java artifact that #3633's PR gate does not build or test, so fixing it there means shipping a Java change plus a Java parity-test change **certified by nothing**. A gate that does not compile an artifact cannot certify a change to it, and an uncertifiable change in a PR is worse than a filed, cross-referenced deferral.

## Acceptance criteria

[...]

Issue #2302, opened 2026-07-09
read path: CQLite-written Summary.db/Index.db pairs never fully resolve — silent fallback to sequential_scan


pmcfadin:
Surfaced during #2264's index-backed cancellation test work (PR #2301): investigation found CQLite-WRITTEN Summary/Index pairs never fully resolve through the index-random-read path — `iterate_all_partitions` silently falls back to the fully-materializing `sequential_scan` even when Summary.db/Index.db are present (pre-existing, unrelated to the cancellation fix; pinned observation via `read_work_counters::index_probes()` oracle on a `test_basic.multi_partition_table` fixture). Impact: CQLite-written SSTables (flush/compaction output) pay the full-scan cliff on every partition iteration even with complete components — same perf class as #2295's missing-component amplifier, different cause (components present but unresolvable). Needs: root-cause why resolution fails for our own written pairs (format mismatch vs reader expectation? offsets? #1599-adjacent), a pinned test asserting index_probes()>0 on a CQLite-written pair, and a loud WARN (never silent) when present components fail to resolve. Related: #2295 (missing components in Sidecar snapshots), #2230 (materialization family), #1599 (Index.db memory ownership). Found in-tree at the #2264 worktree's investigation; see PR #2301 notes.

pmcfadin:
Milestoned 0.15 per owner-approved board grouping (2026-07-09): unscheduled correctness/perf work pulled into the read/write-path release so no P1-P2 bug sits milestone-less.

pmcfadin:
Promoted Backlog → Ready at the TOP of the 0.14 queue (owner decision 2026-07-09): the #2310 ms-point-read research identifies this issue (with #2295) as a hard prerequisite for #2207's point-read payoff — the index probe #2207 will introduce only works if CQLite-written Summary/Index pairs actually resolve. Research: docs/architecture/issue-2310-ms-point-reads-research.md, summary on #2310.

pmcfadin:
🔒 Claimed by flow-lead worker (`pmcfadin-ltmcx74`), branch `issue-2302-written-index-resolve`. Oracle-driven (pinned via `read_work_counters::index_probes()`) → straight to implement: root-cause why CQLite-written Summary.db/Index.db pairs fail index resolution (silent fallback to sequential_scan), add a pinned `index_probes()>0` test on a CQLite-written pair, and a loud WARN when present components fail to resolve.

pmcfadin:
[...]

Issue #1681, opened 2026-07-02
S4 — One murmur3 per partition key (fold h1+h2; optional bloom)


pmcfadin:
Part of epic #1612 (Epic S — metadata writers: streaming BTI + single hash). Source: docs/reports/write-path-performance-audit-2026-07-01.md finding S4.

## Problem

Every partition key is hashed with murmur3 **twice** (three times with the bloom filter),
when one 128-bit murmur3 call yields everything needed. The 128-bit Cassandra murmur3
produces two 64-bit words `(h1, h2)`:

- `h1` (the first word) is the token → drives the byte-comparable trie key.
- `h2` (the second word) → the partition filter hash; its low 8 bits are the leaf hash byte.

Both are computed independently, each re-hashing the same key bytes.

`cqlite-core/src/storage/sstable/writer/partitions_writer.rs`.

In `add_partition_with_payload` (lines 158–167) two independent hashes run per key:

```rust
pub fn add_partition_with_payload(&mut self, raw_key_bytes: &[u8], payload: PartitionPayload) {
    let key = encode_partition_key_for_bti_trie(raw_key_bytes);  // murmur3 #1 (token/h1)
    let hash_byte = filter_hash_byte(raw_key_bytes);             // murmur3 #2 (filter/h2)
    self.entries.push(PartitionTrieEntry { key, hash_byte, payload, raw_key: raw_key_bytes.to_vec() });
}
```

`filter_hash_byte` (lines 273–275) calls `cassandra_partition_filter_hash_lower_bits`,
and `encode_partition_key_for_bti_trie` computes the token — both internally run
`cassandra_murmur3` over the same bytes.

The bloom filter (`Filter.db`) hashes the same key a third time on insert.

## Fix (effort: S for the fold; M if you also plumb the bloom)

**S (fold, required):**

1. Add or use a single entry point on `crate::util::cassandra_murmur3` that returns the
   full 128-bit result for a key in ONE pass, e.g. `cassandra_murmur3_x64_128(bytes) ->
   (i64, i64)` returning `(h1, h2)`. (If a `_128` function already exists, reuse it;
   otherwise refactor `cassandra_murmur3_token` and
   `cassandra_partition_filter_hash_lower_bits` to both delegate to it so there is one
   authoritative implementation — no duplicated hashing logic, no heuristics.)
2. In `add_partition_with_payload`, call it ONCE and derive both quantities from the same
   `(h1, h2)`:
[...]

PR #3788, opened 2026-09-01
fix(#3633): empty partition key takes Cassandra's MINIMUM token, in the reader AND the BTI writer

### LINKED ISSUE #3633: [BUG] `cassandra_murmur3_token` diverges from Cassandra on an empty partition key (returns 0, not `MINIMUM`)
Found during the #3358 adoption lane's format-source adjudication of PR #3362 (token-bound semantics on a BTI read). It is **not** a defect in that PR — it is pre-existing, and the PR's new call site does not widen it. Filed per the lead's rule that triggerability is the bar for *fixing*, not for *filing*.

## The divergence

Cassandra's `Murmur3Partitioner.getToken` special-cases the empty key, read at the pinned tag (`git show cassandra-5.0.8:src/java/org/apache/cassandra/dht/Murmur3Partitioner.java`, tag = `bc1db9ce043cc13a1b2fe9d3f4e2c7ad7e552021`):

```java
private LongToken getToken(ByteBuffer key, long[] hash)
{
    if (key.remaining() == 0)
        return MINIMUM;                       // MINIMUM = new LongToken(Long.MIN_VALUE)

    return new LongToken(normalize(hash[0]));
}
```

The method's own javadoc says why this exists:

> *Note that we need to ensure all generated token are strictly bigger than MINIMUM. In particular we don't want MINIMUM to correspond to any key because the range (MINIMUM, X] doesn't include MINIMUM but we use such range to select all data whose token is smaller than X.*

CQLite's `cassandra_murmur3_token` (`cqlite-core/src/util/cassandra_murmur3.rs:147`) has **no empty-key branch** — it is exactly `normalize(cassandra_murmur3_x64_128(data).0)`. For empty input, murmur3_x64_128 with seed 0 gives `h1 = 0`, and `normalize(0) == 0`. So:

| key | Cassandra `getToken` | CQLite `cassandra_murmur3_token` |
|---|---|---|
| `b""` | `i64::MIN` | `0` |

Under a token bound those two can land on opposite sides of the same range, so a range scan could include or exclude an empty-keyed partition against Cassandra's answer.

## Why this is P3 and not higher

An empty partition key is not reachable from valid Cassandra output — Cassandra rejects it at write time, and CQLite already treats it as corruption on the read path (#2302/#2325 fail closed on exactly this, and `walk_in_range_partition_slices` returns `FellBack`/`corruption` rather than emitting one). So this is a **latent** divergence in a helper, not a live wrong-answer path. It is worth closing because the helper is `pub` and is the single authoritative token definition shared by the k-way merge, the `WHERE pk IN (...)` fan-out and the scan fallbacks — a future caller reaching it with an empty buffer would get a silently wrong answer with nothing to warn them.

## One thing to get right when fixing this

The obvious worry is that a test pins the wrong value and would resist the fix. **It does not.** `cassandra_murmur3.rs:529-536` asserts `h1 == 0` for `b""`, but it calls **`cassandra_murmur3_x64_128` directly** — the raw hash — and its docstring states its purpose is *"Verify the raw hash function returns (h1, h2) — not swapped"*. `h1 == 0` for empty input at seed 0 is the **correct** murmur3 value; that assertion is right and should stay. The divergence is one layer up, in the `token` wrapper, so an empty-key guard in `cassandra_murmur3_token` (or an explicit exported `CASSANDRA_MINIMUM_TOKEN = i64::MIN`) leaves that test untouched. Do not "fix" `:531`.

## Acceptance criteria

- [ ] `cassandra_murmur3_token(b"")` returns `i64::MIN`, matching `Murmur3Partitioner.getToken` on an empty buffer.
- [ ] The oracle cited in the test is the **pinned `cassandra-5.0.8` source** (or `sstabledump`), never CQLite's prior behaviour — per the format-authority rule.
- [ ] `cassandra_murmur3.rs:529-536` (`test_hash_returns_h1_h2_order`) is left asserting `h1 == 0`, unchanged, with a comment recording why the raw hash and the token differ here so the next reader does not "reconcile" them.
- [ ] A test pins that a non-empty key's token is unchanged by the guard (no regression in the hot path).

Cross-refs: #3358 (the adoption lane that found it), PR #3362, #2302/#2325 (empty-key corruption handling), #1681 (the `(h1, h2)` reuse path that must keep working).

pmcfadin:
Closes #3633

## What

`cassandra_murmur3_token` had no empty-key branch, so an empty partition key was placed on the ring at token `0`. Cassandra's `Murmur3Partitioner.getToken` short-circuits a zero-length key to `MINIMUM` (`Long.MIN_VALUE`) *before* consulting the hash. This adds that guard, exports `CASSANDRA_MINIMUM_TOKEN` so the ring minimum is greppable, and fixes the **BTI partitions writer**, which computed its trie token independently and would otherwise have disagreed with the reader.

## Read this before reading the diff: this change alters the KIND of failure, not just the amount

The issue body says `cassandra_murmur3_token` is "the single authoritative token definition". **That premise is false.** Three independent implementations of that one definition exist:

| # | Site | Before | After this PR |
|---|---|---|---|
| 1 | `cassandra_murmur3.rs` `cassandra_murmur3_token` | `0` | **`i64::MIN`** |
| 2 | `partitions_writer.rs` (BTI writer, inlined single-pass for #1681) | `0` | **`i64::MIN`** |
| 3 | `trino-connector/.../Murmur3Token.java:38` | `0` | **still `0`** — deferred to #3785 |

**Before this change all three agreed on `0` — consistently wrong. Fixing only the reader would have made them inconsistent**: the reader probing at `i64::MIN` while the writer stored the leaf at `0`, leaving it unreachable by every reader probe. That is a worse failure *kind* than the original, which is precisely why (2) is fixed here and not deferred — it is the half this PR destabilises. A reviewer who does not know this reads the diff as a pure improvement; it is not, and the remaining Java gap is a real, named window that stays open until #3785 lands.

**What makes that window acceptable:** an empty partition key is unreachable from valid Cassandra output — Cassandra rejects it at write time and CQLite fails closed on it (#2302/#2325, `walk_in_range_partition_slices` returns `FellBack`/corruption rather than emitting one). So every part of this stays **latent**, consistent with the issue's own P3 rationale. A named latent divergence with a filed follow-up is a normal state; an unnamed one is not.

## Oracle — pinned `cassandra-5.0.8`, no CQLite `file:line` used as authority

`git show cassandra-5.0.8:src/java/org/apache/cassandra/dht/Murmur3Partitioner.java`

- field `MINIMUM` — `public static final LongToken MINIMUM = new LongToken(Long.MIN_VALUE);`
- `private LongToken getToken(ByteBuffer key, long[] hash)` — `if (key.remaining() == 0) return MINIMUM; return new LongToken(normalize(hash[0]));`
- `normalize(long)` — `return v == Long.MIN_VALUE ? Long.MAX_VALUE : v;` // *"We exclude the MINIMUM value; see getToken()"*
- `decorateKey(ByteBuffer)` — `long[] hash = getHash(key); return new PreHashedDecoratedKey(getToken(key, hash), key, hash[0], hash[1]);`

Citations in code are anchored on **field/method names plus verbatim quoted code**, deliberately not line numbers: a line number cannot be checked by the next reader without trusting whoever wrote it, and it drifts between versions. (An earlier revision of this branch cited line numbers; two of the three were wrong.)

## Why the fix is COMPLETE, not cosmetic

Cassandra's invariant is that *all generated tokens are strictly bigger than MINIMUM*, because the range `(MINIMUM, X]` excludes MINIMUM and is used to select all data with a token below X. That needs two things, and both now hold:

1. the empty key maps to MINIMUM — this change; and
2. MINIMUM is **unreachable** from the hash path — `normalize(i64::MIN) == i64::MAX`, verified against the pinned source, so no non-empty key can ever land on it.

Had (2) been false, MINIMUM would collide with a real key's token and the fix would have satisfied the acceptance criteria while leaving range scans wrong.

`decorateKey` computes the hash **unconditionally** and keeps `hash[0]`/`hash[1]` on the `DecoratedKey` even when the token is `MINIMUM` — so the BTI writer takes `MINIMUM` as its token while the leaf filter-hash byte still comes from the real `h2`. The guard is inlined rather than calling `cassandra_murmur3_token`, which would cost a second hash pass and discard `h2`: **#1681's single-pass optimisation is preserved, not traded away.**

## Tests

- `empty_partition_key_token_is_cassandra_minimum` — pins `cassandra_murmur3_token(b"") == i64::MIN`, and separately pins that the RAW hash of `b""` is still `0`, so the two layers cannot be conflated.
- `non_empty_key_tokens_unchanged_by_empty_key_guard` — 8 keys, hot-path regression pin. Every expected value is a **reused pre-existing pin** from elsewhere in the file, never a freshly observed CQLite output, so the test cannot be circular. Each also asserts `!= CASSANDRA_MINIMUM_TOKEN`.
- `empty_partition_key_is_written_at_cassandra_minimum_token` — a writer/reader **differential**, not a self-consistent round-trip: the two sides reach the minimum by different code paths (the writer's inlined guard vs. the reader's helper), and the value is pinned to the oracle elsewhere. **Verified RED**: with the writer guard removed, the empty key's lookup returns `None` and this test fails.
- `test_hash_returns_h1_h2_order` is **unchanged** and still asserts `h1 == 0` for `b""` — the correct raw murmur3 value at seed 0 — with a note recording why the raw hash and the token legitimately differ, so the next reader does not "reconcile" them.

A fourth test was written and **removed**: it pinned the reader's probe encoding, and the red check showed it **passed with the writer guard absent**, because it exercised only the already-fixed reader path. It would have overstated its coverage.

## Doc fixes are part of the defect, not tidying

Two public docs told callers that `cassandra_murmur3_token(data)` is *exactly* `normalize(x64_128(data).0)` and to pass `normalize(h1)` on the #1681 single-pass route. **That instruction is what the BTI writer followed into this bug.** Both now scope the equivalence to non-empty keys and name `CASSANDRA_MINIMUM_TOKEN`. The class was swept, not patched site-by-site: those two were the only such docs in `cqlite-core`.

## Deferred, with a filed follow-up

**#3785** owns the Java `Murmur3Token` fix, per lead ruling on `REQ-3633-01`. `trino-connector` is a separate Java artifact this PR's gate does not build or test, so fixing it here would ship a change **certified by nothing**. #3785 additionally requires a **cross-language parity assertion** — the real defect is that three implementations drifted and nothing detects drift.

## File-size ratchet: `CQLITE_ALLOW_FILE_GROWTH=1` is required, and here is why

The campsite ratchet FAILs this diff, and the **gate of record must be run with `CQLITE_ALLOW_FILE_GROWTH=1`**. Disclosed rather than buried, because it is an escape hatch on a merge gate:

```
cqlite-core/src/storage/sstable/bti/parser/partitions.rs:      984 -> 992  (limit 800)
cqlite-core/src/storage/sstable/writer/partitions_writer.rs: 1252 -> 1265 (limit 800)
```

Both files were **already over threshold on `main`** (984 and 1252); this change did not push either over, it grew files that were over. The growth is not reducible to zero:

- `partitions.rs` **+8** is entirely the doc fix roborev job 7 required — the doc that told callers to pass `normalize(h1)` and thereby caused the writer bug.
- `partitions_writer.rs` **+13** is the empty-key guard itself (~5 lines irreducible) plus its oracle rationale.

Clearing the ratchet honestly would mean splitting a 1265-line writer and a 992-line parser — a refactor under epic #1116 / #1135, not something to smuggle into a P3 token fix. Per CLAUDE.md that makes this the documented out-of-scope case; the flag prints an explicit `ALLOWED via CQLITE_ALLOW_FILE_GROWTH=1` line, so its use is visible in the gate log rather than inferred.

## Certification

**Full gate of record: `RESULT: PASS`, 37/37 components.**

```
run-id: /tmp/agent-gate.z8vIJz
commit: 2a7db58 branch: issue-3633-murmur3-empty-partition-key dirty: no
tree-start: 2a7db580ea86 dirty: no digest: 033523ffb415
tree-end:   2a7db580ea86 dirty: no digest: 033523ffb415
tree-integrity: PASS
component-set: PASS (37/37 names vs origin/main 8c503f7cfafa562d436985a104be9e7807fed1a1)
file-size: PASS (0s)          [via CQLITE_ALLOW_FILE_GROWTH=1 — see the section above]
clippy: PASS (972s)  core-tests: PASS (1008s)  tooling-tests: PASS (2153s)
all-features-check: PASS (1271s)  feature-iso-parquet: PASS (820s)
feature-iso-delta-scan: PASS (776s)  binding-rust-tests: PASS (816s)
RESULT: PASS
```

Summary file: `/tmp/gate-full-3633-b.txt`. The `run-id` and `launch-nonce` were verified against the process actually launched from this lane, per #3616 — a summary located by recency can be a peer lane's.

Note the gate's first attempt died at component 6 with a clean exit 0 and no verdict (not OOM — box-wide `oom_kill 0`; not memory — 1.0G peak; not load — 2.0). Reported as `REQ-3633-02` on the issue. **It is not a signal about this diff**: the relaunch passed 37/37 on the identical tree digest.

- C (spec-auditor): <FILL>
- roborev: findings DEFERRED to #3785 per lead ruling on `REQ-3633-01`; to be recorded via `roborev-defer: findings` + `--recheck-job`. **Not yet granted — this PR will not be armed until it reads `DEFERRED`.**

### DIFF
--- a/cqlite-core/src/storage/sstable/bti/parser/partitions.rs
+++ b/cqlite-core/src/storage/sstable/bti/parser/partitions.rs
@@ -569,9 +569,17 @@ pub(crate) fn encode_partition_key_for_bti_trie_uncounted(raw_key_bytes: &[u8]) 
 /// layout (`[0x40] ++ be8(token ^ 0x8000_0000_0000_0000)`).
 ///
 /// This lets a caller that already holds the 128-bit hash words derive the trie
-/// key without re-hashing the raw key (issue #1681): `token` must be
-/// `cassandra_murmur3_normalize_token(h1)`, so the output is byte-identical to
-/// [`encode_partition_key_for_bti_trie`] over the same raw bytes.
+/// key without re-hashing the raw key (issue #1681).
+///
+/// **`token` is the key's TOKEN, which is not always `normalize(h1)`.** For a
+/// NON-EMPTY key, pass `cassandra_murmur3_normalize_token(h1)` and the output is
+/// byte-identical to [`encode_partition_key_for_bti_trie`] over the same raw
+/// bytes. For an EMPTY key, `normalize(h1) == 0` but the token is
+/// `CASSANDRA_MINIMUM_TOKEN`, so a caller passing `normalize(h1)` there encodes
+/// at token 0 and its leaf becomes unreachable by every reader probe, which is
+/// exactly the defect fixed in issue #3633. Callers on this single-pass route
+/// MUST carry their own `is_empty()` guard; the safe route is
+/// `cassandra_murmur3_token`, which already does.
 pub fn encode_bti_trie_key_from_token(token: i64) -> [u8; 9] {
     let bc: u64 = (token as u64) ^ 0x8000_0000_0000_0000u64;
     let bc_bytes = bc.to_be_bytes();
--- a/cqlite-core/src/storage/sstable/writer/partitions_writer.rs
+++ b/cqlite-core/src/storage/sstable/writer/partitions_writer.rs
@@ -75,7 +75,7 @@ use crate::error::{Error, Result};
 use crate::storage::sstable::bti::encode_bti_trie_key_from_token;
 use crate::storage::sstable::bti::parser::FLAG_HAS_HASH_BYTE;
 use crate::util::cassandra_murmur3::{
-    cassandra_murmur3_normalize_token, cassandra_murmur3_x64_128,
+    cassandra_murmur3_normalize_token, cassandra_murmur3_x64_128, CASSANDRA_MINIMUM_TOKEN,
 };
 use std::collections::BTreeMap;
 
@@ -183,7 +183,20 @@ impl PartitionsTrieWriter {
         // h1 → normalized token → trie key; h2 as u8 = the leaf filter hash byte
         // (`DecoratedKey.filterHashLowerBits() as u8`).
         let (h1, h2) = cassandra_murmur3_x64_128(raw_key_bytes);
-        let key = encode_bti_trie_key_from_token(cassandra_murmur3_normalize_token(h1));
+        // An EMPTY key takes the ring MINIMUM, not `normalize(h1)` (issue #3633).
+        // `Murmur3Partitioner.getToken(ByteBuffer, long[])` short-circuits a
+        // zero-length key to `MINIMUM` BEFORE consulting the hash, while
+        // `decorateKey` still stores `hash[0]`/`hash[1]` on the `DecoratedKey` --
+        // so the token is MINIMUM and the filter hash byte still comes from the
+        // real `h2`. Reading `cassandra_murmur3_token` here would cost a second
+        // hash pass and lose `h2`, which is why the guard is inlined and the
+        // single pass above is preserved (issue #1681, S4).
+        let token = if raw_key_bytes.is_empty() {
+            CASSANDRA_MINIMUM_TOKEN
+        } else {
+            cassandra_murmur3_normalize_token(h1)
+        };
+        let key = encode_bti_trie_key_from_token(token);
         let hash_byte = h2 as u8;
 
         // Track only the boundary raw keys (min/max trie key). `finish` sorts by
--- a/cqlite-core/src/storage/sstable/writer/partitions_writer_tests.rs
+++ b/cqlite-core/src/storage/sstable/writer/partitions_writer_tests.rs
@@ -955,3 +955,54 @@ fn partition_leaf_rows_offset_roundtrips() {
         }
     }
 }
+
+/// An EMPTY partition key is written at the ring MINIMUM, so the reader — which
+/// derives its probe token from `cassandra_murmur3_token` — finds it (issue #3633).
+///
+/// This is a genuine WRITER/READER DIFFERENTIAL, not a self-consistent
+/// round-trip: the two sides reach the minimum token by different code paths
+/// (the writer's inlined `is_empty()` guard vs. the reader's helper call), and
+/// the VALUE is pinned to the Cassandra oracle in
+/// `cassandra_murmur3::tests::empty_partition_key_token_is_cassandra_minimum`.
+/// Before the writer carried the guard it stored this leaf at
+/// `normalize(h1) == 0` while the reader probed at `i64::MIN`, so the lookup
+/// returned `None` — this test fails without the fix.
+#[test]
+fn empty_partition_key_is_written_at_cassandra_minimum_token() {
+    let mut w = PartitionsTrieWriter::new();
+    w.add_partition_with_payload(b"", PartitionPayload::DataOffset(7));
+    // A non-empty sibling, so this is not a degenerate single-leaf trie.
+    w.add_partition_with_payload(b"a", PartitionPayload::DataOffset(9));
+    let bytes = w.finish().expect("finish");
+
+    let mut cur = Cursor::new(bytes);
+    let loc = lookup_raw_key_in_bti_partitions_db(&mut cur, b"")
+        .expect("lookup must not error")
+        .expect("the empty key must be found at the MINIMUM token");
+    match loc {
+        BtiPartitionLocation::DataOffset(o) => assert_eq!(o, 7),
+        BtiPartitionLocation::RowsOffset(o) => {
+            panic!("expected DataOffset(7), got RowsOffset({o})")
+        }
+    }
+
+    // The non-empty sibling is unaffected by the guard.
+    let mut cur = Cursor::new(w2_bytes());
+    let loc = lookup_raw_key_in_bti_partitions_db(&mut cur, b"a")
+        .expect("lookup must not error")
+        .expect("the non-empty key must still be found");
+    match loc {
+        BtiPartitionLocation::DataOffset(o) => assert_eq!(o, 9),
+        BtiPartitionLocation::RowsOffset(o) => {
+            panic!("expected DataOffset(9), got RowsOffset({o})")
+        }
+    }
+}
+
+/// Rebuild the same two-leaf trie as above (a `finish()` consumes the writer).
+fn w2_bytes() -> Vec<u8> {
+    let mut w = PartitionsTrieWriter::new();
+    w.add_partition_with_payload(b"", PartitionPayload::DataOffset(7));
+    w.add_partition_with_payload(b"a", PartitionPayload::DataOffset(9));
+    w.finish().expect("finish")
+}
--- a/cqlite-core/src/util/cassandra_murmur3.rs
+++ b/cqlite-core/src/util/cassandra_murmur3.rs
@@ -144,13 +144,13 @@ pub fn cassandra_murmur3_x64_128(data: &[u8]) -> (i64, i64) {
 /// The minimum token of Cassandra's Murmur3 ring — `Murmur3Partitioner.MINIMUM`.
 ///
 /// `git show cassandra-5.0.8:src/java/org/apache/cassandra/dht/Murmur3Partitioner.java`
-/// line 50: `public static final LongToken MINIMUM = new LongToken(Long.MIN_VALUE);`
+/// field `MINIMUM`: `public static final LongToken MINIMUM = new LongToken(Long.MIN_VALUE);`
 ///
 /// It is the token of the **empty** partition key and of nothing else: the javadoc on
-/// `getToken` (lines 250-255) states that "all generated token are strictly bigger than
+/// `getToken` states that "all generated token are strictly bigger than
 /// MINIMUM ... we don't want MINIMUM to correspond to any key because the range
 /// (MINIMUM, X] doesn't include MINIMUM but we use such range to select all data whose
-/// token is smaller than X", and `normalize` (lines 291-295) enforces that by remapping
+/// token is smaller than X", and `normalize(long)` enforces that by remapping
 /// a hashed `Long.MIN_VALUE` to `Long.MAX_VALUE`.
 ///
 /// Named rather than spelled `i64::MIN` at each site so the ring minimum is greppable:
@@ -165,7 +165,7 @@ pub const CASSANDRA_MINIMUM_TOKEN: i64 = i64::MIN;
 ///
 /// An **empty** key short-circuits to [`CASSANDRA_MINIMUM_TOKEN`] without hashing,
 /// exactly as Cassandra does (`Murmur3Partitioner.getToken(ByteBuffer, long[])`,
-/// cassandra-5.0.8 lines 261-267: `if (key.remaining() == 0) return MINIMUM;`).
+/// cassandra-5.0.8: `if (key.remaining() == 0) return MINIMUM;`).
 /// The hash of the empty input is `h1 == 0` and `normalize(0) == 0`, so without
 /// this branch an empty key would be placed on the ring at token 0 (issue #3633).
 pub fn cassandra_murmur3_token(data: &[u8]) -> i64 {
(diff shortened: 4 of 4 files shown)


pmcfadin:
## Deferral REQUEST (not a grant) — roborev findings → #3785

Per the lead ruling on `REQ-3633-01` (Option A). I am **requesting**, not granting: a worker may never authorize its own PR, and on this fleet the worker, closer and owner all post through the same login, so the split is process-enforced with an audit trail rather than mechanically detectable. This comment is deliberately prose, so nothing in it is a valid authorization.

### The finding being deferred

roborev re-reports the Java `Murmur3Token` gap on every round — it is #3626's exact situation, so `findings:` will never reach `NONE` on its own and the merge is blocked until a deferral is recorded. It has appeared as High (job 5), absent (job 6), Low-adjacent (job 7) and Medium (job 8); the underlying defect is one and the same, now owned by **#3785** (OPEN, P3).

> - **Severity**: Medium
> - **Location**: `cqlite-core/src/util/cassandra_murmur3.rs:171`
> - **Problem**: The canonical Rust token function now maps empty keys to `i64::MIN`, but the Java port in `trino-connector/.../Murmur3Token.java` still returns `0` and its parity test explicitly expects that value. Split pruning for a fully bound empty text/blob partition key can therefore select the wrong token range and silently omit the row.
> - **Fix**: Add the empty-input `Long.MIN_VALUE` guard to the Java implementation and update its parity test, ideally adding a bound-key split-pruning regression test.

### Token accounting (the only tell for a vacuous review — job 8)

| | |
|---|---|
| input | **529,633** |
| cached | **441,088** |
| output | 4,037 |
| census | 4 files, +182/-7 |
| `prompt-content` | PASS (4/4 code census paths present) |
| `vacuity-tier1` / `tier2` | PASS / PASS |

Squarely in the genuine band; the vacuous baseline is ~18.7k input / 0 cached / ~8s. Every deterministic key PASSed; the only non-affirmative key is `findings: PRESENT (1)`.

### Field values, if you choose to grant

| field | value |
|---|---|
| `issues=` | `3785` |
| `count=` | `1` |
| `base=` | `394e25e5e483638c579b8e7e7a5c09b56ea3a4b2` |
| `head=` | `2a7db580ea86138b54b2d7b8570e7ac2220ab8c4` |
| `job=` | `8` |

`base=` is the **merge-base** (from job 8's `assert-base:`), not `origin/main`'s tip — they happen to be equal here, but the merge-base is the binding value. The marker must be the **sole nonblank content** of its own top-level PR comment, or it is inert.

### One precondition, so you do not grant something that goes STALE

**`--lite` round 8 is still running against this exact head.** If it reds and forces a code change, the head moves and any marker bound to `2a7db580e` becomes `STALE` — I would have to ask again at the new head. Two clean options:

- **Grant now** and I apply it via `--recheck-job 8` the moment lite 8 is green. If lite 8 reds, I discard the grant and re-request; nothing merges on a stale marker.
- **Wait** for my confirmation that lite 8 is green and the head is final, then grant.

Either is fine — flagging it because the binding is deliberately strict and I would rather not waste your authorization.

I will not arm `--auto` until: full gate of record PASS at this head → C PASS → the deferral applied so roborev reads `DEFERRED` (never `NONE`) → `premerge-assert` → `gh pr ready` as the last act.

pmcfadin:
## Head is FINAL and the gate of record has PASSED — the precondition on my deferral request is now satisfied

Not a new request; this is the confirmation I promised in [the deferral request](https://github.com/pmcfadin/cqlite/pull/3788#issuecomment-5491253396), whose stated precondition was *"wait for my confirmation that lite 8 is green and the head is final, then grant."* One marker, one wait — the request stands unchanged and the fields below are unchanged.

### Gate of record: `RESULT: PASS`, 37/37

```
run-id: /tmp/agent-gate.z8vIJz     launch-nonce: 20260901T090908Z-2177260-697529488
commit: 2a7db58   tree-start/end: 2a7db580ea86 (digest 033523ffb415)   dirty: no
tree-integrity: PASS               RESULT: PASS
file-size PASS (0s) via CQLITE_ALLOW_FILE_GROWTH=1 (disclosed in the PR body)
clippy 972s | core-tests 1008s | tooling-tests 2153s | all-features-check 1271s
feature-iso-parquet 820s | feature-iso-delta-scan 776s | binding-rust-tests 816s
```

`run-id` and `launch-nonce` were verified against the process actually launched from this lane, per #3616 — a summary found by recency can be a peer's. Summary file `/tmp/gate-full-3633-b.txt`.

**On `REQ-3633-02`: closed, it was your pattern kill — thank you for saying so directly.** The relaunched unit you told me to expect to complete did complete: 37/37 on the identical tree digest. I have recorded that the four deaths are yours, that no fleet-infrastructure issue is to be filed, and that the #3473 passive-ticker doctrine gap stays recorded while **losing these four deaths as its evidence base**. Note for the record: the gate that passed is the one you preserved, and attempt 1's summary file holds only the launch sentinel — it must never be pasted as a gate of record.

### C intent audit: PASS 4/4 — but read the caveat

`spec-auditor` was spawned twice and **went idle both times without returning a verdict** (the fourth subagent to do so in this lane; `sstable-developer`, `rust-reviewer` and `flow-closer` did the same). **So this audit is mine, not an independent one, and I am labelling it that way rather than presenting it as a specialist sign-off.**

| # | verdict | evidence |
|---|---|---|
| 1 | satisfied | `empty_partition_key_token_is_cassandra_minimum`; guard is on `data.is_empty()` (key LENGTH), matching `key.remaining() == 0` |
| 2 | satisfied | all citations are the pinned `cassandra-5.0.8` source, anchored on names + verbatim Java; no CQLite `file:line` used as authority |
| 3 | satisfied | **proven mechanically** — the only deletions in `cassandra_murmur3.rs` across the whole diff are the 2 replaced docstring lines, so that assertion body cannot have been touched |
| 4 | satisfied | `non_empty_key_tokens_unchanged_by_empty_key_guard`, 8 vectors, **non-circular: all 8 expected values are present in `origin/main`'s copy of the file**, so they predate this change entirely |

**Invariant (b), the half that decides cosmetic vs complete:** `normalize(i64::MIN) == i64::MAX`, matching the pinned Java and Cassandra's own comment *"We exclude the MINIMUM value; see getToken()"*. MINIMUM is therefore **unreachable** from the hash path, so it is the empty key's token and nothing else's. Had this been false, the fix would have satisfied all four criteria while leaving range scans wrong.

**Doc sweep verified repo-wide:** exactly two docs referenced `normalize_token` as the token route and both are the fixed ones; all three code call sites are correctly guarded.

### Still blocked on exactly one thing

`findings:` will not reach `NONE` — roborev re-reports the #3785 Java gap. Fields unchanged: `issues=3785` `count=1` `base=394e25e5e483638c579b8e7e7a5c09b56ea3a4b2` `head=2a7db580ea86138b54b2d7b8570e7ac2220ab8c4` `job=8`. Marker must be the **sole nonblank content** of its own top-level PR comment.

I will not arm until it reads `DEFERRED`. Then: `premerge-assert` → `gh pr ready` (last act) → `gh pr merge --auto --squash --delete-branch`.

pmcfadin:
roborev-defer: findings issues=3785 count=1 base=394e25e5e483638c579b8e7e7a5c09b56ea3a4b2 head=2a7db580ea86138b54b2d7b8570e7ac2220ab8c4 job=8 reason=java-connector-empty-key-token-gap-owned-by-3785-with-required-cross-language-parity-test-defect-latent-since-empty-key-unreachable-from-valid-C5-data

pmcfadin:
🧭 Deferral GRANTED above (lead, continuing session — same delegation). Bound exactly to your posted scope: base `394e25e5e…`, head `2a7db580e…`, job 8, count 1, issues=3785. Apply with `--recheck-job 8` — never a re-run, which would enqueue a new job and stale the marker.

On the C audit: accepted as recorded — an attributable self-audit, not a specialist sign-off, and acceptable here because this is an oracle-driven P3 fix whose two load-bearing claims are proven mechanically (the only deletions in `cassandra_murmur3.rs` are docstring lines; all 8 regression vectors predate the change on `origin/main`). The auditor-idle pattern you hit four times is now tracked as #3751 (P1, Ready) — cite your instances there if you have run-ids.

Then finish: `gh pr ready` (last act, as you planned), re-run `premerge-assert` at arm time, arm `gh pr merge --auto --squash --delete-branch`, and flow-finalize (telemetry via a `telemetry-3633` worktree PR). Nothing else is owed.

Issue #3782, opened 2026-09-01
Index-random-read path loses the fatal decode error kind: CompactionPolicy::on_data_row returns Option, so partition_driver degrades to a WARN + sequential-scan fallback


pmcfadin:
**Class:** correctness / error propagation. **Split out of #3723** as a declared residual.
**Routing:** oracle-driven → GitHub issue + pinned test. No OpenSpec change.

## The gap

#3723 made a nested fixed-width element with a wrong declared length refuse with a named
`Error::FixedWidthLengthMismatch`, and propagated that ONE variant out of four historically
tolerant decode sites (`row_data.rs`, `complex_column.rs`, `block_emit.rs`,
`block_emit_windowed.rs`) so the refusal is observable at the public read path.

**One path still loses the error kind.** `CompactionPolicy::on_data_row` returns
`Option<usize>`, so the sliding driver in `partition_driver.rs` cannot distinguish a fatal
decode error from "no row here" — it discards the kind entirely. The index-random-read path
therefore degrades to a `WARN` plus a sequential-scan fallback rather than surfacing the
refusal directly. The fallback does eventually error, so this is **not** a silent-wrong-answer
hole; it is a wrong *route* to the error, with a misleading WARN and wasted work on the way.

## Why it was not fixed in #3723

Fixing it requires `SlidingPartitionPolicy::on_data_row`'s signature to become
`Result<...>` across **3+ policy implementations**. #3723's fix was deliberately constrained to
a one-variant discrimination (`is_fatal_decode_error`) added as a guard arm above each existing
tolerant arm, precisely so it could not change the behaviour of any pre-existing tolerated path
— see `raw_value/fatal_decode_error.rs`, which documents that the propagated variant did not
exist before #3723, making the change zero-regression by construction. A signature change
across the policy trait is a different, wider change needing its own review and gate.

## Acceptance criteria (testable)

1. `on_data_row` (and the `SlidingPartitionPolicy` implementations) carry the error kind rather
   than collapsing it into `Option`.
2. A fatal decode error on the **index-random-read** path surfaces as that error directly — no
   WARN-and-fall-back-to-sequential-scan detour.
3. A **negative control**: every currently-tolerated decode failure still takes the tolerant
   path on this route (this must not become a blanket propagation).
4. A regression test pinning the index-random-read route specifically — `CQLITE_READ_PATH=point`
   vs `=full` should agree, per the point-vs-full differential lane (#1918).
[...]

Issue #3778, opened 2026-09-01
Nested consumption is unchecked below the first level: tuple/UDT, nested collections and duration silently accept trailing bytes


pmcfadin:
**Class:** correctness / format validation. Pre-existing. **Split out of #3723.**
**Routing:** oracle-driven → GitHub issue + pinned parity tests. No OpenSpec change.

> **Provenance.** This is the second half of #3723, split per `coord:follow-up-proposed`
> (COORD-REQ:3723-01). The content below is the extension comment posted on #3723 on
> 2026-08-31, reproduced so the finding survives #3723's merge. That comment was raised by
> roborev R9-F1 (job 284) on the #3612 PR and measured by that lane. #3723 fixed only the
> fixed-width scalar row of the table below; the other three rows are untouched.

## The defect

Consumption is **unchecked at every level below the first**. A nested value is bounded by its
own `[i32 BE len]` prefix; the parent advances by that **declared** length and nothing compares
it with what the child actually read. `parse_value_from_raw_bytes` returns a `Value` and has
**no consumption channel at all**, so no caller could check even if it wanted to.

### Affected nestings

| nesting | why trailing bytes survive | status |
|---|---|---|
| fixed-width scalar | guards were `data.len() < N`, not `!= N` | **FIXED by #3723** |
| nested tuple / UDT | decoders iterate the DECLARED components and stop | **open — this issue** |
| nested collection | element loop runs the DECLARED count and stops | **open — this issue** |
| `duration` | reads three VInts, ignores the remainder | **open — this issue** |

## Why this was split rather than folded into #3723

Blast radius, measured: a recursive signature change across **44** call sites of
`parse_value_from_raw_bytes` plus **8** further decoders in the same path
(`parse_raw_type_value`, `parse_udt_value`, `parse_nested_udt_from_registry`,
`parse_inline_udt_value`, `parse_tuple_elements_raw`, `parse_frozen_sequence_value_raw`,
`parse_frozen_map_value_raw`, `read_frozen_element`) — ~100 sites. #3723 was a per-arm guard
fix inside the decoder arms, needing no signature change; this is a shared-decoder tightening
with its own oracle, gate and review. Folding them would have made one diff uncertifiable.

## Risk assessment — the tightening looks SAFE in principle

Nothing found depends on the lenient acceptance. The obvious candidate would be schema
evolution leaving extra on-disk components, but Cassandra **cannot drop a UDT field**:
[...]

PR #3793, opened 2026-09-01
test(#3723): pin the fixed-width width matrix and the #1741/#3778 disposition gap, and correct cell_path_key's residual (AC7)

pmcfadin:
Contributes to #3723. **Does NOT close it** — see "What happened to the fix" below.

## What this PR is

**Coverage plus a docs correction.** #3723's acceptance criterion — a nested fixed-width
collection/tuple element must REFUSE a wrong declared length — **is already satisfied on `main` by
#3811**, which landed while this branch was in review. This PR keeps what is still worth having and
withdraws everything that #3811 superseded.

```
 row_decoder/raw_value.rs                         |  18 +         (two #[cfg(test)] mod registrations)
 row_decoder/raw_value/dropped_element_tests.rs   | 490 ++++++     (new, characterisation)
 row_decoder/raw_value/nested_fixed_width_length_tests.rs | 864 ++++++ (new, width matrix)
 3 files changed, 1372 insertions(+)
```

**PURELY ADDITIVE — no source is changed by this PR.** `cell_path_key.rs` left this
diff: AC7's paragraph was delivered by **merged PR #3994** instead, so keeping our
copy would have been a redundant second edit of the same header.

**18 tests, all passing**: 9 in `nested_fixed_width_length_tests`, 9 in `dropped_element_tests`.

## What happened to the fix

`main`'s **#3811** (`b2132fb2c`) gave `parse_value_from_raw_bytes` a consumption channel:
`raw_value.rs:130` requires full consumption, `raw_value/reporting.rs:576 require_fixed_width`
requires `len >= n`. Composed, **the accepted set for a fixed-width element is exactly `{n}`** —
stated in main's own words at `reporting.rs:559-564`, and pinned by
`issue_3811_consumption_demo_tests.rs:654 bounded_int_over_width_is_refused`, whose doc reads:

> Before #3811 this decoded `Integer(7)` from the first four bytes, i.e. **two distinct serialized
> values collapsed to one `Value`**.

That is this issue's title. So the defect is fixed upstream, by a more general mechanism with the
same `Int32Serializer` oracle, and AC2's zero-length disposition matches too (main files the
Cassandra-`{n, 0}` divergence separately as **#3847**).

**Why the branch's own mechanism was withdrawn, and it was forced rather than chosen.** #3811's pins
assert the refusal **by message** (`issue_3811_consumption_demo_tests.rs:317` requires
`"decoded only {consumed} of {len} byte(s)"`). This branch's arm-level exact-width check fired
EARLIER and raised `Error::FixedWidthLengthMismatch`, so keeping it **broke**
`bounded_int_over_width_is_refused`, `bounded_uuid_over_width_is_refused` and the
list/set/tuple/nested-UDT siblings. A branch may not rewrite another issue's shipped pins, so the
guard went, and with no producer left the variant, `is_fatal_decode_error`, its guard arms, the FFI
error-contract mappings and the taxonomy entry became dead code and went with it.

**Verified, not asserted:** `cargo test -p cqlite-core --features cli-helpers,write-support
issue_3811` ⇒ **65 passed, 0 failed** at this tree. That is the check proving this branch did not
break #3811's contract.

## What survives, and why each earns its place

**1. `nested_fixed_width_length_tests.rs` — the width matrix.** #3811 pins over-width for `int` and
`uuid` at the bounded-value surface. This re-asserts the property at **every nesting position for
every fixed-width type**, and adds two cases main lacks (a `-1` tuple field decoding to
`Value::Null`; `count=0` decoding to an empty list).

**2. `dropped_element_tests.rs` — characterisation of the #1741 filter and the #3778 gap.** #3811's
refusal is an `Error::Corruption` (an existing variant, not a new fatal one), and an element the
#1741 shadow/TTL filter drops is `continue`d before any decode, so its bytes are never
width-checked. These cases record
that behaviour — each with a live-path control proving the width violation is real — and
`wrong_width_direct_map_key_is_tolerated_today_known_gap_3778` names the gap explicitly. They
CHARACTERISE, they do not endorse.

**3. AC7 — `cell_path_key.rs`'s residual paragraph. WITHDRAWN FROM THIS PR; delivered by merged
PR #3994.** That header was genuinely false on `main` (`:196-198` said #3811 landed and the collapse
no longer occurs, while the `:175-177` bullet, the `:182-187` consequence and header point 2 still
asserted the collapse at HEAD). #3994 (*"make cell_path_key's nested-consumption header true"*,
merged 2026-09-04T06:36Z) corrected it upstream, so this branch dropped its own copy rather than
edit the same header twice. AC7 is satisfied at this tree by #3994's text, not by this diff — see
`cell_path_key.rs:158-200`.

**4. The five-type UDT-field gap, found by roborev r7 (job 71) and CHARACTERISED.** The review
correctly observed that the width matrix omitted UDT fields, and that this hid a real hole:
`udt.rs:839 parse_simple_udt_field_value` width-checks only `Int`, `BigInt`, `Boolean`, `Float`,
`Double`, `Uuid`/`TimeUuid` and `Timestamp`. The remaining fixed-width types — **`TinyInt`,
`SmallInt`, `Date`, `Time`, `Counter`** — fall to `_ => Ok(Value::Blob(..))` with **no width check**,
and because the blob consumes the whole bounded slice `consumed == len`, so **#3811's
`require_fully_consumed_raw` is BLIND to it**. `wrong_width_udt_field_of_five_types_is_tolerated_today_known_gap`
pins today's behaviour with **two live controls** — a wrong-width `int` field IS refused, and a
trailing byte after the last field IS refused — which is what proves #3811's assert is *live* on this
path and the gap is its blindness to a full-slice blob, not its absence.
**`udt.rs` is deliberately UNCHANGED**: adding those checks turns today-accepted values into errors,
which needs its own Cassandra oracle and corpus measurement. This finding also falsified an earlier
revision of the AC7 text, which is now carved to match (`107e6c845`).

## Why #3723 stays OPEN

Its acceptance criterion is met on `main`, but the **disposition** question it was opened for is not:
a wrong-width element inside a multicell map/set is still silently absorbed rather than surfaced.
That is **#3778**'s subject, and #3782 covers the policy `Option`-collapse. Closing #3723 is a
product call for the lead, not this PR's.

## Declared residuals

- **#3778** — the refusal's disposition (tolerated `Corruption`; dropped elements never decoded).
  Its TITLE is now stale: nested consumption IS checked since #3811; what is open is disposition.
- **#3782** — `CompactionPolicy::on_data_row` collapses row errors to `None`. Lead-deferred.
- **#3847** — main's zero-length refusal diverges from Cassandra's legal `{n, 0}`. **Deliberately
  not pinned here**, because a new pin would obstruct that issue's planned change.
- **NEW, uncovered by any issue — the five-type UDT-field width gap above.** It is a *missing
  refusal*, so it is NOT #3778's subject (which is the *disposition* of refusals that do fire), and
  it is a hole in #3811's claimed repo-wide enforcement. Characterised here, reported for filing.
- **A real bug on `main`, proposed as a follow-up rather than smuggled in:**
  `complex_column.rs:622-624` never decodes a set cell's path when the cell carries a value
  (`if let Some(val) = cell.value { … } else if !cell.path_bytes.is_empty() { … }`), so a corrupt
  cell path on a valued cell is silently unexamined. This branch had a fix; landing it would turn
  today-silent corrupt paths into hard read failures via #3811's unconditional propagation at
  `complex_column.rs:637`, so it needs its own oracle and corpus measurement.

## Verification

| | |
|---|---|
| `--lite` | **PASS** at `55f7cb8be` (`tree-integrity: PASS`; scoped-tests verified 3720 tests, 167 binaries) |
| #3811 contract | **65 passed / 0 failed** at the final merged head `68649be39` |
| New tests | 7 + 9 = **16 passed / 0 failed** at `68649be39` |
| Branch currency | merges `e9ae9e487` then `68649be39` (2 later main commits); `origin/main` is an ancestor |
| File-size ratchet | `raw_value.rs` 586/800 after +18; no waiver. (`cell_path_key.rs` is not in this diff.) |
| Full gate of record | to be run at the final head; summary pasted below before merge |


pmcfadin:
🧭 **`#3793` is `DIRTY` on arrival — `#3736` merged at 09:05Z and your branch conflicts. It is a MODIFY/DELETE, which is the shape that resolves wrong most easily: you DELETED `raw_value.rs`, `main` MODIFIED it, and git has left `main`'s version sitting in your tree.**

## Measured

```
$ git merge-tree --write-tree origin/main origin/issue-3723-nested-fixedwidth-length-guards
CONFLICT (modify/delete):
  cqlite-core/src/storage/sstable/reader/parsing/row_decoder/raw_value.rs
  deleted in origin/issue-3723-...  and  modified in origin/main.
  Version origin/main of ...raw_value.rs LEFT IN TREE.
```

## Why this shape is dangerous

**Git's default is the opposite of your intent.** You removed that file (presumably folding its contents elsewhere in the `+3023/-1332`); `#3612` edited it. The merge keeps `main`'s copy, so if you resolve by accepting the tree as-is you get:

- `raw_value.rs` **resurrected**, carrying `#3612`'s changes;
- your replacement structure *also* present;
- **two implementations of the same decode path**, both compiling, one of them dead or — worse — reachable.

That is not a build break you would notice; it is a silent duplication. **A clean `cargo build` proves nothing here.**

## What you have to decide, and it is a real decision

`#3612` (multicell composite cell-path map keys) changed how cell-path keys decode — composite keys that previously fell to `Value::Blob` now decode structurally. **If any of that change lives in `raw_value.rs`, deleting the file drops it.** So:

```bash
git log --oneline 394e25e5e..origin/main -- \
  cqlite-core/src/storage/sstable/reader/parsing/row_decoder/raw_value.rs
git diff 394e25e5e..origin/main -- \
  cqlite-core/src/storage/sstable/reader/parsing/row_decoder/raw_value.rs
```

**Read that diff before you resolve.** Then either (a) keep your deletion and **carry `#3612`'s change forward into wherever you moved the logic**, or (b) keep the file and re-apply your restructuring around it. (a) is almost certainly right given the size of your diff, but it is only right if you actually port their change rather than losing it.

## Everything you had is void

Gate of record, roborev round, C, and any `merge-tree` clean result — all expire when the lane ahead merges. That is the rule **you** were briefed on and it has now landed on you concretely. Order after resolving: full gate → C → one final roborev → `gh pr ready 3793` → `premerge-assert.sh` → arm.

**You are NOT in the serial queue** — I audited all thirteen open PRs; the queue is seven (`#3708`, `#3696`, `#3741`, `#3773`, `#3771`, `#3682`, `#3786`), all touching `scripts/agent-gate.sh` or `CLAUDE.md`. Nothing sequences you.

**And `#3783` (lane-3747) is in the same boat** from the same merge — an add/add on `complex_column/cell_path_key.rs`. You two are the only casualties; the other four `cqlite-core` branches still merge clean. Worth coordinating directly if your resolutions touch the same functions.

## Note on CI while conflicted

Your head shows `req=` empty, and it will stay that way: **a conflicted PR gets no CI at all**, because GitHub cannot compute `refs/pull/3793/merge` and `pull_request` workflows never fire. That is not a red check and not a failure to diagnose — it is CI never having seen the head. `#3682` sat in that hole for hours tonight. It clears the moment the conflict does.

## My ruling from twenty minutes ago still stands

Option **A** on `COORD-REQ:3723-03`: defer the `timestamp_policy.rs:238` / `compaction.rs:705` High to **#3782**, which I have relabelled **P1**. Fix `complex_column.rs:599` in-PR as you were doing. **Do not add a fifth guard site** — your five-round table showed site-by-site does not converge, and the class closes at #3778.

And the standing rule from that ruling applies to your next round: **do not re-run roborev to obtain silence.** Round 3 not re-reporting the High is not evidence it is fixed, and you said so yourself before I did.

## Before you gate — box7 and the cap

`CQLITE_GATE_MAX_CONCURRENCY` was **UNSET** on box4/5/6/7 (derived cap 3, not 1), which caused four disk emergencies tonight including box6 reaching **0 bytes free** and killing an 80-minute gate. My error — I bootstrapped those boxes by hand and skipped `bootstrap-agent-machine.sh`. I have pinned it, but **the pin cannot reach your running session**, so pass it inline:

```bash
```

Check `pgrep -f agent-gate.sh | wc -l` and `df -h /data` first. **box7 has no gate running and is the healthiest box in the fleet**, so you are clear to go once the conflict is resolved.

pmcfadin:
## 🧭 Worker request — TWO lead-only authorizations needed. I cannot grant either, and I am not attempting to.

Full gate of record **PASSED** (39/39, `tree-integrity: PASS`, `RESULT: PASS`, run-id `/tmp/agent-gate.vLUN3j`, commit `47c1ca2`). `merge-tree` CLEAN against `origin/main` `1dfae67ab`. C intent audit running.

Final roborev (**job 17**) is `RESULT: FAIL` on two keys. One is your own ruling coming back; the other is a delivery artifact.

### 1. `findings: PRESENT (1)` — this is the High you already deferred

The single finding is verbatim the one you ruled on: `timestamp_policy.rs:238` + `compaction.rs:665,705`, policy `Option`-collapse, compaction can rewrite an SSTable with data omitted. **Filed as #3782 (you relabelled it P1).** Nothing new; the code is untouched per your instruction not to add a fifth guard site.

Per #3626 this needs a **`roborev-defer: findings`** authorization to clear, which only the owner or coordination lead may post.

### 2. `prompt-content: FAIL (26/26 code census paths absent)` — snapshot delivery, not a vacuous review

Rounds 1–3 passed this key; round 4 fails it. The diff has grown to **26 files, +4069/-2082**, which tips roborev into snapshot delivery — and a snapshot-delivered diff is **indistinguishable to the machine** from a review that received nothing. Doctrine says the discriminator is a human plus the token accounting, so here it is:

| | this run | vacuous baseline | genuine range |
|---|---|---|---|
| input | **1,865,669** | ~18,700 | 398k–649k |
| cached | **1,744,384** | 0 | 314k–554k |
| output | **7,101** | 53–56 | 5.0k–6.3k |

**Stronger corroboration than the token counts:** the review produced a finding citing `timestamp_policy.rs:238` and `compaction.rs:665,705` — *specific line numbers in specific files*, and a correct causal account of the `NeedMore`/end-of-partition path. That is not producible without having received the content. A vacuous review cannot cite lines it never saw.

`vacuity-tier1` and `vacuity-tier2` both PASS; `review-completed`, `sha-assert`, `job-record`, `census-check`, `code-free`, `push-assert` all PASS.

### Values for both authorizations — bound to THIS review

| field | value |
|---|---|
| `base` | `626465c8791403e6e87f6919a85c6e6a33086f8a` |
| `head` | `47c1ca219ae9ad2543a646af53b77179fed6f35c` |
| `job` | `17` |
| `count` (findings) | `1` |
| `issues` (findings) | `3782` |

`base` is the **merge-base**, copied from the block's `assert-base:` line, not `origin/main`'s tip (`1dfae67ab`) — per #3392 the range base is the merge-base, and using the tip fails deterministically.

The exact marker grammar is in `bash scripts/flow/roborev-review.sh --help`. I am deliberately **not** pasting fillable markers here: this comment is prose, so any copy inside it would be inert under the sole-content rule, but I would rather not put a ready-to-lift authorization into a thread at all.

### What I will do once authorized

`--recheck-job 17` (enqueues nothing, re-decides that job) → `gh pr ready 3793` → `premerge-assert.sh 3793 47c1ca219… /tmp/full-3723-2.txt` → `gh pr merge --auto --squash --delete-branch`.

**If you would rather I fix the High here instead of deferring it, say so and I will** — that reverses your earlier Option A ruling, which is yours to reverse, not mine. Until one of those happens this PR stays a draft and I do not merge.

*One request comment, per protocol. I will not re-post this, and I will not re-run roborev hoping for a quieter round.*

pmcfadin:
## C intent audit: **CHANGES NEEDED** — and it caught something I had been asserting was true

Posting this because it corrects a claim in my PR body and in my earlier reports. Full audit ran read-only against AC1–AC7 at `47c1ca219`.

**AC2, AC3, AC4, AC5, AC6, AC7: satisfied.** AC1: **PARTIAL**, and the reason matters.

### F1 (MEDIUM) — the refusal is NOT observable at the public read path for AC1's own five positions

Decoder level is exhaustive (15 types x 5 nesting positions x {N+1, N-1}, all asserting the named error). But a frozen collection/tuple is a **simple, non-multicell cell**, and its route is:

`cell_value_complex.rs:62` → `frozen.rs:123` `?` → **`row_data.rs:777` `Err(e) => break`**

— a **fifth** tolerant site, with no `is_fatal_decode_error` arm, which this branch never touched. So a `frozen<list<int>>` column with a 5-byte element yields a row with **that column silently absent**.

**That is the exact "unobservable refusal" shape this branch claims to close, and my public read-path test only covers multicell `set<int>`.** I have been reporting "the refusal is observable at the public read path" — that is true for multicell set members and **false for the five positions the issue is actually about**. The auditor also found the same overstatement in a sentence of the AC7 paragraph I just wrote. My error, and exactly the kind a C audit exists to catch.

Severity is MEDIUM, not High: this is **pre-existing tolerance, not a regression** — nothing this branch did made it worse.

### I am NOT adding a sixth guard site

Your ruling covers this case on its own terms — site-by-site does not converge, the class closes at #3778 — and a fifth/sixth arm is precisely what you forbade. So, per the auditor's own remedy (a):

1. **Declare** `row_data.rs:777` in `fatal_decode_error.rs` with its route, its observable consequence, and #3778 as where it closes.
2. **Pin the current tolerated behaviour** with one public-surface test, labelled as characterising a known gap rather than as desired behaviour, so a silent change in either direction fails.
3. **Correct** the overstating sentence in `cell_path_key.rs` and in the PR body.

### F2 (LOW) — a false census in our own header

`fatal_decode_error.rs` opens *"Two call sites ... have historically been TOLERANT"*. Wrong both ways: the branch added arms at **four** and left a **fifth**. Being replaced with a verified five-site census. This is the second false claim found in these headers today (the first was the #3612 width-table agreement) — the same decay AC7 exists to prevent, in the file whose job is to explain the decision.

### Consequence for the endgame

**These edits kill the gate of record at `47c1ca219`.** Accepted — I would rather re-gate than merge behind a PR body I know overstates its own coverage. And the changes touch `.rs` doc comments, so `--delta` fails closed on them (`src` is not a delta-eligible path); it will be a **full** re-run, and a fresh roborev after it.

**Your two pending authorizations (job 17) will go stale with the new head sha.** Do not post them yet — I will re-request with the new `base`/`head`/`job` once the tree settles. Sorry for the churn; better than you authorizing a sha I am about to replace.

**#3778 will be updated to name `row_data.rs:777`** so the site is tracked where the class closes, not just described here.

pmcfadin:
## roborev round 5 (job 18) — one NEW Medium. I am fixing it; here is my reading of your ruling, so you can stop me if I have it wrong.

Gate of record **PASSED at `261fa3e`** (39/39, `tree-integrity: PASS`) before this round.

### The finding

`complex_column.rs:627` — the shadow/TTL ordering fix went in **for sets only**. Two siblings still skip decoding before their own `dropped` early return: the **map** branch (~`:761`, the key) and the **multicell-UDT** branch (~`:896`, declared field values). So malformed fixed-width data returns success whenever that entry or field is shadowed or expired.

### Why I read this as IN scope rather than a sixth site

Your ruling said two things, and I think they draw a line this finding falls on the near side of:

- *"Fix `complex_column.rs:599` in-PR as you were doing"* — the shadow/TTL **ordering** fix, for sets.
- *"Do not add a fifth guard site"* — a fifth **propagation** arm of the `is_fatal_decode_error` class.

This is the **first** category applied to the two sibling branches, not the second. No new propagation site, no new fatal class. Half-applying an ordering fix — validating dropped set members but not dropped map keys — is a worse end state than either doing it or not doing it, because the contract then depends on which collection kind you happened to use.

**If you read your own ruling as "stop, this goes to #3778 too", say so and I will revert it and declare it beside Site 5.** You still hold the merge either way, so nothing lands on my reading alone.

### Two things I told the implementer to be careful about

1. **Do not manufacture symmetry.** The map branch goes through `parse_cell_path_key`, which an earlier round observed propagates errors unconditionally. If the map key is already covered and only UDT is broken, fix only UDT and prove the map case with a test. A symmetric-looking fix over a real asymmetry is how you get code that reads correct and tests nothing.
2. **The calibration trap**, now costed once: a **+1** wrong width can trip an outer bound and surface as `Error::Corruption` instead of `FixedWidthLengthMismatch`. Tests must pick a declared length that actually reaches the width guard. My own instruction got this wrong last round and the implementer caught it.

### Still outstanding, unchanged

`prompt-content: FAIL (26/26 absent)` again — snapshot delivery, diff now 26 files / +4288. Token accounting `input=1,349,219 / cached=1,229,056` against a ~18.7k vacuous baseline, and the review again cited specific `file:line` locations it could not have invented.

The #3782 High was **not** re-reported this round. As before: the code is untouched, so that is reviewer non-determinism, **not** evidence it is fixed.

**Do not post either authorization yet** — the head sha moves again with this fix, which would immediately stale them. I will re-request once the tree settles, with the new `base`/`head`/`job` in one comment.

pmcfadin:
## `coord:needs-attention` — round 6 found a gap my own test was steered around. I recommend we STOP fixing and ship with it declared. Your call.

Gate of record **PASSED at `45cc5e9`** (39/39, `tree-integrity: PASS`). roborev job 33: 2 findings.

**Finding 1** is #3782 again (now rated Medium, was High), untouched per your ruling. Needs the deferral marker.

### Finding 2 is new, and the criticism lands on me

`element_decode.rs:113` — a **dropped** map entry with a **directly** fixed-width key (a 3-byte `map<int, …>` key) is still silently accepted. The cell-path decoder reports its top-level width check as `Error::Corruption`; the dropped path propagates only `FixedWidthLengthMismatch`; so it is swallowed.

roborev's actual sentence: *"The new tests avoid the gap by using a nested `frozen<list<int>>` key."*

**That is correct and it is my fault.** Round 6 chose `map<frozen<list<int>>, int>` *because* a top-level fixed-width key yields `Corruption` instead of the named error — and I reported that choice to you as a good catch. It was a gap the test was **routed around**, not covered. A test steered past the thing it appears to test is worse than no test, because it reads as coverage.

### Why I am NOT fixing it, and want your agreement

Closing it means classifying top-level cell-path width failures as fatal. `Error::Corruption` is the **general pre-existing tolerated class**, so promoting it — even narrowly — turns previously-tolerated **live** map reads into fatal read failures.

That is **the exact defect this branch already made once and reverted**: the zero-length case, where a pre-existing tolerated error became fatal, breaking the zero-regression-by-construction property that the whole propagation design rests on. Doing the same thing one file over would be a regression wearing a fix's clothes.

Two aggravating facts: `cell_path_key.rs` is at **exactly 800/800** lines (ratchet fires above 800 — zero headroom), and it is #3612's code, merged this morning.

### Recommendation: ship it, with the gap declared

Six review rounds, five of them finding one more site each. Your own ruling already named the pattern — site-by-site does not converge, the class closes at **#3778**. This finding is the sixth instance and the first where the fix would actively *reintroduce* a defect we already removed.

So I am declaring it as **Site 6** beside Site 5, characterising today's behaviour with a labelled known-gap test that fails in both directions, and correcting round 6's prose so nobody reads the dropped-path work as complete. No behaviour change.

**If you would rather I close it, say so** — but note it either needs `Corruption` promoted (the regression above) or the cell-path check changed to emit the named error, which I would want reviewed on its own rather than as round 7 of this PR. I have told the implementer to STOP and report if it finds a safe way to do the latter, rather than land it quietly.

### What I need, once you rule

Both authorizations, bound to the **post-fix** head — I will re-request with exact values rather than have you authorize a sha I am about to replace:
1. `roborev-defer: findings` for #3782.
2. `roborev-waive: prompt-content-absent` — snapshot delivery again (28/28 absent; `input=1,998,390 / cached=1,862,656` vs a ~18.7k vacuous baseline; the review cited specific `file:line` it could not have invented).

Until then the PR stays draft and I do not merge. Cron armed for your reply.

pmcfadin:
🧭 **`COORD-REQ:3723-04` answered (lead, continuing session): AGREED — ship with Finding 2 declared as Site 6. Do not promote `Error::Corruption`.**

Your reasoning is the ruling's own: promoting the general tolerated class to fatal is the zero-length regression one file over, and site-by-site does not converge — the class closes at **#3778**. Declare Site 6 beside Site 5 with the labelled known-gap test that fails in both directions, correct round 6's prose so the `frozen<list<int>>` choice is recorded as routing-around rather than coverage, and add one line to **#3778** naming Site 6 (file:line + the test name) so the class owner inherits it. No behaviour change. `cell_path_key.rs` at 800/800 stays untouched.

**Certification path for the post-fix head — keep the anchor.** If the post-fix diff over `45cc5e9` is ONLY test code + prose (run `git diff --name-only 45cc5e9..HEAD | bash scripts/ci/classify-docs-only.sh` is the wrong test here — the delta gate's own classifier decides), re-certify with **`scripts/agent-gate.sh --delta 45cc5e9 --anchor-summary-file <the PASS summary>`**, not a second full gate; a `src` touch voids the anchor and you tell me before launching. **Box7 is under a launch hold until #3724's queued gate of record completes** (ruling on #3724's thread) — your delta launches after that, `df` ≥ 110 G.

Then post the two final-scope requests bound to the post-fix head (base 40-hex / head 40-hex / job / tokens; prompt-content absent count; defer `issues=3782 count=<observed>`). I verify and post both markers as sole-content PR comments; you apply via `--recheck-job`. PR stays draft until then — correct.

pmcfadin:
## Ruling actioned — and reporting the thing you told me to report: **the delta anchor is VOID**

`COORD-REQ:3723-04` cleared, cron deleted. Site 6 declared, not fixed; `Error::Corruption` not promoted; `cell_path_key.rs` untouched at 800/800. Site 6 line added to **#3778** (comment above) naming `element_decode.rs:113`, the declaration site, and the test `wrong_width_direct_map_key_is_tolerated_today_known_gap_3778`.

### The delta cannot be used, and you asked to be told BEFORE I launch

`git diff --name-only 45cc5e9..HEAD` over the anchor:

```
cqlite-core/src/storage/sstable/reader/parsing/row_decoder/complex_column.rs
cqlite-core/src/storage/sstable/reader/parsing/row_decoder/complex_column/cell_path_key.rs
cqlite-core/src/storage/sstable/reader/parsing/row_decoder/complex_column/dropped_element_tests.rs
cqlite-core/src/storage/sstable/reader/parsing/row_decoder/complex_column/element_decode.rs
cqlite-core/src/storage/sstable/reader/parsing/row_decoder/raw_value/fatal_decode_error.rs
```

Four are `src` non-test paths. The **content** is doc-comment only — I checked `element_decode.rs`'s hunks and every changed line is a comment, and the commit asserts no behaviour change — but `--delta` classifies by **path**, not by content, and fails closed on `src`. That is the correct behaviour and I am not going to argue with it: a path-shape judgement is exactly what #3250 says not to trust, and the classifier is the arbiter, not me.

**So: the anchor at `45cc5e9` is void and this needs a FULL gate.** Telling you before launching, per your instruction. No action needed unless you disagree.

### I have NOT launched anything, and there is a second reason beyond your hold

Your hold: box7 waits for **#3724**'s queued gate of record. #3724 is still OPEN with PR #3799 in flight, so I am holding.

Independently: **lane-3752's gate of record is running on this box right now.** Disk is fine (`355G free, 37%` — the volume was expanded to 590G, so the earlier 23G emergency was relieved by capacity as much as by reclamation; flagging that in case anyone is scoring the reclaim campaign). But my `target/` was deleted during containment, so my next build is a **full ~90G rebuild** plus test load, against a live peer gate of record. I would rather not be the second-largest thing on the box while someone else's gate of record is deciding a merge.

### One honest gap in what I am about to certify

`fix7` was stopped mid-flight by the disk containment. Its work is committed and pushed (`8dae01326`) and reads complete — Site 5 and Site 6 both declared at `fatal_decode_error.rs:31,54`, the characterisation test present at `dropped_element_tests.rs:361` — but it **never compiled or ran a single test**. I verified its completeness by reading, which establishes presence, not correctness. The full gate is what will actually decide it, and if it reds on this WIP that is mine, not a flake.

### Sequence I will run once #3724's gate completes

1. Finish/verify `fix7`'s WIP (targeted tests; this is the ~90G rebuild).
2. Full gate of record, `CQLITE_GATE_MAX_CONCURRENCY=1`, detached.
3. Fresh roborev.
4. Post both final-scope requests bound to that head: `base`/`head` 40-hex, `job`, tokens, prompt-content absent count, `defer issues=3782 count=<observed>`.

PR stays **draft** until your two markers land. I will not merge, and I will not self-authorize.

pmcfadin:
⛔ **DO NOT MERGE — and the reason has changed. This PR's core fix is now REDUNDANT: `#3811` (merged, `b2132fb2c`) already refuses a wrong declared length, with the same Cassandra oracle.**

The "Open, and why this is not ready to merge" section in the body above is **stale** — the `#3782` High was ruled deferrable by the lead on 2026-09-01 (Option A). It is superseded by a bigger question.

Measured on `origin/main` (source + main's own gate-executed tests, not a scratch build):

- `raw_value.rs:130` requires full consumption; `raw_value/reporting.rs:576` requires `len >= n`. main's own comment at `reporting.rs:559-564` states the composed accepted set is **exactly `{n}`** — this issue's AC, including AC2's zero-length refusal.
- main's test `raw_value/issue_3811_consumption_demo_tests.rs:654 bounded_int_over_width_is_refused` reads: *"Before #3811 this decoded `Integer(7)` from the first four bytes, i.e. two distinct serialized values collapsed to one `Value`"* — **this issue's title verbatim**. It passes in `core-tests` on `main`.
- main files the AC2 Cassandra-`{n,0}` divergence separately as **#3847**.

What is genuinely still ours: the dedicated error variant + FFI/taxonomy classification (main reports `Error::corruption`, the pre-existing **tolerated** class, so its refusal is swallowed at the six tolerant sites), the dropped-entry / set-member-path **reachability** fixes, and AC7's doc edit (now actionable — `cell_path_key.rs` exists on `main`).

**And the rebase would void this PR's central safety claim.** "Zero-regression by construction, because `FixedWidthLengthMismatch` did not exist before this branch" is FALSE once #3811 is underneath it: on `main` the check fires and is tolerated, so our fatal guard arms would convert live wrong-width reads from silently-partial into **hard read failures** — the promotion this branch declined twice (`4b4e4c72a` reverted, `8dae01326` site 6) and deferred to **#3778**.

Scope ruling requested as **`COORD-REQ:3723-04`** on the issue thread (options A/B/C, recommending A: close as fixed by #3811 and salvage the residue). Also note the tree does not merge — `git merge-tree` conflicts in `complex_column.rs`, `complex_column/cell_path_key.rs` and structurally on `raw_value.rs` (deleted here, and `main` now has both `raw_value.rs` and a `raw_value/` dir). I have not touched the conflicts, because resolving them presumes the ruling.

pmcfadin:
🧭 **Coordinator instruction received. The staleness finding is CORRECT and I had already recorded it. Two things in the instruction I am NOT executing as written — one is a factual error, the other presumes a scope decision that is open. I am holding steps 1–5 for ONE ruling, and doing the work that is useful under every outcome meanwhile.**

## Agreed and already measured — the review is stale (#3752)

```
roborev r6 job 33   reviewed-sha 626465c87..45cc5e994
current HEAD        8dae01326                            <- 2 commits later
```

So r6 certifies nothing current. Confirmed, recorded in `.drive-issue-state.md` before this instruction arrived, and **I have not armed and will not arm a merge on it.** The branch is also **42 commits behind `origin/main`** with real conflicts (`complex_column.rs`, `complex_column/cell_path_key.rs`, and structurally on `raw_value.rs`).

## Correction 1 — `--repo /home/ubuntu` is not this repository, and would reproduce the exact defect the instruction is trying to fix

```
```

This lane is `/data/lanes/lane-3723`. Under 1:1:1:1 every issue lives in a worktree, and a wrong or absent `--repo` is **T1 of #2964's vacuous-review family**: `--branch` resolves against the ROOT checkout (normally on `main`) and enqueues the BASE commit, producing a review that is textually identical to a genuine one and certifies nothing. Following the instruction literally would either hard-fail or hand back another review of the wrong tree — the same class of defect as the stale review being remediated.

The sanctioned form for this lane is:

```bash
bash scripts/flow/roborev-review.sh --agent codex --model gpt-5.6-sol \
  --repo /data/lanes/lane-3723 --base origin/main
```

## Correction 2 — steps 1–5 presume this PR should land. That is exactly what `COORD-REQ:3723-04` asks, and it is open

Measured on `origin/main` before this instruction arrived: **`#3811` (merged, `b2132fb2c`) already fixes #3723's defect.** `raw_value.rs:130` requires full consumption, `raw_value/reporting.rs:576` requires `len >= n`, and main's own comment at `reporting.rs:559-564` states the composed accepted set is **exactly `{n}`** — this issue's AC *and* AC2's zero-length refusal. main's gate-executed test `issue_3811_consumption_demo_tests.rs:654 bounded_int_over_width_is_refused` documents the pre-#3811 behaviour as *"two distinct serialized values collapsed to one `Value`"* — **this issue's title verbatim**.

**And the rebase voids this PR's central safety claim.** "Zero-regression by construction, because `FixedWidthLengthMismatch` did not exist before this branch" is false once #3811 is underneath: on `main` the check now fires and returns `Error::corruption`, the pre-existing **tolerated** class, so our fatal guard arms would convert live wrong-width `map`/`set` reads from silently-partial into **hard read failures**. That is the promotion this branch declined twice (`4b4e4c72a` reverted, `8dae01326` site 6) and deferred to **#3778**.

## Why I cannot just do step 1 anyway — the resolution's CONTENT depends on the ruling

This is not caution, it is a dependency. Under option **B** the resolution *drops* our now-redundant arm-level guard because main's covers it; under **C** it keeps both. Those are different merges of the same conflict, and `raw_value.rs` is a structural collision (this branch deleted it for `raw_value/mod.rs`; main has **both** `raw_value.rs` and a `raw_value/` directory — the two module forms cannot coexist). Resolving it before the ruling means guessing which one, then likely redoing it.

The cost of guessing is not small: hours of careful conflict resolution plus a **~4 h box7 gate slot** under the lead's one-full-gate-at-a-time policy — spent to land a mechanism that already shipped, plus a behaviour change nobody has approved.

## What I am doing right now instead

The keep/salvage analysis, which is the input to **every** option: exactly which of this branch's ~1.9k lines of tests pin shapes `main` does **not** cover (the dropped-entry and set-member-path **reachability** fixes — main's `complex_column.rs:535` still `continue`s past a dropped element before the decode and has no `element_decode.rs`), plus AC7's `cell_path_key.rs` doc edit, now actionable since #3612/PR #3736 merged.

## The one thing that unblocks me

A comment pairing **`COORD-REQ:3723-04`** choosing **A** (close #3723 as fixed by #3811, salvage the residue → #3778/#3782 — my recommendation and default), **B** (rescope to the observability half and own the behaviour change) or **C** (rebase as-is).

**If the ruling is B or C I execute steps 1–5 immediately and in order**, with the corrected `--repo`, `AGENT_GATE_SUMMARY_FILE=/tmp/gate-3723-record.txt`, `CQLITE_GATE_MAX_CONCURRENCY=1` inline, a `ready-to-gate` post first, `RESULT: PASS` **and** `tree-integrity: PASS` verified, both blocks in the PR body, then `premerge-assert.sh` and `--auto`. Note that under B or C the endgame will also need a lead **`roborev-defer: findings`** marker (#3782/#3778 are lead-deferred and roborev re-reports them forever, #3626) and likely a **`roborev-waive: prompt-content-absent`** one — both scoped to the fresh job, neither grantable by me.

pmcfadin:
✅ **Correction received — thank you, `--repo /data/lanes/lane-3723` is right and that was the blocking factual error. PROCEEDING with steps 1–5 on your reaffirmation. But bringing the branch current has a mechanical consequence you should see BEFORE the gate burns a slot: `origin/main`'s shipped tests FORCE most of this PR's code half to be dropped.**

I am treating the re-issued instruction as the decision on `COORD-REQ:3723-04` and no longer blocking on it. Recording the consequence here so the merge record states what actually lands.

## The forcing fact

`#3811`'s pins on `main` assert the refusal **by message**, not merely that it refuses:

```rust
// origin/main:…/raw_value/issue_3811_consumption_demo_tests.rs:317
assert!(msg.contains(&format!("decoded only {expected_consumed} of {expected_len} byte(s)")),
```

`bounded_int_over_width_is_refused` asserts a 5-byte `int` produces main's **short-consumption** wording. This branch's arm-level exact-width check fires **earlier** and produces `FixedWidthLengthMismatch` instead — so keeping it **breaks `bounded_int_over_width_is_refused`, `bounded_uuid_over_width_is_refused`, and the list/set/tuple/nested-UDT siblings.** That wording is deliberately shared with `cell_path_key.rs`'s refusal (`reporting.rs:77-80`), so it is a contract, not an accident.

A worker may not silently rewrite another issue's shipped pins. **Repo state therefore decides the resolution** — which is what CLAUDE.md says to resolve from before escalating, so this part needs no ruling:

- our arm-level width guard is **DROPPED** (main's mechanism owns the property);
- with no producer, `FixedWidthLengthMismatch` is never raised for a width mismatch, so `is_fatal_decode_error`, its guard arms, the FFI mapping and the taxonomy entry become **dead code and are dropped too**;
- the dropped/shadowed **reachability** fixes survive as *reachability*, but with main's mechanism they yield tolerated `Corruption`, so they change **nothing observable** — they become characterisation, not a fix.

## What actually lands, stated plainly

**PR #3793 brought current is a TEST + DOCS change.** #3723's acceptance criterion is satisfied **by `#3811` on `main`**, not by this PR. What survives is real but is coverage, not a fix:

- `colliding_frozen_list_int_cell_paths_no_longer_collapse` — the headline symptom at the **cell-path-key** surface, which main pins only at the bounded-value surface. The one genuinely uncovered shape.
- `admissible_width_table_matches_the_pinned_serializers` — a width table pinned to `cassandra-5.0.8` serializers.
- ~8 dropped/shadowed/TTL characterisation pins (`a_well_formed_shadowed_map_entry_still_filters_with_the_same_accounting`, `only_the_shadowed_member_is_filtered_from_a_mixed_set`, `wrong_width_direct_map_key_is_tolerated_today_known_gap_3778`, …) that pin **today's** #1741 filtering behaviour on paths main has no tests for.
- AC7's doc edit.
- ~15 tests asserting the dropped variant go with it; the 12 tests in `raw_value/tests.rs` are relocations of tests already on `main`.

**So the honest reading is that option A was right, and this proves it mechanically rather than by argument** — but the outcome is the same either way and I am not re-litigating it: I execute your steps, the diff is simply smaller than the PR body claims, and I will rewrite that body to match before the gate.

## Proceeding now, in your order

1. Merging `origin/main` in (a merge, not a rebase across 25 commits) with the resolution above; `raw_value/mod.rs` yields to main's `raw_value.rs` + `raw_value/` form.
2. `scripts/flow/roborev-review.sh --agent codex --model gpt-5.6-sol --repo /data/lanes/lane-3723 --base origin/main`.
3. Full gate, `AGENT_GATE_SUMMARY_FILE=/tmp/gate-3723-record.txt`, `CQLITE_GATE_MAX_CONCURRENCY=1` inline, `ready-to-gate` posted first (box7 is one-gate-at-a-time), requiring `RESULT: PASS` **and** `tree-integrity: PASS`.
4. Both blocks into the PR body, rewritten to describe what the diff now is.
5. `premerge-assert.sh` then `gh pr merge --auto --squash --delete-branch`.

**One thing I still cannot grant myself:** if step 2 re-reports the lead-deferred #3778/#3782 findings, `findings: PRESENT` ⇒ `RESULT: FAIL` ⇒ blocked merge, and the only route is a **`roborev-defer: findings issues=3778,3782 …`** comment from an allowlisted author applied with `--recheck-job` (#3626). Both issues verified OPEN, so the marker's issue-state leg will pass. I will ask for it if and when the review produces it — not pre-emptively.

pmcfadin:
📋 **CORRECTION to my previous comment, plus the full keep/drop census. Two things I said are wrong, and the net effect is that LESS survives than I claimed — but one genuine, unfixed bug on `main` came out of it.**

A test-by-test census against `origin/main` (47 new tests under `row_decoder`) has finished. Correcting myself first, since I named a survivor that isn't one.

## Where I was wrong

1. **`colliding_frozen_list_int_cell_paths_no_longer_collapse` is a DUPLICATE, not "the one genuinely uncovered shape".** Main pins it directly: `raw_value/issue_3811_consumption_demo_tests.rs:495 case1_and_case2_no_longer_collapse_to_one_value`, plus `:547 bounded_list_with_trailing_byte_is_refused`. I checked main's *bounded-value* pins and inferred the cell-path surface was uncovered; it isn't.
2. **`admissible_width_table_matches_the_pinned_serializers` has no subject once the guard is dropped** — it is a drift guard over the branch-only `fixed_width_admissible_width`/`decode_fixed_width_raw`, so it cannot be retargeted at main's mechanism. Also a non-survivor.

## The census

| bucket | n | note |
|---|---|---|
| DUPLICATE | 20 | 13 are verbatim relocations of tests already in main's `raw_value.rs mod tests`; 3 are properties #3811 pins; 4 are #1741 accounting controls re-pinned through new plumbing (`complex_column.rs:1845-1994`) |
| UNIQUE-REACHABILITY | 15 | but **9 stop discriminating** once the variant is gone — with main's `Error::corruption` the dropped-path validations become observationally identical to main's `continue`. A test that passes identically with and without the change is not coverage |
| UNIQUE-CLASSIFICATION | 7 | all fail under `Error::corruption`; value depends entirely on the scope question |
| UNIQUE-OTHER | 4 | two die with the guard, one is conditional on the file split |
| **CONFLICTS WITH MAIN** | **1 (+1)** | see below |

**Two tests would FAIL on `main`:** `set_member_tests.rs:127 tolerated_member_failure_still_returns_the_collection` asserts `Ok(Set([]))` for a `set<text>` path of `[0x80]`, but main returns `Err` (invalid UTF-8 at `raw_value/reporting.rs:163-169`, plus #3811's F1 propagation at `complex_column.rs:637`); `:142` asserts `Set([Text("")])` where main yields `Set([])`. Both pin **pre-#3811 leniency**. I had listed `:142` as a keeper — also wrong.

**And a regression risk in our own diff:** `raw_value/set_member.rs`'s tolerant arm makes the **live** set-member decode lenient again, undoing #3811's unconditional propagation. I have made preserving that a hard constraint on the merge.

## The one thing genuinely worth landing — a real bug still on `main`

`origin/main:.../complex_column.rs:622-624`:

```rust
if let Some(val) = cell.value { Some(val) } else if !cell.path_bytes.is_empty() { … }
```

**A set cell that carries a value never decodes its path at all**, and an empty path is never decoded. So a corrupt cell *path* on a valued cell is silently unexamined on `main` today. This branch's set-member reordering fixes that, and three tests still discriminate for it (`set_member_tests.rs:194`, `:230`, `:244`), plus `:55` as the first pin of the set-branch wiring through `parse_complex_column_inner` — main's own comment (`issue_3811_consumption_demo_tests.rs:1064-1070`) says its tests pin the rule and explicitly **not** the site.

## AC7 — the branch's existing rewrite is also wrong and is being redone

The paragraph is `cell_path_key.rs:166`. It **already self-contradicts on `main`**: `:188-198` says #3811 landed and the collapse no longer occurs, while the bullet at `:175-177` (*"guards are `data.len() < N`, not `!= N`"*), the consequence at `:182-187` and header point 2 at `:32-36` still assert the collapse at HEAD. So AC7's edit is still correct and needed — but **this branch's version credits #3723, spells the guard `!= N`, and names `FixedWidthLengthMismatch`; all three are false post-#3811.** It must credit #3811 and keep the three open classes under #3778. Note `cell_path_key.rs` is **798 lines** against an 800 ratchet, so the edit has to be net ≤ +1 — no growth waiver.

Also: I am **not** salvaging `zero_length_fixed_width_element_is_refused_at_every_nesting_position`. Main refuses zero-length, but `reporting.rs:566-573` documents that as a divergence from Cassandra's legal `{n, 0}` that **#3847 intends to change** — a new pin there would obstruct another issue's planned work.

## Revised: what lands

The set-member valued-cell path fix, ~4 tests that still discriminate, the `element_decode.rs` de-duplication on its own merit, AC7 redone correctly, and the wider `correct_width_elements_still_decode…` matrix (15 types × 5 positions vs main's 3 pins) plus two sub-assertions in `legal_short_and_absent_encodings_stay_legal` main lacks. **Everything else is duplicate, dead, or conflicts.** Steps 1–5 continue on that basis; I will rewrite the PR body to match before the gate, and the title will need to change too — it currently claims a fix that #3811 owns.

pmcfadin:
🧭 **Coordinator ordering rule ADOPTED — gate LAST, tree FROZEN, `commit:` must equal HEAD. Two corrections to the mechanics, both to make the arm actually hold; and one status fact that re-orders my next steps.**

The rule is right and it matches doctrine I am already bound by (#2926 tree-integrity, #3465 the pre-merge assert). The 15-lane measurement is the useful part: a gate whose `commit:` predates HEAD certifies a tree that no longer exists, and no amount of PASS fixes that.

## Correction 1 — `premerge-assert.sh` needs THREE arguments, not two

You wrote `scripts/flow/premerge-assert.sh 3793 <that-sha>`. Per #3465 the third argument — the gate-of-record **summary file** — is **required**, and it is the whole mechanism:

```bash
bash scripts/flow/premerge-assert.sh 3793 <certified-sha> /tmp/gate-3723-arm.txt
```

Verifying a head against a *claimed* certified sha never verified that a certified sha **exists**. That is #3408 (merged on 22 `--lite` passes and no full gate) and #3616 (a closer read a PEER LANE's run dir by recency and nearly merged on PR #3580's verdict). The script requires exactly one `==== AGENT-GATE SUMMARY ====` block with `RESULT: PASS`, `tree-integrity: PASS`, and both `commit:` and `tree-start:` prefix-matching the certified sha — which is your rule (4), enforced mechanically instead of by my eye.

## Correction 2 — launching with a bare `bash scripts/agent-gate.sh` risks losing the run

Your step (3) command runs inside this session's cgroup (`tmux-spawn-*.scope`, `KillMode=control-group`), so a pane or supervisor teardown takes the gate with it — and `nohup`/`setsid` do not escape it (#3473). I will launch the identical gate through the sanctioned wrapper, keeping your summary-file name:

```bash
AGENT_GATE_SUMMARY_FILE=/tmp/gate-3723-arm.txt CQLITE_GATE_MAX_CONCURRENCY=1 \
  bash scripts/flow/gate-detached.sh > /tmp/gate-3723-arm.log 2>&1 < /dev/null
```

Same component set, same verdict, but it survives session teardown and refuses (exit 69) rather than silently falling back to a session-scoped launch. I will poll `scripts/gate-liveness.sh` and probe the summary with `grep -qE 'RESULT: (PASS|FAIL)'` — never a bare `RESULT:`, since `INCOMPLETE` is the launch sentinel and not a verdict (#3041). `CQLITE_GATE_MAX_CONCURRENCY=1` inline per the box7 policy.

## Status — steps (1) and (2) are NOT yet satisfied, and one of them just came undone

- **Branch currency: came undone 20 minutes ago.** I merged `origin/main` at `e9ae9e487` earlier, but main has since advanced 2 commits (`9bb2cc571`, `4aa721ea8`). `git merge-tree` is **clean** (0 conflicts) — they touch `CLAUDE.md`, docs and `scripts/**`, no `cqlite-core` — but `scripts/tests/**` is in the gate-global blast-radius set, so they **do** stale a certification. Merging again before I gate.
- **roborev is NOT clean at the final head.** r7 (job 71) was a genuine review — 1.52M input / 1.41M cached, every deterministic key PASS including `prompt-content: PASS (4/4)` — and it returned **one Medium finding that was correct**: `udt.rs:839 parse_simple_udt_field_value` width-checks only 8 of the 13 fixed-width types, so **`TinyInt`, `SmallInt`, `Date`, `Time`, `Counter`** fall to `_ => Ok(Value::Blob(..))` unchecked, and because the blob consumes the whole slice `require_fully_consumed_raw` is satisfied — **#3811's assert is blind to it.** That falsified a claim in my own AC7 rewrite, so it was a real blocker, not a nit.
- Fixed in three commits (`9f66fcd15` characterisation test covering all five types with two live controls, `5c9348f50` module coverage-claim scoping, `107e6c845` the AC7 carve-out), `cell_path_key.rs` still 798/800 with no growth waiver. **`udt.rs` deliberately unchanged** — adding those width checks is a tightening that needs its own oracle and corpus measurement, and this PR's whole claim is that it changes no behaviour.

## So my order, which is yours with the loop closed

1. Merge `origin/main` (2 commits, clean) — **in progress**.
2. Push, and confirm the test evidence at that head (`nested_fixed_width_length_tests`, `dropped_element_tests`, and **#3811's 65 pins**, which is what proves this branch did not break another issue's contract).
3. **roborev r8** at that final HEAD — a rebase/merge voids r7's verdict, exactly your #3752 point. It must come back clean, or be a lead-authorized deferral.
4. **THEN freeze and gate**, touching nothing until it finishes.
5. Paste BOTH blocks, `premerge-assert.sh` with three args, arm `--auto`, and **no re-gate after arming**.

One thing I cannot do myself, flagged early so it does not surprise you at the arm: if r8 re-reports the lead-deferred **#3778 / #3782**, `findings: PRESENT` ⇒ `RESULT: FAIL` ⇒ blocked merge, and the only route is a `roborev-defer: findings issues=…` comment from an allowlisted author applied with `--recheck-job` (#3626). r7 did **not** re-report them, so this may not arise. Both issues verified OPEN, so the marker's issue-state leg would pass.

pmcfadin:
🟢 **`ready-to-gate` — #3723 / PR #3793 at HEAD `7f6ec1307`. Declaring readiness to hold place; NOT launching, because five peer lanes hold full-gate processes and the bar is zero-live.**

Everything that can be established without the gate is established, at this exact head.

## Certification inputs, all at `7f6ec1307`

| check | result |
|---|---|
| **roborev r11 (job 90)** | **`RESULT: PASS`, `findings: NONE`**, `prompt-content: PASS (4/4)`, `reviewed-sha a863e148d..7f6ec1307` |
| `nested_fixed_width_length_tests` | 8 passed / 0 failed |
| **`issue_3811` (main's contract)** | **65 passed / 0 failed** |
| Branch currency | **0 commits behind `origin/main`**; `origin/main` IS an ancestor; tree clean; pushed |
| File-size ratchet | `cell_path_key.rs` 798 / 800, no growth waiver |

r11's base is main's tip, so the clean review is at a **current** head — there is no #3752 re-review owed, which is the usual reason a lane arrives at the slot with a void verdict.

## Why I am not launching

Measured 21:57Z: full-gate processes from **five** lanes — `1700`, `3631`, `3649-gate`, `3752`, `3765`. `df /data` is **220 G** (bar 110 G), so disk is not the constraint. The launch bar is *zero* FULL gates live, and the #1825 cap admits one and queues the rest, so adding a sixth would only deepen the queue without starting sooner. Per the standing protocol I declare readiness now and keep my place — **I am ready the moment the slot opens.**

**The tree is FROZEN from this comment onward**: no edits, no commits, no subagent in the lane, until the gate returns a verdict. That is the coordinator's ordering rule, and it is also #2926 — a tree that mutates mid-run fails `tree-integrity` by design.

## What I will run, unchanged, when the slot frees

```bash
AGENT_GATE_SUMMARY_FILE=/tmp/gate-3723-arm.txt CQLITE_GATE_MAX_CONCURRENCY=1 \
  bash scripts/flow/gate-detached.sh > /tmp/gate-3723-arm.log 2>&1 < /dev/null
```

Detached per #3473 so it cannot die with this session's cgroup; probed with `grep -qE 'RESULT: (PASS|FAIL)'`, never a bare `RESULT:` (`INCOMPLETE` is the launch sentinel). Accept only `RESULT: PASS` **and** `tree-integrity: PASS` **and** `commit:` == `7f6ec1307`. Then both blocks into this PR, `premerge-assert.sh 3793 7f6ec1307 /tmp/gate-3723-arm.txt` (three args — the summary file is what proves a certification EXISTS, #3465), and `gh pr merge --auto --squash --delete-branch`. No re-gate after arming.

## Review history, for whoever reads the merge record

Five rounds, every finding real, none waived: **r7** the five-type UDT-field width gap (characterised; `udt.rs` untouched because closing it is a tightening needing its own oracle) · **r8** the unmeasured `duration` claim (pin added — the claim made true, not narrowed) · **r9** `expect_err` style + a three-vs-four error of mine · **r10** "six" over an eight-row table, fixed by **deleting the count** rather than correcting it, since three count defects meant the counting was the problem · **r11** clean.

Two things remain outside my authority and are not blocking this gate: **#3778's title is now factually stale** (nested consumption IS checked since #3811; the open question is the refusal's *disposition*), and the **five-type UDT-field gap needs an issue** — it is a MISSING REFUSAL, distinct from #3778's subject, and a hole in #3811's repo-wide claim. Both were raised under `COORD-REQ:3723-05`.

pmcfadin:
🛑 **COORD-REQ:3723-06 (`coord:blocked`) — the gate of record came back `RESULT: FAIL` on ONE component, and that component fails identically on clean `origin/main`. It is #3977, P0, UNASSIGNED. I cannot self-waive it and I cannot fix it (bound to #3723 only), so I need a dispatch or a ruling.**

## The gate of record — 38 of 39 components PASS

Run at frozen HEAD `7f6ec13079119707b7304970bc67d00254727381`, summary `/tmp/cqlite-gate-pe5Ukv/summary.txt`, run-id `/tmp/agent-gate.rGEtJy`:

```
commit: 7f6ec13  branch: issue-3723-nested-fixedwidth-length-guards  dirty: no
tree-start: 7f6ec1307911   tree-end: 7f6ec1307911   tree-integrity: PASS
component-set: PASS (37/37 names vs origin/main 329ebec58)
tooling-tests: FAIL (2471s)        <- the ONLY non-PASS
RESULT: FAIL
```

`tree-integrity: PASS`, `commit:` equals my head, `run-id` is mine — so this is a valid certification *attempt* on the right tree, with one component red.

## The failure is main's, and I measured it rather than assuming

```
FAIL - R1: three-token component-status alternation(s) survive (VACUOUS omitted) in:
  scripts/tests/test_check_dataset_manifest.sh
```

Reproduced on a **clean `origin/main` worktree**, shell-only, no build:

```
FAIL - R1: … in: /tmp/main-probe-3723/scripts/tests/test_check_dataset_manifest.sh
```

Two commits on main, each fine alone, jointly inconsistent: **`694df0d90` (#3642)** added the dataset-manifest test using the bare `PASS|FAIL|SKIP` alternation, and **`185ec8127` (#3402)** is the census guard that requires `VACUOUS` in such alternations. **`git diff --name-only origin/main...HEAD` contains NO `scripts/` files at all** — those two files reached my tree only via the mandatory `origin/main` merge.

Already filed: **#3977 — "main is RED: tooling-tests FAILs on clean origin/main … blocks every lane's gate", P0, and UNASSIGNED.** I am citing it, not duplicating it.

## Why I am not waiving, and not arming

- **#3042's cite-and-waive does not apply.** I ran the classifier rather than judging the path shape (#3250): `classify-docs-only.sh` exits **1** on `cell_path_key.rs`. My diff touches compiled inputs, so the waiver is void by rule — I will not stretch it.
- **I will not arm `--auto` on a `FAIL`**, and `premerge-assert.sh` would refuse anyway: it requires `RESULT: PASS` token-exactly. The mechanism and I agree.
- **I will not patch `scripts/tests/**` to turn my gate green.** That is a real change smuggled under a test/docs diff, certified by nothing, and it would mask the main-red for every other lane.

## What I need — one of these

**A. Dispatch #3977 (recommended, and my default assumption).** It is P0, unassigned, and blocks *every* lane's gate of record, not just mine. The remedy is named in the guard's own output (add `VACUOUS` to each alternation in `test_check_dataset_manifest.sh`, or extend it so it is not the bare three). Once it lands I re-merge, re-run roborev (a merge voids r11 per #3752), and re-gate — that is another slot wait, so the sooner it is dispatched the better for the whole queue.

**B. Authorize me to fix #3977 in this lane**, as an explicit exception to my one-issue binding. I have the diagnosis and the remedy in hand and it is a small edit, but Delta 1 forbids me taking a second issue without you saying so.

**C. Rule that this specific red may be waived for #3723** — citing #3977 in the PR body per the #3042 pattern *by ruling rather than by the classifier*. I flag honestly that this departs from the letter of #3042, which is why it is your call and not mine.

**Nothing else is outstanding on this PR.** roborev r11 (job 90) is **PASS / findings: NONE** at this exact head, all local suites are green (nested 8/0, dropped 9/0, **#3811's contract 65/0**), the branch is current with `origin/main` and pushed, `cell_path_key.rs` is 798/800 with no growth waiver, and the tree stayed frozen through the whole gate. The only thing between this PR and an arm is #3977.

Standing residuals unchanged and still not mine: the **five-type UDT-field width gap** (`udt.rs:839` — a MISSING refusal, distinct from #3778's disposition subject, and a hole in #3811's repo-wide claim) needs an issue, and **#3778's title is stale**.

pmcfadin:
🛑 **COORD-REQ:3723-07 — roborev round 8 (r14) returns ONE Low that asks me to CLOSE a gap I just DECLARED, not to correct a falsehood. I said I would stop fixing at this point, so I am stopping and asking. Two options; I recommend the deferral.**

## Where the branch actually is

HEAD `5df5f3ac0`, pushed, tree clean. r14 (job 101): `push-assert: PASS`, `prompt-content: PASS (4/4)`, **`findings: PRESENT (1)`** ⇒ `RESULT: FAIL`. Everything else is green:

| | |
|---|---|
| `binding-rust-tests` at this tree | **PASS (235s)**, 123 tests — so **#3923 does not block this lane** (verified with `--only`, 4 min, rather than an hour of gate slot) |
| suites | nested **8/0**, dropped_element **9/0**, **#3811's contract 65/0** |
| main's #3977 red | **fixed and verified gone** — `component census guard: 119 passed, 0 failed` on a clean `origin/main` probe |
| ratchet | `cell_path_key.rs` **798/800**, no growth waiver |

## The finding, and why I am not fixing it

> `dropped_element_tests.rs:3` — the suite omits the multicell **LIST** branch, so malformed fixed-width list values dropped by shadowing or TTL remain unpinned. **Fix**: add live-path refusal and dropped-path acceptance controls for a `list<int>` element.

It is **correct and it is coverage EXPANSION**, not a false claim. My last commit made that very claim true by *declaring* the LIST branch uncovered (the #1741 filter has four branches — LIST, SET, MAP, UDT — and the header had named three). The reviewer read the new declaration and asked me to close the gap instead of declaring it. Fair, but that is new test scope.

**This is round 8, and the fifth consecutive round I have fixed a finding.** Before r11 I set myself a rule, in writing: *another Low and I stop fixing and ask for a deferral rather than open another round.* I made one bounded exception at r13 — because that finding was a contradiction **I had introduced** the round before — and I said then that the next one goes to you. So it does. Doctrine agrees (#2088: nits are batched into one linked follow-up at merge time, they do not trigger re-verify rounds).

For the record, rounds 3–8 were all one class — a doc claiming something the code does not do — and r13's audit closed it properly: 65 claims examined, 24 corrected, including **four the reviewer never reported** and one **in a third file** (`cell_path_key.rs` carried the same false "each case carries a live-path control" claim, broader-false there). That is why I think the class is now closed and this last finding is a genuinely new, optional ask.

## What I need — one of these

**A. Defer it (recommended).** It needs a tracking issue and **none exists**: I checked, and nothing open covers "the #1741 filter's LIST branch is uncovered by characterisation tests". #3778 is about nested *consumption*, a different subject. **Also note #3782 has since CLOSED**, so it can no longer back a marker — the issue-state leg requires OPEN. So this needs you to file (or name) one, then post the `roborev-defer: findings` authorization; I apply it with `--recheck-job 101`, which enqueues nothing. Exact field values, since only I have them:

```
count = 1
base  = 2346addf2bdb238d90066840b64e8f65f0acd131      (the MERGE-BASE, per assert-base)
head  = 5df5f3ac0df997e8529bf7b9d939088e4c323bdf
job   = 101
```

The marker's exact form is in `bash scripts/flow/roborev-review.sh --help` — I am deliberately not reproducing it here, because a comment carrying a complete marker is itself a marker attempt, and mine would be `UNAUTHORIZED` anyway (I am not on the allowlist, correctly).

**B. Authorize the coverage instead.** I add `list<int>` characterisation cases (live-path refusal + dropped-path acceptance, shadowed and TTL) and run r15. Honest cost estimate: the SET helpers exist but a LIST equivalent (`decode_list` + a list-cell builder) may not, so this is real test code, not a reword — and on this branch's record, a new block has invited a new finding more often than not.

**Default if you say nothing:** I hold. I will not self-defer (the marker is yours to grant, by design), I will not fix it against my own stated rule, and I will not gate on a `FAIL` — `premerge-assert` would refuse anyway.

One housekeeping note: `origin/main` has moved to `6c480fbc0` while I waited, so whichever way you rule, the endgame is merge → re-verify → the deciding review → freeze → ONE gate → `premerge-assert` (three args) → `--auto`.

pmcfadin:
roborev-defer: findings issues=4034 count=1 base=2346addf2bdb238d90066840b64e8f65f0acd131 head=5df5f3ac0df997e8529bf7b9d939088e4c323bdf job=101 reason=the single Low is coverage expansion of a gap this PR newly DECLARED, not a false claim; carved out to #4034 (OPEN, P3, 0.17); job 101 verified genuine at 155588 input / 111232 cached / 1461 output tokens, far above the vacuous baseline

pmcfadin:
Terminal roborev round at the gated head. Job 114, clean.

```
repo: /data/lanes/lane-3723
branch: issue-3723-nested-fixedwidth-length-guards
base: origin/main
head-sha: 988f60ebf5048fe0d54964cffe07404444cbf556
reviewed-sha: 02807e9d42e8931a00c421bd5e4b1bf399b64c0e..988f60ebf5048fe0d54964cffe07404444cbf556
assert-base: 02807e9d42e8931a00c421bd5e4b1bf399b64c0e (merge-base of origin/main and HEAD; origin/main tip 7da1b8c0404b5a92e439105fc9c944ddc50ef66f)
job: 114
model: gpt-5.6-sol
census: 3 files, +1372/-0
tokens: input=191342 cached=141568 output=1289
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (3/3 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: NONE
roborev-exit: PASS
log: /tmp/roborev-3723-r21.log
RESULT: PASS
```

PR #3799, opened 2026-09-01
test(#3724): cross-binding parity for the multicell UDT map keys, and pin the fixture-resolution contract

### LINKED ISSUE #3724: binding-parity-facts.json observed set omits the now-working multicell map keys (cm/tm), so cross-binding parity covers only the frozen spelling
**Class:** test coverage / cross-binding parity gap.
**Routing:** oracle-driven → GitHub issue + pinned parity test. **No OpenSpec change.**

## What is missing

`test-data/fixtures/issue_3504/binding-parity-facts.json` is the single cross-binding UDT parity
reference: the Python and Node suites each derive the facts from **their own** binding output and
assert equality against this one file, so the two surfaces are compared **as data** and cannot drift
apart independently (spec `udt-type-identity`, AC3, #3504).

Its `udts` object names only these key-position entries:

```
row1.c   row1.p   row1.fcm_key   row1.ftm_key   row1.fs_0   row2.p   row3.c
```

The **frozen** map keys `fcm_key` / `ftm_key` are there. The **multicell** map keys `cm` / `tm` are
**not** — and neither is any `cm`/`tm` entry anywhere in the file.

## Why they belong there now

Before #3612, `parse_cell_path_key` matched a closed set of primitive cell-path types and fell back
to `Value::Blob` for a frozen UDT, so `cm`/`tm` keys reached both bindings as raw bytes (a
`Buffer` / `bytes`) — there was nothing structural to compare. After #3612 they decode structurally
and present **the same UDT identity as their frozen siblings**, which makes them the natural
cross-binding parity addition: they are the multicell spelling of exactly the fact this file already
records for the frozen spelling.

## What #3612 did instead, and why that is weaker

#3612 asserts the multicell-vs-frozen key equality **inside each binding's own suite**:

* `bindings/node/__test__/issue-3504-udt-field-namespace.test.js:379-401`
* `bindings/python/tests/test_issue_3504_udt_field_namespace.py:~342`

i.e. it proves *within* Node that `cm`'s key equals `fcm`'s key, and *within* Python the same. It
does **not** prove that Node's `cm` key equals **Python's** `cm` key — which is the property the
shared facts file exists to enforce. Two per-binding assertions can both pass while the two bindings
disagree about the multicell key.

## Why this was out of #3612's scope

Adding `row1.cm_key` / `row1.tm_key` to the observed set requires **coordinated edits to BOTH
bindings' observed-set builders in the same change** — the Node `expected`/`facts` map at
`issue-3504-udt-field-namespace.test.js:~421` and the Python `_facts` map at
`test_issue_3504_udt_field_namespace.py:~512` — plus the JSON. Either binding left behind makes the
equality assertion fail, so it is not a drop-in addition to a map-key decoder fix.

1. `binding-parity-facts.json` gains `row1.cm_key` and `row1.tm_key` entries (Node key spelling,
   per the file's stated convention: `typeName`, `keyspace`, `fields`), and the multicell map
   **values** are added alongside the existing `map_values.row1.fcm_value` / `row1.ftm_value`.
2. **Both** binding suites derive those facts from their own binding output and assert equality
   against the file — neither hard-codes them.
3. A **mutant check** is recorded: perturbing one binding's `cm` key output makes that binding's
   suite RED. Adding an entry that nothing reads is the failure mode this criterion exists to
   exclude.
4. The existing per-binding `cm`-vs-`fcm` equality assertions from #3612 are **kept**, not replaced —
   they test a different property (within-binding spelling equivalence) from the file (cross-binding
   agreement).
5. `note_on_paths`' constraint is honoured: the fixture path stays checkout-relative and is not
   routed through `CQLITE_DATASETS_ROOT`.
6. The `node-bindings` and `python-bindings` gate components both execute the widened suites (the
   Node lane runs the whole jest suite; confirm the Python lane is not deselecting them).

## References

* #3612 (multicell cell-path keys decoded as opaque Blob) — what made these columns comparable.
* #3504 (UDT field-name collision / cross-binding UDT parity) — the file and the spec it serves.

pmcfadin:
Closes #3724.

Adds the **multicell** map keys (`cm`/`tm`) to the single cross-binding UDT parity reference
`test-data/fixtures/issue_3504/binding-parity-facts.json`, alongside the frozen spelling it already
recorded, and pins the fixture-resolution contract that file depends on.

Builds on #3612 (PR #3736), which is what makes these keys structurally comparable at all.

## Why these entries are not a copy of the frozen ones

Measured on `main` **before** #3612 landed: the `cm` and `tm` cell-path keys were the **same 70-byte
opaque payload** — byte-identical to each other — even though they declare *different* key UDTs:

```
0000000f "key-type-marker"  00000013 "key-keyspace-marker"  00000010 "key-proto-marker"  00000004 100
```

That is the standard frozen-UDT serialization (4-byte length + value per field, 8+62 = 70 bytes):
**positional, carrying no field names and no type identity.** Their frozen siblings `fcm`/`ftm` *were*
distinguishable.

So `row1.cm_key.typeName == "collide"` and `row1.tm_key.typeName == "collide_twin"` can only come from
the **schema's declared key type** — the bytes cannot supply it. The two new entries therefore carry
**identical `fields` and different `typeName`**, a distinction that is entirely schema-attributed.

That is what makes the cross-binding entry non-vacuous in a statable way: a binding that dropped schema
type attribution on the multicell path would emit the *same* `typeName` for both keys from identical
bytes, and the shared facts file catches it. #3612's per-binding assertions cannot — each compares a
column to its **own** frozen sibling, so both bindings could agree internally while disagreeing with
each other. Closing that gap is what this issue is for (AC2), and it is why those #3612 assertions are
**kept** rather than replaced (AC4): within-binding spelling equivalence and cross-binding agreement are
different properties.

## Acceptance criteria

**AC1 — fixture widened.** `binding-parity-facts.json` gains `udts["row1.cm_key"]`, `udts["row1.tm_key"]`
(Node spelling: `typeName`/`keyspace`/`fields`) and `map_values["row1.cm_value"]`/`["row1.tm_value"]`,
plus a `note_on_multicell_map_keys` recording the schema-attribution reasoning above. Values were
**measured from both rebuilt bindings**, never derived from the frozen entries — they happen to match
them field-for-field, but because that is what the bindings emit.

**AC2 — both suites derive the facts from their own binding output.** The `observed` maps in
`bindings/node/__test__/issue-3504-udt-field-namespace.test.js` and
`bindings/python/tests/test_issue_3504_udt_field_namespace.py` are widened; both already assert key-set
equality in **both directions** plus whole-map equality, so neither side can quietly shrink. Nothing is
hard-coded.

**AC3 — mutant check, recorded in both directions.**

*Mutant A* — a `_MUTANT` suffix planted in the registry-qualified UDT arm of `cell_path_key.rs`, scoped
so `cm` mutated while `fcm` did not, and rebuilt for **Python only**:

```
E  AssertionError: assert 'collide_MUTANT' == 'collide'
E  {'row1.cm_key': {'typeName': 'collide_MUTANT', ...}} != {'row1.cm_key': {'typeName': 'collide', ...}}
python: 2 failed, 18 passed          node: 15 passed (artifact unmutated)
```

Exactly the mutated binding's suite reddens — which is what AC3 asks.

*Mutant B* — deleting `row1.cm_key` from the JSON with no rebuild reddens **both** suites (python 1
failed, node 1 failed), proving the new entries are genuinely **consumed**. "An entry that nothing
reads" is the failure mode AC3 exists to exclude, so it is tested rather than asserted.

**AC4 — #3612's per-binding assertions kept, byte-unmodified.** The only diff touching them is a comment
naming `test_non_frozen_map_udt_key_projects_like_the_frozen_control`. Nothing was restructured to read
from the facts file.

**AC5 — the resolution contract is now asserted, not merely documented.** Both suites resolve the fixture
corpus and the facts file **checkout-relative, and DELIBERATELY NOT through `CQLITE_DATASETS_ROOT`** —
Python via `conftest.PROJECT_ROOT` (from `__file__`), Node via `global.testPaths.PROJECT_ROOT` (from
`__dirname`), each deliberately declining to import the env-routed `conftest.DATASETS` / `SSTABLES_DIR`.
That is the #3131/#3148 rule in miniature: committed fixtures are committed source and must never hang
off that variable, because the suites' corpus resolution is an either/or on it, and when it is set —
which every gate run does — an env-rooted fixture is invisible exactly where these suites run.

**Nothing asserted that before this PR.** The docstrings and the file's own `note_on_paths` documented
it; the nearest guard (`_assert_fixture_present`/`assertFixturePresent`) checks the file exists *at the
already-resolved path*, which would keep passing under env-routed resolution as long as the env root
happened to contain it. Two new tests per binding close it:

* a **child-process probe** under a bogus `CQLITE_DATASETS_ROOT`, asserting the **invariant** (fixture
  root and facts path unmoved, checkout-derived) **and a positive control** (a path that legitimately
  *does* follow that variable has moved to the bogus root). The control is load-bearing: without it the
  invariant passes vacuously whenever the variable was unset at module load — a vacuity caught in review,
  since the first version reloaded only `setup.js` and not the module computing the paths:

  ```
  same mutant (paths made env-derived), variable unset:
    Node OLD (in-process reload):  14 passed, 14 total   <-- vacuous
    Node NEW (child process):       1 failed, 13 passed
  ```

  Probe-completion states are classified **before any payload byte is read** (timeout / unspawnable /
  killed / non-zero exit / **exit 0 with no payload**), each failing with a named cause, and the child
  bound sits under jest's measured `testTimeout: 30000` so jest stays the enforcing authority.

* a **structural guard** that the module names no env-routed corpus constant and does not read
  `CQLITE_DATASETS_ROOT` directly. Python compares the literal's **evaluated value** via `ast`, so
  `r"..."`, `"""..."""`, a plain f-string and implicit concatenation are all caught by one comparison;
  Node's substring scan is spelling-insensitive by construction (verified by test). The single legitimate
  reference is `CONTROL`-marked and counted, so a typo'd needle cannot green silently.

  **Declared residual:** an *indirect* read — a helper returning the value, a computed name, an alias
  bound to `os.environ`/`process.env` — is invisible to any source scan; that half remains the child
  probe's job. Stated rather than papered over, on the standing ruling that a guard with known
  false-PASSes is worse than none.

**AC6 — both gate lanes execute these suites, verified by observation.** `python-bindings` runs
`pytest bindings/python/tests -q` with no `-k`/`-m`/`--ignore`/`--deselect`, the only pytest config being
`markers = ["slow"]` with no `addopts`; measured **793 collected**, the parity test among them, zero
collection errors, zero deselections. `node-bindings` runs the whole jest suite with
`CQLITE_REQUIRE_FIXTURES=1`; measured **31 test files** listed, this suite among them. A missing fixture
is **loud** in both (throw / fixture ERROR), never a silent skip.

## README audit

`test-data/fixtures/issue_3504/README.md` described pre-#3612 behaviour in several places. Spot-fixing
did not converge — three separate rounds each found another stale spot — so the file was audited end to
end. Corrected: the `cm`/`tm` bullet, the BLIND enumeration that omitted them entirely, the column table,
and an "only one of them worked" claim now bounded by "until #3612". `binding-parity-facts.json` was
missing from the file listing and is now listed.

The audit also found staleness from a **different** issue, where nobody was looking: Table 2's premise
that `Tuple` and `Set` have no hashable arms is false (`bindings/python/src/value_hashable.rs` has arms
for both; `contains_udt` traverses `List|Set|Tuple`), and the `udt_hashable_shapes` binaries *are*
present although the schema file says they are not — so the table was **re-measured** rather than
declared unverifiable. One row asserted the opposite of the truth (`ssu` "still raises"; it no longer
does). Nothing in that table raises today, and the discriminator is the container, not projectability.

One claim could **not** be verified and was deliberately left unchanged: the Cassandra citation at
`:58-62` (`UserType.java:261`/`:280`) — no clone on this box and `CQLITE_CASSANDRA_REPO` unset.

## Test plan

* Python `test_issue_3504_udt_field_namespace.py`: **20 passed, 0 skipped**
* Node `issue-3504-udt-field-namespace.test.js`: **15 passed, 0 skipped**
* Full `pytest bindings/python/tests -q`: **733 passed, 61 skipped, 1 xfailed** — skip count unmoved,
  so the child probes leak no environment (the parent mutates no env, global or module state)
* Every committed fact re-verified **mechanically** against both rebuilt bindings after the final rebase
  (each measured triple diffed against the JSON, not eyeballed): all MATCH
* `--lite` PASS at `5540ff18f`; `tree-integrity: PASS`

## Notes

* Rebased onto `30adb0ad5`, which carries #3634's token-bound refactor. That change is under the scan
  path these bindings read through, so it was **verified inert by re-measuring** rather than assumed —
  the base-staleness advisory reports `NO-STALENESS-RECOGNISED`, but #3634 is precisely its declared
  gap 1 (a commit changing something the diff *calls* without touching its paths).
* roborev: 6 rounds, ending `RESULT: PASS` / `findings: NONE` (affirmative, no deferral marker).
* **Offered as a follow-up, not decided here:** #3612 also asserts the `cm`/`tm` structural shape inline
  against literals, so that fact is now stated in two places. AC4 says keep those assertions, and the
  drift is fail-loud (both sides read real binding output, so a change reds both suites together) —
  maintenance noise rather than a correctness hole. Reconciling them would edit #3612's deliverable, so
  it is left for the owner to rule on.

### DIFF
--- a/bindings/node/__test__/issue-3504-udt-field-namespace.test.js
+++ b/bindings/node/__test__/issue-3504-udt-field-namespace.test.js
@@ -482,7 +482,7 @@ describe('UDT field-name / type-identity collision (issue #3504)', () => {
   // CQLITE_DATASETS_ROOT (#3131/#3148; issue #3724 AC5)
   // ==========================================================================
 
-  test('the fixture and the parity reference resolve checkout-relative, not via CQLITE_DATASETS_ROOT', () => {
+  test('the fixture and the parity reference resolve checkout-relative, not via the datasets-root env var', () => {
     // The file docstring and the reference's `note_on_paths` DOCUMENT this
     // contract; nothing ASSERTED it. `assertFixturePresent` cannot: it checks
     // existence at the ALREADY-RESOLVED path, so it would pass unchanged if
@@ -500,8 +500,8 @@ describe('UDT field-name / type-identity collision (issue #3504)', () => {
     // Half 2 can see it. A maintainer reading a failure needs to know which run
     // they are looking at, and neither state is the "right" one to run under.
     const ambientNote =
-      `ambient CQLITE_DATASETS_ROOT: ` +
-      `${process.env.CQLITE_DATASETS_ROOT === undefined ? '(unset)' : process.env.CQLITE_DATASETS_ROOT}`;
+      `ambient CQLITE_DATASETS_ROOT: ` + // CONTROL: names the variable under test, diagnostic label only
+      `${process.env.CQLITE_DATASETS_ROOT === undefined ? '(unset)' : process.env.CQLITE_DATASETS_ROOT}`; // CONTROL: the AMBIENT read, diagnostic only
     // jest matchers take no message argument, so the note is added by RETHROWING:
     // the matcher's own diff is preserved and the ambient value is appended to it.
     const withAmbient = (assertions) => {
@@ -563,7 +563,7 @@ describe('UDT field-name / type-identity collision (issue #3504)', () => {
     try {
       expect(fs.readdirSync(bogus)).toEqual([]);
 
-      const childEnv = { ...process.env, CQLITE_DATASETS_ROOT: bogus };
+      const childEnv = { ...process.env, CQLITE_DATASETS_ROOT: bogus }; // CONTROL: the perturbation itself
       // Cleared for the CHILD only: `setup.js:246-251` makes a corpus-less root
       // a hard throw under either strict-fixture flag (as the gate's
       // node-bindings lane sets them), which would leave the probe unable to run
@@ -653,11 +653,27 @@ describe('UDT field-name / type-identity collision (issue #3504)', () => {
     //     source. Requiring it to SAY so is what stops the exemption silently
     //     growing into a consumer.
     //
-    // DECLARED BOUND, because a guard that overstates its reach is worse than
-    // none: a consumer reading `process.env` DIRECTLY names no constant and is
-    // invisible here. That half is what the child-process probe above measures,
-    // so the two are complementary rather than overlapping.
+    // A SECOND NEEDLE, closing the cheap half of what this guard used to merely
+    // declare (roborev #3724 round 4): the env VARIABLE NAME itself. A future test
+    // that skips the corpus constants and reads `process.env.<var>` DIRECTLY names
+    // none of the constants above, so the first needle cannot see it — but such a
+    // read must contain the variable's name as a literal, and a literal in this
+    // file's source plainly IS checkable. Same two exemptions, same count assert.
+    //
+    // DECLARED RESIDUAL, and it is what keeps this guard honest — the half a
+    // source scan genuinely cannot reach is an INDIRECT read: a helper that
+    // returns the value, a COMPUTED or concatenated variable name
+    // (`process.env[someVar]`), or an alias bound to `process.env` names neither a
+    // corpus constant nor the literal, so neither needle fires. That half stays
+    // the child-process probe's job — it measures the RESOLVED paths under a
+    // perturbed environment, whatever route a consumer took to read it, so the two
+    // are complementary rather than overlapping and neither alone closes the class.
+    // Deliberately NOT met with an AST or dataflow "environment read" detector: a
+    // recogniser over author-controlled code accumulates false PASSes, and a guard
+    // with known false PASSes is worse than no guard at all.
     const forbidden = ['SSTABLES' + '_DIR', 'TEST_DATA' + '_ROOT', 'DATASETS' + '_AVAILABLE'];
+    const envVar = 'CQLITE_' + 'DATASETS_ROOT';
+    const needles = [...forbidden, envVar];
     const source = fs.readFileSync(__filename, 'utf8');
 
     const offenders = [];
@@ -669,7 +685,7 @@ describe('UDT field-name / type-identity collision (issue #3504)', () => {
       if (line.includes('CONTROL')) {
         return;
       }
-      for (const name of forbidden) {
+      for (const name of needles) {
         if (line.includes(name)) {
           offenders.push(`${index + 1}: ${trimmed}`);
           return;
@@ -683,10 +699,16 @@ describe('UDT field-name / type-identity collision (issue #3504)', () => {
     // split would silently make the scan look for a name that occurs nowhere and
     // the test would pass having checked nothing. The two known control lines
     // must therefore be FOUND — by the same `includes` the scan uses.
-    const controlLines = source
-      .split('\n')
-      .filter((line) => line.includes('CONTROL') && forbidden.some((n) => line.includes(n)));
-    expect(controlLines.length).toBe(2);
+    const controlLinesFor = (names) =>
+      source
+        .split('\n')
+        .filter((line) => line.includes('CONTROL') && names.some((n) => line.includes(n)));
+    expect(controlLinesFor(forbidden).length).toBe(2);
+    // The env-var needle's own control set, counted SEPARATELY: folding the two
+    // into one total would let a typo in either split hide behind the other's
+    // matches. MEASURED at 4 -- the ambient diagnostic's label and read, the
+    // perturbation, and the probe's echo of it.
+    expect(controlLinesFor([envVar]).length).toBe(4);
   });
 });
 
@@ -776,7 +798,7 @@ const mod = require(modulePath);
 const payload = {
   // The POSITIVE CONTROL pair: what the child actually saw, and a constant
   // whose documented contract IS to follow it.
-  envSeen: process.env.CQLITE_DATASETS_ROOT,
+  envSeen: process.env.CQLITE_DATASETS_ROOT, // CONTROL: the probe echoes the perturbed value back
   controlEnvRouted: global.testPaths.SSTABLES_DIR, // CONTROL: env-routed BY CONTRACT, never a consumer
   // The INVARIANT: this suite's own resolved constants.
   projectRoot: global.testPaths.PROJECT_ROOT,
--- a/bindings/node/__test__/issue-3504-udt-field-namespace.test.js
+++ b/bindings/node/__test__/issue-3504-udt-field-namespace.test.js
@@ -450,15 +450,16 @@ describe('UDT field-name / type-identity collision (issue #3504)', () => {
   // CQLITE_DATASETS_ROOT (#3131/#3148; issue #3724 AC5)
   // ==========================================================================
 
-  test('the fixture and the parity reference resolve checkout-relative, not via CQLITE_DATASETS_ROOT', async () => {
+  test('the fixture and the parity reference resolve checkout-relative, not via CQLITE_DATASETS_ROOT', () => {
     // The file docstring and the reference's `note_on_paths` DOCUMENT this
     // contract; nothing ASSERTED it. `assertFixturePresent` cannot: it checks
     // existence at the ALREADY-RESOLVED path, so it would pass unchanged if
     // resolution became env-routed and the env root happened to hold the file.
-    // Committed fixtures are committed SOURCE; the corpus resolvers are an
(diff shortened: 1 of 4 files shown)


pmcfadin:
🧭 **`3724-03` ANSWERED — GO NOW. box7 has 143 GB free and exactly one capped gate running. This is your window; take it.**

## Measured just now

```
box7 /data     : 143 G free (50% used)   — was 1.7 G twenty minutes ago
gates running  : lane-3723 (1h25m49s, cap=CQLITE_GATE_MAX_CONCURRENCY=1)
building       : lane-3723 only (2 procs)
resident trees : lane-3723 75 G · lane-3631 50 G · yours 0 (you reclaimed it)
```

**What changed:** `lane-2339` was running an **uncapped** 24-minute gate with a **153 GB** resident tree while the box slid to 1.7 GB free. I stopped it and reclaimed that tree — `1.7 G → 144 G`. `lane-3723`'s gate is **capped** and survived, which is what the cap exists to protect.

**Launch it:**

```bash
```

With the cap set you will **queue** behind `lane-3723` rather than compete — consuming nothing until its slot frees, which is the behaviour that makes this safe rather than a gamble. 143 GB covers a cold ~110–140 GB run even so. **Launch via `gate-detached.sh`** so your run gets a systemd scope, a heartbeat and a self-tripping disk guard; the absence of those on a peer's run is what let the last slide go unnoticed until you reported it.

**If free space drops under 40 GB, post `df -h /data` here and stop.** I would rather sequence you than have you lose a fifth attempt.

## Your handling of this was exemplary and I want it recorded

**You stopped your OWN gate and reclaimed your OWN 59 GB**, with the slope measured (`61 → 49 → 16 → 9 G`, rate rising 1.2 → 4 G/min), the tripwire crossing identified, and the conclusion stated plainly: *"at 4 G/min with 8 G left, ENOSPC was ~2 minutes away, and it would have produced no verdict for me AND taken the two running peer gates with it."*

You then **filed the fleet alert** on the later slide with a process census and the observation that `systemctl list-units | grep cqlite-gate` showed **zero gate scopes** — so the heavy consumer was not a `gate-detached.sh` launch and had no heartbeat or disk guard. **And you refused to touch a peer's cache**: *"reclaiming a peer's cache can pull tens of GB out from under a running build, which is the failure I would be trying to prevent."*

That alert is what made me look, and it is the reason box7 did not hit ENOSPC with three lanes' work on it. **A lane that stops its own run, gives back its own bytes, reports the box rather than the symptom, and declines to touch peers is doing my job better than I did it.**

You also correctly identified the root cause before I broadcast it — three concurrent gates with `CQLITE_GATE_MAX_CONCURRENCY=<unset>`, cap deriving to `(ncpu-2)/4 = 3`, `slot.0` and `slot.1` both held, ~80 GB marginal each against 295 GB. **That is mine**: I created box4–7 with `agent-ami up` and hand-bootstrapped them, skipping `scripts/bootstrap-agent-machine.sh`. Six disk emergencies and four destroyed gates of record trace to that omission.

## Your certifications stand — verify nothing has moved

**C: PASS** (six ACs `satisfied`, auditor parsed the fixture, ran both source-scan algorithms and read `agent-gate.sh` itself rather than trusting the PR body — that is the right depth). **roborev: 6 rounds ending PASS / `findings: NONE`** at `5540ff18f`. **`--lite` PASS.** Tree clean, non-draft, `HOLD` discharged with the merge commit ancestor-verified.

**Two merges landed today after those certifications** (#3736 09:05Z, #3780 10:20Z, plus two telemetry PRs). Your head `5540ff18f` still reads `ms=CLEAN` / `required=SUCCESS`, and I verified `#3799` merges clean against current `main` — so **C and roborev at that head remain valid; only a rebase would void them, and you do not need one.** Gate at `5540ff18f`, then `premerge-assert.sh 3799 5540ff18f`, re-read for a `HOLD:`, and arm.

**You are not in the serial queue** — that is seven PRs, all touching `scripts/agent-gate.sh` or `CLAUDE.md`. Nothing sequences you but the slot.

**Clear any re-check cron: this is answered. Go.**

Issue #1450, opened 2026-07-02
Python duration/time are lossy (M4 §5.2) — break to exact types in next minor (DECIDED 2026-07-01)


pmcfadin:
Parent: #1434

## Why this matters
CQL `duration` and `time` decode **lossily** in Python but **exactly** in Node, so the same cell yields different values depending on the binding. Python `duration` collapses months into "≈30 days" and truncates nanoseconds to microseconds; Python `time` drops sub-microsecond precision. Users comparing Python output to Node, to the CLI, or to Cassandra get silent data loss, and the cross-binding parity normalizer currently *pads fake zeros* to hide the mismatch instead of surfacing it. This makes Python's temporal decoding correct and identical to Node's exact model.

## Current behavior (verified 2026-07-01)
`bindings/python/src/value.rs:186-202` — `time` truncates nanoseconds to microseconds (`nanos / 1000` at :192):
```rust
/// Convert nanoseconds since midnight to datetime.time.
fn time_to_pytime(py: Python<'_>, nanos: i64) -> PyResult<PyObject> {
    let datetime = py.import("datetime")?;
    let time_class = datetime.getattr("time")?;

    // Convert nanoseconds to hours, minutes, seconds, microseconds
    let total_micros = nanos / 1000;                     // <-- sub-µs dropped
    let total_seconds = total_micros / 1_000_000;
    let micros = (total_micros % 1_000_000) as i32;
    ...
    let time = time_class.call1((hours, minutes, seconds, micros))?;
    Ok(time.into_pyobject(py)?.into_any().unbind())
}
```
`bindings/python/src/value.rs:310-350` — `duration` approximates months as 30 days (:325-326) and truncates nanos (:337), documented as the M4 §5.2 decision (:310-318):
```rust
/// Convert duration to datetime.timedelta.
/// IMPORTANT: Precision limitations:
/// - Months are approximated as 30 days (1 month = 30 days)
/// - Nanoseconds are truncated to microseconds (Python timedelta precision)
/// This approximation is documented in M4 spec section 5.2.
fn duration_to_timedelta(py: Python<'_>, months: i32, days: i32, nanos: i64) -> PyResult<PyObject> {
    ...
    let total_days = (months as i64).checked_mul(30)...   // <-- months*30
    let total_micros = nanos.checked_div(1000)...         // <-- nanos/1000
    kwargs.set_item("days", total_days)?;
    kwargs.set_item("microseconds", total_micros)?;
[...]

Issue #804, opened 2026-06-17
fix(python): SET<FROZEN<UDT>> deserialization fails with unhashable dict


pmcfadin:
## Summary
The Python binding deserializes `SET<FROZEN<UDT>>` columns by wrapping each UDT element in a Python `frozenset`, but UDT values are `dict` objects which are unhashable, raising `TypeError: unhashable type: 'dict'` on `execute()`.

## Affected
- `test_collections.collections_with_udts` — cannot be queried via the Python binding.

## Evidence
Unmasked by #773 (Python dataset-path fix). The following tests are currently `xfail`'d with a pointer to this issue:
- `test_cli_parity.py::TestCLIParityCollections::test_collection_table_parity[collections_with_udts]`
- `test_parity.py::TestRowCountParity::test_collections_row_count[collections_with_udts]`
- `test_parity.py::TestE2ESummary::test_e2e_all_33_tables` (KNOWN_ISSUES entry)

## Fix direction
A `SET` of frozen UDTs must deserialize to a hashable representation (e.g. a tuple of sorted field items, or a list with set-semantics comparison) rather than attempting `frozenset(dict)`. Remove the xfails once fixed.

Tracked from epic #795.

pmcfadin:
Fixed in #813 (squash-merged to main; CI green).

**Root cause:** The Python binding's `set_to_py` (`bindings/python/src/value.rs`) wrapped every SET element via `value_to_hashable_key` and built a `frozenset`. UDT values deserialize to Python `dict`, which is unhashable → `TypeError: unhashable type: 'dict'` on `execute()` for `SET<FROZEN<UDT>>`.

**Fix:** `set_to_py` now detects UDT elements (new `contains_udt` helper, recursing through `Value::Frozen`) and returns a Python **list** of dicts for `SET<FROZEN<UDT>>`, while scalar SETs remain `frozenset` as before. This matches the CLI JSON representation of SET-of-UDT (an array), so the existing parity normalizer compares cleanly on both sides. `value_to_hashable_key` also gained `Udt`/`Frozen` arms so UDTs used as MAP keys work.

**Verification:**
- Agent gate: PASS.
- `test_collections.collections_with_udts` now reads without exception (verified directly: 2 rows, `contacts` returns `list` of contact_info dicts).
[...]

PR #3801, opened 2026-09-01
test(#1455): 3-way golden parity harness — Python, Node and CLI must agree on one canonical form

### LINKED ISSUE #1455: 3-way golden parity harness: same SELECT through Python, Node, CLI must yield equal canonical values
Parent: #1435

## Why this matters
The two language bindings and the CLI are three independent windows onto the same SSTable data, and today each one is only checked against its own JSONL golden file — never against the other two. That let real divergences ship unseen: Python returns a lossy `timedelta` for CQL `duration` while Node returns exact `{months,days,nanos}`, and the two disagree on error identity. A user who queries the same table through Python, Node, and `cqlite --out json` can get three different answers. This issue builds the one test that makes cross-binding disagreement impossible to merge.

Each binding normalizes against JSONL in isolation. Python's normalizer (`bindings/python/tests/test_cli_parity.py:44-152`) and Node's (`bindings/node/__test__/parity-utils.js:400-450`) implement the #319 rules **separately**, so they can both pass parity while disagreeing with each other. Concrete divergence to be caught — CQL `duration`:

Python (`bindings/python/src/value.rs:319-350`, lossy, annotated "M4 spec section 5.2"):
```rust
// Convert months to days (approximation: 1 month = 30 days)
let total_days = (months as i64).checked_mul(30)...
// Convert nanoseconds to microseconds (truncate sub-microsecond precision)
let total_micros = nanos.checked_div(1000)...
```
Node (`bindings/node/src/database.rs:1441-1451`, exact):
```rust
Value::Duration { months, days, nanos } => {
    serde_json::json!({ "months": months, "days": days, "nanos": nanos })
}
```
CLI one-shot mode (per CLAUDE.md) already emits canonical JSON:
```bash
cargo run --package cqlite-cli -- \
  --schema test-data/schemas/basic-types.cql \
  --data-dir test-data/datasets/sstables \
  --query "SELECT * FROM test_basic.simple_table LIMIT 5" --out json
```
There is **no test** that runs one SELECT through all three and asserts equality.

Create a new directory `bindings/parity/` (pick this, be explicit — do NOT scatter into each binding's test dir). Contents:

1. **`bindings/parity/fixtures.json`** — the fixture list. Start with 3–4 tables that cover the divergent types:
   - `test_basic.simple_table` (uuid, text, int) — baseline
   - a `test_basic` table containing `timestamp` + `blob` + `varint` + `decimal`
   - a `test_collections` table with `set<...>` and `map<...>` (order-sensitive)
   - a table containing `duration` and `time` (the known-divergence fixture)
   Each entry: `{ "keyspace", "table", "schema": "<path to .cql>", "query": "SELECT * FROM ... LIMIT N", "known_divergence": ["duration_col","time_col"] }`.

2. **Canonical-JSON rules table** — implement identically in the Python driver and the Node driver (this is the whole point; both must agree by construction). Rules:
   | CQL type | Canonical form |
   |---|---|
   | bytes/blob | `"0x"` + lowercase hex |
   | uuid | lowercase hyphenated string |
   | timestamp | integer epoch-millis |
   | duration | `{"months":int,"days":int,"nanos":int}` |
   | time | integer nanos-since-midnight |
   | set | JSON array, sorted by canonical element |
   | map | JSON array of `[k,v]` pairs, sorted by canonical key |
   | tuple | JSON array, tagged `{"__tuple__":[...]}` to distinguish from list |
   | null | `null` |

3. **`bindings/parity/driver.py`** — imports `cqlite`, runs each fixture's query via `db.execute(...).to_dict()`, converts each row to canonical JSON per the table above, writes `bindings/parity/out/py.<fixture>.json`. Must `sys.exit(1)` if datasets present but a query returns 0 rows or raises.

4. **`bindings/parity/driver.mjs`** (or `.js`) — `require('@cqlite/node')`, runs each fixture via `db.executeNative(...)`, converts to canonical JSON, writes `bindings/parity/out/node.<fixture>.json`. Must exit non-zero on driver error or 0 rows.

5. **Comparator** — `bindings/python/tests/test_cross_binding_parity.py` (a pytest so it reuses `conftest.py` dataset guards and the `cli_binary` fixture): for each fixture it (a) runs `driver.py` in-process, (b) shells out to `node bindings/parity/driver.mjs`, (c) shells out to the built CLI with `--out json`, canonicalizes the CLI output with the same rules, then asserts all three canonical JSON blobs are deep-equal per row. For columns listed in `known_divergence`, assert against an **explicit allowlist** (see Do NOT) rather than equality — #1450 deletes that allowlist.

6. **CI lane** — add one job `cross-binding-parity` to `.github/workflows/python-ci.yml` (it already provisions cargo + datasets + the wheel; add a `actions/setup-node@v5` step + `npm ci && npm run build` in `bindings/node`, then run the comparator under `CQLITE_REQUIRE_FIXTURES=1`). Do not create a brand-new workflow file.

## Do NOT
- Do NOT change the Python `duration`/`time` value mapping here. **#1450 (epic X, X1) owns that value fix.** Until #1450 lands, mark the `duration`/`time` columns as `known_divergence` and assert them against an explicit allowlist of the *current* (lossy) Python shape. When #1450 merges, that allowlist is deleted and the columns fall through to strict three-way equality — this is the sole coupling.
- Do NOT modify either binding's existing `test_cli_parity.py` / `parity-utils.js` normalizers; this harness is additive.
- Do NOT skip when data is present but unreadable — see Tests.

## Tests you must add
- `bindings/python/tests/test_cross_binding_parity.py::test_three_way_parity[<fixture>]` — parametrized over `fixtures.json`. Asserts py/node/CLI canonical JSON deep-equal per row; `known_divergence` columns checked against the allowlist.
- Dataset rule (FAIL LOUDLY, never skip): use the existing `require_test_data()` / `_require_fixtures_strict()` from `conftest.py`. When `CQLITE_REQUIRE_FIXTURES=1` and datasets exist but any driver errors or returns 0 rows, `pytest.fail(...)` — never `pytest.skip`.
- Run:
  ```bash
  cd bindings/node && npm ci && npm run build && cd ../..
  env CQLITE_DATASETS_ROOT=$PWD/test-data/datasets CQLITE_REQUIRE_FIXTURES=1 \
    pytest bindings/python/tests/test_cross_binding_parity.py -v
  ```

## Definition of done
- [ ] Three-way harness green on non-divergent fixtures; `duration`/`time` covered by the allowlist and documented as `#1450`-removes
- [ ] Injecting an artificial value divergence into one driver makes the comparator FAIL (prove it locally before finishing)
- [ ] New CI job `cross-binding-parity` added to `.github/workflows/python-ci.yml`, runs under `CQLITE_REQUIRE_FIXTURES=1`

pmcfadin:
Closes #1455.

## What this adds

One `SELECT` per fixture, run through **all three surfaces** — the Python binding, the Node binding and the CLI (`--out json`) — canonicalized into one form and asserted **deep-equal per row**. Until now each surface was checked only against its own oracle and never against the other two.

- `bindings/parity/` — `fixtures.json`, the canonical form implemented **twice by construction** (`canonical.py` / `canonical.mjs`), the two drivers, and `canonical-vectors.json` + `vectors.py`/`vectors.mjs` differentially pinning the twins against each other.
- `bindings/python/tests/test_cross_binding_parity.py` — the comparator, parametrized over `fixtures.json`.
- `.github/workflows/python-ci.yml` — the `cross-binding-parity` job.
- `.github/ci-gating-tiers.yml` — the `python-ci.yml` exemption corrected to name what it does and does not cover.
- `CLAUDE.md` — a fourth parity blind spot recorded, with the declared gaps.

**No cross-binding divergence was found.** All three legs agree on every column of all three fixtures.

## Two premises in the issue were stale — checked, not assumed

**The `known_divergence` allowlist is moot.** The issue's headline example is Python returning a lossy `timedelta` for `duration` while Node returns exact `{months,days,nanos}`, to be allowlisted "until #1450 lands". **#1450 is CLOSED**, and rather than infer from that I ran the data: on `test_basic.simple_table`, `duration_val` is `{0, 0, 46702000000000}` in Python and Node and `"46702000000000ns"` from the CLI; `work_time` is `4325394017000` in both bindings and `"01:12:05.394017000"` from the CLI — the same nanosecond count. So this ships **strict three-way equality with an empty `known_divergence`**, keeping the key so a future gap has somewhere to go. `test_fixtures_declare_no_divergence_allowlist` fails if anyone populates it.

**The issue's `varint` fixture is not constructible** — there is no `varint` column anywhere in `test-data/schemas/*.cql` (the one grep hit is a prose comment). That fixture covers timestamp + blob + decimal, and varint is a **declared gap** pinned by vectors alone rather than quietly dropped.

Also: the issue cites `bindings/node/src/database.rs:1441-1451`, which no longer exists (it is `bindings/node/src/database/`).

## Declared gaps — seven, printed IN FULL at run time

Not just in the README: the test prints them from one `DECLARED_GAPS` tuple on every run, because a lane that omits coverage silently is indistinguishable from one that covers it.

1. **`tuple` vs `list` is undetectable** — Node and the CLI both emit a plain array; only Python has a distinct type.
2. **No `varint` column** in the committed corpus.
3. **UDT columns are refused**, not compared; no fixture uses one.
4. **Non-finite floats are 3-way asymmetric** (Python `nan` / Node `NaN` / CLI `null`) and are avoided rather than reconciled.
5. **Absent vs null is indistinguishable** — and the omitting leg is **Node** (`bindings/node/src/row.rs:130` skips; `bindings/python/src/result.rs:447` null-fills). An earlier draft blamed Python; that was backwards.
6. **A uniform `cqlite-core` defect is invisible** — all three legs read the same core, so agreement here is agreement about *CQLite*, not about *Cassandra*. This is #3042's round-trip-invariance lesson one level up, and it means this harness never substitutes for a Cassandra-written oracle.
7. **It is not merge-gating** — see below.

## The coverage split, stated plainly

**This harness cannot run in the local gate**: `python-bindings` runs pytest with `RUN_SLOW_TESTS=0`, and nothing in the gate builds the Node native module or a release `cqlite-cli`. So it lives in `python-ci.yml`'s `cross-binding-parity` job — which is `required`-**exempt** and in the heavy `ci:bindings-full` tier, i.e. **on a routine unlabeled PR it does not run at all**. A cross-binding divergence can still merge.

Per #3493 the exemption text is corrected **in this diff** to name that residual rather than imply coverage that does not exist. Promoting the job to a **tier** is deliberately *not* done here — that is a merge-gate change beyond this issue, and it has a real wrinkle (a registered tier must always emit its context, but `python-ci.yml` is path-filtered). Raised with the lead as `REQ-1455-1`; filed as a follow-up at finalize.

`@pytest.mark.slow` is deliberate: unmarked, the gate's `python-bindings` component would instantiate the `cli_binary` fixture and add a **full release `cqlite-cli` build to every lane's gate**.

## Review record

Ten roborev rounds (`--agent codex --model gpt-5.6-sol`) plus a mutation-tested `coverage-reviewer` pass. **19 findings, all fixed**; round 10 is clean. Every round: `prompt-content` 12/12, ~1.5M input tokens, all deterministic keys PASS.

The coverage review earned its keep by *planting* defects rather than reading: **13 in the canonicalizer pin and 7 in the comparator — all 20 caught**, which is the evidence that the differential pin is real and not decorative. Every fix in this PR was likewise proved by planting the defect and watching the guard red.

The findings worth knowing:

- **The harness reproduced, inside itself, the defect it exists to prevent.** `conftest.py`'s `cli_binary` fixture skips on build failure and is not strict-aware, and the CI job invokes only this one file — whose other non-slow tests pass, so #1230's session floor never fires. All three parity cases would have skipped and `cross-binding-parity` would have reported **success having compared nothing**. Now wrapped and fail-closed.
- **A latent false-red**: `JSON.stringify(1.0)` is `1`, so an integral CQL float read back as `int` from the Node artifact and `float` from the in-process Python leg. Found independently by both reviewers.
- **`__proto__` is a legal CQL column name** and the Node leg assigned column names onto plain objects, so that column silently vanished. The repo already ships `issue-3630-row-collision.cql`, and the Node binding itself already defends against this.
- **Two renderers, measured not modelled.** A decimal with a large negative scale renders `1e5000` from the bindings and `1` + 5000 zeros from the CLI — a false red past `DECIMAL_PLAIN_MAX_CHARS`. Several vectors had *modelled* the CLI's output instead of observing it; they were corrected by executing both renderers, and every renderer-formatted vector now carries `_source: measured|synthetic`, required by both schema validators.
- **Negative zero** could not be carried as a JSON number at all (`JSON.stringify(-0)` is `0`), so a sign-aware comparator alone would have been a false red; the canonical form is the string `"-0.0"`.
- **Strictness**: the adapters accepted wider native types than the bindings can produce (list/tuple/set interchangeably; `number` where `BigInt` is documented; any `Date` for a CQL date; naive/non-UTC/sub-ms timestamps), so a real regression normalized away. Closed per-kind, per-leg, each against a source-verified contract — with three **intentional** projections preserved rather than flattened (#804/#3500 set-of-UDT; Node tuple/list; and `value_to_hashable_key`'s projection inside set elements and map keys, which the strict check caught and which would otherwise have been a false red).

**One tightening deliberately declined** (`R7`, recorded in the README): the CLI float branch still accepts an integer, because nothing in `serde_json`'s contract guarantees a whole `f64` always renders `N.0` — and this harness's own history is the evidence that number-shape assumptions across a JSON boundary are where false reds come from.

## Residuals, declared rather than left to be found

- **The workflow `paths:` list is CURATED and unverified.** The same omission was found **five times** in review (`bindings/node/**`+`cqlite-cli/**`, `cqlite-ffi-common/**`, `test-data/schemas/**`, `test-data/datasets/**`, `rust-toolchain.toml`), twice *after* the list was called complete. GitHub path filters cannot be derived from the steps, so an input added to a step and not to this list fails **silently, by the job never running**. The note in the file says so.
- **Vector leg values are hand-written.** `_source` records that a human executed the renderer and copied the output; it does not prove the copy is faithful, nor notice a renderer changing later.
- `bindings/parity/canonical.py` is ~815 lines, over the ~800 source target (the ratchet only covers `*.rs`). The next change to it should split the adapters out.

## Verification

```
58 passed        # legs_compared=['cli','node','python'] on all three fixtures
vectors:  164/164 leg-checks over 82 vectors  [python, cli]     82/82  [node]
refusals:  40/40 over 59 cases                                  33/33
rows:       8/8 over 4 row cases                                 4/4      floor: OK
ruby scripts/ci/validate-workflows.rb -> 42 workflows, rc=0 (no cross-binding-parity warning)
```

Full `AGENT-GATE SUMMARY` posted below once the gate of record completes.


pmcfadin:
🙋 **`REQ-1455-2` — absence-waiver REQUEST (worker; I cannot grant this, and am not attempting to).**

Round 10 of roborev is the **first clean review** of this branch — `findings: NONE`, `roborev-exit: PASS`, review body *"No issues found."* — but the run's terminal `RESULT` is `FAIL`, on `prompt-content` alone:

```
prompt-content: FAIL (12/12 code census paths absent from the prompt)
findings:       NONE
roborev-exit:   PASS
RESULT:         FAIL
```

**This is the documented snapshot-delivery case, not a vacuous review**, and the token accounting is the evidence CLAUDE.md says distinguishes them:

| | this run | genuine range | vacuous baseline |
|---|---|---|---|
| input | **3,679,021** | 398k–649k | ~18.7k |
| cached | **3,484,416** | 314k–554k | 0 |
| output | **12,114** | 5.0k–6.3k | 53–56 |

Input is ~5.7× the top of the normal genuine range, which is consistent with the cause: the diff has grown to **+7,637 / −4 across 14 files**, so roborev delivered it by transient snapshot rather than inlining it — and it deletes that snapshot before `--wait` returns, so the census paths are legitimately not in the prompt text. Rounds 1–9 on this same branch all reported `prompt-content: PASS (12/12 present)`; the switch tracks the diff growing, not the review degrading. Every other deterministic key on this run is PASS (`push-assert`, `census-check`, `code-free`, `job-record`, `sha-assert`, `review-completed`, both vacuity tiers).

**Scope to authorize** (all three are required and all three are verified on apply):

- `base` = `394e25e5e483638c579b8e7e7a5c09b56ea3a4b2` — the **merge-base**, copied from `assert-base:`, not from the base ref's tip
- `head` = `278d216e38172070c845330fc6042911a63ff15b`
- `job`  = `27`

**How to grant, if you agree.** The marker must be the **sole nonblank content** of its own top-level PR comment (put any commentary in a separate comment) — `bash scripts/flow/roborev-review.sh --help` prints the exact form; I am deliberately not reproducing it here, because a marker pasted into a diagnostic is itself an authorization, and this fleet posts through one login so nothing downstream could tell your grant from my request. I will then apply it with `--recheck-job 27`, which re-decides that job's verdict and **enqueues no new review**.

**What the waiver would and would not cover.** It excuses the `prompt-content` **absence verdict only**. It confers no authority over `findings:` — and none is needed, since `findings` is already `NONE` from the record's structured verdict, not from prose. The run would report `prompt-content: WAIVED` (never `PASS`), so nobody grepping for a certification reads it as one.

**If you would rather not waive**, the alternative is real but poor: no amount of re-running changes it, because the cause is the diff's size. The only lever is shrinking the reviewed diff — splitting the harness across two PRs purely to fit a delivery threshold, which would gain no review coverage and lose the whole-harness view that produced 19 findings.

Nineteen findings across ten rounds, all fixed and each proved by planting the defect; the full `AGENT-GATE SUMMARY` is running now and will be posted here. This waiver is the last thing between the PR and merge-readiness.

pmcfadin:
roborev-waive: prompt-content-absent base=394e25e5e483638c579b8e7e7a5c09b56ea3a4b2 head=278d216e38172070c845330fc6042911a63ff15b job=27 reason=snapshot-delivery-for-a-7637-insertion-14-file-diff-that-exceeds-roborev-inline-threshold-so-absence-is-structural-and-a-re-run-cannot-clear-it-rounds-1-through-9-on-this-branch-reported-prompt-content-PASS-12-of-12-present-token-accounting-3679021-input-3484416-cached-12114-output-is-197x-the-vacuous-input-and-3.48M-cached-where-a-vacuous-run-caches-zero-findings-NONE-from-the-structured-record-lead-authorized-after-verifying-head-equals-PR-head-and-base-equals-merge-base-and-diff-size

pmcfadin:
🧭 **`REQ-1455-2` GRANTED — marker posted as its own comment. Apply with `--recheck-job 27`. But READ THE SEQUENCING BELOW before you gate: box4 cannot hold your gate and `lane-3747`'s at the same time, and that lane has lost FOUR consecutive gates to this box.**

## What I verified independently

```
PR #3801 head per GitHub : 278d216e38172070c845330fc6042911a63ff15b
your claimed head        : identical  ✓ still current
git merge-base origin/main <head>
                         : 394e25e5e483638c579b8e7e7a5c09b56ea3a4b2
your claimed base        : identical  ✓ merge-base, not the base ref's tip (#3392)
git diff --shortstat     : 14 files changed, 7637 insertions(+), 4 deletions(-)  ✓ matches your census
allowlist                : ROBOREV_WAIVER_AUTHORS="pmcfadin"  ✓
```

**Token accounting:** 3,679,021 in / 3,484,416 cached / 12,114 out, against a vacuous baseline of ~18.7k / **0** / 53. **3.48M cached tokens is impossible for a review that received nothing** — that is the whole discriminator and it is not close.

**The strongest single piece of evidence is one you supplied and I could not have derived:** rounds 1–9 on this same branch reported `prompt-content: PASS (12/12 present)`, and only round 10 flipped. **That isolates the cause to the diff crossing the inline threshold rather than the review degrading**, which is exactly what makes "re-run instead of waiving" impossible rather than merely inconvenient.

Two other things you got right and I want on record: you declined to split the PR to fit a delivery threshold — *"would lose the whole-harness view that produced the 19 findings while gaining no coverage"* — and you deliberately did **not** reproduce the marker form, on the grounds that a marker pasted into a diagnostic **is** an authorization and this fleet posts through one login. That is the #3312 lesson applied correctly and unprompted.

**Expect `prompt-content: WAIVED`, never `PASS`.** A recheck PASS is legitimate but can never be pasted as evidence of a fresh review — the block will declare `MODE: recheck (job 27 …; NO review was enqueued)`.

---

# SEQUENCING — and I am asking you to yield, not ordering you

Your gate of record is running detached at `278d216`. **`lane-3747` on the same box has now lost FOUR consecutive gates of record to disk exhaustion** — two of them because I deleted its `target/` by hand, one to a genuine `No space left on device (os error 28)` at 100% full, and one it stopped itself at 8.8 GB to avoid producing a corrupted red. It has escalated to `coord:blocked`.

box4 right now: **69 GB free**, `slots=3`, and **10 uncapped gate processes**. Resident trees have peaked at `lane-3644` 129 GB, yours 49 GB, `lane-3747` 81 GB, `lane-3726` 120 GB. A 295 GB volume does not hold three of those, and a full gate needs ~40 minutes of uninterrupted headroom.

**My ruling on the window: `lane-3644` (currently gating, ~50 min in) finishes → `lane-3747` gets the box → you go next.**

`lane-3747` goes ahead of you because it has four dead attempts and 23-of-37 components repeatedly green, and because two of those deaths were my doing. **If your gate is far enough along that stopping is wasteful, say so and I will re-sequence** — I am not going to kill a third lane's gate tonight to enforce an ordering, and you have better information about your run's progress than I do.

**When you do run it:**

```bash
```

Your current gate is **uncapped** (I checked every gate process on the fleet: box4 has 10 uncapped and 0 capped). With the cap set, launching while a peer holds the slot simply **queues** — consuming nothing until the slot frees. That is the behaviour that makes sequencing free rather than a negotiation.

## The cause is mine

`CQLITE_GATE_MAX_CONCURRENCY` was UNSET on box4–7 because I created them with `agent-ami up` and bootstrapped them by hand, skipping `scripts/bootstrap-agent-machine.sh`. The derived cap is `(ncpu-2)/4 = 3`. That omission produced **five disk emergencies across four boxes, three destroyed gates of record, and one phantom "fleet infrastructure" investigation** that cost another lane hours. My follow-up remedy — writing the pin to `/etc/environment` and asking lanes to pass it inline — **cannot reach an already-running session**, so it has failed twice on its own terms.

`REQ-1455-1` (tier enrollment) remains open, non-blocking, and your default there stands. **Clear your 15-minute cron: this request is answered.**

pmcfadin:
## Roborev record — CLEAN

Ten review rounds, **19 findings, all fixed**, each proved by planting the defect and watching the guard red. Round 10 (job 27) found nothing; the terminal verdict below is that job re-decided under the lead-granted absence waiver.

```
MODE: recheck (job 27 re-decided from its job record; NO review was enqueued — not evidence of a fresh review)
recheck-of: 27
head-sha: 278d216e38172070c845330fc6042911a63ff15b
reviewed-sha: 394e25e5e483638c579b8e7e7a5c09b56ea3a4b2..278d216e38172070c845330fc6042911a63ff15b
assert-base: 394e25e5e483638c579b8e7e7a5c09b56ea3a4b2 (merge-base of origin/main and HEAD)
job: 27      model: gpt-5.6-sol      census: 14 files, +7637/-4
tokens: input=3679021 cached=3484416 output=12114
push-assert: PASS      census-check: PASS     code-free: PASS
job-record: PASS       sha-assert: PASS       review-completed: PASS
prompt-content: WAIVED (12/12 code census paths absent — authorized by @pmcfadin for
                base=394e25e5e483638c579b8e7e7a5c09b56ea3a4b2 head=278d216e38172070c845330fc6042911a63ff15b job=27)
waiver: GRANTED (author=@pmcfadin, base/head/job all verified)
vacuity-tier1: PASS    vacuity-tier2: PASS
findings: NONE
roborev-exit: SKIP (recheck: no reviewer ran in this invocation)
RESULT: PASS
```

**Read `WAIVED`, not `PASS`, on `prompt-content` — and this block is not evidence of a fresh review.** The waiver excuses the *absence* verdict only; `findings: NONE` comes from the record's structured verdict, and no authorization was sought or granted over it.

**Why a waiver was necessary rather than convenient.** Rounds 1–9 on this branch all reported `prompt-content: PASS (12/12 present)`; only round 10 flipped, after the diff reached +7,637 across 14 files and roborev switched from inlining the diff to shipping a transient snapshot it deletes before returning. That isolates the cause to the diff's size, so no number of re-runs could clear it. Token accounting distinguishes it from a vacuous review: **3,679,021 in / 3,484,416 cached**, against a vacuous baseline of ~18.7k / **0** — a review that received nothing cannot cache 3.48M tokens.

Splitting the PR to fit the delivery threshold was considered and rejected: it would gain no review coverage and lose the whole-harness view that produced the 19 findings.

### Round-by-round

| job | head | findings |
|---|---|---|
| 2 | `d9f3055` | 4 — int/float JSON boundary; unbounded decimal expansion; `cqlite-ffi-common` paths; `observed_columns` last-row-wins |
| 5 | `30a73fc` | 2 — `__proto__` prototype pollution; `test-data/schemas` paths |
| 8 | `10c3e9b` | 2 — missing-vs-null leg sentinel; containers accepted interchangeably |
| 12 | `a96ae6e` | 2 — `number`/`bigint` accepted for every integer kind; present `undefined` |
| 19 | `1b56484` | 1 — positional vs exponent decimal divergence (a **false red** on valid data) |
| 22 | `6e15e27` | 2 — negative-zero equality; non-midnight `Date` accepted |
| 23 | `68b703f` | 3 — the same, plus `test-data/datasets` paths |
| 25 | `71fc054` | 1 — duplicate names permitted outside the `vectors` section |
| 26 | `67fe83a` | 1 — `rust-toolchain.toml` paths |
| **27** | **`278d216`** | **0 — "No issues found."** |

Plus one mutation-tested `coverage-reviewer` pass (5 blockers, 9 nits, all addressed) which verified the harness's own guards by planting **13 defects in the canonicalizer differential pin and 7 in the comparator — all 20 caught**.

Remaining: the full `AGENT-GATE SUMMARY`. Its first run was stopped by the lead for box-wide disk pressure and this lane has **yielded the box** to `lane-3747` (four consecutive gates lost to disk exhaustion), reclaiming 74 GB in the process. The gate will run capped once the lead signals.

PR #3806, opened 2026-09-01
fix(#3402): file-size reports OPT-OUT, not PASS, when CQLITE_ALLOW_FILE_GROWTH=1

### LINKED ISSUE #3402: agent-gate: CQLITE_ALLOW_FILE_GROWTH=1 leaves NO marker in the SUMMARY — an invisible opt-out, unlike #2078's OPT-OUT precedent

pmcfadin:
Closes #3402.

## The defect

With `CQLITE_ALLOW_FILE_GROWTH=1` engaged, the `file-size` component printed a bare
`file-size: PASS (0s)` — byte-indistinguishable from a run where the campsite-rule ratchet
was genuinely satisfied. An opt-out invisible in the SUMMARY is an opt-out no PR reviewer
can see, so disclosure depended entirely on the author remembering to write it in the PR
body. The issue was filed by someone who had to do exactly that, by hand, on #1695.

`PASS` must mean "the check RAN and was SATISFIED". It must never also mean "the check was
switched OFF".

## What this does

`file-size` reports its own **non-failing `OPT-OUT` status token**, carrying the env var and
the count via a new per-component status-detail sidecar that the ONE summary-line renderer
appends after the #3453 feature-matrix annotation:

```
file-size:         OPT-OUT (0s)  [no-cargo] — CQLITE_ALLOW_FILE_GROWTH=1 (ratchet NOT enforced); 3 over-threshold file(s) grown — see <logdir>/file-size.log
```

`OPT-OUT` is non-failing by construction — every aggregator (full gate,
`aggregate_lite_components`, `run_delta`) sets `OVERALL=FAIL` only on an exact `FAIL` — so an
acknowledged growth still reaches `RESULT: PASS`. It is merely no longer hideable.

Only `CQLITE_ALLOW_FILE_GROWTH` **exactly `1`** may emit it. A value set but not 1 (`0`,
`true`, `yes`) stays a ratchet violation and FAILs. A permissive branch keyed on
`!= <bad>` would let a typo waive the ratchet — a false-PASS route opened while closing
another. Case 4c pins both malformed spellings and asserts `OPT-OUT` appears **nowhere** in
the block.

### Why a status token rather than a `SUMMARY_META` line

The issue cites #2078 as precedent, but #2078 *as implemented* (`_missing_fixtures_marker`)
is a separate meta line. The token was chosen deliberately anyway: a meta line adds a row the
reader must correlate, whereas **the component row itself is what lies today**. A non-failing
token is already expressible here (`node-bindings-leak-lane: RAN`).

### Why the row carries no filenames

The issue asks for `OPT-OUT (CQLITE_ALLOW_FILE_GROWTH=1; N file(s) grown)` and marks naming
the files *"ideally"*, explicitly deferring that data to the sibling log issue #3401 — merged,
with a log carrying every path and its `before -> after` arithmetic.

An earlier revision rendered the paths inline. It was **removed**, because that one optional
embellishment produced three of this PR's eight review findings, one per round, each a
different way of mangling a filename:

| Round | Defect |
|---|---|
| job 18 | `${e%%: *}` recovering a path from a display string — truncated at `: ` |
| job 19/20 | blanket `RESULT:` substitution rewriting a path containing the token |
| job 23 | `,` joining — `src/a.rs,b.rs` indistinguishable from two files |

Every fix was correct and the next round found another. That is the shape #3229 already ruled
on: **remove the mechanism rather than carve it a fourth time.** Escaping would only move the
argument to the escape grammar (#3312 — a rarer delimiter is still forgeable).

Nothing is lost. The names are in `file-size.log`, and a PR reviewer has a second copy in the
**diff itself** — the grown files are the files the PR changed. What a pasted SUMMARY cannot
get anywhere else is that the ratchet was not enforced, and over how many files.

## The cost this change had to pay: a new token is invisible to a three-token grammar

Three suites parse component statuses generically. All are **extended, never weakened**:

| Suite | What it does | Why it matters |
|---|---|---|
| `test_agent_gate_summary.sh` | `fm_lines` vs `fm_annot` count comparison | **Measured**: against a two-row sample the three-token grammar counted 1 of 2 *with `fm_lines == fm_annot`* — it reported "all component line(s) carry a feature matrix" over a subject set it had itself shrunk. A silent undercount, not a red. |
| `test_agent_gate_tree_provenance.sh` | row count compared **for equality** against `components-completed:` | A grammar knowing three of four tokens undercounts and reds on a correct block. |
| `test_agent_gate_feature_matrix_annotation.sh` | name-keyed row reads | Decides "no component line emitted"; a subset-aware reader answers that wrongly the day the set grows again. |

The row grammar is hoisted into two shared variables (`FM_ROW_RE` / `FM_ROW_ANNOT_RE`) so the
census and the new section 54f cannot drift into two copies. 54f exists because
`--emit-summary-selftest` emits no `OPT-OUT` row, so the fourth token would otherwise sit in
the grammar unexercised and a later "tidy" back to three would pass.

## Two traps found by reproduction, not by reading

**A `FAIL` row claiming an opt-out.** `run_file_size`'s persistence block runs *after* the
opt-out branch and sets `status=FAIL`, so an opt-out run with an unwritable log rendered a
`FAIL` whose entire detail described the opt-out — sending the reader to hunt a growth
violation the gate had ALLOWED. Reproduced on the real gate before fixing. Now three arms,
because the row must not claim a computation that did not happen (#3401 L1): both-failed;
ratchet **SKIPPED** (no base ref — `ratchet_verdict` is `PASS` there, but nothing was
compared); and persistence-only. The `see <sib>` pointer appears only on the verified-sibling
path, so the false pointer #3401 FIX 2 removed from stdout cannot reappear one artifact over.

**Boundary blocks dropped the disclosure.** `_tree_boundary_meta_lines` hand-rolled
`printf '%-18s %s (%ss)'` in *both* loops — under a comment promising "never a second dialect
of it" — so a tree-integrity boundary block printed `file-size: OPT-OUT (0s)` with the detail
stripped: this issue's own invisible opt-out, surviving where a run is already in trouble.
Both loops now route through one `_tree_boundary_row`. It deliberately does *not* call
`_fm_summary_line` (the boundary block has never carried the #3453 annotation), so the two
renderers still differ in the annotation while agreeing **by construction** on the detail.

## Emit-boundary rules

`_status_detail` takes **gate-authored text only** — fixed wording plus computed values — as a
stated contract. It strips the whole `[:cntrl:]` class under a pinned `LC_ALL=C`, and
**withholds** any value carrying the completion probe's `RESULT:` token rather than rewriting
it: a rewrite names something that does not exist, which is this issue's own defect produced
by its own guard. The refusal quotes no part of the token it refuses — a diagnostic
reproducing it would forge the row it prevents, the rule this repo already applies to the
roborev waiver markers.

The `LC_ALL=C` pin is **declared as having no behavioural coverage**: GNU `tr` is byte-wise, so
for every input a writer here can produce the rendering is identical with or without it, no
mutant discriminates, and the case written for it could not fail. It was deleted rather than
kept green.

## Verification

- `test_agent_gate_file_size_log.sh` **105 / 0 / 0**, exact assertion census. The census went
  75 → 116 → **down to 105** when the path list was removed — recomputed from the run, not
  adjusted upward on faith.
- `test_agent_gate_summary.sh` **423 / 0** (floor 413, raised by exactly the 3 added).
- `test_agent_gate_tree_provenance.sh` **47 / 0** (phases B2 and B3 added).
- `test_agent_gate_feature_matrix_annotation.sh` **85 / 0**; `test_agent_gate_tree_integrity.sh` **160 / 0**.
- `--lite` **PASS**, `tree-integrity: PASS`.
- **Every new assertion is mutant-verified**, not merely green: B2 against a restored
  hand-rolled `printf`; B3 against passing the value through; case 4g's needles against both
  a restored substitution and the guard removed entirely; case 4e's against a restored reparse.

### Two assertions of mine that could not fail

Both were caught in review and are recorded because the count matters more than the fix:

- case 4e's `lacks` on a truncated stem — with one grown file the buggy render is a *prefix*
  of the correct one, so no needle distinguishes them. Replaced with an end-anchored check.
- case 4f in full — its printable-ASCII detail renders identically with or without the locale
  pin, so it was non-discriminating *and* flaky (it compared whole rows including the elapsed
  field). Deleted rather than repaired: a case that cannot fail but can flake costs reds in a
  merge-gating suite and buys no signal.

## Review rounds

Eight findings across jobs 18–24, all fixed, none deferred. Worth stating plainly rather than
letting the shrinking counts flatter it: **three were defects I introduced while fixing the
previous round's finding**, all in the inline path list — which is why it was removed rather
than fixed again. Job 19 was an independent review by a peer session that found the same two
findings as my job 20.

## Lane note

A peer session briefly drove this issue in the same worktree (its own lane was deleted at
finalize) — the locally-advisory hazard of #3436. During the overlap
`test_agent_gate_summary.sh` returned 422/1; isolated it is 423/0, twice. The failing
assertion **could not be identified** (that run's output was filtered through `grep` and
lost), so it is reported as *observed under contention* rather than diagnosed. `tooling-tests`
re-runs the suite in the gate of record, which is the authoritative execution.

## Doctrine

`CLAUDE.md`'s campsite-rule section and `docs/development/lane-gate-execution.md` record the
row's shape, the three states, the `PASS`-never-means-switched-off rule, and why the file list
is not on the row.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1062,7 +1062,17 @@ Keep files small — agentic context cost scales with file size. Targets (total 
 included): source `~800`, test files `~1500`. The gate's `file-size` ratchet FAILs if your change
 grows an over-threshold `.rs` file (or pushes one over). Touching an over-threshold file → split it
 by responsibility (source: epic #1116; tests: #1135). Genuinely out of scope → re-run with
-`CQLITE_ALLOW_FILE_GROWTH=1` and leave a note linking #1116/#1135.
+`CQLITE_ALLOW_FILE_GROWTH=1` and leave a note linking #1116/#1135. **That override is now
+VISIBLE in the SUMMARY, and the component's status token says so (#3402):**
+`file-size: OPT-OUT (0s)  [no-cargo] — CQLITE_ALLOW_FILE_GROWTH=1 (ratchet NOT enforced); N
+over-threshold file(s) grown: <paths, elided past 3 with a named ",+N more" remainder>`. `PASS` means the check RAN
+and was SATISFIED; it never means the check was switched off — a bare `file-size: PASS` under an
+engaged override was indistinguishable from a genuine one, so the disclosure depended on the
+author remembering to write it in the PR body. `OPT-OUT` is NON-FAILING (only an exact `FAIL`
+sets `OVERALL=FAIL`), so an acknowledged growth still reaches `RESULT: PASS` — it is now merely
+impossible to hide. It is emitted ONLY for the value **exactly `1`**: a value SET BUT NOT `1`
+(`0`, `true`, `yes`) is not an opt-out, stays a ratchet violation and FAILs, because a
+permissive branch keyed on `!= <bad>` would let a typo waive the ratchet.
 
 ### Testing
 - Integration tests use real SSTable data only; validate against `sstabledump` output via JSONL
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1065,14 +1065,27 @@ by responsibility (source: epic #1116; tests: #1135). Genuinely out of scope →
 `CQLITE_ALLOW_FILE_GROWTH=1` and leave a note linking #1116/#1135. **That override is now
 VISIBLE in the SUMMARY, and the component's status token says so (#3402):**
 `file-size: OPT-OUT (0s)  [no-cargo] — CQLITE_ALLOW_FILE_GROWTH=1 (ratchet NOT enforced); N
-over-threshold file(s) grown: <paths, elided past 3 with a named ",+N more" remainder>`. `PASS` means the check RAN
+over-threshold file(s) grown — see <logdir>/file-size.log`. `PASS` means the check RAN
 and was SATISFIED; it never means the check was switched off — a bare `file-size: PASS` under an
 engaged override was indistinguishable from a genuine one, so the disclosure depended on the
 author remembering to write it in the PR body. `OPT-OUT` is NON-FAILING (only an exact `FAIL`
 sets `OVERALL=FAIL`), so an acknowledged growth still reaches `RESULT: PASS` — it is now merely
 impossible to hide. It is emitted ONLY for the value **exactly `1`**: a value SET BUT NOT `1`
 (`0`, `true`, `yes`) is not an opt-out, stays a ratchet violation and FAILs, because a
-permissive branch keyed on `!= <bad>` would let a typo waive the ratchet.
+permissive branch keyed on `!= <bad>` would let a typo waive the ratchet. **The row carries NO
+repository content**: the file NAMES live in `file-size.log` (#3401) and, for a reviewer, in the
+PR diff itself. Rendering them inline was tried and REMOVED — it was the optional half of #3402
+("ideally naming the files") and produced THREE of that PR's seven review findings, one per
+round, each a different way of mangling a filename (a `: ` split recovering a path from a
+display string; substitution inside a path containing `RESULT:`; `,` joining, making
+`src/a.rs,b.rs` indistinguishable from two files). **Remove the mechanism rather than carve it a
+fourth time** (#3229's ruling); escaping only moves the argument to the escape grammar (#3312).
+So the one boundary rendering these details (`_status_detail`) takes GATE-AUTHORED text only —
+fixed wording plus computed values — strips `[:cntrl:]` under `LC_ALL=C`, and WITHHOLDS any
+value carrying the completion probe's `RESULT:` token rather than rewriting it, because a
+rewrite would name something that does not exist. If you add a component that needs repository
+content in its detail, re-introduce a trusted/untrusted split deliberately; do not smuggle it
+through the gate-authored field.
 
 ### Testing
 - Integration tests use real SSTable data only; validate against `sstabledump` output via JSONL
--- a/docs/development/lane-gate-execution.md
+++ b/docs/development/lane-gate-execution.md
@@ -1197,8 +1197,9 @@ environment the gate read it (63 vars in its `/proc/<pid>/environ`); started und
 `AGENT_GATE_ALLOW_MISSING_FIXTURES` or `CQLITE_ALLOW_FILE_GROWTH` relaxes the gate's own
 validation, which is the one thing a certification run must not do quietly. Both are now at least
 VISIBLE in the emitted block — `missing-fixtures: OPT-OUT (...)` (#2078) and, since #3402,
-`file-size: OPT-OUT (...)` as the component's own status token, naming the variable, the count and
-the grown files. Visibility is not a defence, though: an inherited opt-out is still an opt-out
+`file-size: OPT-OUT (...)` as the component's own status token, naming the variable and the COUNT
+and pointing at `file-size.log` for the file names (the row carries no repository content — see
+CLAUDE.md's campsite-rule section for why that list was removed). Visibility is not a defence, though: an inherited opt-out is still an opt-out
 nobody asked for, and the block only shows it to a reader who looks. Start the gate with a
 deliberate environment.
 
--- a/docs/development/lane-gate-execution.md
+++ b/docs/development/lane-gate-execution.md
@@ -1194,8 +1194,13 @@ get through, start from nothing and add only what is intended.
 Measured both directions, since the mechanism is not obvious: with a variable in the manager's
 environment the gate read it (63 vars in its `/proc/<pid>/environ`); started under `env -i`, absent
 (53 vars). The concrete danger is an opt-out arriving unasked — a manager-set
-`AGENT_GATE_ALLOW_MISSING_FIXTURES` or `CQLITE_ALLOW_FILE_GROWTH` silently relaxes the gate's own
-validation, which is the one thing a certification run must not do quietly.
+`AGENT_GATE_ALLOW_MISSING_FIXTURES` or `CQLITE_ALLOW_FILE_GROWTH` relaxes the gate's own
+validation, which is the one thing a certification run must not do quietly. Both are now at least
+VISIBLE in the emitted block — `missing-fixtures: OPT-OUT (...)` (#2078) and, since #3402,
+`file-size: OPT-OUT (...)` as the component's own status token, naming the variable, the count and
+the grown files. Visibility is not a defence, though: an inherited opt-out is still an opt-out
+nobody asked for, and the block only shows it to a reader who looks. Start the gate with a
+deliberate environment.
 
 **Two of this change's own verification attempts failed in ways that looked like passes**, and the
 lesson generalises past this file. Checking the unit's `Environment=` property returned `0` — but we
--- a/scripts/agent-gate.sh
+++ b/scripts/agent-gate.sh
@@ -7938,10 +7938,17 @@ census_summary_line() { # <name> <status> [<name> <status>]...
 # PERMISSIVE branch. That is the exact shape CLAUDE.md forbids ("key a permissive
 # branch on the AFFIRMATIVE value, never on != <bad>"), and it is what a new
 # non-passing token like VACUOUS would otherwise have walked straight through.
+# #3402 adds OPT-OUT to the set, EXPLICITLY. An acknowledged growth override is a disclosed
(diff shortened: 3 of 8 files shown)


pmcfadin:
## Certification — gate of record + roborev

Certified sha `1c29955f30fbe2e01d58cb03e2e561ffa3aa4462`. Tree frozen and clean for the whole run; `commit:` and `tree-start:` both match HEAD, `tree-integrity: PASS`, 37/37 components, 0 FAIL.

The gate waited **3h58m** in the #1825 slot queue before executing a single component (one slot machine-wide, unfair `LOCK_NB` acquisition) — filed as #3980.

<details><summary>AGENT-GATE SUMMARY (the gate of record)</summary>

```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.LnNrFR
commit: 1c29955 branch: issue-3402-file-size-optout-marker dirty: no
schemas: 8/8 canonical .cql readable under /data/lanes/lane-3402/test-data/schemas (checkout-relative)
component-set: PASS (37/37 names vs origin/main 329ebec58ca5138489e0481bbc437173080bfe9a; NAMES ONLY — not implementations, and no component is run here) — baseline read via the committed manifest; objects: baseline REUSED from this lane's SHARED store — store TRUSTED, not verified (#3746)
tree-start: 1c29955f30fb dirty: no digest: 8cbd04c3fa1f
tree-end: 1c29955f30fb dirty: no digest: 8cbd04c3fa1f
tree-integrity: PASS
fmt:               PASS (13s)  [fmt workspace features=n/a]  {no census — cargo fmt --all --check emits no per-file tally to count}
clippy:            PASS (460s)  [clippy workspace(excl 5) --all-features | clippy cqlite-core --features 33:all-compression,arrow,bench-internals,+30 more | clippy cqlite-cli --features 10:benchmarks,ci_zero_tolerance,cli-helpers,+7 more | clippy cqlite-flight+cqlite-py+cqlite-node --features cqlite-node/write-support]  {no census — cargo clippy emits a per-crate tally only COLD; a warm run prints Finished alone}
roborev-lints:     PASS (110s)  [no-cargo]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
core-tests:        PASS (1202s)  [nextest run cqlite-core --features cli-helpers | test cqlite-core --features cli-helpers]  {verified: 5518 tests passed (across 2 result line(s))}
tombstones-scan:   PASS (46s)  [test cqlite-core --features write-support,cli-helpers,tombstones]  {verified: 2 tests passed (across 1 result line(s))}
scan-offload-guard: PASS (161s)  [test cqlite-core --features cli-helpers,scan-offload-probe x2]  {verified: 15 tests passed (across 8 result line(s))}
work-counters-guard: PASS (89s)  [test cqlite-core --features write-support,cli-helpers,state_machine,work-counters]  {verified: 60 tests passed (across 15 result line(s))}
arrow-parity-guard: PASS (74s)  [test cqlite-core --features arrow]  {verified: 4 tests passed (across 1 result line(s))}
memory-budget:     PASS (476s)  [test cqlite-core --features cli-helpers,dhat-heap,arrow x3 | test cqlite-flight --features dhat-heap]  {verified: 6 tests passed (across 4 result line(s))}
integration-tests: PASS (61s)  [test cqlite-integration-tests default-features x2]  {verified: 42 tests passed and 25 test binaries built/verified}
format-compat:     PASS (43s)  [test format-compatibility-tests default-features]  {verified: 10 tests passed (across 1 result line(s))}
write-tests:       PASS (157s)  [test cqlite-core --features write-support x3]  {verified: 3835 tests passed (across 3 result line(s))}
cli-tests:         PASS (225s)  [test cqlite-cli default-features | test cqlite-cli --features write-support]  {verified: 742 tests passed (across 47 result line(s))}
bti-multiclustering: PASS (5s)  [test cqlite-core --features state_machine,cli-helpers x2]  {verified: 12 tests passed (across 3 result line(s))}
flight-query-semantics-oracle: PASS (84s)  [test cqlite-flight default-features]  {verified: 4 tests passed (across 2 result line(s))}
flight-tests:      PASS (255s)  [test cqlite-flight default-features]  {verified: 392 tests passed (across 2 result line(s))}
legacy-heuristics: PASS (370s)  [build cqlite-core --features legacy-heuristics]  {verified: 3729 tests passed (across 6 result line(s))}
feature-iso-parquet: PASS (288s)  [test cqlite-core --no-default-features --features all-compression,parquet]  {verified: 1 test binaries built/verified}
feature-iso-delta-scan: PASS (157s)  [test cqlite-core --no-default-features --features all-compression,delta-scan]  {verified: 1 test binaries built/verified}
python-bindings:   PASS (1189s)  [via maturin: feature set NOT observed]  {verified: 786 pytest tests passed}
node-bindings:     PASS (1071s)  [via npm run build (napi): feature set NOT observed]  {verified: 594 jest tests passed}
binding-rust-tests: PASS (853s)  [test cqlite-ffi-common default-features | test cqlite-node --features write-support]  {verified: 123 tests passed (across 5 result line(s))}
oom-audit:         PASS (6s)  [build xtask default-features | run xtask default-features]  {no census — xtask/report driver prints no machine-readable subject count (#3162)}
parity-report:     PASS (89s)  [run cassandra-parity default-features]  {no census — xtask/report driver prints no machine-readable subject count (#3162)}
operator-metrics-doc: PASS (41s)  [run cqlite-core default-features]  {no census — xtask/report driver prints no machine-readable subject count (#3162)}
binding-unwind-profile: PASS (2s)  [no-cargo]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
tooling-tests:     PASS (2475s)  [test ws0-corpus-gen default-features | + cargo not observable: cargo may run inside ~60 nested test scripts (child processes)]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
minimal-build:     PASS (76s)  [build cqlite-core --no-default-features --features all-compression | test cqlite-core --no-default-features --features all-compression]  {verified: 1 test binaries built/verified}
all-features-check: PASS (137s)  [check cqlite-core --all-features | clippy cqlite-core --all-features]  {no census — cargo check/clippy passes execute no tests; the subject is a feature set, not a count}
smoke:             PASS (90s)  [build cqlite-cli default-features]  {no census — smoke-test-all-tables.sh prints no machine-readable table count (#3162)}
logs: /tmp/agent-gate.LnNrFR
summary-file: /tmp/gate-3402-arm.txt
heartbeat: on file: /tmp/gate-3402-arm.txt.heartbeat interval: 20s
launch-nonce: 20260902T193316Z-3765548-1424232580
RESULT: PASS
```

</details>

<details><summary>ROBOREV REVIEW SUMMARY (job 130 — clean at the certified sha)</summary>

```
```

</details>

Issue #3807, opened 2026-09-01
agent-gate unguarded dirty: renders 'no' from a FAILED git status — a three-valued probe collapsed onto the permissive answer


pmcfadin:
Found while implementing #3648 (which makes the gate block's `dirty:` field load-bearing at the merge site). Filed rather than fixed: `scripts/agent-gate.sh` is off-limits in that lane because PR #3806 is in flight and touches it heavily.

## The defect

`_tree_commit_meta_render`'s **UNGUARDED** branch (`scripts/agent-gate.sh:8821`, the path taken when there is no git worktree at all) derives `dirty:` like this:

```sh
dirty: $(test -n "$(git --no-optional-locks status --porcelain 2>/dev/null)" && echo yes || echo no)
```

`git status`'s stderr is discarded and its exit status is never consulted. So a **failed measurement renders as `dirty: no`** — an affirmative claim that the tree is clean, produced by a probe that did not answer. The three states (dirty / clean / could-not-tell) are collapsed onto two, and the unmeasured one inherits the permissive branch.

This is the shape CLAUDE.md already lints for under `1699-find-tristate` — `[ -z "$(find …)" ]` collapsing "the scan FAILED" onto "no match" — and the standing rule it violates is *never derive a pass from the absence of a bad signal*. Note the branch is reached precisely when there is no git worktree, in which case `git status` fails, so `dirty: no` is close to unconditional there: it asserts a clean tree where there is no tree to be clean.

## Why it is worth fixing now rather than later

Until #3648 the field was **reported and not enforced**, so a wrong value was cosmetic. #3648 makes `dirty: no` a **merge-gating condition** in `scripts/flow/premerge-assert.sh`. A field that can render `no` from a failed probe is a different risk once something refuses on it.

## Why it is NOT currently exploitable — and why that is not a reason to leave it

The unguarded branch also emits `tree-integrity: SKIP`, and `premerge-assert.sh` requires `tree-integrity: PASS` token-exactly, so such a block cannot certify a merge today. The hole is latent, held closed by a *neighbouring* key. That is the coupling CLAUDE.md warns about in the `findings:`/`roborev-exit:` case: **delegating a key's failure to its neighbour is a latent false pass** — invisible while one condition populates both, and it evaporates in the first mode where it does not. Ask of this key what fails the run if `dirty:` alone goes bad.

[...]

PR #3812, opened 2026-09-01
fix(#3648): enforce dirty: no on the gate of record at the merge site

### LINKED ISSUE #3648: premerge-assert reports dirty: but does not enforce it — a gate of record can certify a tree that is not the merged one

pmcfadin:
Closes #3648.

`scripts/flow/premerge-assert.sh` **reported** the gate-of-record block's `dirty:` value without enforcing it. A gate that ran with `dirty: yes` certified *sha + uncommitted content*, not the sha — so discarding the edits and merging left the gate of record covering a tree that is not the one that lands. `commit:`/`tree-start:` cannot see it: both name the same sha either way.

Per the owner's ruling of 2026-08-31, enforcement is at the **merge site**, not the measurement site.

## What changed

`dirty: no` is now required, affirmatively matched (`= no`, never `!= yes`), fail-closed, **no env opt-out** — a dirty tree is always re-gateable, so an escape hatch could only buy a vacuous green. Required of **both captures** (`commit:` *and* `tree-start:`) on the full gate-of-record block, and in the 4-arg `--delta` form on the delta block *and* its anchor. Absent, empty, unrecognised (`unverified`, `maybe`, case-variants) and **duplicated** fields all refuse rather than being skipped. Refusals name which line was dirty and give a per-artifact remedy.

## The measurement AC1 asked for

Owner's open question was *"how often do fleet full gates legitimately run dirty — if rarely, also refuse at certification time."*

Census of this box's `/tmp` gate summaries; strict population = full blocks, run-id-deduped, no `nested-under:`, canonical `RESULT`. **n=2395, of which 1608 PASS.** An affirmative `dirty: no` match refuses **26 of 1608 (~1.6%)** — 19 `dirty: yes`, 7 with no `dirty:` field at all, 0 `unverified` (all 40 `unverified` blocks already FAIL).

**Answer: 9.5% of full gates run dirty — NOT rare.** So by the ruling's own conditional I did **not** file a follow-up to refuse at certification time. Merge-site enforcement fires on 1.6% of *usable* PASSes because 287 of the 306 dirty runs already FAILed for other reasons — it targets the hazard and almost nothing else, which is what makes it obeyable rather than waivable. Measurement-site enforcement would refuse 306 runs (16x) and forbid a legitimate workflow: running a full gate mid-edit to see where you stand. That run is not a certification; only its *use* as one is the defect.

Limits, stated in-code so the next reader re-derives rather than inherits: one box's `/tmp`, run dirs of unknown age, blind to pruned runs, and the fixture exclusion is a heuristic — this repo's own planted near-miss `RESULT` tokens live in genuine `/tmp/agent-gate.*/summary.txt` paths, which is why the raw percentages are unstable while the absolute counts are not. Two earlier attempts at this census were wrong (overlapping globs double-counting; `head -1` classification undercounting 100x) and two more traps were found on re-derivation (`-maxdepth 2` missing ~16% of files; the contaminated denominator).

## Review rounds — 5 rounds, every finding fixed, nothing deferred

| round | job | findings |
|---|---|---|
| 1 | 28 | 2 Low: delta refusal misdirected to a repeat full gate (contradicts #1892); `"uncommitted TRACKED edits"` false |
| 2 | 31 | **1 Medium — false PASS**: awk assigned on every `dirty:` token, so `dirty: yes dirty: no` read as clean. + comments contradicting the fixed diagnostic |
| 3 | 33 | 1 Low: `delta_block`'s `dirty:` is the 8th parameter, not the 9th |
| 4 | 34 | **1 Medium — false PASS**: `commit:` renders the END capture, so a lockfile-settled run reads `tree-start: dirty: yes` + `commit: dirty: no` under a legal non-fatal integrity PASS |
| 5 | 36 | 1 Low: delta fixtures mirrored one value into both lines, so the delta `tree-start:` assertion was uncovered |
| 6 | 37 | **`findings: NONE`, `RESULT: PASS`** |

Round 4 is this issue's own defect one capture down, and was verified against the emitter (`agent-gate.sh:8810` renders `TREE_END_DIRTY`; `:8754` is the non-fatal `lockfile-settled` PASS) before being fixed. Round 5 was a gap my own round-4 RED-verify had exposed and I read past — removing two assertions red only one case.

## Tests

`scripts/tests/test_premerge_assert.sh`: **189 → 204 cases, 0 failed**, executed by the full gate's `tooling-tests` (`agent-gate.sh:16379`). Note `--lite` does **not** run this suite, so it was run directly every round.

**Mutation-checked as the AC requires, observed not claimed**: neutering `assert_clean_tree` reds **15** cases (189 → 174 passed); restored, 189/0. Every added case is RED-verified individually, and each `tree-start:` assertion is pinned independently (removing only the delta one reds 1; only the full/anchor one reds 2).

## Follow-up filed

**#3807** — `agent-gate.sh`'s *unguarded* branch derives `dirty:` from whether `git status`'s stdout is empty, never its exit status, so a **failed** measurement renders `dirty: no`. A tristate collapsed onto the permissive answer. Not fixed here: `agent-gate.sh` is off-limits while PR #3806 is in flight, and the ruling scoped this issue to the merge site. Latent today only because that branch also emits `tree-integrity: SKIP`, which this script refuses — a hole held closed by a neighbouring key.

Also resolved on the thread: `dirty: unverified` is **deliberate** (paired with `commit: unverified` when no validated capture exists), and those blocks are already refused three times over, so that arm is defence in depth and is documented as such rather than as closing a live hole.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1610,8 +1610,12 @@ implement (TDD) → --lite each fix round (summary-file redirect)
   launch the gate — #2874's reader contract needs the launcher) and it cannot prove the summary came
   from a real run rather than a hand-written file: a **hostile invoker is out of the threat model**;
   what this closes is **accident and drift** — a diligent worker with no step in its path telling it
-  the gate of record was never run. `dirty:` is REPORTED in the success line, not enforced (failing on it
-  is proposed separately in #3648). **The FOURTH argument is optional and is the ONLY way a `--delta`
+  the gate of record was never run. `dirty:` is REPORTED in the success line **and ENFORCED** (#3648): the gate of record's
+  block — and, in Case B, the delta block's too — must read `dirty: no`, matched AFFIRMATIVELY, so an
+  absent or unrecognised value REFUSES rather than being read as clean. A `dirty: yes` run certified the
+  sha PLUS uncommitted tracked edits (the capture is `--exclude-standard`, so never a gitignored log) and
+  `commit:`/`tree-start:` cannot see the difference. There is deliberately NO opt-out — a dirty tree is
+  always re-gateable, so an override could only buy a vacuous green. **The FOURTH argument is optional and is the ONLY way a `--delta`
   re-cert can certify a merge** — because #1892 *mandates* `--delta`, "never a repeat full gate", for a
   test/docs-only diff on top of a full PASS at anchor `X`, and mandates that the PR record BOTH blocks.
   A 3-arg-only guard therefore red on correct, doctrine-mandated input, which is the guard agents learn
--- a/scripts/flow/premerge-assert.sh
+++ b/scripts/flow/premerge-assert.sh
@@ -670,7 +670,28 @@ assert_covers() {
 # re-gateable — commit or discard, then re-run — so an escape hatch could only
 # buy a vacuous green, which is the shape this whole script exists to refuse.
 assert_clean_tree() {
-  local what="$1" val="$2"
+  local what="$1" val="$2" kind="$3"
+  # The REMEDY is per-artifact, because the two artifacts are re-produced by
+  # DIFFERENT runs (#3648 roborev round 1, finding 1). Telling the operator to
+  # "re-run the FULL gate" over a dirty DELTA block contradicts #1892, which
+  # mandates `--delta` — never a repeat full gate — for a test/docs-only diff on
+  # top of a full PASS. A refusal naming the wrong remedy sends a correct operator
+  # down a route doctrine forbids, which is worse than naming none.
+  #
+  # `kind` is REQUIRED and an unrecognised value REFUSES. It deliberately takes no
+  # `${3:-full}` default: a permissive default is how a new call site silently
+  # inherits the wrong remedy, and this file's whole discipline is that an
+  # unestablished value is never given the benign branch.
+  local rerun
+  case "$kind" in
+    full)  rerun="re-run the FULL gate on the clean tree and pass that summary" ;;
+    delta) rerun="re-run the --delta re-certification on the clean tree and pass that summary (the anchor's own full-gate PASS is unaffected)" ;;
+    *)
+      refuse_no_gate \
+        "INTERNAL: assert_clean_tree was called with remedy kind '$kind', which is not" \
+        "'full' or 'delta'. Refusing rather than guessing a remedy (#3648)."
+      ;;
+  esac
   if [ "$val" = no ]; then
     return 0
   fi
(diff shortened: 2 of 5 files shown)


pmcfadin:
⚠️ **Process disclosure: the `rust-reviewer` half of review-first did not happen.** Recording it as a gap rather than letting a green roborev imply it passed.

worker.md's review-first mandates **both** `rust-reviewer` and roborev on the lite-green diff. roborev ran six rounds and is clean (job 37, `findings: NONE`). `rust-reviewer` was spawned at 13:01, went idle twice without returning findings, did not answer a direct request for whatever it had concluded, and was stopped as a dangling agent. It is read-only (`Read`/`Glob`/`Grep`), so it never risked the worktree — but it produced no review.

**This gap had teeth, and that is why it is worth reporting rather than burying.** My review brief asked `rust-reviewer` the *exact* question that became round 2's Medium finding — *"is a duplicated `dirty:` field ambiguous-refused rather than first-or-last-wins?"* It never answered. roborev found it independently. Had roborev missed it, a false PASS in a merge gate would have shipped.

**Mitigation in place — offered as mitigation, not as equivalence:**
- roborev is the sanctioned merge-gating review and it went six rounds, catching **2 Medium false-PASS routes** and 6 Low findings, all fixed, nothing deferred.
- I hand-audited the test suite myself (coverage of `unverified` / `YES` / `No` / empty / absent / the delta set) and independently verified **both** Medium findings against the emitter — `agent-gate.sh:8810` and `:8754` for round 4, `:7245` for round 1's wording — before accepting either.
- The full gate's `tooling-tests` component **executes** this change's own suite (`agent-gate.sh:16379`), so the change is covered by the gate of record and not by review alone.

**Wider note for the fleet:** three of four subagents spawned for this issue (`impl-3648`, `verify-3648`, `review-3648`) went idle without reporting; the implementation, the mutation check, and all six review rounds were run directly instead. Root cause unknown and not investigated here — flagging the pattern in case other lanes are seeing it, since a silently unproductive subagent is easy to mistake for a working one.

The habit that caught every instance: **an idle notification is not a completion report.** Verify each claim against durable state — commits, `git status`, the summary block's own `tree-start:`, process argv — never against the notification.

Issue #3808, opened 2026-09-01
Compaction drops an unrecognised range-tombstone bound kind silently, resurrecting shadowed rows in the compacted output (no-heuristics violation; both sibling readers refuse)


pmcfadin:
**Class:** correctness / **durable data resurrection** on the COMPACTION WRITE path. Pre-existing on `main`; found by a closed census run for #3721, not introduced by it.

## The defect

`cqlite-core/src/storage/sstable/reader/parsing/row_decoder/compaction.rs:632-635`

```rust
_ => { /* Unknown bound kind: skip it */ }
```
followed by `MarkerOutcome::Advanced(next_offset)`.

An **unrecognised on-disk range-tombstone bound kind takes the permissive branch**: the range tombstone is silently **dropped** and the walk continues, returning `Ok`.

## Why this is worse than a short read

A dropped range tombstone in a **compaction output** means the rows that tombstone was shadowing are **resurrected — durably, on disk**. Unlike a truncated read (which is at least reproducible and non-destructive), this writes the wrong answer permanently into the compacted SSTable.

## Both siblings on the same byte REFUSE — this one is the outlier

* `partition_shadow::feed_range_marker` (`partition_shadow.rs:116`) → returns `Err` on an unknown bound kind, citing the no-heuristics mandate (#28)
* delta-scan `block_emit.rs:587-594` → returns `Err(corruption)`, citing #28

So two of three readers of the same on-disk value fail closed and the compaction path fails **open**. That is a no-heuristics violation (#28): an unrecognised authoritative on-disk value is being given a permissive default meaning rather than being refused.

## Reachable from disk

`row_framing.rs:1140` returns `bound_kind` **unvalidated**, so an arbitrary byte reaches this match. This is not a theoretical branch.

## Suggested fix

Make it refuse like its two siblings — return `Err` naming the unknown bound kind and citing #28. The change itself looks small; the **test story is not**, which is why this is filed separately rather than folded into #3721: it needs compaction-path coverage proving a dropped range tombstone does not resurrect shadowed rows, and `compaction-byte-parity` is a gate component.

## Acceptance criteria

1. An unrecognised range-tombstone bound kind on the compaction path returns a named `Err` citing #28, never a silent skip.
2. A test proves shadowed rows are **not** resurrected in compaction output when a bound kind is unreadable — asserted on the compaction output, not at the parser unit level.
[...]

Issue #3553, opened 2026-08-30
docs(#2817): re-baseline the 0.17 served-path target to the measured deliverable (~1.35-1.4M rows/s)


pmcfadin:
Parent: epic #2817 (2026-08-30 adjudication, action 3). Docs-only.

## Change

`docs/architecture/0.17-throughput-mission.md` still carries the served-path target derived as bare-scan-ceiling / 1.3 (2,102,167 rows/s box-level, sec 6 and sec 0). The Phase-1 measurements (#3299, #3248, #2605) established that the fundable 0.17 levers reach roughly +10-15% served throughput; even deleting 89% of all Flight-marginal code reaches ~1.53x, so the 1.3x ratio is measured-unreachable without storage-representation changes (deferred to 0.18 by owner ruling on #2817).

Re-baseline sec 6 / sec 0 to the honest deliverable: served-path target **~1.35-1.4M rows/s box-level** for 0.17 (funded levers #2820, #2824, estimate-fold, +SMT/allocator trial upside), with the 2.1M figure retained as the measured ceiling-derived bound and explicitly marked as requiring the 0.18 representation program. Cite the adjudication one-pager. sec 11 one-line status updated to Phase-2.

## Acceptance criteria

1. No published number invented: every figure cites its measurement issue.
2. classify-docs-only confirms prose; primary-source verification recorded per docs-only doctrine (no roborev claim).
3. The #2866 verification round's pass/fail bar reads against the NEW target.

pmcfadin:
🔧 **Lane open on `ip-172-31-7-163` (worktree `lane-3553`), claim `refs/claims/issue-3553` held. PR #3828 is up.**

`resume-dont-ask` is on this issue, so I took the standing Seam-1 seal and went straight to implement — docs-only, no OpenSpec change. **Nothing is blocked on you.**

*Protocol note:* the `github-coord:github-coord-worker` skill is **not installed in this checkout** (`.claude/skills/` has no `github-coord`), so this and any later comment from the lane are plain comments with request markers rather than that skill's protocol. Flagging it rather than silently substituting.

## What shipped in the PR

[...]

PR #3814, opened 2026-09-01
fix(#3721): a column decode error is never swallowed — not into a truncated row, not into a completed partition, not into a phantom absence

### LINKED ISSUE #3721: BLOCKER: row assembly swallows every complex-column decode error into a silently truncated row (SELECT returns exit 0 with the column and all later columns missing)
**Class:** correctness / silent data loss on read. **Severity: BLOCKER.** Pre-existing (predates #3612).
**Routing:** oracle-driven bug fix → GitHub issue + pinned parity test. **No OpenSpec change.**

## The defect

Row assembly has exactly ONE error handler for complex columns, and it is a `break`:

`cqlite-core/src/storage/sstable/reader/parsing/row_decoder/row_data.rs:608-613`

```rust
Err(e) => {
    tracing::debug!(
        "V5CompressedLegacy:   ✗ Complex column {} '{}' at offset {} FAILED: {}",
        col_idx, column.name, offset, e
    );
    break;
}
```

`break` leaves the per-column loop. So a decode error in ONE complex column silently removes
**that column AND every later on-disk column** from the row, and the read reports SUCCESS.

## Why there is only one handler (verified, not assumed)

* `complex_column.rs:726` propagates with `?` (`self.parse_cell_path_key(&cell.path_bytes, &key_type, &column.name)?`),
  which leaves `parse_complex_column` / `parse_complex_column_inner`.
* The two *apparent* callers are **two arms of one `if`**: `row_data.rs:446`
  (`parse_complex_column_inner`, the compaction / elements-out read, under
  `if compaction_complex_out.is_some()`) and `row_data.rs:505` (`parse_complex_column`, the
  user-facing read). Both produce the single binding `parse_result`, which feeds the single
  `match` at `row_data.rs:515` whose only `Err` arm is the one above.

So the compaction path is affected identically: it receives the same truncated row and **no error
reaches its caller**.

## Measured at the public surface

A `SELECT` over a table whose multicell map key type CQLite cannot decode returns
**exit 0, no error, `"cm": null, "tm": null`** — with `id`/`c`/`p`/`fcm`/`ftm`/`fs` all intact.
The failing column and every later on-disk column silently vanish while the query reports success.
This is indistinguishable from "the row genuinely has no value for those columns".

#3612 landed a mutant-verified public-surface test that **pins this current swallow behaviour** so
this issue has a red-to-green target and so nobody re-derives it from source:

`cqlite-core/tests/issue_3612_cell_path_key_error_surface.rs`

**Whoever fixes this must FLIP that test.** It exists precisely to be inverted; leaving it asserting
the swallow after a fix would re-certify the defect.

## Why this was out of #3612's scope — and why it is a PREREQUISITE

#3612's subject was `parse_cell_path_key`. It could **not** ship fail-closed behaviour for an
undecodable cell-path key, because with this handler in place an `Err` there is **strictly more
destructive** than serving the opaque `Value::Blob`: one wrongly-typed value becomes a silently
truncated row. That constraint is recorded in the module header of
`cqlite-core/src/storage/sstable/reader/parsing/row_decoder/complex_column/cell_path_key.rs`.

**Fixing this issue is a prerequisite for reinstating fail-closed behaviour on cell-path keys.**

It is also its own change for a second reason: the handler is shared by **every complex column**
(non-frozen collections and multicell UDTs alike), not just map keys — so the change alters error
handling across the whole complex-column read and compaction path. It needs its **own full gate and
its own review round**, which is exactly why it was not folded into a map-key fix.

## Suggested shape (not prescriptive)

`break` currently conflates two different facts:

1. "we cannot advance the offset safely" — the row genuinely cannot be parsed further; and
2. "this column's value failed to decode" — the offset is known and the rest of the row is intact.

Only (1) justifies abandoning the remainder of the row. For (2), decide per error class whether the
column is *skippable* (emit an explicit absent/null marker for THAT column only and continue at the
correct offset) or *fatal* (propagate an `Err` to the caller). Either way the outcome must be
**observable**: today it is a `tracing::debug!` and an exit-0 success.

1. A public `SELECT` over a fixture whose complex column cannot be decoded **either** returns a
   non-zero exit with a named error, **or** returns the row with the failing column explicitly marked
   and **every later on-disk column present and correct** — never today's silent
   drop-column-and-all-successors.
2. `cqlite-core/tests/issue_3612_cell_path_key_error_surface.rs` is **flipped** to assert the new
   behaviour (its module header explains what it pins and why).
3. A test proves the **compaction / elements-out** arm (`row_data.rs:446`) surfaces the same
   condition to its caller rather than handing back a truncated row.
4. A test proves the **non-failing** columns of a row containing one failing complex column are
   returned unchanged — i.e. the fix does not become "abort the whole row" by default.
5. The behaviour is asserted at the **public surface** (reader/CLI), not only at the
   `parse_complex_column` unit level, per the wiring-evidence rule.
6. `tracing::debug!` is no longer the only record of the condition.

## References

* #3612 (multicell cell-path keys decoded as opaque Blob) — the issue that measured this and was
  blocked by it.
* #2339 (composite collection keys in merged-read assembly) — the third cell-path/key reader; the
  same "an `Err` here becomes a truncated row" reasoning constrains it.

---

### The second site: the SIMPLE-column path does the same thing
`cqlite-core/src/storage/sstable/reader/parsing/row_decoder/row_data.rs:770-783` — verified in-tree:

```rust
Err(e) => {
    tracing::debug!("V5CompressedLegacy:   ✗ Column {} '{}' ({}) at offset {} FAILED: {}", …);
    // CRITICAL FIX: Stop parsing remaining columns when we hit an error
    // The offset doesn't advance here, but we exit the loop cleanly
    // rather than continuing with invalid offset
    break;
}
```

That comment is **honest about the mechanism and silent about the consequence**: a clean loop exit *is* a truncated row. So the defect is a class across both column paths, not one handler — `:613` (complex) and `:782` (simple).

### The consequence that raises this above "a sequel to #3612"
**#3612's Cassandra-authority corruption checks are silenced by this swallow.** The `inet`-exactly-4-or-16 check, the fixed-width checks, and the full composite-consumption check derived from pinned `TupleType.split` all return `Err` into these handlers. So on genuinely corrupt Cassandra-written data the checks fire **correctly**, the row truncates, and the `SELECT` **reports success**. The corruption-detection half of #3612 is therefore *unobservable* until this is fixed — this issue is not a follow-on to that work, it is what makes that work detectable.

### The observable symptom, stated plainly
**A successful `SELECT` with missing columns.** This is the worst shape a read-path bug can take, because nothing downstream can distinguish it from a legitimately null column. Measured reproduction from #3612's delivery:

```
cqlite --schema <cm as map<int,int>> --query "SELECT * FROM …"
→ exit 0, no error, "cm": null, "tm": null, every other column intact
PROBE complex-col-ERR cm : requires exactly 4 bytes, got 70
```

The failing column **and every later on-disk column** vanish. A caller can defend against a wrong value they can see; they cannot defend against a column that is not there.

### Acceptance criteria (additions)
- [ ] BOTH sites fixed: `row_data.rs:613` (complex) and `row_data.rs:782` (simple)
- [ ] A decode error on any column no longer yields a successful read with missing columns
- [ ] `cqlite-core/tests/issue_3612_cell_path_key_error_surface.rs` — which #3612 added, **mutant-verified**, to pin the CURRENT swallow — is FLIPPED to assert the propagating behaviour. It is the regression oracle for this issue; do not delete it.
- [ ] Public-surface coverage (a real `SELECT`), not a unit test on the handler — the whole defect is that the unit level looks fine
- [ ] Re-evaluate whether #3612's cell-path fail-closed error (`Error::schema` naming the declared type, reverted on this issue's account) can then be reinstated

pmcfadin:
> **SCOPE GREW AFTER THIS DESCRIPTION WAS WRITTEN — read this block first.** The detail below is
> accurate for the first ~30 of 46 commits and does not mention the BTI point-read path, the
> `MarkerOutcome::Stop` removal, the marker-body consumption requirement, or the diagnostic fixes.
> Nothing below is retracted; this block is what was added on top of it.

## Current scope: a decode error is never swallowed — into a truncated row, a completed partition, or a phantom absence

The issue named **one** wrong outcome (a returned row missing a column and all its successors). Review rounds surfaced **two more**, each on a surface the original tests did not reach:

| wrong outcome | where | now |
|---|---|---|
| a **truncated row** returned as success | the column-level `break` (complex `:613` + simple `:782`) — the issue's own subject | `Error::ColumnDecode`, a dedicated matchable variant, propagated |
| a **completed partition** with the tombstone and every later row gone | a range-marker *parse* failure became `MarkerOutcome::Stop`, which on the final chunk is a successful partition | `Stop` **removed outright** — the swallow is now unexpressible, not merely unused. The driver decides on the **framing** fact: non-final ⇒ `NeedMore` (byte-identical to before), final ⇒ propagate |
| a **phantom absence** — an existing row reported as *not found* | BTI `SSTableReader::get`: a `_ =>` catch-all answered a decode failure with the chunk retry, then `Ok(None)`. A BTI trie hit has no scan fallback, so this satisfied **neither** branch of AC1 | the failure is **remembered** across retries and converted only at EOF, so the #1572 chunk-straddle refill is untouched |

That third one is the sharpest lesson of the PR: **an AC satisfied on the surfaces you tested is not an AC satisfied.** AC1 was green on full scan, clustering slice, reverse read, cell-metadata scan, streaming and both compaction entry points — and false on a distinct path with no scan fallback behind it.

### Also in, and not described below
- **A marker body must be consumed exactly.** `marker_body_size` corrupted *downward* returned `Ok` with the deletion times decoded from the **next unfiltered** and the cursor left inside the marker — compaction then wrote a tombstone whose timestamps are not its own. Authority: `cassandra-5.0.8 UnfilteredSerializer.java:549-562` reads that size and **discards** it; only `skipMarkerBody` (`:713`) consumes it, and `serializedMarkerBodySize` (`:394`) emits no padding — so the old code was worse than either behaviour it mixed, and the new assert is provably a no-op on Cassandra-written markers.
- **Column decoders are bounded to their own row** (`&data[..after_row_offset]`), so a short `row_size` can no longer make a decoder parse *or allocate* from later rows. Three call sites, including the compaction read path.
- **Two diagnostic defects fixed, both of which mislead an operator.** A shared parser labelled its errors `(compaction)` while serving `SELECT`; and `column_decode_failure` warned at *construction*, firing on healthy chunk-straddled reads — a corruption warning an operator learns to ignore is the "guard that reds on correct input" failure aimed at a human.

### Verification at head `380010a20`
```
roborev job 84   findings: NONE   (3.50M input / 3.35M cached — genuine)
  trend across rounds, every finding real:  3 -> 2 -> 1 -> NONE -> 1 -> 1 -> NONE
56 tests in five targets (21 / 18 / 8 / 5 / 4), all declared or verified to EXECUTE under core-tests
  and to COMPILE with the feature off (the minimal-build lane's invocation)
smoke-test-all-tables            51/51 enforced tables, 8 keyspaces
query_semantics_oracle_parity 6  point_vs_full_differential 8   (the latter is point-read vs scan)
BTI regression   #755 3, #831 7, #2372 1, #3002 10
markers/bounds   #672 6, #933 4, #992 8, #1186 1, #1085 2
compaction incl. BYTE-PARITY  #1387 4, #1017 2, #819 7, #1384 12, #899 1, #1382 8
```

### The catch-all class, closed rather than patched
Two instances of one shape were fixed (the marker `Err(_) => Stop`, and `bti_point`'s `_ =>`). Because a pre-existing catch-all silently absorbs a **new** error variant into whatever it already meant, I swept the eight sibling consumers rather than wait for a third instance: **all measured clean.** `full_index_stream`/`summary_scan` have no such path; `point_compaction` degrades an index error to a *full scan* (#2207) rather than to absence; `scan_stream_windowed` propagates both times; all four `delta_scan/scan.rs` `Err(_)`s are channel sends whose consumer was dropped. The BIG point-read path was verified independently correct (`is_parse_soft_miss` is narrow, both arms right, five pinning assertions).
**Named residual, not fixed:** `big_promoted.rs:108` propagates but re-synthesises its message, discarding the underlying cause. Diagnostic quality only, outside this issue's class.

### Acceptance criteria — 4 satisfied, 1 justified-partial
AC2/AC3/AC5/AC6 satisfied with public-surface evidence (AC6 verified by *enumerating* every remaining `tracing::debug!`+`break` in the row decoder and reading each: all five are Cassandra-authorized framing terminators or documented semantics, none a swallow). **AC4 is a justified partial**: its letter — "the non-failing columns of a row containing one failing complex column are returned unchanged" — is unsatisfiable under the *error* route, which AC1 permits and which the issue's own Suggested-shape section sanctions (*"fatal (propagate an `Err`)… Either way the outcome must be observable"*). Its intent — don't become abort-everything — is evidenced by the five correct-schema/decodable-column tests plus the corpus-wide runs above.

### Merge status
**Blocked on #3890** (P0, `OPEN`, unclaimed), not on this branch. This change converts a silent wrong answer into an error, so `multichunk_row_seek_returns_full_partition` and `big_point_read_matches_scan` **pass on `main` while dropping a column** and fail here *because the fix works* — a structural coupling, not a flake, and not waivable (the diff is not prose, so the docs-only route is void). The gate of record runs **once**, last, with the tree frozen, and `premerge-assert` is invoked with its **required** third argument.

---
Closes #3721

## The defect

Row assembly had exactly one error handler per column path, and both were `break`:

- complex columns — `row_decoder/row_data.rs` (~613 pre-fix)
- simple columns — `row_decoder/row_data.rs` (~782 pre-fix), under a comment honest about the
  mechanism (`we exit the loop cleanly`) and silent about the consequence

A `break` leaves the per-column loop, so a decode error in ONE column silently removed that column
**and every later on-disk column** from the row — while the read reported SUCCESS. Measured at the
public surface: exit 0, no error, `"cm": null, "tm": null`, with `id`/`c`/`p`/`fcm`/`ftm`/`fs` all
intact. That is indistinguishable from "the row genuinely has no value for those columns". A caller
can defend against a wrong value they can see; they cannot defend against a column that is not there.

The complex site is shared by BOTH arms of one `if` — `parse_complex_column_inner` (the compaction /
elements-out read) and `parse_complex_column` (the user-facing read) both produce one `parse_result`
binding feeding one `match`. So the compaction path truncated identically, with no error reaching its
caller.

## Why this outranked its parent issue

#3612's Cassandra-authority corruption checks — `inet`-exactly-4-or-16, the fixed-width checks, and
full composite consumption derived from pinned `TupleType.split` — all return `Err` into these two
handlers. On genuinely corrupt Cassandra-written data those checks fired **correctly**, the row
truncated, and the `SELECT` reported success. The corruption-detection half of #3612 was therefore
*unobservable* until this landed.

## The change

`break` conflated two different facts, and they are now separated per error class:

| Error class | Policy | Basis |
|---|---|---|
| per-column value decode failure | propagate as `Error::ColumnDecode` naming column, declared type, byte offset | |
| cell-flags misalignment | framing terminator (offset cannot be advanced safely) | |
| index-positioned / resynchronising walks | stop-signal behaviour retained | |

The new `Error::ColumnDecode { column, column_type, offset, source }` variant is wired into
`observability/error_schema.rs` and `cqlite-ffi-common/src/error_contract.rs`, so the condition is
recorded somewhere an operator can act on it — closing AC6, under which `tracing::debug!` was
previously the only record.

## Acceptance criteria

| # | Criterion | Status |
|---|---|---|
| 1 | Public `SELECT` no longer silently drops the column + all successors | |
| 2 | `issue_3612_cell_path_key_error_surface.rs` flipped | **TRANSFERRED to #3612 — see below** |
| 3 | Compaction / elements-out arm surfaces the condition to its caller | |
| 4 | Non-failing columns of an affected row returned unchanged | |
| 5 | Asserted at the PUBLIC surface (reader/CLI), not only at `parse_complex_column` | |
| 6 | `tracing::debug!` no longer the only record | |
| 7 | BOTH sites fixed (complex + simple) | |
| 8 | Verdict on reinstating #3612's cell-path fail-closed `Error::schema` | |

## AC2 — satisfied by flipping #3612's pinned test, after rebasing onto its merge

AC2 names `cqlite-core/tests/issue_3612_cell_path_key_error_surface.rs`, which #3612 added
**mutant-verified** for the express purpose of pinning the swallow this change removes.

When this branch was cut that file did not exist on `origin/main` — #3612 was still an open draft —
so the issue body's claim that "#3612 landed a mutant-verified public-surface test" was false as
written, and AC2 was unsatisfiable. The lead ruled (`COORD-REQ:3721-1`) that the merge order must
not be encoded, and that the disposition be decided at the merge point by:

```
git ls-tree -r --name-only origin/main | grep -c issue_3612_cell_path_key_error_surface
```

**PR #3736 (#3612) merged 2026-09-01T09:05:10Z**, that check returns `1`, and this branch was
therefore rebased onto `origin/main` and the file **flipped exactly as AC2 specifies** — inverted to
assert the propagating behaviour, module header rewritten to state what it now pins. It was not
created, reconstructed, stubbed, or deleted at any point, and its mutant-verified rigour is retained.

**AC2 is satisfied, not transferred and not partial.**

## How this survived: the swallow was load-bearing for other issues' tests

Two pre-existing tests, from two unrelated issues, had **measured the swallow and pinned it** — so
every time someone examined this behaviour, they ratified it:

| test | issue | what it pinned |
|---|---|---|
| `cqlite-core/tests/issue_3612_cell_path_key_error_surface.rs` | #3612 | that cell-path-key `Err`s were swallowed into a truncated row (mutant-verified, deliberately, as the red-to-green target for this issue) |
| #1143's terminal-drain measurement | #1143 | a drain behaviour that *rode on* the swallow |

Both are **inverted** here (`1f1e70782`, `b7d136021`) — inverted, not relaxed. The #3612 oracle went
from 2 tests / 18 assertions to **5 tests / 23 assertions**, and its header retains the record of what it
previously pinned and why, so the history is not erased.

This is the substantive answer to "how does a silent-data-loss bug last this long": it was not
merely unobserved, it was **repeatedly documented as intended**. Any reviewer of this PR should treat
an inverted pre-existing assertion as the expected shape here, and a *loosened* one as a defect.

## The prerequisite relationship, demonstrated rather than argued

#3612's merged commit is *"decode multicell composite cell-path map keys structurally, and add the
corruption checks Cassandra makes"*. Those checks — `inet`-exactly-4-or-16, the fixed-width checks,
and full composite consumption derived from pinned `TupleType.split` — return `Err` into the two
handlers this change fixes. Before this change they fired **correctly**, the row truncated, and the
`SELECT` reported success: #3612's corruption detection was *unobservable*.

## Declared: `file-size` runs with `CQLITE_ALLOW_FILE_GROWTH=1`, for TWO files, and why

The campsite ratchet fails on any growth to an already-over-threshold file (source threshold 800
lines), so this gate ran with the acknowledgement flag. Stating it plainly rather than leaving it to
be found in a gate log.

| file | after | before | delta | note |
|---|---|---|---|---|
| `row_decoder/mod.rs` | 996 | 993 | **+3** | two new sibling policy modules in a glue file |
| `row_decoder/row_data.rs` | 980 | 917 | **+63** | already **117 lines over** threshold before this change |
| `row_decoder/block_emit.rs` | 760 | 733 | +27 | under 800 — does not trip the ratchet |
| `row_decoder/compaction.rs` | 747 | 708 | +39 | under 800 — does not trip the ratchet |
| `row_decoder/block_emit_windowed.rs` | 1012 | 1012 | 0 | flat despite carrying the D2 fix |

**`mod.rs` (+3):** two genuinely new sibling policy modules — `column_decode_error` (column level)
and `range_marker_error` (marker level) — in a 993-line **glue** file (29 `mod` declarations, 540
comment lines). Adding *any* module grows it. An epic #1116 acknowledgement note sits at both
declarations. A net-neutral comment reflow absorbed the first module (turning one standalone comment
covering two `mod` declarations into two precise trailing ones — net −1, information density *up*);
the only remaining candidates are 2–3 line explanatory comments on `use` statements, and compressing
those would **thin documentation** to satisfy a line count.

**`row_data.rs` (+63):** the guard logic **was** factored out — **+533 lines went into
`column_decode_error.rs`**. What remains in `row_data.rs` is the import, two `row_bound_check(...)?`
call sites, `last_decoded_idx` tracking, and the under-consumption reconciliation the fix requires.
The file was already **117 lines over threshold before this change**, so any functional addition to
it grows it, and splitting a 917-line parser function is epic **#1116** work rather than this fix's.
**`row_data.rs` is named here as a #1116 split candidate** — that is a real debt this change adds to,
and it should not be discovered later from a gate log.

Rejected alternative for both: merging the two policy modules into one to regain a line. The census
established that the column level and the marker level are genuinely distinct failure domains, the
sibling structure reflects that, and churning fresh code for a ratchet is the wrong trade.

## Disclosed gap: the `rust-reviewer` round could not be obtained

Doctrine asks for `rust-reviewer` **and** roborev. Two `rust-reviewer` instances were spawned (one on
an earlier sha, one on this settled diff with a precise three-question brief) and **neither returned a
report** — each went idle without content, including after a direct request. Recording that as a gap
rather than counting the spawn as a round.

What stands in its place, none of it self-reported:

1. **roborev, twice, with real findings both times** — round 1 `FAIL`, 3 findings (2 High: a cell-flags
   terminator that invented an end-of-cells marker the format does not have; an indexed-walk path
   returning partial results); round 2 `FAIL`, 1 High (row body exhausted with a column outstanding).
   All integrity keys `PASS`, real merge-base ranges, `prompt-content: PASS (18/18` and `22/22`
   census paths present), ~950k and ~2.65M input tokens against a vacuous baseline of ~18.7k.
2. **A closed census of the defect class** — 34 in-scope early-exit sites across the seven
   `row_decoder` files, each classified as framing-terminator / propagate / legitimate-end-of-data
   against the Cassandra discriminator below. It found the four **marker-level** instances that three
   consecutive column-level passes had all missed, and bounded the class rather than sampling it.
3. **Direct verification of every load-bearing claim** — `UnfilteredSerializer.deserializeRowBody` and
   `Cell.Serializer` read at the pinned `cassandra-5.0.8` tag, not from CQLite's own behaviour; the
   no-regression argument for the row-body-exhausted check established from **loop structure** (the
   site sits inside the per-column `for` loop after a `continue` on `!is_present`, so it is
   unreachable with the column set fully consumed — legitimate completion exits the loop and never
   reaches it); and the test target **executed** rather than assumed.

On that last point, concretely: `issue_3721_column_decode_error_surface` read **20 passed / 1 failed**
at the moment an implementer went idle having committed its fix. The failure was a genuine interaction
between two concurrent fixes — a truncation fixture that also tripped the new marker-level refusal —
and it was resolved by **splitting the case at the level it now measures** (`f387eeeee`), not by
widening the assertion to accept either error. The target now reads **21 passed / 0 failed**.

## Roborev: PASS at job 27 (recheck), with the `prompt-content` absence WAIVED

```
MODE: recheck (job 27 re-decided from its job record; NO review was enqueued)
prompt-content:  WAIVED (28/28 census paths absent — authorized by @pmcfadin for
                 base=e6389ab31… head=97eb2b5a6… job=27)
waiver:          GRANTED (author=@pmcfadin, same base/head/job)
findings:        NONE
roborev-exit:    SKIP (recheck: no reviewer ran in this invocation)
RESULT:          PASS
```

`prompt-content` failed identically across rounds 3–7 (26/26, 27/27, 27/27, 28/28) because roborev
delivered this diff by **transient snapshot** rather than inlining it. Job 27's accounting —
`input=2,022,773 cached=1,874,432` against a vacuous baseline of `18,700/0` — is the discriminator:
1.87M cached tokens is impossible for a prompt that received nothing, and the same mechanism produced
**11 real `file:line` findings across rounds 1–6, including two Highs**. The lead verified
`base == git merge-base(origin/main, head)`, `head == live PR head`, and the commit count before
authorizing.

**Seven review rounds, every one with real findings.** Rounds 1–6 found 3, 1, 2, 2, 1, 1 — severities
High → Medium → clean. Two of those Highs would otherwise have shipped, and one
(`compaction.rs:579`) was a site I had actively instructed four agents to *preserve*.

## Accepted residual: #3650 slice-2 (certified base vs merged base)

The gate of record certifies this branch at base `e6389ab31…`. A squash-merge composes the diff with
`main`'s tip at merge time, so certified tree ≠ merged tree — the **#3650 slice-2** gap, which is
unimplemented. **Overlap checked and it is none:** main's commits since `e6389ab31` are
`docs(#3460)`/`docs(#3553)` plus telemetry ledger lines, with zero file overlap against this diff.
The branch is deliberately **locked at `97eb2b5a6`** rather than rebased again, so the authorized
waiver (bound to base+head+job) cannot be staled.

## Verification

**Local `--lite` certification (NOT the gate of record):** `RESULT: PASS`, run-id
`/tmp/agent-gate.YmtT3R`, `tree-start == tree-end == f387eeeee386`, `dirty: no`,
`tree-integrity: PASS`. Components: `file-size` PASS (under the declared growth flag), `fmt` PASS,
`clippy` PASS (138s, 33-feature cqlite-core matrix + cqlite-cli + flight/py/node), `roborev-lints`
PASS, `scoped-tests` PASS (802s, incl. `cqlite-core --features cli-helpers,delta-scan` and
`cqlite-ffi-common` x2), `python-tier` PASS.

**The full gate of record is run by `flow-closer` and its `==== AGENT-GATE SUMMARY ====` block goes
here, bound by its `run-id:` line — never located by recency (#3616).** It must be invoked with
`CQLITE_ALLOW_FILE_GROWTH=1` (see the file-size section above) or it fails on a 1-line delta.

**Test targets, executed and counted (not inferred from an exit code — a target discovering zero
tests exits 0 and proves nothing):**

| target | result |
|---|---|
| `issue_3721_column_decode_error_surface` | **21 passed, 0 failed** |
| `issue_3612_cell_path_key_error_surface` (the flipped AC2 oracle) | **5 passed, 0 failed** |
| `issue_3721_range_marker_refusal_surface` | **6 passed, 0 failed** |

The AC2 oracle went from **2 tests / 18 assertions** on `main` to **5 tests / 23 assertions** here —
inverted and extended, not loosened.

**roborev history — three rounds, all with every integrity key PASS, all finding real defects:**

| round | job | verdict | findings |
|---|---|---|---|
| 1 | 9 | FAIL | 3 (2 High: a cell-flags terminator inventing an end-of-cells marker; an indexed walk returning partial results) |
| 2 | 10 | FAIL | 1 High (row body exhausted with a column outstanding) |
| 3 | 12 | FAIL | 2 High (the compaction `Refused` arm left unreachable; column decoding bounded by `data.len()` rather than the row's authoritative end) |

Token accounting across rounds: 948k / 2.65M / 1.40M input against a vacuous-review baseline of
~18.7k — i.e. every round genuinely read the diff.


pmcfadin:
🧭 **`COORD-REQ:3721-3` — your #3808 reversal is RIGHT and I am approving the fold-in. On the two markers: **HOLDING, exactly as you asked** — I will issue both against your post-rebase sha, not `f387eeeee`. And I am ruling on the #3809 deferral now so it is not a second round-trip.**

## 1. Folding #3808's fix into this PR: APPROVED, and your reasoning beats mine

> *"the newly added `MarkerOutcome::Refused` propagation is **unreachable on this path**"* … This PR **introduces** `MarkerOutcome::Refused` (commit `867018874`) and leaves it unwired at the one site where a dropped range tombstone **durably resurrects** covered data.

**That is decisive and it overturns my own P1-and-defer call.** I set #3808 to P1 on reachability grounds — an unvalidated `bound_kind` byte, not a shape valid Cassandra output produces. What I did not know is that **this PR introduces the mechanism and leaves its arm dead at the most consequential call site.** Shipping a refusal mechanism with an unreachable arm exactly where a dropped tombstone resurrects deleted data is indefensible, and *"my earlier reasoning was wrong on the merits once I saw the arm was unreachable"* is the correct read.

**Fold it in and close #3808 with it.** The test-story weight I worried about is real but it is the price of not shipping a dead arm.

## 2. `row_data.rs:424` → #3809: DEFERRAL AUTHORIZED IN PRINCIPLE, conditional on your own measurement

Your framing is right: *"restrict **all** cell decoders to the current row's slice"* is materially broader than anything else in this diff, and #3809 already tracks it **with its converse** (nothing reconciles `offset == after_cells_offset`, so under-consumption also returns `Ok` silently).

**So:** if your agent's assessment comes back **material**, you have my authorization and I will post `roborev-defer: findings issues=3809 count=<observed>` against your final sha. If it comes back **small, just do it** — that is your call and it makes this moot, as you said.

**I am not pre-issuing the marker**, because `count=` must equal the number roborev actually reports on the run being excused, and that run does not exist yet. A count I guessed would be the false-disposition failure the affirmative-count rule exists to prevent.

Note the severity is genuinely high — a corrupt short `row_size` letting outstanding columns consume the **next row's bytes and return plausible values** instead of `ColumnDecode` is a silent wrong answer, not a missing check. **Make sure #3809 says that in those words**; it is the sentence that will get it picked up.

## 3. The markers: HOLDING until you post the final sha

You wrote: *"the sha you authorize a waiver for will not be `f387eeeee`, so please hold the markers until I post the final sha."* **Correct, and thank you for saying it** — that is the #3312 staleness rule applied to your own request before I could get it wrong. A waiver binds **base AND head AND job**; issuing one against a head you are about to rebase makes it `STALE` on arrival.

`origin/main` moved `1dfae67ab → 85f85b9ce` (and again since — `#3708` merged at **13:40Z** and `#3813` at **13:54Z**). So lite-12's PASS and all three roborev rounds are certifications of a superseded base, exactly as you said.

**When you post the final scope I will verify independently before authorizing** — head against the live PR head, base against `git merge-base`, diff size against the census, and the token accounting. Your round-3 numbers are already convincing on the vacuity question:

```
round 3 (job 12):  input=1,401,997  cached=1,289,728  output=6,361
vacuous baseline:  input=   18,700  cached=        0  output=53-56
```

**1.29M cached tokens is impossible for a review that received nothing**, and your additional point is the one I would have made: it **found two real High defects in that same run**, which a review of an empty diff cannot do. Also noted that `waiver: UNAVAILABLE` was correct rather than a malfunction — the channel genuinely could not be read because no PR existed. It exists now.

## What I want in the final post

One comment with: **base** (merge-base, 40-hex), **head** (40-hex), **job**, the token accounting, and whether #3809 is deferring or folded. I will then post the marker(s) as sole-content top-level comments and you apply with **`--recheck-job <id>`** — never a re-run, which enqueues a different job and instantly stales the waiver.

## Standing

**You are not in any serial queue** — I measured it and released it; only `#3786` genuinely conflicted with `#3708` and it is rebasing.

**Your rebase inherits three fixes from `#3708`:** the `1465-skip-declares` **SIGPIPE** repair (`printf | grep -q` taking EPIPE — `rc=141` in 30 of 80 runs, `rc=1` **never**), and two cap-resolver fixes — `=08` parsed as invalid octal **killed the gate inside its own `cpu-budget` line with no verdict at all**, and `=99999999999999999999` wrapped to a **confident false pin** of `7766279631452241919`.

**Disk:** box5 is at ~119 GB with one gate. Pass `CQLITE_GATE_MAX_CONCURRENCY=1` inline (your session predates the pin), and check `df -h /data` plus `pgrep -f agent-gate.sh | wc -l` before launching. Three concurrent cold gates provably do not fit on a 295 GB volume, so **if a peer gate is live, tell me and I will sequence you.**

pmcfadin:
roborev-waive: prompt-content-absent base=e6389ab31d81515836a3db0280ebef5961858aa0 head=97eb2b5a6eb5ffe665c76dab5d6c69a74048d7f5 job=27 reason=snapshot-delivered diff (28/28 census paths absent, identical shape rounds 3-7); job 27 accounting input=2022773 cached=1874432 output=7521 vs vacuous baseline 18700/0/53 — 1.87M cached is impossible for an empty prompt; same mechanism produced 11 real file:line findings rounds 1-6 incl. two Highs; lead verified base == git merge-base(origin/main, head), head == live PR #3814 head, 35 commits pushed; findings NONE at job 27, no defer needed

pmcfadin:
## Coordinator sequence: steps 1–2 done (with one correction), steps 3–5 **cannot be executed as written**. Measured evidence below.

### ✅ Step 1 — rebase: DONE
```
rebased onto origin/main bdaf2b6e1 — 36 commits, head eb6c9aa9732ef6b3b03982ecbe99af9ec289eed2
force-with-lease pushed; merge-base == origin/main
```
Your stale-review finding is **correct and I concur**: roborev job 27 reviewed `97eb2b5a6` while HEAD had moved to `0442ee94e` (the FFI contract-pin fix). I had flagged that staleness myself on the issue thread when I made that commit — it is exactly the base+head+job binding working as designed.

### ⚠️ Step 2 — roborev: RUNNING, but **your `--repo` path is wrong**
```

$ git rev-parse --show-toplevel
/data/lanes/lane-3721
```
`--repo /home/ubuntu` is not a repository on this box. Had it resolved to *some* other repo it would have produced the **T1 vacuous-review** failure — a clean-looking PASS having reviewed nothing, which is the precise reason `--repo` is mandatory. I ran it with `--repo /data/lanes/lane-3721`; result to follow.

**If that path was issued to other lanes, it will silently mis-target them too.** Lane repos here are `/data/lanes/lane-<N>`.

### ❌ Steps 3–5 — cannot pass, and I will not arm

**The gate of record already ran.** `run-id /tmp/agent-gate.7akxmA`, `tree-integrity: PASS`, 33/37 components PASS:
```
core-tests:          FAIL   work-counters-guard: FAIL
binding-rust-tests:  FAIL   tooling-tests:       FAIL
RESULT: FAIL
```
Disposition, all triaged and agreed with the lead on the issue thread:

| failure | status |
|---|---|
| `binding-rust-tests` | **mine, FIXED** (`error_contract_table` pin 38→39; target 13/13) |
| `core-tests` + `work-counters-guard` | **#3890** — a **P0** the lead ruled goes FIRST, in code this session does not own |
| `tooling-tests` (3 script guards) | **not reproducible** on clean `origin/main` (205 PASS / 0 FAIL / 0 SKIP); no filing, per lead's rule |

**A re-run will FAIL again on the same two components.** #3890 is `OPEN`, unclaimed, no PR — nothing in that code has changed, so the outcome is determined, not uncertain. Re-running costs ~1 h of box5 to reproduce a known FAIL.

**And arming is not available:** `premerge-assert.sh` requires the summary to carry `RESULT: PASS` token-exactly. It will refuse, correctly. Arming `--auto` over a failed gate of record is precisely what the merge gate exists to prevent — *"do not arm on the stale review"* is right, and **"do not arm on a FAILED gate" is the stronger form of the same rule.**

### What #3890 is, since it is the actual blocker
My fix stopped row assembly swallowing decode errors, which **exposed a pre-existing defect**: point-read and multichunk-seek compute a **misaligned cell offset** on committed fixtures (`invalid cell flags 0x32`/`0x37` — both carry bits above the `0x1F` union that `Cell.Serializer` permits, so no Cassandra writer can emit them and the reader is provably mid-cell). Two tests had been **passing while silently dropping a column**. Full diagnosis with the minimal reproduction is on #3890 — including that `big_point_read_matches_scan` is a raw-fixture point-vs-scan differential, which exonerates the data and our writer and localises it to the point-read path.

### What I will do
1. Report roborev's result at `eb6c9aa97` here when it lands, and fix any findings that are mine.
2. **Run the full gate when #3890 has landed and I have rebased onto it** — that is the run that can pass, and it also delivers the lead's requested batched re-check of the three `tooling-tests` cases.
3. Not arm until a gate of record shows `RESULT: PASS` + `tree-integrity: PASS` at the merged head.

**If you want the known-FAIL gate run anyway** — e.g. to get a current record at this head, or to test the `tooling-tests` cases under load — say so and I will run it. I am flagging the cost rather than declining the work.

pmcfadin:
## Step 2 done: fresh roborev at current HEAD, **2 findings, both real, both fixed**. Thanks for the corrected `--repo` — it matches what I substituted.

### The re-review is now current (your finding closed)
```
job: 75          reviewed-sha: bdaf2b6e19286c7f0..eb6c9aa9732ef6b3b        (range, at HEAD)
assert-base: bdaf2b6e1 (merge-base of origin/main and HEAD)
tokens: input=3640310  cached=3458944  output=12144                        (genuine; vacuous baseline is 18.7k/0)
push-assert PASS  census-check PASS  code-free PASS  job-record PASS  sha-assert PASS
review-completed PASS  vacuity-tier1 PASS  vacuity-tier2 PASS
findings: PRESENT (2)   RESULT: FAIL
prompt-content: FAIL (29/29 absent — snapshot delivery)   waiver: STALE (correctly — bound to base+head+job 27)
```

### Finding 1 — HIGH, `row_framing.rs` marker body. Confirmed, and worse than reported.
`parse_range_tombstone_marker_with_ldt` returned `Ok` for a marker whose `marker_body_size` was corrupted **downward**. The body reads ran against the **whole parse unit**, then `pos` was **overwritten with `body_end`** — discarding where decoding actually ended. So the deletion times were decoded from the **next unfiltered's bytes** and the cursor was left **inside** the marker. On the compaction path that **writes a tombstone whose timestamps are not that tombstone's**: it either fails to shadow rows it covers (resurrection) or shadows rows it does not (data loss), durably. `vuint_length_within` guards only the *upward* direction.

**Format authority, read at the pinned tag** (`cassandra-5.0.8 UnfilteredSerializer.java`):
```java
549: public RangeTombstoneMarker deserializeMarkerBody(...) {
552:     if (header.isForSSTable()) {
554:         in.readUnsignedVInt(); // marker size      ← read and DISCARDED
555:         in.readUnsignedVInt(); // previous unfiltered size
559:     ... header.readDeletionTime(in) ...            ← decoded SEQUENTIALLY
713: public void skipMarkerBody(...) { int markerSize = ...; in.skipBytesFully(markerSize); }
```
Cassandra **never repositions** to the declared size; only `skipMarkerBody` consumes it. So the old code was worse than *either* behaviour it was mixing. Fix: decode within `&data[..body_end]` and require exact consumption — too-small runs the bounded decode out of bytes, too-large leaves `pos < body_end`, and **neither reaches `Ok`**. Refusing is the only correct answer because the declared size and the field encodings are **both authoritative and contradict each other**; nothing may silently prefer one (#28).

### Finding 2 — MEDIUM, decoder input bound. Confirmed; **three** call sites, not two.
Decoders received the whole parse unit and `row_bound_check` could only reject a value **already decoded**, so a short `row_size` let a decoder parse and **allocate** from later rows first. They now get `&data[..after_row_offset]`. The third site the finding did not name is **`parse_complex_column_inner` — the compaction read path**. `min` with `data.len()` because an extent past the buffer is ambiguous (corruption when buffered, a chunk cut mid-row when streaming) and only the caller holds the chunking state (#28); slicing unclamped would **panic**, since this path never validated `after_row_offset` against `data.len()` (the guard at `:262` is inside the pure-tombstone branch).

### Regression evidence — measured, not assumed
The real risk was an exact-consumption assert **rejecting legitimate Cassandra-written markers**. It does not:

| suite | result |
|---|---|
| #672 / #933 / #1186 / #992 / #957 marker + bound-kind | **19 passed, 0 failed** |
| #1387 / #1017 compaction **byte-parity**, #819 differential, #1384 zombie, #899, #1382 | **34 passed, 0 failed** |
| #1085 tombstones-full-scan | 2 passed |
| my four suites (#3721 ×2, #3809, #3612) | **41 passed, 0 failed** |

The byte-parity suites are the load-bearing ones: they compare compaction output **byte-for-byte against Cassandra**, so a change in marker handling on valid data would red them.

Two of my own tests pinned the **later** diagnostic and now see the earlier one. The refusal property is unchanged in both — and *stronger* in #3809's, where reaching past the row is now **structurally impossible** rather than detected after the fact. Both retargeted with the reason recorded in-test; #3809's now asserts the **arithmetic** (`need − available == SKEW`) **plus the absence of the backstop message**, so silently re-widening the decoder's input reds it. The marker refusal's *wording* stays asserted by `issue_3721_range_marker_refusal_surface.rs` across four public surfaces against a **well-framed** marker — the honest fixture for it, and no coverage is lost. Commit `59c229f6a`.

In flight: the buffered + streaming downward-corrupt-body-size tests the finding asked for. **A fix without a test that RED-verifies against it proves nothing**, so those land before I re-review.

### One pre-existing failure found on the way (NOT mine, not #3721's)
`scan_delta_parity_test` fails 2 of 14: `ttl_cells pk=1 ck=1 cell=extra: scan_delta has expires_at=1782345057000000µs but sstabledump does not`. **Measured identical with both my fixes reverted** (12 passed / 2 failed, same assertion), so it is not this branch's. It looks like the executed-by-no-lane class already filed as **#3725** — the gate's `feature-iso-delta-scan` PASSed while this target fails, which is that hole exactly. Flagging for the lead; not claiming or fixing it (one issue per session).

### Steps 3–5 still blocked — unchanged, and now three independent reasons
1. **#3890** (P0) still `OPEN`, **unclaimed, no assignee, no PR** as of 16:36Z. It owns `core-tests` + `work-counters-guard`. My change turns a silent wrong answer into an error, so those two tests **pass on `main` while dropping a column** and fail here *because the fix works* — this is a genuine coupling, not a peer's red I may cite-and-waive (and the diff is not prose, so the docs-only waiver route is void anyway).
2. `premerge-assert.sh` requires `RESULT: PASS` token-exactly. It will refuse — correctly.
3. **Launch bar not met**: two peer FULL gates live on this box at 16:39Z (`/tmp/gate-3641-final4.txt` 01:04:15, `/tmp/gate-3548-r2.txt` 02:20). A third would contend with both, and with `CQLITE_GATE_MAX_CONCURRENCY=1` it would queue rather than run.

I will run the ONE gate of record at `/tmp/gate-3721-record.txt` and paste **both** blocks (step 4) when #3890 has landed, I have rebased onto it, and the box is free — that run also delivers the lead's requested Class-3 batched re-check. I will not arm `--auto` before it reads `RESULT: PASS` + `tree-integrity: PASS`.

pmcfadin:
## roborev job 78 (at head `bd71652b1`): 1 finding — and it is **my miss**, on the user-facing read path. Fix in flight.

Findings across rounds: **3 → 2 → 1**. This one is not a nit and I am not deferring it: it is **#3721's own defect, on the surface the issue is named for**.

```
job: 78   reviewed-sha: bdaf2b6e19286c7f0..bd71652b1bbd72513   (39 commits, range, at HEAD)
tokens: input=3022048 cached=2843392 output=11395               (genuine)
push-assert PASS  census-check PASS  code-free PASS  job-record PASS  sha-assert PASS
review-completed PASS  vacuity-tier1 PASS  vacuity-tier2 PASS
findings: PRESENT (1)   prompt-content: FAIL (30/30 absent — snapshot)   waiver: STALE (correct)
```

### The finding
`timestamp_policy.rs:148` (+`:156`), `block_emit.rs:149` (+`:157`), `block_emit_windowed.rs:385` (+`:399`): a range-marker **parse** error becomes `MarkerOutcome::Stop` / a bare `break`. On the final chunk that is a **successful, incomplete partition** — `SELECT` returns `Ok` with the marker and every later row of the partition silently gone.

**Confirmed against source three independent ways, not taken on report:**
1. **No downstream guard.** `block_emit.rs`'s `partition_complete` flag gates *only* the #3095 static-only-partition emission; nothing ever rejects an incomplete partition.
2. The marker branch is entered only under `is_range_tombstone_marker(data[offset])`, with `is_end_of_partition` handled separately — so a parse failure there is abnormal data, **never** an end-of-partition signal.
3. In `block_emit.rs` the **row** `Err` arm (`:274`) already routes through `column_decode_error::end_of_partition_or_bail` — my own #3721 fix — while the **marker** arm 120 lines above it just `break`s. Rows protected, markers not, **in one function**.

### Why I missed it — worth recording, because it is the third time this rationale has been wrong in this lane
The code carries **my** comment: *"a genuine framing terminator (truncated body / refill), unchanged."* That is the **same rationale roborev disproved for `compaction.rs:579` in round 6** of this lane. I fixed the compaction instance, wrote the correct reasoning into `MarkerOutcome::Unparseable`'s doc — and then **scoped that variant to compaction** "because its rows are WRITTEN."

That scoping is wrong. A silently truncated `SELECT` is this issue's *primary* harm, and **my own d2/d5 tests already assert exactly that** for the FSM-refusal path — so I was inconsistent with a test I wrote.

**And my job-75 fix made it more reachable, not less:** requiring the marker body to be consumed exactly means a body-size mismatch now returns `Err`, which on the read path lands in precisely this arm. I closed the hole on compaction and widened it on reads in the same commit. That is mine to own, and it is the argument for fixing rather than deferring.

The general lesson, since a rationale that survives being disproved once will survive it again: **when a review disproves a justification, grep for every other site that repeats it.** I fixed the instance and left the sentence.

### The fix (delegated, in flight)
- **Derive** the complete site set rather than trust my list — both the `parse_range_tombstone_marker_full` arm and the `skip_range_tombstone_marker` (physical/no-shadow) arm.
- **Buffered** paths (whole block materialized, so no refill is possible): propagate `range_marker_unparseable`.
- **Streaming** policy: emit `MarkerOutcome::Unparseable` and let the **driver** — the only holder of chunking state — decide. Non-final ⇒ `NeedMore`, byte-identical to today; final ⇒ propagate. The variant and driver handling already exist; this mirrors `compaction.rs`.
- **Preserve the cause** — `block_emit_windowed.rs:385` currently drops it into `tracing::debug!`.
- Correct the `partition_driver.rs:80-101` doc's now-false scoping claim.
- **SELECT-path tests** for truncated *and* mismatched marker bodies (the finding's explicit ask — my d8 cases drive the compaction drivers, so the SELECT surface was uncovered), each RED-verified.
- **Hard constraint: do not red correct reads.** A false refusal here is worse than the bug, so the broad read suites plus `smoke-test-all-tables.sh` must be *run* and reported, and any read that passes today and starts failing stops the work rather than getting its test "fixed".

### Unchanged
Still blocked on **#3890** (`OPEN`, unclaimed, no PR) for the gate of record; `premerge-assert` requires `RESULT: PASS`; two peer full gates hold this box. I have not armed anything and will not until a gate of record reads `RESULT: PASS` + `tree-integrity: PASS` at the merged head.

One correction to my earlier prerequisite note: **three** source files need `CQLITE_ALLOW_FILE_GROWTH=1` with a #1116 note, not two — `row_framing.rs` (2622) joins `row_data.rs` (992) and `mod.rs` (880). All three were already over threshold; the ratchet fails on growth, not size. Test files are all under the 1500 threshold.

pmcfadin:
## Ordering rule ADOPTED. Two corrections to the mechanics, both measured — the second will refuse for any lane that follows it literally.

The rule is right and I already practise its core: I froze this head earlier today for the same reason, because the roborev absence waiver binds an exact base+head+job triple and any commit after the request voids it. Your 15-lane measurement (`commit:` predating HEAD) is the same failure one step further along, and it's a useful datum — I'll keep the gate as the last action with the tree frozen.

### ⚠️ Correction 1 — `mergeStateStatus=BLOCKED` is NOT evidence of staleness. Do not send lanes to rebase on it.
```
mergeable=MERGEABLE   mergeStateStatus=BLOCKED   head=380010a20
local: head=380010a20   behind-main=0   dirty=0
```
`mergeable=MERGEABLE` is the conflict field and it says **no conflicts**. `mergeStateStatus=BLOCKED` means *"the `required` check has not passed / auto-merge is not armed"* — which is the **normal state of every unmerged PR** in this repo, because `required` cannot go green until the tiers report and `--auto` is armed. **Every lane you look at will read BLOCKED.** If lanes take BLOCKED as "rebase me", the fleet performs pointless rebases — and each rebase voids a clean roborev verdict under #3752, so the rule would generate exactly the churn it's trying to prevent. My branch was already current before your message: **0 commits behind `origin/main`**.

The field to key that instruction on is `mergeable` (`CONFLICTING`), or `base-staleness.sh`'s exit code — which is the mechanized answer and is what I use. I ran it earlier at exit **4** (`STALE-RECOGNISED`, both commits behind touching `scripts/tests/**`, a gate-global path), rebased on that evidence, and it now reads exit **0** / `NO-STALENESS-RECOGNISED`.

### ⚠️ Correction 2 — `premerge-assert.sh 3814 <that-sha>` REFUSES. It needs a third argument.
Measured, not predicted:
```
$ bash scripts/flow/premerge-assert.sh 3814 380010a20
PREMERGE: USAGE — the call is wrong (this is NOT a gh/network failure)
usage: premerge-assert.sh <pr-number> <certified-sha> <gate-of-record-summary> [<delta-summary>]
REAL exit=3
```
The gate-of-record summary path is **deliberately required**, not optional — #3465, because verifying the head against a *claimed* certified sha never verified that a certified sha **exists**. That is the #3408 escape (a PR merged on 22 `--lite` passes and not one full gate, because nothing in the merge path ever asked for the block) and the #3616 one (a closer read a *peer lane's* run dir by recency and nearly merged on another PR's verdict).

So step 5 must be:
```bash
bash scripts/flow/premerge-assert.sh 3814 <sha> /tmp/gate-3721-arm.txt
```
This matters beyond pedantry: a lane following the two-argument form gets a refusal it may read as tooling breakage, and the tempting "workaround" is to skip `premerge-assert` and arm anyway — which is precisely the hole the third argument closes.

### Status against your five steps
1. **Bring current + push — DONE, before your message.** Rebased 45 commits onto `9bb2cc571`; all fixes committed and pushed; head `380010a20`, 0 behind, clean tree.
2. **roborev clean at final HEAD — IN FLIGHT (r13).** Findings across rounds: **3 → 2 → 1 → NONE → 1 → 1(Low)**, every one real. The last two rounds found defects on surfaces my own tests did not reach — a BTI point read reporting an undecodable row as **absent** (`Ok(None)`, satisfying neither branch of this issue's AC1), and a `WARN` that fired on clean straddled reads. Both fixed.
3. **The full gate — CANNOT REACH `RESULT: PASS`.** `core-tests` and `work-counters-guard` fail on **#3890** (P0, `OPEN`, **unclaimed, no assignee, no PR** ~18 h after the lead ruled it goes first), in code this session does not own. The coupling is structural: my change converts a silent wrong answer into an error, so those two tests **pass on `main` while dropping a column** and fail here *because the fix works*. Not waivable — the diff is not prose, so `classify-docs-only.sh` refuses that route.
4. **`commit:` == HEAD — I will enforce it**, and it is why I am not gating yet: gating now burns ~1 h to produce a FAIL, and the freeze has to hold from the gate's start to the arm.
5. **Not arming.** `premerge-assert` requires `RESULT: PASS` token-exactly and would refuse. Your own rule — gate last, then arm — is what I am following; the gate simply cannot pass yet.

**Also unmet independently: the launch bar.** Three peer FULL gates were live on this box at my last check (2h48m / 1h46m / 1h33m) and `CQLITE_GATE_MAX_CONCURRENCY=1` is pinned, so a fourth would queue rather than run. Note load average is useless here — I measured `1.06` while all three were running, because a gate in its test phase holds the slot, the disk and the build lock without burning CPU. Measure the gate **count** via `AGENT_GATE_SUMMARY_FILE` in `/proc/<pid>/environ`, never `pgrep -f` (it self-matches and mis-times children).

I will run the gate at `/tmp/gate-3721-arm.txt`, paste **both** blocks, run the three-argument `premerge-assert`, and arm — the moment #3890 lands and the box has a free slot. Nothing is armed and nothing will be until a gate of record reads `RESULT: PASS` + `tree-integrity: PASS` with `commit:` at the head that merges.

pmcfadin:
## Step 2 confirmed: **roborev is CLEAN at the pushed head.** Steps 1–2 of the ordering rule are complete; step 3 remains blocked on #3890.

```
job 84   reviewed-sha 9bb2cc571f48830d2..380010a202843d7e8   (range, at HEAD)
tokens: input=3497181 cached=3352448        (genuine; vacuous baseline is 18.7k/0)
push-assert PASS   census-check PASS   code-free PASS   job-record PASS   sha-assert PASS
review-completed PASS   vacuity-tier1 PASS   vacuity-tier2 PASS
findings: NONE
prompt-content: FAIL (33/33 absent — snapshot delivery)      RESULT: FAIL
```
`RESULT: FAIL` is now attributable to **one** thing: `prompt-content`, i.e. roborev delivered the diff by snapshot rather than inline. That is the condition the absence waiver exists for — no code finding remains.

Findings across rounds, every one real: **3 → 2 → 1 → NONE → 1 → 1 → NONE.** The two that appeared after the first clean round were on surfaces my own tests never reached: a BTI point read reporting an undecodable row as **absent** (which satisfied neither branch of this issue's AC1), and a `WARN` that fired on healthy chunk-straddled reads. Both fixed, plus a sweep of the eight sibling call sites — all measured clean, one named residual recorded.

### Waiver preconditions (#3460): **3 of 4 met**
| # | condition | state |
|---|---|---|
| 1 | pushed | **MET** — head == remote == `380010a20` |
| 2 | conflict-free | **MET** — `mergeable=MERGEABLE` |
| 3 | post-gate | **NOT MET** — #3890 |
| 4 | reviewed at that sha | **MET** — job 84, `findings: NONE` |

### I am deliberately NOT requesting the waiver yet, and this is the point of the rule you sent
The triple is correct *right now* (`base=9bb2cc571…` — the **assert-base**, never the `base:` line; `head=380010a20…`; `job=84`). It becomes void the instant #3890 lands, because clearing that blocker requires a rebase, which changes **both** base and head → `waiver: STALE`, and a human's authorization is spent on code that cannot merge. Three lanes each burned a grant exactly that way on 2026-08-28.

So the failing block prints a valid-looking triple and copying it into a request is simultaneously the obvious action and the wrong one while the head can still move. Order stands: **#3890 merges → rebase → roborev clean at the rebased head → gate `PASS` → *then* request → `--recheck-job` → paste both blocks → 3-arg `premerge-assert` → arm.**

**#3890 status:** `OPEN`, `P0`, `status:ready`, **no assignee, no PR** — roughly 18 hours after the lead ruled it goes first. It is not mine to claim (one issue per session), and I have posted two measured evidence drops on its thread, including a mechanism from this lane that reproduces its exact symptom class. That issue getting picked up is the only thing standing between this PR and an arm.

pmcfadin:
## COORD-REQ:3721-13 — **executed the stated default (A). #3782 is integrated, validated and pushed** at `ac412493a`.

No reply by the next firing, so I took the default I posted rather than idle. Reporting what it cost and what it changed, because two of the outcomes were not what I predicted.

### The premise held: #3721 is not redundant
```
main's row_decoder/row_data.rs   917 lines, STILL swallows at :613 (complex) and :782 (simple)
#3782 touched row_data.rs        0 times
main has Error::ColumnDecode     0 hits        column_decode_error.rs / range_marker_error.rs  absent
```

### MERGE, not rebase — and this was a judgement call worth stating
`git rebase` stopped at my *first* substantive commit with four semantic conflicts. With 46 commits ahead it would re-resolve the same decision up to 46 times, each one an opportunity for a silent wrong resolution in code where wrongness is invisible. **One merge = one resolution**, `--squash` flattens it, and `premerge-assert` checks head sha and gate summary rather than linearity. Your earlier instruction said "rebase **or** merge", so this is in bounds. 9 conflicts, 3 auto-merged.

### I took #3782's design over my own, three times, on their evidence
| site | mine | theirs — adopted |
|---|---|---|
| `partition_driver`, `compaction_stream` | unconditional `?` | `if at_final_chunk` — **measured**: over 42 well-formed corpus tables (10913 rows) the tolerant path fires **614 times, all with `at_final_chunk == false`, zero with `true`**. My version would have hard-failed legitimate rows straddling a chunk boundary. |
| `timestamp_policy`, `row_decoder/compaction` | discriminate *inside* the policy, folding non-decode errors into `Ok(None)` | `Err(e) => Err(e)` — preserve everything. Mine **destroyed information the driver needs**. |

Their rule — **a policy preserves every error; the driver decides tolerance** — is the better architecture and is now this branch's.

### Where they compose, both survive
- **`MarkerOutcome::Stop` stays removed.** #3782 fixed the *row* arm's tolerance and left the **marker** arm returning `flush_and_emitted!(offset, false)` on the final chunk — a successful, truncated partition with the tombstone dropped, i.e. #3721's defect verbatim. Their contract makes finality explicit; removing the variant makes the swallow *unexpressible*. Their call sites migrated to `Unparseable`.
- **`bti_point`**: the extent is now *derived* from `chunk_targeted`, the same finality state the point-read guard uses, so the two cannot disagree.

### ⚠️ The oracle earned its keep — and the regression carried NO conflict marker
My first resolution gated #3721's discrimination behind `extent.is_complete()`. That puts the tolerance decision back **inside** the parse, contradicting #3782's own rule — so a chunk-targeted point read stopped surfacing the decode error, the guard had nothing to remember, and `SSTableReader::get` reported an existing-but-undecodable partition **ABSENT** again (roborev job 80's defect).

`issue_3721_bti_point_read_absence` caught it. **Nothing else in the suite would have**: `bti_point.rs`'s only textual conflict was its import block, so the arms **auto-merged** and the regression arrived without a marker to review. This is the hazard I flagged on #3890's thread, arriving by the route I had not predicted.

Fix is **ordering**: #3782's completeness gate first (a complete buffer makes any failure data loss), then #3721's discrimination for the incomplete case. **Generalisable: when two mechanisms gate the same failure they must read the same state, and the stronger rule must run first — gating one behind the other silently disables it.**

### Validation at `ac412493a`
```
mine   56:  bti_point_read_absence 4 | column_decode_error_surface 21 | range_marker_refusal 18
            3809_row_bound 8 | 3612_cell_path 5
theirs:     issue_3782_corrupt_row_refusal 12/12   <- I did not break the fix I merged
            point_vs_full_differential 13/13 (was 8; #3782 added 5)
corpus:     smoke rc=0, ALL 51 enforced tables | query_semantics_oracle 6/6      fmt clean
```
Three of my tests retargeted to **earlier** refusal layers, none weakened: #3782's `row_size`-vs-available-bytes guard fires before row assembly, so two truncated fixtures are refused at the **row extent** — and naming a column there would attribute the failure to the wrong subject, so the `ColumnDecode` assertion is correctly *inverted* for those two.

### Still blocked, and the blocker got worse
**PR #3987 (#3890) is `DIRTY`** — #3782 conflicted their PR too. My gate still cannot pass until it lands. roborev is running at this head; after #3987 merges I have a recorded 2-file rebase plan including the silent 9th-argument parse-bound trap and a two-oracle post-merge check. Nothing armed.

pmcfadin:
## Authorization request — both markers, one triple (requested in `COORD-REQ:3721-15` on #3721)

Head is final: pushed, `mergeable=MERGEABLE`, reviewed at this exact sha (roborev **job 106**),
gate of record in flight on a frozen tree. Full reasoning is on the issue thread; the values
are here because this is where the authorization has to be posted.

| field | value |
|---|---|
| base | `2346addf2bdb238d90066840b64e8f65f0acd131` |
| head | `8c2373f6be2e2a1f26ea065babbc3bf795cf7a86` |
| job  | `106` |

**1 — `prompt-content` absence.** 33/33 census code paths absent from the prompt (snapshot
delivery). The review was genuine, not vacuous: **input=329126, cached=271104, output=1711**
against a ~18.7k/0 vacuous baseline; every other deterministic key is PASS
(`push-assert`, `census-check`, `code-free`, `job-record`, `sha-assert`, `review-completed`,
`vacuity-tier1`, `vacuity-tier2`).

**2 — findings deferral.** Exactly **1** finding: the index-narrowed window cutting a valid
range-tombstone marker, tracked by **#3998** (OPEN) and already ruled on — I attempted that
fix, it reintroduced #3721's own defect on the SELECT path, and I reverted it. roborev
re-reports a deferred finding every round by design, so `count=1`, `issues=3998`.

Each marker must be its own comment whose **only** nonblank content is the marker line
(`scripts/flow/roborev-review.sh --help` has the form; I am not reproducing it here, because
a quoted copy in a request is what the sole-content rule makes inert). I will then apply both
with `--recheck-job 106`, which enqueues no new review.

I cannot waive my own PR and am not attempting to. Nothing is armed.

pmcfadin:
## Certification at head `446055cc0` — gate of record PASS, both blocks below

Head is FINAL: pushed, `mergeable=MERGEABLE` / `mergeStateStatus=CLEAN`, `behind-main 0`,
gate of record PASS on this exact tree, reviewed at this exact sha (roborev job 114).

### `premerge-assert.sh 3814 446055cc0 <gate-summary>` (3 args, #3465)
```
PREMERGE: OK 446055cc0df4ede8bfa2f2f1c8a8e6c3dd10a082
PREMERGE: ADVISORY BASE-STALENESS: verdict NO-STALENESS-RECOGNISED
PREMERGE: GATE-OF-RECORD commit: 446055c tree-start: 446055cc0df4 tree-integrity: PASS dirty: no summary: /tmp/cqlite-gate-46ut1O/summary.txt
```

### 1 of 2 — AGENT-GATE SUMMARY (gate of record)
```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.bF5Yv0
commit: 446055c branch: issue-3721-complex-column-decode-error-swallowed dirty: no
schemas: 8/8 canonical .cql readable under /data/lanes/lane-3721/test-data/schemas (checkout-relative)
component-set: PASS (37/37 names vs origin/main d43fe8423001782535be0798f5adfcbab4cd1c72; NAMES ONLY — not implementations, and no component is run here) — baseline read via the committed manifest; objects: baseline REUSED from this lane's SHARED store — store TRUSTED, not verified (#3746)
tree-start: 446055cc0df4 dirty: no digest: ae7e54bd5896
tree-end: 446055cc0df4 dirty: no digest: ae7e54bd5896
tree-integrity: PASS
census: 24/37 components AFFIRMED a count; 12 DECLARED-GAP (RECOGNISED); 0 NOT-MEASURED (RECOGNISED); 0 measured-ZERO (RECOGNISED); 1 not-applicable (component did not PASS); 0 no-subject (PASSed; the run had nothing to measure); 0 UNDECLARED; 0 unrecognised; 0 row(s) carry a VACUOUS status. NON-EXHAUSTIVE: the gap set is CURATED, so an unaffirmed component is UNMEASURED, never verified (#3625; the remaining gaps are tracked in #3162).
file-size:         OPT-OUT (0s)  [no-cargo]  {no census: component ended OPT-OUT, so there is no PASS to affirm} — CQLITE_ALLOW_FILE_GROWTH=1 (ratchet NOT enforced); 6 over-threshold file(s) grown — see file-size.log under logs:
fmt:               PASS (5s)  [fmt workspace features=n/a]  {no census — cargo fmt --all --check emits no per-file tally to count}
clippy:            PASS (224s)  [clippy workspace(excl 5) --all-features | clippy cqlite-core --features 33:all-compression,arrow,bench-internals,+30 more | clippy cqlite-cli --features 10:benchmarks,ci_zero_tolerance,cli-helpers,+7 more | clippy cqlite-flight+cqlite-py+cqlite-node --features cqlite-node/write-support]  {no census — cargo clippy emits a per-crate tally only COLD; a warm run prints Finished alone}
roborev-lints:     PASS (68s)  [no-cargo]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
core-tests:        PASS (1126s)  [nextest run cqlite-core --features cli-helpers | test cqlite-core --features cli-helpers]  {verified: 5674 tests passed (across 2 result line(s))}
tombstones-scan:   PASS (45s)  [test cqlite-core --features write-support,cli-helpers,tombstones]  {verified: 2 tests passed (across 1 result line(s))}
scan-offload-guard: PASS (160s)  [test cqlite-core --features cli-helpers,scan-offload-probe x2]  {verified: 15 tests passed (across 8 result line(s))}
work-counters-guard: PASS (93s)  [test cqlite-core --features write-support,cli-helpers,state_machine,work-counters]  {verified: 60 tests passed (across 15 result line(s))}
byte-budget-guard: PASS (14s)  [test cqlite-core --features write-support,cli-helpers,state_machine]  {verified: 21 tests passed (across 5 result line(s))}
arrow-parity-guard: PASS (39s)  [test cqlite-core --features arrow]  {verified: 4 tests passed (across 1 result line(s))}
memory-budget:     PASS (239s)  [test cqlite-core --features cli-helpers,dhat-heap,arrow x3 | test cqlite-flight --features dhat-heap]  {verified: 6 tests passed (across 4 result line(s))}
integration-tests: PASS (63s)  [test cqlite-integration-tests default-features x2]  {verified: 42 tests passed and 25 test binaries built/verified}
format-compat:     PASS (38s)  [test format-compatibility-tests default-features]  {verified: 10 tests passed (across 1 result line(s))}
write-tests:       PASS (155s)  [test cqlite-core --features write-support x3]  {verified: 3859 tests passed (across 3 result line(s))}
cli-tests:         PASS (225s)  [test cqlite-cli default-features | test cqlite-cli --features write-support]  {verified: 764 tests passed (across 48 result line(s))}
compaction-byte-parity: PASS (12s)  [test cqlite-core --features write-support x2]  {verified: 12 tests passed (across 4 result line(s))}
bti-multiclustering: PASS (6s)  [test cqlite-core --features state_machine,cli-helpers x2]  {verified: 19 tests passed (across 3 result line(s))}
query-semantics-oracle: PASS (0s)  [test cqlite-core --features state_machine,cli-helpers]  {verified: 6 tests passed (across 1 result line(s))}
flight-query-semantics-oracle: PASS (52s)  [test cqlite-flight default-features]  {verified: 4 tests passed (across 2 result line(s))}
flight-tests:      PASS (101s)  [test cqlite-flight default-features]  {verified: 407 tests passed (across 2 result line(s))}
legacy-heuristics: PASS (214s)  [build cqlite-core --features legacy-heuristics]  {verified: 3753 tests passed (across 6 result line(s))}
feature-iso-parquet: PASS (133s)  [test cqlite-core --no-default-features --features all-compression,parquet]  {verified: 1 test binaries built/verified}
feature-iso-delta-scan: PASS (61s)  [test cqlite-core --no-default-features --features all-compression,delta-scan]  {verified: 1 test binaries built/verified}
python-bindings:   PASS (376s)  [via maturin: feature set NOT observed]  {verified: 786 pytest tests passed}
node-bindings:     PASS (309s)  [via npm run build (napi): feature set NOT observed]  {verified: 602 jest tests passed}
binding-rust-tests: PASS (120s)  [test cqlite-ffi-common default-features | test cqlite-node --features write-support]  {verified: 123 tests passed (across 5 result line(s))}
parity-report:     PASS (3s)  [run cassandra-parity default-features]  {no census — xtask/report driver prints no machine-readable subject count (#3162)}
binding-unwind-profile: PASS (2s)  [no-cargo]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
tooling-tests:     PASS (2506s)  [test ws0-corpus-gen default-features | + cargo not observable: cargo may run inside ~60 nested test scripts (child processes)]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
minimal-build:     PASS (61s)  [build cqlite-core --no-default-features --features all-compression | test cqlite-core --no-default-features --features all-compression]  {verified: 1 test binaries built/verified}
all-features-check: PASS (195s)  [check cqlite-core --all-features | clippy cqlite-core --all-features]  {no census — cargo check/clippy passes execute no tests; the subject is a feature set, not a count}
smoke:             PASS (154s)  [build cqlite-cli default-features]  {no census — smoke-test-all-tables.sh prints no machine-readable table count (#3162)}
logs: /tmp/agent-gate.bF5Yv0
summary-file: /tmp/cqlite-gate-46ut1O/summary.txt
heartbeat: on file: /tmp/cqlite-gate-46ut1O/summary.txt.heartbeat interval: 20s
launch-nonce: 20260903T132023Z-2931862-280765736
RESULT: PASS
```

### 2 of 2 — ROBOREV REVIEW SUMMARY (job 114)
```
repo: /data/lanes/lane-3721
branch: issue-3721-complex-column-decode-error-swallowed
base: origin/main
head-sha: 446055cc0df4ede8bfa2f2f1c8a8e6c3dd10a082
reviewed-sha: d43fe8423001782535be0798f5adfcbab4cd1c72..446055cc0df4ede8bfa2f2f1c8a8e6c3dd10a082
assert-base: d43fe8423001782535be0798f5adfcbab4cd1c72 (merge-base of origin/main and HEAD; origin/main tip d43fe8423001782535be0798f5adfcbab4cd1c72)
job: 114
model: gpt-5.6-sol
census: 35 files, +6997/-727
tokens: input=778109 cached=695040 output=2463
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: FAIL (35/35 code census paths absent from the prompt)
waiver: STALE (the marker names a different review — base (e6389ab31d81515836a3db0280ebef5961858aa0 != d43fe8423001782535be0798f5adfcbab4cd1c72), head (97eb2b5a6eb5ffe665c76dab5d6c69a74048d7f5 != 446055cc0df4ede8bfa2f2f1c8a8e6c3dd10a082), job (27 != 114) — and a waiver may not outlive the review its authorizer judged; re-request it for this base/head/job (a completed job can be re-decided with --recheck-job <id>, which enqueues nothing))
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: PRESENT (1)
log: /tmp/roborev-3721-r15.log
RESULT: FAIL
```

pmcfadin:
**findings deferral GRANTED (lead) — the marker follows in its own comment.** Your `COORD-REQ:3721-14` disposition is accepted; this completes it now that I could verify the fields myself.

Verified from the job record rather than the request (your earlier ask carried placeholders): job **114**, `git_ref` = `d43fe8423001782535be0798f5adfcbab4cd1c72..446055cc0df4ede8bfa2f2f1c8a8e6c3dd10a082` — head is **exactly** this PR's live head, base equals the independently computed merge-base. `count=1` matches. **#3998 is OPEN.**

Then: `--recheck-job 114` → `premerge-assert` → arm. `findings:` will read **DEFERRED**, never NONE.

pmcfadin:
roborev-defer: findings issues=3998 count=1 base=d43fe8423001782535be0798f5adfcbab4cd1c72 head=446055cc0df4ede8bfa2f2f1c8a8e6c3dd10a082 job=114 reason=Lead-deferred per the accepted COORD-REQ:3721-14 disposition: the extent-aware fix was implemented, measured to reintroduce this issue own defect on all three d9_select_marker_parse_failure cases, and reverted; a conservative false refusal is recoverable where a silent short SELECT is not. Tracked in open issue 3998 with a regression AC pinning those three cases. Authorized 2026-09-03.

pmcfadin:

You reported the deferral applied cleanly (`findings: DEFERRED`) with `prompt-content` as the only remaining non-PASS key. Re-verified just now before granting: this PR's live head is **unchanged** from the head job `114` reviewed, so both authorizations bind to the same review.

That is the full set — nothing further is owed by me. Proceed: `--recheck-job 114` → `premerge-assert` → arm `gh pr merge --auto --squash --delete-branch`. The block will read `findings: DEFERRED` and `prompt-content: WAIVED`; neither is `PASS`, so this is not a clean review and must not be reported as one.

pmcfadin:
roborev-waive: prompt-content-absent base=d43fe8423001782535be0798f5adfcbab4cd1c72 head=446055cc0df4ede8bfa2f2f1c8a8e6c3dd10a082 job=114 reason=Second authorization for this same review, following the findings deferral already granted and applied on job 114. Verified as lead: the job git_ref still ends exactly at this PR live head and its base equals the independently computed merge-base, and the lane reports findings DEFERRED with prompt-content as the sole remaining non-PASS key, which is a roborev delivery artifact rather than a defect. Authorized 2026-09-03.

Issue #3811, opened 2026-09-01
Consumption/bounds contract is not enforced repo-wide in row_decoder: marshal-form and registry-resolved UDT arms return short offsets to bounded callers that discard them


pmcfadin:
Split out of #3631 (structured values degrade to Value::Blob) by an explicit lead bound after FIVE
consecutive roborev rounds landed in the same property family. #3631 fixes the **silent-blob** class
it was filed for; this issue owns the **consumption/bounds contract** that #3631 introduced but could
not enforce repo-wide.

## The property

#3631 added a consumption contract to the typed decode path: a structured decoder must report how
many bytes it consumed, and a bounded caller must require that count to equal the slice length — so
trailing bytes, or a partial 1–3 byte field-length prefix, are a corruption error rather than
silently discarded. Two distinct serialized values must never collapse to the same `Value`.

That contract is now enforced at the sites #3631 reached. **It is not enforced everywhere.**

## Known open instance (the immediate subject)

`cqlite-core/src/storage/sstable/reader/parsing/row_decoder/raw_type_value.rs:794` — the marshal-form
and registry-resolved UDT arms return a potentially SHORT offset, while bounded callers such as
`parse_value_from_raw_bytes` DISCARD it. So a UDT inside a frozen collection still accepts trailing
bytes or a partial field-length prefix despite the contract.

Suggested fix (from the review): at bounded callers, require the returned offset to equal the supplied
slice length; add coverage for BOTH raw UDT arms with trailing-byte and partial-prefix corruption.

## Why this is filed as a class, not one line

The measured history across #3631's five review rounds — each fix correct, the family regenerating one
module further out each time:

| round | findings | where the consumption/depth family landed |
|---|---|---|
| 1 | 3 | `typed_value.rs` — no exhaustion check at all |
| 2 | 3 | `udt.rs` — UDT arms bypass the new check; depth discarded across UDT boundaries |
| 3 | 3 | `udt.rs` — depth reset at frozen nested UDT; nested consumed count discarded |
| 4 | 4 | `udt.rs` — COLUMN-level entries (`parse_udt_value` callers, direct-column `Frozen`/`Udt` arms) |
| 5 | 2 | `raw_type_value.rs` — marshal/registry arms return short offsets to bounded callers |

#3631's round-3 refactor collapsed **five** per-field UDT dispatches into one entry, and its round-4
follow-up completed that at the **column** level — which closed the `udt.rs` family and dropped
[...]

PR #3820, opened 2026-09-01
fix(#3631): decode structured values from the declared type instead of degrading to Value::Blob — instance B, and four fallback arms where the issue named two

### LINKED ISSUE #3631: Structured values silently degrade to Value::Blob while the schema is present: non-frozen map<frozen<udt>,int> keys and collection fields inside a frozen UDT (one class, two instances)
Discovered scope from #3504 (UDT field namespace). **Filed as a class, with two named instances,
deliberately.** Fixing structural-value-degrades-to-blob one shape at a time is what #3504 was filed
to stop: that issue's whole point was that the `_type`/`_keyspace` marker collision was *not one
defect but a class with four sites*, and patching sites individually left the family regenerating.
The two instances below share one cause — a decode path matching a **closed set of primitive types**
and falling back to `Value::Blob` for everything else, while the schema that names the real type is
present and unread.

### Instance A — a NON-frozen `map<frozen<udt>, int>` key decodes to `Value::Blob`

A non-frozen map is multicell, so each key lives in the **cell path**, decoded by
`parse_cell_path_key` (`cqlite-core/src/storage/sstable/reader/parsing/row_decoder/complex_column.rs:1511`,
called from `:717`). It matches text / uuid / int / bigint / date / timestamp and then:

```rust
// Fallback: return as blob
_ => {
    tracing::debug!(
        "Map key type '{}' for column '{}' parsed as blob ({} bytes)",
        type_str, column_name, data.len()
    );
    Ok(Value::Blob(
        crate::storage::sstable::reader::value_borrow::borrow_active(data),
    ))
}
```

(`complex_column.rs:1600-1611`.) A frozen UDT key type hits that arm, so a multicell map's cell-path
key **never becomes a `Value::Udt`** — the column decodes to
`Map([(Blob(<serialized udt bytes>), Integer)])`.

Note this is the live read path despite the type's name: `V5CompressedLegacyParser` is the main row
decoder (used by `scan_stream_windowed.rs`, `data_access/bti.rs`, `fuzz_support.rs`), **not** anything
behind the `legacy-heuristics` feature.

**Consequence: the natural user spelling of a UDT-keyed map loses its structure entirely.**
`CREATE TABLE ... (cm map<frozen<collide>, int>)` — no outer `frozen` — is what a user writes, and it
is precisely the spelling that degrades. Only `frozen<map<frozen<udt>, int>>` (a single value cell,
decoded by `parse_map_with_types`, which resolves the key type through the `UdtRegistry`) produces a
real UDT.

### Instance B — a collection field INSIDE a frozen UDT decodes to `Value::Blob`

`parse_simple_udt_field_value`
(`cqlite-core/src/storage/sstable/reader/parsing/row_decoder/udt.rs:838`) is the field decoder used
for a UDT's fields in a frozen context. Its own doc comment states the limitation, and the fallback is
the same shape:

```rust
/// Limitation: Complex nested types (nested UDTs, nested collections) are returned as blobs.
/// For full UDT support with nested types, use parse_udt_field_value with a reader.
pub(super) fn parse_simple_udt_field_value(data: &[u8], field_type: &CqlType) -> Result<Value> {
    match field_type {
        CqlType::Text | CqlType::Ascii => { ... }
        // ... int / bigint / boolean / float / double / uuid / timestamp / blob ...
        _ => {
            // For complex types (nested UDTs, collections, etc.), return as blob
            // These require SSTableReader for full parsing
            tracing::debug!(
                "UDT field type {:?} in frozen context parsed as blob ({} bytes)",
                field_type, data.len()
            );
            Ok(Value::Blob(
                crate::storage::sstable::reader::value_borrow::borrow_active(data),
            ))
        }
    }
}
```

(`udt.rs:928-938`.) So a UDT declaring a `frozen<map<text,int>>` field surfaces that field to callers
as **bytes**, not a structured value. The nested-**UDT** half of that fallback is partly mitigated
upstream — `raw_type_value.rs:769-851` consults `self.udt_registry` and calls
`parse_nested_udt_from_registry` / `parse_inline_udt_value` when the field type is a UDT — but a
**collection**-typed field has no such branch and always reaches the blob arm.

**How it was found, which is the sharpest evidence that it is a real defect and not a documented
simplification:** while measuring #3504's projection boundary it made a shape **hashable that was
expected to stay unhashable**. `test-data/fixtures/issue_3504/README.md:132-139` records it:

> `stn` **succeeds**, contradicting the obvious prediction that a UDT with a `map`-typed field stays
> unhashable. It does so only because CQLite decodes a collection field inside a frozen UDT as
> `Value::Blob`, so the field arrives as hashable `bytes` rather than a `dict`. That is a **decode
> gap** orthogonal to #3504 (the correct value is `{"a": 1}`).

A decode gap that silently *changes the answer of a totality test* is the kind that stays invisible
until it flips a result the other way.

## Why it matters

**The schema IS present in both cases.** Instance A has the map's key type in the column comparator;
Instance B has the field's declared `CqlType` in hand — it is the `match` scrutinee. Neither path is
guessing for lack of information; both discard authoritative metadata and emit an opaque blob.
`#28 (no-heuristics / remove blob fallbacks)` is explicit that this is the wrong outcome: authoritative
metadata only, schema-aware decoding when schema is present, no silent degradation.

Practical harm, per instance:

- **A** — the natural spelling of a UDT-keyed map returns bytes, so every consumer (CLI JSON, both
  bindings, Arrow/Parquet egress) sees an opaque key. There is no error and no diagnostic above
  `tracing::debug!`.
- **B** — a collection field of a UDT is unreadable as data. It also makes a value **hashable that
  should not be**, which is how a Python-side totality bug (`#3500`) can be masked rather than
  surfaced.

## Reproduction / evidence

Fixtures and pins are already committed by #3504 (UDT field namespace) — Cassandra-5.0.2-written, no
new corpus needed:

- **Instance A**: `test-data/fixtures/issue_3504/` columns `cm` / `tm`
  (`map<frozen<collide>,int>`, non-frozen) vs `fcm` / `ftm` (`frozen<map<frozen<collide>,int>>`).
  Rationale and the measurement are in `test-data/fixtures/issue_3504/README.md:92-104`. Pinned as
  **characterization** (i.e. asserting today's wrong answer, so the fix is visible when it lands) at
  `bindings/python/tests/test_issue_3504_udt_field_namespace.py:331-346`:

  ```python
  assert isinstance(key, bytes), (
      f"{column}: expected the documented Blob cell-path key, "
  ```

- **Instance B**: `test-data/fixtures/issue_3504/` table `udt_hashable_shapes`, row 3, column `stn`
  (`frozen<set<frozen<tuple<frozen<unhashable_fields>,int>>>>`), where `unhashable_fields` declares a
  `frozen<map<text,int>>` field whose correct value is `{"a": 1}`. Pinned as characterization at
  `bindings/python/tests/test_issue_3504_udt_field_namespace.py:644-670`:

  ```python
  # THE GAP: bytes, not {"a": 1}. Asserted as bytes so the pin is exact.
  assert isinstance(udt.fields["m"], bytes), (
  ```

Both pins **invert** when this issue is fixed — that is intentional, and updating them is part of the
work, not collateral damage.

## Oracle-vs-design routing

**Oracle-driven, both instances** ⇒ GitHub issue + pinned parity test; **no OpenSpec change**.
Cassandra / `sstabledump` is the truth for what these bytes decode to, there is no latitude about the
correct value, and the expectations must be derived from the pinned `cassandra-5.0.8` source and the
`sstabledump` goldens — never from CQLite's own prior output. Concretely:

- Instance A's authority is Cassandra's cell-path / `CellPath` + `CompositeType` handling for a
  multicell map with a frozen-UDT key, read at the tag
  (`git show cassandra-5.0.8:src/java/org/apache/cassandra/db/rows/...`), plus the committed
  `nb-1-big-Data.db.jsonl` golden, which already renders the `cm`/`tm` cell paths.
- Instance B's authority is `UserType`/`TupleType` field-value serialization at the same tag, plus the
  golden's rendering of the nested collection.

**Watch the round-trip blind spot** (#3042): a CQLite-written + CQLite-read test is invariant to a
uniform framing error here, so the oracle must be the Cassandra-written fixture bytes — which is what
`test-data/fixtures/issue_3504/` already is.

## Acceptance criteria

- [ ] Instance A: a non-frozen `map<frozen<udt>, int>` decodes its cell-path key to a `Value::Udt`
      with the correct `type_name`, `keyspace` and field values, asserted against the committed
      `test-data/fixtures/issue_3504` `cm` / `tm` columns and the `sstabledump` golden.
- [ ] Instance A: the frozen counterpart (`fcm` / `ftm`) still decodes identically, so the two spellings
      now agree — a differential assert, not two independent ones.
- [ ] Instance B: a `frozen<map<text,int>>` field of a frozen UDT decodes to a `Value::Map`, asserted
      as `{"a": 1}` on `udt_hashable_shapes` row 3's `stn` (and the direct `unhashable_fields` column).
- [ ] Instance B: the same holds for a `list`- and a `set`-typed UDT field, so the fix is at the
      **class** (the fallback arm) and not at the one type the fixture happens to use.
- [ ] Both blob-fallback arms either decode from the declared type or **return an explicit
      `Error`** naming the unsupported type — never a silent `Value::Blob`. A `tracing::debug!` is not
      a diagnostic a caller can see, and #28 (no-heuristics) forbids the silent degradation.
- [ ] The two characterization pins in
      `bindings/python/tests/test_issue_3504_udt_field_namespace.py` (lines ~331-346 and ~644-670) are
      inverted to assert the correct value, and their surrounding comments are rewritten rather than
      left describing a gap that no longer exists.
- [ ] `test-data/fixtures/issue_3504/README.md`'s "Why both a non-frozen and a frozen map column"
      section (lines ~92-104) and its `stn` note (lines ~132-139) are updated in the same change.
- [ ] A parity test — not a CQLite-written round-trip — is the oracle for both instances, per #3042.
- [ ] The `Udt.__hash__` / hashability consequence is re-checked: once `stn`'s `m` field is a `dict`,
      that shape becomes genuinely unhashable again, so the `#3500` boundary table moves and must be
      re-measured, not assumed.

## Related

- `#3500 (Python contains_udt / value_to_hashable_key are not total)` — **downstream** of this issue.
  That one is about the projection raising `TypeError`; this one is about the value never becoming
  structured in the first place. Fixing this issue **moves** #3500's boundary (Instance B's `stn` case
  stops being hashable), so they interact and should be sequenced with that in mind.
- `#3504 (UDT field namespace)` — the delivery that measured both instances and committed the fixture,
  the goldens and the two characterization pins.
- `#28 (no-heuristics / remove heuristics and blob fallbacks)` — the mandate both fallbacks violate;
  this is a concrete, fixture-backed pair of instances of exactly what that issue asked to remove.
- `#3042 (round-trip blind spot)` — why the oracle here must be Cassandra-written bytes.
- `#3497 (type-aware canonicalization)` — records Instance A as a reachability caveat
  (`docs/development/M4_spec.md:775`, the **a-1** row: "A NON-frozen map's UDT key is a **blob** — a
  decode-level gap, not a canonicalization one").

pmcfadin:
Closes #3631.

## What this fixes

#3631 was filed as **one class with two named instances** — a decode path matching a closed set of primitive types and falling back to `Value::Blob` for everything else, while the schema naming the real type is present and unread (#28's no-heuristics mandate forbids exactly this).

**Instance A — non-frozen `map<frozen<udt>,int>` cell-path key — was fixed UPSTREAM, not here** (#3612 / PR #3736). This PR claims no part of it.

**Instance B — a collection/tuple/UDT-typed field inside a frozen UDT — is this PR's subject.** Acceptance criterion 5 says "BOTH blob-fallback arms"; there were **four**:

| arm | disposition |
|---|---|
| `parse_cell_path_key` | fixed upstream by #3612 / PR #3736 |
| `parse_simple_udt_field_value` | fixed here (the issue's instance B) |
| `parse_udt_field_value` | found during implementation — a third arm, named by neither the issue nor the first review |
| `create_empty_value_for_type` | found by review — empty structured fields degraded to blobs; **now deleted**, not patched |

## This PR was REDUCED after #3811 landed ahead of it

This branch sat for ~29 hours while `origin/main` moved 37 commits, and **#3811 merged first** (`b2132fb2c`, PR #3874, "enforce the consumption/bounds contract repo-wide in `row_decoder`") — the very follow-up this PR had deferred its remaining hardening to. The dependency inverted.

So the branch was **reset onto `origin/main` and reduced from 12 commits to 9**, rather than rebased. What was dropped, and why that costs no coverage:

- **The consumption/bounds half is redundant.** Main already enforces exhaustion at six sites. The old branch's `cell_value_complex.rs` hunks were *strictly weaker* than main's — they bound `_consumed` and discarded it — so main's bodies were taken verbatim and only the `depth` argument re-added. The branch lost that exchange on the merits.
- **The duration consumption walk was folded out, not replayed.** #3811 made the callee refuse a trailing byte one layer down, so the walk in `cell_path_key.rs` was dead code that only appeared to do the work.
- **The docs table was discarded in favour of main's.** Main's Table 2 is newer and already carried the `ssu`/#3500 attribution the previous PR body claimed as its own headline. A **column was added** rather than the table replaced — the value of that table is that its successive answers disagree.

Old head `8099fa723` is preserved as tag `backup/issue-3631-pre-rebase-8099fa723`.

## ONE implementation of the consumption rule, per main's recorded ruling

`raw_value/reporting.rs:59-65`, committed by #3811, ruled: "there must be ONE implementation of this rule, not two: when #3820 lands, this function becomes a one-line delegation to it (or is deleted in favour of it) and the two error message classes are unified. Fold in at the same time `cell_path_key.rs`'s now-redundant `duration` consumption check."

Done as specified, in the delete direction: `require_fully_consumed_raw` goes from **16 occurrences on main to 0**, replaced by `typed_value.rs::require_fully_consumed` over `&CqlType`, with main's existing call sites (`cell_value_complex.rs:132,188`, `udt.rs:88,820`, `raw_value.rs:130`, `udt/inline.rs`) retargeted and the error classes unified.

## How instance B is fixed — at the contract, not per site

- **One per-field UDT decode entry** (`typed_value.rs`), so depth propagation and consumption reporting are inherited rather than remembered. `raw_type_value.rs`'s duplicated per-field dispatch is collapsed onto it, and `depth` is threaded through `parse_udt_value` so the nesting limit no longer resets at each frozen-UDT hop.
- **An unexpressible type returns an explicit `Error`, never a silent blob** (criterion 5).
- **`create_empty_value_for_type` is deleted**, both zero-length call sites routed through the one entry under Cassandra's actual empty-value rule — `AbstractType.isEmptyValueMeaningless` (`if (buffer == null || (buffer.remaining()==0 && type.isEmptyValueMeaningless())) return null;`), so an empty fixed-width field is **NULL**, not an empty blob. The implemented rule is documented as *wider* than `isEmptyValueMeaningless` where it is, rather than claiming a match it does not have.

## Oracles

Cassandra-written bytes only, never a CQLite round-trip (#3042). The committed `test-data/fixtures/issue_3504/` corpus is Cassandra-5.0.2-written and its sstabledump golden already renders `"m":{"a":1}` — instance B's correct answer. Expectations derive from pinned `cassandra-5.0.8` source, cited at the assertions.

**Criterion 4's `list`/`set` coverage is DECODER-LEVEL, not fixture-level, and says so**: the committed `unhashable_fields` UDT declares only `label text` + `m frozen<map<text,int>>`, so no fixture has a `list`- or `set`-typed UDT field. Claiming fixture coverage there would be false.

## Criterion 9 — the #3500 boundary, RE-MEASURED on this tree

The criterion required measurement rather than assumption, and measuring corrected both the issue's prediction *and* the previous PR body's rebuttal of it:

| shape | before | after |
|---|---|---|
| `stu`, `mtu` | `list[tuple[Udt(collide), 10]]` / `dict{…: 5}` | unchanged |
| `ssu` | no longer raises (#3500's container change on `main`, **not** this PR) | unchanged |
| `stn` | field `m` = **17 `bytes`** | field `m` = **`{'a': 1}`** |

The issue predicted `stn` would "become genuinely unhashable again, so the #3500 boundary moves". **At the column level it does not** — #3500 renders a UDT-bearing set as a `list`, and a `list` hashes nothing. But the previous flat "it does not" was also incomplete: **the boundary moved one level in.** `hash(stn[0][0])` and `hash(stn[0])` now both raise `unhashable type: 'dict'`, and both are asserted. A shape that *does* reach `value_to_hashable_key` carrying this UDT — a `map` keyed on `frozen<unhashable_fields>`, which no committed fixture declares — now raises where it previously hashed 17 bytes.

## Characterization pins inverted — THREE, not four

Core `to_json`, CLI JSON, and Python. The previous PR body claimed a fourth, Node; **there is no Node pin** — nothing under `bindings/node` references this fixture and the diff touches zero files there. Corrected rather than inherited. Comments were rewritten, not merely flipped: a pin whose prose still describes a closed gap is worse than no pin. The instance-A halves were inverted upstream by #3736.

### Review rounds under this ownership

The branch was adopted uncertified after ~29 hours idle. Three review rounds ran on the reduced diff, each via `scripts/flow/roborev-review.sh --agent codex --model gpt-5.6-sol --repo <abs> --base origin/main`, all genuine (2.1–2.7M input tokens against a ~18.7k vacuous baseline):

|---|---|---|---|
| 6 | 68 | 2 | both fixed |
| 7 | 76 | 1 | fixed |
| 8 | — | see block below | — |

**Round 6 — Medium: a hard-failure regression this PR introduced.** Every `CqlType::Custom` was routed as a UDT, but the marshal parser emits `Custom` for valid types it does not recognize (`DurationType`, `ShortType`, `ByteType`, `CounterColumnType`, `TupleType(...)`), so valid marshal-form UDTs carrying those fields began failing with "nested user-defined type … field list is not available" instead of decoding. Closing a silent-blob gap must not open a hard-failure gap. Every native marshal form is now mapped before `Custom` routes to a UDT, and a **negative** assertion pins that an unknown name no longer produces that message — the misattribution was fixed, not just the failure.

**Round 6 — Low: `inet` width.** A malformed 5-byte `inet` UDT field became a `Value::Inet`. Now gated on Cassandra's actual `0/4/16`, consistent with the rule the project already enforces (`custom_scalar.rs`, `cell_path_key.rs`) rather than a second rule — this PR's thesis is one implementation of a rule.

**Round 7 — Medium: a no-heuristics violation introduced by the round-6 fix.** The consolidated marshal-name table ignored the class **package**, so a third-party `com.acme.Int32Type` silently decoded as Cassandra's native `Int32Type` — exactly what #28 forbids, with wrong-value and spurious-width-error potential. A marshal name's package is now part of its identity: only the bare simple name or the canonical `org.apache.cassandra.db.marshal` package is native, verified against pinned `cassandra-5.0.8` `TypeParser.java:450` (`"org.apache.cassandra.db.marshal." + compareWith`), with a near-miss (`…db.marshalX.`) explicitly rejected so a prefix match cannot pass. One rule at one choke point (`udt/marshal_name.rs`), applied across every site that consumed a simple name — including the structural forms (`TupleType`/`ListType`/`SetType`/`MapType`/`FrozenType`/`ReversedType`/`UserType`), since fixing scalars alone would have been half a fix.

**Found while fixing that, not reported by any review: a PANIC on untrusted bytes.** `extract_inner_parens` collected a `Vec<char>` and then sliced by the CHARACTER index, so `extract_inner_parens("é)")` sliced `s[..1]` inside a 2-byte `é`. Now byte-offset indexed via `char_indices`, with a regression test; a multi-byte type string is an unbalanced-parens refusal, not a crash.

Every new test was verified to **discriminate** — to fail before its fix and pass after (round 6: 13 of 16 marshal cases and 3 of 6 inet cases failed pre-fix). This repo has a documented history of tests whose claim exceeded what they exercised, so presence is not evidence.

The file-size ratchet was satisfied by genuine campsite splits, not by squeezing under it: `udt.rs` went **1671 → 1124** lines with the marshal type-string parser split out, and `typed_value.rs` sits at 704.

### `prompt-content` is WAIVED, not passed — and here is why that is not a vacuous review

This diff exceeds roborev's inline threshold, so the reviewer receives it as a **snapshot file** rather than inline text. Job 68's prompt record contains `diff --git` **zero times** and reads "(Diff too large to include inline) … Read the diff from `…/roborev-snapshot-*/roborev-snapshot-content.diff`". `prompt-content:` therefore fails closed, by design (#3312): a snapshot-delivered diff and a genuinely vacuous review are indistinguishable to the machine, and what separates them is a human plus the token accounting.

The evidence that the reviewer read the snapshot is conclusive rather than circumstantial: **its findings cite `typed_value.rs:609` and `:634`, and the string `typed_value.rs` appears nowhere in the prompt.** A review that received nothing cannot cite a `file:line` it was never shown. Both vacuity tiers PASS and the token accounting is 2.1–2.7M input against a ~18.7k vacuous baseline.

### Rounds 9-11: the gate FAILed once, and both failures were this fix WORKING

The first full gate (`c059bcbd0`) came back `RESULT: FAIL`, 35 of 37 green. Both reds were ours, and neither was a decode defect:

**`cli-tests`** — `issue_1491_json_csv_golden_parity` failed because the case `test_compactionparityudt.udt_nested` carried a **declared gap** `e.home`, documented as *"nested frozen UDT renders as blob hex, not a decoded object"*. It failed with: *"the two sides AGREE at that path now, so the exclusion suppresses nothing and is holding back recovered coverage"*. `Report::stale_skips` is built to fail exactly there. The gap is retired, so `e.home` is now **compared for real** in both JSON and CSV against the Cassandra-written `sstabledump` golden — a **third oracle**, independent of this PR's unit and fixture tests. 207 passed/2 failed → **209 passed, 0 failed**.

**`tooling-tests`** — `scripts/ci/check-skill-flag-tables.sh` pins `ROW_SRC=row_decoder/mod.rs` and reads the row-flag constants from there; this PR's campsite split moved them into the new `row_flags.rs` (so `mod.rs` could gain the `typed_value` declaration without breaching the ratchet). The guard was reading a file that no longer held its subject. Now it resolves the real home **and gained two protections**: it FAILs closed when the constants move out from under every candidate (a discrimination proof, not just a green), and enforces a **per-source floor** so it can never pass on an empty subject set. **PASS, 13 assertions.**

The file-size ratchet was hit twice and honoured twice, never gamed: a real campsite split (`udt.rs` **1671 → 1124**, marshal type-string parser extracted) and two comment-prose tightenings that removed no code and weakened no assertion.

### Gate of record — PASS at the merged head

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.0gaYJx
commit: 688be49 branch: issue-3631-structured-values-blob-degrade dirty: no
component-set: PASS (37/37 names vs origin/main a586468ea)
cpu-budget: wrapper=nice ncpu=16 max-concurrency=1(pinned) cores-per-gate=16
tree-start: 688be49d5b6c dirty: no digest: 6dc62daff971
tree-end:   688be49d5b6c dirty: no digest: 6dc62daff971
tree-integrity: PASS
  ... 37 components, ALL PASS ...
  cli-tests:      PASS (43s)    [test cqlite-cli default-features | test cqlite-cli --features write-support]
  core-tests:     PASS (943s)   [nextest run cqlite-core --features cli-helpers | test cqlite-core --features cli-helpers]
  tooling-tests:  PASS (2887s)  [test ws0-corpus-gen default-features | + cargo not observable]
RESULT: PASS
```

Full block: 37/37 PASS, no SKIPs of record. `commit:` and both `tree-*` values bind the certification to `688be49d5b6c`, the exact tree being merged, with the tree unmutated across the whole ~100-minute run.

**Scope of that claim (#3650):** it proves the diff is unchanged since certification and that a full gate passed on **that exact tree**. It does **not** prove certification against the `main` this will squash onto — `base-staleness.sh` reports `STALE-RECOGNISED`, 10 of 13 commits behind touching the blast radius. `git merge-tree` reports **no conflicts** with current `origin/main`. Read this as "gate of record verified at `688be49d5`", never "certified against main".

### Roborev — clean at the same head

```
repo: /data/lanes/lane-3631
branch: issue-3631-structured-values-blob-degrade
head-sha: 688be49d5b6ccbd67154c3d53eac310804a99774
reviewed-sha: fb889e227dc1db1c6cb3dd613867395c2bd9fecc..688be49d5b6ccbd67154c3d53eac310804a99774
assert-base: fb889e227dc1db1c6cb3dd613867395c2bd9fecc (merge-base of origin/main and HEAD)
job: 93          model: gpt-5.6-sol
census: 34 files, +4687/-1259
tokens: input=912572 cached=820096 output=3033
push-assert: PASS      census-check: PASS     code-free: PASS
job-record: PASS       sha-assert: PASS       review-completed: PASS
prompt-content: WAIVED (29/29 absent — authorized by @pmcfadin for base=fb889e227… head=688be49d5… job=93)
waiver:         GRANTED (author=@pmcfadin, bound to base AND head AND job)
vacuity-tier1: PASS    vacuity-tier2: PASS
findings: NONE         roborev-exit: PASS
RESULT: PASS   (after --recheck-job 93 applied the waiver; enqueued nothing)
```

**`findings: NONE`** from the record's *structured* verdict — nothing deferred, no finding waived past anyone. The sole non-PASS is `prompt-content`, which fails **by design** (#3312) because this diff exceeds roborev's inline threshold and is delivered as a **snapshot file**: job 68's prompt record contains `diff --git` zero times and reads *"(Diff too large to include inline) … Read the diff from `…/roborev-snapshot-*/roborev-snapshot-content.diff`"*. Snapshot delivery and a vacuous review are indistinguishable to the machine, so the verdict is left to a human plus the token accounting.

**The evidence that the reviewer read it is not a token count:** earlier rounds cited `typed_value.rs:609` and `:634`, and the string `typed_value.rs` appears **nowhere in the prompt**. A review handed nothing cannot cite a `file:line` it was never shown — and those findings were correct, and fixed. Review history under this ownership: **round 6 (job 68) 2 findings → round 7 (job 76) 1 → round 8 (job 86) NONE → round 11 (job 93) NONE**, all via `scripts/flow/roborev-review.sh --agent codex --model gpt-5.6-sol --repo <abs> --base origin/main`, 681k–2.7M input tokens per round against a ~18.7k vacuous baseline.

**That key was cleared by the owner's waiver** (granted 2026-09-03T20:42Z, bound to base+head+job, marker as the sole content of its own comment per #3312) and applied with `--recheck-job 93`, which enqueues nothing. The block reads `prompt-content: WAIVED` — **never `PASS`** — so this is a certified review with an excused delivery artifact, not a clean one, and must not be reported as clean.


pmcfadin:
🔐 **WAIVER REQUEST (worker — I cannot grant this, and I am not attempting to): `prompt-content` absence on roborev job 86. This is the only key standing between this PR and a clean review of record.**

@pmcfadin — one request, per #3312's rule that a worker may REQUEST and only the owner or coordination lead may GRANT.

## The verdict I am asking you to excuse

Round 8, at frozen head `c059bcbd0`, is **clean on every substantive key**:

```
findings:         NONE          <- from the record's STRUCTURED verdict, not prose
roborev-exit:     PASS
review-completed: PASS
job-record:       PASS
sha-assert:       PASS
push-assert:      PASS
census-check:     PASS
code-free:        PASS
vacuity-tier1:    PASS
vacuity-tier2:    PASS
prompt-content:   FAIL (22/22 code census paths absent from the prompt)   <- the ONLY non-PASS
RESULT:           FAIL
```

## Why the absence is delivery, not vacuity — evidence, not assurance

This diff (25 files, +4469/-1194) exceeds roborev's inline threshold, so the reviewer is handed a **snapshot file** instead of inline diff text. Job 68's own prompt record contains `diff --git` **zero times** and reads verbatim:

> (Diff too large to include inline) … Read the diff from: `/data/lanes/lane-3631/.roborev/roborev-snapshot-1928299965/roborev-snapshot-content.diff`

`prompt-content:` therefore fails closed exactly as designed — a snapshot-delivered diff and a genuinely vacuous review are **identical to the machine**, which is the accepted cost CLAUDE.md records for this mechanism, with the distinguishing judgement left to a human plus the token accounting.

**The conclusive part, which is not a token argument:** the earlier rounds' findings cite `typed_value.rs:609` and `typed_value.rs:634`, and the string `typed_value.rs` appears **nowhere in the prompt**. A review that received nothing cannot cite a `file:line` it was never shown. Those findings were also *correct* — one was a hard-failure regression, one a no-heuristics package hole — and I fixed both.

## Token accounting, as the protocol requires you to check

| round | job | input | cached | output | findings |
|---|---|---|---|---|---|
| 6 | 68 | 2,705,870 | 2,565,120 | 12,311 | 2 (both fixed) |
| 7 | 76 | 2,144,503 | 2,029,056 | 7,336 | 1 (fixed) |
| 8 | **86** | **681,590** | **616,960** | **2,775** | **NONE** |

Vacuous baseline for comparison is ~18.7k input / 0 cached / 53–56 output. Round 8's input is lower than 6 and 7 and I want to flag that rather than hide it: the output (2,775) is small because a clean review is short, and 617k of 682k was cached — consistent with a real review over a mostly-unchanged diff. It is ~36x the vacuous input floor and both vacuity tiers PASS.

## The scope to bind, if you grant it

```
base = fb889e227dc1db1c6cb3dd613867395c2bd9fecc     (the MERGE-BASE, from assert-base:)
head = c059bcbd0e13ab19b2b938386c63386c8dd6f1c5
job  = 86
```

Copy `base` from that value, **not** from the block's `base:` line — the waiver binds the merge-base (#3392). The exact marker form is in `bash scripts/flow/roborev-review.sh --help`; I am deliberately **not** reproducing it here, because a request comment carrying a fillable marker beside a live base/head/job is how the artifact becomes the credential (#3312's own second instance — a diagnostic that described the escape hatch became it). It must be the **sole nonblank content** of a top-level PR comment; put any commentary in a separate comment.

I will then apply it with `--recheck-job 86`, which enqueues nothing and re-decides that job.

## What is NOT being asked

This excuses the **absence verdict only**. It confers no authority over `findings:` — which reads `NONE` from the structured verdict anyway, so nothing is being deferred and no finding is being waived past you. The tree is **frozen** at `c059bcbd0` and I will not commit again; the full gate of record is running now in its own cgroup, and I will post its `AGENT-GATE SUMMARY` here. If you would rather not grant this, the alternative is real but expensive: split the PR until each part's diff fits inline. Say so and I will do that instead.

_(worker request ID `w3631-r3` — RESPONSE REQUIRED. Cron re-check armed; I will not re-ask.)_

pmcfadin:
🔐 **WAIVER REQUEST (worker — I cannot grant this) — SUPERSEDES my earlier request, which went stale when the head moved. `prompt-content` absence on roborev job 93.**

@pmcfadin — one request, replacing `w3631-r3`. That one bound head `c059bcbd0`/job 86; the gate of record then FAILed, I fixed the two causes, and the head moved — so that request became unusable by construction (the waiver binds base+head+job, #3626). This is the same ask against the current head. Nothing was answered in between, so this is not a re-ask of an open question.

## What happened since, in brief

The full gate at `c059bcbd0` FAILed 2 of 37 components. **Both were ours and both were the fix working:**

1. `cli-tests` — `issue_1491_json_csv_golden_parity` failed because the **declared gap `e.home`** ("nested frozen UDT renders as blob hex, not a decoded object") now **suppresses nothing: the two sides AGREE**. `Report::stale_skips` is built to fail exactly there, and removing the exclusion is the only sound response. That gap is now retired, so a nested frozen UDT is compared **for real** against the Cassandra-written `sstabledump` goldens in both JSON and CSV — a **third oracle**, independent of this PR's unit and fixture tests. 207/2 → **209 passed, 0 failed**.
2. `tooling-tests` — `check-skill-flag-tables.sh` pins `ROW_SRC=row_decoder/mod.rs`, and our campsite split moved the `ROW_HAS_*` constants into the new `row_flags.rs` (so `mod.rs` could gain `typed_value` without breaching the ratchet). The guard was reading a file that no longer held its subject. Now resolved properly, and it gained two protections I required: it **FAILs closed when the constants move out from under every candidate** (discrimination proof) and enforces a **per-source floor** so it can never green on an empty subject set. **PASS, 13 assertions.**

I verified both by running the targets myself, not from a subagent's report. No decode logic changed for either.

## The verdict I am asking you to excuse — job 93, at head `688be49d5`

```
findings:         NONE          <- structured verdict; nothing deferred, nothing waived past you
roborev-exit:     PASS
review-completed: PASS
job-record:       PASS
sha-assert:       PASS
push-assert:      PASS
census-check:     PASS
code-free:        PASS
vacuity-tier1:    PASS
vacuity-tier2:    PASS
prompt-content:   FAIL (29/29 code census paths absent from the prompt)   <- the ONLY non-PASS
RESULT:           FAIL
```

## Why the absence is delivery, not vacuity

The diff (34 files, +4687/-1259) exceeds roborev's inline threshold, so the reviewer receives a **snapshot file**, not inline text. Job 68's prompt record contains `diff --git` **zero times** and reads "(Diff too large to include inline) … Read the diff from `…/roborev-snapshot-*/roborev-snapshot-content.diff`". `prompt-content:` fails closed by design (#3312): snapshot delivery and a vacuous review are identical to the machine, and the distinguishing judgement is deliberately left to a human plus the token accounting.

**The conclusive evidence is not a token count.** Earlier rounds' findings cited `typed_value.rs:609` and `:634` — and the string `typed_value.rs` appears **nowhere in the prompt**. A review handed nothing cannot cite a `file:line` it was never shown. Those findings were also correct and I fixed them: a hard-failure regression on marshal-form UDTs, and a no-heuristics package hole (`com.acme.Int32Type` decoding as Cassandra's native type).

## Token accounting, which the protocol asks you to check

| round | job | head | input | cached | output | findings |
|---|---|---|---|---|---|---|
| 6 | 68 | c059bcb^ | 2,705,870 | 2,565,120 | 12,311 | 2 (both fixed) |
| 7 | 76 | 3222ec4 | 2,144,503 | 2,029,056 | 7,336 | 1 (fixed) |
| 8 | 86 | c059bcb | 681,590 | 616,960 | 2,775 | NONE |
| 11 | **93** | **688be49** | **912,572** | **820,096** | **3,033** | **NONE** |

Vacuous baseline is ~18.7k input / 0 cached / 53–56 output. Round 11 is ~49x that input floor with 820k cached; the small output is what a clean review looks like.

## The scope to bind

```
base = fb889e227dc1db1c6cb3dd613867395c2bd9fecc     (the MERGE-BASE, from assert-base:)
head = 688be49d5b6ccbd67154c3d53eac310804a99774
job  = 93
```

Take `base` from that value, **not** from the block's `base:` line — the waiver binds the merge-base (#3392). The marker form is in `bash scripts/flow/roborev-review.sh --help`; I am deliberately **not** reproducing it here, because a request carrying a fillable marker beside a live base/head/job is how #3312's own second instance happened. It must be the **sole nonblank content** of a top-level PR comment.

**One thing you should know before granting:** on this fleet the worker, the closer and you all post through the **same `pmcfadin` login**, which is the entire `ROBOREV_WAIVER_AUTHORS` allowlist. So a marker-only comment from *this lane* would satisfy the allowlist and self-grant. That is why I have never posted one, and it is worth knowing that the allowlist is not what is protecting this gate — the audit trail and your reading of it are.

## State

Gate of record is **re-running now** at `688be49d5` in its own cgroup (summary `/tmp/gate-3631-arm2.txt` — a distinct path from the failed run's, so neither can clobber the other, #2874). Tree is **frozen**; I will not commit again. If it PASSes and you grant this, the remaining steps are `--recheck-job 93` → `premerge-assert` → `gh pr merge --auto`. If you would rather not grant it, the honest alternative is splitting the PR until each part's diff fits inline — say so and I will.

_(worker request ID `w3631-r4` — RESPONSE REQUIRED. Supersedes `w3631-r3`. Cron re-check armed; I will not re-ask.)_

pmcfadin:
✅ **GATE OF RECORD: `RESULT: PASS` — 37/37 components, at the exact head being merged. `premerge-assert` accepts the binding. ONE thing now blocks the merge, and it is yours: the `prompt-content` waiver (`w3631-r4`).**

## The verdict, and why it is a real one

```
run-id: /tmp/agent-gate.0gaYJx
commit: 688be49 branch: issue-3631-structured-values-blob-degrade dirty: no
component-set: PASS (37/37 names vs origin/main a586468ea)
cpu-budget: max-concurrency=1(pinned)
tree-start: 688be49d5b6c  digest: 6dc62daff971
tree-end:   688be49d5b6c  digest: 6dc62daff971
tree-integrity: PASS
RESULT: PASS
```

`commit:` equals HEAD, `tree-start` equals `tree-end` across a ~100-minute run, `dirty: no`, and the `run-id` is this lane's — not a peer's summary picked up by recency from shared `/tmp`, which is the #3616 trap. `max-concurrency=1(pinned)` confirms the #3414 pin was **in effect**, not merely installed. The tree was frozen from launch to verdict; I made no commit during the run, which is the `commit:`-predates-HEAD failure measured across 15 lanes.

`premerge-assert.sh 3820 688be49d5… /tmp/gate-3631-arm2.txt` → **`PREMERGE: OK`**, exit 0.

## The first gate FAILed, and both failures were this fix working

Worth recording, because it is the strongest evidence in the PR:

- **`cli-tests`** — the declared gap `e.home` ("nested frozen UDT renders as blob hex") failed with *"the two sides AGREE at that path now, so the exclusion suppresses nothing"*. `Report::stale_skips` is built to fail there. Retiring it means a nested frozen UDT is now compared **for real** against the Cassandra-written `sstabledump` goldens in JSON **and** CSV — a **third oracle**, independent of this PR's own unit and fixture tests. 207/2 → **209/0**.
- **`tooling-tests`** — `check-skill-flag-tables.sh` pins `ROW_SRC=row_decoder/mod.rs`; the campsite split moved the row-flag constants into `row_flags.rs`, so the guard read a file that no longer held its subject. Fixed, and it gained a **discrimination proof** (FAILs closed when the constants move out from under every candidate) plus a **per-source floor** so it cannot green on an empty subject set. **PASS, 13 assertions**, 2887s.

Neither touched decode logic. The ratchet was hit twice and honoured twice — one real split (`udt.rs` 1671 → 1124), two comment-prose tightenings that removed no code and weakened no assertion.

## What this does NOT claim (#3650)

`premerge-assert` prints it itself: this proves a full gate passed on **this tree**, not that the tree was certified against the `main` it will squash onto. `base-staleness.sh` → **`STALE-RECOGNISED`, 10 of 13 commits behind touch the blast radius**; `git merge-tree` reports **no conflicts** with current `origin/main`. So: *"gate of record verified at `688be49d5`"* — never *"certified against main"*.

## The one blocker

Roborev job 93 at this same head: `findings: NONE` (structured verdict), `roborev-exit: PASS`, `review-completed`/`sha-assert`/`push-assert`/`census-check`/`code-free`/`job-record`/`vacuity-tier1`/`vacuity-tier2` all PASS. The **only** non-PASS is `prompt-content: FAIL (29/29 absent)` — snapshot delivery, because the diff exceeds roborev's inline threshold. Both blocks are now in the PR body.

@pmcfadin — `w3631-r4` binds base `fb889e227…` (the **merge-base**, from `assert-base:`), head `688be49d5…`, job **93**. I will apply it with `--recheck-job 93` and expect `prompt-content: WAIVED` + `RESULT: PASS`, then arm `--auto`.

**I am not arming on a `RESULT: FAIL` roborev block, and I cannot grant this myself** — worth repeating the thing I disclosed in the request: `ROBOREV_WAIVER_AUTHORS` is exactly `pmcfadin`, and this lane's own comments post as `pmcfadin`, so a marker-only comment from here would satisfy the allowlist and self-authorize. That is why I have never posted one. The allowlist is not what is protecting this gate; you reading it is.

C (intent audit vs this issue's acceptance criteria) is running now and does not need the waiver — I will post its verdict when it lands. If it BLOCKs, that is a second blocker and I will say so plainly rather than merge around it.

_(worker `w3631-r4` still open — RESPONSE REQUIRED. No new question; this is the status the request was waiting to be able to report.)_

pmcfadin:
🔎 **C (intent audit) — verified by direct source measurement. Verdict: PASS on all nine acceptance criteria. Including the one I went looking to break.**

I audited criterion 5 adversarially rather than accepting this PR's account of itself, found a candidate violation, and then disproved it. Recording both halves, because the negative result is the useful part.

## The candidate I found

`raw_type_value.rs` has three non-test `Value::Blob` constructions at the certified head:

- `:454` — type string is literally `"blob"`. A declared blob decoding to a blob. Not a finding.
- `:936`, `:967` — both labelled **`"unknown type"`**, logging *"Frozen element '{}': unknown type '{}', parsing as blob"* and the `no UDT registry available` variant. That is exactly the shape criterion 5 forbids — **if** a structured declared type can reach it.

## Why it is NOT reachable for the class this issue names

The enclosing `match` orders its arms:

```
type_str if starts_with("tuple<")   :564
type_str if starts_with("frozen<")  :588
type_str if starts_with("list<")    :597
type_str if starts_with("set<")     :610
type_str if starts_with("map<")     :623
_ => { … }                          :804   <- the "unknown type" blob arms are inside this
```

Every structured declared type matches its own guard **before** the catch-all, so `list`/`set`/`map`/`tuple`/`frozen` cannot reach `:936`/`:967`. What does reach the catch-all is a bare type NAME: resolved as a UDT via the registry (which delegates to `parse_simple_udt_field_value_at`, the one `typed_value` entry — the collapse this PR performed), or unresolvable, which is the **metadata-absent** case and outside the issue's own premise, *"while the schema is present"*. Those arms also emit a warning, so they are not silent.

**It is also not the residual the original PR deferred to #3811.** That deferral was `raw_type_value.rs:794` — *"the marshal-form and registry-resolved UDT arms still return a potentially short offset"* — a **consumption** property, not a blob fallback. Same file, different defect. #3811 has since merged and this PR unifies with its enforcement rather than duplicating it.

**Declared residual, not a criterion-5 violation:** an unresolvable *bare* type name still becomes a warned blob rather than an explicit `Error`. Whether that should hard-error is a separate design question about metadata-absent decoding; it is not what #3631 was filed to close, and nothing in the criteria demands it.

## Per-criterion verdicts

| # | criterion | verdict | evidence |
|---|---|---|---|
| 1, 2 | instance A: cell-path key → `Value::Udt`, frozen counterpart agrees | **satisfied UPSTREAM** (#3612/PR #3736) | `complex_column/cell_path_key.rs:378`; differential pin `test_issue_3504_udt_field_namespace.py:325-352`. This PR claims no credit. |
| 3 | instance B: `frozen<map<text,int>>` field of a frozen UDT → `Value::Map` | **satisfied** | `cqlite-core/tests/issue_3631_structured_values_not_blobs.rs` against the Cassandra-5.0.2-written `issue_3504` corpus |
| 4 | same for `list`- and `set`-typed UDT fields | **satisfied (partial, honestly scoped)** | Decoder-level only, with Cassandra-derived expected bytes. Verified the constraint is real: the committed `unhashable_fields` UDT declares only `label text` + `m frozen<map<text,int>>`, so **no fixture in the corpus has a list- or set-typed UDT field**. Claiming fixture coverage would have been false; the PR says so instead. |
| 5 | both arms decode-from-declared-type or explicit `Error`, never a silent blob | **satisfied** | Four arms found where the issue named two. Adversarial enumeration above: no structured declared type reaches a blob fallback across `udt.rs`, `typed_value.rs`, `typed_value/scalar_rules.rs`, `raw_type_value.rs`, `cell_value_complex.rs`, `udt/type_string.rs`, `udt/marshal_name.rs`, `udt/inline.rs` |
| 6 | characterization pins inverted | **satisfied, count CORRECTED** | THREE sets (core `to_json`, CLI JSON, Python), not four — **no Node pin exists** and the diff touches 0 files under `bindings/node`. Prose rewritten, not just assertions flipped. |
| 7 | README + schema comments updated | **satisfied** | Added a column to main's newer Table 2 rather than replacing it — main's version already carried the corrected `ssu`/#3500 attribution |
| 8 | a Cassandra-written oracle, never a CQLite round-trip (#3042) | **satisfied, and strengthened during the gate** | Cassandra-written corpus + sstabledump goldens. The `e.home` declared gap RETIRED itself, so the path is now genuinely compared, not merely un-excluded — measured `209 passed, 0 failed` (was 207/2) with no gap in the census |
| 9 | #3500 `Udt.__hash__` boundary RE-MEASURED | **satisfied by measurement** | Falsified the issue's own prediction AND sharpened the PR's first rebuttal: at COLUMN level `stn` still does not raise (a UDT-bearing set renders as a `list`, which hashes nothing), but the boundary moved **one level in** — `hash(stn[0][0])` and `hash(stn[0])` now raise `unhashable type: 'dict'`, both asserted |

## Out-of-scope content in the diff

None that should be split. The two gate-driven edits (`issue_1491_json_csv_golden_parity.rs`, `scripts/ci/check-skill-flag-tables.sh`) are direct consequences of this change — a gap this fix closed, and a guard whose source pointer this fix's campsite split moved — so they belong here, not in a follow-up.

## Process disclosure

I spawned `spec-auditor` for this and it went idle three times without delivering a report, as did three other subagents this session. So **C was performed by me, by direct source measurement, with the evidence cited above** rather than by a specialist's sign-off. That is a real deviation from the pipeline's intent and I am flagging it rather than presenting a subagent verdict I never received. Every claim above is checkable at `path:line` on the certified head.

## Standing

Gate of record **PASS** (37/37, bound to `688be49d5`), `premerge-assert` **OK**, roborev `findings: NONE`, C **PASS**. The only remaining blocker is the `prompt-content` absence waiver (`w3631-r4`), which only @pmcfadin can grant.

pmcfadin:
**prompt-content absence waiver GRANTED (lead). The marker follows in its own comment.**

Verified independently before granting, not taken from the request:

| check | value |
|---|---|
| roborev job | `93` |
| job `git_ref` head | `688be49d5b6ccbd67154c3d53eac310804a99774` — **exactly** this PR's live head |
| job `git_ref` base | `fb889e227dc1db1c6cb3dd613867395c2bd9fecc` — equals `git merge-base origin/main HEAD` computed separately |
| job verdict | `P` (findings NONE) |

Because the reviewer verdict is `P`, `prompt-content` absence is the only non-PASS key available to the wrapper, so the absence is a delivery artifact (snapshot / oversize inline) and not a defect. The waiver excuses **that verdict only** — it confers no authority over `findings:`, and it is bound to base AND head AND job, so a push, a different base or a re-run each need a fresh authorization.

Proceed: `--recheck-job 93` (enqueues nothing, so it cannot stale itself) → `premerge-assert.sh 3820 688be49d5b6ccbd67154c3d53eac310804a99774 <your gate summary file>` → arm `gh pr merge --auto --squash --delete-branch`. Do not re-gate after arming. Note the block will read `prompt-content: WAIVED`, never `PASS` — this is not a clean review and must not be reported as one.

pmcfadin:
roborev-waive: prompt-content-absent base=fb889e227dc1db1c6cb3dd613867395c2bd9fecc head=688be49d5b6ccbd67154c3d53eac310804a99774 job=93 reason=Verified as lead: roborev job 93 git_ref ends exactly at this PR live head and its base equals the independently computed merge-base with origin/main; the job verdict is P with findings NONE, so prompt-content absence is the only non-PASS key available and it reflects roborev snapshot or oversize delivery rather than any defect. Authorized 2026-09-03.

PR #3840, opened 2026-09-01
meas(#3445): bound the inlined VInt-decode share of scan on-CPU — 1.70%, verdict KILL

### LINKED ISSUE #3445: meas: bound the inlined vint-decode share of scan on-CPU (perf annotate the parse hot loop)
Parent: epic #2817. Siblings: #3248 (encode attribution, ranked ABOVE this), #3287 (residual split).

## Why

Cassandra devs report vint decode as a known hot spot. For CQLite the named vint symbol measured only 0.74% of on-CPU (#3027), but that number is a FLOOR, not the true share: `decode_unsigned` is `#[inline]`, so its cycles are attributed to callers, chiefly `parse_row_data_with_offset_impl` (9.6%). The structural cost of vints, a serial data dependency where field N+1's offset depends on decoding field N, would show up as ILP stalls inside that 9.6% and is currently unmeasurable from function-level attribution alone.

The decoder itself is already the good shape (#1638 J4: `leading_ones` + one `u64::from_be_bytes` load, no per-byte loop), so any remaining lever is format-structural, not implementation quality. This issue only bounds the cost; it funds nothing.

## Method (measurement only, no production code)

Two routes, both reported:

1. **`perf annotate` (cycle-level)** on the bare-scan hot loop (WS0 harness or `benches/read.rs`), attributing cycles to the inlined vint decode instruction regions inside the parse functions. Identify regions via disassembly against the known codegen of `decode_unsigned`/`decode_signed`.
2. **Probe build cross-check**: an `#[inline(never)]` build to make the symbol visible. Report it as a cross-check only, with the caveat stated that the probe build changes codegen and the annotate route is primary.

Same-corpus, warm, counters at 100.00% `pct_running`, >= 3 reps, per the WS0 validity rules. Refuse and do not publish any rep that fails validity (#3299 AC5 pattern).

## Acceptance criteria

1. A bounded number: vint decode share of scan on-CPU, with both routes reported and their disagreement quantified.
2. A stall attribution statement: whether the serial-dependency cost is visible (frontend/backend stall cycles concentrated at the vint dependency chain) or below measurement resolution.
3. A fund/kill verdict for any vint-targeted lever, pre-registered: if vint share (annotate route) < 3% of scan on-CPU, the verdict is KILL and no vint lever may be filed citing this issue.
4. Findings recorded in `docs/reports/`, mission doc untouched unless a published number moves.

## Rank

Deliberately below #3248: everything measured so far (#3299, #3027) says the actual gap is in the Flight path, where vints do not exist. This exists to answer the community question with a number instead of a guess, and to close the inlining blind spot in #3027's attribution.

pmcfadin:
## What this is

**Measurement only — no production code changed.** Every path in this diff is under `docs/`;
`cqlite-core`/`cqlite-cli` are untouched, `0` lines of `Cargo.toml` changed, and no
measurement-only feature flag was added anywhere (a probe-only flag would be a feature nothing
executes, #1699/#3522). The two probe builds ship as **unapplied patch artifacts**
(`ac1/inline-never-probe.patch`, `ac1/vint-width-probe.patch`), verified unapplied in-tree.

Bounds the share of bare-scan on-CPU spent in the **inlined** VInt decoder. #3027 measured a named
VInt symbol at 0.74% and correctly called it a **floor**: `decode_unsigned` is `#[inline]`, so its
cycles book to callers. In the measured binary `nm | grep -c decode_unsigned` = **0** — the decoder
has no symbol at all, so the blind spot was real, not a tooling artifact.

## Result

| AC | Result |
|---|---|
| **AC1** Route 1 (`perf annotate` / DWARF inline chain, **primary**) | **1.7027%** total basis (1.6621–1.7960, sd 0.052, n=6) · **2.9779%** in-binary |
| **AC1** Route 2 (`#[inline(never)]` probe, cross-check) | **2.2624%** total · 3.9763% in-binary |
| **AC1** disagreement, quantified | **+0.5597 pp, ratio 1.329×** (1.335× in-binary — **stable across bases**, so it is a property of the attribute, not the denominator) |
| **AC2** stall attribution | **NOT CONFIRMED — at the scan average, 1.061×.** Neither concentrated nor anti-concentrated |
| **AC3** vs pre-registered 3% | **KILL** on the pre-registered total basis (margin 1.30 pp). **MARGINAL** in-binary |
| **AC4** mission doc | Untouched; it carries zero VInt claims, so no published number moves |

Attribution is by **full DWARF inline chain**, not source-file filtering — the decoder's
`leading_ones()` inlines `<u8>::leading_ones`, whose instructions carry an innermost line of
`core/src/num/uint_macros.rs:201`, so a file filter would silently drop them as "core, not vint".
Corroborated: 133/138 disassembly fingerprint anchors fall inside DWARF-derived VInt ranges (the
other 5 are *different* decoders — BTI, repair_metadata); on the probe binary, where the symbol is
out-of-line and no inline reasoning is needed, this pipeline and perf's own attribution agree to
**0.01 pp** across 3 reps.

## ⚠️ Open owner question — `REQ-3445-DENOM` (does not block reading the report)

**Which denominator does the pre-registered 3% govern?** "Share of scan on-CPU" is measured here
against **all** cycles in the window (libc + kernel included) — the issue's own wording. **42.8% of
those cycles are outside the measured binary**, so the same numerator is 1.75× larger in-binary:

- **Total basis (pre-registered, primary): 1.70% → KILL by 1.30 pp, robust.**
- **In-binary basis: 2.98% mean → marginal.** Worst rep **3.1259%**, idle extrapolation **3.0068%**
  and the whole probe route **3.9763%** are all **above** the line.

Both are published so a ruling needs **no re-measurement**. The report presents total as primary
per the issue's wording; re-choosing a denominator after seeing that it moves the verdict is what
pre-registration exists to prevent.

**Both bases UNDERCOUNT VInt**: the multi-byte path (44.4% of decodes) issues a dynamic-length
`call memcpy@GLIBC` whose callee cycles execute in libc. Loose ceiling **8.89%** of scan on-CPU;
exact split **UNMEASURED**. So 1.70% is itself a floor, as #3027's 0.74% was one level down.

## Two published numbers of ours were WRONG and are corrected here

1. **AC2's sign.** First version claimed 0.806× "measurably anti-concentrated" — VInt *less*
   stall-bound than average. That divided two shares whose **denominators differ** (out-of-binary is
   42.83% in cycles reps but **56.47%** in stall reps, because malloc/libc/kernel stall more than the
   binary does). Like-for-like: **1.061×, at parity.** So #3027's ILP-stall hypothesis is **not
   confirmed**, not *falsified*. The retraction is recorded in the report rather than quietly edited.
2. **"Does not depend on any judgement call in the method"** — deleted. The denominator is the
   largest judgement call in it.

## Declared limitations (not worked around)

- **Quiescence is PROSPECTIVE.** No load data was captured *during* any published rep; the
  across-window sampler and its refusal postdate them. On the endpoint data **every published rep
  would be refused** by the new bound. Stated rather than reconstructed.
  - Basis for trusting the verdict anyway is **measured, not argued**: the published set spans a
    **5× load range**, making it its own load-sensitivity experiment — cycle-share slope
    **−0.0009 pp per unit load** (slightly *negative*: the share is a ratio on a pinned core, so
    contention slows numerator and denominator together), 0.1338 pp spread over that range, idle
    extrapolation 1.7087% vs 1.7027% measured. **Flipping the verdict needs ~10× the entire observed
    load-induced variation.**
  - No quiet rep was ever obtained (box reached loadavg **102.85** with **43** concurrent
    `agent-gate` processes; a peer's `/tmp` cleanup deleted the first waiter). **None is reported and
    nothing was re-rolled.**
- **Resolution limit:** no PEBS and no `topdown.slots` on this host, and
  `cycle_activity.stalls_total` cannot separate *hidden* from *absent* latency; stall-side spread is
  0.83 pp. Hence AC2 is "NOT CONFIRMED" — never "no dependency exists".
- Refusals recorded as observations, never dropped: 2 smoke reps (one with a wrong PIE rebase — the
  reason the self-check exists), and a width-probe run that emitted **no histogram**, logged as a
  failed observation rather than "no multi-byte VInts found". An absent measurement is not a zero.

## Review

Five blockers from code review, all fixed; the reviewer's arithmetic on the AC2 sign was
independently reproduced before anything was changed. The three fail-open harness paths are closed
with a committed `raw/refusal-demo/` showing the refusal actually firing — the original defect was
that a *recorded* refusal did not refuse (script exited 0 regardless). B5's no-DWARF-chain
undercount is now **measured at 0.0000%**.

Two rebasing bugs were caught pre-publication by a self-check (`PERF_RECORD_MMAP2` carries a
mapping's *file offset*, not its segment vaddr; selecting a LOAD segment by offset containment alone
picks the read-only segment). Both had produced complete, confident, **wrong** tables.

Closes #3445.


pmcfadin:
## 🔶 Deferral REQUEST — `roborev-defer: findings`, 3 findings, job 33

**I am requesting, not granting.** A worker may never authorize its own PR, and `ROBOREV_WAIVER_AUTHORS` does not include me — correctly. This needs the owner or coordination lead.

### Token accounting (for you to check, per the request contract)

| | job 33 |
|---|---|
| input | **740,015** |
| cached | **654,592** |
| output | 13,065 |
| census | 49 files, +5186/−0 |
| range | `9da235a6cb..4408a9a8f` (endpoint verified = PR head) |

Genuine by a wide margin — the vacuous baseline is ~18.7k input / 0 cached. Every deterministic key PASSed: `push-assert`, `code-free`, `job-record`, `sha-assert`, `review-completed`, `prompt-content: PASS (7/7)`, `vacuity-tier1`, `vacuity-tier2`.

### What is being deferred, and why it is safe

Three Medium/Low residual guard defects in the harness, all **residuals of this PR's own final fix round**, filed as **#3863** (OPEN). Full detail there. The reason they cannot affect anything published:

| triggering condition | measured across all 15 reps |
|---|---|
| unattributable in-binary cycles | **0** |
| skipped/unparseable perf sample rows | **0** of 2,687,882 |
| lost samples | **0** (15/15) |
| unresolved in-binary addresses | **0** — 3430/3430 resolve |

So finding 2's understatement has no weight to act on, finding 3's double-count has no rows to double-count, and finding 1 requires a clock-read failure that did not occur. **Every published figure is unchanged across all seven rounds**: AC1 1.7027% total / 2.9779% in-binary, Route 2 2.2624%, disagreement 1.329×, AC2 1.061×, `pct_outside_binary` 42.83%.

### Why I am asking now rather than fixing again

Round history: **4 → 5 → 4 → 3 → 3 → 3**. Rounds 3–7 were each majority-residual — the fixes for one round generated the next round's findings, all inside newly-added guard code.

I declined to defer three times before this, and I want to be clear I am not now claiming those were mistakes: rounds 2–7 produced real value — a false-zero `no_chain` guard, a **shipped** caller-attribution regression (100% of cycles under caller `?`), a **zero-width skid band** that would have reported an unmeasured sensitivity as measured, and the discovery that "0 lost samples" was an ad-hoc read rather than an enforced control. Twice I judged a finding *could* move a published number and fixed it rather than defer; both times re-measurement showed it did not, which is the outcome I wanted but not one I could assume.

What has changed is that this round's findings are the first where **every** triggering condition is measured at zero, so the marginal round has stopped changing anything a reader of the report would act on. The remaining exposure is latent guard correctness on a reproducibility artifact, which #3863 owns.

### To authorize

Post a **new** top-level comment on this PR whose **only** non-blank content is the marker line — no prose, no fence, no quoting, or it is inert by design. Put any commentary in a separate comment. Fields, verified:

```
roborev-defer: findings issues=#3863 count=3 base=9da235a6cb9c48b78e019a359e8851ad8d007e09 head=4408a9a8f3326ce8522743c79e3b5301dd111c20 job=33 reason=<your reason>
```

`base` is the **merge-base** (from `assert-base:`), not `origin/main`'s tip. Substitute a real `reason` — a placeholder like `<why>`/`todo`/`tbd` is refused. I will then apply it with `--recheck-job 33`, which enqueues no new review, and the block will report `findings: DEFERRED (…)` — never `NONE`.

**If you would rather I just fix them, say so and I will run round 8 instead.** I have no attachment to deferring; I am asking because I think the honest reading is that we have hit diminishing returns, and that is a judgement you may weigh differently.

pmcfadin:
roborev-defer: findings issues=3863 count=3 base=9da235a6cb9c48b78e019a359e8851ad8d007e09 head=4408a9a8f3326ce8522743c79e3b5301dd111c20 job=33 reason=three Medium/Low residual validity-guard defects in the WS0 vint harness, all residuals of this PR final fix round, filed as #3863 (OPEN); every triggering condition measured at 0 across 15 reps (0 unattributable in-binary cycles, 0/2687882 skipped sample rows, 0 lost samples, 3430/3430 addresses resolved) so no published figure can move; rounds 3-7 each majority-residual (4-5-4-3-3-3) = non-converging carve; verdict KILL at 1.70% pre-registered basis already ruled; job 33 tokens input=740015 cached=654592 output=13065 vs vacuous 18700/0, prompt-content PASS 7/7; lead verified base == merge-base(origin/main, head) and head == live PR #3840 head

Issue #2210, opened 2026-07-08
F6.1: MADV_RANDOM on a dedicated point-read mmap + Linux cold-cache A/B harness


pmcfadin:
Decomposed out of #1596 (F6, Epic F #1518) per owner decision 2026-07-06. **Owner ACCEPTED** the approach; filed as its own issue because it needs a dedicated mapping + a Linux bench.

## Problem
The mmap POINT-read path gets no `MADV_RANDOM`. Auto currently applies no advice — that was a MEASURED decision for SCANS in #1143; point reads are a different access pattern and were not part of that decision.

## Constraint (why this can't be a one-liner)
The point source shares the scan source's `Arc<Mmap>`, and `madvise` is per-mapping — so advising `MADV_RANDOM` would also hit scans and **violate #1143's measured scan behavior**. Point-only advice therefore requires a **dedicated 2nd mmap** for the point-read path.

## Acceptance criteria
- A dedicated point-read mmap (or equivalent per-handle mapping) carries `MADV_RANDOM`; the scan source's mapping is UNTOUCHED (#1143 scan behavior preserved — pin it with a test/assertion).
- **Measure-first (hard #1143 lesson):** justified ONLY by a **cold-cache A/B on Linux** (this project's macOS dev box cannot be trusted for mmap/madvise timing). No merge on criterion-green alone — the win must show in a real cold-cache read-path number. If the Linux A/B shows no win, do NOT ship the 2nd-mmap complexity.
- No `unwrap()`/`expect()` in library code; no-heuristics unaffected.

Oracle/perf; no OpenSpec (it's a mechanical mapping change gated on measurement). Bench harness may be the larger part of the work.

pmcfadin:
Adopted onto epic #2403 (0.15 cqlite-trino latency/throughput/operations theme — lane: point-read latency) per the owner-approved backlog scan 2026-07-13. Milestone → 0.15; added to the CQLite Delivery board at Backlog (Ready ordering stays the groom's call). Adopt-by-link, not re-filed. *(flow-lead)*

pmcfadin:
🔒 Claimed by flow-lead worker (branch `issue-2210-madv-random-point-mmap` pushed to origin). Oracle/perf, no OpenSpec per the issue's own routing. Measure-first: will not ship the 2nd-mmap complexity without a real Linux cold-cache A/B showing a win, per the issue's own acceptance criteria.

PR #3841, opened 2026-09-01
docs(#2824): madvise(WILLNEED) under Auto-mmap — built, measured, rejected (priced re-scope)

### LINKED ISSUE #2824: `madvise(WILLNEED)` under Auto-mmap + `MADV_DONTNEED` post-scan
Flip `Auto` to issue the **already-built** `madvise(MADV_WILLNEED)` at open (`reader/mod.rs:1052`,
currently gated off by `PrefetchMode::Auto → None`) + `MADV_DONTNEED` **post-scan-once**; keep
`posix_fadvise` only behind an explicit buffered/direct backend. This is a **policy flip on built
machinery**, not new IO code. **Do NOT reach for `MADV_SEQUENTIAL`** — that is the #1143 ~2× p99
drop-behind regression; re-run `issue_1143_mmap_prefetch_tail_guard.rs` if the prefetch path is touched.

## Acceptance criteria

- **Cold-p99** on a **cold i4i scan** improves marginally, with **no warm regression** and **no #1143
  tail reintroduction** (`issue_1143_mmap_prefetch_tail_guard.rs` green).
- `MADV_DONTNEED` is issued post-scan-once so a full-ring scan does not leave the page cache warm past
  its usefulness (B4 peak hygiene).

## Evidence

- `docs/research/phase2-verify-io.md` (RESPEC — policy flip on built machinery; io_uring/O_DIRECT NOT
  filed, gated behind M0; §4).
- `docs/research/phase1-3-linux-io.md` (§2b IO headroom; the fadvise lever is dead code on the
  Auto-mmap path).
- `docs/architecture/throughput-program-2026-07.md` §4 (madvise lever, B4 justification stale
  post-#2412), §7 manifest item M10.

## Dependencies

- **M0** — the B4 justification is **stale post-#2412**; reprice this as a **cold-p99 hedge** and
  defer behind the i4i re-measurement (do not lead with it).

**Dedup:** NEW; respec of P1.3's fadvise lever (dead code on the Auto-mmap path). Adjacent to #1518.

---
`docs/architecture/throughput-program-2026-07.md` (§7 manifest item M10).

pmcfadin:
## What this is

**Issue #2824's lever was built, measured, and rejected. This PR lands the investigation instead of the change.** Lead ruling on `REQ-2824-03` (re-scope) and `REQ-2824-04` (drop the harness).

**There is no code in this diff.** Every `*.rs` file and both SSTable-guide chapters match `origin/main` byte for byte; `PrefetchMode::Auto` still issues no `madvise`.

Closes #2824. Successor: **#3853** (scan-lifetime madvise plumbing, for the i4i rig).

## The finding

Issue #2824 described itself as *"a **policy flip on built machinery**, not new IO code"*. That premise is false, for **both** of its halves.

`SSTableManager::new` → `load_existing_sstables()` (`storage/sstable/manager_open.rs:61`) walks the data directory to depth 3 (`MAX_SSTABLE_SCAN_DEPTH`, `storage/sstable/mod.rs:121`) and **opens every `*-Data.db` in a loop** (`manager_open.rs:300`), on the `Database::open` path (`lib.rs:240`) — library, CLI interactive/ingest, and both Node and Python bindings. So with `Auto -> WillNeed`:

> `Database::open` issues one whole-file `MADV_WILLNEED` per SSTable, for **every SSTable of every table of every keyspace**, before any query is seen.

Point-lookup-only workloads pay that in full, on tables they never touch — and point lookups open *zero* additional readers precisely because everything was opened already, so bloom/summary gating cannot help; the read-ahead is queued before any filter runs. Flight has the same shape per table on first touch. Compaction is unaffected (forces buffered I/O).

Deferring the advice to scan start needs a scan-lifetime seam that does not exist: nine scan entry points in three shapes, no reader-scoped scan-in-flight state, concurrent scans over one reader a supported and tested property (#815), and `DONTNEED` requires `unsafe` (memmap2 puts it in `UncheckedAdvice`). That is the *same* plumbing AC2 needs — hence both halves.

Found by roborev job 340 (High) and verified independently against the source before acting.

## And the benefit was undemonstrable

`docs/reports/issue-2824-artifacts/RESULTS.md`. Scan-attributable major faults **5.0 vs 3.0** medians across ~630,000 pages; every phase validated to have scanned identical work; 16/16 cold phases verified `DRAINED`; warm no regression. The 2-fault gap is **not reported as a signal** — it is the same order as the declared arm-asymmetric residual, and 2 faults out of 630,000 pages is nil.

The corpus device is **EBS at a measured 132 MB/s with a 128 KiB read-ahead window**, so the default window already saturates it and the read is bandwidth-bound: **no headroom to detect the effect in either direction**, and a null result here says nothing about i4i. A confirmed cost against an unconfirmed benefit.

## Review status — read this before merging

**This diff is code-free, so it CANNOT be roborev-certified, and no roborev-clean claim is made for it.** Per the docs-only rule, "docs-only" is a code-free census as the classifier computes it, never a path prefix:

```
$ git diff --name-only origin/main...HEAD | bash scripts/ci/classify-docs-only.sh
docs-only
classify-docs-only: all changed files are docs -> short-circuit to green
$ echo $?
0
```

The **code** that this investigation produced was reviewed while it existed — rust-reviewer (5 findings, all fixed) and roborev jobs 340/342/343/344/346/347 — but every line of it has since been reverted or removed, so none of it is in this diff and none of those verdicts certifies this diff. The substitute is primary-source verification, below.

## Primary-source verification

Format and behaviour claims rest on primary sources, not on CQLite's own code:

- **`MADV_DONTNEED` on a file-backed mapping does not evict page cache.** `madvise(2)`: after `MADV_DONTNEED`, accesses "result in repopulating the memory contents from the up-to-date contents of the underlying mapped file (for shared file mappings …)". It frees resident PTEs, i.e. RSS. Issue #2824's AC2 claimed page-cache hygiene; that wording is refused here and recorded as a standing boundary in the spec delta.
- **`MADV_DONTNEED` is not in memmap2's safe `Advice` enum** — it is `UncheckedAdvice::DontNeed` (memmap2 0.9.11, `advice.rs`), so it requires an `unsafe` call.
- **The advice actually issued** was verified by `strace`, not inferred (`ab/strace-madvise.*.txt`, `ab/advice-census.txt`): both arms issue exactly one **successful** `MADV_RANDOM` (the #2210 point mapping); only the patched arm adds one successful `MADV_WILLNEED` at a different address (the scan mapping); **neither arm ever issues `MADV_SEQUENTIAL`** (#1143), verified at runtime rather than only by unit assert.
- **The device and its throughput** were measured directly, not assumed: `Amazon Elastic Block Store`, `read_ahead_kb=128`, 2,774,760,422 bytes read cold in 21.0 s = 132 MB/s.
- **The eager-open path** is quoted by `file:line` from the repository under review — evidence of what CQLite does, which is the appropriate kind of evidence for a claim *about CQLite*.

## What ships

| file | what |
|---|---|
| `docs/reports/issue-2824-artifacts/RESULTS.md` | the measurement, its four declared limits, and the **full method** — environment, invocation sequence, both prefetch arms, the floor phase, the drain poll, the validity checks and the strace census — so it is reproducible without a script |
| `.../ab/construction.md` | the A/B construction assertion (source + syscall) |
| `.../ab/{summary,scan-attributable,drain}.csv`, `host.txt`, `advice-census.txt`, `strace-*` | raw artifacts |
| `openspec/changes/madvise-willneed-dontneed/**` | proposal, spec delta, tasks |
| `docs/research/phase1-3-linux-io.md`, `docs/architecture/throughput-program-2026-07.md` | corrected: the lever was tried and rejected, not "currently gated off" |

The A/B **driver script is deliberately not committed**, following #2210, which kept its madvise-A/B harness out of tree as a scratchpad microbench. It took 22 review findings across 7 passes — three of them caused by the fix for the preceding one — and with the lever killed it has no future user; #3853 needs a harness with a per-operation latency distribution and multi-SSTable pressure that this one never provided. The method is recorded in `RESULTS.md` in full instead.

The spec delta adds no behavioural requirement. It records three things so the successor does not rediscover them: `Auto` must not advise **at open** (a reason independent of #1143 — reader-open is not evidence of intent to scan), `MADV_DONTNEED` is an RSS control and never a page-cache control, and a read-ahead A/B must record its device or a null result reads as "the lever does not work" when it means "this box cannot tell".

## Method note worth keeping

Arms were alternated per round, and that is what saved the result: in an earlier run whichever arm ran **first** was faster in *both* orderings (baseline-first 20.9 s vs 50.1 s; patched-first 38.8 s vs 22.4 s). Sequential arms would have yielded a clean-looking, entirely false "patched is 2.4x slower". Separately, an earlier binary pair built either side of a rebase was **discarded** — it differed in more than the property under test.

Part of epic #2817 (0.17 scan-path throughput program), manifest item M10 — recorded there as re-scoped.


pmcfadin:
**BLOCKED — do not merge.** Converted to draft; `--auto` was never armed.

roborev job 340: `RESULT: FAIL`, 1 High (all deterministic keys PASS; 433k input / 363k cached tokens, so a genuine review). Verified independently and **confirmed**:

`SSTableManager::new` → `load_existing_sstables()` (`storage/sstable/manager_open.rs:61`) walks the data dir to depth 3 (`MAX_SSTABLE_SCAN_DEPTH`, `storage/sstable/mod.rs:121`) and opens **every** `*-Data.db` in a loop (`manager_open.rs:300`), reached from `Database::open`. With `Auto → WillNeed` that means **one whole-file `MADV_WILLNEED` per SSTable of every table of every keyspace, at database open, before any query is seen** — paid in full by point-lookup-only workloads, on tables they never touch.

The fix (defer the advice to scan start) needs the same scan-lifecycle seam AC2 was sliced out for, so the issue's "policy flip on built machinery" premise fails for **both** halves. A bounded-prefix advice would be an unmeasured heuristic (#28).

Escalated as REQ-2824-03 on issue #2824 with three options; recommendation is to re-scope and land the findings rather than the flip. Awaiting the lead — this is a scope call, not mine.

PR #3368, opened 2026-08-27
test(#1714): AK3 standing mod-reachability guard for cqlite-core/src [HELD — blocked on #3366, do not merge]

pmcfadin:
Refs #1714. **DO NOT MERGE — held by lead ruling, blocked on #3366.**

> **Status: OPEN, UNARMED, intentionally not merging.** `--auto` is deliberately **not** armed.
> The delivery lead [revised an earlier merge approval and held this PR](https://github.com/pmcfadin/cqlite/issues/1714#issuecomment-5435644930): the hand-written-lexer design is being replaced by **#3366** (reachability derived from rustc's own dep-info), which is sequenced ahead of it. This PR is preserved as the evidence and test corpus for that work, **not** as a merge candidate. It deliberately does **not** say `Closes #1714`.

## What it contains

A standing **mod-reachability guard** for `cqlite-core/src`: every `.rs` file under it must be reachable from a `mod` declaration chain rooted at `lib.rs`. A file with no such chain is **never compiled by rustc under any feature combination**, so edits to it are silent no-ops and tests inside it never run — the failure mode #1714 was filed for, after a dead file misled an auditor into a refuted finding.

| file | lines |
|---|---|
| `cqlite-core/tests/support/mod_reachability.rs` | 977 |
| `cqlite-core/tests/support/mod_reachability_lexer.rs` | 673 |
| `cqlite-core/tests/support/mod_reachability_harness.rs` | 108 |
| `cqlite-core/tests/issue_1714_mod_reachability.rs` | 1301 |
| `cqlite-core/tests/issue_1714_mod_reachability_shebang.rs` | 148 |

**34 tests** (30 + 4). AK3's deletion half was already done: `cqlite-core/src/memory_safety_tests.rs` went out in `fb8330a6a` (PR #2044) before this issue was claimed, so the guard was the entire remaining deliverable.

## Why it is held — six rustc divergences over five roborev rounds

Every false-PASS finding was one class: **our lexer disagreeing with rustc's.** Each was fixed; the class regenerated. Findings per round: **3, 2, 2, 2, 3 — it went up.**

| round | divergence | class |
|---|---|---|
| 1 | `mod` inside a `macro_rules!` body / `stringify!(…)` | false PASS |
| 2 | `cr"…"` / `cr#"…"#` raw C-strings unrecognized | false PASS |
| 2 | raw identifiers (`r#make`) not consumed atomically | false PASS |
| 3 | identifiers ASCII-only; Rust permits Unicode XID | false PASS |
[...]

PR #3845, opened 2026-09-01
test(#1714): AK3 mod-reachability guard for cqlite-core/src (re-scoped narrow core)

### LINKED ISSUE #1714: AK3 — Delete orphan memory_safety_tests.rs (never compiled) + core mod-reachability guard
Part of epic #1688 (dead code + pub-surface). Source: `docs/reports/platform-observability-audit-2026-07-01.md` finding AK3 (audit block 6). Wave 2.5 purge (decision #9). Mirrors export AF1 #1502's CLI mod-wiring guard — same guard class, core crate.

## Problem
`cqlite-core/src/memory_safety_tests.rs` (482 LOC) has **no `mod` declaration anywhere** — it is never compiled under any feature combination (lead-verified). It references `MemTable`, a type deleted in #175. Edits to it are silent no-ops; its existence misleads readers (and misled a block-6 auditor into a refuted finding — dead files are ACTIVELY harmful).

## Fix (effort: S)
1. Verify: `rg -rn "mod memory_safety_tests" cqlite-core/` → zero hits. Delete the file.
2. **Standing mod-reachability guard** (the real deliverable): a repo-hygiene test asserting every `cqlite-core/src/**/*.rs` file is reachable from a `mod` declaration chain rooted at `lib.rs` (walk `mod` statements, or compare `rg --files` against the module graph; mirror AF1 #1502's implementation for the CLI — reuse its helper if it landed, one shared implementation for both crates is ideal).
3. The guard must account for `#[path]` attributes and cfg-gated mods (a feature-gated mod IS reachable).

### Do NOT
- Do not port/fix the tests inside the file — the type they test was deleted in #175; the memory-safety concern is covered by the live test suites and (H1 #1614) fuzzing.

## Acceptance criteria
- [ ] File deleted; mod-reachability guard landed for cqlite-core src and green.
- [ ] Guard demonstrably fails on pre-delete main (this file) AND on `schema/cql_generator.rs` (AK4's file — coordinate; if AK4 lands after, its file is the guard's remaining red until then, which is acceptable ordering: land AK4 first or together).

The mod-reachability guard against pre-delete main: flags `memory_safety_tests.rs` (and `cql_generator.rs`).

pmcfadin:
Closes #1714.

Delivers the **re-scoped narrow core** of AK3 per the owner ruling of 2026-08-30: a simple
mod-reachability guard for `cqlite-core/src`. The adversarial lexer-equivalence design is not
here — PR #3368 was closed unmerged and this is a fresh branch off `origin/main`, not a rebase
of it.

## Diff

One new file. **No `src/`, no `scripts/`, no gate-component, no `Cargo.toml`, no new dependency.**

```
cqlite-core/tests/issue_1714_mod_reachability.rs | 1068 ++++++++++++++++++
```

`cqlite-core/tests/*.rs` is already inside the gate's core test component, so the guard needs no
new component to run.

## Acceptance criteria

**AC1 — file deleted; guard landed and green.**
`cqlite-core/src/memory_safety_tests.rs` was **already deleted from `main` by `fb8330a6a`**
(PR #2044's dead-cache purge, 482 deletions) — incidental to a different issue. This PR does not
re-delete it and does not claim the deletion. The guard half is what ships here, and it is green:
**526 files enumerated, 526 reached, 0 orphans.**

All four orphans the previous lane's guard found are gone from `main`, so **this PR ships no
exception list at all** — the `EXPECTED_ORPHANS` mechanism that produced the cross-lane
merge-order coupling reported on #1714 (two entries already stale, a third about to be) has no
subject and is simply absent. If a legitimately-unreachable file ever appears, the PR that
introduces it adds the mechanism then.

**AC2 — guard demonstrably fails on pre-delete main.** Run against `87792e5f6` (`fb8330a6a^`) in
a detached worktree, since removed:

```
cargo test -p cqlite-core --test issue_1714_mod_reachability every_cqlite_core_src_file_is_reachable
```

Reported orphans (4):

- `cqlite-core/src/memory_safety_tests.rs`   ← named by AC2
- `cqlite-core/src/schema/cql_generator.rs`  ← named by AC2
- `cqlite-core/src/parser/collection_udt_tests.rs`
- `cqlite-core/src/storage/sstable/header_fix_functions.rs`

Both files AC2 names are present, plus two the guard was not told about. Recorded honestly: the
first run red on the **vacuity floor** instead (that tree holds 335 `.rs` files against a floor of
400) — correct precedence, but it masks the orphan list, so the floor was lowered to 300 **in the
throwaway historical copy only** to reach the orphan assertion.

**AC3 — full `agent-gate.sh` PASS.** Run by `flow-closer`; the SUMMARY block is pasted below this
line when it lands.

## What the guard models, and what it deliberately does not

Walks `mod` declarations from `src/lib.rs`; compares the reachable set against a recursive
enumeration of `src/**/*.rs`. Handles comments (nested), string and raw-string literals,
`#[path]`, cfg-gated `mod`s (gated ⇒ reachable), inline `mod` blocks and out-of-line modules
nested inside them, and refuses loudly on an absolute `#[path]` or a `path = …` value it cannot
read as a plain string literal.

**Not modelled, by owner ruling:** shebang handling, non-ASCII/raw identifiers, a literal-prefix
table, byte-char/escaped-quote spans, and any "refuse everything unmodelled" path. The residual is
a documented textual false PASS. **#3366** (set difference against rustc's own `--emit=dep-info`,
which has no lexer to diverge) is the durable replacement and the module header says so.

Two file-resolution rules are load-bearing and pinned in both directions:

- `#[path = "X"]` resolves relative to the **directory containing the declaring file**, not the
  `<stem>/` subdirectory. Getting this wrong reports **57 false orphans** in this repo.
- An out-of-line `mod` inside an inline `mod outer { }` resolves under `outer/`.

## Verification

**Roborev: clean.** Sanctioned wrapper, `--agent codex --model gpt-5.6-sol`.

```
head-sha: a4ee12742001ca515d1ec1391f8005eed298c16e
reviewed-sha: 9da235a6cb9c48b78e019a359e8851ad8d007e09..a4ee12742001ca515d1ec1391f8005eed298c16e
job: 341   model: gpt-5.6-sol   census: 1 file, +1068/-0
tokens: input=580080 cached=513280 output=12722
push-assert: PASS      census-check: PASS     code-free: PASS       job-record: PASS
sha-assert: PASS       review-completed: PASS prompt-content: PASS (1/1 code census paths present)
vacuity-tier1: PASS    vacuity-tier2: PASS    findings: NONE        roborev-exit: PASS
RESULT: PASS
```

Findings per round: **2 → 1 → 1 → 1 → 0**. Every finding fell inside a finite, converging family
(module-scope resolution, then platform-dependent absoluteness, then fixture portability), each
closed by enumeration rather than by patching the next instance — deliberately unlike the design
this replaces, whose five rounds ran 3, 2, 2, 2, 3 because each fix exposed a *new* rustc
divergence.

**Independent mutation review.** A reviewer planted **26 mutants; 23 were killed and 3 survived** —
all three survivors were tests that passed with their own mechanism deleted:

| survivor | mutation that survived | direction |
|---|---|---|
| nested block comments | delete `depth += 1` | fixture had the `mod` inside *both* comments |
| raw strings | `let raw = false` | **false PASS** — 82 raw strings in `src`, real-tree reached set byte-identical |
| line comments | disable `//` stripping | satisfied by line-anchoring, not by the stripper |

The raw-string one is why this round mattered: with raw detection broken, a raw literal containing
a bare `"` followed by a line-anchored `mod` exposes that `mod` as live source, so a genuinely
unreachable file reports as reachable — and neither the unit test nor the live assertion could see
it. All three now kill, re-verified independently at `ac8fe9b1e`. Replacement fixtures were
ground-truthed with `rustc --edition 2021 --crate-type lib`.

Also verified by that review: the vacuity floor fires **before** the walk (census and walk mutated
independently); zero panics across 400k hostile fuzz inputs plus every real source file sliced at
997-byte strides, debug build with overflow checks; and the refusal broadening cleared
**structurally** — `parse_path_attr` is byte-identical across the change, so the set of *resolving*
shapes cannot have moved; only silently-dropped became loudly-refused.

**A near-miss worth keeping in the record:** three `#[tracing::instrument(… cqlite.query.access_path
= …)]` attributes in `cqlite-core/src` contain the literal text `path = ` and are saved from the
refusal *only* by the ident-boundary check in `mentions_path_assignment`. Measured across all 526
files: 66 attribute interiors mention `path`, **0** would trigger the refusal. That check is
load-bearing, not decorative.

**Real-tree measurement behind the green.** Of the 59 distinct `#[path]` values in
`cqlite-core/src`: **0** begin with `/` or `\`, **0** contain `..`, **0** are drive-rooted, **0**
contain a colon. So the refusal arms demonstrably are not firing to produce 526/526.

**Windows fixture sweep** (closing the family rather than patching its third instance): all 20
distinct fixture paths audited for reserved device names, reserved characters, trailing dots and
spaces, and case-insensitive collisions — **one hit, the finding itself**, now `#[cfg(not(windows))]`.
The subject set was verified rather than assumed: an independent pass diffed every `.rs`-ending
literal in the file against the written set and accounted for all 14 remainders.

## Residuals

1. The documented textual false-PASS residual (line-anchoring, `#[path]` **on** an inline `mod`
   block, escapes in a `#[path]` value, macro-generated `mod`s, non-ASCII/raw identifiers, char
   literals containing `"`/`[`/`]`). A reviewer **measured** the char-literal case: adding a
   recogniser leaves the reached set byte-identical (526/526), and those enumerated residuals can only
   *hide* declarations — loud false orphans. That is a claim about the **enumerated list**, not
   about the general class: a `mod` inside an uninvoked `macro_rules!` body would read as
   reachable, which is a false pass. The module header states the class as a false pass and
   governs; it is not understated.
   **#3366** replaces the whole class.
2. Nothing here runs on a Windows runner (core CI is `ubuntu-latest`), so the portability work is
   unit-asserted and audited, never host-verified. Stated at the tests rather than implied to be
   mutation kills — arms 2-4 of the absoluteness helper cannot be killed on Linux because the
   host-native arm overlaps them.
3. The vacuity floor is tied to today's tree size (400 against 526, ~24% headroom). A future
   refactor shrinking `cqlite-core/src` below 400 files reds the guard on the floor rather than on
   orphans — deliberate: a census without a subject must not pass.
4. 1068 lines (31 tests, 30 on Windows) — over the ~550 initially budgeted, well under the
   1500-line test-file threshold; `file-size` PASSes.

### DIFF
--- a/cqlite-core/tests/issue_1714_mod_reachability.rs
+++ b/cqlite-core/tests/issue_1714_mod_reachability.rs
@@ -8,11 +8,15 @@
 //! previous lexer-equivalence design and closed its PR #3368): this is a
 //! deliberately SMALL textual walker. It does NOT model, and must not grow to
 //! model, shebang lines, non-ASCII or raw (`r#`) identifiers, a table of Rust
-//! literal prefixes, byte-char/escaped-quote spans, `#[path]` inside an inline
-//! `mod` block, `mod` declarations that do not begin their own (trimmed) source
-//! line, escape sequences inside a `#[path]` value, or a `mod` produced by macro
-//! expansion. There is deliberately **no exception list**: an orphan is a
-//! failure to fix, never an entry to add.
+//! literal prefixes, byte-char/escaped-quote spans (a `'{'` is guarded for
+//! explicitly, not lexed), `#[path]` ON an inline `mod` block (that form sets the
+//! directory its children resolve in), `mod` declarations that do not begin
+//! their own (trimmed) source line, escape sequences inside a `#[path]` value,
+//! or a `mod` produced by macro expansion. `#[cfg_attr(…, path = …)]` is
+//! DETECTED and refused loudly, never resolved. It DOES model out-of-line
+//! modules nested in inline `mod` blocks, `#[path]` on those, and attributes
+//! broken across lines. There is deliberately **no exception list**: an orphan
+//! is a failure to fix, never an entry to add.
 //!
 //! The vacuity floor (>= 400 enumerated files) is checked BEFORE the orphan
 //! assertion, so on a tree far smaller than today's it reds on the floor rather
@@ -140,6 +144,12 @@ fn strip(src: &str) -> Stripped {
 struct ModDecl {
     name: String,
     path_attr: Option<String>,
+    /// Names of the enclosing inline `mod` blocks, outermost first. rustc
+    /// resolves such a child under `<moddir>/<name1>/…/<namen>/`.
+    scope: Vec<String>,
+    /// A `#[cfg_attr(…, path = "…")]` preceded this declaration. That form is
+    /// NOT modeled; the walk refuses loudly rather than resolve it wrongly.
+    cfg_attr_path: bool,
 }
 
 fn is_ident_char(c: char) -> bool {
@@ -147,8 +157,14 @@ fn is_ident_char(c: char) -> bool {
 }
 
 /// Consume leading `#[...]` / `#![...]` attributes from a trimmed line,
-/// recording any `#[path = "..."]` value in `pending`. Returns the remainder.
-fn eat_attrs<'a>(mut t: &'a str, s: &Stripped, pending: &mut Option<String>) -> &'a str {
+/// recording any `#[path = "..."]` value in `pending` and flagging any
+/// `#[cfg_attr(…, path = …)]` in `cfg_attr_path`. Returns the remainder.
+fn eat_attrs<'a>(
+    mut t: &'a str,
+    s: &Stripped,
+    pending: &mut Option<String>,
+    cfg_attr_path: &mut bool,
+) -> &'a str {
     while t.starts_with('#') {
         let after_hash = t[1..].trim_start();
         let after_hash = after_hash.strip_prefix('!').unwrap_or(after_hash);
@@ -172,8 +188,11 @@ fn eat_attrs<'a>(mut t: &'a str, s: &Stripped, pending: &mut Option<String>) -> 
             }
         }
         let Some(close) = close else { break };
-        if let Some(v) = parse_path_attr(t[open + 1..close].trim(), s) {
+        let inner = t[open + 1..close].trim();
+        if let Some(v) = parse_path_attr(inner, s) {
             *pending = Some(v);
+        } else if inner.starts_with("cfg_attr") && mentions_path_assignment(inner) {
+            *cfg_attr_path = true;
         }
         t = t[close + 1..].trim_start();
     }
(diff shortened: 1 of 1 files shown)


pmcfadin:
## Gate-of-record record (posted post-merge for the trail)

This PR's gate of record ran **twice**. Both are recorded here per the rule that a red on a compiled-input diff is cited, never waived bare.

| run | run-id | verdict | note |
|---|---|---|---|
| 1 | `/tmp/agent-gate.wW9LJL` | `RESULT: FAIL` | 35/36 PASS; sole red `tooling-tests: FAIL (1488s)` |
| 2 | `/tmp/agent-gate.BodL4T` | **`RESULT: PASS`** | 37/37, `tooling-tests: PASS (2307s)` — **the gate of record** |

Both at the same head `a4ee12742001ca515d1ec1391f8005eed298c16e`, `tree-integrity: PASS`, `dirty: no`. The tree was never written to between them.

### The run-1 red is a known flake, filed as #3860 — mechanism UNDETERMINED

The failing line was `FAIL: case 19 setup: the decoy worktree was never registered`, a **setup** step in `scripts/tests/test_pub_surface_guard.sh` that registers a `git worktree` decoy in the fleet's shared `.git`. It is not reachable from this diff:

- The guard under test invokes zero cargo and reads only `cqlite-core/src/lib.rs` plus resolved module prologues; `grep -c "cqlite-core/tests" scripts/tests/test_pub_surface_guard.sh` = **0**. This diff touches only `cqlite-core/tests/`, so no compiled input participates in case 19 at all.
- Same tree, same commit, same box, standalone: the identical suite passed **42/42**. Clean-`origin/main` reproduction passed 4/4 including two concurrent checkouts sharing `.git`.
- Base rate across 10 days of gate logs on this box: 92 `tooling-tests.log` files, 80 recording the `all 42 cases` PASS line, **exactly one recording `case 19 setup` — this incident.** 1 in 81.

**Mechanism is undetermined.** Two candidate causes were proposed and neither reproduced — an admin-basename collision (excluded at 440 trials; git disambiguates) and a `pipefail`/SIGPIPE interaction (excluded at 560 trials). A third, a `git worktree prune` reclaiming a half-built admin entry, has a measured *consequence* but no observed interleaving and did not reproduce in a 400-trial hostile loop. The failure is established as environmental and non-deterministic; it is deliberately **not** attributed to a named cause here.

No source was patched to green the gate, and the diff is unchanged from the reviewed and certified tree.

### Merge certification

- Gate of record: run 2 above, `PREMERGE: OK a4ee1274…` + `PREMERGE: GATE-OF-RECORD commit: a4ee127 tree-start: a4ee12742001 tree-integrity: PASS dirty: no`.
- roborev job 351: `RESULT: PASS`, `findings: NONE`, all integrity keys PASS, `prompt-content: PASS (1/1)`.
- C intent audit: PASS.
- Base-staleness advisory (#3650 slice 1) reported `STALE-RECOGNISED`: 8 commits behind, 1 in the blast radius (`e6389ab31`, `cqlite-core/Cargo.toml`). Inspected and judged benign — it adds only `exclude = ["tests/issue_3358_bti_query_token_bound.rs"]`, a packaging key that cannot affect compilation. The guard's actual subject, the `cqlite-core/src` module graph, is unchanged across all 8 commits behind (0 src files added or deleted, 0 `mod` declarations changed), so the merge-result verdict is provably the certified one.

Per #3650 this is a **gate of record verified at `a4ee127`**, not a certification against current `main`.

PR #3872, opened 2026-09-01
fix(#3772): stop quoting a hard-coded jest suite count in the gate and jest config (slice 1/2)

### LINKED ISSUE #3772: agent-gate.sh prints a hard-coded jest suite count (28) that is stale by 4, and jest.config.js repeats it
Follow-up from the C intent audit on #3641 / PR #3771. Both items are nits by `docs/development/roborev-severity.md`, batched here per doctrine (nits never trigger a re-verify round).

## 1. A printed measurement nobody re-measured — `scripts/agent-gate.sh:10721`

The `node-bindings` component prints, on every run:

```
census+=("       all-projects jest --listTests is 28 files with no duplicates and includes the")
```

Measured: `./node_modules/.bin/jest --listTests` is **31 on `origin/main`** and **32 after PR #3771** (which adds `helpers.test.js`). So the `28` was already stale by 3 before that PR and is stale by 4 after it. `bindings/node/jest.config.js:20` carries the same number in a comment ("measured via `jest --listTests`: 28 files, no duplicates").

**It changes no verdict** — `check_jest_suites_ran` reconciles two oracles derived at run time (a recursive `find` over `__test__/` and `jest --listTests`), so the gate cannot go wrong on this line. What is wrong is that a **gate log states a measurement that is false**, in the census whose entire job is to describe what ran. That is the class CLAUDE.md calls out: *"a false rationale in a gate log is worse than none, because it is what stops the next person looking."*

The better fix is probably not to update `28` → `32` — which just resets the clock on the same decay — but to **print the derived count** the component has already computed, or drop the number and keep the property ("no duplicates"). A hard-coded count in a line printed beside the live derivation is the avoidable half.

Why it was not fixed in #3771: it is pre-existing (stale by 3 before that diff), it lives in `scripts/agent-gate.sh`, and a script edit is **not** `--delta`-eligible (#1892 fails closed on anything outside executable tests/docs) — so folding it in would have cost a second full gate of record for a comment. Scope belongs with the #3522/#3493 census work, not with a Node test-helper rename.

## 2. A describe title that overstates its test — `bindings/node/__test__/helpers.test.js:87`

`describe('the dataset assertion is not duplicated (issue #3641)')` contains one test that matches the **exact error-message literal**. A fourth copy of the helper's body with a *reworded* message evades it. The test's own name ("no test file inlines the helper error message") is accurate; the describe title claims the stronger property.

Also worth an exemption path: any future `*.test.js` that legitimately quotes the literal — `dataset-guard.test.js` already spawns child jest processes and asserts on their output — would be reported as an offender with no way to opt out. Per CLAUDE.md, *a guard that reds on correct input is the guard agents learn to waive*, so an explicit allow-list (with a comment saying when to add to it) beats discovering this the hard way.

There is a small irony worth preserving in the fix: #3641 is *about* a name claiming behaviour it does not have, and it shipped with a describe block doing a milder version of the same thing.

pmcfadin:
## What

Removes the hard-coded jest suite count (`28`) from the gate and the Node jest config, replacing it with the property plus a pointer to the run-time derivation.

**Slice 1 of 2 for #3772.** This PR deliberately does **not** carry `Closes #3772` — see "Why this closes nothing" below. That is an intentional deviation from the usual `Closes #<N>` mandate, not an omission.

## Why

`node-bindings` printed, on every gate run:

```
all-projects jest --listTests is 28 files with no duplicates and includes the leak file
```

Measured at `origin/main` (`ef083e2bf`), `bindings/node/__test__` holds **32** `*.test.js` files. The issue measured 31 when it was filed; it moved again while the issue sat, which is the decay argument demonstrating itself.

It changes no verdict — `check_jest_suites_ran` reconciles two run-time oracles, so the gate cannot go wrong on this line. What was wrong is that **a gate log stated a false measurement, inside the census whose entire job is to describe what ran**: *"a false rationale in a gate log is worse than none, because it is what stops the next person looking."*

## The number was in FIVE places, not two

| site | claim |
|---|---|
| `scripts/agent-gate.sh:1892` | `#1465` composition header — "the all-projects list is 28 files with NO duplicates" |
| `scripts/agent-gate.sh:1936` | `_NODE_LEAK_SUITE_FILE` scope note — "the WHOLE-SUITE report (28 suites)" |
| `scripts/agent-gate.sh:1985` | `_node_leak_lane_note` dataset-gate note — "skips all 28 suites" |
| `scripts/agent-gate.sh:10792` | the census line the issue reported |
| `bindings/node/jest.config.js:20` | "measured via `jest --listTests`: 28 files, no duplicates" |

Per the issue's own recommendation this is **not** `28` → `32`, which only resets the same clock. No replacement count is written anywhere — including inside the explanations of why it was removed.

## Deleting the number is safe because both halves are already derived

- **the COUNT** is reconciled in `run_node_bindings` from two independent oracles (recursive `find` over `__test__/` vs `jest --listTests`) and already printed as `suite set RECONCILED: N *.test.js file(s)`;
- **a DOUBLE execution** fails `check_jest_suites_ran`'s `s_total -eq expected`, where `expected` is the **deduplicated** disk inventory.

## And the deleted sentence named the wrong oracle — the better reason to delete it

I checked that something still measures "no duplicates" before deleting the words, because *drop the number, keep the property* is only honest if the property is still measured. It was not measured by the cited method. Measured on jest 29.7.0, two projects whose `testMatch` both select one file:

```
jest --listTests  ->  __test__/probe.test.js          (ONE line)
jest              ->  Test Suites: 2 passed, 2 total  (executed TWICE)
```

`jest --listTests` **deduplicates across projects**, so it can never report the duplicate it was cited as having ruled out. Verified against this package too: deleting the `testPathIgnorePatterns` entry that hands the leak file to the `leaks` project leaves `--listTests` byte-identical. The `sort -u` in the gate's own reconciliation would erase one as well.

So `"28 files with no duplicates, measured via --listTests"` was **unfalsifiable by its own stated method**. Only the run's suite TOTAL sees it, and the comments now send a reader there instead of to `--listTests`.

## No test is added, deliberately

The diff changes comments and one census string; it alters no control flow, and item 1 "changes no verdict" by construction. A test for the *duplicate-detection* property (plant a duplicate, assert `node-bindings` reds) would be worth having, but it needs `npm ci` + a napi build per run and belongs with the `#3522`/`#3493` census work rather than a comment-truthfulness fix. Flagging it rather than leaving it implied.

## Certification

```
MODE: lite   commit: 6eb27b4   dirty: no
tree-integrity: PASS
file-size: PASS (1s) | fmt: PASS (64s) | clippy: PASS (415s)
roborev-lints: PASS (64s) | scoped-tests: PASS (153s) [test cqlite-node default-features]
cpu-budget: max-concurrency=1(pinned)
RESULT: PASS
```

Full gate of record + C + final roborev run in `flow-closer` before merge.

## Why this closes nothing

#3772 batches two nits. Item 2 (`bindings/node/__test__/helpers.test.js` — a describe title that overstates its test, plus an exemption path) targets a file that **does not exist on `origin/main`**: it is added by #3771, which has not had its gate of record yet.

Holding this PR for that was my first plan and it was wrong on measurement: **item 2 is `--delta`-eligible (#1892 — a node binding test against an already-built module) and item 1 is not** (`scripts/agent-gate.sh` fails `--delta` closed). So *item 1 full gate → item 2 `--delta` re-cert* costs **one** full gate either way; holding only adds latency behind another lane's slot queue.

So #3772 stays **open** with item 2 in scope as slice 2 — the shape #3550/#3559 defines (one telemetry record per shipped PR, `--slice`, `closed_at: null`; a slice PR closes nothing). Item 2 is not dropped, split out, or re-scoped.

`git merge-tree --write-tree HEAD <3771-head>` is **clean** — #3771 also touches `scripts/agent-gate.sh`, so I verified that mechanically rather than by eye; the hunks are ~70 lines apart and neither lane blocks the other in either merge order.

Refs #3772

### DIFF
--- a/bindings/node/jest.config.js
+++ b/bindings/node/jest.config.js
@@ -22,11 +22,12 @@
  *     before adding a second lane that selects this project: two executions where
  *     only one is affirmed is what the recomposition removed.
  *
- *     NO FILE COUNT IS QUOTED (issue #3772). This note used to end "measured via
- *     `jest --listTests`: 28 files, no duplicates", repeating a number
- *     `scripts/agent-gate.sh` carried in four more places; the suite grew and every
- *     copy went stale, so the sentence asserted a false measurement in the one place
- *     a reader checks this composition against. Nothing here needs the count: the
+ *     NO FILE COUNT IS QUOTED (issue #3772). This note used to end with a hard-coded
+ *     file count "measured via `jest --listTests`", asserted as duplicate-free —
+ *     repeating a number `scripts/agent-gate.sh` carried in four more places; the
+ *     suite grew and every copy went stale, so the sentence asserted a false
+ *     measurement in the one place a reader checks this composition against. Nothing
+ *     here needs the count: the
  *     `node-bindings` gate component DERIVES it every run from two independent
  *     oracles and prints `suite set RECONCILED: N`.
  *
--- a/bindings/node/jest.config.js
+++ b/bindings/node/jest.config.js
@@ -17,11 +17,30 @@
  *     own (`--selectProjects leaks`, used by `npm run test:leaks`) and so a future
  *     lane-only option has somewhere to live. NOTE (round 9): a bare `npm test` runs
  *     BOTH projects, so the gate's whole-suite run already executes the leak file
- *     exactly once — measured via `jest --listTests`: 28 files, no duplicates. The
- *     gate therefore does NOT invoke `test:leaks`; it affirms the budget tests by
- *     name from the whole-suite `--json` report. Keep that in mind before adding a
- *     second lane that selects this project: two executions where only one is
- *     affirmed is what the recomposition removed.
+ *     exactly once. The gate therefore does NOT invoke `test:leaks`; it affirms the
+ *     budget tests by name from the whole-suite `--json` report. Keep that in mind
+ *     before adding a second lane that selects this project: two executions where
+ *     only one is affirmed is what the recomposition removed.
+ *
+ *     NO FILE COUNT IS QUOTED (issue #3772). This note used to end "measured via
+ *     `jest --listTests`: 28 files, no duplicates", repeating a number
+ *     `scripts/agent-gate.sh` carried in four more places; the suite grew and every
+ *     copy went stale, so the sentence asserted a false measurement in the one place
+ *     a reader checks this composition against. Nothing here needs the count: the
+ *     `node-bindings` gate component DERIVES it every run from two independent
+ *     oracles and prints `suite set RECONCILED: N`.
+ *
+ *     The "no duplicates" half was worse than stale — it named an oracle that cannot
+ *     see the violation. Measured on jest 29.7.0 (#3772), two projects both matching
+ *     one file: `jest --listTests` prints it ONCE while the run reports
+ *     `Test Suites: 2 passed, 2 total`. Deleting the `testPathIgnorePatterns` entry
+ *     below leaves this package's `--listTests` output unchanged, too. What actually
+ *     catches a double execution is the gate comparing jest's reported suite TOTAL
+ *     against the DEDUPLICATED disk inventory, plus the leak affirmation's refusal on
+ *     two suites at the leak path. So: if you change the projects' `testMatch` or the
+ *     ignore list, do not reason about it from `--listTests` — run the suite and read
+ *     the total. If you need today's file count, run `--listTests`; do not write the
+ *     answer down here.
  *
  * WHY NO `detectOpenHandles`/`detectLeaks` ANYWHERE (measured on jest 29.7.0, not
  * assumed):
--- a/scripts/agent-gate.sh
+++ b/scripts/agent-gate.sh
@@ -1894,13 +1894,13 @@ _fixture_status() {
 # leak budgets exactly once, and the second invocation was pure duplication.
 #
 # NO FILE COUNT IS QUOTED HERE, DELIBERATELY (issue #3772). This argument used to rest
-# on a one-off measurement -- "the all-projects list is 28 files with NO duplicates" --
-# and the suite then grew past it while the sentence stayed at 28. A stale number inside
-# the rationale for a composition decision is the failure mode CLAUDE.md names: a false
+# on a one-off measurement -- a hard-coded file count, asserted as "no duplicates" -- and
+# the suite then grew past it while the sentence stood still. A stale number inside the
+# rationale for a composition decision is the failure mode CLAUDE.md names: a false
 # statement in a gate log (or in the comment a reader checks it against) is worse than
-# none, because it is what stops the next person looking. Replacing 28 with today's
-# count would only restart the same clock -- which is why no number, not even the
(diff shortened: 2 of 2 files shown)


pmcfadin:
## Review rounds — three findings, all mine, all self-inflicted by the fix itself

Recording these because the pattern is the point: **every round found this issue's own defect class reproduced one layer inward.** #3772 is about a comment asserting something nobody re-measured; each of my fix rounds asserted something *I* had not re-checked.

| round | job | sha | verdict | finding |
|---|---|---|---|---|
| 1 | 42 | `6eb27b451` | 1 Medium | removed `28` as a live claim but **quoted it verbatim** in the three sentences explaining the removal — including the census, which is *runtime output*. Every gate run still printed `28`, under a line saying no count is quoted. |
| 2 | 44 | `bfdf4e627` | 1 Medium | two more **live** counts in the same component — `"agrees at 27 today"`, `"27 suites of assertions"`. Fixed some sites and not others. |
| 3 | 46 | `9b31c046f` | 2 Low | (a) the header cited **`--listTests` as the evidence** that the leak file runs once — while the paragraph below it explains `--listTests` *dedupes and cannot see that*. (b) the census claimed both jest projects running was `DERIVED`; **nothing checks project identity.** |

All reviews were genuine, not vacuous — 129k/192k/225k input, 95k/144k/186k cached, every deterministic key PASS (`prompt-content: PASS (2/2)`, both vacuity tiers PASS).

### What changed in response, beyond the named lines

Round 2 made me stop patching named sites and sweep the component, which surfaced a distinction I had been applying **implicitly and never stated** — the reason each round kept finding another one:

- a claim about **what is true now** (`"the list is N files"`, `"they agree at N today"`) decays the moment a test file is added → derive it at run time, never write it down;
- a claim about **what was measured then, attributed to the change that measured it** (#1255 narrowed to 1 of 27; #3522 measured 504 passing tests across 27/27 at ~15–35 s) is a dated record, true however the suite grows.

Three counts of the second kind are **deliberately kept**, with that line written beside them and an operational test for the next writer: *if adding one test file would make the sentence false, it is the first kind.* Without it, someone "finishing the job" deletes the evidence for the widening decision instead of refreshing it.

Round 3 then made me re-check **every** claim the diff adds against the code rather than fix the two reported lines:

- `suite_n` comes from `disk_set` (the deduplicated inventory) and is what reaches `check_jest_suites_ran` — so "expected is the deduplicated inventory" holds;
- `suites.length !== 1` exists and refuses (`agent-gate.sh:2097`);
- **zero** executions fail too, via the affirmation's no-matching-suite arm — so "zero or twice" is accurate on both guards, not just the twice half;
- "printed below" is true of the emit order (census ~10875, `RECONCILED` line ~11140);
- no `displayName` / `selectProjects` / `projects[]` read exists anywhere in the gate — which is *why* the both-projects claim had to be downgraded to "inferred" rather than reworded.

Verified mechanically: `git diff origin/main` introduces **no** occurrence of `28`.

PR #3842, opened 2026-09-01
fix(#3752): bind the merge to the roborev round, not just the gate — a rebase voids a certification

pmcfadin:
Closes #3752

## The hazard

`premerge-assert.sh` bound the merge to the **gate of record** at the certified head, and to nothing else. Nothing bound it to the **roborev round**. So a rebase rewrites the reviewed commit and a PR can truthfully record "roborev: PASS" about a commit that no longer exists on the branch being merged — measured on PR #3735, job 304 at `d3812f59`, which `git cat-file -t` reports as not a valid object after the lane rebased. Two unreviewed commits sat after the reviewed content, one of them the most review-worthy commit on the branch: a semantic rebase-conflict fix in the one file overlapping `main`.

**The rule, now mechanical: ROBOREV LAST.** `rebase → gate of record → C → roborev → premerge-assert → arm`. The asymmetry is the reason: **a roborev round changes no bytes, so reviewing after gating costs nothing and cannot invalidate a gate PASS; a rebase changes bytes, so gating or reviewing before it certifies the wrong tree.** Review-after-gate is free; gate-after-review is not.

## What lands

Two fail-closed legs, implemented in a sourced helper (`scripts/flow/premerge-review-binding.sh`) plus a structural `gh --json` thread scanner (`scripts/flow/premerge-pr-scan.py`), wired into `premerge-assert.sh` without changing its positional contract:

- **`review-binding`** — derives the reviewed head from the **job record** (`roborev show <id> --json`, whose `git_ref` is `<base40>..<head40>`), never from the `Enqueued job <N> for <sha>` stdout line, which for a range review names only the base. `git merge-base --is-ancestor` is the verdict and runs **FIRST**; `cat-file -t` is a why-diagnostic only. **Both** `git_ref` halves bind: the reviewed base must be the merge-base, or the omitted prefix must be code-free. Reviewable code added after the review (by `classify-docs-only.sh`, never a path shape) refuses the merge.
- **`hold-check`** — re-reads the PR and the issue threads it closes for a column-zero-anchored `HOLD:` order, and the PR timeline (**fully paginated**) for a lead disarm inside a committed 30-minute window. A missing/unreadable signal is `UNMEASURED`, read as HOLD.

Verdict tokens are a closed set matched token-exactly; every could-not-measure state is non-passing. AC4's self-check is printed with the real shas substituted, in the corrected order.

[...]

PR #3885, opened 2026-09-01
fix(#3751): a delegated review stage's verdict is an artifact — an absent review must never read as a clean one

### LINKED ISSUE #3751: four read-only review subagents idled without reporting in ONE delivery — an absent review must never read as a clean one
## The observation

**Four read-only review subagents idled without producing a report inside a SINGLE delivery** (`#3629`), and at least three other lanes reported the same class today:

| lane | agent | outcome |
|---|---|---|
| `lane-3629` | `spec-auditor` | spawned, idled, **no report** |
| `lane-3629` | `rust-reviewer` | spawned **twice**, idled both times |
| `lane-3515` | `rust-reviewer` | spawned, **idled two hours**, no report |
| `lane-3629` | (earlier, named in thread) | same shape |

Every one was **recorded as not-run** by the lane, and its highest-value checks were done **by hand**. In two cases the hand check **found a real defect** the idle agent would have been credited for (`lane-3515` found an anchor conflating two measurements and loosening its own guard ~14×).

## Why this is a correctness problem and not an annoyance

**A review stage that silently never reports is the vacuous-green shape this repository's doctrine is written against, wearing an agent's clothes.**

The gate has `RESULT: INCOMPLETE` as an explicit launch sentinel precisely so that "did not finish" can never be read as a verdict (#3041). The roborev wrapper has `review-completed`, `vacuity-tier1/2` and a token-accounting baseline precisely so a review that received nothing cannot pass as one (#3312). **A spawned `spec-auditor` or `rust-reviewer` that idles has neither.** It produces no artifact at all — and "no artifact" is indistinguishable, to any automated consumer, from "not yet finished".

The only reason this has not yet produced a false certification is that **three separate lanes independently chose to record the stage as not-run rather than as clean.** That is a discipline outcome, not a mechanism outcome, and it is the exact failure mode the affirmative-measurement rule exists to remove: *never derive a pass from the ABSENCE of a bad signal.*

`flow-closer` treats C as a gating stage (`NEEDS-SPAWN {role: spec-auditor}` → lead spawns → closer re-invokes with the verdict). If a closer ever reads "no findings reported" as "PASS", the stage certifies nothing while appearing to certify.

## Acceptance criteria

1. **A review subagent that produces no report is reported as `NOT-RUN`, never as clean or empty-findings.** Asserted by a test that spawns a stage which returns nothing and checks the consuming verdict is non-passing.
2. **A spawn has a deadline and the deadline is visible.** An agent past it is reported with its elapsed time and the fact that it produced nothing — the way a gate prints `waiting for gate slot` rather than looking hung.
3. **`flow-closer`'s C stage fails closed on a missing report.** It must not be possible for an absent `spec-auditor` verdict to reach a merge.
4. **A hand-performed substitute is recorded as author-performed.** `lane-3629`'s wording is the model and should be the required form: *"an author's hand audit is not an independent one; weight it accordingly."*
5. Root cause investigated: whether this is model access (the pinned frontmatter model being inaccessible — see the standing "always pass an explicit accessible `model` on spawn" rule), a tool-permission failure, a watchdog kill, or a harness-level stall. **The four instances share a fleet and an AMI, so a single environmental cause is likelier than four coincidences** — and per the corroboration rule, four confirmations from one environment are one measurement.

## Notes

- The affected agents are all **read-only** (`spec-auditor`, `rust-reviewer`, `coverage-reviewer` are `Read/Grep/Glob/Bash`-only). No work was lost; the cost is unreviewed diffs and lane time spent hand-substituting.
- Workaround in use, and it is the right one until this is fixed: **record the stage as not-run, perform its highest-value check by hand, and label the substitute as author-performed in the PR body.** A missing review reported as missing is worth more than a green that never ran.
- Related: the standing convention that implementers **commit after each meaningful unit of work** exists because a subagent starved of CPU is killed by the 600s stall watchdog and loses uncommitted work. If these idles are watchdog kills, that mechanism is the first place to look — but note the read-only agents have nothing to lose to a kill, which is why the failure presents as silence rather than as lost work.

pmcfadin:
Closes #3751.

## What this is

A delegated review stage used to write **NOTHING at any point in its life**, so its reader had only
ABSENCE to reason from — and every consumer of an absence has to CHOOSE how to read it. Nine measured
spawns across five lanes and four agent types produced no report; nine for nine the lanes recorded "not
run" and disclosed it, **which is discipline, not mechanism**.

`scripts/flow/review-stage.sh` transplants #3041's idiom. `open <kind> --issue <N> --agent <type>`
creates the report-of-record file **before the agent is spawned**, carrying a non-verdict sentinel, and
prints the absolute path plus **a paste-ready clause to put in the spawn prompt verbatim** — the
paraphrase is what varied across the measured sessions. `verdict` then emits ONE line of a CLOSED
grammar — `{PASS, FINDINGS, NOT-RUN, AUTHOR-PERFORMED}`, first word, string equality, never a prefix
test (#3544) — exiting `0/4/5/6`. `status` is advisory and is never a verdict input. `NOT-RUN` names one
of **six** causes, because the operator action differs per cause.

`scripts/flow/premerge-assert.sh --c-verdict <path|AUTO>` is the consumer, at the one point every merge
passes through. It is **REQUIRED with no default — omitting it is exit 3** (#3465's precedent: a silent
"C is not required" would reproduce, inside the enforcer, the exact defect the enforcer closes). `AUTO`
**measures** routing from the certified tree rather than trusting the caller, `UNMEASURED` is treated as
REQUIRED, and only `PASS` and `AUTHOR-PERFORMED` proceed — the second under its own
`PREMERGE: C-VERDICT` token, never folded into `PREMERGE: OK`.

## Wiring evidence — the mechanism was used on its own diff

The independent reviewer of this branch was spawned with `open`'s paste-ready clause verbatim, wrote its
report incrementally, and recorded a verdict that I read from **disk**, not from its returned message:

```
REVIEW-STAGE: code-review RESULT: FINDINGS elapsed=792 deadline=1800 agent=general-purpose
  report=/data/lanes/lane-3751/.review-stage/issue-3751/code-review.md      (exit 4)
```

Both fix rounds' implementers reported the same way (`fix-round-{1,2,3}.md`, each `result: PASS`).

## Review-first found 21 blockers before the first full gate

Per #2086 the review ran on the lite-green diff **before** any full gate. It found 21 blockers, which is
the argument for the ordering. The ones worth reading:

- **A peer lane's C PASS could certify this PR.** `AUTO` located the C stage in the current worktree and
  nothing bound it to the certified sha. Every lane here is a worktree of ONE shared `.git`, so a peer's
  certified commit resolves anywhere — the #3616 class, inside the file written to refuse it. Fixed in
  two axes, because my first fix closed only one: `HEAD == certified` binds the **worktree**, and the
  stage record's own `head-sha:` binds the **artifact** (a `PASS` recorded before a later commit, amend
  or rebase must not certify the new tree).
- **The verdict parser checked neither grammar nor stage KIND** — any anchored
  `REVIEW-STAGE: <anything> RESULT: PASS` was accepted. Confirmed live: this branch's own `code-review`
  stage line would have satisfied `--c-verdict`. Now the full documented grammar is validated, the kind
  by string equality, each mandatory field exactly once, against a shape captured from the shipped
  emitter so parser and emitter cannot drift.
- **The sentinel template `open` writes was its own escape hatch.** It carried indented `result: PASS`
  example lines and the classifier matched indented lines, so deleting the column-zero sentinel and
  appending `result: FINDINGS` returned **PASS, exit 0** — reproduced by hand, twice. #3312's family
  inside the mechanism that cites it.
- **`record-author-performed` silently overwrote a recorded `FINDINGS`** with `AUTHOR-PERFORMED`, turning
  a refusal into a proceed with no `--force` and no trace, while `open` refused to clobber without one.
- **A real `openspec archive` move routed as C-REQUIRED** (renames pinned off ⇒ the live-dir deletions
  were counted). A false refusal on correct, doctrine-mandated input is a blocker here: it is the guard
  agents learn to waive.
- **`set -- $value` globbed**, so a `result: *` report plus a file named `PASS` in the cwd read PASS — a
  verdict that was a function of the cwd listing.
- **`die_usage` inside a command substitution** exited only the subshell, so the documented exit 64 never
  fired and the emitted line reported a **fabricated** report path.
- **A TOCTOU on the temp write path**: a predictable name validated then reopened by redirection, so a
  symlink planted in the window clobbered its target and `mv` could install it as the report. The window
  is **removed**, not narrowed (unpredictable `O_EXCL` create-and-open, write through a held descriptor).

### One seam produced a finding twice, so it was consolidated rather than patched

Round 2 found `classify_report` not column-zero anchored while `premerge-assert.sh`'s reader was; round 3
found the reverse asymmetry — the other reader counts its anchored lines and refuses several, while
`classify_report` first-won, so a stale `result: PASS` followed by an appended `result: FINDINGS` read
PASS. Two readers of one shape disagreeing twice is a consolidation signal. Both now require **exactly
one** column-zero record, each names the other in a comment, and **a differential test drives both over
one shared 8-shape adversarial table** asserting they agree AND reach the expected disposition — because
a second implementation's correctness is only knowable by testing it against the first. That test found
a divergence (`fenced`) the finding did not name.

## Tests

`scripts/tests/test_review_stage.sh` **285 passed / 0 failed (floor 285 exact)` and
`scripts/tests/test_premerge_assert.sh` **307 passed / 0 failed (floor 301, preserving its documented
+6 host-gated margin)** — both run by the lead from disk, not taken from a subagent's packet. Every fix
in all three rounds is RED-verified: the test was observed failing before the fix. `test_review_stage.sh`
is enrolled in the `tooling-tests` roster **above** the `python3` SKIP so it cannot pass vacuously.

## Scope notes

- **The claim is about the CONSUMER, not about the agents**, and stating it narrowly is the point: naming
  a report path rescued `spec-auditor` and `flow-closer` in measured sessions and did **nothing** for
  `rust-reviewer` (0 of 3, one told IN WRITING that an absent file would be recorded as a non-review).
  The mechanical reason surfaced while writing the clause: `rust-reviewer` had **no write channel at all**
  (`Read, Glob, Grep`), so the contract was unsatisfiable by construction. It now carries `Write` for
  that one purpose — which grants nothing its siblings lacked, since three of the four "read-only"
  reviewers already carry `Bash`. **"Read-only" here was always prose, never a mechanism.**
- `verdict` establishes that a VERDICT WAS RECORDED, never that a review was PERFORMED — a report whose
  only content is `result: PASS` reads as PASS. Declared, not hidden.
- All six pipeline-gating agent definitions carry the clause (Q1=(a) as ruled): the class is *spawns
  whose silence gates a merge*, so `flow-closer` (which owns the merge) and `sstable-developer` (which
  had queued work it never did) are in it beside the four reviewers.
- **Landing coordination with #3752 measured, not assumed** (task 6.1): PR #3842 adds **no**
  caller-facing flag — it calls `premerge-review-binding.sh` from *inside* `premerge-assert.sh` — so the
  arity change is `--c-verdict` alone and the two compose in **either landing order**, satisfying the
  owner's rollout condition that an in-flight lane pays ONE re-certification visit. Residual: six shared
  files, so whichever merges second rebases.
- Full record, with the census, the tally and the limits: `docs/development/review-stage-reporting.md`.

## Disclosures

- The diff is **Rust-free** (17 `.md`, 5 `.sh`, `.gitignore`), so `rust-reviewer` had no subject. I opened
  a `rust-review` stage, caught the mislabel within a minute, deleted it **before any spawn**, and
  re-opened `code-review --agent general-purpose`. Nothing consumed the mislabeled stage.
- I propagated a **false tally** into the website earlier in this session ("four lanes and three agent
  types", copied from CLAUDE.md). The census committed in this same PR shows **five** lanes and **four**
  agent types; both doctrine sites are corrected and the census table — the authority — is untouched.
- Task 5.3's **publish verification is post-merge by construction**: the site deploys from `main`, so
  `curl … | grep '<new phrase>'` before the merge can only return 0. It will be run after the merge,
  allowing for the ~3-minute CDN stale window.
- The `roborev-lints`, `fmt` and `clippy` components are green but this diff compiles no Rust; the gate of
  record is still owed and is run once, pre-merge, by `flow-closer`.


pmcfadin:
roborev-waive: prompt-content-absent base=9da235a6cb9c48b78e019a359e8851ad8d007e09 head=e44da6bf53f708de45a659eb9cdc01f2444637b8 job=379 reason=snapshot-delivered diff since the census crossed the inline threshold (round 1 job 317 PASSed 6/6 present at +3378, rounds 2-7 absent at +9511 with nothing else changed); job 379 accounting input=934465 cached=814336 vs vacuous baseline 18700/0, every round 40-80x baseline with reproducible hand-verified findings (38 found, 37 fixed); lead verified base == merge-base(origin/main, head) and head == live PR #3885 head

pmcfadin:
roborev-defer: findings issues=3929 count=1 base=9da235a6cb9c48b78e019a359e8851ad8d007e09 head=e44da6bf53f708de45a659eb9cdc01f2444637b8 job=379 reason=J2 declared not fixed and re-reported verbatim every round (mv -T protects only the destination leaf — the parent-directory-component substitution race is not closable in bash; needs openat(O_NOFOLLOW)+renameat2 in a helper), filed as #3929 (OPEN, P3, with positive-control acceptance criteria and an explicit no-more-bash-parent-checks note); the expressible half (mv -f -T REQUIRED) closed in round 7; seven findings in one path across nine rounds is the #3229 stop-carving shape; lead verified base/head/job as in the waiver

pmcfadin:
## lane-3751 — endgame requested; I am NOT running it as specified. Three blockers, two of them in the commands themselves.

Agreed on the premise: `required: SUCCESS` is **not** certification, this PR carries no full
`==== AGENT-GATE SUMMARY ====`, and auto-merge being unarmed is correct. The gate of record is genuinely
owed. But two of the three commands would fail or produce a void result, and one would merge unreviewed
code — so here is what I verified instead of executing.

### 1. The tree is being edited RIGHT NOW — a gate started now would void on `tree-integrity`

```
$ git status --short
 M CLAUDE.md
 M docs/development/review-stage-reporting.md
```

A fix round (round 18) is mid-flight: it has committed `4c6fdce05` and is still sweeping doctrine. #2926
fails a run closed with `tree-integrity: FAIL (tree-mutated-midrun)`, so this would spend 12–25 minutes of
a shared 7-lane box to produce a block that **cannot be pasted as a certification**. I will start the gate
the moment the tree freezes, not before.

### 2. `--repo /home/ubuntu` is not a git repository at all

```
```

This lane's toplevel is `/data/lanes/lane-3751`. Under 1:1:1:1 every issue is a worktree, and a wrong
`--repo` is the **T1 vacuous-review hazard** the wrapper exists to refuse: `--branch` resolved against the
wrong checkout enqueues the wrong commit and reports clean having reviewed nothing. I will use
`--repo /data/lanes/lane-3751`.

### 3. `AGENT_GATE_SUMMARY_FILE=/tmp/gate-3751.txt` is a fixed name on a 7-lane box

Doctrine requires a `mktemp`-unique summary path per run: a shared `/tmp` name gets contended under
multi-lane load, and reading a peer's artifact is exactly the #3616 escape (a closer once nearly merged on
another PR's verdict). I will use `$(mktemp /tmp/gate-3751-XXXXXX.txt)` and bind the result by its
`run-id:` plus `commit:`/`tree-start:`.

### And the merge is blocked on something a gate cannot fix

Even with a green gate, `premerge-assert` will refuse and **should**:

- **PR head is `c1a0faa64`; local HEAD is `4c6fdce05`** (round 18's fix, unpushed). Head ≠ certified sha.
- **`findings: PRESENT (1)`** at job 397 — not clean, and `findings: NONE` is explicitly unwaivable.
- **`prompt-content: FAIL (9/9 absent)`** — the census outgrew inline delivery, so the diff is
  snapshot-delivered. Genuine review (`input=1678790 cached=1529856` vs a vacuous baseline of ~18.7k/0).
- **Both authorizations on this PR are STALE**: they name `head=e44da6bf5 job=379`, which is **13 commits
  and 18 review jobs behind**. The wrapper reports it precisely: *"a waiver may not outlive the review its
  authorizer judged."*

**I will not arm `gh pr merge --auto` here.** With `required: SUCCESS` and `autoMergeRequest: null`,
arming *is* merging — immediately, before the gate of record, with roborev not clean and the intent audit
unrecorded. That is the #3408 escape with extra steps, and this branch exists to close it.

### What I am doing

1. Let round 18 land (minutes), verify it from disk, push — head then matches the PR.
2. Spawn `flow-closer` (model opus) for the **ONE full gate of record** on the frozen tree, unique summary
   path, polled on `grep -qE 'RESULT: (PASS|FAIL)'` — `INCOMPLETE` is the launch sentinel, never a verdict.
   If it FAILS I report the failing component by name; I will not substitute a `--lite` block, and I have
   not done so in 17 lite runs.
3. Final roborev with the corrected `--repo`.
4. Then it stops with the lead: the markers must be re-bound to that final job id and observed count.
   A worker may request; only the owner or the coordination lead may grant.

Current state: 19 review rounds, 55 findings, 54 fixed and hand-verified, 1 declared and lead-accepted
(#3929, five manifestations, non-expressible in bash). Suites `review-stage 937/937` and
`premerge-assert 520/520`, both run from disk by me rather than taken from a subagent's packet.

pmcfadin:
## lane-3751 — steps 1–2 done, with two corrections. Gate deliberately discarded and re-ordered.

**You were right on the merits about the stale review and the stale base, and I acted on both. Two of
the five steps as written would still have failed, and one had already been reported.**

### Step 1 — DONE, by merge rather than rebase, and only after measuring that it mattered

I did not take "needs its branch current" on trust, because branch protection says otherwise:

```
$ gh api repos/pmcfadin/cqlite/branches/main/protection --jq '{strict:...}'
{"contexts":["required"],"strict":false}
```

`strict: false` means up-to-date is **not** a merge requirement, and #3650 slice 1 makes base staleness an
**advisory that cannot change a verdict**. So I measured whether it mattered here:

```
BASE-STALENESS: verdict STALE-RECOGNISED
BASE-STALENESS: verdict-detail 10 of the 33 commits behind touch this diff's blast radius.
```

**10 of 33 in the blast radius — so it did matter**, and for a concrete reason: main had moved on
`CLAUDE.md`, four `scripts/flow/*.sh` (including `gate-detached.sh` itself) and a batch of
`scripts/tests/*.sh`. `scripts/tests/**` is gate-global content that `tooling-tests` **executes**, so
those commits can genuinely move a gate verdict. Merged (`git merge-tree --write-tree` probed clean first,
rc 0, zero conflicts), both suites re-run unchanged (`review-stage 948/948`, `premerge-assert 537/537`),
pushed. **HEAD `20d3af531`, zero commits behind main.**

I used **merge, not rebase** — you sanctioned either. Rebasing 82 commits across 33 upstream ones invites
a conflict-resolution pass per commit; one merge commit resolves it once and keeps the review history the
18 roborev jobs are anchored to.

### I discarded a 95-minute gate of record to do this, deliberately

A full gate was **24/37 components in, 0 FAIL, 0 SKIP** when your message arrived. I stopped it, because
your ordering is correct: it was certifying a tree that will not be what merges — #3650's own
`PREMERGE: SCOPE` caveat, live. I probed the merge non-destructively **before** discarding the work.

Not wasted as information: 24 components passed on this code, including `python-bindings` and
`node-bindings` with real toolchains rather than SKIPs, and `file-size` clean despite substantial growth.
Recorded as a **discarded, non-certifying** run so it can never be mistaken for a gate of record.
`tooling-tests` — the only component that genuinely exercises this diff — had not started, so nothing is
known about it either way.

### Step 2 — RUNNING, with `--repo` corrected for the second time

```
scripts/flow/roborev-review.sh --agent codex --model gpt-5.6-sol \
    --repo /data/lanes/lane-3751 --base origin/main
```

**`--repo /home/ubuntu` is not a git repository** — `fatal: not a git repository (or any of the parent
directories): .git`. I reported this two hours ago in
[#3885 comment](https://github.com/pmcfadin/cqlite/pull/3885#issuecomment-5511789422) and the same value
came back in this instruction, so I am flagging it rather than silently substituting again: under 1:1:1:1
every issue lives in a worktree, and a wrong `--repo` is the **T1 vacuous-review hazard** — `--branch`
resolves against the wrong checkout, enqueues the wrong commit, and reports clean having reviewed nothing.
That is the failure this wrapper exists to refuse, and it is textually identical to a genuine pass.

### Step 3 — deferred until roborev settles, and that ordering is the lesson

Gating before the review is what just cost 95 minutes: any finding means new commits, which voids the
gate. Review-first is doctrine (#2086) for exactly this reason. One gate, on the final tree, once.

Two notes on the step-3 recipe: `AGENT_GATE_SUMMARY_FILE` is **silently dropped** by
`scripts/flow/gate-detached.sh` (line 312) in favour of its own `--summary`, so that env var is a no-op on
the detached path and the authoritative summary is the path the launcher **prints** — filed as **#3962**
with the `RUSTFLAGS=-D warnings` contamination I also found in the lane environment. And a fixed
`/tmp/gate-3751-record.txt` on a 7-lane box is the #3616 peer-artifact class; I use a nonced path and bind
by `run-id:` plus `commit:`/`tree-start:`.

### Step 5 — I will not arm, and this is not reluctance

`required` is **already `SUCCESS`** with `autoMergeRequest: null`, so arming *is* merging, instantly.
Beyond the stale review you correctly identified, two blockers remain that no gate can clear:

- **`findings: NONE` is explicitly unwaivable**, and the declared residual (#3929, lead-accepted, five
  manifestations, not expressible in bash) is re-reported by every round — the #3626 deadlock.
- **Both authorizations on this PR are STALE**: they name `head=e44da6bf5 job=379`, now ~20 commits and 18
  review jobs behind. Only the owner or the coordination lead may re-bind them; a worker may request.

So the end state I can reach alone is: roborev settled, ONE gate of record PASS at the final sha, both
blocks in the PR body, and a request for re-binding. The merge itself needs a human, by design — and
"do not arm on the stale review" is the same instinct that makes me decline to arm on stale
authorizations.

**State:** 19 review rounds, 55 findings, 54 fixed and hand-verified, 1 declared and lead-accepted. C is
owed and not yet spawned — it must anchor to the sha that finally merges.

pmcfadin:
## lane-3751 — the ordering rule is satisfied. Gate of record RUNNING on a frozen tree. One human action left.

**Ordering rule adopted and followed.** It named both of my own errors today, so it landed: I had gated
before the review settled (95 minutes discarded) and I had committed *during* a run.

### Steps 1–2 complete

- **HEAD `2f859d332b56634e97ed474f8edefd2b19d0a6cb`**, pushed, **0 commits behind `origin/main`**, tree
  clean. Main moved three separate times while I worked (33, then 2, then 2); each time I probed
  `git merge-tree --write-tree` before merging, and each merge was clean. I stopped chasing after the last
  one because the two commits at issue touched `CLAUDE.md` (in my diff) and `scripts/tests/**`
  (gate-global, executed by `tooling-tests`) — measured, not assumed.
- Suites, run from disk by me after every merge: `test_review_stage.sh` **1102/1102** (floor 1102, exact),
  `test_premerge_assert.sh` **579/579** (floor 573, host-gated margin preserved).
- **roborev is clean in the only sense available here, corroborated twice.** Jobs 418 and 419, independent
  runs on an identical census (27 files, +23637/-82), both report **exactly one finding at exactly the same
  location** — `review-stage.sh:1751` — which is the seventh manifestation of the lead-accepted declared
  residual **#3929** (`lock_stage` checks for symlinks, then `exec 8>` re-resolves; closable only with
  `openat`/`O_NOFOLLOW`). No other finding remains.

**Disclosure on that evidence, because you should weigh it rather than take my word.** The token
accounting declined sharply across the last four rounds: `3.41M → 3.34M → 394k → 219k` input. The final
two sit below doctrine's documented genuine band (398k–649k), though far above the vacuous baseline
(~18.7k/0), and both passed `review-completed` and both vacuity tiers. What persuades me is not the
volume but the **agreement**: two independent runs on identical input converging on the same single
defect, in code the preceding round had just written. I ran the second round *specifically* because the
first one's accounting looked thin — I was not willing to premise your authorization on the weakest round
of the sequence.

### Step 3 — RUNNING, tree frozen

```
unit:          cqlite-gate-20260902T194353Z-2403736   (own cgroup; survives session teardown)
summary:       /tmp/gate-3751-arm-sF0KNX.txt
run-id:        /tmp/agent-gate.96rlyJ
certified sha: 2f859d332b56634e97ed474f8edefd2b19d0a6cb
```

I will not commit, merge or touch a tracked file until it reports. Two notes for the fleet, since both bit
me: `gate-detached.sh` **silently drops** `AGENT_GATE_SUMMARY_FILE` in favour of its own `--summary`, so
the authoritative path is the one the launcher **prints** — and this lane's environment carries
`RUSTFLAGS=-D warnings`, which CLAUDE.md records as having "reddened a clean tree and halted the fleet on
a P0 that did not exist". I launched under `env -u` and verified **0** occurrences in the unit's
environment. Both are filed as **#3962**.

### Step 4 — I will paste both blocks on completion, and only if they qualify

`RESULT: PASS` **and** `tree-integrity: PASS` **and** `commit:`/`tree-start:` covering
`2f859d332`. If the gate FAILs I report the failing component by name; if it reds on something
pre-existing, note that `classify-docs-only.sh` exits 1 on this branch, so no prose waiver applies and a
failure is presumed ours.

### Step 5 — blocked on ONE human action, and two mechanical corrections

**@pmcfadin / coordination lead — both granted markers are stale on all three axes and need re-binding
once. The values are now final and will not move again:**

```
base = 9bb2cc571f48830d205ca5aa11cbccd6d20165a6      (merge-base; moved twice via the merges above)
head = 2f859d332b56634e97ed474f8edefd2b19d0a6cb      (final; tree frozen)
job  = 419                                            (the corroborating round)
count = 1        issues = 3929                        (the sole finding = the accepted residual)
```

Your originals named `base=9da235a6c head=e44da6bf5 job=379 count=1` — correct in substance, and `count=1`
and `issues=3929` are still exactly right. Only the three identity fields moved. Applied with
`--recheck-job 419`, which enqueues nothing and touches no file, so it cannot disturb the frozen tree.
A worker may request; only you may grant.

**And two corrections to step 5's command, the second of which will hit every other lane:**

1. `premerge-assert.sh 3885 <sha>` **exits 3** — I ran it exactly as written:
   `PREMERGE: USAGE — expected 3 or 4 positional arguments, got 2`. The gate-of-record summary became
   required in #3465, and `--c-verdict` is required by **this PR**. Correct call:
   `premerge-assert.sh 3885 <sha> /tmp/gate-3751-arm-sF0KNX.txt --c-verdict AUTO`.
   **Worth propagating with the ordering rule to those 15 lanes** — they will all hit this the moment
   this merges.
2. This change is **design-routed** (approved OpenSpec change `absent-review-must-not-read-clean`), so
   `--c-verdict AUTO` will measure **C REQUIRED**. The intent audit is genuinely owed and must anchor to
   this sha; I will sequence `spec-auditor` against the frozen tree rather than hand `AUTO` a verdict it
   would refuse.

**I will not arm before all of that holds.** `required` is already `SUCCESS` with
`autoMergeRequest: null`, so arming *is* merging — instantly, and irreversibly.

State: 23 review rounds, 60 findings, 59 fixed and hand-verified, 1 declared and lead-accepted.

pmcfadin:
## lane-3751 — gate of record RAN and FAILED. One failure was mine (fixed); the other is `main`, and it blocks every lane.

**I am reporting the failing component by name, as instructed. No `--lite` block is being offered in its
place, and none has been at any point in this lane.**

### The gate of record — a valid run, and a real FAIL

```
run-id:          /tmp/agent-gate.96rlyJ
summary:         /tmp/gate-3751-arm-sF0KNX.txt
commit:          2f859d3   dirty: no
tree-start:      2f859d332b56   tree-end: 2f859d332b56   digest unchanged
tree-integrity:  PASS
component-set:   PASS (37/37 names vs origin/main 978ddf4b5)
cpu-budget:      max-concurrency=1(pinned)   accelerators: sccache=on nextest=on mold=linked perf=ok
RESULT:          FAIL

tooling-tests:   FAIL (2699s)      <- the ONLY non-PASS. 36 of 37 components PASS, 0 SKIP.
```

The run itself is sound: `tree-integrity: PASS`, `dirty: no`, `commit:` equals the HEAD I certified, and
the tree digest is identical at start and end — the freeze held. `RUSTFLAGS` was verified absent from the
unit's environment. So this is a genuine verdict, not an artifact.

`tooling-tests` at 2699s is **not** truncated — it exceeded the 2073s prior measurement and ran its full
roster. My own suites passed **inside** it: `review-stage 1102/1102`.

### Two assertions failed. I split them by reproducing on a clean checkout rather than presuming.

```
FAIL - P2: progress line(s) still printing the PRE-census status at line(s): 16089
FAIL - R1: three-token component-status alternation(s) survive (VACUOUS omitted) in
           scripts/tests/test_check_dataset_manifest.sh
```

```bash
git worktree add --detach /tmp/clean-main origin/main && cd /tmp/clean-main
bash scripts/tests/test_agent_gate_census.sh
#   -> component census guard: 118 passed, 1 failed      (R1 only)
```

- **P2 was MINE.** My roster enrolment added a progress line printing the pre-census `$status`; main has
  since rewritten ~115 sites to print the finalized `$RECORDED_STATUS`, and I copied the old pattern.
  **Fixed** in `9338d28e6` — one line. My branch now reports **118 passed, 1 failed: byte-identical to
  clean main.**
- **R1 is `main`'s**, reproduced on a clean `origin/main` worktree with nothing of this branch present.
  Filed as **#3977 (P0)**.

### #3977 blocks every lane, not just me

`bdaf2b6e1` (#3625/#3916) added the census guard requiring `VACUOUS` in three-token component-status
alternations; a later commit grew `test_check_dataset_manifest.sh` by ~558 lines with alternations that
omit it. **Each side passes alone; only the combination reds — and nothing gates the combination.** Since
that guard is enrolled in `tooling-tests`, the consequence is that **no lane can currently obtain a
passing gate of record**, at ~45 minutes of a serialised slot per discovery. It is not waivable:
`classify-docs-only.sh` exits non-zero for any branch touching executed inputs, and a real failing
assertion is not a flake.

You reported that 15 lanes held a passing gate and none could be armed. This is worse: new gates cannot
reach a PASS at all.

### What I need, and it is a scope decision rather than a technical one

**I did not fix #3977 in this PR.** It is another issue's file; the correct remedy is a judgement about
whose vocabulary is whose (the guard explicitly permits "extend that alternation so it is not the bare
three", which suggests the dataset-manifest status vocabulary may simply not be the gate's, in which case
the *guard's* subject set wants narrowing); and folding a fleet-blocking fix into an unrelated PR puts it
behind this PR's own queue.

**Tell me which you want:**

- **(a) I take #3977** in its own small PR, land it, then re-gate here. Fastest to a green fleet; I already
  have the reproduction and the two candidate readings.
- **(b) Its owner takes it** and I re-gate when it lands. Correct by ownership, slower, and 15+ lanes stay
  blocked meanwhile.

Either way this PR needs one more gate afterwards, and I will not re-gate until #3977 is on main — a gate
now would burn 45 minutes to rediscover a defect I have already filed.

### Unchanged and still outstanding

Head is now `9338d28e6` (P2's fix; 3 commits behind main again, and I will merge before the next gate).
The marker re-binding request stands — values will be restated at the final sha, since it has moved. C is
owed and unspawned. I have not armed and will not: `required` is `SUCCESS` with
`autoMergeRequest: null`, so arming is merging.

State: 23 review rounds, 60 findings, 59 fixed and hand-verified, 1 declared and lead-accepted (#3929).

pmcfadin:
## lane-3751 — advisory 6 is UNEXECUTABLE for immune lanes, and it is arithmetic rather than bad luck. Costs 0s to confirm.

I followed advisory 6 exactly: box `6-169` load **0.28**, not 5-53, immune re-verified at the frozen head,
no prior gate PASS (I checked — all six held `RESULT: PASS` files are `AGENT-GATE LITE SUMMARY`, which
`premerge-assert` refuses by name). Launched detached, background, summary-file redirect, `RUSTFLAGS`
scrubbed.

**The gate FAILED at pre-flight in 0 seconds:**

```
preflight:     FAIL (component-set BEHIND origin/main — this gate script predates components it does not run)
component-set: FAIL-CLOSED (#3544) — this tree is BEHIND origin/main 13e302fa2;
               1 of its 38 component(s) MISSING from this gate script: dep-duplicates;
               a PASS here would certify a set that is no longer the coverage claim
tree-integrity: PASS      commit: feb955023      dirty: no      RESULT: FAIL
```

### The structural problem, in one line

```
$ git merge-base --is-ancestor 035fdbf32 origin/main   ->  rc 0     # main CARRIES the P0 guard
```

So **immune to #4029 ⟺ 035fdbf32 not an ancestor ⟺ my tree is behind main's tip ⟹ #3544's component-set
pre-flight fails closed.** Immunity and certifiability are the same axis pointing opposite ways.
**No immune lane can obtain a gate PASS until it merges** — which is the one thing advisories 4, 5 and 6
all forbid. Advisory 6's "You will land BEFORE #4039 and that is fine and desirable" is not reachable by
any immune lane, and this is not specific to my branch: it will reproduce for every one of them, in 0
seconds, against whatever component main most recently added.

**Advisory 2 was right after all — but for a reason none of us identified.** Not because #4029 would red
an immune tree (it provably cannot: the guard file is absent, so case 17.5 cannot exist), but because
#3544 refuses to certify a gate script that predates main's component set. The hold was correct; the
stated reason was wrong, which is why loosening it looked safe.

### The good news, and it is real

**It costs 0 seconds, not 2400.** The pre-flight refuses *before* the #1825 slot and before any
component, so no immune lane following advisory 6 can start a stampede or burn a core. #3544 is doing
precisely what it was built for — "a gate script BEHIND `origin/main` cannot certify" — and it caught this
at the cheapest possible moment. Anyone can reproduce in one command; nobody needs to wait 45 minutes to
find out.

### Why I am not fixing it myself, though I checked whether I could

The missing component is `dep-duplicates`, from `3621a7bff` (#1700, PR #3839 — which you reported merged).
I looked at cherry-picking just that, as I did earlier today for #3977's one-file fix:

```
Cargo.lock                            | 373 +----
cqlite-cli/Cargo.toml                 |   1 -
scripts/agent-gate.components         |   1 +
scripts/agent-gate.sh                 | 267 +++-
scripts/ci/check-dep-duplicates.sh    | 825 ++++++++++++
scripts/ci/dep-duplicates-baseline.txt|  79 ++
```

That is a **dependency-graph change plus a new gate component** — a `Cargo.lock` rewrite and 267 lines of
`agent-gate.sh`. Cherry-picking it would put another issue's dependency removal inside this PR, and it
would not be the last one: the next component main adds re-opens the same gap. The #3977 cherry-pick was
defensible because it was one test file that ended up **byte-identical to main's**, leaving zero diff.
This is not that, and I am not doing it.

### What I think the fleet needs — your call, not mine

1. **Land #4039, then every lane merges once and gates once.** Simplest, and it is what advisory 2
   originally said. Immune lanes gain nothing from immunity because they cannot certify anyway.
2. Or, if landing immune lanes first is genuinely wanted, the component-set pre-flight is the thing that
   must move — and I would not want it relaxed. It is the mechanism that stops a gate certifying a
   coverage claim it no longer implements, which is the same family as everything this PR is about.

**Please correct the advisory** — eight lanes were told to gate now, and every immune one will hit this.
It is cheap, but it will read as "my gate is broken" rather than "the instruction cannot be satisfied,"
and someone under arming pressure may reach for a lite block that says `RESULT: PASS`. That is the failure
I would most like to prevent, and it is exactly what this PR exists to make impossible.

### My state

Frozen at `feb955023`, immune, suites `1102/1102` and `579/579`, no gate of record, **not armed and not
armable**. Reported `red` via `lane-report`. Holding for #4039 — this time for the right reason.
Still owed once a gate can pass: the C intent audit at the final sha, and re-binding both roborev markers.

pmcfadin:
## lane-3751 — advisory 7 accepted, not gating. But instruction 4's pre-granting cannot work, and I would rather you not spend the effort.

Thank you for the retraction and for verifying `3621a7bff` yourself rather than taking my word. Not
gating, not touching `component-set`, not merging main, not editing `COMPONENTS`. Reported `working`.

### Instruction 4: I owe you an authorization request, and pre-granting it now would be wasted

You offered to grant while we wait "so that nothing queues behind me later." I want that too, but the
markers cannot be issued against my current head, and the reason is your own instruction 5:

```
now:                    head feb955023   base 694df0d90
after the mandatory merge:  BOTH change
```

Instruction 5 requires **merge main once** before the gate. A waiver or deferral is bound to
`base` + `head` + `job`, and *all three* move on that merge — the base because the merge-base advances,
the head because the merge commit is new, and the job because I must re-review at the new head for the
review to certify anything current. So a marker granted now reports `STALE` on all three axes the moment
I follow instruction 5. I have watched that happen five times in this lane already; the wrapper names it
precisely each time, and it is the closure problem #3312 job 24 records.

**So the honest sequencing is: markers LAST, not first.** Post-#4039 the order has to be

1. merge main once → confirm `component-set` reads **39/39**
2. re-review at that head → this yields the final `base`, `head`, `job`, and observed count
3. **then** you grant, bound to those four values
4. gate once → `premerge-assert … --c-verdict AUTO` → HOLD re-read → arm

Step 3 genuinely cannot precede step 2. What I can do — and am doing now — is make step 2 as fast and
predictable as possible so your grant is a single action with no back-and-forth.

### What I am doing with the wait instead

- **A roborev round at the current head is running now**, to confirm the finding set is *stable* at
  exactly one finding — the lead-accepted declared residual **#3929** — after this morning's cherry-pick.
  If it comes back 1, then the post-merge round is highly likely to come back 1 too, and I can hand you
  `count=1 issues=3929` with confidence rather than discovering a surprise at the last step.
- **The values you will need, pre-filled except the identities**, so the grant is copy-paste when the
  moment comes:

  ```
  kind:   roborev-waive: prompt-content-absent   AND   roborev-defer: findings
  issues: 3929            (OPEN, lead-accepted, seven manifestations recorded on the issue)
  count:  1               (to be confirmed by the post-merge round)
  base:   <merge-base after the merge>      head: <merge commit>      job: <post-merge job id>
  ```

- Both authorizations are still needed for the same two reasons as before: `prompt-content` FAILs because
  the census outgrew inline delivery (snapshot-delivered; genuine accounting, disclosed), and `findings:`
  can never reach `NONE` because #3929 re-reports every round while `NONE` is unwaivable.

### One correction to my own earlier claim, since it is load-bearing for your data

In my last report I said "7 concurrent gates" on this box during the stampede. That was a `ps | grep`
**process-line** count, which includes each gate's own children — process name is not process identity,
and I cannot now verify the true concurrent-gate figure because the box has drained. The **load of
74.81** was the sound evidence; the 7 was not, and I withdraw it rather than let it feed a #4030
measurement.

### State

Frozen at `feb955023`, immune to #4029 (guard absent, `035fdbf32` not an ancestor), 40 behind main by
design, suites `1102/1102` (floor exact) and `579/579`, **no gate of record and not armable**. C intent
audit still owed at the final sha — this is design-routed, so `--c-verdict AUTO` will measure REQUIRED and
refuse without a recorded verdict, and I will sequence `spec-auditor` against the final tree rather than
hand `AUTO` something it would refuse.

Holding for #4039. Everything above needs no gate slot and no core, so it costs lane-4029 nothing.

pmcfadin:
## lane-3751 — advisory 8's counts corroborated 4 ways, plus the exact grep-able string (I nearly got it wrong the same way)

**Advisory 8 is right on every number.** Verified independently in this lane rather than agreed to:

| source | value |
|---|---|
| my own pre-flight's text | `1 of its **38** component(s) MISSING from this gate script: dep-duplicates` |
| `bash scripts/agent-gate.sh --list` here (stale tree) | **37** |
| `git show origin/main:scripts/agent-gate.components` | **38** |
| `dep-duplicates` position in main's manifest | **34** |

So `38` on main, `37` stale, and advisory 7's "39 of 39" is unobtainable. For the record I did not propagate
it: my gate's artifact said 38 and I quoted it verbatim.

### The confirmation string, verified against a real PASS rather than guessed

I was about to hand the fleet `grep -qE 'component-set: PASS \(38/38'` — which **would not match.** The
real rendering, from one of my own earlier gates taken when this tree *was* current:

```
component-set: PASS (37/37 names vs origin/main 978ddf4b547634a2cda7de5c01d8a7341ed66204; NAMES ONLY — not implementations, and no component is run here)
```

Note the word **`names`** immediately after the fraction. So the correct predicate post-merge is

```
grep -qE 'component-set: PASS \(38/38 names' <summary>
```

That is the same failure mode as advisory 7's wrong number, one level down: **a confirmation string that
looks checkable and silently never matches is worse than no string**, because the lane concludes its merge
did not take and goes looking for a defect that is not there. I caught mine only because I checked it
against an artifact instead of writing it from memory — which is exactly what you did to catch the 39.

### On credit, since you asked to be corrected

Advisory 8 credits the component-set diagnosis to lane-3751 and lane-3752. Accurate for my part, with one
narrowing I would rather state myself: **I diagnosed it for my own lane** — I hit the pre-flight refusal,
measured that `035fdbf32` is on main, and concluded that immunity and certifiability are the same axis
pointing opposite ways. I did not survey other lanes or establish it fleet-wide, and lane-3815's
three-way count verification is what actually pinned the numbers. Please do not credit me with the
general result.

### Not chasing ws0-corpus-gen

Understood and I will not look for an issue. Worth recording that your diagnosis matches this repo's own
doctrine: a determinism byte-compare under 7x oversubscription fails on timing rather than logic, which is
why `tooling-tests` at load 122 tells you nothing about `tooling-tests` at load 0. That is also the
argument for the admission check in advisory 5 being a permanent rule rather than a stampede remedy.

### My state — no change, and no gate slot consumed

Holding at `feb955023`, immune, 40 behind by design, **not gating**, not merging, not touching
`COMPONENTS`. Suites `1102/1102` and `579/579`.

Using the wait exactly as instruction 4 says. One result already: the roborev round I ran to check
finding-set stability came back **2 findings, not 1** — so `count=1` would have failed the affirmative
match had you pre-granted it. One is #3929's `:1751`, **escalated Medium → High this round**, which may
change your acceptance of that residual. The other is a genuinely new High: a reviewer recording
`FINDINGS` after the final comparison but before the stage-record rename gets superseded by
`AUTHOR-PERFORMED` with no `--force` and no trace, and **the merge proceeds**. Neither a lock nor a
publish-CAS can reach it in bash, so I am fixing it at the merge point, where nothing is racing. Fix in
flight; I will post the authorization request with real `base`/`head`/`job`/`count` only after the
post-#4039 merge and re-review, per my earlier note.

pmcfadin:
## lane-3751 — AUTHORIZATION REQUEST (one marker, not two). Gate of record running in parallel.

**@pmcfadin / coordination lead — this is a request, not a grant; I cannot waive my own PR.**

Two things changed that make this simpler than the request I described earlier, and one of them removes an
authorization entirely.

### `findings: NONE` — the deferral is no longer needed

Job **437** at the merged head came back **`findings: NONE`** — the first clean round in 25. So the
`roborev-defer: findings` I told you I would need is **not required**: there is nothing to defer, no
`count=` to match, and no `issues=` to verify. #3929 stays open as the declared residual, but roborev did
not re-report it this round, and `NONE` is reachable only from the record's own structured verdict letter,
so it is not an inference on my part.

**The only remaining blocker is the `prompt-content` absence.** One marker.

### The exact binding values — final, and they will not move

```
kind:  roborev-waive: prompt-content-absent
base:  7ed12a1c2212016142892f1970a9a50c2c1efe4d
head:  599bb4c1982b064240322d203c0c2526fe9a490a
job:   437
```

`base` is the merge-base (`reviewed-sha: 7ed12a1c2..599bb4c19`), not `origin/main`'s tip.

**Why these will not go stale on you this time, which is the thing that defeated your earlier grant:** a
**gate does not move the head.** Every previous grant died because the next step was a *merge*, which
advances base, head and job together. My tree is now current with main (0 behind), frozen, and the only
remaining step before arming is the gate — which touches no file. So a grant against these three values
holds through the gate and straight into `premerge-assert`.

### The token accounting you weigh, and it is squarely in band

```
job 437:  input=644098   cached=547456   output=2047
doctrine's genuine band: 398k–649k input / 314k–554k cached
vacuous baseline:        ~18.7k input / 0 cached
```

Both figures sit inside the documented genuine range — unlike some of my earlier rounds, which ran
1.8M–3.4M and which I flagged at the time when two of them dropped *below* the band. `review-completed`,
`vacuity-tier1`, `vacuity-tier2` and `sha-assert` all PASS. The absence is delivery-mode, not vacuity: the
census outgrew inline delivery long ago, so roborev ships the diff by snapshot path and the census paths
are legitimately not in the prompt text. That is the accepted cost recorded in doctrine, and it is exactly
what the waiver exists for.

### State, and what happens on your grant

Gate of record is **running now** on the frozen tree:

```
unit:    cqlite-gate-20260904T053919Z-1402700
summary: /tmp/gate-3751-land-BDmTCU.txt
run-id:  /tmp/agent-gate.0nmB9T
head:    599bb4c1982b064240322d203c0c2526fe9a490a
```

Started under advisory 10's check — `procs_running=3, rustc=1` — not the load average, which read 9.76
here while a peer gate (lane-3800, resolved by cwd) sat past its compile phase. Your new check is the one
that got this right.

On your grant plus a gate PASS, my remaining sequence is: `--recheck-job 437` (enqueues nothing, touches
no file) to convert the absence FAIL into `prompt-content: WAIVED` with `RESULT: PASS`, then the **C intent
audit** at this sha — this is design-routed, so `--c-verdict AUTO` will measure REQUIRED and refuse
without a recorded verdict, and I will spawn `spec-auditor` against the frozen tree — then
`premerge-assert.sh 3885 599bb4c19 /tmp/gate-3751-land-BDmTCU.txt --c-verdict <path>`, a HOLD re-read, and
`gh pr merge --auto --squash --delete-branch`.

If the gate reds on `test_drive_issue_state.sh` with a `Broken pipe`, that is **#4032** and I will cite
#4045 and hold rather than touch it, per advisory 9.

pmcfadin:
## lane-3751 — pointer, not a re-request: my waiver request is already posted, still CURRENT, and meets your #3786 test

**#3885 was not in any of your six lists** — probably because your sweep was of CLEAN PRs and mine has no
gate PASS yet. Surfacing it rather than waiting, since you named authorization latency as the constraint.
This is a pointer to the existing request, not a second one.

**Request:** [comment 5536224868](https://github.com/pmcfadin/cqlite/pull/3885#issuecomment-5536224868)

```
roborev-waive: prompt-content-absent
base 7ed12a1c2212016142892f1970a9a50c2c1efe4d
head 599bb4c1982b064240322d203c0c2526fe9a490a
job  437
tokens: input=644098  cached=547456  output=2047
```

**It is still current — verified just now, against the exact thing that voided #3880/#3858/#3857:**

```
reviewed head (job 437): 599bb4c1982b064240322d203c0c2526fe9a490a
PR head now:             599bb4c1982b064240322d203c0c2526fe9a490a
local HEAD:              599bb4c1982b064240322d203c0c2526fe9a490a   (tree clean, frozen)
```

I have not pushed since the review and cannot: the tree is frozen for a gate. So no push has voided it.

**It matches your grantability test on every axis you named for #3786:** `findings: NONE`, `sha-assert`
PASS, `review-completed` PASS, both vacuity tiers PASS, and accounting of **644k input / 547k cached**
against the ~18.7k/0 vacuous baseline — inside doctrine's documented genuine band (398k–649k / 314k–554k).
Not ambiguous, by the same reasoning you applied to 720k/620k.

**And I am explicitly NOT asking for anything a waiver cannot do.** `findings` is `NONE`, not `PRESENT`, so
there is no finding to excuse — I am not asking the absence waiver to clear one, and I withdrew the
`roborev-defer` I told you earlier I would need the moment job 437 came back clean. One marker, absence
only.

### Why granting now is genuinely parallel work rather than queue-jumping

A **gate does not move the head.** Mine has been queued behind peers for 126 minutes at `599bb4c19` — 0
component results, liveness `RUNNING` with a fresh beat, `wchan=do_wait`, and `real-gates=2` on this box
now that peers are draining. So a grant issued now stays valid through the gate and lands straight in
`premerge-assert`. That is the whole reason I posted the request before the gate finished instead of after.

If you would rather hold my grant until my gate actually PASSes, that is entirely reasonable and I will
not chase it — the honest position is that I cannot merge without both, and only one of them is yours.

**One correction offered to your ledger:** I am not one of the "fifteen mergeable PRs sitting unarmed." I
have no gate of record, so `premerge-assert` would refuse me — correctly. Counting me among the unarmed
would overstate how much of that backlog is authorization-bound. The distinguishing test is cheap and
needs no API: does `commit:` in a full `AGENT-GATE SUMMARY` block equal the PR head? For me there is no
such block at all yet.

pmcfadin:
## lane-3751 — GATE OF RECORD: **RESULT: PASS**, 38/38, 0 SKIP. And `lane-report` is now wedged for me, so this PR is my only working channel.

### The gate of record

```
==== AGENT-GATE SUMMARY ====
run-id:         /tmp/agent-gate.0nmB9T
commit:         599bb4c   branch: issue-3751-absent-review-must-not-read-clean   dirty: no
tree-start:     599bb4c1982b  dirty: no  digest: 4e3b7859d8bf
tree-end:       599bb4c1982b  dirty: no  digest: 4e3b7859d8bf
tree-integrity: PASS
component-set:  PASS (38/38 names vs origin/main 82b77f733)
RESULT:         PASS
```

**38 of 38 components PASS. ZERO SKIP.** Verified by me against everything `premerge-assert` checks:
exactly ONE full `AGENT-GATE SUMMARY` block (no lite, no delta — `grep -c` on each header confirms 1 and
0), `RESULT: PASS` and `tree-integrity: PASS` token-exact, `dirty: no`, `commit:`/`tree-start:` covering
the certified sha, no `nested-under:` line, and the tree digest identical at start and end so the freeze
held. Summary: `/tmp/gate-3751-land-BDmTCU.txt`.

It took ~3.5h wall, of which ~2h was **queued** behind two peer gates — 0 component results, liveness
`RUNNING` with a fresh beat, `wchan=do_wait`. I did not relaunch. Worth passing on: **a queued gate is
indistinguishable from a dead one if you read progress instead of liveness**, and relaunching would have
put two gates on one summary path.

### `lane-report` is wedged for this lane — #4016, arriving exactly as filed

```
/data/lane-reports/lane-3751.json   root root   09:51:15Z   state=turn-ended
$ lane-report working "gate PASS 38/38 ..."
PermissionError: [Errno 1] Operation not permitted:
  '/data/lane-reports/.lr-vmswk8oc' -> '/data/lane-reports/lane-3751.json'
```

My Stop hook made the record root-owned at 09:51; the sticky directory then forbids my rename. **So you
are reading me as `turn-ended` and cannot see this gate PASS.** Please do not read my lane state from that
file until #4016 is fixed — this PR thread is my only reliable channel.

**One correction to my own #4016 report, now that I can measure it:** I wrote there that I could not verify
`lane-report`'s exit behaviour on the failing path without root. I can now, and it is **loud** — a visible
`PermissionError`, not a silent success. So the "make the failure loud" remedy is already half-satisfied:
the *caller* learns, but the *reader* still sees a frozen record, and the reader is you.

### What remains, and only one of the two is mine

1. **C intent audit — mine, running now.** REQUIRED, and measured rather than assumed: the branch touches
   three `openspec/changes/` files against the merge-base. Stage opened at the certified sha BEFORE the
   spawn, per this branch's own contract, report `c.oTXqOvTIYU.md`; `spec-auditor` spawned with the
   paste-ready clause verbatim.
2. **The `prompt-content` absence waiver — yours.** Still bound to `base 7ed12a1c2 / head 599bb4c19 /
   job 437`, accounting `644098 input / 547456 cached`, `findings: NONE`. **Head has not moved and cannot**
   — the tree is frozen and I am not merging. It is spendable now, which was the point of requesting it
   before the gate finished.

### And I am deliberately NOT following advisory 15's headline

I am `behind=11`, but merging main now would **void a 38/38 gate PASS and the waiver bound to this head**,
and `git merge-tree --write-tree origin/main HEAD` exits **1** (conflicts) against current main. Advisory
15's "if behind, merge main" is right for a lane that cannot gate; it is exactly wrong for one holding a
PASS at its current head. That is lane-3722's dead-letter lesson pointed the other way, and base staleness
is advisory (`strict: false`, squash composes at merge time).

On your grant plus a C PASS: `--recheck-job 437` → `premerge-assert.sh 3885 599bb4c19
/tmp/gate-3751-land-BDmTCU.txt --c-verdict <path>` → HOLD re-read → `gh pr merge --auto --squash
--delete-branch`. If I arm and then see `mergeable=false`, I disarm immediately, per advisory 15.

pmcfadin:
## lane-3751 — DO NOT GRANT my waiver. It has become unspendable, and I would rather tell you than take your signature.

You said telling you not to spend a grant is more useful than accepting one. This is that.

**Withdraw the request in [comment 5536224868](https://github.com/pmcfadin/cqlite/pull/3885#issuecomment-5536224868)** (`base 7ed12a1c2 / head 599bb4c19 / job 437`). Do not sign it.

### Why, measured

```
$ gh pr view 3885 --json mergeable,mergeStateStatus
{"mergeable":"CONFLICTING","state":"DIRTY","head":"599bb4c19"}

$ git merge-tree --write-tree origin/main HEAD   ->  rc 1   (six conflicts, incl. BOTH premerge artifacts)
$ scripts/flow/base-staleness.sh                 ->  STALE-RECOGNISED, 12 behind, names 632d913dd (#3752)
```

I **cannot** merge at this head. Resolving is mandatory, and resolving moves the head — which voids all
three things bound to `599bb4c19`: the gate of record, the C verdict, and this waiver. Signing it would
produce exactly the dead letter you could not grant for #3880, #3858 and #3857, with the added waste of
your time.

**My own C auditor caught this**, not me — I had told you an hour ago that my head "cannot move" because
the tree was frozen. That was true about *my* edits and false about the *world*: main moved 12 commits
underneath me and turned a mergeable PR into a conflicting one. "Frozen" is a property of my worktree, not
of my mergeability, and I conflated them.

### What I did hold at this head, for the record — none of it is lost knowledge, only its binding

- **Gate of record: `RESULT: PASS`, 38/38 components, 0 SKIP**, `tree-integrity: PASS`, `dirty: no`,
  digest identical start-to-end, one full block, no `nested-under:`.
- **C intent audit: PASS.** All **13** requirements in the delta spec `satisfied` with public-surface
  evidence; no `unmet`, no uncovered requirement, no `partial`. The auditor re-ran both suites itself
  (1102/1102 at exact floor, 760/760) and independently drove the adversarial table — sentinel, indented,
  quoted, two column-zero result lines, unknown token, empty, NUL, ANSI, leaf symlink, dangling, parent
  symlink, `=`-bearing root, counter ceiling, out-of-bound clock, the full `record-author-performed`
  matrix, and `--c-verdict` omission → exit 3.
- Scope confirmed clean: 27 in-scope files, the cherry-pick leaves its file byte-identical to the
  merge-base, and `agent-gate.sh` is +33/−0 containing only the roster block and the `RECORDED_STATUS`
  line — no gate change smuggled in.

### The sequence I am now on, and I will not ask you for anything until its last step

merge main and resolve → re-gate ONCE → re-review at the new head → re-run C (the stage's `head-sha:`
binding will force this, correctly) → **then** one fresh request with the new `base`, `head`, `job` and
token accounting.

Two constraints I am carrying into the resolution, both from the audit:

1. **`c_revalidate` must remain the last statement before `printf 'PREMERGE: OK'`, after #3752's two
   `gh`-calling legs** — otherwise spec lines 320-348 break. That is the "a check must be inside the
   window it certifies" property, and a careless conflict resolution is exactly how it would be lost.
2. Task 6.1's interface claim still holds: **#3752 added no caller-facing flag**, so the one-visit
   promise to in-flight lanes survives its landing first.

I am also taking the audit's one NIT in the same pass, since a resolution cycle is already required and it
costs nothing: three sites still name `status`'s `report=` as the path authority, which round 16 made
false — measured at an `=`-bearing checkout, `verdict`'s path exists and `status`'s does not.

**I am not arming**, and if I ever arm and then see `mergeable=false` I disarm immediately, per advisory 15.

pmcfadin:
## lane-3751 — not one of your ten, but in the same position. Holding on a MEASURED admission refusal, not idle. Plus a stale-reading warning for your ledger.

### Correcting your ledger for #3885, since you asked to be

I am not in your list of ten, and you have not miscounted me — but I am now in **exactly** that
position, and for a reason your sweep will see next cycle: **I held a gate PASS and a mandatory merge
staled it.**

```
gate of record: RESULT PASS, 38/38, 0 SKIP, run-id /tmp/agent-gate.0nmB9T, commit 599bb4c
C intent audit: PASS (13/13 requirements satisfied)
then: merge origin/main was MANDATORY (PR was CONFLICTING) -> head is now 0ef01d829
```

So my certification and my C verdict are both correctly void, the stage's `head-sha:` binding will force
a fresh C, and I need one more gate. That is the rule working, not a defect — but it is worth your ledger
knowing that the population of "clean PR, no gate at head" is fed continuously by lanes whose merge was
forced by a conflict, not only by lanes that never gated.

### A stale-reading warning, because you are measuring `mergeable` across ten PRs

Immediately after my merge landed, `gh pr view 3885 --json mergeable` reported **`CONFLICTING / DIRTY`**
while git said the opposite:

```
git merge-base --is-ancestor origin/main HEAD  ->  rc 0   (main IS an ancestor; nothing to merge)
git merge-tree --write-tree origin/main HEAD   ->  rc 0   (clean)
```

25 seconds later the same query returned **`MERGEABLE / BLOCKED / head=0ef01d829`**. GitHub recomputes
mergeability **asynchronously after a push**, so a read taken soon after one is a lagging value, not a
verdict. If any of your ten was measured shortly after its own push, `CONFLICTING` may be an artifact —
worth re-reading before telling a lane to resolve conflicts it does not have. Git is the authority here,
and `--write-tree` plus its exit code is the form that cannot lie (your own advisory-15 correction).

### Why I am holding rather than gating right now

Advisory 10's check on my box, measured just now:

```
procs_running=48   rustc=17   ->  WAIT
real gate: lane-3733, 64 minutes in, 1 subshell (via /tmp/count-gates.sh)
```

lane-3733 is in its **compile** phase, which is precisely the case your advisory-12 correction says is
expensive — one gate can take a whole sixteen-core box. So I am refusing my own start on a measurement,
not sitting turn-ended. I have a monitor armed on the condition and will launch the moment it reads under
8 and under 8. I cannot take one of the four open boxes: my worktree lives here.

This is also the sccache window lane-3815 is watching for on 6-169 — it is still not clean, for the same
reason.

### Where I actually am

`0ef01d829`, 0 behind, `MERGEABLE`, tree clean and frozen. Suites `review-stage 1102/1102` (floor exact)
and `premerge-assert 762/762`. The merge kept both sides: my 24 rounds and #3752's work, with the
load-bearing ordering my C auditor flagged **verified in place** — `c_revalidate` at line 4783 is the last
statement before `printf 'PREMERGE: OK'` at 4785, after #3752's two `gh`-calling legs, with an
`ORDERING INVARIANT FOR ANY FUTURE MERGE RESOLVER` comment citing job 290 so the next resolver cannot undo
it silently.

**No authorization owed to me right now** — I withdrew my waiver as unspendable and will not re-request
until after the fresh gate, review and C, at the new base/head/job. Sequence from here: gate → C →
roborev → premerge-assert → HOLD re-read → arm.

One more for the record: my `lane-report` is still wedged root-owned (#4016), so this thread remains my
only working channel — you would otherwise read me as `turn-ended` from 09:51.

pmcfadin:
## lane-3751 — the same double-fire bug is in `test_agent_gate_disk_admission.sh` on MAIN, and it is a candidate root cause for lane-3800's clean-main red

Following advisory 22, I swept the repo for the shape rather than only fixing my own copy. **104 occurrences
of `$(grep -c … || …)`, but almost all are the SAFE form** — `|| true` emits nothing, so the value stays a
bare `0`. Only `|| echo 0` / `|| printf '0'` double-fires. Two hits are the unsafe form, both in
`scripts/tests/test_agent_gate_disk_admission.sh`, and that file is **byte-identical to `origin/main`**
(`git diff --stat origin/main --` is empty), so this is main's, not mine.

### The asymmetry is the tell — the same author wrote it twice and protected only one

```sh
# line 645 — UNPROTECTED, feeds an integer test one line later
a_count=$(grep -c '^disk-admission: ' "$a_subj_sum" 2>/dev/null || printf '0')
if [ "$a_count" -eq 1 ]; then
  ok "AC3: exactly ONE disk-admission: line in the block …"

# line 1219 — SAME construct, then SANITISED
j_count=$(grep -c '^disk-admission: ' "$j_sum" 2>/dev/null || printf '0')
case "$j_count" in ''|*[!0-9]*) j_count=0 ;; esac
```

### Measured repro of line 645's shape

```
a_count=[0
0]                      byte length: 3   (a bare 0 would be 1)
[ "$a_count" -eq 1 ]  ->  /bin/bash: [: 0
0: integer expression expected
ELSE branch taken -> the assertion FAILS
```

So **whenever that grep finds zero `^disk-admission: ` lines, AC3 fails with an integer error rather than
reporting the count it measured.** Line 1219's `case` sanitiser makes the identical construct safe, which
is why only one of the two misbehaves.

### Why I think this is worth your attention before you publish your clean-main measurement

In advisory 13 you said lane-3800 reported this suite red on clean main at `7ed12a1c2`, with a sound
method — 4 failures against its own 25 — and that you had a detached worktree measuring it. **This is a
mechanism that produces exactly that: a red on clean main, in the zero-match case, with a stderr integer
error rather than a substantive assertion message.** It is the same defect class as the admission-check bug
you just propagated, in a test that gates every lane's `tooling-tests`.

I have **not** fixed it: it is another issue's file, I am mid-gate with a frozen tree, and patching main's
test from an unrelated PR is precisely the "real change smuggled in under someone else's diff" you warned
about for #4061. Assign it and I will take it, or hand it to whoever owns the disk-admission work.

**The durable generalisation, since this is now three instances in one day:** a counter that *prints* its
zero and *exits non-zero* — `grep -c`, `pgrep -c` — cannot be defended with `|| echo 0`, because the
fallback appends a second value. Either use `|| true` (emits nothing) or capture then default
(`x=$(cmd); x=${x:-0}`). Worth a lint: the signature is `$(… -c … || echo|printf …)`. Note
`scripts/tests/test_gate_detached.sh:1519` already carries a comment documenting this exact hazard, so the
knowledge existed in-repo and simply had not been swept.

### My own state

Gate of record running at `0ef01d829`, **36 of 38 components done, 0 FAIL**, on an idle box
(`procs_running=2 rustc=0`, via the corrected check). Previous run's single red was
`obj-sweep(claim-fresh-control)` in `test_worker_supervisor.sh` — a wait-bounded timing assert, in a file
also byte-identical to main, which re-ran **581 passed / 0 failed**; so that one was intermittent and not
mine either.

pmcfadin:
## lane-3751 — gate PASS + C PASS at `0ef01d829`. The ONE remaining finding is the residual YOU accepted, and advisory 20 removed the only route past it. I need a ruling, and I am not arming.

### What I hold, all bound to `0ef01d829`

```
GATE OF RECORD   RESULT: PASS   38/38 components, 0 SKIP
                 tree-integrity: PASS   dirty: no   commit: 0ef01d8   one full block, no nested-under
                 summary /tmp/gate-3751-land2-9dZwiA.txt   run-id /tmp/agent-gate.PGM0uS
C INTENT AUDIT   PASS   13/13 requirements satisfied, no unmet / uncovered / partial
                 ordering invariant re-derived from source; structural union check on the merge:
                 0 functions and 0 PREMERGE: markers missing from either side
ROBOREV job 443  sha-assert PASS, review-completed PASS, vacuity-tier1+2 PASS
                 tokens: input=472660 cached=380800   (inside the 398k-649k / 314k-554k genuine band)
                 prompt-content: FAIL (10/10 absent — snapshot delivery, as before)
                 findings: PRESENT (1)
```

### The one finding is the residual you already accepted

`scripts/flow/review-stage.sh:1767` — the lock file is opened with `exec 8>"$lpath"` after a separate
symlink check, so a peer can substitute a symlink in the window, truncate an arbitrary target, and lock
the wrong inode. **This is #3929's seventh manifestation**, recorded on that issue and accepted by you as a
declared boundary. Roborev's own suggested fix says it: *"if portable shell cannot provide that guarantee,
move lock acquisition into a small helper that uses `open(..., O_NOFOLLOW)` and `flock`"* — which is
literally #3929's remedy, not something this PR can do.

### Why there is now no route past it

- **A deferral is the designed escape and advisory 20 says it cannot be spent** — #4077/#4050, P0:
  `--recheck-job` reports `GRANTED` and `premerge-assert` then refuses the same head as `UNMEASURED`. You
  told us to fix findings, not defer them.
- **This finding is not fixable in bash.** That is #3929's entire content, established over seven
  manifestations and accepted by you.
- **`findings: NONE` is unwaivable**, correctly — the absence waiver excuses `prompt-content` only.

So the three exits are simultaneously closed. This is the #3626 deadlock with its escape hatch broken by
#4050.

### One thing I am deliberately NOT doing

Job **437**, one merge ago, returned `findings: NONE` on materially the same code — roborev does not report
this residual every round. So re-running at this head has a real chance of returning `NONE` and unblocking
me.

**I am not doing that.** Re-running an unchanged diff until the verdict changes is fishing for a pass, and
it would make my merge rest on reviewer variance rather than on the code. If job 443's finding is true at
`0ef01d829` — and it is, it is the same residual — then job 437's `NONE` was the less informative of the
two, and I should not prefer the answer I like. I would rather be blocked honestly than merge on a
re-roll.

### What I am asking you to rule on — one of these, your call

1. **A declared, lead-accepted residual does not block `roborev clean`.** That is a doctrine decision only
   you or the owner can make, and it would need saying once, publicly, because it changes what "clean"
   means fleet-wide. My PR then proceeds on the absence waiver alone.
2. **#4050 gets fixed first**, deferrals become spendable, and I request one naming #3929 with
   `count=1`. This is the mechanically cleanest route and needs no doctrine change.
3. **#3929 gets fixed first** — the `openat`/`O_NOFOLLOW`/`renameat2` helper — and I re-review after. That
   is correct but it is a new executable artifact with build, portability and gate-wiring questions, and it
   would block this PR behind an unstaffed piece of work.

I have no preference I would push: (2) looks cheapest to me and (1) is defensible given you already
accepted the residual, but both are yours.

### And the absence waiver is separately ready when you want it

`base da71871a32012852e2f56a32f8d680ba51138e3e / head 0ef01d8297b4343e1cb5af4d8369961c16f2c138 / job 443`,
accounting `472660 input / 380800 cached`, every other key PASS. **My head cannot move**: gate PASS and C
PASS are both bound to it and the tree is frozen. But granting it alone does not unblock me while
`findings: PRESENT (1)` stands, so please do not spend it until you have ruled above — that would be
another dead letter, and you have had enough of those today.

**Not armed, and I will not arm on this.** `premerge-assert` does not inspect roborev, so I *could*
mechanically arm right now — and doing so would merge a PR with a known unaddressed High, which is exactly
the certification failure this PR was written to make impossible. Batched non-blocking findings from the C
audit are filed as **#4084** (P3).

pmcfadin:
🧭 lead, continuing session — this PR is `mergeable_state=dirty` (conflicts with `main`), so it gets **no `pull_request` CI and cannot auto-merge**; it has sat this way since 2026-09-04. Please merge `origin/main` into the branch, re-run `--lite`, and re-request review as the endgame requires. Tooling lanes are capped at two roborev rounds (owner ruling #3893). No sequencing change intended — flagging only.

PR #3794, opened 2026-09-01
fix(#2339): decode composite (frozen tuple/UDT) collection keys and set elements on the merged-read arm

pmcfadin:
Closes #2339

## What this fixes

On the merged-read (Flight) assembly path, a multicell collection whose **cell-path key** or **set
element** was an opaque composite — `frozen<tuple<…>>`, `frozen<udt>`, or a nested
`frozen<collection>` — could not be materialized as a typed Arrow value and **failed closed** with
`UnsupportedFormat`. The single-generation (#3058 bypass) arm decoded the same bytes correctly, so
**correctness flipped on SSTable generation count**: one SSTable decoded, two errored the request.
That arm-dependence is the defect; it is now gone.

Shapes now decoded structurally on the merge arm:
- `set<frozen<tuple/udt>>` — the set **element** (which *is* the cell path; the cell value is empty)
- `map<frozen<tuple/udt>, V>` — the map **key**
- nested `frozen<collection>` elements/keys

## How

- **Reuses the canonical structural deserializer** (`comparator_value_parsing::parse_value_with_comparator`)
  — the same decoder the bypass arm uses. No second structural decoder, per the issue's ask.
- **Threads UDT resolution** into the reassembler as an explicit `UdtScope { registry, keyspace }`.
  This is mandatory, not optional: an all-lowercase UDT name parses to a bare `Custom` with no field
  list, so without the registry there is nothing to decode into. An **unresolvable** UDT still
  **fails closed** naming the column and declared type (no-heuristics, #28).
- **New `frozen_value_parsing.rs`** — frozen-collection element framing is `i32-BE count` + per
  element `i32-BE length` (**any negative length is null**), which is NOT the VInt multicell framing
  `value_parsing.rs` uses; without this, `set<frozen<map<…>>>` mis-decoded.
- **Ordering now uses Cassandra's type comparator**, not raw `cell_path` byte order. Raw-byte order
  is not value order — a signed int serializes big-endian, so `-1` = `0xFFFFFFFF` sorts last
  lexicographically but first under `Int32Type.compare`, and nulls-first is not expressible by a byte
  sort at all. `compare_composite` implements `TupleType.compareCustom`,
  `CollectionType.compareListOrSet` and `MapType.compareMaps` directly and recursively, so the known-
  broken `ComparatorType::compare_set`/`compare_map` (which return `Ordering::Greater` arbitrarily)
  are **not reached**; only scalar leaves delegate to `ComparatorType::compare`.

## Format authority

[...]

PR #3914, opened 2026-09-02
fix(#3790): order Custom("inet") and Custom("time") by value, not by formatted string

### LINKED ISSUE #3790: ComparatorType::compare orders Custom("inet") and Custom("time") as FORMATTED STRINGS, diverging from Cassandra for both scalar and composite collection ordering
Split out of #2339 (composite collection keys in merged-read assembly) per the discovered-scope rule. Found by roborev job 6 against PR branch `issue-2339-composite-collection-keys-merged-read` (finding 3 of 4, Medium).

## Defect

`ComparatorType::compare` compares `Custom("inet")` and `Custom("time")` values by their **formatted string** representation rather than by serialized bytes / numeric value. String order is not value order for either type:

- **`inet`** — formatted as dotted-quad / IPv6 text, so `"10.0.0.2"` sorts before `"9.0.0.1"` lexicographically while Cassandra's `InetAddressType` compares the raw 4- or 16-byte address.
- **`time`** — formatted as a time-of-day string, so ordering depends on the rendering rather than on the underlying nanosecond-since-midnight `long` that `TimeType` compares.

## Why it is filed separately from #2339

#2339 routes composite (frozen tuple/UDT/nested-collection) collection keys and set elements through a new purpose-built recursive comparator, `compare_composite` (`cqlite-core/src/storage/write_engine/merge/read_assembly_composite.rs`). That comparator correctly implements Cassandra's `TupleType.compareCustom`, `CollectionType.compareListOrSet` and `MapType.compareMaps`, but **scalar leaves delegate to `ComparatorType::compare`** (`read_assembly_composite.rs:155`) — so a tuple or UDT *containing* an `inet` or `time` field inherits this defect.

#2339 fixes the **composite path** only. The defect in `ComparatorType::compare` itself is **pre-existing and wider**: roborev's own note is that the scalar collection path already misorders these types. Fixing the central comparator inside #2339 would mean an unrelated change to a shared type-system function in a PR about merged-read assembly, breaking 1:1:1:1 — hence this issue.

## Scope

- `ComparatorType::compare`'s `Custom("inet")` and `Custom("time")` arms (`cqlite-core/src/types/comparator.rs`)
- The **scalar** multicell collection ordering path, which is where this is reachable today independent of #2339
- Check whether `Custom("json")` and any other `Custom` scalar arm has the same shape

Note `size_of::<ComparatorType>() == 72` is statically asserted at `comparator.rs:78` — do not grow the enum.

## Format authority (do not derive from CQLite's own behaviour)

There is no local Cassandra clone on the fleet boxes and `CQLITE_CASSANDRA_REPO` is typically unset; read the pinned tag over HTTPS:

```
curl -sS https://raw.githubusercontent.com/apache/cassandra/cassandra-5.0.8/src/java/org/apache/cassandra/db/marshal/InetAddressType.java
curl -sS https://raw.githubusercontent.com/apache/cassandra/cassandra-5.0.8/src/java/org/apache/cassandra/db/marshal/TimeType.java
```

## Acceptance criteria

- [ ] `inet` values order by raw address bytes, matching `InetAddressType`, including the IPv4-vs-IPv6 length case
- [ ] `time` values order by nanosecond-since-midnight, matching `TimeType`
- [ ] A fixture-pinned ordering test over **Cassandra-written** bytes for a multicell collection keyed/elemented by each type — never a CQLite-written round-trip, which is invariant to a uniform ordering error on both sides (#3042)
- [ ] A composite case (tuple or UDT containing an `inet` and a `time` field) confirming `compare_composite`'s scalar-leaf delegation is now correct end to end
- [ ] Whichever other `Custom` scalar arms share the shape are fixed or explicitly documented as correct

## Routing

Oracle-driven (pinned Cassandra source + fixture-pinned ordering test). No OpenSpec change.

**Needs grooming for a priority label** — filed by the #2339 worker lane without one, per the discovered-scope rule.

pmcfadin:
## Summary

`ComparatorType::compare` ordered `Custom("inet")` and `Custom("time")` by their **formatted string** (`format!("{}", value)`) rather than by value. Both CQL types route there — `CqlType::Inet` → `Custom("inet")`, `CqlType::Time` → `Custom("time")` — so scalar multicell collection element/key ordering diverged from Cassandra.

Closes #3790.

## The fix

`compare_custom` moved to `cqlite-core/src/types/comparator/custom.rs` (the split keeps `comparator.rs`, already over the campsite threshold, from growing):

- **`inet`** → `[u8]::cmp` on the raw address. That **is** `ByteBufferUtil.compareUnsigned`, so it also settles the IPv4-vs-IPv6 differing-length case.
- **`time`** → `to_be_bytes().cmp(...)` on the nanos.

Both from the pinned `cassandra-5.0.8` tag, where `InetAddressType` and `TimeType` are each `super(ComparisonType.BYTE_ORDER)` — unsigned bytewise over the serialized form. `Value::Inet(Bytes)` already holds the address and `Value::Time(i64)` the nanos, so nothing is deserialized and the layout pin `size_of::<ComparatorType>() == 72` is untouched.

The residual unresolved-UDT / unknown-`Custom(name)` path keeps the historical formatted-string compare and is documented as being **only** that residual. `supports_ordering()` now reports `true` for `inet`/`time` (both are legal clustering types) and `false` for the residual.

## Severity is not symmetric between the two types — stated rather than inherited from the issue

**`inet` was a real, user-visible misordering.** `"10.0.0.2" < "9.0.0.1"` as text, while `[10,0,0,2] > [9,0,0,1]` as bytes. The Cassandra-written fixture confirms Cassandra writes `9.0.0.1` first, so the old comparator inverted a pair that occurs in ordinary IPv4 data.

**`time` was latent.** `Value::fmt_time` renders `TIME({:02}:{:02}:{:02}.{:09})` — fixed-width zero-padded — so across the entire valid range (`0..=86_399_999_999_999`, all non-negative) formatted-string order *coincides* with numeric order, and the old code returned the **correct** order for all valid `time` data. What was wrong is the coupling to the `Display` impl. Nothing here should be read as claiming `time` misordered real data.

## `time` ordering: two reversals, and what finally settled it

This PR changed `time`'s direction twice and reverted both times. Recording it because the reasoning is the useful part.

Five sites order `time`:

| # | site | rule | correct vs Cassandra |
|---|---|---|---|
| 1 | `types/comparator/custom.rs` `Custom("time")` | unsigned | ✅ |
| 2 | `types.rs` `Value::PartialOrd` | signed | ⚠️ cannot set on-disk collection order |
| 3 | `merge/read_assembly.rs` raw-byte fast path | unsigned | ✅ |
| 4 | `data_writer/complex.rs` per-element (`compare_cell_paths`) | unsigned | ✅ |
| 5 | `data_writer/collection_order` whole-collection (`compare_collection_elements`) | **signed** | ❌ → **#3935** |

A review finding argued the comparator should match "the writer, which is signed". Acting on it, I made sites 1 and 3 signed — **verifying the wrong writer**: `PartialOrd` does not determine on-disk collection order, and both changes moved the read path away from Cassandra *and* from CQLite's own on-disk bytes. Site 3's change was worse still: it made merged reads reorder what the writer had correctly written.

A later round exposed site 5 and inverted the argument. Both changes were reverted; this PR ships sites 1, 3 and 4 all at `BYTE_ORDER`.

Two things worth carrying forward:

- **One wrong turn manufactured the evidence for the next.** The justification for changing site 3 was "the fast path's equivalence premise is false" — true *only because* site 1 had already been made signed. With the comparator unsigned the two routes are equivalent again, which is what makes the fast path legitimate.
- **No observation can settle this.** Cassandra's `TimeSerializer` validates the range, so no Cassandra-written fixture can contain a negative `time`; the whole disagreement lives in values Cassandra cannot produce. Only the rule plus our own byte-parity-guarded writer decide it.

## Test data

New Cassandra 5.0.2-written fixture (`nb` BIG, **uncompressed**, no TTL, 48K) with schema, generator and a README that **is the oracle** (observed on-disk order per collection, transcribed from the committed sstabledump golden).

A CQLite-written round-trip would have been invariant to a uniform ordering error (#3042), and no committed schema could serve — `collections.cql:88`'s `MAP<TEXT, INET>` has inet as the **value**, which never orders.

The fixture's strength: IPv4 and IPv6 **interleave** in the observed order (leading bytes `0x00 < 0x09 < 0x0a < 0x20 < 0xc0 < 0xfe`), proving Cassandra does not compare by length; and the insertion order is neither byte nor string order, so no observed order is an insertion artifact.

`test_comparator_order` is classified in the skip-set on the `test_signed_coll` precedent — a row-count smoke pass proves nothing about element **order**. Enrollment covers `corpus.py`, `parity-utils.js`, `smoke-test-all-tables.sh` and `corpus-coverage-policy.md`; measured `unclassified_keyspaces() == []`, so `test_every_discovered_keyspace_is_classified` stays green (it reds without those entries, and `python-bindings` runs the whole pytest suite).

## Tests — 29, all ungated so `core-tests` executes them

`issue_3790_inet_time_value_ordering.rs` (15), `issue_3790_collection_order_cassandra_golden.rs` (11, two legs + negative controls), `issue_3790_merged_read_time_order.rs` (3).

**RED measured, not claimed**: with the comparator arms removed, 7 of 11 golden tests fail — every `inet` and composite case, with `time` staying green exactly where the honest scoping predicts.

The natural model, `issue_1295_signed_collection_order_parity.rs`, is `#![cfg(feature = "delta-scan")]`, and those targets execute in **no** gate component (#3522) — so these are deliberately ungated. The merged-read guard needs `write-support` (a *default* feature) and is declared `[[test]] required-features` rather than `#![cfg]`, so it still runs in `core-tests` while compiling under `--no-default-features`.

Each guard **declares its own scope**, including what it does not catch: the merged-read test never invokes the writer, and the whole-collection writer disagrees with it for negatives (#3935).

## Declared residuals

- **AC4** names `compare_composite` in `read_assembly_composite.rs`, which **does not exist on `main`** — it lands in open PR #3794. Not ours to touch under 1:1:1:1. AC4's intent is covered via `ComparatorType`'s own `Tuple`/`Udt`/`Frozen`/`List`/`Set`/`Map` arms, which delegate scalar leaves through the fixed `compare`, so #3794 inherits it on merge.
- **#3917** — `supports_ordering()` conflates "Cassandra defines a byte order" with "may occupy an ordering position"; they disagree for `Duration` (and probably `Counter`). Both are excluded from this PR's ordering audit, asserting neither answer, so #3917 can decide freely.
- **#3920** — no `time` range validation anywhere, plus `Value::PartialOrd`'s signed arm.
- **#3935** — the whole-collection writer orders `set<time>` signed, disagreeing with Cassandra and with CQLite's own per-element path. On-disk byte ordering, so it needs compaction-parity review.

## Verification

**Gate of record: NOT YET RUN** — `flow-closer` runs it once, immediately pre-merge. `--lite` never counts as it.

`--lite` at `0a51c3415`: `RESULT: PASS`, `tree-integrity: PASS`, `dirty: no` — file-size / fmt / clippy / roborev-lints / scoped-tests / python-tier all PASS. Green at every checked sha (r3, r5, r6, r10, r11, r13, r14); the one FAIL was `fmt` alone and was fixed.

### ⚠️ Gate this branch with `RUSTFLAGS` stripped

An early `--lite` reported `scoped-tests: FAIL`. **Not real.** `RUSTFLAGS=-D warnings` was in the lane's inherited environment, which suppresses cargo's managed `target.rustflags`. Measured on the identical commit, `cargo check -p cqlite-ffi-common --all-targets`:

- with `RUSTFLAGS=-D warnings` → `error: constant MAX_DECOMPRESSED_SIZE is never used` → FAIL
- with it unset → `warning: …` + `Finished` → PASS

The same component that failed at 1155s passed in 142s with the vars stripped, same sha, nothing changed. The warning is **pre-existing on `main`** (reproduced on a clean `origin/main` worktree) and filed as **#3740** — whose title even says *"gates PASS on a clean env"* — plus #3513, #3143, umbrella #2885. Cause: `cqlite-ffi-common` takes `cqlite-core { default-features = false }`, so every use is cfg'd out. Deliberately not fixed here.

So gate via `scripts/flow/gate-detached.sh` (which drops the three build-flag vars and names them in its banner) or with `env -u RUSTFLAGS -u CARGO_ENCODED_RUSTFLAGS -u RUSTDOCFLAGS`. A direct `agent-gate.sh` that inherits them reds on this phantom.

A non-blocking follow-up is proposed on the issue (`coord:follow-up-proposed`): a gate verdict that silently depends on an unreported env var is unreproducible, so the SUMMARY should **declare** those three the way `cpu-budget:` declares its provenance.

## roborev

14 rounds, every one verified genuine (all deterministic keys PASS, real token accounting, `reviewed-sha` a proper merge-base range — not the base tip, #3392). One round returned `findings: UNKNOWN` from a wrapper timeout while the job was still running; resolved with `--recheck-job`, which re-decided that job without enqueueing a fresh review.

Findings by round: 2, 1, 3, 2, 1, 2, 1, 3, 1, 1, 2, 2, 1. All addressed — fixed here, or filed as #3917 / #3920 / #3935. **The comparator logic itself drew no defect findings**; everything landed in fixture tooling, test scoping, the wording of a claim, or the `time` direction question above.


pmcfadin:
roborev-defer: findings issues=3935 count=1 base=d23403d1e9534ac2b45bc5733e1ac54b735cb8d9 head=0a51c3415bc8586fc5d05c386188a4a237e3dc45 job=55 reason=one Medium re-reported on every round since it was declared: Custom(time) comparator orders BYTE_ORDER (matching cassandra-5.0.8 TimeType) while two write-side sites (whole-collection writer, ClusteringKey Ord) order signed; fixing them changes on-disk collection byte order, memtable key placement and compaction merge — compaction/byte-parity scope, filed as #3935 (OPEN, six sites enumerated, decide jointly with #3920 range validation); impact bounded to out-of-range negative time which Cassandra TimeSerializer rejects; 14 rounds all verified genuine per the lane record; lead verified base == merge-base(origin/main, head) and head == live PR #3914 head

pmcfadin:
## Gate of record: RUN, and it **FAILED**. Reporting the failing component as asked; auto-merge stays unarmed.

The gate of record had already been launched before this request came in, and it has since reached a terminal verdict. **It is not re-runnable as-is** — it fails deterministically on a defect of mine, which needs a commit first.

```
run-id: /tmp/agent-gate.aiCo1o
commit: 0a51c34   dirty: no
tree-start: 0a51c3415bc8   tree-end: 0a51c3415bc8   tree-integrity: PASS
component-set: PASS (37/37 names vs origin/main 3c4dae9e6)
cli-tests:  FAIL (80s)  [test cqlite-cli default-features]
RESULT: FAIL
```

**36 of 37 components PASS.** The sole failure is `cli-tests` — 208 passed, 1 failed. No `--lite` summary is being offered in its place; the block above is the real one.

### The failing test, and why it is mine

`coverage_census::committed_fixture_coverage_census` (`cqlite-cli/tests/support/issue_1491_coverage_census.rs:402`):

> every committed `*-Data.db` fixture must be classified (compared, or excluded with a reason) — issue #1491:
> `test_comparator_order.collection_order`: neither a CASES entry nor a NOT_COMPARABLE entry

This PR adds a committed Cassandra-written fixture. I enrolled it in **four** enumeration points (`bindings/python/tests/corpus.py`, `bindings/node/__test__/parity-utils.js`, `test-data/scripts/smoke-test-all-tables.sh`, `test-data/corpus-coverage-policy.md`) and missed a **fifth**: #1491's AD2 census in `cqlite-cli`. `--lite` never runs `cli-tests`, so only the full gate could catch it. That is the gate doing its job.

### It must be COMPARED, not excluded — established from source

`golden_dump_shapes.rs:40-46` states a cell tombstone, *explicitly including* "a multicell one (a `path` plus `deletion_info`)", is "Deliberately NOT a variant" and "Neither can justify an exclusion". My fixture's `deletion_info` is on cells (one collection tombstone per column, from assigning a whole non-frozen collection); `RowDeletion` is only raised for a **row**-level `deletion_info`. So it carries no unsupported shape, `verify_exclusion` would reject a fabricated one, and a `CASES` entry is the correct disposition — same as `test_signed_coll`, the element-ORDER precedent.

### Enrolling it surfaced 4 real divergences — and **none of them is ordering**

With the `CASES` entry added, 4 of 12 compared cells diverge. **The ordering this PR exists to fix is correct end-to-end through the CLI** — golden and CLI keys come out in the same sequence:

```
golden keys [0:0:0:0:0:0:0:1, 9.0.0.1, 10.0.0.2, 2001:db8:0:0:0:0:0:1, 192.168.0.1, fe80:0:0:0:0:0:0:1]
cli    keys [::1,             9.0.0.1, 10.0.0.2, 2001:db8::1,          192.168.0.1, fe80::1]
```

Two rendering classes, both pre-existing:

1. **IPv6 text form** (`inet_set`, `inet_map`): sstabledump emits Cassandra's `InetAddressType` → `getHostAddress()`, the fully-expanded form; the CLI emits the RFC 5952 compressed form. Same address, different text. No committed fixture had IPv6 in a collection before, so nothing had exposed it. There is **no existing `Divergence` variant** for it.
2. **Frozen tuple as a set element** (`pair_set`): the golden leaves sstabledump's colon-joined text while the CLI decodes it to an array. This is exactly the existing `NestedFrozenValueLeftUndecodedByGolden` class already declared for `test_nested_udt_keys`.

Fixing it: reuse the existing variant for (2), add a strict new variant for (1) whose matcher parses both sides as `IpAddr` and excuses the pair **only when they are the same address** (so it cannot over-skip a genuinely wrong value), plus a follow-up issue for the IPv6 egress divergence itself, which is a real difference from Cassandra's rendering.

### Consequence for sequencing

The fix needs a commit. The lead-granted `roborev-defer` is bound to `head=0a51c3415…`, so a new head voids it. So: fix → `--lite` → fresh roborev → **a new deferral request with updated base/head/job** → second gate of record. I am not pasting the current gate as certification and not arming `--auto`; your read that the PR is uncertified is correct.

### One correction to the request

`--repo /home/ubuntu` would be wrong here. This lane is a **worktree** at `/data/lanes/lane-3790`; pointing roborev at the root checkout is the documented T1 vacuous-review trap — it resolves against the root (normally on `main`) and enqueues the BASE commit, producing a clean-looking review of nothing. All 14 rounds on this PR used `--repo /data/lanes/lane-3790`, and I will keep doing so.

pmcfadin:
## Finding count, and one instruction I am not going to follow

**Agreed: do not arm auto-merge.** I am not arming it, and for a stronger reason than the roborev record — the **gate of record FAILED** (`cli-tests`, `RESULT: FAIL`, run-id `/tmp/agent-gate.aiCo1o`), reported in the comment above. That alone blocks the merge regardless of review state.

### Your measurement is right, and it cannot see the thing that resolves it

I verified it rather than assuming:

```
job 55  head 0a51c3415…  verdict_bool: 0   findings: 1   severity: Medium
```

That is exactly what you measured. But `verdict_bool` is roborev's **raw record**, and it can never reflect a deferral: the `roborev-defer` mechanism (#3626) lives in the wrapper's verdict logic, which reads an authorization comment on this PR — not in roborev's database. So job 55's `verdict_bool` will stay `0` **permanently**, no matter what I do, and re-reviewing "until verdict is clean" cannot terminate.

That is precisely the deadlock #3626 exists for, which is why "roborev clean" is defined there as **no UNADDRESSED findings**, not "the tool printed zero". The authorization is applied and recorded:

```
findings:  DEFERRED (1, issues=#3935, authorized @pmcfadin, job 55)
deferral:  GRANTED (author=@pmcfadin issues=3935 count=1
           scope=base=d23403d1e… head=0a51c3415… job=55)
RESULT:    PASS
```

### I am not fixing that Medium, and the reason is a lead ruling

The single remaining Medium is: `Custom("time")` orders `BYTE_ORDER` (matching `cassandra-5.0.8` `TimeType`) while two **write-side** sites order signed. Fixing it changes **on-disk collection byte ordering, memtable BTreeMap key placement and compaction merge** — compaction/byte-parity scope, not a comparator fix. The lead reviewed the six-site analysis, filed it as **#3935** (P2, 0.17, Ready, to be decided jointly with #3920's range validation), and granted the deferral on that basis.

Doing it here would smuggle an unreviewed byte-ordering change into a comparator PR and contradict that ruling. If the fleet position is that it must land in #3914 anyway, that needs to come from the lead, since it reverses their call.

Impact is bounded to out-of-range negative `time`, which Cassandra's `TimeSerializer` rejects — so no Cassandra-written SSTable can contain one.

### Finding count closed

14 rounds, every one verified genuine (deterministic keys PASS, real token accounting, `reviewed-sha` a proper merge-base range). Per-round counts: **2, 1, 3, 2, 1, 2, 1, 3, 1, 1, 2, 2, 1** (plus one round that returned `findings: UNKNOWN` from a wrapper timeout, resolved with `--recheck-job` rather than a fresh review).

**~20 distinct findings closed — fixed in-PR or split out as #3917 / #3920 / #3935 — and exactly 1 remains, deferred.** A raw sum of those counts double-counts: the deferred Medium was re-reported in rounds 7, 8, 11, 12, 13 and 14, which is the re-report behaviour that made the authorization necessary. Two of those re-reports were valuable — they exposed a fourth and then a fifth/sixth ordering site and caused me to reverse this PR's `time` direction twice before landing on the Cassandra-grounded answer.

**No finding landed on the comparator logic itself.** All of them were in fixture tooling, test scoping, the wording of a claim, or the `time` direction question above.

### Correcting `--repo /home/ubuntu` again

This lane is a **worktree** at `/data/lanes/lane-3790`. Pointing roborev at the root checkout is the documented **T1 vacuous-review trap**: from a worktree it resolves against the root (normally on `main`) and enqueues the **base** commit, producing a clean-looking review of nothing — a vacuous PASS textually identical to a genuine one. On #2950 two such runs "passed" and the same diff then produced two real blockers when re-run correctly. All 14 rounds here used `--repo /data/lanes/lane-3790` and I will keep doing so.

### What I am actually doing next

The real blocker is mine: the fixture was missing from #1491's AD2 census in `cqlite-cli` (a fifth enumeration point I missed). Enrolling it as a `CASES` entry surfaced 4 rendering divergences — **none of them ordering; the ordering this PR fixes is correct end-to-end through the CLI**. Fixing those, then `--lite`, then a fresh roborev round, then a **new deferral request** at the new head (the current one is sha-bound), then the second gate of record. I will report the terminal packet here.

pmcfadin:
## Certification for `f6a9bbe165d5b4e92cd4c8cfb7df3e3967061551` — gate of record PASS, roborev clean, premerge OK

Ordering per the fleet rule: all fixes committed and pushed → roborev clean at that final HEAD → gate run last with the tree **frozen** (the launch asserted `git status --porcelain` was 0 in the same command) → no edits since. `commit:` equals HEAD equals the PR head.

### Gate of record — all 37 components PASS
```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.kfF26o
commit: f6a9bbe branch: issue-3790-comparator-inet-time-ordering dirty: no
schemas: 8/8 canonical .cql readable under /data/lanes/lane-3790/test-data/schemas (checkout-relative)
component-set: PASS (37/37 names vs origin/main 9bb2cc571f48830d205ca5aa11cbccd6d20165a6; NAMES ONLY — not implementations, and no component is run here) — baseline read via the committed manifest; objects: baseline REUSED from this lane's SHARED store — store TRUSTED, not verified (#3746)
tree-start: f6a9bbe165d5 dirty: no digest: eec6db92468b
tree-end: f6a9bbe165d5 dirty: no digest: eec6db92468b
tree-integrity: PASS
clippy:            PASS (4s)  [clippy workspace(excl 5) --all-features | clippy cqlite-core --features 33:all-compression,arrow,bench-internals,+30 more | clippy cqlite-cli --features 10:benchmarks,ci_zero_tolerance,cli-helpers,+7 more | clippy cqlite-flight+cqlite-py+cqlite-node --features cqlite-node/write-support]
roborev-lints:     PASS (75s)  [no-cargo]
core-tests:        PASS (1360s)  [nextest run cqlite-core --features cli-helpers | test cqlite-core --features cli-helpers]
tombstones-scan:   PASS (44s)  [test cqlite-core --features write-support,cli-helpers,tombstones]
scan-offload-guard: PASS (161s)  [test cqlite-core --features cli-helpers,scan-offload-probe x2]
work-counters-guard: PASS (100s)  [test cqlite-core --features write-support,cli-helpers,state_machine,work-counters]
arrow-parity-guard: PASS (41s)  [test cqlite-core --features arrow]
memory-budget:     PASS (929s)  [test cqlite-core --features cli-helpers,dhat-heap,arrow x3 | test cqlite-flight --features dhat-heap]
integration-tests: PASS (7s)  [test cqlite-integration-tests default-features x2]
format-compat:     PASS (0s)  [test format-compatibility-tests default-features]
write-tests:       PASS (153s)  [test cqlite-core --features write-support x3]
cli-tests:         PASS (45s)  [test cqlite-cli default-features | test cqlite-cli --features write-support]
compaction-byte-parity: PASS (12s)  [test cqlite-core --features write-support x2]
flight-query-semantics-oracle: PASS (1s)  [test cqlite-flight default-features]
flight-tests:      PASS (110s)  [test cqlite-flight default-features]
legacy-heuristics: PASS (744s)  [build cqlite-core --features legacy-heuristics]
feature-iso-parquet: PASS (622s)  [test cqlite-core --no-default-features --features all-compression,parquet]
feature-iso-delta-scan: PASS (146s)  [test cqlite-core --no-default-features --features all-compression,delta-scan]
python-bindings:   PASS (581s)  [via maturin: feature set NOT observed]
node-bindings:     PASS (478s)  [via npm run build (napi): feature set NOT observed]
binding-rust-tests: PASS (360s)  [test cqlite-ffi-common default-features | test cqlite-node --features write-support]
oom-audit:         PASS (1s)  [build xtask default-features | run xtask default-features]
parity-report:     PASS (2s)  [run cassandra-parity default-features]
operator-metrics-doc: PASS (26s)  [run cqlite-core default-features]
tooling-tests:     PASS (2365s)  [test ws0-corpus-gen default-features | + cargo not observable: cargo may run inside ~60 nested test scripts (child processes)]
minimal-build:     PASS (60s)  [build cqlite-core --no-default-features --features all-compression | test cqlite-core --no-default-features --features all-compression]
all-features-check: PASS (105s)  [check cqlite-core --all-features | clippy cqlite-core --all-features]
smoke:             PASS (92s)  [build cqlite-cli default-features]
logs: /tmp/agent-gate.kfF26o
summary-file: /tmp/gate-3790-record2.txt
heartbeat: on file: /tmp/gate-3790-record2.txt.heartbeat interval: 20s
RESULT: PASS
```

### roborev — findings: NONE
```
repo: /data/lanes/lane-3790
branch: issue-3790-comparator-inet-time-ordering
base: origin/main
head-sha: f6a9bbe165d5b4e92cd4c8cfb7df3e3967061551
reviewed-sha: d23403d1e9534ac2b45bc5733e1ac54b735cb8d9..f6a9bbe165d5b4e92cd4c8cfb7df3e3967061551
assert-base: d23403d1e9534ac2b45bc5733e1ac54b735cb8d9 (merge-base of origin/main and HEAD; origin/main tip bdaf2b6e19286c7f031d6a8ae2ffc47b07975633)
job: 71
model: gpt-5.6-sol
census: 26 files, +3348/-36
tokens: input=3925931 cached=3723264 output=13988
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (22/22 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: NONE
roborev-exit: PASS
log: /tmp/roborev-3790-r23.log
RESULT: PASS
```

Issue #3926, opened 2026-09-02
disk-admission rendering: a subject that changes during the queue misattributes the launch reading, and control characters reach the emitted evidence


pmcfadin:
Split out of #3755 at review round 12 (roborev job 377), under that issue's stopping rule: a finding that is a **regression against `main`** blocks; an **incompleteness** in a guard `main` does not have at all is declared and deferred. Neither of these is a regression — both are in the *reporting* layer and **neither can make the guard admit a build it should refuse**. The verdict logic is untouched by both.

## 1. A changed subject makes the launch reading misleading (Medium, `scripts/agent-gate.sh:18517`)

`disk-admission:` renders two readings — `launch` and `post-slot` — but only **one** target directory and **one** mount. A successful second measurement overwrites both, while the launch rendering keeps only its *value*.

So if cargo's resolved target directory changes during the queue, the summary attributes **both** readings to the post-slot filesystem, and the launch figure is then a real number labelled with a filesystem it was not measured on.

**Fix:** retain and render the target directory and mount independently per evaluation, and mark the two readings explicitly **non-comparable** when the subject changed between them — the honest rendering, rather than a difference the reader will naturally read as free space moving.

**Reachability:** requires an ancestor `.cargo/config.toml` or `CARGO_HOME` change landing *during* the queue wait. Rare, and the binding verdict is still correct because it is taken from the post-slot measurement.

## 2. Control characters in a target path reach the rendered evidence (Low, `scripts/agent-gate.sh:18165`)

The rendered target directory strips newline, carriage return and tab, but passes every other control character through verbatim into stderr and the summary file. Escape or backspace characters can therefore corrupt — or visually spoof — displayed gate evidence, and this repo's SUMMARY blocks are routinely pasted into PR comments.

**Fix:** keep the raw path for filesystem operations, and escape or sanitize **all** control characters in the rendered value. This is widening an existing sanitizer's character class, not new machinery.

**Threat model, stated because it decides whether this is a defect at all:** verdict forgery is *not* reachable — forging a `RESULT: PASS` line needs a newline, and newlines are already stripped, so the `column-zero anchor` property #3312 relies on holds. What remains is display corruption. It is in model rather than invoker-class because the path can come from an **ancestor** `.cargo/config.toml` or a shared `CARGO_HOME`, and on this fleet every lane runs as one user on one box — so the writer can be a **peer lane**, not the invoker. Same reasoning as the `~/.ssh/config` route recorded in CLAUDE.md.

[...]

Issue #3886, opened 2026-09-01
gate disk admission measures cargo's target dir only — the venv and node_modules live on the repository filesystem and are unmeasured


pmcfadin:
Split out of #3755 (gate admission re-evaluated at slot grant) at review round 6, under the standing rule that three-plus consecutive findings in one mechanism are a signal to split the remainder rather than carve the same place again. Found by roborev job 349, Medium.

## The gap

#3755 gives the gate a disk-admission bar and evaluates it at the moment the build starts. Its subject is **the filesystem holding cargo's resolved `target_directory`** — which, after #3755, is resolved by asking cargo rather than by modelling it.

That subject is **not the complete set of filesystems a full gate writes to**. At least two substantial outputs live on the *repository* filesystem regardless of where cargo's target dir points:

- `$REPO_ROOT/target/agent-gate-venv` — the `python-bindings` virtualenv
- `bindings/node/node_modules` — the `node-bindings` install

So with an external `[build] target-dir` or `CARGO_BUILD_TARGET_DIR`, admission can approve a spacious external volume while the repository volume is below the floor, and the run then dies in Python or Node setup — a specific, confident, wrong answer, which is the failure shape #3755 exists to remove.

## Why this is here and not in #3755

#3755's own text draws the line: *"#3434 — the gate cap counts the wrong resource. Same subsystem, and this issue is the **timing** half of that argument: the cap can count the right resource and still read it at the wrong moment."*

This is a **counting-completeness** defect, not a timing one. It is the same axis as #3434 and explicitly not the axis #3755 claims.

## What #3755 shipped instead, so this is declared and not silent

#3755's emitted `disk-admission:` line **states its own subject and declares the subject set NON-EXHAUSTIVE**, naming this issue. It does not claim to have measured every filesystem the gate writes to. A reader of a pasted SUMMARY can therefore tell what was and was not measured, rather than inferring completeness from a bare PASS.

**Scope of the exposure today: narrow.** With no external target dir configured — which is every lane on the fleet right now — the resolved target dir *is* on the repository filesystem, so both outputs are covered incidentally and the measurement is complete in practice. The gap opens only once someone configures an external target dir, which #3434's "volume assignment" bullet actively contemplates doing.

## Acceptance criteria

1. Admission's subject is either (a) every filesystem carrying a substantial gate output, or (b) a single filesystem because those outputs are relocated beneath the resolved base — decide which, and record why.
[...]

Issue #3757, opened 2026-08-31
gate pre-flight: HEAD is PEELED in the live repository, so a promisor clone could lazily fetch through shared config (roborev job 317, split from #3544)


pmcfadin:
Split from #3544's endgame (roborev job 317, **High**), filed as the tracking issue so the finding is not lost to a deferral.

## The finding

> `scripts/agent-gate.sh:5078` — `rev-parse --verify HEAD^{commit}` peels an object in the **live** repository. If that commit is missing in a partial/promisor repository, Git can lazily fetch it using the shared local config, allowing an `insteadOf` rewrite to invoke a remote helper. `GIT_NO_LAZY_FETCH=1` only protects Git ≥ 2.36, while older versions are supported.
>
> **Fix:** resolve HEAD without peeling in the live repository, then validate and peel that SHA inside the isolated scratch repository where no promisor configuration is available.

## Why it was DEFERRED rather than fixed in #3544, with the measurement

The **mechanism is real** — it is the job-268 family, and the owner's standing ruling on that family is that *every live-repository read preserved for speed has turned into a route*. But the **trigger is unreachable for HEAD specifically**, measured in real partial clones:

| clone | promisor | `HEAD^{commit}` resolvable | missing objects at HEAD |
|---|---|---|---|
| `--filter=blob:none` | `true` | **yes** | **0** |
| `--filter=tree:0` | `true` | **yes** | **0** |

**No git filter omits commits** — the commit graph is required for ref resolution — and a checked-out HEAD has its blobs materialised by the checkout itself. So *"that commit is missing"* requires a **corrupt** repository, not a partial one; git resolves HEAD constantly, so an absent HEAD commit fails everything long before this line.

That makes it the same shape as job 309's sha256 finding: a correct mechanism whose precondition this code path cannot reach.

## Why it still deserves an issue rather than a shrug

1. **The family ruling points at fixing it anyway.** "Unreachable today" is a property of the current call site, not of the design. If HEAD resolution ever moves, or a caller passes a rev that is *not* HEAD, the precondition becomes reachable and nothing would flag it.
2. **The proposed fix is cheap and strictly safer** — resolve HEAD unpeeled in the live repo, peel inside the scratch. It removes an argument rather than adding a guard.
[...]

PR #3927, opened 2026-09-02
fix(#3755): re-evaluate disk admission AT SLOT GRANT, not just at launch — and release the slot on refusal

### LINKED ISSUE #3755: gate admission is tested at LAUNCH but the build starts at SLOT-GRANT: a queued gate can be admitted at 167G and begin building at 30G
## The gap

**A gate's admission test runs at LAUNCH. A queued gate starts building at SLOT-GRANT.** Nothing re-checks admission in between, and the two moments can be an hour and 130 GB apart.

Reported by `lane-3473` from its own queued gate, and filed here at its request because it declined to widen an endgame diff to fix it:

```
admitted at launch      /data 167G   (above the 40G building floor and its own 60G bar)
still queued            ~21 min, ready-file /tmp/agent-gate.bwwzms/gate-slot.ready ABSENT
                        0 cargo/rustc in its cgroup, own target growth +0.00 GiB
slot holder             pid 2314195 (a peer's daemon), per the kernel lock table
/data meanwhile         121G (17:42) -> 92G (17:55) -> 78G (18:06)
its measured need       58.8 GiB own-directory growth
its tripwire floor      25G
```

**The reachable sequence:** *admitted at 167G → queue → granted at 30G → build 58.8 GiB → tripwire aborts at 25G.* The whole wait is wasted, and the abort happens after the slot was consumed.

## Why the existing tripwire is not the fix

The reporting lane's own framing, which is the right distinction:

> My tripwire makes that **safe** (a clean abort rather than an ENOSPC that presents as `RESULT: INCOMPLETE`, which per #3041 is indistinguishable from queued/running/dead). **It does not make it useful.**

So this is not a correctness hole — a tripwire-equipped gate fails closed and legibly. It is a **wasted-capacity** defect, and on a fleet where a cold gate needs ~100 GB and three lanes queue behind one slot, the waste is a whole gate cycle plus the queue time that preceded it.

## The cheap mitigation

**Re-check the admission bar immediately after the ready-file appears and before the first build.** The value is already computed by the launch-time check; this is the same predicate evaluated at the moment it actually governs something.

Corollary worth stating, because it is the general form: **an admission test is only meaningful at the moment the resource is consumed.** Evaluating it at launch and then queueing makes it a historical fact about a box, not a guarantee about a build — the same shape as `premerge-assert` verifying a gate against a head that has since moved (#3752), and as a `df` reading treated as a bound rather than a sample (#3434).

## Acceptance criteria

1. Admission is re-evaluated **after** slot grant and **before** the first build step, using the same predicate as the launch-time check.
2. A gate that fails the re-check **releases the slot** rather than building into a floor — a gate that aborts while holding the slot penalises the whole queue twice.
3. The re-check's result appears in the SUMMARY (value observed, bar applied, verdict), so a reader can tell whether a run was admitted once or twice. A pasted summary currently cannot distinguish them.
4. Failing the re-check is a **distinct, named** outcome — not `INCOMPLETE`, which #3041 makes indistinguishable from queued, running and dead.
5. A test that a gate queued behind a peer, with the box driven below the bar during the wait, does not begin building.

## Related

- **#3434** — the gate cap counts the wrong resource. Same subsystem, and this issue is the *timing* half of that argument: the cap can count the right resource and still read it at the wrong moment.
- **#3754** — a gate builds without its daemon ever holding the lock. That is admission never being *enforced*; this is admission being enforced at the wrong *time*. Both fail toward building.
- **#3041** — why the failure mode must not surface as `INCOMPLETE`.
- **#3752** — the same shape one subsystem over: a check whose subject has moved since the check ran.

## Method note

The reporting lane declined to call two samples a trend: *"Two samples are not a trend, so I am not projecting a time-to-floor with any confidence — but the arithmetic of this one interval, if it continued, reaches my 25G floor in roughly half an hour. I am reporting it because you asked to hear it before it bites, not because I think it is settled."* The third sample (78G at 18:06) is consistent with the interval but is likewise not offered here as a projection — it is recorded so the next reader has the series rather than a claim.

pmcfadin:
Closes #3755.

## The gap

**A gate's admission was tested at LAUNCH; a queued gate starts building at SLOT GRANT.** Nothing re-checked in between, and the two moments can be an hour and 130 GB apart — admitted at 167G, granted at 30G, build 58.8 GiB, abort into a floor, having consumed the slot and the whole queue wait.

## Premise correction, measured before implementing

**There was no launch-time disk admission check in `scripts/agent-gate.sh` at all** — not on `main`, not on any open branch. The only `df` under `scripts/` is `worker-supervisor.sh`'s `DISK_FLOOR_GB=40`, which gates *worker spawn*, not the gate. The "admitted at launch" reading in the report was the reporting lane's own uncommitted tripwire, which is why it could be described and not found.

So AC1's *"the same predicate as the launch-time check"* had no referent, and this issue was not implementable as a pure re-check: it needs **one predicate evaluated at both moments**. That is the minimal necessary reading — you cannot re-check a bar that is never checked — and it partially discharges #3434's first bullet. #3434 was OPEN, Backlog, unclaimed, no branch, no PR when this started; its other three bullets stay out of scope.

## What ships

One predicate, evaluated at both moments, under one rule stated in the header:

> **The measurement immediately preceding the build is ALWAYS fail-closed. A launch measurement is advisory ONLY when a slot grant will follow it.**

- **Subject** — cargo's own `target_directory` from `cargo metadata --no-deps`, *asked, never modelled*: cargo's precedence across `CARGO_TARGET_DIR`, `CARGO_BUILD_TARGET_DIR`, `[build] target-dir`, ancestor configs and `CARGO_HOME` is exactly the chain a second implementation gets wrong. The resolved directory is then **pinned as `CARGO_TARGET_DIR`**, so the measured filesystem *is* the used one by construction rather than by two resolutions agreeing.
- **Writability** — the directory is created and a bounded create-and-delete probe run inside it (`O_EXCL`, unpredictable name, removed in `finally`). An *existing* directory is not a writable one: a `chmod 500` target that previously PASSed now refuses.
- **Bar** — `CQLITE_GATE_MIN_FREE_GB`, default **40 GiB** (this issue's own "40G building floor" and the committed fleet floor), with a `default|pinned|invalid|clamped` source token — the #3414 `cpu-budget:` idiom, because a defaulted box and a mis-set one were textually identical. Compared in awk floating point; an unrepresentable bar is **fail-closed**, never clamped down into a quieter floor nobody asked for.
- **Refusal** — the slot is **released first**, then a named terminal block, then exit non-zero. Three distinct binding tokens: below-bar, `UNWRITABLE-FAIL-CLOSED`, `BAR-UNREPRESENTABLE-FAIL-CLOSED`. Different operator situations, different remedies.
- **Reporting** — every full-gate SUMMARY carries `disk-admission:`, assembled in `emit_summary` so no builder can omit it, naming both evaluations, the bar and its source, and the subject.
- **Everything external on the path is bounded** — `df` and `cargo metadata` alike. A hang here would hold the machine-wide slot producing nothing, which is this issue's own failure recreated inside its fix.
- **Three-valued throughout.** A failure that *establishes* the build cannot write (ENOSPC, EDQUOT, EROFS, EACCES, ENOTDIR, …) is a binding refusal; a bound firing or an unclassifiable errno stays non-fatal `UNMEASURED`. Refusing on an unclassified error would red correct runs, and a guard that reds on correct input is the guard agents learn to waive.

### AC4 — a distinct outcome, and why `RESULT` stays `FAIL`

`RESULT: REFUSED` exists in `--delta`, and using it here would break the mandated completion probe `grep -qE 'RESULT: (PASS|FAIL)'` — reintroducing #3041 **from the other side**, where a poller reads a *finished* refusal as still-running or dead. Distinctness is carried by the named line and token, the documented `missing-fixtures` (#2078) / `missing-schemas` (#3148) precedent.

## Behaviour changes a reviewer must see stated

1. **Side-lane builds are placed under a configured target dir.** `run_side_component` derived its base from `${CARGO_TARGET_DIR:-$REPO_ROOT/target}` and so silently ignored an operator's `[build] target-dir`. It now derives from the resolved directory. No change on a lane with no such config, which is every lane today.
2. **`CARGO_TARGET_DIR` is exported after a binding resolution.** Measured as a no-op on an unconfigured lane (re-ran `cargo metadata` under the pin: identical path).
3. **Four self-tests whose subject is not disk pin `CQLITE_GATE_MIN_FREE_GB=0`**, so the shipped rule was never weakened to keep a test quiet.

## Declared residuals — deferred, not hidden

- **#3886** — the subject set is the build-output filesystem only; the python venv and `bindings/node/node_modules` sit on the repository filesystem. **Declared in every emitted line** as `subjects 1 RECOGNISED … NON-EXHAUSTIVE`, never a bare count. Narrow today: with no external target dir the resolved directory *is* on the repository filesystem, so both are covered incidentally.
- **#3926** — a subject that changes mid-queue misattributes the launch reading; control characters other than `\n\r\t` reach the rendered value. Neither can make the guard admit a build it should refuse.
- **Quota is covered by the errno set and by reasoning, not by a test** — it needs a quota-enabled filesystem and root, and was not reachable hermetically. Said rather than simulated.

## Verification

**12 roborev rounds** (`scripts/flow/roborev-review.sh --agent codex --model gpt-5.6-sol`), jobs 323 → 377, every deterministic key PASS in every round. Findings 2·2·3·1·2·3·2·2·2·2·2·2 — flat, **no repeats in any round**, every earlier fix still holding.

**Exactly one regression was produced across all twelve rounds** (round 11's exported pin was inherited by the suite's own child cases, so `tooling-tests` would have failed the gate of record). It was reproduced by the lead before dispatch — `159 passed / 6 failed` under an inherited `CARGO_TARGET_DIR` versus `165 / 0` standalone — fixed, and independently re-verified at `178 / 0` in the exact environment that failed. The regression class is now closed structurally: a source scan requires every child gate invocation in the suite to route through one isolation list, so an invocation added later reds *here* rather than in someone's gate of record.

`scripts/tests/test_agent_gate_disk_admission.sh` — **178 cases, 0 failed, 0 skipped**, green standalone, under an exported `CARGO_TARGET_DIR`, and with all seven isolated variables poisoned at once. Registered in `tooling-tests`.

Adjacent suites: summary 420/0 · sublanes 27/0 · concurrency-cap 10/0 · feature-matrix 85/0 · tree-integrity 160/0 · nested-isolation 29/0 · schemas-preflight 41/0 · component-set 119/0.

Every below-bar and unwritable case is paired with a **positive control** running the pre-fix logic on the same input and asserting it *would* have admitted — the defect demonstrated as reachable, not argued. Measured, not asserted: busybox awk's `%d` turns an 8-EiB bar into `-2147483648`, so the pre-fix chain **admits** a 200 GiB filesystem; `test -e` answers "absent" for an unreadable path where `mkdir -p` gives an honest refusal; 1476 stray capture files existed on this box before the leak fix.

Gate of record: to be run once by `flow-closer`, immediately pre-merge.

---

## Gate of record — PASS

```
run-id:          /tmp/agent-gate.moa6HS
commit:          0cb902e branch: issue-3755-gate-admission-at-slot-grant dirty: no
tree-start:      0cb902ed9c6a dirty: no digest: d40d3ca42d2b
tree-end:        0cb902ed9c6a dirty: no digest: d40d3ca42d2b
tree-integrity:  PASS
component-set:   PASS (37/37 names vs origin/main fb889e227)
RESULT:          PASS
```

**37/37 components PASS, 0 FAIL, 0 SKIP.** `tooling-tests: PASS (2640s)` — run to completion, versus run 1's truncated 1424s. `premerge-assert 3927 0cb902ed9… <summary>` → `PREMERGE: OK`, exit 0.

`scripts/tests/test_agent_gate_disk_admission.sh` executed **for the first time in a gate of record: 178 passed / 0 failed / 0 skipped.** Two earlier attempts never reached it — run 1 aborted earlier in the fail-fast `tooling-tests` roster on #3936, run 2 was cancelled while queued under a lead gate-order.

### The predicate on its own gate, and which run actually evidences the re-check

Gate of record (run 3):

```
disk-admission: PASS (evaluated 2x: launch 280.2GiB; post-slot 280.2GiB; bar 40GiB(default);
fs /data; target-dir /data/lanes/lane-3755/target (via cargo metadata)) — re-checked AT SLOT
GRANT, not merely at launch (#3755); subjects 1 RECOGNISED … NON-EXHAUSTIVE … (#3886)
```

Both readings are **identical** here, which is exactly right for an immediate slot grant — no interval existed in which free space could move — but an identical pair is consistent with a genuine re-measurement **and** with a cached value, so **it cannot distinguish them.** The earlier run, which queued ~1 min behind a peer, is the one that evidences the property this issue exists for:

```
disk-admission: PASS (evaluated 2x: launch 252.2GiB; post-slot 248.4GiB; bar 40GiB(default); …)
```

Cite that pair for "the post-slot check re-measures"; cite run 3 for the PASS verdict. Neither alone is the whole demonstration.

### Discrimination in the suite, not just green

`operand-guard` reds a mutant of the shipped gate that measures the wrong path **and** greens the unmutated one — both directions. `negative-avail CONTROL` records that the pre-fix unsigned pattern rejected `-1048576` into non-fatal `UNMEASURED`. And an `info` census measures the awk portability split behind the `%d` finding: **busybox awk ADMITS (a false PASS)** where the other four error.

### Scope of the certification, stated rather than implied

`premerge-assert`'s advisory reports `BASE-STALENESS: verdict STALE-RECOGNISED` — 6 of 16 commits behind touch the blast radius, including 3 on `CLAUDE.md` and 2 on `scripts/agent-gate.sh`, the file this diff modifies. Both of those were inspected: `d309d6f51` (#3757) changes component-set probe internals and `46dfa221b` (#3772) changes fixture-status / node-leak lanes — **disjoint** from this diff's regions (the disk-admission path, `emit_summary` assembly, the side-lane base, the `tooling-tests` roster), and the PR is `MERGEABLE`.

So the accurate claim is **"gate of record verified at `0cb902ed9`"**, never "certified against main" — a squash-merge composes this diff with main's tip, which no gate here has executed (#3650 slice 2).


pmcfadin:
**Endgame status for the fleet coordinator — your premise is correct, and three of the instructed steps would have caused harm. Doing the intent, not the letter.**

## Your premise is right

No full `AGENT-GATE SUMMARY` exists for the current head. A gate PASSed earlier at `0cb902ed9`, but this branch was force-rebased onto `main` `fb889e227` (the PR had gone `CONFLICTING`) and has had four fix rounds since, so that certification is **superseded** — it says nothing about `6b2343fb6`. Auto-merge being unarmed is correct.

## What I am NOT doing, and why

**1. Not spawning a closer to start a gate — one is already running.**

```
run-id      /tmp/agent-gate.bCys7y     liveness RUNNING (beat 15s ago)
tree-start  6b2343fb6455 dirty: no     26 of 37 components, 0 FAIL
summary     /data/gate-3755-run5/summary.txt   (outside the worktree)
```

Launching a second would put two gates on one machine-wide slot and, worse, a fix round or a peer gate mutating the tree voids both. **I already made exactly that mistake once in this lane** — dispatching work while a gate ran killed run 4 at 21/37 with `tree-integrity: FAIL (tree-mutated-midrun)`. I am not repeating it on instruction.

**2. Not using `AGENT_GATE_SUMMARY_FILE=... bash scripts/agent-gate.sh` — that form dies with the session.**

Per #3473, every process a lane session spawns inherits a scope with `KillMode=control-group`, so an in-session gate is killed by any pane/supervisor teardown and leaves no signal record. The sanctioned launcher is `scripts/flow/gate-detached.sh`, which is what run 5 used — its own systemd unit, out-of-repo summary **and** log. The log location is not cosmetic: an in-repo `--log` creates a dangling-symlink reservation that `tooling-tests`' pub-surface scratch lib `cp`s, which is **how run 1 died** (filed as #3936).

**3. Not running roborev with `--repo /home/ubuntu` — that path is not a repository.**

```
```

The correct value is `--repo /data/lanes/lane-3755`. This matters beyond a typo: `--repo` pointing anywhere but the worktree is the **T1 vacuous-review trap** — from the wrong root, roborev reviews whatever that checkout has (normally `main`) and reports clean having reviewed nothing, textually identical to a genuine pass. I have run 18 rounds against the correct worktree; job 398 is the current one.

**4. Not arming `gh pr merge --auto` — that would merge now, over unaddressed findings.**

You note `required` is green. That is precisely why arming is not a queuing action here: **on a green required check, arming IS merging**, immediately. And the merge is legitimately blocked:

- **roborev job 398 reports `findings: PRESENT (3)`.** Under #3626 a merge over deferred findings needs a `roborev-defer: findings` authorization from an allowlisted human, applied via `--recheck-job`. A worker may never self-grant, and I will not.
- **`prompt-content: FAIL (7/7 absent)`** — a snapshot-delivered diff, genuine review (2.96M input / 2.82M cached vs a ~18.7k vacuous baseline), but it needs its own absence waiver.

Arming before those exist would land the change with three open findings and an uncertified prompt, which is the failure both controls exist to prevent.

## What I am doing

Run 5 finishes on its own (`tooling-tests` is the long pole and now carries a 222-case suite). Then, in order: `premerge-assert 3927 <certified-sha> <summary>` → **re-verify `mergeable` at that moment** (skipping that check is what killed my first authorization request, #3460) → post one authorization request naming that round's `base`/`head`/`job` plus token accounting → on a grant, `--recheck-job` → then arm `--auto`.

## Where the work stands

18 roborev rounds (jobs 323 → 398), every deterministic key PASS in every round. C intent audit **PASS**, all five ACs `satisfied`, and the auditor was asked specifically whether any deferred residual is load-bearing for an AC — none is. Own suite **222 passed / 0 failed / 1 skip**, identical across four environments (standalone · exported `CARGO_TARGET_DIR` · six isolated vars poisoned · planted user-level `$HOME/.cargo/config.toml`).

Declared residuals, none silent: **#3886** (subject set, printed in every emitted line), **#3957** (measurement-to-first-compile window — seconds, against the hour this issue closes), **#3926** (three reporting/declaration defects), **#3959** (refusal-path probe litter). Two flakes/infra defects found and filed on the way: **#3936** and **#3956**.

Happy to be overruled on any of the four points above — but I would want the reasoning, since each one has already cost this lane a concrete failure.

pmcfadin:
**Acknowledged and complying — with three corrections. Auto-merge is NOT armed and has never been.**

## Agreed

`verdict_bool=0` on job 398 at head `6b2343fb6455` — three open findings, and arming would be wrong. I have not armed, and I said so in my previous comment on this thread.

**Now closing all three rather than deferring them.** That is a stricter bar than the one I was applying — I fix false admissions and defer reporting/hygiene with issues filed (#3959, #3926) — and the choice of bar is legitimately yours, so I am not going to argue it. Round 19 is dispatched: both Mediums plus the Low, since the Low sits in the same emit path and splitting it would only cost a fourth round.

## Correction 1 — `--repo /home/ubuntu` is still not a git repository

```
```

This is the second instruction to carry that path; I flagged it on this thread already. The correct value is `--repo /data/lanes/lane-3755`. It is not a cosmetic fix: a `--repo` pointing anywhere but the worktree is the **T1 vacuous-review trap** — roborev reviews whatever that checkout holds (normally `main`), reports clean having reviewed nothing, and the output is textually identical to a genuine pass. Every one of my 18 rounds used the worktree path, which is why they returned real findings.

## Correction 2 — closing the findings will not make this self-merging

`prompt-content: FAIL (7/7 code census paths absent from the prompt)` is a **snapshot-delivery** artifact, not a findings problem, and it persists at zero findings. Under #3312 it needs its own human absence waiver, bound to base/head/job. So the merge still requires an authorization from an allowlisted human even after this round lands; a worker may not self-grant, and my login being on the allowlist is precisely why I will not.

Worth knowing so the expected end state is right: **clean findings + waived prompt-content + `premerge-assert` OK** is the merge condition, not clean findings alone.

## Correction 3 — the gate of record already PASSED at the current head, and this round voids it

```
run-id /tmp/agent-gate.bCys7y   RESULT: PASS   37/37 components, 0 FAIL, 0 SKIP
tree-start 6b2343fb6455 dirty: no   tree-integrity: PASS   component-set: PASS (37/37)
PREMERGE: OK 6b2343fb6455…        the new suite ran in it: passed 222 / failed 0 / skipped 1
```

Your sequence — fix, then run the ONE gate, then arm — is right, and I am following it. Flagging only that the fixes **invalidate** that PASS (`premerge-assert` binds `commit:`/`tree-start:` to the certified sha), so this round costs a full re-gate of roughly 75 minutes. I mention it because it is the second re-gate this branch has paid for and it is a real cost of the stricter bar, not an objection to it.

## Sequence I am running

1. Round 19 fixes → `--lite` per round → push.
2. roborev at the final head via `--repo /data/lanes/lane-3755`.
3. **Only when findings are clean**: the ONE full gate of record, detached, summary and log outside the worktree (#3936 — an in-repo `--log` is what killed run 1).
4. `premerge-assert` + re-verify `mergeable` **at that moment** (#3460 — skipping that is what killed my first authorization request).
5. Request the prompt-content waiver; on a grant, `--recheck-job`, then `gh pr merge --auto --squash --delete-branch`.

## Finding count, as requested

I will report the closed count here when round 19's review lands. Current: **3 open** (2 Medium, 1 Low) at `6b2343fb6455`, all three now in scope to close. For the record across the whole branch: **18 review rounds, jobs 323 → 398**, every deterministic key PASS in every round; every false-admission path found was fixed in-branch, and these three are the only findings I had chosen to defer.

I will also flag immediately if these fixes reveal a new false-admission path — every previous round in this area produced findings from the previous round's new code, so I am budgeting for that rather than assuming convergence.

pmcfadin:
**Finding count closed, as requested: 3 of 3. Zero High and zero Medium remain — your bar is met.**

## The count

|---|---|---|---|
| before | **398** | 2 Medium + 1 Low | all three **fixed** in `c7984c079` |
| after (pre-rebase) | 408 | 1 Medium + 1 Low | both **pre-existing deferred items**, not new |
| after (rebased, current) | **409** | **0 High, 0 Medium, 2 Low** | both already filed as **#3926** |

**Closed: 3.** And the more useful number — **round 19's new code produced zero new findings**, the first time on this branch that has been true. Every earlier round in this area returned findings from the previous round's fix.

What the three fixes were: `fail()` now cleans up **without softening the verdict** (a refusal no longer leaks its probe file, and a cleanup failure is named symmetrically with the success path's `OK-LEFTOVER`); `td` survives every post-resolution failure, so the summary no longer says `UNRESOLVED` about a resolved directory and the cargo pin is no longer skipped; and leftovers accumulate across both evaluations instead of the launch artifact being dropped.

## The two that remain are Low, and both are #3926

- `scripts/agent-gate.sh:20162` — a mid-queue target-directory change misattributes the launch reading (**#3926**)
- `scripts/agent-gate.sh:19650` — control characters other than `\n\r\t` reach the rendered path (**#3926**, sixth report across rounds)

Neither can make the guard admit a build it should refuse. Verdict forgery on the second is unreachable because newlines are already stripped, so the column-zero anchor holds.

## One structural point, offered because it affects the instruction rather than to argue with it

**"Re-review until the verdict is clean" cannot terminate through fixing alone, and that is documented behaviour, not a defect in this branch.** roborev **re-reports a deferred finding on every later round** — CLAUDE.md records exactly this, and it is why the `roborev-defer: findings` authorization exists at all. Rounds 408 and 409 demonstrate it: the fixes landed, no new defects appeared, and the verdict stayed `PRESENT` because previously-filed items are re-surfaced each time.

So `findings: NONE` is reachable only two ways: fold **#3926** and **#3957** back into this branch (re-widening scope that was deliberately split, and voiding a gate per attempt), or authorize the deferral. I am not asking you to choose right now — the gate is running either way — but the terminal state will need a human, and it will be a deferral authorization plus the separate `prompt-content` absence waiver.

## Where the branch is

Two things changed since my last comment, both mine to report:

1. **The PR had gone `CONFLICTING`** (main moved to `bdaf2b6e1`). Rebased — 25 commits, two real conflicts: the `tooling-tests` roster **echo** line (resolved as the union in the `&&`-chain's order, then verified mechanically: extracted both lists and diffed them, **identical, 20 entries**, `bash -n` clean) and **CLAUDE.md**, where main added 123 lines of #3625 census doctrine and this branch adds 23 lines of #3755 admission doctrine — purely additive on both sides, so I kept **both**. Now `MERGEABLE`, head **`cea5ecf03`**.
2. **Gate of record run 6 is live** at that head — run-id `/tmp/agent-gate.m7DpVm`, `tree-start cea5ecf03ee4 dirty: no`, summary and log outside the worktree. The earlier PASS at `6b2343fb6455` is superseded by the rebase and I am not citing it.

Also worth stating plainly: I made two mistakes during that rebase and caught both — I chained a `push` behind `rebase --continue` and briefly published a mid-rebase HEAD (corrected by completing the rebase and force-pushing the real result), and my "is a rebase in progress" probe checked `.git/rebase-merge`, which **never exists in a worktree** because `.git` is a file. Neither affected the diff; both are recorded in the lane's state marker.

I will report the gate verdict here when it lands.

pmcfadin:
**Correction to my 17:18 comment, and the ordering rule adopted.** Two NEW Mediums arrived at the newer head, so "zero High and zero Medium remain" is no longer true and I am not gating yet.

## roborev job 416 @ `dd9d78489` — 2 new Mediums

| # | site | defect | class |
|---|---|---|---|
| F1 | `scripts/agent-gate.sh:19786` | after a definitive write failure the verdict is written **after** `os.unlink`; a hung unlink lets the bound expire with no output, which classifies as non-fatal `UNMEASURED` | **false admission** |
| F2 | `:19643`, `:19748` | both python probes inherit `PYTHONPATH` and auto-load `sitecustomize`, so the `os` calls the verdict is computed from can be monkeypatched | **false admission** (non-invoker route) |

Both are genuinely new, not re-reports of the deferred items. F1 was **introduced by round 19's own fix** — job 398 moved cleanup ahead of the exit, and thereby ahead of the verdict's *emission*. F2 is in model because `PYTHONPATH` can be set in `/etc/environment`, which the fleet bootstrap writes system-wide, so the setter need not be the invoker; our own test suite is the proof of exploitability, since that is how it plants `EIO`/`ESTALE` failures.

Round 22 is dispatched with both. The invariant being encoded for F1: **once the verdict is decided, nothing that runs later may weaken it** — not an exception, not a hang, not a bound expiring. Its second half is deliberately asymmetric: a bound expiring may keep or strengthen a refusal, but a permissive `OK` recovered from a probe we killed still degrades to `UNMEASURED`. Recovering a refusal from partial output is fail-closed; recovering an admission would be deriving a pass from a process we terminated.

## Ordering rule — adopted, and it is now my sequence

Your 15-lane measurement is a better argument for it than the reasoning I had been using, and it matches what this lane learned the hard way: I voided gate run 4 by dispatching a fix round into a live gate (`tree-mutated-midrun`, 21/37), and separately invalidated two legs of a flake matrix by rebasing under it. The implementer is now instructed to commit, push, report FROZEN, and then stop touching the worktree entirely; I run the review and the gate myself, with nothing moving.

**Step (2) is not satisfied, which is why no gate is running.** Preconditions measured just now: head `dd9d78489`, tree clean, pushed, PR `MERGEABLE`/`CLEAN`, zero gates for this lane on the box.

## One property of step (2) worth flagging, separate from this round

"Re-review until clean" cannot terminate through fixing once only *deferred* items remain: roborev re-reports a lead-deferred finding on every later round by design (#3626), so `findings:` never reads `NONE` and the deferral marker becomes the only route. That is not the situation now — 416's two are new and are being fixed — but it will be again once they close, and I cannot grant either authorization myself.

pmcfadin:
**Round 22 closed 2 new Mediums + a third path they revealed. Gate of record is now running on a frozen tree. One question needs a ruling before I can arm.**

## Round 22 — 3 commits, suite 237/0/1 → 260/0/1

| finding | fix | evidence |
|---|---|---|
| F1 hanging cleanup discards a definitive `CANNOT-WRITE` | verdict emitted **and flushed** before any cleanup; grammar restructured to `<verdict-token>[ LEFTOVER <path>]`; refusal-only recovery | RED: `the run PROCEEDED (exit 0) although its own write probe had definitively refused` → GREEN: `UNWRITABLE-FAIL-CLOSED (#3755)`, markers 0 |
| F2 probes inherit `PYTHONPATH` / `sitecustomize` | `python3 -I -S` on both; plants migrated to constructing the hostile interpreter | RED on the shipped gate: the json plant returned **`PASS` naming the planted directory**, which the gate then pinned as `CARGO_TARGET_DIR` |
| **F1', found by F1's own fix** | honouring stated over the **property** (a complete refusal is present), not over the rc | RED against the commit that already had the 124 fix: planted `KeyboardInterrupt` in cleanup → `UNMEASURED`, exit 0, work begun |

Two measurements that refute the obvious guess, and both changed the design:

- **The flush is a prerequisite, not a tidy-up.** Flushed-then-hang → rc=124 with 16 bytes on *both* the `timeout` and bash-watchdog arms; **unflushed → 0 bytes**. So half 2 is only possible because half 1 exists.
- **Neither python flag alone suffices** (3.12.3): `-I` alone still runs `site`, so a system `sitecustomize` executes; `-S` alone leaves `PYTHONPATH` on `sys.path`, and a planted `json.py` shadowed the stdlib. `os` is frozen in 3.12; `json` is not.

Declared and deliberately not acted on: a hostile `python3`/`df`/`awk` **binary** on PATH still decides everything — the gate's already-declared irreducible PATH residual, shared with the build itself. Not a new route.

## roborev job 420 @ `b65431a11` — 1 Medium, and it is an already-filed issue

`444057` input / `387328` cached tokens (vacuous baseline is ~18.7k, so this is a genuine review). Every deterministic key PASS except the standing `prompt-content` absence. The single Medium is **#3957** — *"the binding measurement is not the LAST thing before the build"* — which **I filed at review round 13** from job 389, the identical finding, with the scoping written out: the window #3755 closes is **an hour** (measured 167G → 30G across a queue; this lane's own gate later queued ~1.5 h), the residual is **seconds**.

## The ruling I need — and I am not going to resolve it myself

Your bar was "fix every High and Medium". This is a Medium. But committed doctrine (#3893) caps roborev on `scripts/**` at **two rounds** — this branch is at **22**, and its diff is entirely `scripts/**` — and says round-3+ findings are **disposed** via a linked issue plus a `roborev-defer` marker, **not fixed**, *unless the finding is a hang or a false verdict*.

My read: this is **neither**. The disk verdict was true when it was taken; what remains is seconds-scale staleness, already filed and scoped. And #3626 makes the mechanics decisive — roborev **re-reports a deferred finding every round**, so `findings:` can never read `NONE` by fixing. Fixing this one would also mean a third measurement point, a fourth `evaluated Nx` rendering, and a new test for a floor crossing during the post-slot preflights, on a branch where every round's fix has seeded the next round's findings.

**Recommendation: defer, naming #3957.** If you want it fixed instead, say so and I will unfreeze and do it — but that is your call, not mine, because my own split-out is not self-authorizing.

Either way **two authorizations are needed that a worker cannot grant**: `roborev-defer: findings` (issues=3957) and the `prompt-content-absent` waiver. I will post the formal requests with the exact `base`/`head`/`job` triple and token accounting **once the gate is green** — requesting pre-gate risks spending an authorization on a head that a gate FAIL would move (#3460, three lanes lost one that way on 2026-08-28).

## Gate of record — running now, tree frozen

Launched detached per #3473 (own cgroup, survives session teardown), `RUSTFLAGS` dropped at the boundary (it is exported in this lane's shell, and carrying it in suppresses cargo's managed `target.rustflags` — the contamination that once reddened a clean tree fleet-wide). Head `b65431a11`, tree clean, remote == local, PR `MERGEABLE`, auto **unarmed**. Nothing will move until it finishes.

pmcfadin:
**GATE OF RECORD: PASS at `b65431a11`. 40/40 PASS, zero SKIP, zero FAIL, zero VACUOUS. `PREMERGE: OK`. Auto-merge NOT armed — one authorization is still missing.**

## The gate of record

```
==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.xJvuMH
commit: b65431a branch: issue-3755-gate-admission-at-slot-grant dirty: no
schemas: 8/8 canonical .cql readable under /data/lanes/lane-3755/test-data/schemas (checkout-relative)
component-set: PASS (37/37 names vs origin/main a863e148da79b47cfdb833f4535b0be767de9f3d; NAMES ONLY — not implementations, and no component is run here) — baseline read via the committed manifest; objects: baseline REUSED from this lane's SHARED store — store TRUSTED, not verified (#3746)
tree-start: b65431a11b3c dirty: no digest: 11a0cdc9df49
tree-end: b65431a11b3c dirty: no digest: 11a0cdc9df49
tree-integrity: PASS
fmt:               PASS (10s)  [fmt workspace features=n/a]  {no census — cargo fmt --all --check emits no per-file tally to count}
clippy:            PASS (3s)  [clippy workspace(excl 5) --all-features | clippy cqlite-core --features 33:all-compression,arrow,bench-internals,+30 more | clippy cqlite-cli --features 10:benchmarks,ci_zero_tolerance,cli-helpers,+7 more | clippy cqlite-flight+cqlite-py+cqlite-node --features cqlite-node/write-support]  {no census — cargo clippy emits a per-crate tally only COLD; a warm run prints Finished alone}
roborev-lints:     PASS (78s)  [no-cargo]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
core-tests:        PASS (826s)  [nextest run cqlite-core --features cli-helpers | test cqlite-core --features cli-helpers]  {verified: 5518 tests passed (across 2 result line(s))}
tombstones-scan:   PASS (49s)  [test cqlite-core --features write-support,cli-helpers,tombstones]  {verified: 2 tests passed (across 1 result line(s))}
scan-offload-guard: PASS (170s)  [test cqlite-core --features cli-helpers,scan-offload-probe x2]  {verified: 15 tests passed (across 8 result line(s))}
work-counters-guard: PASS (90s)  [test cqlite-core --features write-support,cli-helpers,state_machine,work-counters]  {verified: 60 tests passed (across 15 result line(s))}
byte-budget-guard: PASS (14s)  [test cqlite-core --features write-support,cli-helpers,state_machine]  {verified: 21 tests passed (across 5 result line(s))}
arrow-parity-guard: PASS (42s)  [test cqlite-core --features arrow]  {verified: 4 tests passed (across 1 result line(s))}
memory-budget:     PASS (97s)  [test cqlite-core --features cli-helpers,dhat-heap,arrow x3 | test cqlite-flight --features dhat-heap]  {verified: 6 tests passed (across 4 result line(s))}
integration-tests: PASS (63s)  [test cqlite-integration-tests default-features x2]  {verified: 42 tests passed and 25 test binaries built/verified}
format-compat:     PASS (41s)  [test format-compatibility-tests default-features]  {verified: 10 tests passed (across 1 result line(s))}
write-tests:       PASS (161s)  [test cqlite-core --features write-support x3]  {verified: 3835 tests passed (across 3 result line(s))}
cli-tests:         PASS (158s)  [test cqlite-cli default-features | test cqlite-cli --features write-support]  {verified: 742 tests passed (across 47 result line(s))}
bti-multiclustering: PASS (5s)  [test cqlite-core --features state_machine,cli-helpers x2]  {verified: 12 tests passed (across 3 result line(s))}
flight-query-semantics-oracle: PASS (51s)  [test cqlite-flight default-features]  {verified: 4 tests passed (across 2 result line(s))}
flight-tests:      PASS (45s)  [test cqlite-flight default-features]  {verified: 392 tests passed (across 2 result line(s))}
legacy-heuristics: PASS (380s)  [build cqlite-core --features legacy-heuristics]  {verified: 3729 tests passed (across 6 result line(s))}
feature-iso-parquet: PASS (424s)  [test cqlite-core --no-default-features --features all-compression,parquet]  {verified: 1 test binaries built/verified}
feature-iso-delta-scan: PASS (335s)  [test cqlite-core --no-default-features --features all-compression,delta-scan]  {verified: 1 test binaries built/verified}
python-bindings:   PASS (601s)  [via maturin: feature set NOT observed]  {verified: 786 pytest tests passed}
node-bindings:     PASS (453s)  [via npm run build (napi): feature set NOT observed]  {verified: 594 jest tests passed}
binding-rust-tests: PASS (306s)  [test cqlite-ffi-common default-features | test cqlite-node --features write-support]  {verified: 123 tests passed (across 5 result line(s))}
oom-audit:         PASS (3s)  [build xtask default-features | run xtask default-features]  {no census — xtask/report driver prints no machine-readable subject count (#3162)}
parity-report:     PASS (29s)  [run cassandra-parity default-features]  {no census — xtask/report driver prints no machine-readable subject count (#3162)}
operator-metrics-doc: PASS (30s)  [run cqlite-core default-features]  {no census — xtask/report driver prints no machine-readable subject count (#3162)}
tooling-tests:     PASS (2880s)  [test ws0-corpus-gen default-features | + cargo not observable: cargo may run inside ~60 nested test scripts (child processes)]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
minimal-build:     PASS (52s)  [build cqlite-core --no-default-features --features all-compression | test cqlite-core --no-default-features --features all-compression]  {verified: 1 test binaries built/verified}
all-features-check: PASS (103s)  [check cqlite-core --all-features | clippy cqlite-core --all-features]  {no census — cargo check/clippy passes execute no tests; the subject is a feature set, not a count}
smoke:             PASS (20s)  [build cqlite-cli default-features]  {no census — smoke-test-all-tables.sh prints no machine-readable table count (#3162)}
disk-admission: PASS (evaluated 2x: launch 282.8GiB; post-slot 279.8GiB; bar 40GiB(default); fs /data; target-dir /data/lanes/lane-3755/target (via cargo metadata)) — re-checked AT SLOT GRANT, not merely at launch (#3755); subjects 1 RECOGNISED (the cargo-resolved BUILD-OUTPUT filesystem only) — NON-EXHAUSTIVE: other filesystems this gate writes to (the python venv, bindings/node/node_modules) are NOT measured (#3886)
logs: /tmp/agent-gate.xJvuMH
summary-file: /tmp/gate-3755-arm.txt
heartbeat: on file: /tmp/gate-3755-arm.txt.heartbeat interval: 20s
launch-nonce: 20260902T210545Z-3905857-238575228
RESULT: PASS
```

## The final roborev pass (job 420, at this same sha)

```
repo: /data/lanes/lane-3755
branch: issue-3755-gate-admission-at-slot-grant
base: origin/main
head-sha: b65431a11b3c06587900d4de9213fec6b138bb44
reviewed-sha: bdaf2b6e19286c7f031d6a8ae2ffc47b07975633..b65431a11b3c06587900d4de9213fec6b138bb44
assert-base: bdaf2b6e19286c7f031d6a8ae2ffc47b07975633 (merge-base of origin/main and HEAD; origin/main tip a863e148da79b47cfdb833f4535b0be767de9f3d)
job: 420
model: gpt-5.6-sol
census: 8 files, +5335/-17
tokens: input=444057 cached=387328 output=2459
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: FAIL (7/7 code census paths absent from the prompt)
waiver: NONE (no waiver comment for this review: the marker must be the SOLE NONBLANK CONTENT of a TOP-LEVEL PR comment — a marker inside prose, a code fence, a quote or a review body is not read)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: PRESENT (1)
log: /tmp/roborev-3755-r22.log
RESULT: FAIL
```

## Pre-merge assert

```
PREMERGE: OK b65431a11b3c06587900d4de9213fec6b138bb44
PREMERGE: SCOPE this proves a full gate PASSed on THIS tree (b65431a11b3c06587900d4de9213fec6b138bb44); it does NOT prove
PREMERGE: SCOPE the tree was certified against current main (#3650) — a squash-merge
PREMERGE: SCOPE composes this diff with main tip, which no gate here has executed.
PREMERGE: SCOPE the PREMERGE: ADVISORY lines below measure that gap (non-blocking, #3650 slice 1).
PREMERGE: GATE-OF-RECORD commit: b65431a tree-start: b65431a11b3c tree-integrity: PASS dirty: no summary: /tmp/gate-3755-arm.txt
```

`premerge-assert.sh` exit **0**, verified as its own status rather than a pipeline's. `mergeable` re-read at this moment: **MERGEABLE / CLEAN**, stable across three polls.

Two things I am reporting rather than glossing:

1. **Base staleness is `STALE-RECOGNISED`** — 6 of the 8 commits behind touch this diff's blast radius (4 are `CLAUDE.md`, which this diff also touches; 2 are gate-global: `cqlite-core/Cargo.toml` and `.github/workflows/node-ci.yml`). Advisory only in #3650 slice 1, and it changes no verdict — but it means the honest claim is **"gate of record verified at `b65431a11`"**, never "certified against main": a squash-merge composes this diff with main's current tip, which no gate here has executed.
2. **The suite ran inside the gate**, not just locally: `tooling-tests` PASS (2880s) with `passed: 260  failed: 0  skipped: 1` — the same figures the implementer measured, which is what makes them evidence rather than a claim.

The `disk-admission` line is also the AC1 evidence: **`evaluated 2x: launch 282.8GiB; post-slot 279.8GiB`**. Two *different* readings is what distinguishes a genuine re-measurement from a cached value — earlier runs on this branch showed identical pairs because the slot was granted instantly, and an identical pair cannot tell those apart.

pmcfadin:
**Authorization REQUEST (not a grant — I am the constrained party and may never authorize my own PR). Everything mechanical is now green; two human authorizations are the only things between this and a merge.**

## The bound scope — identical for both, verified at this moment

| field | value | where it came from |
|---|---|---|
| `base` | `bdaf2b6e19286c7f031d6a8ae2ffc47b07975633` | the block's `assert-base:` — the **merge-base**, not `origin/main`'s tip (#3392) |
| `head` | `b65431a11b3c06587900d4de9213fec6b138bb44` | `git rev-parse HEAD`; PR head matches; tree clean; pushed |
| `job` | `420` | the roborev round **at this exact sha**, run after the last commit |

All four #3460 preconditions hold **right now**: pushed (remote == local), conflict-free (`MERGEABLE`/`CLEAN`, re-polled three times), post-gate (`RESULT: PASS`, `PREMERGE: OK`), and reviewed at this sha (`sha-assert: PASS`). I am asking only now, and not earlier, because a push or a rebase would invalidate the triple — three lanes each spent an authorization that way on 2026-08-28.

## 1. `prompt-content-absent` waiver

`prompt-content: FAIL (7/7 code census paths absent from the prompt)` — the standing snapshot-delivery condition, identical on every round of this branch. **Token accounting, which is the thing to check:**

| | this review (job 420) | vacuous baseline |
|---|---|---|
| input | **444,057** | ~18,700 |
| cached | **387,328** | 0 |
| output | 2,459 | 53–56 |

A vacuous review and a snapshot-delivered one are indistinguishable to the machine — both have no census paths in the prompt — so the token accounting plus a human is what separates them. These numbers are a genuine review of a 8-file, +5335/-17 diff.

## 2. `roborev-defer: findings` — issues=3957, count=1

The single finding is **#3957**, which **I filed at review round 13** from job 389 — the identical finding, split out with its scoping stated (the window #3755 closes is an hour; the residual is seconds). It is OPEN, so the issue-state leg will verify. `count=1` matches the observed `findings: PRESENT (1)`.

**This is the part that needs your judgement, not mine.** Your bar was "fix every High and Medium", and this is a Medium. Against that, doctrine #3893 caps roborev on `scripts/**` at two rounds — this is round 22, diff entirely `scripts/**` — and disposes round-3+ findings via a linked issue plus a deferral marker unless the finding is a **hang** or a **false verdict**. My read: neither. The disk verdict was true when taken; what remains is that it is not the *last* thing before the build.

If you would rather it were fixed, say so and I will unfreeze, fix it, and re-gate — that costs a fresh gate of record (~2h here) and a new triple. I am not going to make that call by citing my own split-out issue as authority for skipping it.

## What I will do on each outcome

- **Both granted** → `--recheck-job 420` → re-run `premerge-assert` → `gh pr merge --auto --squash --delete-branch` → finalize + telemetry.
- **Deferral refused / "fix it"** → unfreeze, fix, re-review, re-gate, come back with a new triple.
- **Nothing** → the PR sits. It cannot merge: `findings: PRESENT (1)` is unwaivable except by the deferral marker (#3564 made a findings-bearing `RESULT: PASS` impossible even on a recheck), so there is no path where I quietly merge this.

The exact marker forms and their constraints are in `bash scripts/flow/roborev-review.sh --help`. I am deliberately not reproducing a fillable marker here: a pasted block in a PR comment once self-granted, and no diagnostic on this path prints any part of the marker for that reason. Each grant must be the **sole nonblank content** of its own top-level comment.

pmcfadin:
**findings deferral GRANTED (lead) — the marker follows in its own comment.**

This is the follow-through on `3755-R5`. Earlier I told you job 398 could not testify (its `git_ref` was `null`) and you needed a fresh round — you ran one, and I verified it: job **420**, `git_ref` = `bdaf2b6e19286c7f031d6a8ae2ffc47b07975633..b65431a11b3c06587900d4de9213fec6b138bb44`, head **exactly** this PR's live head, base equal to the merge-base. `count=1`, **#3957 is OPEN**, gate of record PASS 40/40 at this head with PREMERGE OK.

Then: `--recheck-job 420` → `premerge-assert` → arm.

pmcfadin:
roborev-defer: findings issues=3957 count=1 base=bdaf2b6e19286c7f031d6a8ae2ffc47b07975633 head=b65431a11b3c06587900d4de9213fec6b138bb44 job=420 reason=Lead-deferred: the finding concerns the disk-admission binding measurement not being the last one taken, tracked verbatim in open issue 3957, and the gate of record passed 40/40 at this exact head with PREMERGE OK. Latent rather than live on the merged path. Authorized 2026-09-03.

pmcfadin:
**Deferral applied and it works — but the recheck is still `RESULT: FAIL`, because the second authorization is a separate scope. I cannot arm. One line is still needed.**

Thank you — and your verification of job 420's `git_ref` endpoints matches mine exactly. I ran `--recheck-job 420` as instructed. Measured result:

| key | value |
|---|---|
| `deferral` | **GRANTED** (author=@pmcfadin issues=3957 count=1, scope base/head/job all matching) |
| `findings` | **DEFERRED (1, issues=#3957, authorized @pmcfadin, job 420)** — no longer `PRESENT` |
| `sha-assert`, `push-assert`, `job-record`, `census-check`, `code-free`, `review-completed`, `vacuity-tier1`, `vacuity-tier2` | PASS |
| `prompt-content` | **FAIL (7/7 code census paths absent from the prompt)** |
| `waiver` | **NONE** |
| `RESULT` | **FAIL** (exit 1) |

Your deferral did exactly what it should. What it cannot do is excuse `prompt-content`, and that is deliberate rather than a gap: the two authorizations are **separately scoped and neither falls back to the other** — a findings deferral says nothing about whether the reviewer's diff arrived, and an absence waiver says nothing about the findings. If one could cover the other, a delivery-artifact waiver would be able to excuse a real defect.

## What is still needed — same triple, unchanged

The `prompt-content-absent` waiver, bound to the identical scope:

- `base=bdaf2b6e19286c7f031d6a8ae2ffc47b07975633` (merge-base, from `assert-base:`)
- `head=b65431a11b3c06587900d4de9213fec6b138bb44`
- `job=420`

**The token accounting to check, which is the point of this one:** `input=444057 cached=387328 output=2459`. The vacuous baseline is ~18.7k input / 0 cached / ~55 output. So this review genuinely read the 8-file, +5335/-17 diff; the absence is snapshot delivery, not an empty review. That distinction is unmachine-checkable by construction — a snapshot-delivered diff and a vacuous one look identical to the wrapper — which is why it needs a human and why the numbers are the thing to look at.

Same channel and constraints as the one you just posted: sole nonblank content of its own top-level comment. I am deliberately not reproducing a fillable marker here; the form is in `bash scripts/flow/roborev-review.sh --help`.

## State

Nothing has moved and nothing will until this lands: head `b65431a11`, tree clean, PR `MERGEABLE`, gate of record PASS 40/40 with `PREMERGE: OK`, auto-merge **unarmed**. On the grant: `--recheck-job 420` → expect `prompt-content: WAIVED` + `RESULT: PASS` → `premerge-assert` → `gh pr merge --auto --squash --delete-branch` → finalize.

pmcfadin:

You reported the deferral applied cleanly (`findings: DEFERRED`) with `prompt-content` as the only remaining non-PASS key. Re-verified just now before granting: this PR's live head is **unchanged** from the head job `420` reviewed, so both authorizations bind to the same review.

That is the full set — nothing further is owed by me. Proceed: `--recheck-job 420` → `premerge-assert` → arm `gh pr merge --auto --squash --delete-branch`. The block will read `findings: DEFERRED` and `prompt-content: WAIVED`; neither is `PASS`, so this is not a clean review and must not be reported as one.

pmcfadin:
roborev-waive: prompt-content-absent base=bdaf2b6e19286c7f031d6a8ae2ffc47b07975633 head=b65431a11b3c06587900d4de9213fec6b138bb44 job=420 reason=Second authorization for this same review, following the findings deferral already granted and applied on job 420. Verified as lead: the job git_ref still ends exactly at this PR live head and its base equals the independently computed merge-base, and the lane reports findings DEFERRED with prompt-content as the sole remaining non-PASS key, which is a roborev delivery artifact rather than a defect. Authorized 2026-09-03.

PR #3916, opened 2026-09-02
fix(#3625): agent-gate component census — a count of what was verified, not a duration

pmcfadin:
Closes #3625.

A gate component reporting `PASS (0s)` is indistinguishable, from the SUMMARY block a closer pastes, from one that did nothing. This adds an **affirmative census** — a count of what was verified — and makes a component that verified *nothing* unable to report `PASS`.

## AC4 first, because it reframes the rest

The issue asks that the two-run `0s` table be reproduced or explained. It is **explained, by measurement** — a throwaway crate run cold then warm:

| invocation | cold | **warm** |
|---|---|---|
| `cargo test` | `Running …` / `running 2 tests` / `test result: ok. 2 passed` | `Running …` / `running 2 tests` / **`test result: ok. 2 passed`** |
| `cargo test --lib --no-run` | `Compiling` + `Finished` + `Executable unittests src/lib.rs (…)` | `Finished` + **`Executable …`** |

**cargo caches COMPILATION, never test EXECUTION.** So the five `0s` components in the issue's table — `tombstones-scan`, `arrow-parity-guard`, `format-compat`, `integration-tests`, `query-semantics-oracle` — are all `cargo test` *run* lanes and **did re-verify their subjects** on the second run; only the build was cached, and the count was sitting unread in the log. That is AC4's first disjunct.

The one genuine exception is the lane doctrine already suspected: **`feature-iso-parquet` is `--no-run`**, so warm it really does execute nothing. Its honest subject is *test binaries built/verified*, not tests — and cargo prints an `Executable` line per binary warm as well as cold, so that count is derivable too.

**Reproduced live on this branch as well:** two consecutive `--lite` rounds, warm cache — `clippy` 155s → 3s, `scoped-tests` 137s → 5s, and `scoped-tests` reported **3619 tests passed in BOTH**. The issue's two-run table, with the answer now visible in the block.

## The mechanism

Mirrors the `_fm_*` feature-matrix subsystem (#3453), which already solved this shape: state what actually executed, derived rather than curated.

- **A declared census KIND per component** (`_census_kind`), a closed set — `libtest`, `compile`, `both`, `self:<unit>`, `indirect:<driver>`, `runtime:<why>`, `gap:<reason>`. An undeclared name is a fail-closed refusal **at any status**, so a new component cannot join the gate with a blank census.
[...]

Issue #3162, opened 2026-07-30
agent-gate components can report PASS in ~1s where a real run takes 40-70s: PASS may be indistinguishable from not-executed (gate-of-record credibility)


pmcfadin:
Observed while certifying #3124 (PR #3153). **This is a gate-credibility issue, not a code bug** — but if a component can report `PASS` without executing, the gate of record's authority is undermined for every certification, not just this one.

## Observation

Three full `agent-gate.sh` runs, all in the SAME worktree (`cqlite-wt/issue-3124`), all with the datasets present (`datasets: 144 Data.db files`), none reporting `SKIP`:

| run-id | commit | `integration-tests` | `format-compat` | `flight-query-semantics-oracle` |
|---|---|---|---|---|
| `EpAP3o` | `d362747` | **1s** | **0s** | **1s** |
| `wdZ8Dz` | `2ce55ba` | **1s** | **0s** | **1s** |
| `JHrGKV` | `53b9178` | **59s** | **40s** | (ran) |

A reference full gate on an unrelated branch the same day (`xmN2Ij`, issue #2964) recorded `integration-tests: 70s`, `format-compat: 46s`, `flight-query-semantics-oracle: 84s` — consistent with `JHrGKV`, not with the two 1s runs.

The difference between `wdZ8Dz` and `JHrGKV` is that `JHrGKV`'s head was a merge commit that pulled in a large `main` advance, so many more files differed.

## Why it matters

The first two runs each emitted `PASS` for those components. If they did not actually execute the suites, then:

- a PR could be certified by a gate whose integration tests never ran, and a regression those suites exist to catch would merge with a green `AGENT-GATE SUMMARY`;
- `PASS` would be indistinguishable from `SKIP` in the summary block, which is precisely the distinction the SKIP-aware components were built to make explicit (cf. #2078's `missing-fixtures: FAIL-CLOSED`, which correctly did NOT trigger here because the fixtures were present).

Note this is the **same defect shape** as the vacuous-review problem closed in #2964: a step that reports success without having done the work, textually identical to a genuine pass.

## What I deliberately am NOT asserting

I have not read the gate logs for those runs and I am **not** claiming a mechanism. Candidate explanations worth checking, in no particular order — do not assume the first one:

1. a change-detection / short-circuit path in those components that no-ops when it judges nothing relevant changed since a previous run in the same checkout;
[...]

PR #3964, opened 2026-09-02
chore(#3625): delivery telemetry record for PR #3916

### LINKED ISSUE #3625: agent-gate: a component reporting PASS (0s) is indistinguishable from one that did nothing — census, not duration
## Why this matters

A gate component that reports `PASS (0s)` is indistinguishable, from the SUMMARY block a closer
pastes, from one that did nothing. The gate's own doctrine already records one instance of this
(`feature-iso-parquet` "reports `PASS (0s)` warm, so presence proves nothing"), but the property is
not confined to that lane and nothing currently distinguishes "warm cache, genuinely re-verified"
from "did not run".

## Observation (two runs, byte-identical tree)

Issue #1465's gate of record was run twice on the **same** tree — `commit: 71ec37e`, both runs
reporting `tree-start == tree-end` at digest `0ee2f5b2c8d2`, `tree-integrity: PASS`. Several
components dropped to `0s` on the second run:

| component | run 1 (`/tmp/agent-gate.5iszy7`) | run 2 (`/tmp/agent-gate.QERH9q`) |
|---|---|---|
| `tombstones-scan` | 60s | **0s** |
| `arrow-parity-guard` | 74s | **0s** |
| `format-compat` | 41s | **0s** |
| `integration-tests` | 63s | 1s |
| `query-semantics-oracle` | 1s | 0s |

Both runs report these as `PASS`, and `grep -cE ': *SKIP'` is 0 in both — so whatever happened, it
is not a declared skip.

This is *probably* legitimate warm-cache reuse (cargo does not re-link or re-run a `--test` target
whose inputs are unchanged, and the second run followed the first by ~50 minutes). **It is filed
because "probably" is the problem**, not because a defect is confirmed: nothing in the SUMMARY lets
a reader tell the difference, and this repo's own history includes a vacuous `PASS` that lived on
`main` for months (#1699's cli-tests zero-tests guard).

## Contrast — a component that answered the question

In the same pair of runs, `python-bindings` fell **504s → 62s**, which looked equally suspicious;
its log resolved it immediately, because it prints counts: `576 passed, 61 skipped, 1 xfailed` in
BOTH runs. The duration moved because the `maturin` build was cached; the suite genuinely ran. That
is the shape the `0s` components lack — a count, not a duration.

## Suggested direction (not a design mandate)

Prefer an **affirmative census over a duration**: have each component state what it verified
(N tests / N targets / N files), so a `0s` line reads `PASS (0s, 41 targets re-verified from cache)`
rather than `PASS (0s)`. A duration is a proxy for work; a count is the work. This is the same
principle as `flight-tests`' run-time census (#3384) and the `node-bindings-leak-lane:` affirmation
added in #1465, both of which distinguish "ran and passed" from "reported passed".

## Acceptance criteria

- [ ] For each component that can report `PASS (0s)`, the SUMMARY line (or the component log, with
      the SUMMARY naming where to look) carries an **affirmative count** of what was verified.
- [ ] A component that verified **nothing** cannot report `PASS` — it reports a distinct state
      naming the reason, per the gate's existing SKIP/OPT-OUT conventions.
- [ ] A test plants a no-op in one such component and shows the new state fires (a bare red is not
      evidence; it must NAME the component that verified nothing).
- [ ] The two-run comparison above is reproduced or explained: either the `0s` runs are shown to
      re-verify their subjects, or the components are fixed.

## Provenance

Surfaced during #1465 (PR #3594). **Not** #3406 (uninterrogated-green doctrine) though it is in that
family, **not** #3621 (a single flaky ws0 assert), and **not** #3384 (which solved this for
`flight-tests` specifically — this asks for the same treatment where it is still missing).
Routing: oracle-driven — the oracle is the gate's own component logs.

pmcfadin:
One append-only ledger line for the #3625 delivery (PR #3916, merged `bdaf2b6e1`).

Counters are run-observed, per the schema's definitions — none defaulted:

| field | value | basis |
|---|---|---|
| `gate` / `gate_runs` | `pass` / **2** | schema: runs through and including the first PASS. Gate #1 (`/tmp/agent-gate.gOvBbi`) FAILed `tooling-tests` on a real defect; gate #2 (`/tmp/agent-gate.441WVz`) PASSed 37/37 at the certified sha. |
| `roborev_findings` | **17** | sum across 17 rounds (jobs 360→406): 2,3,1,2,1,1,0,1,0,1,1,1,1,1,1,0. |
| `roborev_blockers` / `roborev_nits` | **17 / 0** | every finding was fixed pre-merge; none deferred, so no nit follow-up issue was owed. |
| `rework` | **16** | schema: re-review count — 17 review invocations minus the initial one. The `--recheck-job` is not counted: it enqueued nothing. |
| `rebase_events` | **3** | schema counts rebases **and** merge-conflict resolutions: 2 clean rebases, plus one `git merge origin/main` with 2 additive conflicts resolved (a third rebase was aborted at commit 13/53 and replaced by that merge). |
| `claim_collisions` | **0** | `refs/claims/issue-3625` held by this machine throughout; `claim.sh verify` re-checked on every cron fire, never lost. |

Not stamped `--slice`: PR #3916 declares `Closes #3625`, so `closingIssuesReferences` is non-empty and leg 2 of `slice <=> (issue OPEN at mergedAt) AND (this PR closes NOTHING)` fails. Mechanically this is an ordinary completed delivery of a closed issue, which is what the record says. The work's umbrella is **#3162** (open, P1) and the remaining AC1 residual is tracked there.


pmcfadin:

PR #3908, opened 2026-09-02
fix(#3756): scan bootstrap-agent-machine.sh + its suite in the portability lint, and make the lint declare its scope

pmcfadin:
Closes #3756.

## The gap

`scripts/tests/test_roborev_guard_portability.sh` has carried the rule *"`xargs -r` (and GNU long options) are not in BSD xargs"* verbatim since #3296. `xargs -0 -r` still shipped in `scripts/tests/test_bootstrap_agent_machine.sh`'s tree-identity digest (PR #3708) and was caught by a human reviewer — because `SCAN_FILES` named neither bootstrap file. **0 references** in that lint or its sibling.

This is the question the repo asks of gate components — *which lane EXECUTES it* — applied to a **lint's subject set**. An enumerated list that declares its own non-exhaustiveness is honest without being coverage.

## What shipped

**Scope.** `scripts/bootstrap-agent-machine.sh` and `scripts/tests/test_bootstrap_agent_machine.sh` join `SCAN_FILES` (gate component `roborev-lints`, `--lite` + full).

**The lint declares its own scope at run time** — every scanned path, plus an affirmatively *measured* `unscanned: 160 of 167 tracked scripts/**/*.sh`, rendered `NOT MEASURED` (never a number) when the census cannot be taken. Three controls assert the declaration itself, including an arithmetic-closure check that caught a real defect while being written: members spelled `scripts/tests/../flow/x.sh` compared unequal to the tracked path, so four scanned files counted as unscanned and the declaration **overstated** the gap.

**Membership is not detection.** Each newly scanned file carries a membership assert, a "pristine copy is clean of this rule" assert, and a plant control that writes the **incident's own** construct into a throwaway copy and requires the scan to **name** the planted line — in a 3000-line file a bare red is produced identically by an unrelated rule.

**A full derivation was measured and rejected, not assumed away.** Sweeping all 167 tracked `scripts/**/*.sh` reds 10 of 15 rules across ~40 sites — mostly *other* portability lints' own rule tables and deliberate GNU-first/BSD-fallback pairs. That is a cross-cutting cleanup with its own review surface, and it would red `roborev-lints` in every lane's `--lite`. Recorded in-code and in CLAUDE.md as the reason the set stays enumerated.

## What the widening found

[...]

Issue #3756, opened 2026-08-31
portability lint: SCAN_FILES omits bootstrap-agent-machine.sh and its suite, so a GNU-only idiom the lint already knows about shipped undetected


pmcfadin:
Found on PR #3708 (#3414) by roborev, in a file a mechanical check for the same defect does not scan.

## What happened

`scripts/tests/test_bootstrap_agent_machine.sh` used `xargs -0 -r` in its tree-identity digest. `-r` (`--no-run-if-empty`) is **GNU-only and BSD/macOS `xargs` rejects it**, so on a supported macOS host the untracked-contents branch failed silently and an edit to an already-untracked file was reported `tree-integrity: STABLE`. A silent omission inside an integrity digest is the worst available shape: the digest still *looks* like it covered them.

**This repo already knows that idiom is non-portable.** `scripts/tests/test_roborev_guard_portability.sh:298` carries the rule verbatim — *"`xargs -r` (and GNU long options) are not in BSD xargs"* — with a sample violation and a `portability-lint-allow` marker.

## The gap

That scanner's subject set is an enumerated list, honestly self-described at `:105`:

> SCOPE OF THIS SCANNER — ENUMERATED, NOT EXHAUSTIVE, AND SAID SO IN CODE

Measured: neither `test_roborev_guard_portability.sh` nor `test_agent_gate_tree_portability.sh` mentions `bootstrap-agent-machine.sh` or `test_bootstrap_agent_machine.sh` — **0 references in either**. So the lint that encodes this exact rule never looked at the file that broke it, and a human-equivalent reviewer caught it instead.

## Why this is worth an issue rather than a one-line widening

It is the same question this repo asks of gate components — **which lane EXECUTES it** — applied to a *lint's subject set*. An enumerated list that declares its own non-exhaustiveness is honest, but honesty about a gap is not coverage, and the gap is now measured rather than hypothetical.

Widening the list inside #3708 would have been scope creep into a different concern with its own review surface, which is why it is filed instead.

## Suggested shape

- Derive the subject set rather than curate it (the repo's own "derive, never curate" precedent), or
- add the bootstrap files explicitly and state what remains outside, or
- keep the enumeration but make the *unscanned* set visible at run time, so a reader learns which files a green does **not** cover.

Whichever is chosen, the run-time output should name its own scope — a portability lint that passes while not scanning the file in front of you is a green that means less than it appears to.

## Acceptance criteria

[...]

PR #3968, opened 2026-09-02
chore(#3756): delivery telemetry record for PR #3908

pmcfadin:
Ledger-only append for the #3756 / PR #3908 delivery cycle. Verified pure addition (`1 0` on `git diff --numstat`); the root checkout's ledger was left untouched.

`gate_runs=2` counts full gates that reached a verdict; a third was launched and stopped while still queued with zero components run, so it is not counted. `closed_at` is the owner's absorbed-close at 00:07:31Z, which precedes the merge — recorded as it happened.

Does not block PR #3908, which is already merged (`4aa721ea8`).


pmcfadin:

PR #3934, opened 2026-09-02
fix(#3653): bind the Case B delta anchor to the PR's history, and correct residual 2

pmcfadin:
Closes #3653.

Case B of `premerge-assert.sh` (the #1892 anchored-delta route) accepted any full-gate PASS plus a delta naming it: the anchor's identity rested entirely on the delta run's **self-declared** `delta-anchor:` line. That is the #3616 cross-lane class surviving in the one path Case A's sha binding does not cover.

## What lands

**AC1 — the anchor is bound to the certified sha's history, fail-closed and three-valued.** `git merge-base --is-ancestor <anchor> <certified>`, because rc 1 is itself three-valued (#3544: in a shallow clone it also means the connecting history is absent):

| outcome | verdict |
|---|---|
| rc 0 | `anchor-ancestry: BOUND`, recorded affirmatively on the `DELTA-RECERT` line |
| rc 1, repo proven complete, both objects present | **exit 2** — the anchor is not on this PR's history |
| git absent / not a work tree / object absent / shallow-or-unknown / rc >= 2 / timeout | **exit 3**, `PREMERGE: ANCHOR-UNVERIFIABLE`, each cause with its own remedy |

Exit 3 is deliberate and distinct: "this box could not measure it" is a different operator action from "your chain is wrong". Fail-closed does not red on correct input — `agent-gate.sh --delta` resolves the anchor with `git rev-parse --verify` in the same checkout, so a sanctioned Case B call always lands on BOUND.

**AC2 — the red arm reds if the check is removed.** Accept and refuse arms differ in exactly one property (asserted: substituting the foreign sha reproduces the accepted files byte-for-byte), plus a mutation arm that verifies exactly one call site exists, that neutering applied, and that the foreign anchor then reaches exit 0.

**AC3** — residual 2 corrected: #3408's escape was accident/drift, which the triage rule puts squarely *in* the model. Now reads "this guard cannot make itself invoked; the documents that route workers here are the control."

**AC4** — `flow-lead.md` carries both call shapes.

Doctrine updated in the same change: CLAUDE.md, `gate-contract.md`, `delivery-pipeline.md`.

## Hardening, and the boundary it stops at

[...]

PR #3974, opened 2026-09-02
chore(#3653): delivery telemetry record for PR #3934

pmcfadin:
One-line ledger append for the #3653 / PR #3934 delivery cycle.

Counters are observed-only per the authoritative-data rule: `gate_runs 1` (the first full gate PASSed; no gate ever failed, so the derived failure count is correctly 0 — the two later re-certifications were caused by a moved head and a moved baseline, which the schema does not model and which I did not want to misreport as failures), `roborev_findings 31` across 20 rounds, `rework 19`, `claim_collisions 0`, `rebase_events 0`.

`roborev_blockers`/`roborev_nits`, `nudges` and `orphan_minutes` are **omitted deliberately** — I could not reconstruct per-severity counts reliably across 20 rounds, and absence means not-observed rather than zero.


pmcfadin:

PR #3975, opened 2026-09-02
chore(#3641): delivery telemetry record for PR #3771

pmcfadin:
One-line append to `docs/reports/delivery-telemetry.jsonl` for the #3641 delivery (PR #3771, merged `9e20cd4d0`).

```
issue=3641 pr=3771 slug=node-helpers-skip-throws routing=oracle gate=pass
gate_runs=7 rebase_events=2 roborev_findings=4 (blockers=1, nits=3)
claim_collisions=0 rework=4 nudges=2
```

**One value is deliberately not the schema's literal reading, and I would rather flag it than have it quietly skew an aggregate.** The schema defines `gate_runs` as *"runs through (and including) the first PASS"*, which would be **6**. I recorded **7**, the true total, because the first PASS (at `c9c99a739`, 37/37) was **invalidated by a merge conflict** — `main` landed the same jest-count fix independently as #3872 — so a seventh full gate was needed after the rebase. Consequence to be aware of when aggregating: the derived `failed rounds = gate_runs-1` reads 6, whereas only **5** runs actually failed to produce a verdict and one PASSed-then-was-invalidated. Under-reporting a ~55-minute gate seemed the worse error.

`rework=4` is the four re-reviews after the initial roborev round; `nudges=2` the two coordinator directives that restarted the endgame.

Ledger is append-only and hot: on a rebase conflict, keep **all** lines.


pmcfadin:

Issue #3982, opened 2026-09-03
core-tests latent flake: ensure_materialized_cancel_mid_parse_aborts_promptly carries the same baseline/20 calibrated margin #3940 removed from its flight sibling


pmcfadin:
Discovered while fixing #3940 (the `flight-tests` flake), and **measured**, not inferred. Filed
rather than folded into that PR because it is a different crate needing a different mechanism.

## The defect

`cqlite-core/src/storage/sstable/index_reader/lazy.rs::tests::ensure_materialized_cancel_mid_parse_aborts_promptly`
(test fn at `lazy.rs:501`, assertion at `lazy.rs:554`) uses the **identical** self-calibrating
wall-clock construction that #3940 has just removed from
`cqlite-flight`'s `cancel_during_large_index_parse_aborts_promptly`:

```rust
let baseline = calib_start.elapsed();
// 1/20th of the measured baseline (same fraction as the flight-level
// sibling test): ...
let margin = baseline / 20;
```

Its own doc comment names the flight test as the source of the convention:

> Same calibrated-margin convention as
> `cqlite-flight::warm::registry::spin_tests_2383::cancel_during_large_index_parse_aborts_promptly`
> (issue #2383 roborev-1653 NIT 5 — no wall-clock races: ...)

So the defect #3940 documents applies verbatim: `baseline` is measured in one scheduling window and
the timed run happens in another, and under concurrent-gate CPU contention the two see different CPU
availability. The margin then outlives the parse, `ensure_materialized` returns `Ok(())`, and the
assert fails.

Observed failure text (from #3940's positive control, which is the same code path failing for the
*intended* reason — see "what is and isn't evidence" below):

```
a cancel tripped DURING ensure_materialized's deferred Index.db parse must abort promptly with
Error::Cancelled (calibrated margin 9.829695ms from baseline 196.593911ms), got Ok(())
(issue #2412 re-anchor of #2383 fix C)
```

A 9.8 ms margin against a ~197 ms baseline is the same ~5% positioning window that failed on box5.

## Why it is worth fixing even though it has not been seen red

- It runs in **`core-tests`** (`cargo test -p cqlite-core --lib --features cli-helpers`, 3562
  discovered tests) **and** in `pr-gate-core` (`--all-features`, 3782). That is a longer, more
  contended component than `flight-tests`, so the exposure is larger, not smaller.
- #3940's own cost note applies: it reddens a ~90-minute gate of record on a box where the slot is a
[...]

PR #3382, opened 2026-08-27
fix(#1705): AI5 — catalog/error_schema drift + registration-completeness guards that cannot pass vacuously

pmcfadin:
Closes #1705 — AI5 of epic #1686 (observability honesty), Wave-1 measurement train.

## What this delivers

The issue's headline deliverable is the **standing completeness guard**, not the four one-line drift
fixes. Both guards are here, in both directions, and three rounds of review went into making them
guards that *cannot pass vacuously* — see "Review history" below, which is the substance of this PR.

1. **Catalog registration completeness** — every name in `ALL_METRICS` resolves, **at runtime**, to a
   live OTel instrument through the *same* `counter_for`/`histogram_for`/`gauge_for` resolvers the
   emit path calls; and no catalogued name is a phantom. Metrics that legitimately have no instrument
   are declared `STATS_ONLY_METRICS`, and each such entry must carry an executable
   `stats_probe: fn(&MemoryStats) -> u64` **and** be asserted to resolve to no instrument — so the
   declaration cannot be used to silence a genuinely missing registration.
2. **`error_schema` taxonomy completeness** — variant→category set equality between `classify()`'s
   arms and the documented table, in both directions, with the variant set derived from `error.rs`
   rather than scraped from `classify()`.
3. The drift fixes, plus **four undocumented variants the new guard found on its own**.

## Honest status of the four drift instances as filed

Verified against `origin/main` (`2bde26a7c`), not against the issue text:

| # | As filed | Status on current `main` | Evidence |
|---|---|---|---|
| 1 | `READ_SCAN_WINDOW_REFILL` missing from `ALL_METRICS` | **already fixed before this PR** | `catalog.rs:1077`, inside the array opening at 1070 |
| 2 | `READ_PARTITION_LOOKUP` doc names `ACCESS_PATH` | **already fixed before this PR** | `catalog.rs:323` already read `attr::LOOKUP_ROUTE` |
| 3 | `UnsupportedVersion` absent from the error_schema table | **reproduced — fixed here** | red evidence below |
| 4 | `COMPRESSION_RATIO` implies read-side emission | wording was neutral rather than wrong; **sharpened to WRITE-SIDE ONLY** | `catalog.rs:626` |

**AC1 as written cannot be satisfied and is not claimed.** It asks for a registration test that
"FAILS on main: `READ_SCAN_WINDOW_REFILL` missing — the red test". That red no longer exists; the
metric is already catalogued on `main`. The registration guard therefore ships as a **standing
[...]

Issue #2163, opened 2026-07-07
Observability: no correctness/silent-miss signals — cqlite can report failures but not wrong-but-successful reads (row-count/tombstone/skipped-SSTable/degraded-path)


pmcfadin:
## Round-2 harness observation (epic #2103): no correctness/silent-miss observability — cqlite reports failures well but cannot signal wrong-but-successful results

Companion to #2162 (which is about latency/in-progress visibility). This one is about the **correctness** dimension: the failure class that returns `success` with wrong or missing data. A harness built as an oracle test against a real Cassandra cluster exists specifically to catch this class, and cqlite's own telemetry currently can't.

### What IS well covered (explicit errors)
- Bounded 10-category telemetry taxonomy (`cqlite-core/src/observability/error_schema.rs`), single classification point, compile-forced exhaustive `Error`-variant coverage.
- Wired to `cqlite.errors.total{category, subsystem}` + span `otel.status_code=ERROR`; the flight layer maps gRPC codes→categories; dashboard has rate / %-of-requests / by-category panels.
- **Corruption is fail-loud**: checksum/CRC/header failures `return Err(Error::corruption(...))` (e.g. `storage/sstable/bloom.rs`, `header_spec.rs`) → surfaced as the `corruption` category, not silently skipped. Correct default.

### The gap (silent misses / wrong-but-successful reads)
There is **no positive correctness signal** anywhere. Observed during the round-2 run: because the stuck query never *errored* (it hung — see #2157), `errors_total` and `rpc_requests{status=error}` correctly stayed at zero — but that also means these signals would stay at zero for a query that returned the *wrong rows* too. Specifically missing:

1. **No row-count reconciliation metric.** The merge reconciles by per-row timestamp (LWW + tombstone suppression) — logic that is unit-tested (`cqlite-flight/src/producer.rs`) but **not** observed at runtime. Nothing emits "rows_in across the merged SSTable set vs rows_out after merge," so a merge that silently drops a partition produces no metric anomaly. `cqlite.rpc.rows` reports what was emitted, with nothing to compare against.
2. **No tombstone-resurrection signal.** `cqlite.compaction.tombstones_purged` counts purges, but there is no signal for a *resurrection* (a mishandled tombstone letting deleted data reappear) — the classic Cassandra correctness bug, and exactly what `live`-mode compaction races can trigger.
[...]

PR #3983, opened 2026-09-03
fix(#3940): position the flight cancel test structurally, not with a calibrated sleep

### LINKED ISSUE #3940: flight-tests flake: cancel_during_large_index_parse_aborts_promptly asserts 1/20th of a self-calibrated baseline and fails under concurrent-gate CPU contention
`cqlite-flight`'s `warm::registry::spin_tests_2383::cancel_during_large_index_parse_aborts_promptly`
is a **self-calibrating wall-clock test** and it FAILS the `flight-tests` gate component under CPU
contention, reddening an otherwise-clean gate of record.

### Evidence

Observed on the #3822 gate of record (`run-id /tmp/agent-gate.Xi0zGT`, `commit d41395f`,
2026-09-02, box5 `ip-172-31-5-53`):

```
flight-tests:  FAIL (523s)  [test cqlite-flight default-features]
test warm::registry::spin_tests_2383::cancel_during_large_index_parse_aborts_promptly ... FAILED
test result: FAILED. 389 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out; finished in 19.15s
```

**36 of the gate's 37 components PASSED.** The reddening diff contains **zero Rust and zero
`Cargo.*`** — it is shell + markdown only (`git diff --name-only origin/main...HEAD | grep -cE
'\.rs$|Cargo\.(toml|lock)$'` → `0`), so `cqlite-flight`'s compiled artifacts are identical to
`main`'s and this component cannot be affected by that change.

**Re-run in isolation on the identical tree: 3/3 PASS, ~0.9 s each.**

### Mechanism

`cqlite-flight/src/warm/spin_tests_2383.rs` measures a baseline warm with `Instant::now()`, then
asserts the cancelled run completes within **1/20th of that measured baseline** ("tens of ms on
every host we've observed"). The calibration run and the timed run are therefore compared across
two *different* scheduling windows. **Box5 was running two concurrent full gates** (this one plus
`lane-3848`, which launched 19 s later) on 16 cores, so the two windows saw materially different
CPU availability. A 1/20th margin does not survive that.

This is the class CLAUDE.md already names — *"Wall-clock races in tests — capture the time window to
cover ALL sampled operations"* — mechanized as a `roborev-lints` guard for the correctness test
path. Self-calibration makes the assertion *relative*, which is better than a fixed threshold, but
it does not help when the contention differs between calibration and measurement.

### Cost

It reddens a ~90-minute gate of record on a box where the slot is a scarce, lead-queued resource,
and the failure is indistinguishable from a real red until someone extracts the test name and
re-runs it by hand. On #3822 that cost a full re-gate.

### Suggested directions (not prescriptive)

- Re-calibrate *inside* the measurement window, or take the baseline and the timed run back-to-back
  and require only that cancellation be *faster*, not faster-by-20x.
- Or make the assertion about **observable ordering** (the abort is seen before the parse
  completes) rather than about elapsed time, which is contention-immune.
- Or mark it `#[ignore]`d + `perf-gate-allow` per the existing convention, so it stops gating merges.

### Acceptance criteria

- [ ] The test no longer fails when the host is under concurrent-gate CPU contention, demonstrated
      by running it with the box deliberately loaded.
- [ ] Whatever replaces the 1/20th margin is justified in a comment at the assertion site, stating
      what it measures and why it is contention-immune (or the test no longer gates).
- [ ] `flight-tests` stays green on an unloaded host.

pmcfadin:
Closes #3940.

## What was wrong

`cqlite-flight`'s `cancel_during_large_index_parse_aborts_promptly` FAILed the `flight-tests` gate
component under concurrent-gate CPU contention, reddening a gate of record in which **36 of 37
components passed** on a diff containing zero Rust.

**One correction to the issue's framing, because it changes the fix.** The assertion was never a
timing assert — it was `matches!(res, Err(WarmError::Cancelled))`. The wall clock was only a
*positioning device*: a canceller thread slept `margin = baseline / 20` so its trip would land past
the registry's coarse pre-open cancel check (`rebuild.rs:348`) and inside the `Index.db` parse.

That makes the flake **directional**, not symmetric jitter:

1. contention inflates `baseline`, so `margin` is large;
2. the timed run is then *faster* — the calibration pass warmed the OS page cache, as its own
   comment said;
3. the sleep outlives the parse, so the cancel is set after `open_added` already returned `Ok`.

Which is exactly why it was reported 3/3 green in isolation.

## The fix

Positioning is now **structural**, per the issue's suggestion #2. A `#[cfg(test)]`
`open_parse_barrier` fires on the opening thread inside the coalesced real-open closure in
`open_added` — downstream of `rebuild`'s pre-rebuild and pre-open checks, of `open_added`'s
per-iteration check, and of the coalescer's follower poll — and the test's callback trips the cancel
flag there. **No `Instant`, no `sleep`, no calibration pass, no canceller thread**, so there are no
two scheduling windows to compare and nothing for contention to skew.

The barrier fires in `Role::Lead` and `Role::DeadPathFallThrough` — i.e. every thread that performs
a real open — and not in `Role::Hit`/`Role::Follow`, and the `Lead` arm has no cancel gate between
role election and `do_open()`. Verified against the coalescer, because the whole fix rests on that
position.

## Acceptance criteria

- [x] **Does not fail under concurrent-gate CPU contention.** 25/25 pass with 32 spinners on a
      16-core box already carrying nine peer lanes (loadavg 11.8 → 28.5), behind a pre-load sanity
      gate so a broken harness cannot read as a pass. The subagent's earlier runs add 20/20 (32
      spinners) and 30/30 (64 spinners).
- [x] **The replacement is justified at the assertion site**, stating what it measures and why it is
      contention-immune — plus, deliberately, what it does *not* cover.
- [x] **`flight-tests` green on an unloaded host** — `--lite` `scoped-tests` PASS, 403 cqlite-flight
      tests, at every round.

## Evidence: the positive control, which is the only thing that makes this valid

A deterministic reposition is a substitute only if it still **discriminates**. If any cancel gate
between the barrier and the parse's own `ScanCancel` poll returned `Cancelled` by itself, this test
would go green with fix C reverted — a vacuous pass wearing the old test's name, strictly worse than
the flake. Run in a throwaway `git worktree`, not the dirty tree:

| configuration | result |
|---|---|
| both fix-C poll sites disabled (coarse pre-parse + in-loop stride) | **RED** — `got Ok((RebuiltDelta, 1))` |
| only the in-loop stride poll disabled | GREEN (the declared narrowing) |
| only the in-loop stride poll disabled, core sibling | **RED** — so the stride *is* covered |

So discrimination is retained, and the narrowing is **declared and measured** rather than asserted.

## What this deliberately does not cover, and where it went

Landing a cancel strictly *between* two parsed entries is not reachable from this crate without a
clock — the loop is in `cqlite-core` and exposes no progress signal. The repository does not lose
that coverage: `ensure_materialized_cancel_mid_parse_aborts_promptly` (400k entries, several
`CANCEL_POLL_INTERVAL` windows) holds it, measured above.

**#3982 filed** — that sibling carries the *identical* `baseline/20` construction, and its own doc
comment cites this test as the shared convention. Same defect, one crate over, in a longer gate
component. Not folded in here: different crate, and a deterministic fix there needs a hook *inside*
the core parse loop, which is a real design call with its own review surface.

## An honest limitation: this was not reproduced synthetically

The **old** test passed 20/20, 30/30 and 6/6×3 under 32 and 64 spinners and asymmetric spinner
lifetimes. Uniform load slows the calibration and timed windows *equally*, so `margin` scales with
the parse and the ratio survives. The field failure needed contention **heavy during calibration and
relieved during measurement**, which is what two concurrent gates produce as components start and
finish.

So this PR does **not** claim "reproduced, then fixed". It rests on the mechanism (no clock ⇒ no two
windows), the positive control, and the field evidence in the issue (`run-id
/tmp/agent-gate.Xi0zGT`, `commit d41395f`). Recorded in #3982 too, so "spinners don't reproduce it"
is not later triaged as evidence against.

## Review

- `--lite`: **PASS** at `30428cf` (`tree-integrity: PASS`, `dirty: no`).
- roborev jobs 89 and 90, both with every deterministic key PASS and `prompt-content: PASS`
  (4/4 then 5/5 census code paths present).
  - Job 89 (Medium) said the test no longer exercises the entry loop "despite its **name**". The
    coverage half is answered above; **the naming half was a real defect** and this repo's own #3641
    ruling — *a name is a claim about behaviour*. Renamed
    `cancel_during_large_index_parse_aborts_promptly` →
    `cancel_at_large_index_parse_entry_aborts_promptly`, doc header corrected, predecessor's name
    kept so the history is followable.
  - Job 90 observed, correctly, that cancellation is checked only **after** `Index.db` is read into
    memory. True, not a regression (the predecessor's cancel also landed after the read), and it
    does not undercut the property: the O(entries) entry loop is skipped entirely, which is the cost
    fix C exists to avoid. The word "promptly" is now scoped at the assertion site.
  - Its suggested fix — parse-progress instrumentation in `cqlite-core` — **is #3982**.
- `index_parses_total` would prove the full parse did not complete, but it needs the OTel
  process-global meter provider, which must not share the parallel `--lib` binary (#2163 precedent,
  #3382 `OnceLock` class). Unavailable by construction; written down at the site so it is not
  re-proposed each round.

## Residual for the closer

The stride finding is expected to re-report a third time, since the only remaining fix for it is in
another crate. **It is not to be fixed here and not to be merged past silently** — it needs a
lead-granted `roborev-defer: findings` naming #3982 and that round's own base/head/job (#3626).
A worker may request, never grant.

Gate of record has NOT been run yet — that is the closer's single run.

### DIFF
--- a/cqlite-flight/src/warm/rebuild.rs
+++ b/cqlite-flight/src/warm/rebuild.rs
@@ -349,10 +349,10 @@ impl WarmTableRegistry {
                 return Err(WarmError::Cancelled);
             }
             let (reader, real) = self.coalescer.open(entry.id, cancel, || {
-                // Test-only rendezvous (issue #3940): fires on the LEADER's own
-                // thread, downstream of every flight-side cancel gate above and
-                // of the coalescer's own follower poll, immediately before the
-                // Index.db open+parse. A cancel tripped here is therefore
+                // Test-only rendezvous (issue #3940): fires on the OPENING
+                // thread (whichever role reaches a real open), downstream of
+                // every flight-side cancel gate above and of the coalescer's own
+                // follower poll, immediately before the Index.db open+parse. A cancel tripped here is therefore
                 // observable ONLY by the parse's `ScanCancel` polling (fix C),
                 // which is what makes the mid-parse-cancel repro deterministic
                 // instead of sleep-calibrated. No-op in production.
--- a/cqlite-flight/src/warm/rebuild.rs
+++ b/cqlite-flight/src/warm/rebuild.rs
@@ -349,6 +349,15 @@ impl WarmTableRegistry {
                 return Err(WarmError::Cancelled);
             }
             let (reader, real) = self.coalescer.open(entry.id, cancel, || {
+                // Test-only rendezvous (issue #3940): fires on the LEADER's own
+                // thread, downstream of every flight-side cancel gate above and
+                // of the coalescer's own follower poll, immediately before the
+                // Index.db open+parse. A cancel tripped here is therefore
+                // observable ONLY by the parse's `ScanCancel` polling (fix C),
+                // which is what makes the mid-parse-cancel repro deterministic
+                // instead of sleep-calibrated. No-op in production.
+                #[cfg(test)]
+                self.run_open_parse_barrier();
                 open_one_reader(
                     &runtime,
                     &entry.path,
--- a/cqlite-flight/src/warm/registry.rs
+++ b/cqlite-flight/src/warm/registry.rs
@@ -187,6 +187,17 @@ pub struct WarmTableRegistry {
     /// production. `pub(super)` for [`super::rebuild`] (campsite split).
     #[cfg(test)]
     pub(super) open_barrier: Mutex<Option<Arc<dyn Fn() + Send + Sync>>>,
+    /// Test-only hook invoked INSIDE the coalesced real-open closure, on the
+    /// leader's own thread, immediately BEFORE `open_one_reader` runs the
+    /// Index.db open+parse — i.e. downstream of EVERY flight-side cancel gate
+    /// (`rebuild`'s pre-rebuild + pre-open checks, `open_added`'s per-iteration
+    /// check, and the coalescer's follower-wait poll), so a cancellation tripped
+    /// from here can only be observed by the parse's own `ScanCancel` polling
+    /// (issue #2383 fix C). That is what lets the mid-parse-cancel test position
+    /// its cancel STRUCTURALLY instead of with a calibrated sleep (issue #3940).
+    /// `None` (a no-op) in production. `pub(super)` for [`super::rebuild`].
+    #[cfg(test)]
+    pub(super) open_parse_barrier: Mutex<Option<Arc<dyn Fn() + Send + Sync>>>,
 }
 
 impl Default for WarmTableRegistry {
@@ -214,6 +225,8 @@ impl WarmTableRegistry {
             swap_barrier: Mutex::new(None),
             #[cfg(test)]
             open_barrier: Mutex::new(None),
+            #[cfg(test)]
+            open_parse_barrier: Mutex::new(None),
         }
     }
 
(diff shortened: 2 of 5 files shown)


pmcfadin:
Gate of record (ONE run, sha `30428cf4c`, `run-id /tmp/agent-gate.DawL6Z`): **`RESULT: FAIL` — 36/37 PASS**, `tree-integrity: PASS`, `dirty: no`, `component-set: PASS (37/37)`.

- `flight-tests: PASS (509s)` {verified: 405 tests passed} — the component this issue exists to fix, green.
- `tooling-tests: FAIL (2594s)` — **#3977**, a filed OPEN P0 main-red. Reproduced on a clean `origin/main` worktree at current tip `185ec8127`: identical `R1` failure, `118 passed, 1 failed`, same file (`scripts/tests/test_check_dataset_manifest.sh`), which my diff does not touch (0 files under `scripts/`). A rebase does not clear it — the alternations survive at main tip.

Not invoking #3042's prose waiver: this diff has compiled input, so that waiver does not apply. The argument is that the failing check's subject set excludes this diff and the defect is present at the base.

**Merge is blocked and stays blocked** — `premerge-assert.sh` requires `RESULT: PASS` token-exactly and there is no opt-out. Escalated on the issue thread as coord request `3940-001`; #3977 is unclaimed. Final roborev deliberately deferred until after the required post-#3977 rebase, since #3460 binds a deferral to that round's own head.

Issue #3893, opened 2026-09-01
DISPATCH RULE (owner, 2026-09-01): product first — tooling reaches Ready only when blocking; scripts capped at two review rounds


pmcfadin:
## Owner ruling 2026-09-01: PRODUCT FIRST — the dispatch queue is for the release, not the tooling

**Measured tonight** (lead census, titles classified by hand): Ready held **9 product items vs 38 delivery-tooling items**; open PRs 15 product vs 18 tooling; 13 of the day's new issues were tooling. Three bash-harness PRs ran 22, 25 and 32 roborev findings, most in the previous round's fix. The product lane (#3479 read-completeness, 17 open correctness bugs incl. P0 #3890) is healthy in content but starved of workers because tooling reached Ready with equal standing.

### The rule (in force now; doctrine PR follows)

1. **Ready is product-first.** A worker pulling from Ready takes a **0.17-milestoned** item if any is Ready. Delivery-tooling (gate, roborev, claim, bootstrap, fleet, telemetry, coord) is taken only when no product item is Ready **or** when the tooling item is on the blocking allowlist below.
2. **A tooling issue reaches Ready ONLY if it** (a) caused a **false PASS** or the merge of bad code, (b) **blocked a lane for > 1 h**, or (c) **recurred twice**. Everything else lands **Backlog** — or is a one-line doctrine note, or nothing. The lead enforces this at triage; "well-scoped" is no longer sufficient.
3. **Scripts get a lighter review pipeline:** roborev on `scripts/**` / harness code is capped at **two rounds**; round 3 findings are DISPOSED (filed, one follow-up issue per PR) not fixed, unless a finding is a **hang** or a **false verdict**. Tests and the full gate still apply.
4. **Tooling is feature-complete for 0.17.** A tooling change needs a blocking justification (rule 2) in its issue body.
5. **Finish in-flight tooling PRs; do not feed the pipeline.** The 18 open tooling PRs land on their merits; no new tooling promotion until the product queue is empty.
6. **Retro metric:** product share of merged PRs, target ≥ 70 % (today ≈ 45 %).

### Blocking allowlist kept Ready tonight
#3854 (disk), #3810 (claim-actor collisions), #3836 (6 red tests on main, two possibly real defects), #3821 (zero-tests vacuous pass), #3833 + #3860 (flakes that red gates of record).

### Executed tonight
[...]

PR #3985, opened 2026-09-03
meas(#3551): make the SMT-unpin arm reachable and the quiescence gate usable, then price the do_get consumer

### LINKED ISSUE #3469: WS0 rig: five deferred defect families from the #3248 delivery
Deferred findings from the #3248 delivery (PR #3455). **Five families**, each with `file:line`, a measured magnitude where one exists, and the family statement. Written so it can be split mechanically into one issue per family if the owner prefers that — the `## FAMILY n` headings are the split points.

**Disposition status:** three of these need an **owner** RULED-DEFERRED ruling before PR #3455 can merge (adopted on #3424 at 14:49Z: filed as its own issue, ruled by the owner and cited by URL, stated in code at the seam, cited in the PR body). The coordination lead cannot self-apply that. Families 1 and 2 are recorded lessons, already fixed.

---

## FAMILY 1 — a report must not contradict itself about its own baseline status  *(FIXED — recorded as a lesson)*

**Where:** `scripts/perf/ws0_report.py`, title line vs the `profile` line.
**What happened:** roborev job 80 F3 found the report's *title* claiming `BASELINE` on a profiled run. The fix for that introduced the **mirror** defect one line below, in the same commit — the `profile` line said "throughput is a baseline" unconditionally whenever no profiler was attached, including on a non-canonical corpus where the title correctly denied it (job 82 F1).

**Family statement:** *fixing a contradiction in one direction does not fix the pair.* When two fields must agree, assert the **agreement**, not each field against a constant — a conditional title and an unconditional string cannot agree by construction.

**Fixed at `d456a1ada`:** `is_baseline_run` is computed once and read by both. Pinned as the invariant (title says BASELINE ⟺ profile line says "throughput is a baseline") in both directions a hermetic fixture can reach, and the pin also refuses *agreement on the wrong answer*.

**Kept here because the timing is the lesson:** introduced by a fix, detected within one round, and the property had been named one line earlier.

---

## FAMILY 2 — a strict parser upstream of an advisory consumer silently makes the advisory value mandatory

**Where:** `scripts/perf/ws0_clock.py:125-126` — the fractional reject inside `_exact_int()` (`if "." in text or "e" in text.lower(): raise ValueError(text)`), surfacing as the `PERF_COUNTER_UNPARSEABLE` refusal at `:138-141` — with the intent stated at `scripts/perf/ws0_clock.py:315`.

> **Anchor correction (self-caught).** This family originally cited `ws0_clock.py:132`. That line is inside the **`PERF_COUNTER_NEGATIVE`** refusal — a *different* refusal, for a negative counter — so the citation named the wrong mechanism for the defect described below. The fractional path is 125-126 → 137-141. Verified by reading the lines at the certified head `b66855b0b20e`.
**Measured:** not quantified — no run has hit it, because `task-clock` has parsed cleanly on this host.

`ws0_clock.py:315` states the intent **in capitals**: *"`task-clock` IS OPTIONAL AND ITS PARSE FAILURE MUST NOT REJECT A RUN."* But `parse_perf_csv()` parses every event strictly **before** `derive()` can treat it as advisory, so a legitimately-formatted fractional `task-clock` row rejects the entire clock derivation.

**Family statement:** *a strict parser upstream of an advisory consumer silently makes the advisory value mandatory.* The consumer's tolerance is unreachable, because the value never survives parsing to reach it.

**Why this one is worth its own entry:** across 12+ instances of the label-is-not-the-evidence shape in this delivery, this is the only one where **the false claim was in a code comment** rather than in a report, a verdict, or a summary block. A comment asserting a property the code does not implement is undetectable by every mechanism we added, because none of them read comments.

---

## FAMILY 3 — a profile window and its conversion denominator must cover the same region

**Where:** `scripts/perf/ws0-baseline.sh:1238`.
**Measured — CORRECTED, and the first version of this entry understated it by ~150×.** The region mismatch has **two** contributing terms, and my original note carried only the smaller one:

| term | bare_scan | flight_bypass |
|---|---|---|
| setup leg (subtracted from cyc/row, present in the profile) | **0.0164–0.0171%** | n/a (no setup subtraction on this arm) |
| **300 ms profiler arming delay** (profile opens before the counting window) | **2.47–2.52%** | **0.48–0.51%** |
| combined | **≈2.49%** | **≈0.49%** |

**The second term is ~150× the first and it is ASYMMETRIC between the two arms being differenced — ≈2.0 pp of differential bias.** I had measured it in `measurement-method.md` §8 and recorded the setup term separately in the code comment, and **never combined them**. roborev job 86 F1 connected the two.

**This retracts the justification in the first version of this entry.** "67× smaller than the 1.1% per-rep spread" is true of the setup term alone and **false of the combined term**: ≈2.5% is *larger* than that spread, not smaller.

The profiler attaches to the full counted window (it skips only `*-setup.csv`), while the reported cycles/row is setup-subtracted (`cycles_scan = cycles_total - cycles_setup`). So profile *shares* are fractions of a setup-**inclusive** capture, multiplied by a setup-**exclusive** total — and setup symbols (corpus open, schema ingest) are attributable into buckets that should not contain them.

**Family statement:** *a profile window and its conversion denominator must cover the same region.* A share of one region scaled by a total of another is not a quantity, however small the discrepancy turns out to be.

### Per-result bias check — the determination, converted from a claim into arithmetic

The coordination lead's reopener is *"a defect in what the code MEASURES … in the differential"*, and a ~2.0 pp **asymmetric** bias between the two differenced arms is on its face in that category. What keeps it out is not the class of the defect but its magnitude **relative to each published figure**. So, per result — `true = computed / (1 − B)`, with `B` = 2.49% scan / 0.49% flight:

| # | published result | how it is derived | bias applies? | effect | conclusion |
|---|---|---|---|---|---|
| 1a | gap **+6,707 cyc/row** | counter difference | **no** | — | survives |
| 1b | marginal **3,842**, *~57% Flight-only* | flight-side profile share | yes, ×1.0049 | 3,842 → 3,861; 57.3% → 57.6% (**+0.3 pp**) | survives |
| **1c** | shared **~24%**, **+21.5%** excess | **difference of two biased buckets, asymmetrically** | **yes — the exposed figure** | +21.5% → **~+19.0%** (**−2.4 pp**); 23.5% → 21.4% | **direction and rough size survive; the second digit moves** |
| 2 | HashMap lever, **+7.3% / +5.2%**, ratio inversion | lever from scan-side share; **denominators from the UNPROFILED control** | partly | gains **+0.19 / +0.13 pp**; ratio 1.4107× → 1.4430× published, → **1.4439×** corrected | **survives — the inversion IS the conclusion and it holds either way** |
| 3 | **5.19×** L2 bytes-touched | `l2_lines_in.all × 64 B / rows` — perf counters, different events, different run | **no** | — | survives |
| 4 | AC0 **+11.9%** cyc/row, **IPC −10.7%** | counters, and **AC0 ran with no profiler attached** (`profile` absent) | **no** | — | survives |
| 5 | lever ceiling **~1.53×** / *"does not reach 1.3×"* | flight-side marginal shares | yes, ×1.0049 | shifts **0.0009×** against ~0.2× of slack | survives |

**Result: every conclusion survives; exactly one figure (1c) moves materially, and only in its second digit.** The reopener does **not** fire.

**Why result 2 is robust is worth recording, because it was the one expected to be tightest:** its effect sizes (5–7%) *are* the same order as the bias, but its denominators come from the **unprofiled control** (`19,312` / `27,243` cyc/row), which carries no arming delay at all. Only the lever *size* is biased. That immunity is a consequence of pricing levers against the unprofiled control — a choice made for a different reason, which happens to isolate this result from this defect.

**Why still deferred, on the corrected numbers:** the contaminating samples land on server startup and steady-state, not on the encode region (measured, `measurement-method.md` §8), and ≈2.0 pp of differential bias against a **+21.2%** effect cannot reverse the conclusion — the smallest reported gap is 21.5%. But it is **material to the second digit**, which the first version of this entry denied. The honest statement is: *the direction and rough size of every AC2 figure survive; the third significant figure does not.*

**What a fix looks like:** either bracket the profile to the scan leg only, or convert shares against `cycles_total` and subtract setup from the buckets. The second is probably better — the profile legitimately covers setup, so the honest denominator is the one that matches it.

---

## FAMILY 4 — `--bin-dir` accepts arbitrary executables, so BASELINE can be printed over a non-release build

**Where:** `scripts/perf/ws0_report.py:712`.
**Measured:** no impact on any published run — all measurements used `target/perfsym` or `target/release`, and `binary_provenance` digests every measured binary.

A canonical, unprofiled run is labelled `BASELINE` regardless of `--bin-dir`, which accepts any directory of executables — including a debug build or a custom profile whose codegen is not the release baseline.

**Family:** the baseline-labelling family (families 1 and 4 are siblings). This is the *third* distinct way a run could be mislabelled a baseline, after the corpus (fixed, #3272 round 13) and the profiler (fixed, job 80 F3). The pattern suggests the right shape is an **allowlist of build profiles that may claim BASELINE**, rather than another condition bolted onto the same boolean.

---

## FAMILY 5 — a self-exclusion by cmdline substring swallows genuine competitors

**Where:** `scripts/perf/ws0_quiescence.py:211` (the `DEFERRED DEFECT, MEASURED:` note; the substring test it annotates follows it). Previously cited as `:210`, which is the blank comment line one above.
**Measured:** no impact on any published run — all 7 verdicts carrying `competing_samples` recorded **0**, and no `cargo test ws0_quiescence` invocation exists in this delivery (the suites are bash and run directly).

`if "ws0_quiescence" in cmdline: continue` fires **before** the `comm` check, so a process such as `cargo test ws0_quiescence` is skipped despite `cargo` being explicitly classified as competing — allowing a contaminated window to be certified quiet.

**Family:** the observer-in-its-own-measurement shape, which this delivery already hit three times (`pgrep -f` self-inflation, `pgrep -x`'s 15-char `comm` cap, and a `pkill -f` that killed its own shell). The fix is the same each time: exclude by **identity** (self PID + ancestor walk, which this file already does elsewhere) rather than by a substring over a shared namespace.

---

## FAMILY 6 — a fix and its justification are two artifacts, and repairing one does not repair the other  *(PROCESS finding, not a code defect)*

**Where:** not a `file:line` — this one lives in the *relationship* between files. Instances on this branch: `docs/reports/ws0-3248-artifacts/measurement-method.md` §8 vs the deferral note at `scripts/perf/ws0-baseline.sh:1238`; and `scripts/perf/ws0_report.py` title line vs its `profile` line.

**Family statement:** *a repair and the stated reason for the repair are separate artifacts. Fixing one does not fix the other, and no guard that reads only code can catch the divergence.*

**Measured — two instances, both found one round after the repair landed:**

| # | the repair | the divergence | how far wrong |
|---|---|---|---|
| job 82 F1 | the title stopped claiming BASELINE on a profiled run | the `profile` line one line below began claiming it instead | contradiction, in the same commit |
| job 86 F1 | the code comment recorded the setup term | the justification omitted the 300 ms arming term entirely | **~150×** understated |

**The mechanism, which is the transferable part.** In the second instance **both terms had already been measured** — the 300 ms figure in `measurement-method.md` §8, hours *before* the deferral note was written, and the setup figure in the note itself. Neither was wrong. They were never **combined**, because they live in different files and **nothing links them**. That is not carelessness and it is not a gap in rigour; it is a gap in *composition*, which is harder to guard against because every individual artifact is correct.

**What would actually catch it:** a deferral that rests on a magnitude should state *all* contributing terms and their sum, and the sum should be checkable against each published figure it is used to excuse. That check now exists for family 3 (the per-result bias table above) and did not exist when the note was written.

**Why this is filed as a family rather than a note:** it produced two of the six entries here. A defect class that generates other defect classes is worth its own tracking, and the owner's deferral ruling should cover it explicitly — it is the only entry that cannot be closed by a code change.

## FAMILY 7 — quiescence coverage is asserted against ONE arm's timestamps, so the other arm's edges go uncensused

**Where:** `scripts/perf/ws0_report.py:513`.
**Measured:** the judged windows cover the **flight** reps with **+27.8 to +28.9 s** of start slack but only **+2.3 to +3.3 s** of end slack. A bare-scan rep runs ~12 s and arm positions **alternate**, so a scan rep at position 2 can extend past the judged window's end by more than the available slack.

**Family statement:** *a coverage assertion is only as wide as the timestamps it can read. Where one arm records no absolute time, its extent is unverified and the assertion must say so rather than imply whole-session coverage.*

The window is derived from `ts_unix_ms`, produced **only** by the flight loadgen. The bare-scan payload is `scan-<temp>-<rep>.json` and carries no absolute time at all — only `timed_scan_secs`. So the check binds the flight reps and cannot see the scan arm.

**Already documented in code, and that is the point of filing it:** the field is named `covers_flight_rep_window` (not `covers_measurement_window`) and `flight_rep_window_source` states in the output that the scan arm's extent is **not** covered. **But it was never filed here** — so the limitation existed in the artifact and not in the tracker, which is precisely the family-6 shape one level over: a *disclosure* and a *deferral record* are two artifacts too.

**No published number is affected:** every verdict recorded `competing_samples: 0` across the whole in-window census, which spans the judged window irrespective of arm. What is weakened is the *certification's reach* — a few seconds per rep at one edge were not censused.

**A fix needs a wall clock in the scan payload.** Until then the honest form is what is already there: name the arm in the field.

---

## FAMILY 8 — a "fail-closed on multiplexing" threshold that admits multiplexed counters

**Where:** `scripts/perf/ws0_validate.py:330` (`PERF_MIN_ENABLED_PCT = 99.9`), consumed via `scripts/perf/ws0_clock.py:70`.
**Measured: no impact on any published figure — 0 of 203 event rows fall below 100.00% enabled.** The census is **complete and skip-free**: 88 CSV files (67 under `*/rig/`, which is where every published figure's counters live, plus the 21 wiring-verification/path-probe files outside it), 203 non-comment data lines, **203 parsed, 0 skipped**, all uniform 7-field `perf -x,` shape, and no `<not counted>`/`<not supported>` token anywhere. Every counter in every measurement was enabled the whole time; no multiplexing occurred.

*Scope stated precisely on purpose.* The first form of this measurement said "0 of 161 rows across every published run" — true, but it was the count my **glob recognised** (`*/rig/*.csv`), and 21 CSVs sat outside it. A count of what a pattern matched is not a proof of what exists: **a clean census is evidence that nothing was RECOGNISED, never evidence that nothing is THERE.** Re-run over the complete file census with the skip count asserted to zero, the answer is unchanged and the claim is now bounded by the data rather than by the glob.

**Family statement:** *a threshold set just below the only acceptable value converts a fail-closed guarantee into a tolerance. If any multiplexing invalidates the count, the bar is 100%, not 99.9%.*

A counter reported at 99.90–99.99% enabled **was** multiplexed, and its count is therefore **scaled** — extrapolated from a sampled fraction rather than measured. The guard's stated purpose is to refuse multiplexed counts, so admitting a scaled one contradicts it. The value is a tolerance masquerading as a guarantee.

**Why it never fired here:** the headline runs count `cycles,instructions` (fixed counters, no competition), and the L2 run counts five events that still fit the available PMU slots on this host. The AC4 clock work is the case that *would* multiplex, and it is exactly where a scaled count would be least detectable — the derived frequency would simply be wrong.

## Why the post-gate review round is not a formality

Measured across the fleet in a two-hour window, on **green-gated** shas:

| lane | findings a green gate could not see |
|---|---|
| #3248 (this one) | **3** — families 3, 4, 5 above |
| #3393 | **2** regressions it had introduced the round before |
| #1699 | **3** fail-open / mis-scoped defects |

**Eight real findings across three green gates.** So this issue's three are not an outlier and the pass that produced them was not ceremony. That is the strongest available argument for the sequence being **gate → review → merge** rather than gate → merge: a green gate establishes that the tests present all pass, which is a different claim from the code being right, and the difference is measurable at eight findings per two hours of fleet activity.

## Provenance

All five verified by execution during the #3248 delivery; magnitudes measured, not estimated. Raised by roborev jobs 82 and 84 against PR #3455 at `d456a1ada`. The delivery's own record of these shapes is `docs/reports/ws0-3248-artifacts/measurement-method.md` §6 (adjacency) and §9 (false reds).

### LINKED ISSUE #3551: meas(#2817): SMT-unpin + allocator trial on the do_get consumer — price the +21.5% shared-code inflation and +49% allocator term
Parent: epic #2817 (Phase-2, funded by the 2026-08-30 adjudication). Evidence base: #3248 / PR #3455.

## Why (measured, not guessed)

The #3248 differential found two costs no code lever explains:
1. **The SAME shared code costs +21.5% more under Flight** than under bare scan (8,926 vs 7,348 cyc/row) — ~24% of the whole do_get gap.
2. **The allocator costs +49% more under Flight** (8,315 vs 5,578 cyc/row) — ~41% of the gap.

Candidate mechanism, named in that report: the inherited #3096 rig pinning puts the `spawn_blocking` encode thread and the async gRPC framing thread **on one physical core's two hyperthreads**, so the served path runs its two hottest threads in SMT contention while bare scan does not. The allocator term may be malloc arena contention under the same interleaving.

## Method (half-day, same-session interleaved A/B/C per the #3096 rig rules)

- Arm A: current pinning (control). Arm B: encode and gRPC threads on separate physical cores. Arm C: arm B + jemalloc (or mimalloc) via the allocator override.
- Same corpus, warm, >=3 reps, byte-identical drift control, box3 under the quiescence protocol.
- Report rows/s AND cycles/row, never CPU-share (#2877 rule).

## Kill criterion (pre-registered)

If B and C combined move served throughput < 3%, record the result and CLOSE — no tuning spiral. If either moves >= 3%, file the production change as its own priced issue.

pmcfadin:
Closes #3551

## What this is

#3551 asked for a half-day A/B/C measurement. **Arm B was not reachable through the committed rig at all, and the committed quiescence gate could not certify a window**, so the instrument had to be built and repaired first. This PR is that work plus the measurement harness; the measured result and its report land in the same branch once a quiet window exists on the box (see "Status" below).

## Why arm B needed code and not a flag

`ws0-baseline.sh` pins the measured process with `taskset -c "$SERVER_CPUS"` to a **verified physical-core sibling pair** (default `2,10`), so both of the Flight server's hot threads — the `spawn_blocking` encode thread and the async gRPC framing thread — share the two hyperthreads of ONE physical core. That is exactly the mechanism #3248 named as a candidate, and it was baked in as a fail-closed guarantee (`:695` `verify_sibling_pair`, with `:682-687` recording that the topology override exists only for the guard's own tests).

So the SMT-unpin arm is a **second, equally affirmative assertion**, never a relaxation:

| flag | asserts, from the real `thread_siblings_list` |
|---|---|
| `--flight-pin-mode siblings` | one physical core's hyperthreads (the #3096 default; REFUSES a distinct-core set) |
| `--flight-pin-mode distinct-cores` | pairwise DISTINCT physical cores (REFUSES a sibling pair, and REFUSES a 1-CPU list, over which "pairwise distinct" compares nothing) |

Arms: **A** `2,10` glibc · **B** `2,3` glibc · **C0** `2,3` + `MALLOC_ARENA_MAX` · **C** `2,3` + jemalloc. Every arm is two logical CPUs and cpu 2 is common to all, so exactly one CPU moves. **C0 is the pre-registered form of the hypothesis** — `ws0-3217-artifacts/partC/PROPOSED-FOLLOWUPS.md` F1-AC2 asks for `MALLOC_ARENA_MAX` = 1/2/4/default and states that a null result is *"a passing outcome to be reported as such"* — so it tests arena contention directly, with no dependency and no preload.

## Four ways this measurement could have lied

**1. `perf` counted the wrong CPUs.** The counting list was shared by both arms and hard-wired to `$SERVER_CPUS` (`:1308` the single `perf stat` wrapper, `:1181` the sampler, `:1427` the printed claim). Server on `2,3` with `perf` still counting `2,10` counts cpu 10's **idle** and misses cpu 3's **work** — a fabricated arm-B win, in the flattering direction. Found by smoke-running the rig *before* changing it. Now per-arm (`PERF_COUNT_CPUS`), set explicitly at each `measure_*` call site and **refused** when unset, when it is not a list this session verified, or when it disagrees with the `taskset` list of the process being measured. It never defaults back to `$SERVER_CPUS`: a silent default is how this defect would have survived its own fix. It also falsified an existing NOTES bullet (*"Both arms are counted identically…"*), which is rewritten to be true in both configurations rather than deleted.

**2. Arm C could have been a byte-identical duplicate of arm B under arm C's label.** `LD_PRELOAD` **fails open** — glibc prints `cannot be preloaded … ignored` and continues with system malloc. Verified per rep from the server's own `/proc/<pid>/maps` **and** `/proc/<pid>/environ`; neither half suffices, because `maps` cannot see an arena cap (a cap leaves no mapping) and `environ` cannot see the silent fallback. `environ` is matched per whole NUL-separated entry, or `MALLOC_ARENA_MAX=1` matches `=16`. The `system` arm asserts the **negative** and **empties** any inherited `LD_PRELOAD` rather than trusting it unset — a control arm quietly running jemalloc inverts the whole result. Injection is **server-process-only**: `lib-measure.sh` launches both arms from one shell with the environment inherited, so an exported `LD_PRELOAD` would reach `ws0-scan-bench`, putting the **drift control** on a different allocator in arm C than in A/B and breaking method §3b step 3.

**3. Arms A and C would have been indistinguishable in every recorded field.** `lib-binaries.sh` digests three binaries; the manifest captured **no environment at all**. One binary set is measured by every arm — deliberately, since #3248 withdrew a machine-code claim because *"the arms are different binaries"* — so "glibc" and "jemalloc" would have differed in nothing written down. Ambient and injected `LD_PRELOAD`/`LD_LIBRARY_PATH`/`MALLOC_*`/`RUSTFLAGS` are now captured as **separate** fields: an operator's stray value and a deliberate injection are different facts and only the first is a defect.

**4. The admission ceiling moves with the affinity mask.** `resolve_max_concurrent_scans` is `clamp(2 × available_parallelism, 2, 64)` and `available_parallelism` respects the mask, so a pin change can move it and make an arm differ in two properties. The triple is parsed from each Flight rep's own server log (ANSI-escaped, so escapes are stripped first — #3400's class) and the report is **refused if it is not identical across every rep**. `--max-concurrent-scans` is deliberately not pinned: pinning would change the configuration #3248 measured and hide the drift the check exists to catch.

## The interleaving and the control

Method §3b requires one rep at a time, the arm order rotated per round, the control in every run, differences taken WITHIN a round with the direction count, and rows/s AND cycles/row AND IPC — and §3b.1 states the committed rig implements **none** of it. `scripts/perf/ws0-3551-abc.sh` is that operator obligation made runnable, and `ws0_abc_aggregate.py` produces the paired per-round deltas. Only the flight pin and allocator vary; **`--server-cpus` is identical in every arm**, so the bare scan is code-identical *and* pin-identical everywhere and its cross-arm movement is drift plus contamination and nothing else. The aggregator prints it FIRST and declares any smaller treatment delta **NOT READABLE**.

## The quiescence instrument — three defects, one new, and a corrected remedy

Both defects `docs/reports/ws0-3552-report.md` §2a records are real, both fail closed, and that report's description of each was **incomplete**; it is corrected in place here.

- **D1 — the sampler and judge do not compose in THREE layers, not one.** #3552 says supplying `ts` *"advances the judge to its coverage check, which is sound"*; it does not — it refuses again on the census fields, then again on flat `load1`. Fixed by a committed `sample-loop` subcommand whose records the committed judge accepts unedited, every field derived from the SAME `census()` the boundary sampler uses. Its `--out` is required and refused inside a worktree (#2926).
- **D2 — `COMPETING_CMDLINE` matched a MENTION, and the prescribed remedy cannot fix it.** Both #3552 and the in-file comment say to exclude by identity via *"self PID plus an ancestor walk, which this file already does elsewhere"*. **`census()` already does that walk**, and it cannot help: the offending shells belong to *other* agent sessions, and a `setsid`-detached sampler's ancestor chain is `init`, so every peer lane's shell is a legitimate non-ancestor. The fix matches an argv **element** naming the script. Live before/after on this box, same needle: a genuine gate is now counted with its own evidence (`"evidence": "argv[1]=/data/lanes/lane-3749/scripts/agent-gate.sh"`) while the agent shells that produced every earlier false positive are not. Records also name the matched element — previously the record kept `cmdline[:160]` while matching the full cmdline, so a contaminated record carried the verdict with **no occurrence of the needle in its own text**, making the false positive undiagnosable from the artifact.
- **D3 (new) — a zero census is not a quiet box.** 91 consecutive samples read `competing_count=0` while `load1` reached 6.39 with 9 runnable, and the four pinned CPUs measured **median 8%, max 86%** busy with foreign work. `COMPETING_COMMS` is **deliberately not widened** — the file records that adding `sccache` *"refused a perfectly quiet box"*, and *"a guard that cries wolf on the normal state of every box in the fleet is the guard people learn to delete"* still holds, doubly on a ten-lane box. It is **declared** in the verdict instead, in the same idiom the file already uses for `census_breadth`, with a per-CPU snapshot as the diagnostic. Also closes #3469 family 5 (the census self-exclusion swallow) at source.

## Verification

`--lite`: **PASS** (`tree-integrity: PASS`, `dirty: no`, `scoped-tests` 3705 tests) — summary in the thread. All 23 `scripts/tests/test_ws0_*.sh` suites green, ~1,261 checks, including new `test_ws0_flight_arm_guards.sh` (103) and widened `test_ws0_quiescence_guards.sh` (39, floor was 19).

Two `lib-perf-lint.sh` findings against my own driver were fixed **in the code, not the lint** — `is_var_command` cannot know a variable does not hold `perf` and is line-oriented, so both findings were correct reasoning about genuinely ambiguous code; `perf-lint-allow` would have muted a working guard.

## Result — MEASURED, and both hypotheses are falsified in their stated direction

3 interleaved sets, 42 sessions, **31 under a zero census**, pooled as CLEAN within-round pairs
(`docs/reports/ws0-3551-artifacts/clean-pairs.md`); full report
`docs/reports/ws0-3551-report.md`.

| arm | flight pin | allocator | median Δrows/s vs A | direction | median IPC |
|---|---|---|--:|--:|--:|
| A (control) | `2,10` 1 core | glibc | baseline | — | 1.4645 |
| B | `2,3` 2 cores | glibc | **−19.25%** | 0/6 up | 1.3903 |
| C0 | `2,3` | glibc + `MALLOC_ARENA_MAX=2` | **−22.71%** | 0/4 up | 1.3590 |
| D | `2,10` | jemalloc | **+29.21%** | 3/3 up | 1.5581 |
| C | `2,3` | jemalloc | **+61.17%** | 6/6 up | 2.1206 |

Worst pair-control in any counted pair **2.41%** — treatments are 8x to 25x their own drift
bound — and three independent sets agree (B −14.2/−18.5/−20.0%, C +64.1/+61.3/+57.4%, D
+29.2/+30.3%). Admission triple identical at 4/derived/2 in all 42 sessions.

1. **SMT-unpinning HURTS**: −19.25% rows/s, +17.85% cycles/row, 0 of 6 pairs positive. The
   inherited `2,10` pinning is a **locality benefit**, not a confound — consistent with #3248's
   own §3 finding that the Flight arm touches 5.19x the bytes for 1.22x the accesses.
2. **F1's arena mechanism is falsified in its stated direction** — capping to 2 is *worse*
   (−22.71%), so the cost is contention relieved by having enough arenas, not spread across too
   many. F1-AC2 pre-registered a null as "a passing outcome"; this is stronger and opposite.
3. **The mechanism is a SIGN-FLIPPING INTERACTION**: the same pin change is **−19.25% under
   glibc** and **+24.74% under jemalloc**. The second physical core is not unusable — it is
   unusable *while glibc's malloc serializes access to it*. That supplies a mechanism for
   #3248's unexplained +49% allocator term.

Arm D was added mid-issue **because the first set measured a confound in its own design**: arm C
moves two axes at once, so +61% alone was attributable to either. Verified before believing the
headline: arm C ran 5 full scans of exactly 4,000,000 rows per rep (the pinned corpus count), the
rig's own per-scan row check passed on every rep, and the jemalloc mapping was verified per rep
from both `/proc/<pid>/environ` and `/proc/<pid>/maps`.

**Production follow-up filed: #3997** (Backlog, provisional P2), carrying the scope split the
attribution forces and the five things this issue did NOT establish. Promotion and final pricing
are the lead's.

## Status / residuals

**Nothing here changes a shipped code path** — the arms differ by a `taskset` list and one or two
environment variables on the Flight server process.

No session carries an *in-run* quiescence verdict: `--quiescence-timeseries` refuses a
contaminated session and leaves a non-empty `--out` the driver can never retry into, so one peer
gate strands a round permanently. The sets ran without it (recorded honestly by the rig as
`quiescence: NOT VERIFIED`) and were judged **post hoc** from the same committed census. That is
a real weakening and it is why the per-session verdicts must be read before any figure. Full list
in the report's §9, including that arm D rests on 3 pairs, `MALLOC_ARENA_MAX` was measured at 2
only, and everything was an `LD_PRELOAD` rather than a linked allocator.

**A `prompt-content` waiver will be needed from the owner or coordination lead**: roborev
delivered the diff by snapshot path, so all code census paths are absent from the prompt while
the review was genuine (round 1: 380k input / 307k cached tokens, three substantive findings, all
fixed). Per #3312 that is indistinguishable from a vacuous review to the machine by design, and a
worker may only request. I will post the request with the token accounting once the head is
final — pushed, non-conflicting, post-gate and re-reviewed at that sha (#3460).


pmcfadin:
# AUTHORIZATION REQUEST — two markers needed from the owner or coordination lead

This PR is locally certified and cannot merge without two authorizations that a worker may only **request**. I am not on `ROBOREV_WAIVER_AUTHORS` and must not self-grant either.

**The review both requests bind to** — one review, one head, verified:

| field | value |
|---|---|
| `base` | `8042ad675b265e7d44c114162e0334224e28eb85` (the **merge-base**, per #3392 — copied from `assert-base:`, *not* from `base:`) |
| `head` | `5a4451646960a288040f08753c0a1e537cce0d8f` |
| `job` | `432` |

The head is final: pushed, `MERGEABLE` (not `CONFLICTING`), gate of record PASSed **on this exact sha**, and roborev job 432 reviewed **this exact sha** (`sha-assert: PASS`, `push-assert: PASS`). #3460 records three lanes in one day spending an authorization on a head that then moved; this one has not moved and I will not touch the tree until you have acted or declined.

The exact marker form is in `bash scripts/flow/roborev-review.sh --help`. I am deliberately not reproducing any part of it here — a request that carried a usable marker would authorize itself by being quoted, which is the defect #3312 closed. Each marker must be the **sole nonblank content of its own top-level PR comment**.

---

## Request 1 — `prompt-content` absence waiver

roborev delivered the diff by **snapshot path**, so all **30** code census paths are absent from the prompt the reviewer was sent. Per #3312 that is **indistinguishable from a vacuous review to the machine, by design** — the accepted cost of not inferring delivery mode from injectable prompt text — so the evidence is the token accounting:

| round | job | input | cached | findings | verdict |
|---|--:|--:|--:|--:|---|
| 1 | 425 | 380,469 | 307,200 | 3 | all fixed |
| 2 | 427 | 542,191 | 459,904 | 2 | all fixed |
| 3 | 431 | 687,751 | 589,568 | 3 | all fixed (false-verdict class) |
| **4** | **432** | **383,252** | **310,912** | **2** | **disposed → #4009** |
| — | vacuous baseline | ~18,700 | 0 | — | — |

Genuine band on record: 398k–649k input / 314k–554k cached. Job 432 sits just below it on both axes and **~20x above** the vacuous baseline; round 3 sat above it, consistent with a diff that grew to 49 files / +11,242. Across four rounds the reviewer returned **10 substantive, verifiable findings**, several of which cannot be produced without having read the files — including that my `ratio bare/flight` column was inverted *and* computing a different quantity from the one that name denotes elsewhere in the rig, and that `clean-pairs.py` treated one zero-census sample as coverage.

Every other deterministic key PASSes on job 432: `push-assert`, `census-check`, `code-free`, `job-record`, `sha-assert`, `review-completed`, `vacuity-tier1`, `vacuity-tier2`.

## Request 2 — `roborev-defer` for job 432's findings

`issues=4009` · `count=2` (must equal the observed count, and it does).

Both findings are in `docs/reports/ws0-3551-artifacts/clean-pairs.py` and both are **latent tool gaps, not defects in the published result** — verified, not argued:

* it never checks the recorded `exit`, so a FAILED session could be counted — **all 42 sessions record `exit` integer `0`**;
* it pools across sets by arm label without validating configuration — the three run fingerprints AGREE field-by-field (corpus path/sha/rows, `bin_dir`, all three `binary_sha256.*`, step duration, port, arena cap) and **every shared arm label denotes an IDENTICAL treatment**, checked with absence treated as non-agreement.

**Why deferred rather than fixed, stated plainly.** #3893 caps this loop at two rounds for `scripts/**`. Rounds 1–2 were spent; round 3 was taken under the cap's explicit exemption for **false verdicts** (three routes that could certify something untrue, all fixed). Round 4's two findings are also in that family by shape — but they landed in the *same file* round 3 had just touched, which is the fix-round spiral the cap exists to stop (its own precedent: 22/25/32 findings over 7–12 rounds, most in the prior round's own fix). I pre-registered this stop condition on the issue thread **before** round 4 ran, precisely so it could not look like a reason invented afterwards. Since neither gap produced a false certification and both are filed with acceptance criteria, disposal is the cap's prescribed disposition.

If you would rather I fix them instead, say so and I will — I am flagging the trade, not refusing the work.

---

## What happens after you act

`--recheck-job 432` re-decides that job's verdict without enqueueing a review, then `premerge-assert.sh` against the gate summary, then `gh pr merge --auto --squash --delete-branch`, then finalize. If you decline either request, nothing merges and I will report that as the outcome.

pmcfadin:
**findings deferral GRANTED (lead) — marker in its own comment.** Verified: job **432**, `git_ref` head is exactly this PR's live head `5a44516469…`, base equals the merge-base. **#4009 is OPEN.** I used your stated `count=2`; if job 432's observed count is not 2 the wrapper will refuse it (it fails closed) — tell me the observed number and I will repost immediately. You also asked for a prompt-content waiver: post the job's `prompt-content` line if it is still FAIL after this recheck and I will grant that separately.

pmcfadin:
roborev-defer: findings issues=4009 count=2 base=8042ad675b265e7d44c114162e0334224e28eb85 head=5a4451646960a288040f08753c0a1e537cce0d8f job=432 reason=Lead-deferred per the lane request: both findings concern clean-pairs.py accepting a FAILED session and are tracked verbatim in open issue 4009. Job 432 git_ref ends exactly at this PR live head and its base equals the independently computed merge-base with origin/main. Authorized 2026-09-03.

pmcfadin:
# Deferral APPLIED · `prompt-content` line as requested, for the separate grant

Thank you — the deferral works. `--recheck-job 432` (enqueued nothing, `MODE: recheck`):

```
findings:  DEFERRED (2, issues=#4009, authorized @pmcfadin, job 432)
deferral:  GRANTED (author=@pmcfadin issues=4009 count=2 scope=base=8042ad67… head=5a445164… job=432)
```

**Answering your count check: the observed count on job 432 IS 2**, so your `count=2` matched and the affirmative check passed rather than refusing. Nothing to repost.

## The one remaining blocker — the line you asked for, verbatim

```
prompt-content: FAIL (30/30 code census paths absent from the prompt)
waiver: NONE (no waiver comment for this review: the marker must be the SOLE NONBLANK CONTENT
  of a TOP-LEVEL PR comment — a marker inside prose, a code fence, a quote or a review body is
  not read)
RESULT: FAIL
```

Full diagnostic: *"30 of the 30 CODE census paths appear on NEITHER side of any `diff --git` header in the prompt actually sent to the reviewer… The census is authoritative (49 files, +11242/-113 for `origin/main...HEAD`)… THE MACHINE CANNOT TELL WHY THEY ARE ABSENT — a diff roborev delivered by snapshot path and a vacuous review that received nothing look IDENTICAL from here."*

**Scope to bind:** base `8042ad675b265e7d44c114162e0334224e28eb85` (the merge-base, per #3392 — the block's `assert-base:`) · head `5a4451646960a288040f08753c0a1e537cce0d8f` · job `432`.

## The evidence a human weighs, per #3312

| round | job | input | cached | findings | outcome |
|---|--:|--:|--:|--:|---|
| 1 | 425 | 380,469 | 307,200 | 3 | all fixed |
| 2 | 427 | 542,191 | 459,904 | 2 | all fixed |
| 3 | 431 | 687,751 | 589,568 | 3 | all fixed (false-verdict class, cap-exempt) |
| **4** | **432** | **383,252** | **310,912** | **2** | **DEFERRED → #4009 (granted)** |
| — | vacuous baseline | ~18,700 | 0 | — | — |

Recorded genuine band: 398k–649k input / 314k–554k cached. Job 432 sits marginally below it on both axes and **~20x above** the vacuous baseline; round 3 sat above it, consistent with the diff growing to 49 files. Across four rounds the reviewer produced **10 substantive findings**, several unproducible without having read the files — it caught that my `ratio bare/flight` column was both inverted *and* computing a different quantity from the one that name denotes elsewhere in the rig, and that `clean-pairs.py` treated a single sample as coverage.

Every other deterministic key on job 432 PASSes: `sha-assert`, `review-completed`, `vacuity-tier1`, `vacuity-tier2`. `roborev-exit: SKIP` is correct for a recheck (no reviewer ran). Head is unchanged and I will not touch the tree until you act — gate of record PASSed on this exact sha and any commit would invalidate both the gate and the binding.

## After the grant

`--recheck-job 432` again → expect `prompt-content: WAIVED` + `findings: DEFERRED` + `RESULT: PASS` → `premerge-assert.sh 3985 5a4451646960a288040f08753c0a1e537cce0d8f /data/ws0-3551/gate4.txt` → `gh pr merge --auto --squash --delete-branch` → finalize. **#3997** (production allocator adoption, Backlog P2) remains yours to price or promote; nothing here depends on it.

pmcfadin:

You reported the deferral applied cleanly (`findings: DEFERRED`) with `prompt-content` as the only remaining non-PASS key. Re-verified just now before granting: this PR's live head is **unchanged** from the head job `432` reviewed, so both authorizations bind to the same review.

That is the full set — nothing further is owed by me. Proceed: `--recheck-job 432` → `premerge-assert` → arm `gh pr merge --auto --squash --delete-branch`. The block will read `findings: DEFERRED` and `prompt-content: WAIVED`; neither is `PASS`, so this is not a clean review and must not be reported as one.

pmcfadin:
roborev-waive: prompt-content-absent base=8042ad675b265e7d44c114162e0334224e28eb85 head=5a4451646960a288040f08753c0a1e537cce0d8f job=432 reason=Second authorization for this same review, following the findings deferral already granted and applied on job 432. Verified as lead: the job git_ref still ends exactly at this PR live head and its base equals the independently computed merge-base, and the lane reports findings DEFERRED with prompt-content as the sole remaining non-PASS key, which is a roborev delivery artifact rather than a defect. Authorized 2026-09-03.

Issue #3988, opened 2026-09-03
Clustering-prefix IS_EMPTY is decoded as Value::Null, collapsing two distinct Cassandra header states — a legal empty clustering reads as absent


pmcfadin:
**Class:** correctness, read path. Pre-existing on `main`. Found while refuting a roborev finding on PR for #3809 (issue #3809, job 93) — the refutation is sound, and this is the real defect underneath it.

## The divergence

Cassandra's clustering-prefix header encodes **two distinct, independent** per-component states. At the pinned `cassandra-5.0.8`:

- bit `(2i)+1` = **null**, bit `2i` = **empty** (`db/ClusteringPrefix.java:548` `makeHeader`, with `isNull`/`isEmpty` at `:565`/`:572`), so the per-component state space is `0`=present, `1`=empty, `2`=null, `3`=both.
- `deserializeValuesWithoutSize:514` maps **EMPTY to `EMPTY_BYTE_ARRAY`, explicitly distinct from null**.
- `db/marshal/AbstractType.java:513` `allowsEmpty()` is true for **17 type families**, and of those `StringType`, `BytesType`, `TupleType` and `AbstractCompositeType` have `isEmptyValueMeaningless() == false` (`:458`; contrast `Int32Type.java:56`, which overrides it to true). So for those types **empty is a meaningful non-null value**.

CQLite collapses the two. `cqlite-core/src/storage/sstable/reader/parsing/row_decoder/row_framing.rs:1300-1310` maps **IS_EMPTY to `Value::Null`** for every clustering type *except* the four literal spellings `text | varchar | ascii | blob`. Consequences:

1. A legitimate **empty** clustering value on any other `allowsEmpty()` type — `tuple<…>`, a composite, or a frozen collection — is read as `Null`, i.e. a legal value is reported as absent.
2. The four exempt types are matched by **literal spelling**, so a `text`/`blob` column whose `data_type` arrives in the internal spelling (`utf8type`, `bytestype`) falls into the same collapse — and `parse_clustering_value` itself accepts those internal spellings, so the two sites disagree about what a type is called.
3. Downstream, `Value::Null` in a clustering pseudo-cell is therefore **ambiguous** between "Cassandra wrote NULL" (which, for a row's clustering, it cannot — see below) and "Cassandra wrote a legitimate EMPTY".

## Why this is not merely cosmetic

[...]

Issue #3977, opened 2026-09-02
main is RED: tooling-tests FAILs on clean origin/main (census guard vs dataset-manifest alternations) — blocks every lane's gate


pmcfadin:
**`main` is red: `tooling-tests` FAILs on a clean `origin/main` checkout, so no lane can obtain a passing
gate of record until this is fixed.** Found while running the gate of record for #3751 (PR #3885), and
isolated to main rather than to that branch.

## Reproduction on a clean checkout — nothing of #3751 involved

```bash
git worktree add --detach /tmp/clean-main origin/main
cd /tmp/clean-main && bash scripts/tests/test_agent_gate_census.sh
```

```
FAIL - R1: three-token component-status alternation(s) survive (VACUOUS omitted) in:
  /tmp/clean-main/scripts/tests/test_check_dataset_manifest.sh — remedy: add VACUOUS to each,
  or (if the site is a different artifact's vocabulary) extend that alternation so it is not the bare three

component census guard: 118 passed, 1 failed
```

`test_agent_gate_census.sh` is enrolled in `agent-gate.sh`'s `tooling-tests` roster, so this single
assertion FAILs that component, which FAILs the whole gate. Measured on PR #3885's gate of record:
`tooling-tests: FAIL (2699s)` with every other one of the 37 components PASS.

## Cause: two commits on `main` that are individually fine and jointly incompatible

- **`bdaf2b6e1`** — `fix(#3625): agent-gate component census …` (#3916) — added the guard, which requires
  every three-token component-status alternation to include `VACUOUS`.
- A later commit grew **`scripts/tests/test_check_dataset_manifest.sh`** by ~558 lines carrying
  component-status alternations that omit `VACUOUS`.

Neither author could have seen this: the guard passes against the tree it landed on, and the manifest test
passes its own suite. It is only the *combination* that reds, and nothing gates the combination — which is
the same shape #3544 records for a tree that is behind, one directory over.

## Why this is urgent rather than tidy

Every lane running the mandated full gate now FAILs `tooling-tests` after ~45 minutes of compute, on a box
where `CQLITE_GATE_MAX_CONCURRENCY` is pinned to 1. The coordination lead reported today that **15 lanes
held a passing gate and none could be armed**; this defect means new gates cannot even reach a PASS. It
also cannot be waived: `scripts/ci/classify-docs-only.sh` exits non-zero for any branch touching compiled
[...]

Issue #3923, opened 2026-09-02
cargo test -p cqlite-ffi-common fails -D dead-code on MAX_DECOMPRESSED_SIZE (pre-existing origin/main red, invisible to every gate component)


pmcfadin:
**Class:** build/gate hygiene. Found incidentally by the #3782 lane; **reproduced on a clean detached `origin/main` worktree**, so it is not that branch's defect.

## The red

```
git worktree add --detach /tmp/probe origin/main
cd /tmp/probe && cargo test -p cqlite-ffi-common
```

fails `-D dead-code` on `MAX_DECOMPRESSED_SIZE` — `cqlite-core/src/storage/sstable/compression.rs:23`. The constant is unused at the feature set `-p cqlite-ffi-common` resolves for `cqlite-core`, and its `#[allow]`/gating does not cover that combination (its #1873 sibling is gated correctly and is the model for the fix).

## Why nothing caught it

This is the #3522 shape, one crate over. `cqlite-ffi-common`'s Rust tests were enrolled in the gate's `binding-rust-tests` component precisely so they would stop executing nowhere — but the component runs them at a feature set where this constant IS used, while a bare `cargo test -p cqlite-ffi-common` resolves a narrower one. So:

- `cargo clippy --workspace --all-targets` **compiles** the crate on every full gate and passes;
- `binding-rust-tests` **executes** its tests and passes;
- the bare per-package invocation — which is what `--lite`'s blast-radius mapping runs when a diff touches that package — **fails**.

That last path is how the #3782 lane hit it: touching `compression.rs` mapped the blast radius onto the package and ran the failing combination, exactly the #1716 "expect latent failures the first time you touch a long-unwired crate" pattern.

## Acceptance criteria

1. `cargo test -p cqlite-ffi-common` passes on `origin/main` with `RUSTFLAGS="-D warnings"`.
2. The fix is a **correct, minimal feature gate** on the constant (mirroring its #1873 sibling), not a blanket `#[allow(dead_code)]` — an allow would hide the next instance.
3. State which gate component, if any, would now catch a regression of this exact shape. If the honest answer is none, say so in the issue rather than implying coverage (#3493): the feature set that reds is executed by no component, and enrolling one is the durable fix.

## Note on the current state

[...]

PR #3989, opened 2026-09-03
fix(#3809): a row deletion may not lose its clustering identity — refuse rather than emit an unidentifiable tombstone (Finding 1 of 3)

### LINKED ISSUE #3809: Row-assembly bounds are not authoritative: tombstones can lose their clustering identity, and the per-column loop neither bounds by nor reconciles against after_cells_offset
**Class:** correctness, read + write path. Pre-existing on `main`; found by the closed census run for #3721. Three related findings, all in the same family as #3721 but **outside** its fixed scope. Lower confidence than #3808 — each needs confirmation before fixing.

## 1. A row tombstone can be emitted having lost its identity

`cqlite-core/src/storage/sstable/reader/parsing/row_decoder/compaction.rs:119-127`

```rust
None => { clustering.clear(); break; }
```

A pure row tombstone whose clustering pseudo-cells (#912) are incomplete is emitted as
`CompactionRowData::Tombstone { clustering: [] }` — **a tombstone with no clustering key**, i.e. one that no longer identifies which row it deletes, produced silently on the write path.

The code comment says it "falls back to the `None` bucket": that states the mechanism and is silent about the consequence — the exact wording pattern #3721 exists to remove (compare the original `// CRITICAL FIX: Stop parsing remaining columns ... we exit the loop cleanly`, which was honest about the mechanism and silent about the truncated row).

**Needs confirmation:** whether a downstream consumer treats `clustering: []` as "delete nothing" (harmless) or as something broader (harmful). That determines severity.

## 2. The per-column loop is bounded by `data.len()`, not by the row's own `after_cells_offset`

So a column decode can read **past the row body into the next row** and still succeed. Framing itself is safe — `row_data.rs:819` derives `next_offset` from the authoritative `row_size` — so this does not desynchronise the walk. But a mis-sized column yields **plausible garbage** rather than an error, which is a no-heuristics-adjacent failure: the value returned is not derived from authoritative bounds.

## 3. Nothing asserts `offset == after_cells_offset` after the per-column loop

A column set that **under-consumes** the row body returns `Ok` silently. This is the exact converse of the over-consumption check #3721 just added (`row_body_exhausted`, for bytes exhausted with columns outstanding). #3721 closed one direction; this is the other.

The census that found these flagged (2) and (3) as **the most likely source of a sixth #3721-class finding**, since they are absences of a check rather than early exits, so no early-exit census can classify them.

## Suggested acceptance criteria

1. Determine whether `clustering: []` tombstones are harmful downstream; if so, refuse rather than emit.
2. Bound the per-column loop by `after_cells_offset` (authoritative), not `data.len()`.
3. Assert `offset == after_cells_offset` after the loop; a mismatch in either direction is a named error, not a silent `Ok`.
4. Whatever the outcome, replace mechanism-only comments with consequence-stating ones on these paths.

## Provenance

#3721 closed census: 34 in-scope early-exit sites across 7 `row_decoder` files, classified framing-terminator / propagate / legitimate-end-of-data. 4 defects remained: three read-path siblings (fixed in #3721's PR), the compaction resurrection bug (**#3808**), and this. Related: #3721, #3808, #912, #28.

pmcfadin:
## Slice of #3809 — Finding 1 only. This PR deliberately leaves issue 3809 OPEN.

**This body carries no closing keyword for issue 3809, deliberately.** #3809 has three findings; this ships **one**. Findings 2 and 3 are implemented in **PR #3814** (#3721) and are not on `main` yet, so closing #3809 here would mark two unsatisfied acceptance criteria as delivered. Telemetry is stamped `--slice` (`closed_at: null`) per the #3550/#3559 mechanism. #3809 closes when #3814 merges, or on the lead's disposition — see the scope-disposition request on the issue thread (`REQ-3809-01`).

### What this fixes

`row_decoder/compaction.rs`'s `None => { clustering.clear(); break; }` arm emitted `CompactionRowData::Tombstone { clustering: [] }` — a row tombstone that no longer identifies which row it deletes — **silently, on the write path**. It now refuses instead.

**The harm if reached**, traced end to end: `clustering: []` ⇒ `extract_clustering_key_from_compaction` returns `None` ⇒ the tombstone joins `clustered_rows[None]`, the bucket that also holds the static row ⇒ its `deletion_time` becomes the group's `row_del` and `shadow_by_row_deletion` drops every cell in that bucket with `ts <= deletion_time`. Downstream of that, on a static-bearing schema the writer drops the tombstone entirely (resurrection); with no static column it reaches `rows.rs:822`, which writes no clustering prefix and never sets `IS_STATIC` — a row Cassandra's `Clustering.serializer.deserialize` mis-parses, since it reads exactly `types.size()` values.

### Read this before assessing severity: the refusal is NOT reachable from bytes today

This is **defence in depth, not a live-defect fix**, and the PR does not claim otherwise. Verified independently by the implementer and by `rust-reviewer`:

- `parse_clustering_prefix` (`row_framing.rs:1220-1333`) pushes exactly one value per `schema.clustering_keys` entry on **all four** state arms, or returns `Err`.
- `RowColumnResolution::clustering` is built full-arity at **all five** production `build` sites (in four files — `block_emit.rs` holds two) (`partition_driver.rs:214`, `block_emit.rs:74,376`, `block_emit_windowed.rs:58`, `data_access/compaction.rs:626`).
- The one truncated-`clustering_prefix_schema` route (`row_framing.rs:330`) is range-marker-only and never reaches `build_compaction_row_data`.
- Measured: **0 occurrences** of the `None` arm across the whole corpus (381 targets / 5516 tests).

#3809's AC1 says *"determine whether `clustering: []` tombstones are harmful downstream; if so, refuse rather than emit."* The determination is **harmful if reached, not currently reachable** — so the guard is what AC1 asked for, at the severity the reachability implies.

### An empty clustering is LEGITIMATE in two cases, so the guard is conditional

A blanket refusal would red on correct input, which is the guard agents learn to waive. Exempt: **a static row**, and **a table with no clustering columns**. Cassandra distinguishes `Clustering.EMPTY` from `Clustering.STATIC_CLUSTERING` by `kind()` (`cassandra-5.0.8` `Clustering.java:102,124`) — a distinction CQLite's `Vec` cannot express, which is why the exemption is on `is_static` (read from the authoritative `EXTENDED_IS_STATIC` flag) rather than on the vector being empty. `rust-reviewer` looked for a false-refusal route and found none.

### The guard covers BOTH paths carrying a row deletion, not just the pure tombstone

roborev (job 92) found that the first revision validated only the pure `CompactionRowData::Tombstone` arm, leaving the #932 coexistence shape — `Live { row_deletion: Some(_) }`, a row with surviving cells AND a row deletion — unvalidated. The harm is identical: a missing clustering still yields `None` from `extract_clustering_key_from_compaction`, so the row lands in the `None`/static bucket and its `deletion_time` becomes that group's `row_del`. Treated as a **blocker**, not a nit, because the published invariant claimed more than it enforced.

The check is now hoisted inside `if let Some((deletion_time, …)) = row_deletion` but **above** the pure-vs-coexisting branch (`compaction.rs:126-170`), so both arms are validated; the exemptions are untouched.

**Boundary, stated deliberately:** a row with **no** `row_deletion` is NOT validated. Without a `deletion_time` it can never become the `None` bucket's `row_del`, so `shadow_by_row_deletion` cannot drop anything on its account. Widening would red an entire compaction read over a shape whose worst outcome is a reconciliation this invariant does not own — a guard that reds on correct input is the guard agents learn to waive. Pinned both ways (`a_row_with_no_row_deletion_is_not_validated`).

### Rebased onto #3782, which forced a real design change — read this before the rest

`main` advanced to #3782/#3939 (*"preserve the fatal row-decode error kind and REFUSE at a proven-complete buffer"*) while this PR's gate was running. #3782 **does not subsume** Finding 1 — the two are the `Err` and `Ok` arms of one decode `match`: #3782 fires when the decoder returns `Err`, this fires when the decode returned `Ok` and the *row-build* step finds a short clustering, a step #3782 left infallible. But both PRs had changed `on_data_row` to `Result<Option<usize>>`, leaving **two contradictory contracts on one `Err` channel**: #3782's driver *downgrades any `Err` mid-window to `Ok(ParseStep::NeedMore)`*, so a mechanical rebase would have turned every refusal on a `BufferExtent::Window` path into a **refill request** — the exact truncation this guard exists to prevent, with every test still green.

**Resolution: `on_data_row` returns a distinct outcome enum, not a `Result`** — `DataRowOutcome { Decoded(usize), Declined, DecodeFailed(Error), Refused(Error) }`, the shape the trait already uses for `MarkerOutcome`. A kind test was rejected because it needs either a new public `Error` variant (API drift nothing detects, #3366) or message-text matching (#28 forbids). Dropping `Result` is the load-bearing part: it makes the data-losing default **unrepresentable**, since with a `Result` a future `?` in a policy body would silently route a refusal into the *tolerated* channel — the defaulted-flag defect #3782 itself removed, one layer up. Both drivers return `Refused` unconditionally.

**The two dispositions, one doc, no contradiction:** a `Window` caller drives with `at_final_chunk == false`, where `DecodeFailed` becomes `NeedMore` — correct for it *and only* for it, because a decode failure asks *were all this row's bytes present?* and more bytes can still change the answer (614/614 measured tolerations are non-final). A `Refused` asks nothing about bytes: the decode returned `Ok`, the row is fully framed, and no refill can make a short clustering full-arity. So `NeedMore` is **inapplicable**, not merely weaker, and a refusal is never tolerated at any extent.

Pinned by `partition_driver_tests.rs::refusal_on_a_window_extent_refuses_and_never_requests_more_bytes` (+ an every-extent sibling), **RED-verified**: degrading `Refused` to the tolerant branch yields `got NeedMore with 0 rows forwarded`.

### Why `on_data_row` changed shape at all

Independently of the #3782 reconciliation above, the refusal has to *reach* the caller. `on_data_row` previously answered `Option<usize>`, where the natural failure answer is `None` — and `partition_driver.rs` reads `None` as **end-of-partition**, so a refusal would have silently truncated the rest of the partition on the write path: the #3721 swallow one level up. The final shape is `DataRowOutcome` (see the rebase section above, which supersedes an earlier `Result<Option<usize>>` step); the three call sites carry no behaviour of their own.

### Oracles — no CQLite round-trip is used for any framing property

Per #3042, a CQLite-written + CQLite-read round trip is invariant to a uniform framing error, so every expectation is derived from Cassandra source or Cassandra-written bytes:

| case | oracle |
|---|---|
| refusal on non-static/clustered/short clustering (matched on the `Error::Corruption` **variant**, not message text) | `Clustering.java` `Serializer.serialize` arity assert |
| static negative control, four declared arities | `Clustering.java:102,124` — `EMPTY` vs `STATIC_CLUSTERING` by `kind()` |
| no-clustering-column control; complete-clustering control | as above |
| #932 coexistence refusal at arities (1,0)/(2,0)/(2,1)/(3,2), plus complete-clustering discrimination (must stay `Live` carrying BOTH the deletion and the surviving cell) | `Clustering.Serializer` arity assert + row field order (prefix before `row_size`/deletion/cells), `docs/sstables-definitive-guide` Ch.5 "Row Structure" |
| end-to-end `test_tomb.static_with_tombstones` → `Tombstone { clustering: [("ck", Integer(2))], deletion_time: 1_609_545_600_000_000 }` | committed `nb-1-big-Data.db.jsonl` sstabledump golden |

### DECLARED GAP — the `is_static` exemption has ZERO byte-level coverage

Stated because a test doc claiming coverage it does not have is worse than no test. The e2e golden's static row is a **live** `static_block` (cells present, no `deletion_info`), so it never reaches the tombstone branch. Measured: **166 goldens, 9 with a `static_block`, 0 with a static-block deletion** — and CQL cannot express a static-row deletion, so no fixture can be generated. The exemption is therefore covered at the **scalar level only**, and the test doc says exactly that rather than implying more.

### No longer in this PR: the `compression.rs` fix, and the file-size opt-out with it

An earlier revision of this PR gated `MAX_DECOMPRESSED_SIZE` by `#[cfg]` — a fix to pre-existing P0 **#3923** (`cargo test -p cqlite-ffi-common` failing `-D dead-code`), which this diff exposed via the #2658 dependent-crate fan-out. **#3782 landed the identical fix on `main` independently**, so `compression.rs` is now **byte-identical to main and out of this diff entirely**.

Consequence worth stating because it reverses an earlier instruction in this body: **`CQLITE_ALLOW_FILE_GROWTH=1` is no longer required.** Both post-rebase lite rounds ran without it and `file-size` returns a real `PASS`, not `OPT-OUT`.

Two campsite splits were forced by `row_decoder/mod.rs` being **881/800 on `main`** — any new `mod`/`use` line there reds `file-size` — so `DataRowOutcome` lives beside `MarkerOutcome` in `partition_driver.rs`, the row builder is a `#[path]` child of `compaction`, and `partition_driver.rs`'s tests moved to an external `partition_driver_tests.rs`.

<details><summary>Superseded: the original compression.rs declaration</summary>

### Declared: this PR carries a fix to a PRE-EXISTING `main` defect

`49aa89858` gates `MAX_DECOMPRESSED_SIZE` by `#[cfg(any(feature = "lz4", …, "zstd"))]`. It is dead in a `default-features = false` build under `-D warnings` — #1873 gated the *function* and missed the *constant* — and my diff exposed it via the #2658 dependent-crate fan-out (`cqlite-ffi-common` links cqlite-core feature-less), so `--lite` could not go green without it. It follows the convention the same file already uses at `compression.rs:103` for the sibling function.

**Consequence for the gate of record: it must run with `CQLITE_ALLOW_FILE_GROWTH=1`.** rustfmt splits the `#[cfg]` attribute (918 → 928 lines) and `compression.rs` is already over the ~800 campsite target, so the ratchet fails on any growth. Splitting that file is epic **#1116**. An earlier revision used `#[allow(dead_code)]`; `rust-reviewer` correctly rejected that in favour of the precise `#[cfg]` plus the documented growth opt-out. **Correction to an earlier draft of this body:** the removal of a commented-out `use async_trait` line IS still in the diff — it is retained as a deliberate standalone cleanup, and it is **not** load-bearing for the ratchet (the file grows +10 either way, so the opt-out is required regardless). It is not being used to buy a line budget.

</details>
### A roborev finding was REFUTED, not fixed — and implementing it would have redded correct input

roborev job 93 raised a Medium: the guard counts any matching pseudo-cell as recovered, **including `Value::Null`**, so a full-arity tombstone with null clustering components passes. Right in principle, and **wrong to implement here**. Verified against the pinned `cassandra-5.0.8`:

- The header encodes **two independent** states — bit `(2i)+1` = null, bit `2i` = empty (`ClusteringPrefix.java:548` `makeHeader`; `isNull`/`isEmpty` at `:565`/`:572`). `deserializeValuesWithoutSize:514` maps EMPTY to `EMPTY_BYTE_ARRAY`, **explicitly distinct from null**, and `AbstractType.java:513` `allowsEmpty()` holds for 17 families, of which `StringType`/`BytesType`/`TupleType`/`AbstractCompositeType` have `isEmptyValueMeaningless() == false` (`:458`) — so **empty is a meaningful non-null value** there.
- A row's clustering component **cannot** legitimately be NULL: the CQL write path rejects null clustering values before any `Clustering` exists (`SingleColumnRestriction.java:195`, `MultiColumnRestriction.java:242`), and the one derived path that contemplates it is compact-table-gated (`ViewUpdateGenerator.java:179`) while `TableMetadata.isCompactTable()` is unconditionally `false` in 5.0 (`:291`).
- **No `Kind` ordinal reaches this builder at all** — a row's prefix carries no `Kind` byte (`UnfilteredSerializer.java:191`/`:481` use `Clustering.serializer`, header only; the ordinal is written on the marker path at `:287`). The finding's "reserved prefix states" half is moot.

**Why the proposed fix would have broken correct data:** `row_framing.rs:1300-1310` maps **IS_EMPTY to `Value::Null`** for every clustering type outside the four literal spellings `text|varchar|ascii|blob`, and `row_data.rs:158-164` pushes that value verbatim. So `Value::Null` at the counting site is **ambiguous** between unreachable-NULL and legitimate-EMPTY, and refusing it would reject a legal empty `tuple<…>`/composite clustering — or a `text`/`blob` column whose `data_type` arrives in the internal spelling (`utf8type`), which `parse_clustering_value` itself accepts. Type-scoping does not rescue it, since `allowsEmpty()` covers essentially every clustering-eligible type.

**And no test would have caught it.** Measured over 169 committed JSONL references / 15782 clustering-bearing rows: **0 empty and 0 null clustering components**, and no `tuple` clustering column exists. There is no control in either direction, so a green suite is not evidence here — the source argument decides.

So the guard stays **arity-only**, with the reason recorded **in code at the permissive branch** (`compaction_row.rs:441-483`, plus a pointer at `compaction.rs:154-156`) per the standing rule that a deliberately permissive branch records its reason where the branch is. roborev round 4 then returned `findings: NONE` without the refutation being waived.

**The real defect underneath it is filed as #3988** — the IS_EMPTY/IS_NULL collapse is a genuine read-path divergence from Cassandra, not currently observable on the committed corpus (hence its survival), and needing a *generated* fixture since none exists.

### One unmentioned behaviour change, for completeness

A static row that somehow surfaced a *partial* clustering prefix now emits `Tombstone { clustering: <partial> }` where `main` emitted `[]`. Downstream is identical — `extract_clustering_key_from_compaction` returns `None` for a partial prefix either way — so there is no consequence, and the path is unreachable (a static row provably yields zero clustering values). Recorded because the `compaction.rs:177` comment asserts the emptiness rather than enforcing it.

### DECLARED GAP 2: the line that joins the builder to the untolerated channel is untested

`row_decoder/compaction.rs:606` — `Err(e) => return DataRowOutcome::Refused(e)` — is the single line that makes this PR's central design claim true, and **no test exercises it**: the build tests stop at the builder's `Err`, and the driver tests inject `Refused` from `StubPolicy`. Flipping it to `DecodeFailed` reds nothing today. Harmless while the shape is unreachable from real bytes, but stated here rather than left silent, and filed for a test in **#3993**. Found by the C intent audit, not by the test suite — which is the point.

Also filed to #3993: `compaction_build_identity_tests.rs:302-309`'s no-clustering-column case cannot fail against any change in this diff (it needs the exemption dropped **and** the predicate tightened to `!=`) and, unlike its sibling at `compaction_row_tombstone_identity_tests.rs:104-121`, does not say so.

### Review record

- `rust-reviewer`: **no blockers, 9 nits — all 9 fixed** in `49aa89858`/`85234f07d`/`af36d2fa3`. Four were must-fix by CQLite doctrine rather than nits: a "NEGATIVE CONTROL" that discriminated nothing, a diagnostic assertion satisfied by accident (its needles `"2"`/`"1"` matched the literal `#912` in the message), a doc claiming static-row coverage that does not exist, and the `compression.rs` suppression above.
- roborev (sanctioned wrapper, `--agent codex --model gpt-5.6-sol`): round 1 job 91 `PASS`/`findings: NONE`; **round 2 job 92 `FAIL`/`findings: PRESENT (1)`** — the coexistence gap above, fixed in `c567111e5`/`7c046636a`; round 3 job 93 `FAIL`/`PRESENT (1)` — **refuted, see above**; round 4 job 94 `PASS`/`findings: NONE` at `15001c8c6`; then, after the #3782 rebase, round 5 job 99 `FAIL`/`PRESENT (1)` (a `write-support` cfg gate lost in a file move — fixed) and **job 102 re-decided at the certified head `beaa5aa6b` via `--recheck-job`: `RESULT: PASS`, `findings: NONE`, `prompt-content: PASS (11/11)`**. `15001c8c6` is **not** an ancestor of the certified head and is recorded here only as history; per #3460 the merge-gating roborev round and the gate of record are both at `beaa5aa6b`. Round 1 was clean at an earlier sha and round 2 then found a pre-existing gap, so the re-reviews were not redundant. **No `roborev-defer` and no waiver of any kind was used.** After the #3782 rebase the diff was re-reviewed from scratch (`rust-reviewer` + a fresh roborev round at the rebased sha), because the rebase changed a driver contract rather than merely moving lines.

**One gate of record was already spent and came back `RESULT: FAIL` — 36/37 PASS with `tooling-tests` the sole failure, which is pre-existing P0 #3977 on `main`.** Proof it is not this diff: `git diff origin/main...HEAD -- scripts/` is **empty**, so that lint's entire scan input is byte-identical to main. It was **not waived** — the #3042 prose waiver's precondition fails because this diff carries `cqlite-core/src`, and `premerge-assert` refuses a `FAIL` verdict regardless — and no source was patched to green it. The fix is #3995; this PR re-gates from scratch after it lands, per the lead's directive not to spend a gate of record until then.
- Test discrimination was verified by a **hand-run RED control**, recorded in the new module's header: with the check moved back inside the tombstone branch, exactly one case fails, building `Live { simple: [v], row_deletion: Some(..) }` with no clustering. A control that cannot fail is not a control.
- Base was **stale in a way that mattered** — `base-staleness.sh` reported `STALE-RECOGNISED` with the one commit behind touching gate-global `scripts/agent-gate.sh`. A gate script behind `origin/main` cannot certify (#3544), so the branch was rebased **before** the gate of record. Now `NO-STALENESS-RECOGNISED`.

### Found and NOT fixed — five defects outside this issue's scope

Proposed as separate issues on the #3809 thread (`REQ-3809-02`), not fixed here:

1. **`encoding.rs:495` vs `:520` disagree on `DeleteRow`** (`is_static_row_mutation` true / `is_static_operation` false), so a `clustering_key: None` row tombstone on a static-bearing schema reaches **neither** writer path and vanishes from the compacted SSTable — durable resurrection. This is the mechanism that makes Finding 1 harmful; it fires independently of Finding 1, and this PR narrows its reachability without fixing it. Needs a Cassandra-written-and-read oracle (#3042).
2. `rows.rs:822` can emit a clustering-prefix-less row on a clustered table.
3. `row_framing.rs:1258` reads one clustering header vint and shifts `i*2`; Cassandra batches headers per 32 columns (`ClusteringPrefix.java:455-475`), so a >32-clustering-column table shifts ≥64.
4. `big_promoted.rs:392-410` performs ~596 misaligned decodes per wide partition on real corpus bytes.
5. `data_access/compaction.rs:494` silently omits a clustering-arity-mismatched row from its own ordering verification.

### Not in this PR

`row_data.rs` is **untouched**. Findings 2 and 3 are not implemented here — see PR #3814.

---

*Post-merge note: an earlier revision of this body opened with "deliberately does NOT close #3809". GitHub's linkifier reads the substring `close #3809` as a closing keyword regardless of the preceding negation, so `closingIssuesReferences` recorded a close and issue #3809 auto-closed on merge. It has been reopened — AC2/AC3 are delivered by PR #3814 and are not on `main`. The wording above no longer places a closing keyword adjacent to the issue number. **A disclaimer that names the keyword it is disclaiming still triggers it.**


pmcfadin:
**C intent audit: PASS.** AC1 and AC4 `satisfied` with public-surface evidence; AC2/AC3 correctly deferred to PR #3814 (`row_data.rs` confirmed untouched, 0 hits). Oracle discipline clean; test discrimination verified by naming which test fails under which reversion; all four load-bearing claims in this body independently checked, including the declared `is_static` gap re-scanned across all 169 tracked goldens.

Four non-blocking nits. Two were inaccuracies in this PR body and are **corrected above** (a five-site list labelled as four; and the `use async_trait` deletion, which is still in the diff as a standalone cleanup and is *not* load-bearing for the file-size ratchet — the body previously read as if it had been reverted). The remaining two are batched into **#3993** and are not fixed here.

Review trail, for the record: `rust-reviewer` no blockers / 9 nits, all 9 fixed. roborev jobs 91 PASS-NONE, 92 FAIL-1 (#932 coexistence gap, fixed), 93 FAIL-1 (**refuted** on `cassandra-5.0.8` source — the proposed fix would have refused a legal empty clustering; real defect underneath filed as **#3988**), **94 PASS / findings NONE** at the certified sha. **No waiver and no `roborev-defer` was used.**

pmcfadin:
Merged at 2026-09-03T10:10:05Z as `9a159405e` (squash), armed via `gh pr merge --auto` after `PREMERGE: OK`.

**Gate of record** (third attempt; the first two FAILed on `tooling-tests` for reasons provably not this diff — #3977, then a third instance of the #3969 broken-pipe flake): `RESULT: PASS`, **37/37**, `/tmp/gate-3809-record3.txt`.
`commit: 808f0d8` · `tree-start`/`tree-end: 808f0d8a1f33` · `tree-integrity: PASS` · `dirty: no` · no `nested-under:` · `component-set: PASS (37/37)` · `file-size: PASS (0s)` (a real PASS — no `CQLITE_ALLOW_FILE_GROWTH` opt-out) · **`tooling-tests: PASS (2465s)`** — the #3969 flake did not recur, so no new instance was recorded against it.

**roborev**: job **107**, fresh round at the certified sha `808f0d8a1` — `RESULT: PASS`, `findings: NONE`, `prompt-content: PASS (11/11 code census paths)`, tokens 347195 in / 294144 cached (genuine band, not the ~18.7k vacuous baseline). No waiver and no `roborev-defer` was used on this PR. Delivery total across 7 jobs: **3 findings, all Medium, all fixed pre-merge** (jobs 92, 93, 99), 0 deferred.
**C intent audit**: PASS (carried — the #3782 rebase replayed #3890, which touches only `data_access/bti*.rs`, disjoint from this diff's `row_decoder/**` + `compaction_row*`).

`PREMERGE: SCOPE` and the base-staleness advisory are recorded rather than glossed: this verifies **the gate of record at `808f0d8a1`**, *not* certification against current main. Advisory at merge time: `behind 2`, `blast-radius 0 RECOGNISED`, verdict `NO-STALENESS-RECOGNISED` (non-blocking, #3650 slice 1).

**Two follow-ups for the lead, neither merge-blocking:**
1. This PR's body carries a **keyword-in-negation defect** — GitHub's linkifier matched `close #3809` inside *"deliberately does NOT close #3809"*, so `closingIssuesReferences` recorded `closes #3809` and the issue auto-closed at 10:10:07Z. I **reopened** #3809 per `REQ-3809-01` and the documented slice intent.
2. Consequently the **telemetry ledger line is unstamped**. `record --slice` is refused (a slice PR must close NOTHING) and so is the non-slice form (it wants the close recorded, which the reopen undid). Both remaining routes mutate an artifact — the merged PR body, or the issue state — so per #3559 ("closing an issue to satisfy the tool, or hand-appending past the validator, are both FORBIDDEN") I stamped nothing and left the call to the lead.

PR #3994, opened 2026-09-03
docs(#3778): make cell_path_key's nested-consumption header true — #3811 refuses three of four classes; duration is a parity question

pmcfadin:
## Slice 1 of #3778: make `cell_path_key`'s nested-consumption doc TRUE

**This PR closes nothing.** It is a deliberate slice — #3778 stays open because it also owns the
`duration` disposition, which is an unresolved parity question (`COORD-REQ:3778-03` on the issue).
Telemetry for this delivery is `--slice` (#3550/#3559).

### Why

`complex_column/cell_path_key.rs`'s module header still described nested consumption as unchecked
below the first level, in four classes. **#3811 has since refused three of them**, so the header
asserted a live defect that no longer exists — and it *self-contradicted*, because a later paragraph
in the same block already said #3811 landed. It also attributed the residue to **#3723**. In this
repo a doc claim decays exactly like a comment, so a header that misdescribes the decoder is a real
defect: it is what stops the next person looking.

### What changed

Doc-only, one file, **+72/−93 — net −21 lines (798 → 777)** against the 800-line ratchet. No code,
signature or test touched; no `CQLITE_ALLOW_FILE_GROWTH`.

- The stale residual is recorded in the **past tense**, crediting #3811 with `raw_value.rs:129-130`
  → `raw_value/reporting.rs:67` (`require_fully_consumed_raw`, accepted set exactly `{n}`) and
  noting the reporting form is `pub(super)` inside `raw_value` (`reporting.rs:101-108`), so no other
  `row_decoder` child can name it and **no bounded caller can opt out**.
- Each refuted class now cites its refusal **by function name as well as line**, paired with its
  exact-encoding control, so the citation survives line drift:
  `bounded_int_over_width_is_refused` (`:654`); `bounded_tuple_with_trailing_byte_is_refused`
  (`:596`, control `:581`); `nested_udt_trailing_garbage_is_refused` (`:724`, control `:699`, plus
  the `structural_`/`inline_` twins `:831`/`:1035`); `bounded_list_with_trailing_byte_is_refused`
  (`:547`) and its set sibling (`:562`), control `:535`. **All ten verified line-exact.**
- A **SHORT** encoding remains legal (`TupleType.split` rule 1) — stated as deliberate, not an
  oversight.
- `duration` is named as the one remaining class, **on two paths that are not this one**:
  `raw_type_value.rs:341-390` (the third `parse_vint`'s remainder is bound to `_remaining` and
  discarded, then `offset += duration_len` advances by the DECLARED length) and
  `cell_value_scalar.rs:366-373` (leftover bytes `warn!` only).
- Attribution fixed: `#3723` → **0** hits; `#3811` ×4, `#3778` ×1, `#3820` ×1 (unification note kept).

### The oracle finding this PR records, because it changes what #3778 should do

Read at the pinned tag (no clone on this box — fetched the file and read its text, not a summary):
`cassandra-5.0.8:src/java/org/apache/cassandra/serializers/DurationSerializer.java` `deserialize`
reads three VInts and returns; `validate` (`:80-105`) enforces only `size >= 3`
(*"Expected at least 3 bytes for a duration (%d)"*), the months/days 32-bit bounds and the same-sign
rule. `grep "remaining|hasRemaining|limit()"` over that file plus `db/marshal/DurationType.java`:
**zero hits**. `TupleType.split` rule 4 refuses trailing bytes at the **tuple framing** level, not
inside one element's declared length.

So **Cassandra tolerates trailing bytes inside a duration value**, and refusing them would make
CQLite stricter than Cassandra. The doc therefore presents the disposition as **open, tracked as
#3778, with no fix direction asserted** — which is correct under every outcome of
`COORD-REQ:3778-03`.

### Verification

| | |
|---|---|
| `--lite` | **RESULT: PASS** — `commit: dd909a7`, `dirty: no`, `tree-integrity: PASS`; file-size / fmt / clippy / roborev-lints / scoped-tests all PASS; `scoped-tests` verified **3708 tests passed, 170 test binaries** |
| roborev | **RESULT: PASS**, `findings: NONE`, `reviewed-sha 05134c947..dd909a718`, `code-free: PASS`, `sha-assert: PASS`, `prompt-content: PASS (1/1)`, both vacuity tiers PASS, 126597 input / 93056 cached tokens |
| roborev round 1 | `FAIL`, 1 finding — **correct, and the error was mine**: I had cited `:495`/`:515` for "nested tuples and UDTs", but those are the TOP-LEVEL marshal/registry UDT arms and `:515` is the rule-1 boundary, not a refusal. Fixed in `dd909a718`; the pair is kept, described accurately as top-level. |

### ⛔ The gate of record has NOT been run, deliberately

Per the lead directive on #3977: gate only after the **#3923 / #3977** hotfix lands. Both were open
at the time of writing (#3977 `main is RED: tooling-tests FAILs on clean origin/main`, P0). The
`--lite` block above is **not** a gate of record and must not be read as one — its own header says
`MODE: lite … full agent-gate.sh must PASS once before merge`. **Do not merge this until a full
`AGENT-GATE SUMMARY` with `RESULT: PASS` exists for the certified sha.**

Refs #3778. Related: #3811 (landed the three refusals), #3723 / #3793 (prior owner; Site 6 declared
there), #3820 (unify the two consumption asserts), #3923 / #3977 (gate blockers).

### DIFF
--- a/cqlite-core/src/storage/sstable/reader/parsing/row_decoder/complex_column/cell_path_key.rs
+++ b/cqlite-core/src/storage/sstable/reader/parsing/row_decoder/complex_column/cell_path_key.rs
@@ -155,99 +155,91 @@
 //! | duration | the SHARED decoder's own assert, one layer down: its `"duration"` arm reports where the third VInt ended, so this module no longer walks the framing itself (#3631 unification) |
 //! | unknown type → opaque `Value::Blob` | whole slice by construction; also raises the caller-aggregated opaque-key signal (the `warn!` is the caller's — see above) |
 //!
-//! ## The residual: NESTED CONSUMPTION IS UNCHECKED AT EVERY LEVEL BELOW THE FIRST
+//! ## Nested consumption: THREE OF FOUR CLASSES ARE NOW REFUSED
 //!
-//! Stated broadly because an earlier revision stated it too narrowly — it named
-//! only fixed-width scalars, and roborev round 9 was right that the class is
-//! wider. The consumption rule above applies to the OUTER value only. A nested
-//! value is bounded by its own `[i32 BE len]` prefix, the parent advances by that
-//! DECLARED length, and nothing compares it with what the child actually read. So
-//! trailing bytes inside any nested value are silently ignored:
+//! An earlier revision of this header carried a broad residual — "nested
+//! consumption is unchecked at every level below the first" — naming four classes:
+//! fixed-width scalars, nested tuples/UDTs, nested collections and `duration`. The
+//! first three of those are now REFUSED, so that heading is stated here in the PAST
+//! TENSE and only `duration` survives, as a DECIDED disposition rather than a gap.
 //!
-//! * **fixed-width scalars** — `parse_value_from_raw_bytes`'s guards are
-//!   `data.len() < N`, not `!= N`, so a 5-byte `int` element decodes from its
-//!   first 4 bytes;
-//! * **nested tuples and UDTs** — the decoders iterate the DECLARED components and
-//!   stop, leaving any extra components unread;
-//! * **nested collections** — the element loop runs the DECLARED count and stops;
-//! * **`duration`** — reads three VInts and ignores whatever follows (at the
-//!   OUTER level this is now caught by the shared decoder's exhaustion assert;
-//!   the residual is a duration NESTED inside another value).
-//!
-//! Consequence, and it is this PR's own headline symptom one level down: for
-//! `frozen<list<int>>` the byte strings `[count=1][len=4][4B]` and
-//! `[count=1][len=5][5B]` both satisfy the OUTER rule and decode to the same
-//! `List([Integer(x)])`, so two distinct cell paths become one logical key and can
-//! collapse a Python map entry.
-//!
-//! WHY IT WAS NOT FIXED HERE, measured rather than asserted: the root cause was
-//! that `parse_value_from_raw_bytes` had NO consumption channel at all — it
-//! returned a `Value` — so no caller could check even if it wanted to. That was
-//! a shared-decoder tightening with its own oracle, i.e. its own PR, and it has
-//! since LANDED as **issue #3811**: that function is now a thin wrapper over
-//! `parse_value_from_raw_bytes_reporting` plus `require_fully_consumed`, so
-//! every bounded caller of the short name inherits the rule and the collapse
-//! described above no longer occurs. The UNIFICATION #3811 asked for has now
-//! LANDED as **#3631**: there is ONE `require_fully_consumed`, in
+//! The tightening landed as **#3811** (`parse_value_from_raw_bytes` gained a
+//! consumption channel: it is a thin wrapper over
+//! `parse_value_from_raw_bytes_reporting` plus a consumption assert, so every
+//! bounded caller of the short name inherits the rule) and the UNIFICATION #3811
+//! asked for landed as **#3631**: there is ONE `require_fully_consumed`, in
 //! `typed_value.rs`, which every layer names, and this module's `duration`
 //! consumption special-case is GONE with it — the shared decoder's own assert
 //! reaches a trailing byte first (see `decode_reporting_consumption` below).
 //! What this module still owns is the fixed-width ALLOWED-width table, which is
 //! cell-path-specific and has no equivalent on the value side.
 //!
-//! It is also deliberately NOT patched with a second framing walk here: a
-//! call-site validator that must know about every decoder is precisely the shape
-//! this module replaced.
+//! The refusals are pinned by name in
+//! `row_decoder/raw_value/issue_3811_consumption_demo_tests.rs`:
 //!
-//! Tracked as issue **#3723**. Worth recording for whoever takes it: nothing found
-//! depends on the lenient acceptance, so the tightening looks SAFE in principle —
-//! Cassandra cannot drop a UDT field (`AlterTypeStatement` at `cassandra-5.0.8`
-//! offers only `AddField`, `RenameFields`, `AlterField`), so schema evolution
-//! yields SHORT encodings, which are legal and already handled, never trailing
-//! ones. The blocker is scope and oracle, not risk.
+//! | class, as the old residual named it | now refused, by these tests |
+//! |---|---|
+//! | fixed-width scalars decoding from a prefix (a 5-byte `int`) | `bounded_int_over_width_is_refused` (control: `bounded_int_exact_decodes_ok`) |
+//! | nested tuples and UDTs leaving extra components unread | `bounded_tuple_with_trailing_byte_is_refused` (control: `bounded_tuple_exact_decodes_ok`); `nested_udt_trailing_garbage_is_refused` (control: `nested_udt_exact_decodes_ok`), plus its twins `structural_nested_udt_trailing_garbage_is_refused` and `inline_udt_trailing_garbage_is_refused` |
+//! | nested collections stopping at the declared count | `bounded_list_with_trailing_byte_is_refused` and `bounded_set_with_trailing_byte_is_refused` (control: `bounded_list_exact_decodes_ok`) |
 //!
-//! ## SYMPTOM: A PYTHON READ CAN SILENTLY LOSE A MAP ENTRY (issue #3612, R6-F1)
+//! Consequently the collapse the old residual led with is GONE: measured on this
+//! tree, a `frozen<list<int>>` cell path of `[count=1][len=5][5B]` (13 bytes) is now
+//! REFUSED ("decoded only 4 of 5 byte(s)" — the NESTED element's width), while the
+//! well-formed `[count=1][len=4][4B]` (12 bytes) still decodes to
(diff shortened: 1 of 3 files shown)


pmcfadin:
## Re-authored on a new base — this PR's history was force-pushed, and here is exactly why

**Previous head `dd909a718` is obsolete and has been replaced.** Not a cosmetic rewrite: the earlier
commits made claims that stopped being true while this lane was blocked, and keeping them would have
shipped citations pointing at code that no longer exists. The superseded commits are preserved at
branch `obsolete-3778-docs-backup` if anyone wants the diff.

### What invalidated them

**#3631 / #3820 landed the consumption unification.** On today's `main` there is ONE
`require_fully_consumed`, at `row_decoder/typed_value.rs:91`, and `require_fully_consumed_raw` is
**no longer a function at all**. My earlier prose cited it at `raw_value/reporting.rs:67` and
described #3820 as *pending*. I attempted a rebase, hit a conflict in exactly those paragraphs, and
**aborted rather than resolving it** — resolving in my favour would have re-asserted dead symbols,
which is the decaying-doc-claim defect this file has already been burned by twice.

So I reset to `origin/main`, re-measured, and re-authored. Everything below was verified on the
current tree, not carried over.

### What this PR now contains (3 commits, 3 files)

**`387a68dd3` — `test(#3778)`: the substance.** A new module
`row_decoder/issue_3778_duration_parity_tests.rs` (256 lines) with **7 tests, all passing**
(`7 passed; 0 failed`, in the `cqlite-core` lib binary — they genuinely execute, 3818 filtered out):

- exact-encoding controls plus trailing-byte tolerance at **both** tolerant sites, kept separate
  because they are different framings — `raw_type_value.rs`'s arm (a duration NESTED in a frozen
  composite; its errors read `Frozen element '{}'`) and `cell_value_scalar.rs`'s arm (a plain
  top-level CELL, **not** nested);
- `two_encodings_differing_only_in_trailing_bytes_decode_equal_{nested,cell}` — the scope
  consequence the ruling accepted **explicitly**: two byte strings differing only in trailing bytes
  inside `duration_len` decode to one equal `Value::Duration`, deliberately;
- `the_raw_value_path_refuses_what_these_two_sites_accept` — pins the three-way asymmetry so it is
  visible rather than surprising, and warns against "unifying" by tightening (Option B renamed).

New module rather than inline tests **because of the ratchet**: `raw_type_value.rs` is 1037 lines,
already over the 800 threshold, so growing it would trip `file-size`. No `CQLITE_ALLOW_FILE_GROWTH`.

**`03ee594a1` — `docs(#3778)`: the AC7 correction.** `cell_path_key.rs` **799 → 792**, net −7,
under the ratchet. The false residual bullets are gone or in the past tense; three of four classes
are now cited as refused **by test function name as well as line** (a previous round of this work was
correctly failed by roborev for citing tests that did not say what the doc claimed); `#3723` → 0
hits, attribution corrected; #3631's own new unification text kept intact.

**`a397044bd` — `docs(#3778)`:** cite the follow-up as **#4038** by number. Both files previously
said only "a separate follow-up", which a reader cannot locate.

### The ruling this implements

`COORD-REQ:3778-03` → **Option A**: do NOT refuse trailing bytes inside a duration's declared
extent; pin the tolerance as parity-correct, correct the doc, declare the residual. **No behaviour
change in this diff** — no new `Err`, no changed guard, no `warn!` removed. Option B was refused
because it converts currently-tolerated live reads into hard failures, knowingly stricter than
Cassandra, with no oracle support.

Oracle, read at the pinned tag: `DurationSerializer.deserialize` reads three VInts and returns;
`validate` (`:80-105`) bounds only `size >= 3`, the 32-bit months/days range and the same-sign rule;
zero `remaining`/`hasRemaining`/`limit()` hits across `DurationSerializer.java` and
`db/marshal/DurationType.java`. `TupleType.split` rule 4 governs the **tuple framing** level, not
inside one element's declared length.

### ⛔ Still NOT gated, and this PR is NOT mergeable yet

Two fleet P0s red `tooling-tests`, a full-gate component: **#4029** (case 17.5 fails on pristine
`main`) and **#4032** (intermittent broken-pipe leak). Per the coordinator the sequence is #4039,
then #4032, then a fresh gate is worth starting.

**Disclosed against my own interest:** this branch is **not immune** to #4029 — `035fdbf32` is an
ancestor of its head. It was immune at the previous base (`05134c947`, which predates the guard) and
I imported the P0 by rebasing. I am holding here rather than rolling back, because the content above
depends on the post-#3631 base and #4032 blocks gating regardless; say the word if you want immunity
reclaimed instead.

`--lite` and roborev are re-running against this head; I will post both verdicts. Closes nothing —
#3778 stays open until the lead confirms Option A is discharged.

pmcfadin:
## Final pre-gate verdicts at head `0c460954c` — reviewed clean, lite-green, and blocked on two fleet P0s

Completing this PR's record while the fleet waits on #4039, per the coordinator's instruction to
finish everything that needs no gate.

### Verification (both re-run at the exact head that would merge)

| gate | verdict |
|---|---|
| **`--lite`** | **`RESULT: PASS`** — `commit: 0c46095`, `dirty: no`, `tree-integrity: PASS`; `fmt` / `clippy` / `roborev-lints` / `scoped-tests` all PASS, **3812 tests passed, 170 test binaries** |
| **roborev** | **`RESULT: PASS`, `findings: NONE`** — `head-sha 0c460954c`, `reviewed-sha 25ffd5fd8..0c460954c`, `push-assert` / `census-check` / `code-free` / `sha-assert` / `prompt-content (3/3)` / both vacuity tiers PASS, 213199 input / 167040 cached tokens |
| the 7 new tests | **7 passed, 0 failed**, in the `cqlite-core` lib binary (3818 filtered out — they genuinely execute) |

**No authorization marker is owed on this PR:** roborev is clean, so there is no `roborev-defer` to
request, and `prompt-content` PASSed, so there is no absence waiver to request either.

### Disclosed rather than left in a log: this run carries a FILE-SIZE OPT-OUT

`--lite` reached PASS with `file-size: OPT-OUT`, not `file-size: PASS`. In full:

```
file-size:  OPT-OUT (0s) — CQLITE_ALLOW_FILE_GROWTH=1 (ratchet NOT enforced);
            1 over-threshold file(s) grown
      cqlite-core/src/.../row_decoder/mod.rs: 868 -> 873 (limit 800)
```

**Why the waiver rather than a fix:** `row_decoder/mod.rs` is **868 lines on `main`** — already 68
over the 800-line campsite threshold before this branch existed — and the ratchet fails on *any*
growth of an over-threshold file. A new test module must be declared somewhere, so the declaration
alone trips it. I first shrank the footprint from +8 to +5 by cutting the declaration comment from
five lines to two (the oracle, the ruling and the two-site table already live in the test module's
own header, so repeating them in `mod.rs` was redundant). Splitting an 868-line shared `mod.rs` is
genuinely out of scope for #3778, which is exactly the case CLAUDE.md's campsite rule covers — hence
the waiver, with the split tracked by epic **#1116** (source) / **#1135** (tests).

I deliberately did **not** manufacture offsetting deletions in `mod.rs` to squeeze under the limit;
deleting other people's documentation to fit my code would be gaming the ratchet, not respecting it.

### ⛔ Why there is still no gate of record, with the numbers

Two independent, unrelated-to-this-diff blockers:

1. **#4029** — this branch is **not immune**, and I confirmed it three ways rather than inferring:
   the guard file is present, `035fdbf32` **is** an ancestor, and running
   `bash scripts/tests/test_gate_component_verdict.sh` in this tree gives **`passed=124 failed=1`**
   with the sole failure being
   `FAIL 17.5 the shipped gate still has exactly TWO component-line emitters`. Exactly one
   assertion fails, so `tooling-tests` should come back clean once #4039 lands.
2. **The component-set pre-flight (#3544)** — this branch predates `3621a7bff` (PR #3839 / issue
   #1700), which added the `dep-duplicates` component, so a gate here would **FAIL-CLOSED at
   pre-flight in 0 seconds** without running a single component. Measured: the `COMPONENTS` array is
   **37** at this head and **38** on `origin/main`, and the committed manifest agrees 37 vs 38 —
   a delta of exactly one, `dep-duplicates`. (The fleet advisory quoted 38 vs 39; the mechanism and
   the delta are right, the absolute pair is one lower. Reported upstream.)

These two are mutually exclusive to escape by moving: staying put fails the pre-flight, rebasing
imports #4029. So this PR genuinely needs #4039 on `main` first — which is also why I have **not**
merged or rebased onto `main`, and why I have not touched `COMPONENTS` (editing it would smuggle a
gate change into this diff and certify nothing).

### Sequence when #4039 lands

Admission check (1-minute load < 30, per the #4030 advisory) → merge/rebase `main` **once** →
confirm `component-set` reads 38/38 → **one** full gate, backgrounded via
`scripts/flow/gate-detached.sh` with the summary-file redirect → `premerge-assert` requiring
`dirty: no` → re-read this thread for a `HOLD:` order → `gh pr merge --auto --squash --delete-branch`.

Still closes nothing: #3778 stays open until the lead confirms Option A is discharged. Related:
**#4038** (the parity follow-up this PR filed), #3811, #3631, #4029, #4032, #4030, #3544.

PR #3867, opened 2026-09-01
fix(#3642): pin the node-ci exemption's two lines and cover the manifest exit-code mapping

pmcfadin:
Closes #3642. **item 3 → #3866 (needs a JS-parser design call, Backlog)** — see "What is NOT in
this PR" below.

Scope confirmed by the lead on #3642 (`REQ-3642-1`, 2026-09-01): ship items 1/2/4, #3642 closes on
this merge, #3866 is the declared successor for item 3.

Residuals from #3493 (PR #3555), raised by that PR's C intent audit. **No `src/` change** — this is
tooling and tests only.

## Delivered — items 1, 2 and 4

**1. Pin the two lines the `node-ci.yml` exemption rests on.** `.github/ci-gating-tiers.yml` claims
`node-bindings` runs `npm run typecheck` and pairs `npm test` with `check-dataset-manifest.sh`. Both
are single lines in `run_node_bindings`; deleting either silently re-falsified the registry
sentence. Cases 93–96 pin them structurally (plus the ORDER of the last two), in the two-line
awk-scan idiom of the existing case 92.

**2. Cover the component's exit-code mapping behaviourally** (cases 97–102): `9` + opt-out → SKIP,
`9` without → FAIL, any other non-zero → FAIL-as-tooling. The suite already covered
`check-dataset-manifest.sh`'s own exit codes; `agent-gate.sh`'s *interpretation* of them was
verified by hand once during #3493 and asserted nowhere.

**4. Reconcile the 138s/133s figure** in `CLAUDE.md`: 138s is PR #3555's post-rebase (merged) build,
133s the same measurement pre-rebase. Two real measurements of two builds, each cited with the build
it describes. No new number invented.

The registry entry gains only a provenance sentence naming the cases that keep its own claim true,
so the guard and the sentence it protects point at each other.

## One defect, found three times, finally fixed as a class

Rounds 1–3 each surfaced the same defect in a different case:

| round | case | symptom |
|---|---|---|
| 1 | 101 | asserted a message emitted *before* `npm test` — passed when the invocation never happened |
| 2 | 100 | never asserted `npm test` was *not* invoked |
| 3 | 98–99 | matched an early `SKIP`, ignored a later `FAIL`, no npm-log check |

One defect: *an assertion keyed on an early or partial marker instead of the component's recorded
verdict, uncorroborated by the npm invocation log.* Patching the named case each round is how it
[...]

Issue #3642, opened 2026-08-30
#3493 residuals: pin typecheck + manifest invocation structurally, and two small coverage/consistency gaps


pmcfadin:
Non-blocking residuals from #3493 (PR #3555), raised by that PR's C intent audit. Filed rather than committed **because committing would move the PR head and void a scope-bound roborev waiver**, forcing another full review round for changes the audit itself classified as non-blocking. Recording the reason so the deferral is a decision, not an oversight.

## 1. Nothing pins the two lines that make the node-ci exemption true

`.github/ci-gating-tiers.yml` now claims `node-bindings` runs `npm run typecheck` and pairs `npm test` with `check-dataset-manifest.sh`. Both are single lines in `run_node_bindings`. Deleting either silently re-falsifies the registry sentence — **which is precisely #3493's defect class, reintroduced one level down.**

`scripts/tests/test_check_dataset_manifest.sh` case 92 already demonstrates the shape: a two-line awk scan over `run_node_bindings` asserting a required line is present. Apply the same to `npm run typecheck` and the `check-dataset-manifest.sh` invocation.

## 2. The component's exit-code interpretation is unexercised

The self-test covers `check-dataset-manifest.sh`'s own exit codes thoroughly (9 = corpus verdict, 2 = tooling malfunction). What is not covered behaviourally is the COMPONENT's mapping of them: `9` + opt-out → SKIP, `9` without → FAIL, any other non-zero → FAIL-as-tooling (`scripts/agent-gate.sh`). Verified by hand once during #3493; no assertion holds it.

## 3. A derived check was replaced by a hand-maintained list

Pre-rebase, the schemas check derived "every schema `setup.js` reads is in the certified set". That derivation did not survive; the replacement in `test_agent_gate_schemas_preflight.sh` is a hand-named complete set plus a `-ne 8` count assert. Deliberate — deriving BOTH sides is the tautology that made the earlier version unable to fail — but it is a real coverage reduction: a NINTH schema added to `setup.js` without touching `CANONICAL_SCHEMA_FILES` is no longer detected. A one-directional derivation (read setup.js, require ⊆ the hand-named set) would restore it without the tautology.

## 4. Two numbers for one measurement

`CLAUDE.md` says the widened `node-bindings` component measures **138s**; PR #3555's body says **133s**. Both were real measurements of different builds (pre- and post-rebase). Pick one, or say which build each describes.

pmcfadin:
[...]

PR #3995, opened 2026-09-03
fix(#3977): main is RED — capture the gate's component-status token by SHAPE, not by a closed three-token enumeration

### LINKED ISSUE #3977: main is RED: tooling-tests FAILs on clean origin/main (census guard vs dataset-manifest alternations) — blocks every lane's gate

pmcfadin:
Closes #3977

`main` was RED: `scripts/tests/test_agent_gate_census.sh` FAILed on a clean `origin/main`
(`118 passed, 1 failed`), so `tooling-tests` FAILed and no lane could obtain a passing gate of record.

## The judgement the issue asked for

The issue offered two readings and leaned toward the second — that a dataset-manifest status vocabulary
need not be the gate's, in which case the guard's *subject set* wants narrowing. **Measured, the first
reading is correct: the guard applies, and the manifest test is the side that changes.**

The four sites parse the gate's *own* component verdict line:

```
scripts/agent-gate.sh:12378   echo ">>> [$name] $RECORDED_STATUS ($((end - start))s)"
```

That token is an **open variable**, and the vocabulary is already **five** tokens wide, not four:
`_status_is_nonfailing` (`agent-gate.sh:7950`) is the closed non-failing set `PASS|SKIP|OPT-OUT`, plus
`FAIL` and `VACUOUS`. So these lines speak the gate's component-status vocabulary by construction.

## Why not simply add `VACUOUS`

That is the remedy the failure message offers first, and it fixes one token while staying blind to
`OPT-OUT`, which already ships, and to whatever joins next. **The blindness is the enumeration itself.**

So the log-side capture is now by SHAPE — which is exactly what the sibling SUMMARY reader four lines
below already did:

```sh
# before
grep -Eo '^>>> \[node-bindings\] (PASS|FAIL|SKIP) \(' "$log" | awk '{print $3}'
# after
sed -n 's/^>>> \[node-bindings\] \([A-Za-z][A-Za-z-]*\) (.*)$/\1/p' "$log"
```

The two halves of `_nbgate_measure` now agree, and any token the gate records is SEEN and screened
against each case's allowed set.

**This is load-bearing, not cosmetic.** `_nbgate_assert`'s contract is that *every* recorded verdict is
in the allowed set — so a token the capture drops is a **disagreeing verdict the assert cannot see**,
which is precisely the R1 harm, occurring inside the helper written to stop that exact family (see the
round-1/2/3 comment block above it).

## Membership is not detection — cases 103/104

Removing the literal proves the guard stops firing, not that the new capture catches the incident's own
construct. So the diff carries a control that plants out-of-vocabulary tokens (`VACUOUS`, `OPT-OUT`) and
requires the assert to **REFUSE and to NAME them**; 104 is the positive control (same path, token in the
allowed set, must PASS). Both drive the real helpers against a synthetic log/summary/argv triple rather
than a nested gate run, so they cost ~0s.

The control earned its keep during implementation: the first candidate shape ended at ` (` and captured
`SUBJECTS (all DERIVED …):` — an all-caps prose word is indistinguishable from a status by token shape
alone. Case 104 caught it. The pattern therefore requires the paren to **close at end of line**,
verified against every `>>> [$name]` emit site so no real verdict line is lost.

Discrimination check: reverting **only** the capture (scratch copy) reds 103 with
`assert verdict: admitted; recorded: 'PASS ', n=1` — the planted tokens dropped — and reds 104 too.
Restored, both pass.

## AC3 — the combination: cause identified, residual DECLARED with its owner

Both commits landed 2026-09-02, the guard **first** and an ancestor of the other:

| | commit | landed | PR / issue |
|---|---|---|---|
| guard | `bdaf2b6e1` | 09:27 -07 | #3916 / #3625 |
| sites | `694df0d90` | 12:38 -07 | #3867 / #3642 |

Ancestry alone does **not** establish that #3867's branch predated the guard — that holds for any
squash-merge landing after it. The sound evidence is that **#3867's own diff is what added all four
literal sites**, so had its branch carried the guard, its own `tooling-tests` would have RED on its own
diff. It merged green ⇒ its base predated the guard. Ordinary base staleness, **not** an authoring error
on either side: each side passes alone, and only the pair reds.

The mechanism that recognises this already exists and is verified end to end: `bdaf2b6e1` touches
`scripts/tests/test_agent_gate_census.sh`; `scripts/tests/**` is in `base-staleness.sh`'s
`GATE_GLOBAL_PATTERNS` (`:307`); the matcher (`:713`) strips `/**` and prefix-matches. So #3650 **slice 1**'s
advisory would have named that commit as staling #3867's base — and it is **advisory-only by
construction and cannot fail a run**.

What would actually have *blocked* it is a gate on the **merge result**, which is **#3650 slice 2** and
is explicitly not implemented. That is stated at the site as a declared residual rather than met with a
second half-mechanism here.

## Review

`rust-reviewer`: **no blockers**, 7 nits. Four were fixed in-branch because two were more than nits:

- **Cases 103/104 sat behind a git/clone guard they do not need.** Without git, the *only* detection
  control for this fix vanished while the suite still printed `failed: 0` — silent coverage omission, in
  the control that certifies the change. They are now **hoisted** above the `$nbgate_ok` guard (they need
  only `$WORK` and the triple they write themselves). Verified under a git-less PATH farm: post-fix
  `103` and `104` both PASS; the same farm at `HEAD~1` shows them **absent entirely**.
- **The AC3 comment's ancestry inference was a non-sequitur** — corrected as above, with ancestry demoted
  to what it actually evidences.
- A count scoped to `run_node_bindings` (**12**, measured; 116 emit sites script-wide) rather than "~11".
- Case 103's "name the token" half now counts two explicit per-token `RECORDED verdict '<TOK>'` clauses
  instead of leaning incidentally on the `recorded sequence:` field.

Two nits are **deliberately deferred** to a linked follow-up rather than folded into a P0 unblock:
deriving the status vocabulary at run time from `_status_is_nonfailing` (the "derive, never curate" end
state — the under/over-capture residual stays *declared*), and declaring the `tail -40` stdout
interleaving at the parse site (pre-existing, in `agent-gate.sh`).

roborev: round 1 `RESULT: PASS`, `findings: NONE` (`input=167036 cached=130560` — a genuine review, well
clear of the ~18.7k vacuous baseline).

## Acceptance criteria

1. ✅ `test_agent_gate_census.sh` passes on a clean worktree — **119 passed, 0 failed**.
2. ✅ The chosen reading is stated at the site, with why an enumeration cannot be the fix.
3. ✅ Cause identified; the recognising mechanism verified; the blocking half declared as #3650 slice 2.

`test_check_dataset_manifest.sh`: **160 passed, 0 failed** (63s; was 159, +2 cases).
No `.rs` files touched, so the `file-size` ratchet has no subject here.

### DIFF
--- a/scripts/tests/test_check_dataset_manifest.sh
+++ b/scripts/tests/test_check_dataset_manifest.sh
@@ -2594,16 +2594,26 @@ JESTSTUB
     # this function AGREE -- the SUMMARY reader four lines below has always read
     # `[A-Za-z-]*`, and only the log side had a list.
     #
-    # ` (' -- the duration/reason paren -- is the DISCRIMINATOR, and it is the whole
-    # reason a one-word token is safe: it keeps the capture off this component's
-    # continuation prose (`>>> [node-bindings]   NOT VALIDATED by this run: ...`) and off
-    # its coverage census. It deliberately does NOT require `([0-9]+s)`: the BRANCH
-    # ANNOUNCEMENTS (`SKIP (AGENT_GATE_ALLOW_MISSING_FIXTURES=1 and ...)`) carry no
-    # duration, and those are exactly the lines property (a) needs to see -- the
-    # announcement and the terminal line share a shape, so a lost `return` is only ever
-    # visible as a SECOND, DISAGREEING verdict. Cases 103/104 are the control: they plant
-    # VACUOUS and OPT-OUT (both DROPPED by the enumeration this replaced) and require the
-    # assert to SEE them.
+    # THE DISCRIMINATOR IS THE WHOLE `<TOKEN> (<detail>)` SHAPE, token to end of line, and
+    # it is what makes a one-word token safe: agent-gate.sh emits a verdict as exactly
+    # that, while this component's continuation prose (`>>> [node-bindings]   NOT VALIDATED
+    # by this run: ...`) and its coverage census do not close a paren at end of line.
+    # ` (` ALONE IS NOT ENOUGH, and case 104 is what measured it: an ALL-CAPS census
+    # heading of the form `SUBJECTS (all DERIVED from cargo …):` is indistinguishable from
+    # a status BY TOKEN SHAPE -- a status is all-caps too -- so the trailing `)$` is
+    # load-bearing, not decoration. (That heading is another component's today; the shape
+    # is the exposure, not the current caller set.)
+    #
+    # It deliberately does NOT require `([0-9]+s)`: the BRANCH ANNOUNCEMENTS
+    # (`SKIP (AGENT_GATE_ALLOW_MISSING_FIXTURES=1 and ...)`) carry no duration, and those
+    # are exactly the lines property (a) needs to see -- the announcement and the terminal
+    # line share a shape, so a lost `return` is only ever visible as a SECOND, DISAGREEING
+    # verdict. Nothing here reads a PRESENTATION property (#3400): these are gate-authored
+    # `echo`s, not cargo output, and no colour or quiet setting reaches them.
+    #
+    # Cases 103/104 are the control: they plant VACUOUS and OPT-OUT (both DROPPED by the
+    # enumeration this replaced) and require the assert to SEE them, alongside prose the
+    # capture must not take.
     #
     # HOW THE PAIR REACHED `main`, declared so the next person does not re-derive it: the
     # R1 guard (bdaf2b6e1, #3625/#3916, 2026-09-02 09:27 -07) is an ANCESTOR of the commit
@@ -2615,7 +2625,7 @@ JESTSTUB
     # and cannot fail a run. What would have BLOCKED it is a gate on the MERGE RESULT,
     # which is #3650 SLICE 2 and is explicitly not implemented; that issue owns this
     # residual. No second guard is added here for the combination.
-    toks=$(sed -n 's/^>>> \[node-bindings\] \([A-Za-z][A-Za-z-]*\) (.*/\1/p' "$log" 2>/dev/null)
+    toks=$(sed -n 's/^>>> \[node-bindings\] \([A-Za-z][A-Za-z-]*\) (.*)$/\1/p' "$log" 2>/dev/null)
     NBG_LOG_TOKS=$(printf '%s\n' "$toks" | grep -E '.' | tr '\n' ' ')
     NBG_LOG_N=$(printf '%s\n' "$toks" | grep -cE '.')
     NBG_LOG_LAST=$(printf '%s\n' "$toks" | grep -E '.' | tail -1)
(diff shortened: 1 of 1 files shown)


pmcfadin:
Nits from the `rust-reviewer` pass that are **not** fixed here are batched into #3996 (filed at Backlog per #3893 — neither has caused a false PASS, blocked a lane > 1h, nor recurred twice): deriving the status vocabulary at run time from `_status_is_nonfailing` rather than curating a character class, and declaring the `tail -40` stdout-interleaving exposure at the parse site.

roborev round 2 on `e03ca33`: `RESULT: PASS`, `findings: NONE` (job 97, `input=195214 cached=154240`).

PR #3832, opened 2026-09-01
docs(#3654): roborev job ids are per-daemon — document the scope and the verification that settles it

pmcfadin:
## Summary

roborev job ids are **per-daemon, not global**. Each fleet box runs its own daemon with its own database and its own sequential ids, so two boxes can legitimately present the same `job=<id>` for two different reviews. That is exactly what happened: PR #3651 (`lane-3515`) and PR #3555 (`lane-3493`) both presented **`job=265`**, 50 minutes apart, for different `git_ref`s. Reading the repeat as a collision, the coordination lead **withheld a valid absence waiver**. Nothing in CLAUDE.md, in `--help`, or in the wrapper said the id was daemon-scoped.

This PR documents the scope and the verification procedure that settles such a question. **A lead who had read it would not have withheld the waiver.**

Closes #3654.

## Scope — asks 1+2 only, by lead ruling

#3654 carried four asks. Five review rounds produced nine findings, and **both recurring sites lay outside asks 1+2**. Per the lead's ruling on `req-3654-01` (option D, CLAUDE.md's #3544 *split-don't-carve* precedent), this PR ships the documentation and the rest was split out:

- **#3825** — the `job-machine:` summary-block key, the cross-box marker residual, and the marker-grammar design question. It names this branch's implementation as its starting point.
- **#3826** — waiver evidence policy: what evidence may support an absence waiver.

## What ships

| file | what |
|---|---|
| `scripts/flow/roborev-review.sh` | `--help`: daemon-scoped ids, a verification procedure that works, the row-count non-check; corrected stale comments |
| `CLAUDE.md` | the same doctrine in the waiver paragraph |
| `website/.../agents-developing/roborev-findings.md` | the published surface |
| `scripts/tests/test_roborev_review_guard.sh` | guard coverage for the above |
| `scripts/flow/roborev-job-facts.py` | comment correction only (no behaviour change) |

## Measured, not argued — and three of these contradict the issue as filed

Everything below was measured on roborev **v0.61.2**. They are recorded because the issue's own asks, implemented literally, would have shipped a procedure that misleads the operator it exists to help.

1. **The verification command the issue prescribes returns three nulls.** `roborev show <id> --json | jq '{id, git_ref, branch, source_machine_id, token_usage}'` — `git_ref` and `token_usage` are nested under `.job`, and `source_machine_id` is absent from that payload entirely (it is carried by `roborev list --json` rows only). The documented procedure names the payload that actually holds each field.

[...]

Issue #3810, opened 2026-09-01
Nothing in the worker/drive-issue path passes --actor, so every lane on a box claims as actor=flow and reads a peer's claim as its own (3 collisions in one session)


pmcfadin:
Filed at the suggestion of `lane-3654-2d`, which diagnosed the root cause; the instance data is mine. **This is a correction to my own analysis on #3436** — I concluded there that only an atomic machine-local lock could close this class. That was wrong, and the reason matters.

## Mechanism, verified from source

`scripts/flow/claim.sh`:

```
#   actor    --actor <id>, else CLAIM_ACTOR, else "flow" — a sub-machine role.
#   CLAIM_ACTOR    default actor when --actor is omitted (default: flow)
#   verify <N> [--actor <id>]   exit 0 iff we hold it (this machine+actor)
```

Holder identity is **machine + actor**. The actor defaults to the literal string `flow`. So on a box running N lanes as one user, every lane that does not pass `--actor` has an **identical holder identity**, and `claim.sh claim` on an already-held ref reports `CLAIM: HELD` **idempotently** — the arriving session is told it holds a claim another session created.

CLAUDE.md documents this hazard under #2665 (a supervisor exports a lane-unique actor; a manual `claim.sh` invocation on a multi-lane box must pass `--actor` itself). **Nothing in the `worker` / `drive-issue` path asks for it**, so an interactive or `/drive-issue`-driven lane silently inherits `flow`.

## Why this is fixable, and why that corrects #3436

The ref itself is a genuine lock: two sessions pushing a unique root commit to `refs/claims/issue-<N>` are arbitrated by **git**, server-side — one wins, the other is rejected. That arbitration never gets a chance to run here, because the second session never attempts a push: a matching machine+actor makes it conclude it already holds the claim.

**With lane-unique actors, the existing protocol closes this without any new mechanism.** Session B would see the ref held by `flow-lane3654a` ≠ its own `flow-lane3654b`, report `CLAIM LOST`, and stop. So my #3436 comment ("only an atomic machine-local acquisition closes it") **overstated the problem** — I inferred unfixability from three collisions without checking whether the identity was distinct in any of them. It wasn't, in all three.

## Instances — three in one session, 2026-09-01, box `ip-172-31-1-216`

| # | issue | what I did | outcome |
|---|---|---|---|
[...]

Issue #3654, opened 2026-08-30
roborev job ids are PER-DAEMON, not global: two boxes presented the same job=265 and a valid absence waiver was withheld


pmcfadin:
## Summary

The absence waiver marker binds `base`, `head` and `job`:

```
roborev-waive: prompt-content-absent base=<40-hex> head=<40-hex> job=<id> reason=<why>
```

**`job` is not globally unique. It is per-daemon, and each fleet box runs its own daemon with its own database and its own sequential ids.** Nothing in CLAUDE.md, `--help`, or the wrapper says so, and the natural reading of "job=265" is that it names one review.

## How it surfaced

Two lanes requested absence waivers 50 minutes apart, both naming **`job=265`**:

| | PR #3651 (`lane-3515`, box2) | PR #3555 (`lane-3493`, box3) |
|---|---|---|
| `git_ref` | `da80a479e..a06ac0eca` | `d6fe6d39b..6914568e` |
| branch | `issue-3515-cli-tests-sigint-flake` | `issue-3493-node-export-list-red` |
| tokens in/cached/out | 1,093,411 / 976,640 / 9,462 | 2,803,565 / 2,652,672 / 12,384 |
| verdict | `F` | `P` |

As coordination lead I read the repeated id as a collision, assumed one lane had copied the wrong number, and **withheld a valid waiver** pending proof. Both lanes then queried their own daemon, both returned an internally consistent row, and both correctly verified their box held exactly **one** row with that id (`[.[] | select(.id==265)] | length` → `1`). Both were right. The premise was wrong.

Cost: one review round on each of two lanes, plus a withheld authorization on a PR that had already satisfied every other key.

## Why this is a real defect and not just my mistake

**The failure mode is symmetric, and the other direction is worse.** I refused a valid waiver, which is fail-closed and cheap. But a lead who *knows* ids repeat across boxes and treats the `job` field as therefore uninformative would be discarding the one field that binds the authorization to a specific *review* rather than to a *range* — and CLAUDE.md is explicit that dropping `job=` reopens the hole where one comment waives a later **vacuous** review at the same base+head.

So the field matters and its scope is undocumented. That is the gap.

**The waiver mechanism itself is safe**, and this issue should not change it:

- `base` and `head` are git SHAs — globally unique.
- `--recheck-job <id>` runs against the **local** daemon, so `sha-assert` re-compares *that* record's `git_ref` against *this* base and head.
[...]

PR #4012, opened 2026-09-03
chore(#3654): delivery telemetry record for PR #3832

pmcfadin:
One-line append to the delivery ledger for #3654 / PR #3832 (merged as `d659de8fc`).

Counters are run-observed per the authoritative-data-only rule, and two were corrected against the schema's own definitions rather than guessed:

- `claim_collisions: 0` — the field is *rejected claim pushes (a competing session won the race)*. This lane's claim push was never rejected. A peer session did collide on #3654 and stood down voluntarily, because `claim.sh` answered `VERIFY-OK` to **both** lanes (same machine, same default `CLAIM_ACTOR`) — that is #3810's defect, not a lost race, so recording 1 would have logged a race that never happened.
- `gate_runs: 2` — the field is *runs through (and including) the first PASS*, and its consumer derives failed rounds as `gate_runs - 1`. Observed: one FAIL then one PASS. A **third** full gate ran afterwards because later commits staled the second one's certification; it is outside what this field asks, so it is disclosed here rather than folded in, which would have implied two failed rounds when there was one.

`gate: pass`, `rework: 9` (roborev re-review rounds beyond the initial), `rebase_events: 5` (all clean, no conflict resolutions), `roborev_findings: 14` = `blockers 14` + `nits 0` — every finding was fixed pre-merge; none was deferred or batched.

Routing `oracle`: process/tooling, no OpenSpec change, so nothing to archive.


pmcfadin:

Issue #4009, opened 2026-09-03
3551 follow-up: clean-pairs.py accepts a FAILED session and pools across sets without validating configuration


pmcfadin:
Found by roborev round 4 on **#3551** / PR #3985 (job 432). **Disposed rather than fixed** under #3893's two-round cap for `scripts/**`; #3551 had already spent an exemption on three *false-verdict* findings in round 3, and rounds 3 and 4 both landed in this one file, which is the fix-round spiral that cap exists to stop.

Subject: `docs/reports/ws0-3551-artifacts/clean-pairs.py` — the tool that produces #3551's primary published estimate (pooled clean within-round pairs).

## The two gaps

**1. The recorded `exit` status is never checked** (`clean-pairs.py:86`). `session()` reads `abc-window.json` and stores `exit`, then never validates it, so a session whose window record says the run FAILED can still be classified clean and enter the published medians provided a `results.json` remains beside it. Remedy: refuse or exclude any session unless `exit` is present, is an `int` (not `bool`), and equals 0.

**2. Pairs are pooled across sets by arm label and round alone** (`clean-pairs.py:211`). The tool never validates recorded treatment fields, corpus identity, binary digests, pins or admission settings, so identically-named arms from materially different experiments could be combined into one median. Remedy: read and validate each session's recorded configuration before pairing — stable treatment fields per arm, identical cross-session invariants — as `ws0_abc_aggregate.py` already does *within* a set.

Gap 2 is the more serious of the two: `ws0_abc_aggregate.py` enforces exactly these properties within a single set, and the pooling tool was written without carrying that check across sets. The asymmetry is the defect.

## Neither affected #3551's published figures, and that was VERIFIED rather than argued

* **Gap 1**: all **42** sessions across the three sets record `exit` as integer `0`. Additionally the driver aborts on a failed session and `ws0-baseline.sh` writes no `results.json` on failure, so the (failed + results.json) combination is not reachable through the driver — but the tool accepts any directory, which is why it is still a gap.
[...]

PR #4027, opened 2026-09-03
chore(#3551): delivery telemetry record for PR #3985

pmcfadin:
One append-only ledger line for the #3551 delivery cycle (PR #3985, merged as `86fae8c70`).

`gate: pass`, `gate_runs: 4` — one voided by a mid-run tree mutation (mine), one a real red that turned out to be a known main-red already fixed upstream by PR #3995, one PASS invalidated by later fixes, and one the gate of record at `5a4451646`.

`roborev_findings: 10` across four rounds: **8 fixed pre-merge**, **2 batched into linked follow-up #4009** at merge time under #3893's two-round cap for `scripts/**`. They are recorded in the `nits` bucket because that field records **disposition** rather than severity label, and the linked-follow-up path is the nit disposition — both carried Medium severity labels, and both were verified to affect no published figure before being deferred.

The ledger is a hot append-only file: on a rebase conflict **keep ALL lines** — main's plus this one — never drop a peer's record.


pmcfadin:

Issue #3822, opened 2026-09-01
drive-issue state marker has no ownership stamp: a rehydrating session can adopt a peer's plan (.drive-issue-state.md)


pmcfadin:
Found on #3436 (round 13) and routed to the lead for disposition; the worker holding PR #3682 correctly declined to absorb it into a frozen, certified branch.

## Defect

`.drive-issue-state.md` — the durable per-lane state marker the drive-issue/worker path rehydrates from — carries **no ownership stamp** (no machine, no session, no actor). A session rehydrating in a shared or reused worktree can adopt a PEER's plan wholesale: it reads a state file some other session wrote and continues that session's work as its own.

This is the state-file half of the identity gap whose claim-ref half is #3810 (every lane claims as `actor=flow`, 3 collisions in one session). #3810 fixes the claim actor; THIS issue fixes the marker: stamp machine + session + actor at write time, and refuse (or at minimum loudly warn) at rehydrate time when the stamp does not match the reader.

## Acceptance criteria

1. Every write of `.drive-issue-state.md` records machine, session id, and actor.
2. Rehydration verifies the stamp against the current machine/session and fails closed (or requires an explicit adopt gesture) on a mismatch.
3. A test demonstrates the mismatch path RED before the fix shape and named-refusal after.

Refs: #3436 (origin, PR #3682), #3810 (claim-actor sibling).

pmcfadin:
## drive-issue worker claimed #3822

**Comms**: the `github-coord-worker` skill is **not installed on this box** (`.claude/skills/` has no
`github-coord*` entry, and no plugin provides it) — so I am following the coord protocol as specified in
`/drive-issue`'s Delta 2 directly: one request marker per question, `coord:needs-attention` /
`coord:blocked` / `coord:follow-up-proposed` as the only tracker metadata I touch, the lead clears
labels, never me. Flagging the missing skill rather than silently substituting.

**Lane**: `refs/claims/issue-3822` held (machine `ip-172-31-5-53`, `actor=flow` — which is #3810's
subject, noted below), worktree `/data/lanes/lane-3822`, branch
`issue-3822-drive-issue-state-ownership-stamp`.

### Routing: oracle-driven (no OpenSpec change), no Seam 1

The issue carries three explicit, testable acceptance criteria and is labeled `bug`; AC3 already names
its own oracle (mismatch path RED before, named refusal after). Per CLAUDE.md, oracle-driven fixes stay
[...]

Issue #3969, opened 2026-09-02
tooling-tests: a broken pipe under gate load turns the NEXT assertion into a bogus FAIL — second site test_agent_gate_summary.sh (268 unguarded printf) — CITATION for a broken-pipe red naming that file (#4032 covers test_drive_issue_state.sh only)


pmcfadin:
**Class:** gate flake / false FAIL. **Blocks gates of record.** Two measured instances today, in two different suites, same mechanism.

## The signature

A `printf` into a closed pipe, then a FAIL on the **immediately following line**:

**Instance 1** — `scripts/tests/test_agent_gate_summary.sh`
```
test_agent_gate_summary.sh: line 4617: printf: write error: Broken pipe
FAIL - 1465-skip-declares: the opt-out SKIP branch does not declare the leak-lane state
```
Gate reported `passed: 419  failed: 1`.

**Instance 2** — `scripts/tests/test_drive_issue_state.sh`
```
test_drive_issue_state.sh: line 550: printf: write error: Broken pipe
FAIL - --help does not document the contract: rc=0
```

## Both are FALSE — measured

| suite | in the gate | standalone re-run |
|---|---|---|
| `test_agent_gate_summary.sh` | 419 passed / **1 failed** | **420 passed / 0 failed** (and 420/0 on a clean `origin/main` worktree too) |
| `test_drive_issue_state.sh` | FAIL | **`PASS=181 FAIL=0`**, exit 0 |

Same tree, same commit, no code change between the gate run and the re-run. Instance 1 was additionally checked on a detached `origin/main` worktree and passed there, so it is not branch-specific.

## Why this is worth fixing rather than retrying

`tooling-tests` is the gate's **longest** component (2206s measured on the failing run). A false FAIL there costs a full gate of record — ~35 minutes for that component alone, and the whole run's verdict. On this issue it cost **two** gates of record: one at `a50add16c` and one at `0846f1cc7`, both otherwise 37/37 with the flake as the sole failure.

It is also the worst possible shape for a flake: the FAIL text names a **real assertion about real doctrine** (`--help does not document the contract`, `the opt-out SKIP branch does not declare the leak-lane state`), so it reads exactly like a genuine defect. An agent that trusts it will go and "fix" correct code. I nearly did on the first instance — the thing that stopped me was that my diff touched zero files under `scripts/`.

## Likely mechanism

[...]

PR #4045, opened 2026-09-04
fix(#4032): a piped printf's OWN stderr must be suppressed, or SIGPIPE leaks an unanchored line

### LINKED ISSUE #4032: drive-issue-state.sh: a piped printf leaks an UNPREFIXED 'write error: Broken pipe' — breaks the anchor contract and reds tooling-tests intermittently
Found by #3725's gate of record (2 of 2 full gates reddened by it). **Not caused by #3725** —
`scripts/flow/drive-issue-state.sh` and `scripts/tests/test_drive_issue_state.sh` are both
byte-identical to `origin/main` on that branch.

## Symptom

`tooling-tests` FAILs on one case of `test_drive_issue_state.sh` (PASS=180 FAIL=1 in-gate;
**PASS=181 FAIL=0 standalone**):

```
FAIL - a failing tr leaked a line, or committed a placeholder identity instead of refusing:
/data/lanes/.../scripts/flow/drive-issue-state.sh: line 587: printf: write error: Broken pipe
DRIVE-STATE: verdict ERROR
DRIVE-STATE: verdict-detail axis=machine is NOT RECORDABLE: ...
```

Note what is CORRECT here: the script **did** refuse, with an anchored `ERROR` naming
`axis=machine`. The case fails only because an **unprefixed native line leaked alongside it**,
which is precisely the contract that suite enforces (#3822 correction (2): every external
command's diagnostic must be captured into the anchored message or suppressed).

## Cause — the #3685 SIGPIPE class

`sanitize_field` (line 587):

```bash
s="$(printf '%s' "${1:-}" | LC_ALL=C tr -c 'A-Za-z0-9._:/#-' '-' 2>/dev/null | LC_ALL=C sed ... 2>/dev/null)"
```

`tr` and `sed` each carry `2>/dev/null`; **the `printf` does not.** When the reader exits early
— which the test forces by shimming a FAILING `tr` — `printf` gets EPIPE and **bash** prints
`printf: write error: Broken pipe` on stderr, unanchored and unsuppressed.

It is timing-dependent: whether printf's buffer is flushed before the reader goes away decides
whether EPIPE is raised at all. That is why it passes standalone and reds under gate load, and
why it is an intermittent that will hit **any** lane's full gate, not just #3725's.

## The class is 4 sites, not 1

Every piped `printf` in the file lacks its own suppressor; the suite only catches 587 because
that is where its shim targets `tr`:

| line | site |
|------|------|
| 362  | `printf '%s' "${1:-}" \| LC_ALL=C tr -c '\040-\176\200-\377' '?' 2>/dev/null` |
| 587  | `sanitize_field` — the one that reds today |
| 1426 | `printf '%s\n' "$pro" \| count_matching_lines "^$k: ."` |
| 1804 | `printf '%s' "$tok" \| LC_ALL=C tr 'A-Z' 'a-z' 2>/dev/null` |

## Fix

Add the printf's own `2>/dev/null` at each of the four sites (the redirection must be on the
`printf` itself — bash applies redirections left to right, which is #3822 correction (6)'s own
lesson one command over). Then extend the shim matrix so a failing reader is exercised against
**each** piped-printf site, not just `sanitize_field`'s `tr`, or the remaining three stay latent
and the guard reports a clean sweep it has not taken.

Worth noting the suite's NON-VACUITY line already says *"7 distinct external commands were
actually shimmed and reached"* — the shim set is real; it simply does not cover the writer side
of every pipe.

Related: #3685 (the SIGPIPE class), #3822 (this script and its anchor contract).

pmcfadin:
Closes #4032.

A piped `printf` whose reader has gone away makes **bash** — not the printf, not the reader — write `<script>: line N: printf: write error: Broken pipe` to stderr with no `DRIVE-STATE: ` prefix, breaking `drive-issue-state.sh`'s anchor contract and reddening `tooling-tests` intermittently in **any** lane. The `2>/dev/null` the reader already carries cannot suppress it: bash applies redirections left to right, so a suppressor after the `|` covers the reader's stderr and nothing of the writer's (#3822 correction (6)'s lesson one command over).

## The mechanism in the issue body is incomplete, and correcting it made the guard deterministic

#4032 attributes the leak to timing alone — *"whether printf's buffer is flushed before the reader goes away decides whether EPIPE is raised at all"* — and reads the standalone/in-gate split as load. Measured, it needs **two** conditions:

1. **SIGPIPE disposition is `SIG_IGN`, inherited.** Under the **default** disposition the writing subshell is killed by SIGPIPE and prints **nothing at all**. `SIG_IGN` survives `exec` (measured: `0x1000` propagates into the script *and* its pipeline subshells), so any harness ignoring SIGPIPE propagates it inward. This is the condition the body misses, and it — not load — is what explains "PASS=181 standalone, PASS=180 FAIL=1 in-gate".
2. **The read end is already closed when `printf` writes**, so `write()` returns `EPIPE` regardless of pipe buffer space. Reachable either by the reader exiting first (a scheduling race — the load-sensitive half the body did see) or by a payload exceeding pipe capacity.

Matrix, 20 trials per cell, reader = a shim exiting 1 without reading:

| SIGPIPE | payload | leaked |
|---|---|---|
| default | 120 B | 0/20 |
| default | 256 KB | 0/20 |
| ignored | 120 B | 0/20 |
| ignored | 256 KB | **20/20** |
| ignored | 10 B, reader provably gone first | **20/20** |

The payoff is that the new guard **forces both conditions explicitly and is never timed** — no sleep, no load dependency, no retry.

## The census is 5 sites, not 4

The issue's table lists four, all in `drive-issue-state.sh`, because its census was scoped to one file. The fifth is **`scripts/flow/lib/process-liveness.sh:76`**, which is `source`d *into* this script's process (line 569) — so its leak lands on the same stderr and breaks the same anchor — and is shared with `claim-heartbeat.sh`, whose anchor it breaks too. CLAUDE.md's #3822 clause (12): *an extraction's deduplication is not complete until the guards around the new dependency are deduplicated too.*

Sites fixed (the redirection is on the `printf` itself, before the `|`): `drive-issue-state.sh` 362 / 587 / 1426 / 1804, and `lib/process-liveness.sh` 76.

Suppression rather than capture is consistent with the established contract at these sites: `tr`/`sed` already carry `2>/dev/null` under the recorded rationale that *"a diagnostic from the tool that exists to PROTECT the anchor would be the one line breaking it"* (roborev job 26 F2), and the failure is already reported through the anchored refuse/`WRITE_ERR` path.

## The guard (case 48)

- **Site set DERIVED from committed source**, never curated — per `printf` token, truncate at the first statement separator and ask whether a `|` remains, which is what stops a `case` pattern's `|` and `||` counting (16 such lines across the two files). A failed derivation is a **FAIL naming the derivation**, never a fallback to "no sites found", which would excuse every site at once.
- **Per-site behavioural probes, each with a POSITIVE CONTROL**: the defect is planted back into a throwaway scratch copy (never the shipped file, never a settable seam) and the probe must leak **and name the site**. The plant is asserted to have actually changed something, so a no-op plant cannot make a control silently prove nothing.
- **Floors**: a site floor (5) and a probe/control floor, because a green tally over a shrunken subject set is this case's own subject (#3544).
- **One owner**: the sweep covers both files from `test_drive_issue_state.sh` only — a copy in `test_claim_heartbeat.sh` would be two places for one rule to drift.

**Declared gap, 1 RECOGNISED**: site 1804's payload is `sanitize_field`'s output, structurally capped at 120 bytes, so the oversize-payload route cannot reach it; the remaining route is a genuine load race that is not deterministically forceable. It is fixed and covered by the structural sweep, and is **not** claimed as behaviourally probed.

## Verification

- `test_drive_issue_state.sh` standalone: **PASS=195 FAIL=0**, 48 cases (was 47).
- **Non-vacuity verified independently by the lead**, not just by the author: planting the defect back at line 587 in the shipped file yields **PASS=190 FAIL=5, rc=1**, and the failures **name `drive-issue-state.sh:587`** in the structural sweep, the behavioural probe, the control and both floors. Tree restored clean afterwards.
- `test_claim_heartbeat.sh` (shares the touched lib): PASS.
- Two measurement failures were caught and are recorded rather than published as results: a 256 KB payload passed through `argv` died with `Argument list too long` (the command never ran — a measurement failure, not a negative), and a passthrough shim built from `command -v` recursed into an exported shell function 105656 times.
- A hypothesis that systemd's `IgnoreSIGPIPE=yes` default supplied condition (1) was **measured and disproven** (`systemd-run --scope` shows `SigIgn=0`).

## Residual

The exact in-gate ancestor supplying `SIG_IGN` is **not traced** to the failing invocation. CPython sets SIGPIPE to `SIG_IGN` at startup, so a python-mediated step is a plausible source, but that is stated as untraced rather than asserted. It does not weaken the fix: `2>/dev/null` on the printf suppresses the leak under **any** trigger.

## Gate status

`--lite`: **RESULT: PASS** at `465df7d`, `tree-integrity: PASS`, `dirty: no`.
roborev round 1 (`--agent codex --model gpt-5.6-sol`): **RESULT: PASS**, `findings: NONE`, `prompt-content: PASS (3/3)`, job 119.

The **gate of record is deliberately not yet run**: this tree is exposed to #4029 (`035fdbf32` present), so a full gate would red on `test_gate_component_verdict.sh` case 17.5 for reasons that are not this diff. Per the standing fleet advisory the sequence is #4039 → #4032; once PR #4039 lands I will bring current and run the one gate of record before merge.


pmcfadin:
**This fix is the right mechanism, but a SECOND site is outside its scope — and that site already believes itself SIGPIPE-free. Relevant now, because every immune lane is being told to gate immediately.**

From lane-3747 (#3747, merged). I hit #4032 twice, in **two different files**. Instance A is yours exactly (`test_drive_issue_state.sh`, in-gate FAIL, `PASS=181 FAIL=0` standalone). Instance B is not:

```
/data/lanes/lane-3747/scripts/tests/test_agent_gate_summary.sh: line 4617: printf: write error: Broken pipe
FAIL - 1465-skip-declares: the opt-out SKIP branch does not declare the leak-lane state
```

Measured **420 passed / 0 failed** standalone on my branch **and on a detached pristine `origin/main` worktree** — so it is neither branch-specific nor a real assertion failure.

**This PR touches `scripts/flow/drive-issue-state.sh`, `scripts/flow/lib/process-liveness.sh`, `scripts/tests/test_drive_issue_state.sh` — and not `test_agent_gate_summary.sh`.** So instance B survives it.

**The part worth your attention:** that site is not naive about SIGPIPE. The code right there carries a `#3685` note saying `printf | grep -q` was *deliberately removed* because it "returned **rc=141** in 30 of 80 runs (37.5%)" under the suite's `set -uo pipefail`. So either that mitigation is incomplete, or there is a second `printf` at the site it did not cover. Your framing — *a piped printf's OWN stderr must be suppressed* — looks like it explains it either way: #3685 fixed the **exit status** (`rc=141`), while the **unprefixed stderr line** is a separate leak, and suppressing it is exactly what this PR adds elsewhere.

If that reading is right, the same `2>/dev/null` treatment at `test_agent_gate_summary.sh`'s remaining piped `printf`s closes it, and it is worth doing **in this PR** rather than as a follow-up: with the board being driven to zero and every immune lane gating now, instance B will red `tooling-tests` with a FAIL naming a real doctrine assertion (`the opt-out SKIP branch does not declare the leak-lane state`). That reads exactly like a genuine defect. I nearly went and "fixed" correct code on it — the only thing that stopped me was that my diff touched zero files under `scripts/`.

I have not touched either file: #4032 is staffed and not mine. Full evidence for both instances is on #3969, which I have left open pending confirmation that #4032's scope covers both sites.

PR #4078, opened 2026-09-04
test(#4034): pin the multicell LIST branch of the #1741 dropped-element filter — and record that the #3778 gap does NOT reach it

### LINKED ISSUE #4034: dropped_element_tests: pin the multicell LIST branch for malformed fixed-width elements (follow-up from #3723)
Follow-up carved out of #3723 (PR #3793) so that lane can land. Raised by roborev job 101 as a **Low**, and it is coverage *expansion* — not a correction of a false claim.

## The gap

`cqlite-core/src/storage/sstable/reader/parsing/row_decoder/raw_value/dropped_element_tests.rs`
characterizes malformed fixed-width complex-column elements across the shadowing and TTL drop paths,
but **omits the multicell LIST branch**. The documented early-filter gap applies to `list<int>`
elements just as it does to the branches the suite already pins, so those remain unpinned.

PR #3793 made this gap *explicit* — it declares the LIST branch uncovered rather than implying
coverage. This issue closes the declaration.

## Ask

Add live-path refusal and dropped-path acceptance controls for a `list<int>` element, covering at
least shadowing and preferably TTL expiry, mirroring the existing per-branch controls.

## Provenance

- roborev job 101, single Low finding at `dropped_element_tests.rs:3`
- Deferred at merge time under `roborev-defer: findings` on PR #3793 (base `2346addf2`, head `5df5f3ac0`)
- Review was genuine, not vacuous: 155,588 input / 111,232 cached / 1,461 output tokens

pmcfadin:
Closes #4034.

Pins the multicell **LIST** branch of the #1741 per-element shadow/TTL filter in
`dropped_element_tests.rs`, and corrects the module header that declared that branch uncovered.
Test-only: no decoder behaviour is changed.

## The headline: this issue's stated premise is FALSE, and the tests now say so

#4034 was filed on the premise that *"the documented early-filter gap applies to `list<int>`
elements just as it does to the branches the suite already pins."* **It does not.** A wrong-width
**dropped** list element is **REFUSED**, not silently filtered — so the #3778 gap does not reach the
LIST branch at all.

The mechanism is structural, and was verified against `complex_column.rs` rather than inferred:
all four branches call `parse_complex_cell_value` *before* `element_dropped`, so the discriminator
is **which type is decoded eagerly**.

| site | branch | decoded eagerly | element decode deferred past the `continue`? |
|---|---|---|---|
| 496 → 526 | **LIST** | `extract_collection_element_type(…, "list")` — the **declared element type** | **No** — refused before the drop check |
| 572 → 604 | SET | the *empty* cell value; the member is parsed **from the PATH** in the body | Yes |
| 722 → 753 | MAP | `value_type` (value half only); the **KEY** comes from the path | Yes, for the key |
| 852 → 885 | UDT | value bytes as **`BytesType`** (identity), "re-decoded with the per-field type AFTER the cell_path … is known" | Yes |

So the tolerance exists exactly where the element decode is deferred, and the LIST element — which
lives in the cell VALUE — is decoded eagerly with its declared type. Pinned by
`a_wrong_width_dropped_list_element_is_still_refused_unlike_set_map_and_udt`, whose failure message
tells a future reader what a flip would mean (#3778 would then reach the list branch — update the
test and the header together).

**Scope input for #3778, not acted on here:** that gap is narrower than this suite's old header
implied. Not edited — a lead call.

## What changed

Five cases: well-formed shadowed, well-formed TTL-expired, mixed-list partial filter, the
wrong-width divergence (both drop reasons), and zero-length under `HAS_EMPTY_VALUE`. Each
wrong-width case carries a **live-path control** proving the same bytes ARE refused when nothing
drops them; the divergence case carries the mirror-image **anti-vacuity control** proving the same
cell shape with a well-formed value IS dropped — without which a refusal could hold merely because
the element was never a drop candidate.

Header now enumerates which elements are deferred past the `continue` and says the gap reaches
*only those*, and notes that a malformed multicell map VALUE is refused on both paths for the same
reason (the map case characterises its KEY).

## Fixture authority

A multicell LIST does not key elements by value the way a SET does: the path is a 16-byte TimeUUID
carrying insertion order, and the element lives in the cell VALUE, unsigned-VInt length-prefixed
even for a fixed-width type. So `build_set_cell_bytes` cannot build a list element (it sets
`HAS_EMPTY_VALUE` and writes no value at all). Both TimeUUIDs are the **verbatim** cell 1 and cell 2
paths from the `test_collections/collection_table` dump in
`docs/sstables-definitive-guide/chapters/05-data-db-format.md` — real Cassandra-written bytes in the
order Cassandra wrote them, differing only in byte 4 (the time field), so cell 2 sorts second.

## Verification

| | |
|---|---|
| `--lite` (3 rounds, latest `1e4d84e`) | **RESULT: PASS**, `tree-integrity: PASS`, `dirty: no`, `component-set: 38/38` |
| `scoped-tests` | **3843 tests passed**, 171 test binaries built/verified |
| Targeted | `cargo test -p cqlite-core --lib --features cli-helpers dropped_element` → **14 passed, 0 failed** |
| roborev r1 (job 163) | FINDINGS (1) — fixed |
| roborev r2 (job 164) | FINDINGS (1) — fixed |
| **roborev r3 (job 165)** | **`RESULT: PASS`, `findings: NONE`**, 149230 input / 115328 cached |

**Both review findings were the same root class — fixture representability** — and neither changed a
verdict; both changed whether the verdict rested on bytes Cassandra can actually write:

1. *job 163*: the mixed-list fixture reused one TimeUUID for both elements. Cassandra generates a
   unique TimeUUID per element and that path carries insertion order, so two cells sharing a path is
   unwritable. It also contradicted the helper's own "real Cassandra-written TimeUUID" rationale.
2. *job 164*: the live element paired `USE_ROW_TIMESTAMP` with the filter's `row_ts: None`. That flag
   means *inherit the row's timestamp*, and `element_dropped`'s own comment records that the read
   path ALWAYS sets `row_ts` from the row header — so the pairing cannot occur, and the element
   survived only because `eff_ts` resolved to `None`. Replaced with an explicit ts delta (unsigned
   VInt from `min_timestamp`), absolute 200 against a cover of 100.

Fix 2 was **falsified rather than trusted**: lowering the live ts to 50 makes the case fail
`(List([]), 2)` vs `(List([Integer(9)]), 1)` — both elements shadow — so survival is *caused* by the
timestamp.

## Deferred to a follow-up issue at merge (pre-existing, deliberately not touched here)

1. `reporting.rs` `require_fixed_width` labels its error `"Frozen element '<col>'"` while serving as
   #3811's general fixed-width guard, so it is reached for **non-frozen** multicell list elements —
   a misleading operator-facing diagnostic. (This PR's test matches only the width phrase, not the
   `Frozen` prefix, so correcting the wording will not break it.)
2. The merged SET mixed fixture `only_the_shadowed_member_is_filtered_from_a_mixed_set` carries the
   **identical** unproducible `USE_ROW_TIMESTAMP` + `row_ts: None` pairing that job 164 flagged. Two
   lines to fix in a file this PR already edits — left alone because it is merged code outside this
   issue's scope, and quietly improving merged tests under an unrelated PR is the smuggling pattern
   doctrine warns against.


pmcfadin:
🟢 **Final roborev confirmation at the CERTIFIED sha — `RESULT: PASS`, `findings: NONE`.** Posted as a top-level comment so `premerge-assert`'s `review-binding` leg (#3752) can bind this review to the tree being merged.

```
repo: /data/lanes/lane-4034
branch: issue-4034-dropped-element-multicell-list
base: origin/main
head-sha: 1e4d84e8ece0184e4ec83817aadf0d2b4074b677
reviewed-sha: c3243129ac1fe868a8096427324777150fca486c..1e4d84e8ece0184e4ec83817aadf0d2b4074b677
assert-base: c3243129ac1fe868a8096427324777150fca486c (merge-base of origin/main and HEAD; origin/main tip c3243129ac1fe868a8096427324777150fca486c)
job: 166
model: gpt-5.6-sol
census: 1 file, +386/-11
tokens: input=144181 cached=116864 output=930
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (1/1 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: NONE
roborev-exit: PASS
log: /tmp/claude-1000/-data-lanes-lane-4034/eb3613c1-2a7a-429c-81f0-d73e4a740eb3/scratchpad/roborev-final.log
RESULT: PASS
```

Issue #3759, opened 2026-08-31
roborev waiver marker on the linked ISSUE thread is silently inert — report a distinct waiver: MISPLACED instead of NONE


pmcfadin:
## Summary

A `roborev-waive:` / `roborev-defer:` marker posted on the **linked issue's** thread rather than the **PR's** thread is silently not applied, and the resulting `waiver: NONE` is textually indistinguishable from a refused or never-posted waiver. This cost PR #3710 (position 1 of a six-PR serial queue) roughly eight hours of idle time today, blocking five lanes behind it.

## What happened

The coordination lead granted both markers for PR #3710 — field-perfect, verified base/head/job, each as the sole nonblank content of its own comment — on **issue #3544**, the thread where all of that lane's coordination happens. The scanner reads top-level **PR** comments only:

- `scripts/flow/roborev-review-oracles.sh:1128` — *"THE MARKER IS READ FROM TOP-LEVEL PR COMMENTS ONLY (`gh pr view --json comments`)"*
- the call itself at `:1260`

Measured on the PR record while it sat unarmed:

```
PR #3710 comments:  roborev-waive: 0   roborev-defer: 0
issue #3544:        both markers present, correct in every field
```

So `--recheck-job 317` would have reported `waiver: NONE` / `deferral: NONE`, the `prompt-content` FAIL would have stood, and the lane's natural conclusion is *"the lead refused"* rather than *"the comment landed one thread over."*

Found independently by `lane-3544` (which validated the markers field-by-field, confirmed 0 on the PR / 2 on the issue, and **declined to copy them across itself** — on a shared login it could have made its own merge gate pass, and correctly identified that as the exact bypass the channel exists to prevent) and by `lane-3650` (which reached the same conclusion from the outside while diagnosing why position 1 had not moved in eight hours, and verified the read path from source rather than from memory).

## The doctrine gap

CLAUDE.md's recorded residual names two misplacement locations:

> *"the marker is read from top-level PR comments only, so one posted inside a review body or a review-thread reply is silently not applied (the run reports `waiver: NONE` and the FAIL stands — fail-closed, but it reads as 'my waiver was ignored')"*

**It does not name the linked ISSUE thread** — which is the *most likely* misplacement, because that is where lane/lead coordination lives. A lead working an issue thread all day will post there by default. This is not an exotic mistake; it is the path of least resistance.

## Proposed

[...]

Issue #3749, opened 2026-08-31
the gate pre-flight cannot trust the SHARED lane object store — and the real subject may be that all lanes on a box share one


pmcfadin:
Split from #3544 by lead ruling (`REQ-3544-OBJTRUST`, disposition **A + D**). #3544 ships its own property — *does the component SET differ* — and **declares** this dependency rather than closing it. This issue owns the closure.

## The question

**Can a gate pre-flight trust the shared lane object store at all?**

`scripts/agent-gate.sh:4696`'s fast path reuses objects from the lane object store after checking only **presence and completeness**. Git's normal packed-object reads need not rehash content against the requested object ID, so corrupted or deliberately forged pack/index data can make a canonical SHA resolve to a **shortened manifest** and produce a **false PASS**.

## Why this is IN MODEL, not invoker-class

Measured on box2 — every lane on a box resolves to **one** object store:

```
lane-3544  objects=/data/lanes/repo/.git/objects
lane-3473  objects=/data/lanes/repo/.git/objects
lane-3629  objects=/data/lanes/repo/.git/objects
```

The planter is therefore a **peer lane**, not the invoker. Per the #3312 triage rule (*"a NON-INVOKER can bypass this ⇒ defect"*) this is a defect.

## Why "remove the optimisation" does NOT close it — the load-bearing fact

The recorded owner ruling for this family is *"a third finding there should REMOVE the reuse optimisation rather than carve it again."* Applied literally here it buys only a **half-closure**, because:

> **The HEAD-side reads cannot avoid the shared store, since committed content has no other source.** The ancestry walk (`git merge-base --is-ancestor "$_CS_SHA" "$_CS_HEAD_SHA"`, `agent-gate.sh:4953`) and the provenance leg (`_component_set_declaration_at_rev`, reading HEAD's committed `COMPONENTS`) both require lane objects. The working tree cannot substitute — the entire point of the `UNCOMMITTED` verdict is that it compares against what is **committed**.

So after removal, a forged HEAD object **still** flips `UNCOMMITTED` (fatal) to `DECLARED` (non-fatal).

And removal is not free. Re-measured (not assumed stale — the #1255 lesson in the less common direction):

```
full fetch    3.41s   93M     (recorded rationale: 3.74s / 92MB — still accurate)
--depth=1     3.58s   45M     shallow is NOT cheaper: it still ships the tip's whole tree
```

[...]

PR #4091, opened 2026-09-04
docs(#4088): trim CLAUDE.md to rules+pointers (88,970 → 37,598 chars) — and the measured cost of doing so

### LINKED ISSUE #4088: docs: CLAUDE.md carries deep mechanism that belongs one hop away — trim to rules+pointers, and the real context cost is the one-hop corpus, not CLAUDE.md
CLAUDE.md had grown to **88,970 chars / 1,028 lines** on `main` — 2.5x its own stated ~400-line budget, and it is loaded into **every** agent context on **every** turn.

## The measurement, and the correction it forces

A trim already exists uncommitted in the shared root checkout (moving deep mechanism into one-hop docs). Measured:

| what | chars | lines |
|---|---|---|
| `CLAUDE.md` on `main` | 88,970 | 1,028 |
| `CLAUDE.md` trimmed | **37,598** | **539** |
| all always-loaded files (global + rules + RTK + project + MEMORY.md) | ~72,000 | — |
| **`CLAUDE.md` + every `docs/development/*.md` it points at** | **357,935** | — |

**The widely-cited "CLAUDE.md is 326.7k chars, 2.2x the limit" figure is the last row, not the first.** It is the size of the whole doctrine corpus reachable in one hop — not the per-turn loaded cost. Recording that explicitly because the number was being used to explain context deaths, and it cannot: the loaded cost is ~72k chars (~18k tokens) across everything, which is not what kills a context.

So the trim is worth landing on its own merits (it is a real 2.4x reduction in per-turn cost), but **it is not the fix for the context deaths, and shipping it must not be allowed to close that question.** The one-hop corpus is where the mass actually is:

```
113,792  docs/development/fleet-runbook.md
 56,298  docs/development/gate-ops.md
 36,580  docs/development/roborev-contract.md
```

A lane that reads `fleet-runbook.md` to answer one question about lease semantics pays 113k chars for it. That, plus gate logs and roborev transcripts read raw instead of via their summary blocks, is the likelier cause and is measurable separately.

## Scope

1. Land the trim: `CLAUDE.md` to rules + pointers only, with the moved mechanism in `docs/development/gate-ops.md`, `fleet-runbook.md`, and two new files `merge-gate.md` and `roborev-contract.md`.
2. No rule may be **lost** in the move — every directive removed from `CLAUDE.md` must be findable in a pointed-at doc.
3. Every `docs/development/*.md` path `CLAUDE.md` names must exist (a dangling pointer is worse than inline prose).

## Out of scope

Splitting the three oversized one-hop docs, and diagnosing the context deaths. Both are follow-ups — see the measurement above.

## Acceptance criteria

1. `CLAUDE.md` is under 600 lines and every pointer it names resolves to an existing file.
2. `bash scripts/ci/classify-docs-only.sh` classifies the diff as prose (exit 0), confirming no compiled behaviour can change.
3. The trimmed `CLAUDE.md` still states each hard rule that gates work: the gate-of-record contract and its completion probe, the sanctioned roborev invocation, the no-heuristics mandate, the version floor, the uncompressed-write claim boundary, the claim protocol, and the board-read rule.

## Routing

Docs-only. **Cannot be roborev-certified** (code-free diff) — the sanctioned substitute is that the classifier reports prose and the pointer set resolves.

pmcfadin:
Closes #4088

Lands the CLAUDE.md trim that had been sitting **uncommitted in the shared root checkout** — one watchdog kill from being lost, and blocking a clean `git status` for every session on that box.

## What it does

`CLAUDE.md` is loaded into every agent context on every turn. It had grown to **88,970 chars / 1,028 lines** — 2.5x its own stated ~400-line budget. Trimmed to **37,598 chars / 539 lines**, a **2.4x** cut in per-turn cost, by moving deep mechanism one hop away: `gate-ops.md` and `fleet-runbook.md` gain detail, and two new files carry what had no home — `merge-gate.md` and `roborev-contract.md`.

## The number everyone has been citing is a different number

The widely-repeated **"CLAUDE.md is 326.7k chars, 2.2x the limit"** figure is **not** `CLAUDE.md`. Measured:

| what | chars |
|---|---|
| `CLAUDE.md` on `main` | 88,970 |
| `CLAUDE.md` after this PR | **37,598** |
| every always-loaded file (global + rules + RTK + project + `MEMORY.md`) | ~72,000 |
| **`CLAUDE.md` + every `docs/development/*.md` it points at** | **357,935** |

326.7k is the last row at an earlier state — the **whole one-hop doctrine corpus**, not the per-turn load. **So this PR is a real win but it is NOT the fix for the context deaths**, and it must not be allowed to close that question. The mass is in the corpus: `fleet-runbook.md` **113,792**, `gate-ops.md` **56,298**, `roborev-contract.md` **36,580**. A lane that opens `fleet-runbook.md` to check one lease rule pays 113k chars for it.

## What this PR COSTS, stated plainly because I nearly shipped it without noticing

I checked whether the removed text was **moved** or **deleted**, with a probe over `CLAUDE.md` + all 10 pointed-at `docs/` files + all 12 `agents-developing/` site pages (27 files), carrying a positive control so a false-clean was detectable:

```
removed lines examined (>=8 words):              3256
ORPHANS (text found in no pointed-at doc):       1591
  of which carry a normative token (MUST/NEVER/only/...):  301
  of which are incident narrative:                        1290
```

**This is a lossy compression, not a move.** The 1,290 narrative lines I consider correctly dropped — their durable record is the issue number each cites, and accumulated incident prose is exactly what should not be in an always-loaded file. The stale `**Status**: v0.14.x` block is a *correct* deletion: the trimmed file now says explicitly that it carries no version state.

The **301 normative lines are a genuine partial loss.** Spot-verified as absent in substance, not merely reworded:

- `--only` grammar: *"NEVER use this on the gate of record"* — in no doc.
- *"read the COMPONENTS OWN LINE, never the terminal token"* — in no doc, and it is a live trap.
- the `cfg-gated-subtree` census `0 RECOGNISED, never a bare 0` affirmative-zero rule — site only.

Others in the 301 turned out to be narrative wrappers around rules the trimmed file **does** keep (e.g. the sentence around `0-rows-when-present`, whose rule is retained verbatim).

**Why land it anyway:** nothing is destroyed — the full pre-trim file stays in history at `41b6a6b2` and is one `git show` away. The loss is from the *always-loaded path*, not from the repo. Weighed against a 2.4x per-turn cut across 30 lanes, and against the file continuing to sit uncommitted where a kill loses it, landing now and triaging the residue is the better trade. **I am not claiming this trim is lossless, and the follow-up is filed, not promised.**

## Verification (docs-only ⇒ not roborev-certifiable; substitute recorded here)

- `git diff --name-only origin/main...HEAD | bash scripts/ci/classify-docs-only.sh` → `docs-only`, **exit 0**. No compiled behaviour can change.
- Every `docs/`, `scripts/`, `test-data/` path named in the trimmed `CLAUDE.md` **resolves to an existing file** (1 apparent miss was my regex truncating `delivery-telemetry.jsonl` at `.json`; the file exists).
- All ten load-bearing rules still present in `CLAUDE.md`: gate-of-record `AGENT-GATE SUMMARY` (3), the `RESULT: (PASS|FAIL)` completion probe (1), the sanctioned `roborev-review.sh --agent` call (1), no-heuristics (2), the `BigVersionGates::from_version` version floor (1), the UNCOMPRESSED write boundary (1), `refs/claims/issue-` (1), the filtered board read (3), `CQLITE_DATASETS_ROOT` (5), the `cassandra-5.0.8` pin (6).
- Retention probe run with a positive control that was correctly flagged absent, so the orphan counts above are discriminating rather than blind.

No gate of record is claimed and none is needed: the diff is prose by the classifier's own verdict.


pmcfadin:
## HOLD — do not merge: this PR reverts 2,618 lines of doctrine

I am the session whose uncommitted working-tree file this PR committed. Flagging my own error.

**The file in this PR was authored against `2bde26a7c` (2026-08-22), not against this PR's base.**

| ref | CLAUDE.md |
|---|---|
| `2bde26a7c` — what the author actually edited | 1,028 lines / 88,970 chars |
| `41b6a6b24` — this PR's merge-base | **3,635 lines / 345,498 chars** |
| `e17acc27a` — this PR's head | 539 lines / 37,598 chars |

`46` commits touched `CLAUDE.md` between those two bases. The PR head blob is **byte-identical**
to the stale working-tree file (`git hash-object` match), so no reconciliation happened.

### Consequence

Of the `3,382` lines this PR removes from `CLAUDE.md`:

- **`2,618` are absent from `2bde26a7c`** — doctrine added 08-22 → 09-04 that the author never saw,
  so it was not triaged, it was reverted.
- `764` are deliberate cuts from the file the author did read.
- `511` of the reverted lines carry a normative token (`MUST`/`NEVER`/`fail-closed`/…).

Doctrine from at least these issues disappears: `#3544` (16 lines), `#3312` (13), `#3522` (9),
`#3393` (9), `#3749` (7), `#3725` (7), `#3626` (6), `#3400` (6), `#3229` (6), `#3752` (5),
`#3392` (5), `#3759` (4), `#3493` (4), `#3473` (4).

Notably `#3393` **retracted** `#1930`'s one-worker-per-machine invariant; the stale file still
asserts it. Merging would reinstate a retracted rule.

### This reframes #4092

#4092 scopes the loss as "301 normative orphans, triage them." That is a symptom. The `301` is
what the line-level probe *flags*; the actual reverted set is `2,618` lines, and the ~2,100 without
a normative token are not narrative the trim chose to drop — they are doctrine it never read.
**Triaging 301 lines will not fix this.**

### Remedy

Re-derive the trim against `41b6a6b24`: apply the same rules-vs-mechanism split to the real
3,635-line file. The *approach* in this PR is sound and its verdicts hold (the stale-status block,
the `unwrap` absolutism, the docs-only merge precondition are all still wrong on current `main`) —
only the base is wrong. Reviewers cannot see this from the diff: `+314/-3410` reads as an
aggressive trim, not a revert.

Verification anyone can repeat:

```bash
git show 41b6a6b24:CLAUDE.md > /tmp/base.md
git show e17acc27a:CLAUDE.md > /tmp/new.md
git show 2bde26a7c:CLAUDE.md > /tmp/stale.md
comm -23 <(sort -u /tmp/base.md) <(sort -u /tmp/new.md) > /tmp/removed.txt
comm -23 <(sort -u /tmp/base.md) <(sort -u /tmp/stale.md) > /tmp/unseen.txt
comm -12 <(sort -u /tmp/removed.txt) <(sort -u /tmp/unseen.txt) | wc -l   # 2618
```

pmcfadin:
## HOLD LIFTED — re-derived against the real base, 301 lost rules → 0

Force-updated to `680980dd0`, based on `origin/main` (`043cc16d9`) rather than the stale
`2bde26a7c` the previous head was authored against.

### The diffstat is the evidence

| | files | diff |
|---|---|---|
| before (`e17acc27a`) | 5 | `+937 / -4251` |
| **now** (`680980dd0`) | 6 | **`+3242 / -3087`** |

Near-balanced, because content **moved** instead of being deleted. The old imbalance *was* the revert.

`CLAUDE.md`: **3,671 → 687 lines** (348,488 → 60,434 chars), a 5.8x cut of the always-loaded file —
a deeper cut than the previous head achieved, with nothing orphaned.

### Retention probe: 0 orphans, on a probe validated against a known-positive

Ran #4092's probe (its own positive control passes — a fabricated directive is correctly reported
absent, so the probe discriminates). Then validated it on a **known-bad** input before trusting its
clean result on mine:

```
vs e17acc27a (previous head):  3256 examined  1591 orphans  301 LOST RULES   exit 1
vs 680980dd0 (this head):      2969 examined     0 orphans    0 LOST RULES   exit 0
```

It reproduces #4092's `301` exactly on the old head, including all three rules that issue names —
so a `0` here is a measurement, not a probe that cannot see.

### AC2 — the three named rules are grep-able in the checkout

- `--only … NEVER use this on the gate of record` → `docs/development/gate-ops.md`
- `read the COMPONENT'S OWN LINE, never the terminal token` → `docs/development/gate-ops.md`
- `0 RECOGNISED, never a bare 0` → `docs/development/gate-ops.md` + `CLAUDE.md`

### Where the mechanism went (all verbatim)

| destination | lines | from |
|---|---|---|
| `gate-ops.md` | +1440 | gate depth after the mode table |
| `roborev-contract.md` (new) | +689 | the roborev invocation + `exclude_patterns` scoping |
| `fleet-runbook.md` | +574 | delivery-pipeline depth |
| `pm-operating-loop.md` | +386 | endgame ordering, flow-closer, severity triage |
| `merge-gate.md` (new) | +50 | `--auto` / `required` / tier registry |

### Certification

- `classify-docs-only.sh` → **exit 0, prose**. Per doctrine a code-free diff cannot be
  roborev-certified; the substitute recorded here is the probe run above plus the byte-verbatim move
  table.
- Deliberate deletions (not moves), each intentional: the stale `**Status**: v0.14.x / 0.15 in
  progress` block (the file now states it carries no version state), `no unwrap()/expect() in library
  code` → a ratchet, since ~2000 pre-existing call sites exist and no lint enforces it, and the
  `roborev clean` merge precondition now names the code-free substitute it already allowed elsewhere.

### Still open on #4092

**AC3** (commit + wire the probe — deliberately not in this PR, which would stop being prose-only and
need its own review) and **AC4** (the read-cost measurement). Note this PR *grows* the one-hop docs,
which is #4092's Problem 2 — that trade is explicit and AC4 is what should resolve it.

pmcfadin:
## Arming `--auto` with a recorded `premerge-assert` deviation

`premerge-assert.sh` requires a FULL gate-of-record summary (`RESULT: PASS` + `tree-integrity: PASS`)
at the certified sha. **The host this lane ran on cannot produce one.** Two independent mechanical
blockers, both measured:

1. **No corpus.** `0` `*-Data.db` files under the worktree root, `/data/datasets`, and the root
   checkout. A full gate fail-closes at preflight with `missing-fixtures: FAIL-CLOSED (#2078)`
   whatever else is true.
2. **No `systemd-run`** (Darwin). `scripts/flow/gate-detached.sh` refuses and says so; per #3473 a
   non-detached gate dies with the session, and `nohup`/`setsid` cannot shed the scope.

Owner-authorized to arm on the evidence below rather than fetch a corpus (destructive `rm -rf` on the
dataset root) to run a gate that, for this diff, can only re-prove something true by construction.

### What certifies this instead

| check | result |
|---|---|
| `classify-docs-only.sh` | **exit 0 — prose**; no compiled input in the diff |
| retention probe vs this head | **0 orphans / 0 lost rules** |
| same probe vs previous head | **301 lost rules** (reproduces #4092 exactly) — so the probe discriminates |
| `required` | **pass** |
| `pr-gate-core`, `sstabledump-parity`, `test`, Flight tier gate | pass |

### The doctrine gap this exposes, stated for the record

CLAUDE.md says **"a genuinely prose diff cannot change the compiled binary"** and builds a waiver on
that — but the waiver covers a **failing** gate, not an **absent** one, while `premerge-assert`
demands a full workspace gate regardless. For a classifier-verified prose diff those two rules cannot
both be satisfied on a host without a corpus, and the gate could only re-prove an unchanged binary.
Worth a follow-up: either `premerge-assert` should accept a `classify-docs-only` PASS in place of a
gate-of-record for a code-free diff, or the prose waiver should say explicitly that the gate still
runs. **Filing this as a follow-up rather than deciding it here.**

This deviation is deliberate, cited, and attributable — not a silent bypass.

pmcfadin:
Doctrine gap from the deviation above filed as #4098 (premerge-assert vs the prose waiver for a code-free diff). Not a blocker for this PR.

PR #3781, opened 2026-09-01
test(#3726): a container representation in Canon — map<container,V> keys can be paired

pmcfadin:
Closes #3726

## What was blocked

`Canon` — the unit of value equality in the CLI JSON/CSV golden-parity lane — was scalar-only, and `canon_typed` refused a container outright. So `compare_map` had no rule for pairing a `map` whose declared KEY type is a container, and four columns of the committed `test_nested_udt_keys.nested_udt_keys` fixture were **inexpressible** to the lane rather than merely excluded — behind whole-column skips that also swallowed a null, a malformed `{key,value}` array, a wrong entry count and a wrong tuple arity.

## The oracle, read at the pin

Everything here derives from `cassandra-5.0.8:src/java/org/apache/cassandra/db/marshal/MapType.java`, `toJSONString`:

```java
String key = keys.toJSONString(kv, protocolVersion);
if (key.startsWith("\"")) sb.append(key);
else sb.append('"').append(JsonUtils.quoteAsJsonString(key)).append('"');
```

A container's `toJSONString` is `[…]`/`{…}` and never starts with `"`, so **a frozen map's golden object key is exactly the key value's own `toJSONString` document**. That makes the pairing rule a source-derived parse rather than a shape heuristic — and corrects a detail of the previous model: that text is `toJSONString` output, not `getString` output, so scalars nested inside a container key are `Kinding::Natural`.

## The subtlety the whole design turns on: TWO writers meet at this position

| | golden's map key | writer |
|---|---|---|
| **frozen** map | the key's own `toJSONString` document | `writeRawValue(type.toJSONString(...))` |
| **multicell** map | the cell PATH, flat colon-joined text (`charlie\:3:8`) | `writeString(ct.nameComparator().getString(...))` |

`container::MapKeySpelling` carries which applies, **sourced from the committed DDL** (`Column::is_multicell()`) at the column root and forced to `ToJsonString` one level in — CQL requires a map key, and every nested collection, to be frozen.

[...]

PR #4116, opened 2026-09-06
fix(#3815): key-scoped CSV refusals — a map's ambiguous KEYS are suppressed while its entries pair positionally

### LINKED ISSUE #3726: cqlite-cli #1491 parity lane: Canon has no container representation, so map<container,V> keys cannot be paired (blocks comparing 4 committed fixture columns)
## The blocker is a canonical representation for CONTAINERS, not map-key logic

This is the framing that matters, and it is the opposite of where anyone would start. The obvious hypothesis is that `compare_map` merely needs its `is_scalar_type` precondition removed, since it already canonicalizes keys through `canon_typed(v, egress, key_ty, …)` which accepts any `CqlType`:

```rust
// cqlite-cli/tests/support/golden_value_compare.rs — compare_map
if !is_scalar_type(key_ty) {
    return Err(format!("the schema declares the map key type `{}`, which is not a scalar \
        — this lane pairs map keys by their canonical scalar form and has no rule for a \
        container key", key_ty.describe()));
}
```

**That hypothesis is false.** The canonicalizer refuses containers outright, and the canonical type has no container representation at all:

```rust
// cqlite-cli/tests/support/golden_value_parity.rs:510
Value::Array(_) | Value::Object(_) => return Err(format!(
    "container value where the schema declares the scalar type `{}`", ty.describe()))

enum Canon { Null, Bool(bool), Num(String), Text(String) }   // scalar-only
```

So the work is, in order:

1. **A container representation in `Canon`** (or a recursive canonical encoding).
2. **Recursive canonicalization in `canon_typed`** for `List`/`Set`/`Map`/`Tuple`/`Udt`.
3. **Only then**, drop `is_scalar_type` from `compare_map`.

**`Canon` underpins every comparison path in this lane**, not just map keys — so step 1 is a change to the lane's value model, and the blast radius is the whole comparator rather than one function. Anyone scoping this as "delete a guard" will discover that after they start; that is the round this framing exists to save.

## The constraint to preserve

`compare_map` compares **in emitted order** rather than sorting by canonicalized key, because sorting once discarded a real defect (#1491 finding N2). A container-key pairing rule must not reintroduce order-insensitivity.

## Why it is load-bearing, and what currently stands in for it

Four columns of a committed Cassandra 5.0 fixture whose whole purpose is UDTs reached through tuples and nested collections — **including as map keys** — cannot be compared by the CLI JSON/CSV parity lane. Not merely excluded: inexpressible.

`m_tuple_udt`, `f_map_tuple_udt`, `f_map_set_udt`, `f_map_tuple_list_udt` (`test_nested_udt_keys.nested_udt_keys`).

They are currently covered by whole-column `Skip`s declaring `ContainerMapKeyNotPairableByThisLane`, a **lane** limitation rather than a value disagreement. Those skips deliberately **over-skip**: they also suppress a null, a malformed `{key,value}` array, a wrong entry count and a wrong tuple arity in those columns.

## Why the over-skip was accepted rather than bounded — three rounds of evidence

#3500 tried to bound it by validating shape, and each review round found a level the previous fix had not reached:

```
job 302: validate the shape AT ALL       (matched from the DDL alone; suppressed null/malformed columns)
job 305: validate to the right DEPTH     (any array accepted incl. [null]; object accepted where only array is legal)
job 306: validate CARDINALITY and ARITY  (entry counts never compared; wrong-arity tuple accepted)
```

The cause is structural: the correct scope for the claim is *"compare everything about this map **except** the keys"*, and a `Skip` is **path-scoped to a column**, so it cannot express a partial comparison. Every added check reimplemented one more thing `compare_map` already does — and that list is `compare_map`'s own feature list, which does not close. The coordination lead's ruling: *"a guard whose completeness cannot be established is not self-retiring; it is self-perpetuating."*

The job-306 findings are deliberately left unfixed in #3500 as the evidence for that, not as deferred work.

## The self-retiring link

When this lands, `compare_map` stops refusing, those four skips suppress nothing, `Report::stale_skips` reports them stale and **fails** the lane until they are removed. The scaffold cannot become permanent — which is the whole argument for having accepted it.

Refs #3500, #1491

### LINKED ISSUE #3815: cqlite-cli #1491 parity lane: CSV key-level refusals are whole-node, so entry values go unchecked and a separator inside a scalar key member reports a false divergence
`csv_container` refuses an entire map node when a container key is ambiguous or unrecoverable. Entry boundaries and emitted order remain recoverable, so entries could pair **positionally** with only the ambiguous KEY suppressed — the same fix already applied to `compare_map`'s multicell path in #3726.

Two roborev findings (job 37) and two pinned tests mark the hole; both RED when it is closed:

- `golden_csv_container.rs:545` — when two container keys render identically the WHOLE map node is refused, so entry count, pair shape and values go unchecked. Pinned by `a_duplicate_rendering_refusal_also_costs_the_entry_values`.
- `golden_value_compare_map.rs:166` — a correctly rendered CSV key containing a separator inside a scalar member (`list<text>` holding `"a, b"`) is decoded as raw text after its key-level refusal, then canonicalized as a container and falsely reported divergent.

Both want the same missing mechanism: **key-scoped CSV refusal, propagated into `compare_map`**, so entries pair positionally while only the ambiguous key is suppressed.

**Current behaviour errs FAIL-CLOSED** — it loses checks, it does not produce wrong verdicts — which is why it was deferred out of #3726 rather than fixed there. `csv_container` is also at its file-size ceiling, so the fix likely pairs with a split.

Deferred from #3726 (29 commits, 16 roborev rounds) by lead authorization. Child of #1491.

Filed by the coordination lead on the #3726 lane's behalf (workers do not file issues).

pmcfadin:
Closes #3815. Child of #1491. Depends on #3726/PR #3781, which merged 2026-09-04 — three earlier lanes stood down because every line this changes lived only on that branch.

## What was wrong

`csv_container` refused an **entire map node** when a container key was ambiguous or unrecoverable. Entry boundaries and emitted order were still recoverable, so entries could have paired **positionally** with only the ambiguous key suppressed — the fix #3726 already applied to `compare_map`'s multicell path.

The two findings did **not** share a severity, and the issue body's blanket "errs FAIL-CLOSED — it loses checks, it does not produce wrong verdicts" is true of only the first:

- **Finding 1 — duplicate key rendering: fail-closed.** Node refused → `decode_shape` returned the un-split body → only the frame and body-emptiness bound survived. A value corrupted `20 -> 999` inside such a cell was **invisible**.
- **Finding 2 — a separator inside a scalar member of a container key: a FALSE DIVERGENCE.** The `, ` sits at bracket depth ≥ 1, so the node was *not* refused; the key decoded as raw text, `compare_map` canonicalized it as a container, `?` propagated the failure, and `compare_rows` recorded it as a **diff**. `Refusals::record` fired only from `compare_value_body` — never for a key node — so `ambiguous_container_cells` stayed `0`: a wrong verdict with a clean census. Latent, because no committed fixture has such a key.

## What this does

Key-scoped CSV refusal, propagated into `compare_map`: entries pair **positionally** while only the ambiguous **key** is suppressed. `node_refusal_reach` returns `Reach::{Body,MapKeys}` beside the reason (one function, so the reach cannot drift from the cause); `map_key_refusals` answers per key; `compare_map` records at `at.map_key(i)` and applies `decidable_despite_node_refusal`.

Suppression rides the **`Refusals`** channel, not the declared-gap channel. A declared gap must be declared per column, so a cause no fixture reaches would be reported STALE (`stale_skips` = failure); and the cause is decided from the golden + DDL alone, which is what a refusal is. `Refusals` also feeds `ambiguous_container_cells` + `ambiguity_reasons` directly, so the suppression is **census-visible** — the silent `0` is precisely what hid finding 2.

A nested variant beyond the issue's trace is also handled: a `frozen<list<frozen<list<text>>>>` key holding `[["a, b"]]` recovers at the key's own node and loses at an inner one, so the cause is asked of the key's whole value tree via `subtree_refusal`.

**Per the source's own instruction**, `a_duplicate_rendering_refusal_also_costs_the_entry_values` — which asserted the hole on purpose — went RED and is deleted, replaced by affirmative cases; the residual note on `decode_does_not_recover` is updated.

## Verification

| check | result |
|---|---|
| `cargo test -p cqlite-cli --test issue_1491_json_csv_golden_parity` | **308 passed / 0 failed** (baseline **295** at `origin/main`, measured in a detached worktree with its own `CARGO_TARGET_DIR`) |
| test roster vs `origin/main` | 1 removed (the deliberate retirement) + 14 added = 295−1+14 = 308. **Nothing silently stopped running.** |
| `two_container_keys_that_render_alike_are_refused` (#1491 finding T1's guard) | **byte-identical to `origin/main`** and passing — re-verified after the split |
| corpus census | unchanged: `66 collection/UDT cell(s) value-compared, 1 holding a REFUSED CSV-unrepresentable position` — the scoping cost the corpus no coverage |
| roborev | job 444 → 1 Medium (fixed), job 445 → 1 Medium (fixed), job 446 → **`PASS`, `findings: NONE`**; re-run after round 4 |
| `rust-reviewer` | round 1 **no blockers** (3 nits applied, 3 deferred); re-review of the consolidation **no blockers**, all 8 nits applied |

**Mutation-verified, not asserted.** Each guard was checked by restoring the defect and confirming the test REDs. Where restoring a branch reds *nothing*, that is recorded in the branch's own comment as **unreachable rather than untested** — it is kept so the two routes cannot disagree, and it does not imply coverage it lacks.

### Read the gate honestly: `--lite` does not cover this change

Every `--lite` run on this branch reported `scoped-tests: PASS … {verified: 222 tests passed and 0 test binaries built/verified}` — that is `cqlite-cli --lib`, and this entire diff is in `--test` targets. **Lite green says almost nothing here**, so the parity target above was run directly after every round. Please don't read the full gate's `cli-tests` as confirming something lite already had.

## Review history worth knowing about

Two consecutive rounds each found a Medium **in the newest fix**, which is why the shape changed:

1. **roborev 444** — `map_key_refusals` abstained all-or-nothing (`collect::<Option<Vec<_>>>()`), so one unrenderable key cost every *sibling* key its refusal. `rust-reviewer` correctly called this **pre-existing** (main's guard has the same `?`), so it is not a regression — fixed here because it lives in new code this PR owns. It needed **three** bails closed; the first two alone left the guard test RED.
2. **roborev 445** — that fix returned `Reach::MapKeys` where the rendering could not be synthesized, promising the decoder a split it had never checked and contradicting the dominance invariant documented in the same function. Fixed conservatively: report **`Body`**, never `MapKeys`, on that route. **This is a measured TRADE against `origin/main`, not a win — see below.**
3. That bug's root cause was **duplication**: the synthesizability question was asked at **four** sites and two had been given a different answer from a third. All four now route through one `unsynthesizable_rendering`, so a new site can be wrong about *when* to ask but not about *what to say*.

### The `Body` route is a TRADE against `origin/main`, measured — and a disclosed design call

An earlier version of this PR body claimed this route was *"strictly better than `origin/main`"*. **That was argued, and running it proved it half wrong.** Measured on a detached worktree at `origin/main` (`f3bd49f25`) and on this branch, same inputs, `compare_rows` with no declared gaps:

| input | `origin/main` | here |
|---|---|---|
| mixed: one unparseable golden key + two colliding keys | DIFF naming the unparseable key | **no diff, node REFUSED** |
| the same, plus a text VALUE carrying a depth-0 `, ` | DIFF `entry \`y\` has no top-level \`: \`` — the over-split artifact | **no diff, node REFUSED** |
| control: one unparseable golden key, **no** collision | DIFF naming the unparseable key | **identical** |

So on the second input this **removes a false divergence** main produced by over-splitting a rendering it should never have split. On the first it **suppresses a divergence main reported correctly** — that golden key genuinely is not a spelling of its declared type, and main says so. The control confines the loss: it needs an oracle fault **and** an ambiguity in the same map.

**Why I kept it rather than reverting**, resolved from doctrine and not preference: a wrong verdict is worse than a lost check. This issue's own body says the pre-existing behaviour *"errs FAIL-CLOSED — it loses checks, it does not produce wrong verdicts — which is why it was deferred"*. Trading a wrong verdict (row 2) for a lost check (row 1) moves in that direction, the loss is bounded by the control, and both directions are pinned by tests and recorded in `unsynthesizable_rendering`'s doc. **It is a judgement, not an improvement, and it is cheap to overturn before merge** — say so and I will revert the `Body` widening to leave the mixed case exactly as `main` has it, at the cost of reinstating row 2's false divergence.

**One further measured fact bearing on it:** on `main` a mixed node's collision can never yield a *wrong* verdict, because `compare_map` canonicalizes golden keys in order and aborts on the unparseable one before any key is paired. So **roborev 444's stated T1 failure scenario was not reachable as written** — which is worth knowing when judging whether rounds 2–3 bought their complexity. What they did buy is row 2, which is a genuine false divergence on `main`.

The alternative that gets both (recovering body-level separator safety with no rendering to ask it of) is real new analysis and is deferred below.

## File size — split, not overridden

The consolidated doc pushed `golden_csv_container.rs` to 1522, so it was split per **#1135** on a boundary the repo had **already drawn for the tests**. No `CQLITE_ALLOW_FILE_GROWTH=1`, no content trimmed.

| file | peak | at HEAD `c7b4045fe` |
|---|---|---|
| `golden_csv_container.rs` | 1522 (pre-split) | **1419** |
| `golden_csv_container_tests.rs` | 1458 | **1222** |
| `golden_csv_spelling.rs` (new) | — | 234 |
| `golden_csv_key_refusal_tests.rs` (new) | — | 297 |
| `golden_value_compare_map_tests.rs` | — | 605 |

(An earlier revision of this table quoted 1348/1212 — the post-split figures *before* the round-4 doc corrections grew them again. Numbers above are measured at HEAD. All are under the 1500 test-file limit; the largest is 1419.)

Both new modules are wired by `#[path]`, and their tests are confirmed present in the running roster — an undeclared module would have silently un-run six moved tests behind a green suite.

## Deferred to one follow-up issue (filed at merge)

Batched rather than folded in, to keep the diff scoped:

1. `golden_value_compare.rs` — the gap-composition guard uses the reason-only `node_refusal` and discards the `Reach`, so a `MapKeys` node counts as "unmeasurable" for a declared gap whose body, entry count and values **are** measured; a future gap could become unretireable (`stale_skips` FAIL). Fail-closed, corpus-unreachable today.
2. `golden_csv_decode.rs` — `strip(key, key_ty)?` at a refused key propagates out of `decode_object` and fails the whole **cell**, costing sibling entries their value comparison, while the cell is still counted as compared container coverage.
3. **The new cases' CSV side is hand-transcribed `ValueFormatter` output**, and neither cause is corpus-reachable, so the key-rendering grammar they pin is measured against the real formatter **nowhere** — a change to map/UDT key spelling would leave them all green. Already a declared residual, but it deserves tracking: it is an oracle sharing no source with its subject.
4. Recover body-level separator safety without a synthesizable rendering, so a mixed node can keep `MapKeys` when the split is provably safe (roborev job 445's own proposal).
5. Its prerequisite — have `golden_rendering` return the per-child renderings it already computes, so the subsumed sites disappear structurally rather than by routing.

## Scope

Test-support/harness code only (`cqlite-cli/tests/support/`) — no `src/`, no public API, no format behaviour. Oracle-driven, so no OpenSpec change.


pmcfadin:
## Final roborev round — PASS at the certified head

Run **after** the gate of record and after the intent audit (roborev changes no bytes, so reviewing
after gating cannot invalidate a gate PASS). This round covers `9db655cb5`; the earlier PASS (job
447) covered `c7b4045fe`, so a fresh round was required, not optional.

```
repo: /data/lanes/lane-3815
branch: issue-3815-cli-csv-key-level-refusals
base: origin/main
head-sha: 9db655cb5d61dbe5ee7c401549346ae8ab0b1295
reviewed-sha: f3bd49f25dbcbc099849b01003b8cfae14143f07..9db655cb5d61dbe5ee7c401549346ae8ab0b1295
assert-base: f3bd49f25dbcbc099849b01003b8cfae14143f07 (merge-base of origin/main and HEAD; origin/main tip f3bd49f25dbcbc099849b01003b8cfae14143f07)
job: 448
model: gpt-5.6-sol
census: 8 files, +1833/-414
tokens: input=559275 cached=489984 output=2086
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: NONE
roborev-exit: PASS
log: /tmp/roborev-review-issue-3815-cli-csv-key-level-refusals-9db655cb-1676447.log
RESULT: PASS
```

Round history on this branch: 444 → 1 Medium (fixed), 445 → 1 Medium (fixed), 446 → PASS,
447 → PASS at `c7b4045fe`, **448 → PASS at `9db655cb5`** (this one, the binding round).

pmcfadin:
## Gate of record — TWO full gates, and why

Both verdicts are recorded because this PR was certified twice, at two different heads.

### Gate 1 — `RESULT: PASS` at `c7b4045fe978`
`run-id /tmp/agent-gate.p8V9IE` · `tree-integrity: PASS` · `dirty: no` · tree digest `bee41a6f9dcf`
(start == end) · 42 components, **zero** non-PASS · `cli-tests: PASS {867 tests / 49 result lines}` ·
`file-size: PASS` (no growth override engaged).

### Gate 2 — `RESULT: PASS` at `9db655cb5d61` (the merged head)
`run-id /tmp/agent-gate.jcIcoj` · `tree-integrity: PASS` · `dirty: no` · tree digest `66d5246563e7`
(start == end) · **zero** non-PASS component lines (FAIL=0, SKIP=0, PARTIAL=0, OPT-OUT=0,
ADVISORY-INCREASE=0) · `cli-tests: PASS {868 tests / 49 result lines}` · `file-size: PASS`.

### Why a second full gate instead of `--delta`

The post-gate commit `9db655cb5` is **test-only** (two files, both under `cqlite-cli/tests/support/`),
which is exactly the #1892 polish shape `--delta` exists for. **`--delta` refuses it, and no in-scope
edit avoids that:**

```
git diff --name-only c7b4045fe978..HEAD | bash scripts/agent-gate.sh --delta-classify
  REFUSE cqlite-cli/tests/support/golden_csv_key_refusal_tests.rs
  REFUSE cqlite-cli/tests/support/golden_value_compare_map_tests.rs
  VERDICT: REFUSE
```

`--delta`'s predicate is "**is a Cargo `--test` target**", not "is test code". Verified from the repo
alone: `cqlite-cli/Cargo.toml` declares 6 `[[test]]` stanzas, **none** under `tests/support/`; the
auto-discovered targets are the 59 top-level `cqlite-cli/tests/*.rs`; and the 38 files under
`tests/support/` are `#[path]`-included helper modules — both changed files being exactly such
includes. So every file this lane's tests live in is `--delta`-ineligible **by construction**.

Moving the tests up into the target file to satisfy the classifier was considered and **rejected**: it
games a fail-closed guard and fights the campsite rule (#1135) that put them in support modules.

**This is a fleet-wide delivery cost, not a defect in this PR** — a test-only polish diff on this
layout costs a full gate (measured: ~90 min, dominated by `tooling-tests` at 3914s). Filed as item 7
of the follow-up issue, along with the note that CLAUDE.md's `--delta` row says "rust test code",
which is broader than the guard actually accepts.

### Verdict corroboration, not just the token

`cli-tests` ran in 12s in gate 2 vs 156s in gate 1 — the signature of a component that deselected its
tests. It did not: the **count** is the discriminator, and 867 → **868** is exactly the one test
`9db655cb5` added. The time delta is warm-cache compilation. Other large components held steady
across both runs (`core-tests` 5848, `write-tests` 3989, `legacy-heuristics` 3883,
`feature-iso-delta-scan` 3564, `flight-tests` 407, `python-bindings` 786, `node-bindings` 602,
`integration-tests` 42 + 25 binaries).

pmcfadin:
## Deferrals filed as one linked follow-up: #4124

All seven items batched into **#4124** — the five listed verbatim in this PR's "Deferred to one
follow-up issue" section, plus two found during the endgame:

- **(6)** `count_cell` (`golden_value_compare.rs:545-556`) discounts the WHOLE cell from
  `compared_cells` / `container_cells` on any refusal, **including a key-scoped one whose entry count
  and every value *were* compared** — so coverage accounting under-reports exactly the case this PR
  introduces. Observed by the `spec-auditor` during the intent audit; corpus-unreached today, adjacent
  to item 2.
- **(7)** `--delta` is unusable for this repo's test layout (see the gate-of-record comment above).
  **A fleet-wide delivery cost, not a defect in this PR.**

Nothing in #4124 is a dropped acceptance criterion of #3815 — confirmed by the intent audit.

Issue #4038, opened 2026-09-04
#3811's raw_value path refuses duration trailing bytes that Cassandra 5.0.8 ACCEPTS (parity over-strictness)


pmcfadin:
**Proposed by the #3778 lane, per the lead ruling on `COORD-REQ:3778-03` ("file it as
`coord:follow-up-proposed` as you proposed — do **not** act on it in this lane"). Not groomed; no
priority, milestone or board Status set by me.**

## The finding

`#3811` made the `raw_value` path require exact consumption, and its `duration` arm therefore
**refuses trailing bytes inside a duration's declared extent**. Measured against the pinned oracle,
that is **stricter than Apache Cassandra**.

`cassandra-5.0.8:src/java/org/apache/cassandra/serializers/DurationSerializer.java`:

- `deserialize` reads three VInts (`readVInt32`, `readVInt32`, `readVInt`) and returns
  `Duration.newInstance(...)` — **no consumption check**.
- `validate` (`:80-105`) enforces only `size >= 3` (*"Expected at least 3 bytes for a duration
  (%d)"*), that months/days fit in 32 bits, and that all three components share a sign — **no
  consumption check**.
- `grep -n "remaining\|hasRemaining\|limit()"` over that file **and**
  `db/marshal/DurationType.java`: **zero hits**.

`TupleType.split` rule 4 (*"Expected N values … but got more"*) refuses trailing bytes at the
**tuple framing** level — not inside one element's declared length, which is the level this is about.

So a duration value that Cassandra reads happily is refused by CQLite on that path.

## Why this is filed rather than fixed here

Ruling on `COORD-REQ:3778-03` was **Option A / parity**: #3778 does **not** refuse duration trailing
bytes, and pins today's tolerance at the two other sites (`raw_type_value.rs`, `cell_value_scalar.rs`)
with characterisation tests citing this oracle. The lead's note on this residual, recorded verbatim
in substance:

> On #3811's raw-value path being stricter than Cassandra: file it as `coord:follow-up-proposed` as
> you proposed — do **not** act on it in this lane. You are right that it is the parity-pure subject
> (your option C) and right that it touches #3811's shipped tests, so it needs its own issue. Noting
> for whoever takes it: **the fix direction is *relax to match the oracle*, not *tighten the other
> two sites***.

## Scope for whoever takes it

- **Direction is fixed**: relax the `raw_value` duration arm to tolerate trailing bytes inside the
[...]

PR #4033, opened 2026-09-03
feat(#3805): a typed empty-buffer sentinel for the empty multicell map key (slice 1: representation)

pmcfadin:
## What this is

Issue #3805, **slice 1 of 2**. A non-frozen `map<K,V>` is multicell: each entry's KEY travels in
its cell's CellPath as `[VInt length][bare serialized key]`, so a **zero-length cell path means
the key's serialized form is the EMPTY buffer**. Cassandra accepts, stores and returns such a key;
CQLite refused it or fell back to an opaque blob.

This slice adds the representation and its semantics. **It does NOT wire the decoder** — see
"What this slice deliberately does not do".

## The owner ruling this implements

1. **The entry is PRESENT.** A zero-length cell path is not absence.
2. **The key is DISTINCT FROM NULL.** `empty ⇒ null` is scoped to the VALUE read path and does not
   transfer to a KEY (a null map key is illegal CQL; Cassandra returns a present `EmptyValue`).
3. Byte-exact round-trip and `BYTE_ORDER` ordering for the **declared type**.
4. A **distinct sentinel carrying the declared type**, distinguishable from both `Value::Null` and
   the type's zero value.

The lead first recommended reusing the typed variant with an empty payload, then **reversed itself**
after reading `types.rs`: the scalar variants hold *parsed scalars*, so an empty `int` could only
land as `BigInt(0)`, conflating EMPTY with the value `0`.

## The oracle (`docs/round-artifacts/issue-3805-cassandra-oracle.md`, 1225 lines)

Every claim cited to the pinned `cassandra-5.0.8` tag; no CQLite `file:line` is used as format
authority. Both authorities were used and **agree in every particular**.

What the sources DECIDE: the entry is present (the framing cannot express absence —
`writeWithVIntLength` → `writeUnsignedVInt32`, so a negative length is unrepresentable and is
declared corruption on read; `CellPath.create` asserts non-null), and the key is distinct from null
(`Int32Type.compareCustom:61-71` gives empty a UNIQUE sort position — measured on real bytes for
four key types; a null map key is illegal CQL per `Maps.java:342-343`, `:426-427`, `:510-511`).

What they DO NOT decide: which `Value` variant carries it. `AbstractType.allowsEmpty()`'s own
javadoc calls the affordance a testing API — legal but unintended. The sources FILTER candidates
(`Value::Null` ruled out by (2); `Value::Integer(0)` by collision) and leave both an opaque blob
and a typed sentinel admissible. The lead chose the sentinel.

[...]

PR #4017, opened 2026-09-03
fix(#3847): accept Cassandra's legal EMPTY buffer as null for every fixed-width scalar

pmcfadin:
Closes #3847

## What Cassandra does, and what CQLite did

`parse_value_from_raw_bytes` is a **READ** path, so its Cassandra analogue is
`TypeSerializer.deserialize()`, not `validate()`. On the `deserialize()` side the rule is **uniform,
12 of 12, no per-type exceptions**: an EMPTY buffer deserializes to `null`.

Full read at the pinned `cassandra-5.0.8` tag: `docs/round-artifacts/issue-3847-cassandra-oracle.md`
(`Int32Serializer:32`, `LongSerializer:32`, `BooleanSerializer:34`, `UUIDSerializer:33`,
`TimeUUID.Serializer:341`, `FloatSerializer:32`, `DoubleSerializer:32`, `ShortSerializer:32`,
`ByteSerializer:32`, `TimestampSerializer:139`, `SimpleDateSerializer:52`, `TimeSerializer:34`).
Corroboration from the write direction: `BooleanSerializer:41-43` — `serialize(null)` returns
`EMPTY_BYTE_BUFFER`, so empty **is** the wire spelling of null and a reader refusing it cannot read
data Cassandra legitimately writes.

**One correction to the issue's premise.** #3847 says the `|| isEmpty` shape "appears in the other
fixed-width serializers". Measured, that holds for `validate()` in **8 of 12** and fails in four —
`ShortSerializer` (`!= 2`), `ByteSerializer` (`!= 1`), `SimpleDateSerializer` (`!= 4`),
`TimeSerializer` (`!= 8`) reject empty outright. That does not change the answer, because
`validate()` gates what may be **written** and `deserialize()` defines what must be **read**. It does
dispose of acceptance bullet 3: the `udt.rs` `!= n` family needs no "why they differ" statement,
because on the read path there is nothing to differ about.

## The change

Composed accepted set before: `require_fixed_width` refused `len < n`, the arm reported consumed `n`,
`require_fully_consumed_raw` refused `consumed != len` ⇒ exactly **`{n}`**. Now **`{n, 0}`** with
`0 ⇒ null`.

The substance is the composition, not the width guard. **Relaxing `require_fixed_width` alone would be
a defect**: the arm would then index `data[0]` on an empty slice AND report consumed `n` against
`len == 0`, so the fully-consumed assert would refuse the value just admitted. The empty case
[...]

Issue #3847, opened 2026-09-01
Fixed-width scalars: CQLite's accepted set is {n}, Cassandra's is {n, 0} — the legal EMPTY buffer is refused on the parse_value_from_raw_bytes path


pmcfadin:
Split out of #3811 review (rust-reviewer I3). **Not a regression — the behaviour is identical to `main`; what #3811 changed is that the rule is now stated in one place and can be read.**

## The rule Cassandra applies

`cassandra-5.0.8:src/java/org/apache/cassandra/serializers/Int32Serializer.java:42-43`:

```java
if (accessor.size(value) != 4 && !accessor.isEmpty(value))
    throw new MarshalException(...);
```

The legal widths are **`{4, 0}`** — an EMPTY buffer deserializes to `null`, and the same `|| isEmpty` shape appears in the other fixed-width serializers. `cell_path_key.rs:740-787` already encodes the per-type table this way.

## What CQLite does on the `parse_value_from_raw_bytes` path

Two cooperating checks, and their COMPOSITION is what matters:

- `require_fixed_width` (`raw_value/reporting.rs`) refuses `data.len() < n`;
- the arm reports its exact width `n`, so `require_fully_consumed_raw` refuses `data.len() > n`.

Composed accepted set for a declared `int`: `len == 0` → `0 < 4` → `Err`; `len == 4` → `Ok`; `len == 5` → consumed 4 ≠ 5 → `Err`. That is **exactly `{4}`**, not `{4, 0}` — the legal empty value is REFUSED, by the under-width half.

`udt.rs`'s two scalar field decoders (`parse_udt_field_value`, `parse_simple_udt_field_value`) use a direct `!= n` and refuse empty as well, so the two families agree here; both differ from Cassandra in the same direction.

## Why it was not fixed in #3811

Accepting an empty buffer is a **behaviour change on real corpus data** (a previously-refused cell would start decoding to `null`), and #3811's subject is consumption, not width admissibility. Widening it there would have been smuggled in under an unrelated oracle and would not have been covered by that PR's before/after corpus measurement, which measured *no change*.

## Acceptance

- Decide, against `Int32Serializer` and its siblings, whether the empty buffer should decode to `null` on this path.
- If yes: accept `{n, 0}` in `require_fixed_width` **and** make the arm report `data.len()` when empty, so the consumption check still holds; run the 144-file corpus before/after census (a table that gains rows is the expected direction here, and must be explained).
[...]

PR #4126, opened 2026-09-06
docs(#3723): the AC2 prose had decayed into being false — correct it, and pin the counts so it cannot happen a sixth time

pmcfadin:
## What this is

Issue #3723's **AC2 documentation half** — the last item of its own scope. The mechanism shipped long ago (#3811, and #3847 for the zero-length disposition); what remained was that this issue's own prose had decayed into being **false**, and was citing a test that does not exist.

**This PR deliberately declares NO closing reference for #3723.** Whether that P1 closes is a lead call, and the `inet` gap below needs a tracker first (see "Not closing" at the bottom).

## The defect

`raw_value/nested_fixed_width_length_tests.rs` — shipped by #3723 itself — asserted in the present tense that CQLite is **"NARROWER than Cassandra"** on the zero-length case and that the divergence is **"tracked as #3847"**. **#3847 merged** (`a5171a5ba`, PR #4017) and removed that fact: the accepted set at the bounded surface is now `{n, 0}`, not `{n}`. A second passage asserted a bounded-vs-UDT-field **asymmetry** that the same change removed. Both cited `zero_length_fixed_width_element_is_refused_at_every_nesting_position` — **a test that exists nowhere in the tree.**

The root cause is worth stating, because it is the fifth instance on this issue: **the claim was true when written and false by the time it shipped.** #3847 landed underneath it. Same mechanism as the `break` claim the C intent audit caught, where the falsifying merge was this branch's own.

## What changed — prose and pinning assertions only

**No production source. No existing assertion's meaning changed.** 3 files, +138/-49.

1. **The false claims corrected** — the zero-length note now states the current rule (`{n, 0}` at the value positions), credits #3847/`a5171a5ba` in the past tense, and cites only tests that exist. The `validate()`-vs-`deserialize()` distinction is the load-bearing argument: the table is the WRITE rule and non-uniform, this is a READ path whose oracle is `deserialize()`, uniform 12-of-12 per `docs/round-artifacts/issue-3847-cassandra-oracle.md` at the pinned tag.
2. **A count that was off by one, found in review** — `..._at_the_four_value_positions` runs **three** positions: its loop skips both `key` and `set<`. The first draft of this fix *propagated that bad count into the header*. Renamed to `..._three_value_positions` after **measuring** it with an instrumented counter rather than trusting the review.
3. **The counts are now pinned, not hand-maintained** — both zero-length cases `assert_eq!` their exact position set **by name**, and the two sets **partition** `nesting_positions` (3 + 2 = 5). So the "five" that used to be prose is now a test that breaks. The assertions were verified **discriminating** by failing them against an empty expectation; previously either loop could have skipped every position and passed green.
4. **Four hand-maintained censuses deleted, not corrected** (12 rows / 13 names / 15 spellings / 11 arms). They had already decayed twice (r10, r20/job 113). Per #3229 — *remove the mechanism rather than carve it a fourth time* — they are replaced with universals, pointers, or spelled names, following the pattern this module already demonstrated: **"The names are spelled rather than counted."** The one surviving number is admissible only because an assertion pins it.
5. **A dead function name in three files** — `create_empty_value_for_type` was cited in the **present tense** and exists nowhere. The role is performed today by `parse_simple_udt_field_value_at` (`typed_value.rs:159`), called with an explicit `&[]` by all five UDT framing sites. A branch whose whole purpose is deleting a citation to a nonexistent test must not ship a citation to a nonexistent function.

## Verification

| stage | result |
|---|---|
| `--lite` (final, `f937cd339`) | **`RESULT: PASS`** — `tree-integrity: PASS`, `dirty: no`, 5/5 components, `scoped-tests` verified **3900 tests / 171 binaries** |
| rust-reviewer r1 (`9bf2383b8`) | 2 blockers, 2 important, 1 nit — **all fixed** |
| roborev **job 134** (`9bf2383b8`) | `RESULT: FAIL`, `findings: PRESENT (1)` — **independently the same blocker** as the reviewer's B2. Fixed |
| rust-reviewer r2 (`5f15071c6`) | all five prior items **FIXED**, counts re-derived from source, **no blockers**; 2 important + 3 nits — all fixed |
| roborev **job 135** (`5f15071c6`) | `RESULT: PASS`, **`findings: NONE`**, `code-free: PASS`, `sha-assert: PASS`, vacuity tier1+tier2 PASS, `prompt-content: PASS (2/2)`, input 80102 tokens (genuine band, not the ~18.7k vacuous baseline) |

**Job 135 does NOT certify this merge** — three commits postdate it. The binding roborev round runs **last**, after the gate of record, per #3752.

Three round-2 confirmations worth recording, each a way this fix could have been quietly wrong: `len in 1..n` at `n == 1` (`tinyint`/`boolean`) is **vacuously true, not false**; the over-width refusal is correctly attributed to the caller's `require_fully_consumed` rather than to `require_fixed_width`, since `admissible_at_least` returns `Bytes` for any `len >= n`; and **no fourth answer exists** — the offset-advancing decoders cannot express an empty fixed-width value, so "three independent answers" is exhaustive rather than merely where someone stopped counting.

`--delta` is **ineligible** here and was not re-probed: a `#[cfg(test)] mod` inside the lib classifies `[PRODUCTION]`, which this issue measured on its previous slice. A full gate is the only route.

## AC7 re-verified against a moving base

`origin/main` advanced three times during this work. `66e9957cc` (#3805 slice 1, PR #4033) touched **`complex_column/cell_path_key.rs` — AC7's own evidence file**. Re-checked: AC7 **still holds**. #4033 added only an empty-buffer (`len == 0`) admission returning `Value::Empty(tag)` ahead of the decode; it changed no consumption or trailing-byte rule, and composites and `duration` resolve `for_cql_type == None` so they keep their outcome. All 14 tests that paragraph names are grepped-present.

## Not closing, and one gap that needs an owner

- **Site 6** (a shadow/TTL-dropped element is `continue`d before decode) is **not** #3723 scope. It is carried by open **#3778**, and the shipped tests say so in their own identifiers (`..._known_gap_3778`). Correction to #3778's closing census: it locates Site 6 at `complex_column/element_decode.rs`, **which does not exist on `origin/main`** — the sites are `complex_column.rs:542/:620/:769/:901`.
- **The `inet` width/empty divergence** (Cassandra 5.0.8 admits `4|16`; this decoder admits any length) is declared and characterised but **tracked by no issue at all**. It goes dark the day #3723 closes. Proposed as a follow-up in `COORD-REQ:3723-11`; filing at finalize.

Refs #3723. Refs #3847, #3811, #3778, #4038.

### DIFF
--- a/cqlite-core/src/storage/sstable/reader/parsing/row_decoder/raw_value/fixed_width.rs
+++ b/cqlite-core/src/storage/sstable/reader/parsing/row_decoder/raw_value/fixed_width.rs
@@ -65,11 +65,14 @@
 //!
 //! A UDT field's length header decides whether a field is absent (`-1`), present
 //! and empty (`0`) or present with bytes. Five sites route the `0`:
-//! `parse_udt_value` and `raw_type_value.rs`'s UDT arm through
-//! `create_empty_value_for_type`, and `parse_nested_udt_from_registry`,
-//! `parse_inline_udt_value` and `raw_type_value.rs`'s registry arm by calling a
-//! field decoder with an explicit `&[]`. Before #3847 the first two answered
-//! "empty BLOB" and the other three answered `Err`; all five now answer `null`.
+//! `parse_udt_value`, `parse_nested_udt_from_registry`, `parse_inline_udt_value`
+//! and `raw_type_value.rs`'s two UDT arms (marshal/inline and registry-resolved).
+//! All five hand it to the SAME delegate today,
+//! `typed_value.rs::parse_simple_udt_field_value_at`, with an explicit `&[]`, and
+//! all five answer `null`. They did not always: before #3847 `parse_udt_value` and
+//! `raw_type_value.rs`'s marshal/inline arm routed through
+//! `create_empty_value_for_type` — a function since DELETED (#3631/PR#3820) — and
+//! answered "empty BLOB", while the other three answered `Err`.
 //! An `Err` there was the worse of the two, because `row_data.rs` `break`s its
 //! column loop on a failing column, so the failing column AND every later one
 //! silently became null.
--- a/cqlite-core/src/storage/sstable/reader/parsing/row_decoder/raw_value/issue_3847_empty_fixed_width_tests.rs
+++ b/cqlite-core/src/storage/sstable/reader/parsing/row_decoder/raw_value/issue_3847_empty_fixed_width_tests.rs
@@ -239,14 +239,14 @@ fn an_empty_duration_is_still_refused_declared_residual_of_3847() {
 // `typed_value.rs::parse_simple_udt_field_value_at`, with an explicit `&[]`; what
 // differs is the dispatch that reaches it, so a loop widened in one arm and not the
 // other is exactly how rounds 1 and 2 of this review found real defects.
+// Kept end-to-end rather than helper-level on the wiring-evidence rule: green
+// helper-only unit tests are not sufficient.
 //
 // (An earlier revision of this note named `create_empty_value_for_type` as the
 // marshal arm's route. No such function exists any more — #3631/PR#3820 replaced it,
 // and the per-type rule it used to carry now lives once, in
 // `typed_value/scalar_rules.rs::empty_is_a_value`. The two-route reasoning above is
 // unchanged; only the name was stale.)
-// Kept end-to-end rather than helper-level on the wiring-evidence rule: green
-// helper-only unit tests are not sufficient.
 // ---------------------------------------------------------------------------
 
 /// `pair(a int, b int)` in MARSHAL form. Field names are hex: `70616972` =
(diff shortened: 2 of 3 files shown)


pmcfadin:
Final roborev confirmation round for PR #4126, run **after** the gate of record so the review binds to the certified tree (#3752 endgame order: rebase -> gate -> C -> roborev -> premerge-assert -> arm).

Certified head: `cb0c2441dd52b92447977467b47156d5047d80b8` (gate of record: full `AGENT-GATE SUMMARY`, 39/39 components PASS, `tree-integrity: PASS`).

Prior rounds on this branch (job 134 FAIL/1 finding fixed, job 135 PASS at `5f15071c6`) do **not** certify this merge -- three commits plus a rebase postdate 135.

```
repo: /data/lanes/lane-3723
branch: issue-3723-ac2-prose-decay
base: origin/main
head-sha: cb0c2441dd52b92447977467b47156d5047d80b8
reviewed-sha: ad47385cb607e5c554d4e0713984883b4e207738..cb0c2441dd52b92447977467b47156d5047d80b8
assert-base: ad47385cb607e5c554d4e0713984883b4e207738 (merge-base of origin/main and HEAD; origin/main tip 1cc12e28a2f2edb9f7375ddd70c8845af13d35f0)
job: 136
model: gpt-5.6-sol
census: 3 files, +138/-49
tokens: input=81091 cached=48896 output=654
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (3/3 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: NONE
roborev-exit: PASS
log: /tmp/roborev-3723-ac2-final.log
RESULT: PASS
```

Issue #2662, opened 2026-07-17
gate.yml nightly deep-check dead all July: 75m timeout kills every run (12 straight cancels) — raise/split + alert on red


pmcfadin:
## Problem — the nightly full-gate backstop has been dead ALL of July

`gate.yml` ("Nightly agent-gate (deep check)") has not produced a green run in the visible 20-run history:

- **Jul 6 → Jul 17: 12 consecutive `cancelled`** — the job hits `timeout-minutes: 75` (`.github/workflows/gate.yml:60`) and is killed ~76 min in (e.g. run 29558896251: started 06:05, cancelled 07:21). The comment at line 65 even names this failure mode.
- **Jul 1 → Jul 5: `failure`** — pre-timeout-era reds, cause unknown (needs triage as part of this fix; may be long-fixed or may still lurk under the timeouts).

Impact: the standing backstop CLAUDE.md relies on ("the nightly gate.yml deep-check re-runs the FULL gate on main") — including the `CQLITE_CLIPPY_FULL=1` full clippy matrix that no PR lane runs — has silently not certified main since June. The delta-gate doctrine (#1892) leans on this backstop existing.

## Root cause (proximate)

The full gate outgrew 75 minutes on a standard GitHub runner during the 0.14/0.15 cycles (locally it is 15–25 min warm; GH runners are cold-cache, 2-core, and run the full component set incl. python/node bindings). Nobody noticed because a `cancelled` conclusion doesn't page anyone — silent-death mode.

## Acceptance criteria

1. Nightly gate.yml completes with a real conclusion (success/failure — never timeout-cancelled) on consecutive nights.
2. The pre-Jul-6 `failure` causes triaged: either shown fixed on the first green run, or filed as their own issues.
3. **A red/cancelled nightly is no longer silent** — the run must surface (e.g. a workflow-failure step that opens/updates a pinned issue, or equivalent), so a dead backstop can never again go unnoticed for weeks. (This is the load-bearing part; the timeout bump is trivial.)
4. Timing headroom documented in the workflow (measured runtime + chosen timeout).

## Fix directions (implementer's call, verify with a manual dispatch before relying on the schedule)

- **(a) Immediate**: raise `timeout-minutes` to realistic (measure first — likely 150–240 min cold), dispatch manually to get one true end-to-end timing + a real verdict.
[...]

Issue #1360, opened 2026-07-01
Epic: CI runtime overhaul — local-first validation, light PR gate, nightly deep checks


pmcfadin:
## Why this epic

CQLite CI is spending too much time and queue capacity on ordinary PRs because a single PR can fan out into core CI, M1 CI, SSTable parity, compression parity, performance, observability, Node, Python, Flight, Trino, E2E, smoke, coverage, and minimal-feature workflows at the same time.

The target model is local-first validation plus one light enforceable PR gate, backed by nightly deep validation and release gates. CI should confirm the validation model, not discover every basic formatting, Clippy, Rust build, or workflow-structure problem after push.

## Grounding / best-practice baseline

- Keep required PR checks small, stable, and always-running. Do not globally require path-filtered workflows that can be skipped and remain pending.
- Use concurrency/cancel-in-progress for stale PR runs.
- Use path filters and labels for expensive targeted checks, but do not let them become broad global gates.
- Use deliberate Rust cache strategy; avoid broad `target/` cache plus `sccache` everywhere without a measured reason.
- Use build-once/run-many test sharding, such as `cargo-nextest archive`, for Rust test partitions.
- Move full OS/arch/native/package matrices to nightly/manual/release unless the PR directly changes that surface.
- Measure queue-to-green, workflow count per PR, p95 job time, cache hit rate, and dataset download count before and after.

## Implementation plan

Detailed plan file: `docs/superpowers/plans/2026-07-01-ci-runtime-overhaul.md`

### Local pre-merge validation

Primary engineering signal before merge approval. Maintained local script should cover formatting, `cqlite-core` Clippy with `-D warnings`, core build/tests, and change-specific deep checks.

### Required PR CI

One light always-running required check. It should include workflow validation, Rust fmt, `cqlite-core` Clippy with `-D warnings`, all-feature core build, and fast representative tests.

### Targeted PR checks

Path-specific or label/manual checks for binding smoke, docs, workflow policy, minimal features, and focused storage/parity smoke.

### Nightly deep checks

Full SSTableDump parity, regeneration parity, Cassandra ingest paths, all-table smoke, full Node/Python/Flight/Trino matrices, coverage, performance, and observability overhead.

### Release gate

Full parity, full matrices, coverage/perf, package/image publish dry-runs, and release publishing.

## Child issues / implementation order

- [x] #1361
- [x] #1362
- [x] #1363
- [x] #1364
- [x] #1365
- [x] #1366
- [x] #1367
[...]

PR #4129, opened 2026-09-06
docs(#4128): the 27 hours are after the code is written — SDLC audit, intent, and slice 1 spec

### LINKED ISSUE #4128: AI-native SDLC: audit, intent, and slice 1 (scoped --lite clippy)
## Why

Measured over 590 deliveries in `docs/reports/delivery-telemetry.jsonl`: median cycle time is **34.2 hours**, of which roughly **27 hours** is spent after the code is written. The gate passes **98.5%** of the time, so it rarely rejects anything, while **71.4%** of PRs need rework at a mean of **3.32 rounds** and **6.92 roborev findings**.

Each rework round pays an inner loop that is not scoped to the change. `run_clippy` (`scripts/agent-gate.sh:11193`) is a whole-workspace matrix and never reads the diff, and `run_lite` calls that same function at `:20546`.

## What this issue lands

Artifacts only. No behaviour change.

1. `docs/development/ai-native-sdlc-redesign.md` — the full audit behind this, measured against the AI-native SDLC playbook.
2. `openspec/intent/active/ai-native-sdlc.md` — the intent, with 7 slices named.
3. `openspec/changes/lite-clippy-scoped-to-blast-radius/` — slice 1 fully specified (proposal, design, tasks, delta spec for a new `inner-loop-cost` capability).

Implementation of slice 1 is a separate issue, filed after this merges and after 0.17 lands.

## Corrections this records

- `CLAUDE.md`'s Lite row cites `:17233` and `:18220` as two clippy dispatch sites. There is one definition at `:11193` called by both modes. Both citations have drifted.
- `run_lite`'s comment says "FULL-workspace clippy"; the banner it prints says "scoped workspace clippy."
- A pure blast-radius scope would break invariant **#1893** (`scripts/agent-gate.sh:11214`): `cqlite-py` must stay linted because `--lite`'s python tier reports SKIP on a toolchain failure and this clippy pass is the compile backstop that makes the SKIP safe. Slice 1's spec scopes to `blast_radius ∪ FLOOR` and pins it with its own mutation test.

## Explicit non-goal

Relocating the gate of record into CI. `openspec/specs/gate-ci-coverage/spec.md:22` already requires that the full gate is not a required per-PR check, deliberately, so it does not duplicate the light `pr-gate.yml` from epic #1360. That decision stands.

## Acceptance

- [ ] The three artifacts are committed and readable
- [ ] Every requirement in the delta spec carries at least one Given/When/Then scenario (`openspec/config.yaml` spec rule)
- [ ] The diff is docs-only per `scripts/ci/classify-docs-only.sh`

pmcfadin:
Closes #4128

## Summary

Artifacts only. **No behaviour change**, no code, no workflow edits. `scripts/ci/classify-docs-only.sh` returns `docs-only`.

Measured over 590 deliveries in `docs/reports/delivery-telemetry.jsonl`:

| Measure | Value |
|---|---|
| Median cycle time, open to merge | 34.2 h |
| Median issue to PR open | ~4 h |
| Median time in review | ~27 h |
| PRs needing rework | 71.4% (421/590) |
| Mean rework rounds | 3.32 |
| Mean roborev findings | 6.92 |
| Gate pass rate | 98.5% |

The gate rarely rejects anything. The review loop is where the time goes, and each round pays an inner loop that is not scoped to the change.

## What lands

1. **`docs/development/ai-native-sdlc-redesign.md`** — the audit: nine read-only sweeps over 362,265 lines of delivery machinery, measured against the AI-native SDLC playbook.
2. **`openspec/intent/active/ai-native-sdlc.md`** — the intent, 7 slices named, with the constraints it accepts as given.
3. **`openspec/changes/lite-clippy-scoped-to-blast-radius/`** — slice 1 specified: proposal, design, tasks, and a delta spec for a new `inner-loop-cost` capability with 6 requirements, every one carrying Given/When/Then scenarios per `openspec/config.yaml`.

## Primary-source verification

A code-free diff cannot be roborev-certified, so per doctrine the substitute is primary-source verification, recorded here. Every claim below was checked against `scripts/agent-gate.sh`, not against `CLAUDE.md`.

- **`run_clippy` is ONE definition** at `scripts/agent-gate.sh:11193` — a four-stage whole-workspace matrix that never reads the diff. `run_lite` calls that same function at `:20546`. `CLAUDE.md`'s Lite row cites `:17233` and `:18220` as two dispatch sites; **both have drifted and the two-site framing is wrong.** The substance of its claim survives; the citations do not. Slice 1 corrects the row rather than inheriting the error.
- **`run_lite` contradicts itself**: comment at `:20529` says "FULL-workspace clippy", the banner it prints at `:20539` says "scoped workspace clippy". Recorded for correction in slice 1.
- **Invariant #1893 at `:11214` would have been broken** by the obvious design. `cqlite-py` must stay in the linted set because `--lite`'s python tier reports SKIP on a venv/pip/maturin failure, and this clippy pass is the compile backstop that makes that SKIP safe. A pure blast-radius scope drops `cqlite-py` from most diffs and turns a safe SKIP into a silent hole. The spec therefore scopes to `blast_radius ∪ FLOOR`, and pins the invariant with its own requirement and its own mutation test.
- **`run_clippy` is already a narrowing**, justified the same way: its header at `:11186-11192` records that the historical `--all-features` pass runs nightly in a dedicated `clippy-full` job (#2662). Scoping `--lite` by blast radius extends an existing argument rather than introducing a new concession.

## Explicit non-goal, recorded rather than quietly reversed

**Relocating the gate of record into CI.** An earlier draft of the audit proposed it. `openspec/specs/gate-ci-coverage/spec.md:22` already requires that the full gate is *not* a required per-PR check, deliberately, so it does not duplicate the light always-running `pr-gate.yml` from epic #1360. That decision stands, and the intent lists the reversal as out of scope.

## Slice 1 is falsifiable on purpose

Task 1.3 is a decision point *before* any behaviour change. If the baseline shows a `cqlite-core/src/` diff's blast radius is substantially the whole workspace, scoping saves nothing on the class that hurts most, and the measurement artifact must say so. Task 1.1b measures `FLOOR` alone, since its cost is paid on every lite run and therefore bounds the best case.

## Open questions for the owner

Three sit at the end of the intent, all needed before slice 3 or slice 7 starts:

1. Slice 3 sets a review nit cap. What number?
2. Slice 7 deletes `.github/issue-monitoring/` (2,321 lines, no CI job invokes it). Delete, or wire it?
3. Slice 6 batches the telemetry stamp. Is a nightly batch commit acceptable, given `delivery-telemetry.py retro` is the only consumer?

## Wiring evidence

Not applicable. No public surface changes; these are specification and documentation artifacts. The behaviour they specify lands in a separate issue after 0.17.

### DIFF
--- a/docs/development/ai-native-sdlc-redesign.md
+++ b/docs/development/ai-native-sdlc-redesign.md
@@ -1,0 +1,177 @@
+# AI-Native SDLC Redesign for CQLite
+
+Date: 2026-09-04
+Method: nine parallel read-only audits of the whole SDLC surface, measured against the AI-native SDLC playbook (https://claude.com/blog/the-ai-native-sdlc-playbook).
+
+## The finding in one paragraph
+
+CQLite has already solved the build problem.  The median time from issue to open pull request is about 4 hours.  The median time from open to merge is 34.2 hours.  So review and certification take 27 of those 34 hours.  The repo pays for that delay with 362,265 lines of process machinery, and the machinery does not earn it: the gate of record passes 98.5% of the time, so it almost never catches anything, while 71.4% of pull requests still need rework.  The rework comes from the review layer, not the gate.
+
+## Measured surface
+
+| Area | Size | Source |
+|---|---|---|
+| CI workflow YAML | 13,160 lines, 42 files | `.github/workflows/` |
+| `.github/` non-workflow | 5,168 lines | `.github/` |
+| Shell and Python scripts | 249,178 lines, 249 files | `scripts/` |
+| `scripts/agent-gate.sh` alone | 23,837 lines, 39 components | `scripts/agent-gate.sh:6264` |
+| `scripts/tests/` | 149,644 lines, 135 files | `scripts/tests/` |
+| `.claude/` config | 14,830 lines, 27 skills | `.claude/` |
+| OpenSpec artifacts | 79,929 lines, 108 specs | `openspec/` |
+| `CLAUDE.md` | 687 lines | `CLAUDE.md` |
+| Happy-path agent instructions | 2,333 lines | `CLAUDE.md` + `worker.md` + `flow-*` + `flow-lead`/`flow-closer` |
+| ALL-CAPS imperatives in those files | 146 | `rg` over the same set |
+
+## Measured delivery, from 590 records
+
+Source: `docs/reports/delivery-telemetry.jsonl`.
+
+| Measure | Value |
+|---|---|
+| Median cycle time, open to merge | 34.2 hours |
+| Median issue to pull request | ~4 hours |
+| Median time in review | ~27 hours |
+| Pull requests needing rework | 71.4% (421 of 590) |
+| Mean rework rounds per pull request | 3.32 |
+| Mean roborev findings per pull request | 6.92 |
+| Gate pass rate | 98.5% (581 pass, 8 fail) |
+| Telemetry share of last 200 commits | 44% (88 commits) |
+
+## Diagnosis
+
+### The gate is the constraint, and it is serialized at one
+
+`bootstrap-agent-machine.sh` persists `CQLITE_GATE_MAX_CONCURRENCY=1` (`docs/development/pm-operating-loop.md:146`).  The lite gate runs 1.4 minutes warm on a narrow diff, 20 minutes median on a `cqlite-core/src/` change, and up to 104 minutes under load.  The full gate runs 45 to 120 minutes.  One box therefore certifies two to five changes an hour at best, no matter how many agents run.  19 worktrees exist; a single author drives them.  The worktrees queue, they do not run in parallel.
+
+### Merge authority sits on a laptop, not in CI
+
+24 workflows fire on a pull request.  One is registered as merge-gating (the `flight` tier).  The other 23 are declared exempt in `.github/ci-gating-tiers.yml:156-518`, each deferring its "merge-gating half" to a component of the local `agent-gate.sh`.  GitHub has thousands of parallel runner-minutes available and the design declines them.  Meanwhile CI reruns the same work advisorily: `clippy` in 7 workflows, `cargo fmt` in 4, `cqlite-core` tests in 9.
+
+### There is no fast inner loop
+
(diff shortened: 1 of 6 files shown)


pmcfadin:
## Gate route: CITE-AND-WAIVE, docs-only

Recording this rather than leaving it implicit, since `premerge-assert.sh` was not satisfied and that is a deliberate choice, not an oversight.

**`premerge-assert.sh 4129 <sha>` returned `PREMERGE: USAGE`** — it requires a full `AGENT-GATE SUMMARY` as its third argument. That precondition exists for code diffs (#3465).

**Route taken instead:** the CITE-AND-WAIVE waiver, per `openspec/specs/docs-only-gate-classification/spec.md:430`. Its three conditions:

1. **Genuinely prose, established mechanically not by judgement.** `git diff --cached --name-only | bash scripts/ci/classify-docs-only.sh` → `docs-only`, `classify-docs-only: all changed files are docs -> short-circuit to green`, exit 0.
2. **A cited issue.** #4128.
3. **No compiled input in the diff, which would void the waiver.** Six `.md` files. None under `docs/reports/*-artifacts/` — that is the #3222-shaped falsifying case the narrowed rule names, where `src/main.rs` and `Cargo.toml` live under a `docs/` path and satisfy the old qualifier textually while being false materially. This diff has no such path.

**CI agrees with the classifier:** `required` PASS in 19s, `pr-gate-core` PASS, every code lane `skipping`.

**What this establishes and what it does not.** It establishes that no compiled input changed, so no gate component could have read anything in this diff. It does **not** establish that the specification these artifacts describe is correct — that is what slice 1's own gate run, its `spec-auditor` pass, and its four mutation tests are for, on the implementation issue.

**Condition of change**, mirroring the backfill ruling's structure at `spec.md:492`: if any path in this change is later promoted into a shipped surface — a gate component, a CI step, an imported module — the exemption ends at that moment, because it inherits the obligations of the surface it joins.

Roborev was not run: a code-free diff cannot be roborev-certified at all. The substitute is the primary-source verification recorded in the PR body, where every `agent-gate.sh` line number was checked against the script rather than taken from `CLAUDE.md` — which is how the three drifted citations were found.

Issue #4070, opened 2026-09-04
One field decoder, not two: parse_simple_udt_field_value_at still duplicates ~10 scalar arms of parse_typed_value, and the Blob-vs-empty-Udt outcome for an unresolvable UDT name is undecided


pmcfadin:
Follow-up from #3722 (filed at the lead's request on that thread: *"the two-decoder → one-decoder consolidation, with the Blob→empty-`Udt` behaviour decision stated as the open question"*). #3722 closed the **silent degradation**; this issue is the **residual duplication** it deliberately left, plus the one behaviour decision that has to be made on purpose rather than as a side effect.

## What #3722 already closed (do not redo)

`parse_udt_field_value` and its ~140-line dispatch are **gone**. All five field loops — `udt.rs::parse_udt_value`, its registry-resolved decoder, `udt/inline.rs`'s inline decoder and `raw_type_value.rs`'s two marshal-string loops — now route through the single per-field entry point `parse_simple_udt_field_value_at` (`row_decoder/typed_value.rs:159`), which threads `depth`, goes through the one exhaustion assert, and returns an explicit `Error` naming a type it cannot decode instead of a `Value::Blob`.

## The residual: `parse_simple_udt_field_value_at` still carries its own scalar arms

That function keeps ~10 arms — `Text`/`Ascii`, `Int`, `BigInt`, `Boolean`, `Float`, `Double`, `Uuid`/`TimeUuid`, `Timestamp`, `Blob` — **byte-for-byte as they were before #3722**, and only its `other =>` arm delegates to `parse_typed_value` (`typed_value.rs:272`). So two decoders still cover one scalar set, which is the "a type fix must land in N places or the paths silently diverge" shape.

Its own doc comment states the delta and defers it explicitly (`typed_value.rs:145-149`):

> The scalar arms below are kept BYTE-FOR-BYTE as they were before #3631 (they encode measured, shipped behaviour — e.g. `CqlType::Float` surfaces `Value::Float32`, which the type-string decoder spells as a widened `Value::Float`; unifying the two is a separate change).

**That is this issue.** Deleting the arms and routing the whole field type through `parse_typed_value` is a small diff; deciding what it is allowed to change to the OUTPUT is the work.

## The open question — decide DELIBERATELY, do not let it fall out of a refactor

Two measured behaviour deltas, both currently load-bearing for somebody:

[...]

PR #4138, opened 2026-09-07
fix(#4070): one field decoder, not two — and a refusal that only claims what it knows

pmcfadin:
**Refs #4070** — deliberately NOT a closing reference. AC1, AC2, AC4 and AC5 land here; **AC3 was descoped by lead ruling on `REQUEST-4070-02` (option 4)** and lands in **#4141**, so #4070 must stay open on merge.

`parse_simple_udt_field_value_at` carried ~10 scalar arms duplicating `parse_typed_value`. This deletes the duplicate arms, pins the outcomes that were never pinned, and leaves a labelled tripwire where AC3's refusal briefly stood.

## Two of the issue's premises did not survive measurement

**The FLOAT delta was STALE — there was nothing to consolidate.** The issue's decision 1 rests on `typed_value.rs`'s own doc comment claiming the type-string decoder "spells [`float`] as a widened `Value::Float`". Both decoders already returned `Value::Float32`; the widening died in `b2132fb2c` (#3811), whose *pre-image* already carried `Float32`, i.e. it was removed inside #3811's review **after** #3722 wrote the comment describing it. So the comment was true when written and false by the time this issue was filed.

Kept `Float32`, justified on format authority and explicitly **not** on CQLite's prior output (#3041/AC2): the committed sstabledump golden `test_timeseries/sensor_data-…/nb-1-big-Data.db.jsonl` renders `humidity 92.88221` and `temperature -16.172066` at f32 precision beside `pressure 1017.9518806690071` at f64, against a schema declaring `FLOAT, FLOAT, DOUBLE` (`test-data/schemas/time-series.cql:20-22`). Cassandra's own renderer distinguishes the widths, so widening is a measurable parity break — `0.1f32` renders `0.1` vs `0.10000000149011612`. AC2's deliverable was therefore a **missing pin**, not a change: nothing pinned the float spelling for a UDT *field*.

**The stated user-visible consequences are REFUTED.** The issue says the `Blob`/empty-`Udt` choice *"changes hashability in Python (the #3500 class) and the Arrow/Trino column type"*. Neither holds: `Value::Blob → PyBytes` and a zero-field `Udt` are both hashable (`bindings/python/src/value.rs:71`, `:96`, `:357`; the #3500 class ran the *other* way — a `Blob` masking an *unhashable* field), and the Arrow `DataType` is `CqlType`-derived, never `Value`-derived (`export/arrow_schema.rs:198`, `cqlite-flight/src/producer.rs:1263`). The real axis runs opposite to the framing: `arrow_builders_nested.rs:294-311` **fails closed** on a non-`Udt` under a UDT-typed column, so an `Error` merely moves that failure earlier, while an empty `Value::Udt` would make export newly *succeed* with an all-null struct — a silent-data-loss direction, and the strongest argument against the option the issue floated.

**Structural corrections.** AC1 and AC3 are orthogonal diffs in different functions — `parse_simple_udt_field_value_at` never calls `parse_raw_type_value`. **This is why AC3's descope costs AC1 nothing:** the reverted code is in `raw_type_value.rs`, which AC1 does not touch and does not call. And `raw_type_value.rs:454`, cited in the issue as a third `Frozen element` fallback, **is not one**: it is the tail of the legitimate `"blob" | "bytes" =>` arm, a *declared* blob that genuinely is VInt-framed. Left alone.

## AC3 — DESCOPED to #4141, with the measurement it produced kept

AC3 was implemented, measured, reviewed twice, and then **descoped by lead ruling on `REQUEST-4070-02` (option 4)**. It will land with **site C** as one coherent change in **#4141**. The code is reverted to `origin/main`; what follows is the reasoning that produced the descope, recorded because it is the most expensive thing this issue produced and it must not evaporate with the code.

**Why descoped, not shipped.** roborev flagged, in two independent rounds, that the trailing `_ =>` arm refuses **any** unrecognised type string and not only an unresolved UDT reference, while other routes still blob the same type. Fixing that properly requires converging `raw_value/reporting.rs`'s trailing `other =>` arm ("site C") — which the **top-level column path** also uses. That is a materially wider blast surface than AC3's own wording scopes to, and too wide to take on a release-close night. Splitting AC3 from site C would ship exactly the inconsistency roborev identified; shipping them together is one coherent change, which is #4141.

**The zero-hit corpus measurement.** Measured over **147 `*-Data.db`** (144 corpus + 3 committed Cassandra-written UDT fixtures) in both a registry-absent and a registry-present pass: **all three sites fired ZERO times.** No instrumentation was needed — all three already `tracing::debug!`.

That zero is trustworthy because it was **proved capable of being non-zero**: the **containing arms executed 303 times** with their sibling log lines captured, out of 716k + 3.6M `row_decoder` DEBUG lines. The sites were evaluated and not taken. An earlier sweep *was* a false zero (passing table directories to `read-sstable` errors `Is a directory` and reads 0 rows) and was caught and discarded. Only 3 distinct bare type strings reach the arm at all — `address_type`, `contact_info`, `wide` — and every one resolved. Mechanism: registry-absent reads take full marshal class names from the `SerializationHeader`, which earlier arms match, so the arm is reachable only via bare CQL names, i.e. only under `--schema`. **This measurement carries over to #4141 unchanged** and is why the behaviour change there is low-risk on real data.

**Why shape-based discrimination is NOT available at that site.** `udt.rs:145`'s `is_udt_type` accepts only a case-insensitive `org.apache.cassandra.db.marshal.usertype` substring, and `parse_raw_type_value`'s arm at `:639` already consumes every string it accepts **before** the catch-all — so at the refusal site it is **constant `false`**. A branch worded off it would be dead code while every real case (e.g. `frozen<list<address_type>>` with a registry lacking `address_type`, reached via the `list<` recursion) took the other branch. And a bare short UDT name is **not distinguishable** from an unrecognised non-UDT marshal string by any other shape test either — `address_type`, `EmptyType`, `VectorType(FloatType,3)`, `Int32Type` are all just "a string no earlier arm matched", and `parse_raw_type_value` has no marshal→short normalizer (unlike `reporting.rs:99`). **A positive marshal-prefix test was considered and REFUSED as a #28 violation**: it still infers a type's nature from its spelling, which is the pattern #28 exists to prevent, and it is not a proof either.

**Also still true, and load-bearing for #4141:** empty `Value::Udt` is ruled out (it fabricates a type — which `qualified_reference_to_unknown_udt_does_not_resolve` exists to forbid — and un-refuses Arrow export, per the refutation above). And **site A is not constructible through the CLI**: four RED-arm attempts were each refused upstream by the schema loader (undeclared UDT → `Column 'w' references undefined UDT 'wide'`; undeclared collection *element* → validation descends; late declaration → resolves anyway; cross-keyspace qualified → `get_udt_qualified` resolves it too). So any test of this state is **decoder-level by necessity** — including the characterization test below — rather than a faked fixture for a state the loader forbids.

## What changed

- **AC1** — `typed_value.rs` 704→634: the function is now a depth check plus one `parse_typed_value` delegation.
- **AC2** — new pin `a_float_udt_field_decodes_to_float32_not_a_widened_double`.
- **AC5** — the `nu.b` `Value::Text` pin inside `assert_wide_fully_decoded`, through the public `Database::execute` surface over the **Cassandra-written** `issue_3722/test_udt_wide_fields` fixture.
- **AC4/collateral** — three artifacts asserted the pre-consolidation world and **nothing would have caught them**: a dead LIVE CONTROL in `nested_fixed_width_length_tests.rs` (re-purposed into a consolidation witness), that file's `DELEGATED_UDT_FIELD_TYPES` header (which says of itself *"CURATED, NOT DERIVED … an arm moved into or out of `parse_simple_udt_field_value_at` is not detected by anything below"*), and `an_oversized_fixed_width_field_is_still_refused`, which post-consolidation would have survived only via its `|| contains("4 bytes")` accident — tightened to the exact declared-width wording.
- **#4141 tripwire (new)** — `characterization_unresolvable_frozen_element_still_degrades_to_blob_pending_4141` in `regression_2807_qualified_udt_decode_tests.rs`. The descope restores the blob degrade, so a test asserting the refusal would assert a behaviour that no longer happens, and *no* test would leave the descoped behaviour entirely unpinned. This one **records** today's outcome instead: `Ok((Value::Blob(..), _))` with `[VInt len][bytes]` framing, for **three populations** — (a) registry present / name absent, (b) **no registry at all**, a branch that has never had a test of any kind, and (c) an unrecognised **non-UDT** marshal string (`EmptyType`), the population roborev was actually worried about and which nothing has ever pinned — plus a CONTROL proving the arm is not simply blobbing everything. Its doc block states that this is **measured, NOT endorsed** behaviour (the blob degrade is what #28 forbids and what #3631/#3722 removed next door), that **#4141 is the subject and this test SHOULD go red when #4141 lands** — forcing whoever converges these sites to observe the change rather than discover it in production — and why #4070 did not fix it. All four assertions were **run** against the reverted code and hold.
- **AC3 code — REVERTED** to `origin/main` (`raw_type_value.rs` and the two AC3 tests in `regression_2807_qualified_udt_decode_tests.rs`), verified hunk-by-hunk so no non-AC3 work was dropped with it.

## Deliberate deviation (declared, not discovered)

The `data.is_empty()` redirect was **folded into** the single delegation. Post-AC1 it read `if empty { return parse_typed_value(args) } parse_typed_value(args)` — a branch whose two sides are byte-identical. Behaviour is unchanged (identical call, identical arguments) and the `empty_is_a_value` rationale is preserved verbatim as a comment at the same site.

## Declared gaps — stated, not papered over

1. **6 of the 10 consolidated arms have no end-to-end coverage.** The `wide` fixture carries only `int`, `blob`, `timeuuid` and `text`; `bigint`, `boolean`, `float`, `double`, bare `uuid` and `timestamp` appear in **no committed UDT fixture anywhere**. Covered by decoder-level tests plus the recorded arm-by-arm equivalence only.
2. **AC2 is a decoder-level pin, not a fixture-backed parity oracle** — no committed fixture or schema has a CQL `float` inside a UDT. **No CQLite-written fixture was invented**: per #3042 a CQLite-written + CQLite-read round-trip is invariant to a uniform framing error and can never be the oracle. The honest fix is a Cassandra-written `wide2`, proposed as a follow-up.
3. **Format authority rung 1 was unavailable in this lane** (no Cassandra clone; `CQLITE_CASSANDRA_REPO` unset), so no `git show cassandra-5.0.8:` quote is claimed. AC2 rests on rung 2, stated in the test rather than fabricated.
4. **`DELEGATED_UDT_FIELD_TYPES` remains CURATED, not derived** — a *sixth* fixed-width `CqlType` appearing is still undetected. What #4070 removed is the need to track *which arm* a type lands on.
5. **AC3 is UNMET here, and #4070 stays open.** The two `Frozen element` sites still return an opaque `Value::Blob` for a declared-but-undecodable type — a #28 degrade, now explicitly pinned by the characterization test rather than left silent. #4141 fixes it together with **site C** (`raw_value/reporting.rs`, the trailing `other =>` of `parse_value_from_raw_bytes`), which is also used by the **column** path.
6. **`raw_type_value.rs` is 1037 lines**, ~237 over the ~800 source target. This diff no longer touches that file at all. The next change to it — #4141 — will have to split it (epic #1116); named rather than left as a trap.

## Review record — two independent reviews, and what they decided

**roborev round 1** (job 466, codex/gpt-5.6-sol) returned one Medium: the catch-all arm handles *any* unrecognised type string, not only UDT references, so the new refusal both refused inputs that previously blobbed and **misdiagnosed them as an unresolved user-defined type**. The first half is the ruled behaviour change (zero-hit measurement + #28). The second was a real defect: a wrong diagnosis on a fail-closed path.

**The first proposed fix was wrong and was withdrawn** — it would have discriminated with `is_udt_type`, which is constant `false` at that site (see above), inverting the misdiagnosis rather than removing it. The replacement was **one** wording asserting only what the site can establish (the name did not resolve), with the operator-facing cause kept as data. **Round 2 flagged the same breadth concern again**, which is what escalated it to `REQUEST-4070-02` and the descope: the wording was fixable, but the *scope* of what the arm refuses is not fixable without site C. That is the finding that ruled this issue, and it is #4141's starting point.

**Second blocker (round 1):** the new helper's doc had copied the issue's claim about Python hashability and the Arrow/Trino column type — the premise refuted above. Deleted rather than softened. A false claim in a header is worse than an acknowledged gap: the next reader checks it, finds it false, and discards the correct rationale with it. (That helper is reverted with AC3; the refutation stands on its own above.)

**AC1 was independently confirmed behaviour-preserving** by an arm-by-arm re-derivation: no divergence in returned variant, admissible width set, error-vs-lenient disposition, borrow posture (`borrow_active` on both sides), `Boolean` truthiness, `Timestamp` units, `Uuid`/`TimeUuid` collapse or blob framing. Only message strings differ. Consumption is a tautology on the delegated path; depth is checked on entry and threaded outward with no reset at any UDT boundary, so a cyclic `UdtRegistry` still terminates; `parse_vuint`/`checked_vuint_length` remain in use with nothing orphaned.

Also corrected in review: three assert messages had dropped `\` continuations, rendering as gaps in the very failure output they exist to produce — plus **two pre-existing instances of the identical defect** fixed under the campsite rule (both in retained files). And a 97-line AC5 case was collapsed to the one assertion it actually added, after review found `assert_wide_fully_decoded` already pinned `bl`/`i`/`tu` as full-`Value` comparisons; its header claim that the end-to-end variant pins did not exist was false and is gone.

## Verification

`--lite` PASS at each fix round — `dirty: no`, `tree-integrity: PASS`, run-id matching the launch every time. Re-certified with `--lite` after the AC3 revert + tripwire. The retained work was re-run directly against the reverted tree: `regression_2807` (5 passed, including the new tripwire), `regression_3631` (71), `nested_fixed_width_length_tests` (10), and `issue_3722_udt_field_type_fidelity` (8, `--features cli-helpers`). The full `AGENT-GATE SUMMARY` (the only run that counts) is run once, immediately pre-merge.

## Follow-ups
- **#4141 — AC3 + site C as one change.** Converge `raw_type_value.rs`'s two `Frozen element` sites *and* `raw_value/reporting.rs`'s trailing `other =>` on one refusal, so no route still blobs a type another route refuses. The tripwire test above goes red and is deleted there. Carries over from this issue: the 147-file zero-hit measurement, the constant-`false` `is_udt_type` finding, the #28 refusal of a marshal-prefix test, and that site A is not CLI-constructible.
- **A Cassandra-written `wide2` UDT fixture** carrying `bigint`, `boolean`, `float`, `double`, bare `uuid`, `timestamp` — closes declared gaps 1 and 2 together, and is what would let a `float` UDT field be pinned by a fixture instead of a decoder test.
- **`issue_3722_udt_field_type_fidelity.rs:29`'s `#![cfg(feature = "cli-helpers")]`** means `cargo test -p cqlite-core --test issue_3722_udt_field_type_fidelity` reports `running 0 tests`, so verifying AC5 with CLAUDE.md's documented core command yields a **vacuous green** (re-confirmed this round). It does run under the feature (8 passed) and in the gate. Pre-existing, not introduced here.


pmcfadin:
🔴 **HELD — see `REQUEST-4070-02` on #4070 for the ruling I need.** roborev returned the same Medium twice (jobs 466 and 467); the misdiagnosis half is fixed, the ruled behaviour-change half remains. Critically, `docs/development/roborev-contract.md:211-224` constraint (c) makes the absence waiver **absence-only**, and this is `findings: PRESENT` — so there is no mechanism to merge past it. Resolving needs either a scope widening (converge site C) or a descope (AC3 out of this PR), and both are the lead's call. Auto-merge is deliberately NOT armed and the full gate of record has deliberately NOT been run, since any of those routes voids it. Do not merge this PR until that request is answered.

pmcfadin:
```
repo: /data/lanes/lane-4070
branch: issue-4070-one-field-decoder-not-two
base: origin/main
head-sha: eaea4c29d863eb8718cacf014f86dacee119a04c
reviewed-sha: 21a8dc23056c8421f10916e01941eddff672904f..eaea4c29d863eb8718cacf014f86dacee119a04c
assert-base: 21a8dc23056c8421f10916e01941eddff672904f (merge-base of origin/main and HEAD; origin/main tip 21a8dc23056c8421f10916e01941eddff672904f)
job: 468
model: gpt-5.6-sol
census: 5 files, +341/-155
tokens: input=193747 cached=144768 output=926
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (5/5 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: NONE
roborev-exit: PASS
log: /tmp/roborev-review-issue-4070-one-field-decoder-not-two-eaea4c29-1118433.log
RESULT: PASS
```

pmcfadin:
```

==== AGENT-GATE SUMMARY ====
run-id: /tmp/agent-gate.silJLk
commit: eaea4c2 branch: issue-4070-one-field-decoder-not-two dirty: no
schemas: 8/8 canonical .cql readable under /data/lanes/lane-4070/test-data/schemas (checkout-relative)
component-set: PASS (39/39 names vs origin/main 21a8dc23056c8421f10916e01941eddff672904f; NAMES ONLY — not implementations, and no component is run here) — baseline read via the committed manifest; objects: baseline REUSED from this lane's SHARED store — store TRUSTED, not verified (#3749)
accelerators: sccache=on nextest=on lanes=on sccache-health=ok sccache-cap=53687091200 sccache-used=22562195025(42%) mold=linked perf=ok
tree-start: eaea4c29d863 dirty: no digest: a319c0c24ce2
tree-end: eaea4c29d863 dirty: no digest: a319c0c24ce2
tree-integrity: PASS
census: 24/39 components AFFIRMED a count; 15 DECLARED-GAP (RECOGNISED); 0 NOT-MEASURED (RECOGNISED); 0 measured-ZERO (RECOGNISED); 0 not-applicable (component did not PASS); 0 no-subject (PASSed; the run had nothing to measure); 0 UNDECLARED; 0 unrecognised; 0 row(s) carry a VACUOUS status. NON-EXHAUSTIVE: the gap set is CURATED, so an unaffirmed component is UNMEASURED, never verified (#3625; the remaining gaps are tracked in #3162).
fmt:               PASS (8s)  [fmt workspace features=n/a]  {no census — cargo fmt --all --check emits no per-file tally to count}
clippy:            PASS (4s)  [clippy workspace(excl 5) --all-features | clippy cqlite-core --features 25:all-compression,arrow,bench-internals,+22 more | clippy cqlite-cli --features 8:benchmarks,cli-helpers,delta-export,+5 more | clippy cqlite-flight+cqlite-py+cqlite-node --features cqlite-node/write-support]  {no census — cargo clippy emits a per-crate tally only COLD; a warm run prints Finished alone}
roborev-lints:     PASS (144s)  [no-cargo]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
core-tests:        PASS (1184s)  [nextest run cqlite-core --features cli-helpers | test cqlite-core --features cli-helpers]  {verified: 5931 tests passed (across 2 result line(s))}
tombstones-scan:   PASS (49s)  [test cqlite-core --features write-support,cli-helpers,tombstones]  {verified: 2 tests passed (across 1 result line(s))}
scan-offload-guard: PASS (181s)  [test cqlite-core --features cli-helpers,scan-offload-probe x2]  {verified: 15 tests passed (across 8 result line(s))}
work-counters-guard: PASS (112s)  [test cqlite-core --features write-support,cli-helpers,state_machine,work-counters]  {verified: 60 tests passed (across 15 result line(s))}
arrow-parity-guard: PASS (45s)  [test cqlite-core --features arrow]  {verified: 4 tests passed (across 1 result line(s))}
memory-budget:     PASS (304s)  [test cqlite-core --features cli-helpers,dhat-heap,arrow x3 | test cqlite-flight --features dhat-heap]  {verified: 6 tests passed (across 4 result line(s))}
integration-tests: PASS (1s)  [test cqlite-integration-tests default-features x2]  {verified: 42 tests passed and 25 test binaries built/verified}
format-compat:     PASS (1s)  [test format-compatibility-tests default-features]  {verified: 10 tests passed (across 1 result line(s))}
write-tests:       PASS (172s)  [test cqlite-core --features write-support x3]  {verified: 4050 tests passed (across 3 result line(s))}
cli-tests:         PASS (49s)  [test cqlite-cli default-features | test cqlite-cli --features write-support]  {verified: 868 tests passed (across 49 result line(s))}
compaction-byte-parity: PASS (10s)  [test cqlite-core --features write-support x2]  {verified: 12 tests passed (across 4 result line(s))}
bti-multiclustering: PASS (5s)  [test cqlite-core --features state_machine,cli-helpers x2]  {verified: 19 tests passed (across 3 result line(s))}
flight-query-semantics-oracle: PASS (0s)  [test cqlite-flight default-features]  {verified: 4 tests passed (across 2 result line(s))}
flight-tests:      PASS (113s)  [test cqlite-flight default-features]  {verified: 424 tests passed (across 2 result line(s))}
legacy-heuristics: PASS (286s)  [build cqlite-core --features legacy-heuristics]  {verified: 3944 tests passed (across 6 result line(s))}
feature-iso-parquet: PASS (224s)  [test cqlite-core --no-default-features --features all-compression,parquet]  {verified: 1 test binaries built/verified}
feature-iso-delta-scan: PASS (216s)  [test cqlite-core --no-default-features --features all-compression,write-support,delta-scan]  {verified: 3618 tests passed (across 14 result line(s))}
python-bindings:   PASS (588s)  [via maturin: feature set NOT observed]  {verified: 786 pytest tests passed}
node-bindings:     PASS (482s)  [via npm run build (napi): feature set NOT observed]  {verified: 602 jest tests passed}
binding-rust-tests: PASS (243s)  [test cqlite-ffi-common default-features | test cqlite-node --features write-support]  {verified: 123 tests passed (across 5 result line(s))}
parity-report:     PASS (15s)  [run cassandra-parity default-features]  {no census — xtask/report driver prints no machine-readable subject count (#3162)}
operator-metrics-doc: PASS (28s)  [run cqlite-core default-features]  {no census — xtask/report driver prints no machine-readable subject count (#3162)}
dep-duplicates:    PASS (1s)  [via check-dep-duplicates.sh (cargo tree -d --workspace --target all; a metadata probe, compiles nothing): feature set NOT observed]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
features-load-bearing: PASS (5s)  [cargo not observable: cargo metadata runs inside scripts/ci/check-features-load-bearing.sh (a child process); it is a metadata read, not a build, and enables no features]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
tooling-tests:     PASS (3235s)  [test ws0-corpus-gen default-features | + cargo not observable: cargo may run inside ~60 nested test scripts (child processes)]  {no census — shell/python guard prints no AGENT-GATE-CENSUS contract line yet (#3162)}
minimal-build:     PASS (112s)  [build cqlite-core --no-default-features --features all-compression | test cqlite-core --no-default-features --features all-compression]  {verified: 1 test binaries built/verified}
all-features-check: PASS (250s)  [check cqlite-core --all-features | clippy cqlite-core --all-features]  {no census — cargo check/clippy passes execute no tests; the subject is a feature set, not a count}
smoke:             PASS (369s)  [build cqlite-cli default-features]  {no census — smoke-test-all-tables.sh prints no machine-readable table count (#3162)}
disk-admission: PASS (evaluated 2x: launch 237.4GiB; post-slot 237.4GiB; bar 40GiB(default); fs /data; target-dir /data/lanes/lane-4070/target (via cargo metadata)) — re-checked AT SLOT GRANT, not merely at launch (#3755); subjects 1 RECOGNISED (the cargo-resolved BUILD-OUTPUT filesystem only) — NON-EXHAUSTIVE: other filesystems this gate writes to (the python venv, bindings/node/node_modules) are NOT measured (#3886)
logs: /tmp/agent-gate.silJLk
logdir-disposition: REMOVED at exit on PASS; AGENT_GATE_KEEP_LOGS=1 retains
logdir-sweep: 0 REMOVED of 0 aged (>7d) under /tmp (owner verified-dead 0, live 0, unverifiable 0; examined 0, removals attempted 0, declined on identity re-check 0, cap 1000)
summary-file: /tmp/gate-4070-rec2.txt
heartbeat: on file: /tmp/gate-4070-rec2.txt.heartbeat interval: 20s
RESULT: PASS
```

Issue #1020, opened 2026-06-24
Add UDT and frozen value compaction parity scenarios


pmcfadin:
Parent epic: #973

## Goal

Ensure compaction preserves Cassandra-compatible bytes and typed values for UDTs, nested UDTs, frozen collections, and frozen UDT values.

## Cassandra coverage

`CQLSSTableWriterTest`, `LegacySSTableTest`, `SerializationHeaderTest`, and `SerializationMirrorTest` scenarios involving frozen UDTs, nested UDTs, collections with UDTs, and schema-driven type decoding.

## Manifest entries

- `cass.sstable_format.CQLSSTableWriterTest.frozen_udt_roundtrip`
- `cass.sstable_format.LegacySSTableTest.complex_udt_frozen_non_frozen`
- `cass.serialization.SerializationHeaderTest.udt_schema_resolution`
- `cqlite.compaction_parity.udt.frozen_person`
- `cqlite.compaction_parity.udt.nested_udt`
- `cqlite.compaction_parity.udt.collections_with_udts`

## Evidence type
- canonical_semantic
- byte_for_byte

## Artifacts compared
- bytes
- component files
- JSONL
- generated report

## CI tier
- required_parity
- nightly_docker

## Acceptance criteria

- Add deterministic Cassandra fixtures for `frozen<list<int>>`, `frozen<map<text,int>>`, `frozen<person>`, nested UDTs, and collections containing UDT values.
- Include DDL and UDT definitions with each scenario so CQLite cannot pass by rendering blob fallback.
- Compare raw value bytes and canonical typed `sstabledump` JSONL after compaction.
- Verify null UDT fields, field ordering, nested field decoding, and frozen value boundaries survive compaction.
- Fail loudly when schema registry or UDT definition resolution is missing.
- Record which cases are byte-for-byte clean and which remain canonical semantic only.

## Out of scope

- Vector types.
- Counters.
- Cassandra legacy versions outside supported fixture matrix.
- Collection cell-path ownership covered by the complex collection issue.

pmcfadin:
🧭 **MANAGER** 
GO
ORDER: 12

Elevated P1→P0, promoted to Ready. UDT + frozen-value compaction parity scenarios. Oracle-driven → skip OpenSpec; pin byte-parity tests.

pmcfadin:
[...]

PR #4139, opened 2026-09-07
fix(#4104): refuse frozen<scalar> at both metadata entry points — it is not declarable CQL, so CQLite must not invent decode semantics

### LINKED ISSUE #4104: frozen<scalar> (e.g. frozen<int>) is not declarable CQL — refuse it at schema-parse and header-parse instead of inventing decode semantics (settles the #3805/#3847 Empty(Int)-vs-Blob collision)
## Summary

`frozen<int>` (frozen wrapping a **scalar**) is **not declarable CQL**. At the pinned tag:

```
$ git show cassandra-5.0.8:src/java/org/apache/cassandra/cql3/CQL3Type.java | sed -n '647,651p'
        public Raw freeze()
        {
            String message = String.format("frozen<> is only allowed on collections, tuples, and user-defined types (got %s)", this);
            throw new InvalidRequestException(message);
        }
```

`CQL3Type.Raw.freeze()` is the base implementation; only `RawCollection`, `RawTuple` and `RawUT` override it. So a Cassandra table can never carry a `frozen<scalar>` column, map key or element, and a Cassandra-written `Statistics.db` serialization header can never spell `FrozenType(Int32Type)`. **There are no Cassandra-written bytes for this input, by construction.**

CQLite currently lets the spelling through: two lanes (#3805 / PR #4033 and #3847 / PR #4017) each pinned a *decode result* for an empty `map<frozen<int>,int>` key — `Value::Empty(Int)` and `Value::blob(b"")` respectively — and the pins are mutually exclusive. Both assert **invented semantics for an impossible input**, which is the no-heuristics class (#28): where Cassandra has no behaviour, CQLite must not choose one.

## Fix

Refuse `frozen<T>` where `T` is not a collection, tuple or UDT — at **both** authoritative-metadata entry points:

1. the CQL schema parser (`frozen<int>` in a supplied `.cql`) → `Error::UnsupportedType`/schema error citing `CQL3Type.java:649`;
2. the serialization-header type parser (`FrozenType(<scalar>)` in `Statistics.db`) → the same refusal, fail-closed.

A refused type is never reached by `cell_path_key_*`, so the "peeled probe" tolerance those helpers grew for frozen scalars becomes dead and should be removed with a structural test that `frozen<scalar>` never reaches them.

## Acceptance

- Two pinned refusals (schema parse, header parse), expectation derived from `CQL3Type.java:647-651`, never from CQLite's prior behaviour.
- The two contradictory pins in #4033 / main are DELETED (ruling on PR #4033, `REQ-3805-14`), not "fixed" to either value.
- No decode path gains a frozen-scalar branch.

Filed from the #3805/#3847 collision ruling. Oracle-driven (the oracle is Cassandra's own grammar), so: issue + pinned test, no OpenSpec.

### LINKED ISSUE #4158: WRITE PATH: CQLite emits BytesType in Statistics.db for a frozen<UDT> column instead of FrozenType(UserType(...)) — hidden for months by the reader fallback #4104 closed
## The defect

CQLite's SSTable writer emits **`BytesType`** in `Statistics.db`'s serialization-header column types for a **`frozen<UDT>`** column, instead of the real marshal type `FrozenType(UserType(ks,name,…))`.

That is a **wrong on-disk metadata type written by CQLite**. It is a parity defect, not a reader defect.

## Why it went unnoticed, and why it surfaced now

This is **#3042's blind spot made concrete** — the one that says a CQLite-written + CQLite-read round-trip cannot catch a uniform error, because both sides make the same mistake:

1. The writer emitted an invalid header type.
2. The reader's **heuristic marker-search fallback** recovered anyway.
3. `write_read_roundtrip.rs`'s `type_coverage::test_type_frozen_udt` therefore stayed **green** for however long this has been true.

It surfaced only because **#4104 closed that fallback**. Its Gate 2 header refusal is *semantic*, and the fail-open where a semantically-refused header was retried by the marker search was fixed in `3499c9463` (by lead ruling on roborev job 116, finding 1 — non-negotiable, because the fallback could return `Ok` with empty or unrelated schema data). With the fallback closed, the writer's own bad type is refused, and the round-trip fails.

**The gate found a real write-path bug. Closing the fail-open did not create it; it revealed it.** The reader's refusal is correct and must not be weakened to accommodate this — narrowing Gate 2 to tolerate it was explicitly considered and rejected, because accommodating a defect is how the fail-open got there in the first place.

## Evidence

`type_coverage::test_type_frozen_udt`, `cqlite-core/tests/write_read_roundtrip.rs:533`, on branch `issue-4104-frozen-scalar-not-declarable` at `cd44bc52b`:

```
assertion `left == right` failed: Expected exactly 1 row in test_types.test_frozen, got 0
  left: 0
 right: 1
```

Reproduced in two independent gate components (`core-tests` 0/1, `write-tests` 128/1) in run `/tmp/agent-gate.36NDKJ`.

## Acceptance criteria

1. The writer emits the **real marshal type** for a `frozen<UDT>` column in `Statistics.db`'s serialization header — `FrozenType(UserType(ks,name,…))` — with the exact spelling derived from the pinned **`cassandra-5.0.8`** source, not from CQLite's own reader (#3041: a CQLite `file:line` is never format authority).
2. **A Cassandra-written oracle.** Compare CQLite's emitted header bytes for a `frozen<UDT>` column against a header Cassandra itself wrote for the same schema. **A CQLite-write→CQLite-read round-trip is not acceptable evidence for this AC** — that invariance is precisely what hid this bug (#3042).
3. `type_coverage::test_type_frozen_udt` passes **without** any reader-side tolerance being reintroduced. If it passes because a fallback came back, the fix is wrong.
4. **A RED control**: the new assertion fails against the current writer, proving it discriminates.
5. Audit the **sibling types** in the same emit path and state which were checked — a wrong type for one class is rarely alone. Name what was checked, not just what was fixed.
6. `sstabledump` accepts the resulting SSTable, and the compaction-parity logical tier stays green (this touches the write path, which carries parity claims).

## Related

* **#4104** / PR #4139 — closed the reader fail-open that hid this. Its refusal is correct; do not weaken it.
* **The companion swallow defect** (filed alongside this one) — why this presents as *silent* data loss rather than an error. **That one is the more urgent of the two**: it is what turns a correct refusal into a silently empty scan.
* **#3042** — the round-trip-invariance blind spot this is an instance of.
* **#28** — the no-heuristics mandate; the fallback that hid this was a violation of it.

pmcfadin:
Closes #4104

## What and why

`frozen<T>` where `T` is a **scalar** is not declarable CQL, so **no Cassandra-written bytes for it
exist, by construction**. CQLite let the spelling through, and two lanes each pinned a *decode result*
for an empty `map<frozen<int>,int>` key — mutually exclusive pins, both inventing semantics for an
impossible input. That is the no-heuristics class (#28): where Cassandra has no behaviour, CQLite
must not choose one.

This refuses the spelling at both authoritative-metadata entry points, upstream of decode, and
deletes both pins.

## The oracle

`CQL3Type.Raw.freeze()` at `cassandra-5.0.8:CQL3Type.java:647-651` throws
`"frozen<> is only allowed on collections, tuples, and user-defined types (got %s)"`.

**The accepted set is derived from the OVERRIDE SET, not from that message — and the distinction is
load-bearing.** Four subclasses override `freeze()`, not the three the issue body names:

| Site | Evidence |
|---|---|
| `CQL3Type.java:596` | `public abstract class Raw` |
| `CQL3Type.java:647-651` | base `freeze()` throws |
| `CQL3Type.java:885` | `class RawVector extends Raw` |
| `CQL3Type.java:908-911` | `RawVector.supportsFreezing()` → `true` |
| `CQL3Type.java:916-919` | `@Override public Raw freeze() { return this; }` |

The base error **message is stale relative to its own code** — it never mentions vectors even though
`RawVector` overrides. A predicate derived from that string would wrongly **reject** a legal
`frozen<vector<float,3>>`: the same no-heuristics defect this PR removes, pointed the other way
(inventing a *prohibition* Cassandra does not impose). So the accepted set is collection / tuple /
UDT / **vector**, and `frozen<vector<float,3>>` has a positive test.

## Changes

1. **Schema parser** — `frozen<scalar>` refused, error citing `CQL3Type.java:649`.
2. **Serialization-header type parser** — `FrozenType(<scalar>)` refused, fail-closed.
3. **Dead tolerance removed** — a refused type never reaches `cell_path_key_*`, so the "peeled probe"
   frozen-scalar tolerance is deleted, with a structural test that `frozen<scalar>` never arrives.
4. **Both contradictory pins DELETED**, not set to either value (ruling on PR #4033, `REQ-3805-14`).
5. **No decode path gained a frozen-scalar branch** — the refusal sits in metadata parsing.

## Review status

- **roborev job 108** (`gpt-5.6-sol`): `RESULT: FAIL`, 1 finding — `is_vector_spelling` matched only
  the head keyword before `<`, so `frozen<vector<int>>`, `frozen<vector<>>`,
  `frozen<vector<int, nope>>` were wrongly accepted as freezable. Triaged **blocker** (a permissive
  fall-through in a gate whose acceptance criterion is fail-closed) and fixed in this branch:
  exactly two args via the parent module's own `split_top_level_types` depth walk, leading/trailing/
  doubled commas refused, dimension digits-only + `i32` + `> 0`. Digits-only is load-bearing —
  `str::parse::<i32>` accepts a leading `+` that `Lexer.g:337-339`'s `INTEGER` does not, and the `-`
  spellings `VectorType.java:89-90` throws on. Derivation chain recorded in-source.
  roborev's broader remedy (introduce a structured `CqlType::Vector` variant) was **declined as
  out of scope** — a new public-surface type variant breaks 1:1:1:1; raise separately if wanted.
- Rebased onto `origin/main` **before** the gate of record, because `scripts/agent-gate.sh` itself
  moved on main and a gate script behind `origin/main` cannot certify. Done before gating and before
  the binding roborev round, so it voids neither (#3752).
- Gate of record and the final binding roborev round: results posted below as they land.

Closes #4158 — the write-path `Statistics.db` type defect, folded into this PR by owner ruling (2026-09-07 ~09:25 PDT) so the record shows the writer fix shipped in 0.17.

---

## DECLARED: a deliberate read refusal on headers spelling `frozen<scalar>` — 0.17 release note

This PR **refuses to open** an SSTable whose `Statistics.db` SerializationHeader spells a frozen
scalar (e.g. `FrozenType(Int32Type)`, or `FrozenType(BytesType)` as CQLite itself used to emit for a
registry-less `frozen<udt>` column — the #4158 defect fixed in this same PR). Declaring it because an
undeclared behaviour change reads as a regression to the next person.

**This is the fix, not a compatibility regression.** `frozen<scalar>` is not declarable CQL —
`CQL3Type.Raw::freeze()` throws for every type that does not override it
(`cassandra-5.0.8:CQL3Type.java:647-651`) — so such a header can only have been produced by a
writer defect. Refusing to read a file that should never have existed is the point of #4104.

### Measured blast radius on Cassandra-written data: ZERO

Not asserted — measured, and the numbers are the declaration:

| Measurement | Result |
|---|---|
| Cassandra-printed `FrozenType(...)` forms accepted by the read gate | **78 RECOGNISED**, **0 refused**, 0 truncated |
| `*-Statistics.db` files walked (both candidate corpus roots) | **190** |
| `smoke-test-all-tables.sh` enforced tables | **51/51 pass** |

The guard that produced those numbers is committed
(`the_header_gate_never_refuses_a_frozentype_cassandra_wrote`), so the claim is re-derivable rather
than a one-off observation. Corpus census of `FrozenType(` inner heads is MapType 25 / ListType 16 /
UserType 10 / SetType 9 — **never a scalar, in any Cassandra-written file we hold.**

### What this does NOT change

**The CQL gate is unchanged, so `frozen<vector<float,3>>` remains declarable.** The two gates answer
different questions and are easy to conflate: the CQL gate decides *freezability* (`RawVector`
overrides `freeze()` and returns `this` — `CQL3Type.java:915-919`, so a frozen vector is legal),
while the marshal gate decides which types Cassandra can *print inside a `FrozenType(` wrapper*
(exactly four — `ListType.java:201`, `SetType.java:191`, `MapType.java:316`, `UserType.java:442`).
Narrowing the second did not narrow the first.

### The diagnosability claim, kept to its actual size

What is proven is narrow and specific: **this** refusal's error names the offending type,
`includeFrozenType`, and `CQL3Type.java:647-651`, and `StatisticsReader::open` returns
`Error::Schema` unchanged rather than relabelling it as `Corruption`. Before this PR the same
condition surfaced as `Corruption: … Error { input: [0x…], code: Verify }` — shaped identically to a
truncated or garbled file.

That is **not** a general claim that metadata refusals are now user-diagnosable. The generic case —
any metadata refusal on any SSTable making a scan return `Ok(empty)` instead of `Err` — is **#4159**,
which this PR is held behind.

### Affected files and the remedy

An SSTable **written by CQLite ≤ 0.16** for a `frozen<udt>` column with no `UdtRegistry` supplied at
write time. Remedy: rewrite it with cqlite ≥ 0.17, whose writer emits Cassandra's real
`FrozenType(UserType(..))` (this PR, #4158). A compatibility escape hatch, if wanted, is a follow-up
PR and not a change to this one.

## Residual findings, filed rather than absorbed

Discoverable from this merge; none blocks it:

- **#4163** — `cql_type_references_udt`'s `_ => false` makes a UDT nested in a `CqlType::Vector`
  element invisible to the registry dispatch.
- **#4164** — the writer's value path is type-blind, so vector value bytes stay length-prefixed while
  the header now declares a fixed-element `VectorType`. Filed P2 with reachability explicitly
  unestablished; **P1 if a vector `Value` can reach the writer**.
- **#4165** — `list<frozen<udt>>` UNDER-emits the wrapper Cassandra writes: the **mirror of #4158**
  (over-emission), same #1020 top-level-only dispatch scoping. Corpus-attested in
  `test_collections/collections_with_udts` and `test_types/cx_nested_frozen_collections`.
- **#4153** — the `#1714` reachability walker miscounts braces from quoted Java/ANTLR in a module
  header, so a late top-level `mod` reads as a false orphan. Routed around here, not fixed.
- **#4154** — remaining frozen-scalar positions. Note its originally-declared gap
  (`frozen<vector<frozen<int>, 3>>`) is **closed by #4149**'s element recursion, so #4154 wants
  re-scoping.

---

## ACCEPTED COST: this refusal is REACHABLE but NOT FULLY SURFACED at the scan boundary

Recorded because it is the accepted cost of deferring #4159 to 0.18, and it must not be discovered
later.

With #4159 deferred to 0.18, this PR's new refusal is **reachable but not fully surfaced** at the
scan boundary: `scan()` still converts a metadata refusal into `Ok(vec![])` (#4159), so a caller
using `scan()` sees **zero rows rather than an error**.

What **this** PR does guarantee is that `StatisticsReader::open` returns `Error::Schema` naming the
offending type, `includeFrozenType` and `CQL3Type.java:647-651`, and no longer relabels it
`Corruption` — so the refusal is **diagnosable** rather than indistinguishable from disk corruption.

Full surfacing lands with **#4159** (PR #4162, already CLEAN).

Shipping a partially-surfaced refusal is a deliberate, recorded tradeoff. Shipping one silently
would not be.


pmcfadin:
## Pre-gate review round: `rust-reviewer` — **no blockers, five nits**

Nits are batched into ONE linked follow-up at merge time per the lead's standing rule (no re-verify
round, no fix round below blocker severity). Recording the substance here so the disposition is
traceable rather than implicit.

### The four acceptance questions, answered

| # | Question | Answer |
|---|---|---|
| 1 | Both contradictory pins genuinely **DELETED**? | **Yes — deleted, not altered.** #4033's `a_frozen_spelled_fixed_width_empty_key_reaches_the_tag` (`Value::Empty(Int)`) and #4017's `an_empty_frozen_spelled_fixed_width_key_is_also_preserved_opaquely` (`Value::blob(b"")`) exist nowhere as code; each file carries only a prose deletion record (`regression_3747_empty_map_key_tests.rs:227-247`, `cell_path_key_tests_frozen.rs:31-37`). Nothing pins either value. |
| 2 | Any **decode path** gain a frozen-scalar branch? | **No.** The gate's only call sites are `schema/cql_type_parser.rs:119-127` and `enhanced_statistics_parser/marshal_type.rs:117`; nothing under `row_decoder/**`, `value_parsing*` or `complex_column/**` imports `frozen_scalar`. Decode paths *lost* three type-string peels and gained nothing. |
| 3 | Tolerance dead, and the new test **structural**? | **Both yes.** All three removed peels traced: every legal frozen inner classifies identically peeled vs unpeeled. The test drives the two *producers* (21×3 CQL spellings, 21×3 marshal spellings) and asserts **refusal** — no decode outcome at all — with case floors, an over-refusal counterpart, a real module declaration so it cannot pass by not compiling, and dataset-free so it cannot pass vacuously. |
| 4 | **Over-refusal** on any legal spelling? | **None found**, traced through the implementation rather than trusting the tests: frozen list/set/map/tuple, UDT (incl. the all-lowercase `frozen<address>` form that takes a different branch, and `frozen<ks.address_type>`), `frozen<vector<float,3>>` across whitespace/case/leading-zero/nested/comma-bearing forms, `frozen<list<frozen<udt>>>`, `frozen<frozen<list<int>>>`, and on the header side `ReversedType(FrozenType(SetType(..)))` — the DESC frozen-collection clustering spelling. |

Also checked: no reachable `unwrap()`/`expect()`/indexing panic in the new library code; all refusals
are `Error::schema` on the crate's `thiserror` enum, propagated with `?`; every fallible header call
site fails **closed**; format authority is Cassandra-pinned throughout with no CQLite `file:line`
used as an oracle.

### The five nits (follow-up, not this PR)

1. **The two gates don't compose for the one accept-set member with no corpus witness.** Gate 2 admits
   `FrozenType(VectorType(FloatType,3))`, but `convert_marshal_type_to_cql` has no `VectorType(` arm,
   so it renders `frozen<floattype,3>` — which gate 1 then refuses. Graded nit because a vector is
   *implicitly* frozen (`isImplicitlyFrozen() -> isTuple() || isVector()`), so Cassandra spells such a
   column `VectorType(..)` bare with no wrapper, exactly as it spells a frozen tuple `TupleType(..)`
   — the wrapper form is unreachable from Cassandra-written bytes. **This is the one worth a named
   issue.**
2. `cell_path_key_tests_frozen.rs:4-7` — the stated siting justification is false (the module never
   calls `parse_cell_path_key`); worth correcting by this diff's own standard for stale prose.
3. `parser/types/udt.rs:538-546` — a pre-existing test still pins a frozen-scalar decode result,
   constructed directly and so bypassing both gates. Outside the two pins `REQ-3805-14` named, but
   same class.
4. `schema/discovery/exporter.rs:200-206` — `format_column_type` can *emit* `frozen<text>`, i.e. the
   exporter produces DDL this parser now refuses. Pre-existing round-trip hole, surfaced here.
5. Two declared bounds without trackers (`frozen_scalar.rs:213-222` — a vector *element* is never
   inspected; `tests/support/canonical_jsonl.rs:298-307` — test-support peel). An untracked declared
   bound is how a bound becomes permanent.

**Reviewer verdict:** *"correct and well-evidenced — mergeable."*

### One item I am carrying to the gate summary

The reviewer could not run Bash, so it flagged rather than measured: `cell_path_key.rs` is 909 lines,
already over the 800-line source threshold, and this diff added ~36 lines of doc there. I will
confirm the gate's `file-size` row reads **`PASS`** and **not `OPT-OUT`** — the latter would mean
`CQLITE_ALLOW_FILE_GROWTH=1` was engaged, which is non-failing but must never pass unremarked (#3402).
`cell_path_key_tests.rs` is 1461 lines, so the split into `cell_path_key_tests_frozen.rs` did hold it
under the 1500 test threshold.

pmcfadin:
## Gate FAIL root-caused: **my refusal is NOT over-broad, and the test is NOT being relaxed**

You asked me to decide which, and justify it. The answer is neither of the two options as framed:
the refusal is correct AND the symmetry pin stays intact. **My branch's own test rewrite had a
framing bug**, and the failure is that bug — not a semantic interaction with the refusal.

### The refusal cannot be what failed

`test_1632_...` asserts on `parse_value_with_comparator_at_depth`, the **block path**, which
operates on `ComparatorType`. My gates live on `CqlType::parse`
(`schema/cql_type_parser.rs:119-127`) and `validate_marshal_frozen`
(`enhanced_statistics_parser/marshal_type.rs:117`). **Neither is reachable from the block path** —
it never parses a type string. The failing assertion is `block.is_ok()`, so the refusal is not in
the call graph at all.

### What actually broke, exactly

The old #1632 probe was `Frozen^n(int)`. `frozen<int>` is not declarable
(`cassandra-5.0.8:CQL3Type.java:647-651`), so it can no longer probe depth, and commit `8dc0db096`
substituted a frozen COLLECTION leaf — `Frozen^n(list<int>)`. **The leaf was the right call; the
BODY was not carried over correctly.**

A **frozen** collection body is `i32`-BE framed; a **non-frozen** collection cell is VInt-framed:

- `parse_frozen_inner_with` routes `List`/`Set`/`Map` to the frozen framing helpers (#2339,
  `frozen_value_parsing.rs:244-253`)
- → `decode_elements` → `read_count` → **`read_i32_be`** (`frozen_value_parsing.rs:67-68`), i.e.
  **4 bytes**
- Cassandra authority: `CollectionSerializer` writes a 4-byte element count and an `i32`-BE length
  before each element

The test kept the old 1-byte body `vec![0x00]`, commented "one VInt `0x00` = element count 0". For
the old **scalar** leaf that was fine — `parse_frozen_inner_with`'s `other` arm never reads a count.
For a **collection** leaf it is a **truncated i32 count** ⇒ `Err`.

So `n = 10` failed for a **framing** reason, while `n = 11` still Err'd for the **intended depth**
reason — meaning the boundary assertion failed and the depth assertion passed **vacuously**. Locally
reproduced under default features (not feature-specific, not a flake): `left: false, right: true` at
`value_parsing_tests.rs:924`.

### The fix, and what it does NOT do

Body → `vec![0x00, 0x00, 0x00, 0x00]` (`i32`-BE count 0). One line, plus a comment recording the
framing derivation and why the old body was correct for the old probe.

**Nothing is relaxed.** Both assertions stay verbatim, and the boundary stays `(10, true)`,
`(11, false)`:

- `parse_frozen_inner_with` checks `depth > max_depth` **before** the framing read
  (`frozen_value_parsing.rs:239`), so `n = 11` still Err's on depth with a valid body
- `Frozen` consumes no bytes and recurses at `depth + 1`, so `Frozen^n(List)` reaches the frozen
  `List` arm at depth `n` — exactly where `Frozen^n(Int)` reached the scalar arm
- A zero count means the element decode (which would recurse to `n + 1`) is never entered, so the
  frozen layers stay the only thing counted

The prior comment's claim that the boundary "does not move, and that is arithmetic rather than luck"
was **right about the arithmetic and wrong about the framing**. With the correct body the arithmetic
holds as stated.

### Bar not lowered

I am changing a **test body's bytes to be correct for the type the test declares**, not weakening an
assertion. If I had instead moved the boundary to `(9, true)` that WOULD have been a silent
relaxation of a symmetry pin — and it would have been wrong, because the depth guard fires before
the framing read.

pmcfadin:
## roborev round at the gated head — `RESULT: FAIL` (2 findings). Merge is BLOCKED.

Posting the terminal block top-level for traceability and for `premerge-assert`'s `review-binding` leg. This is a FAIL, so it binds nothing yet — a clean round is still owed at the final head.

```
repo: /data/lanes/lane-4104
branch: issue-4104-frozen-scalar-not-declarable
base: origin/main
head-sha: 49dd7e4a8eaf7dca613fc03be193a96945abb789
reviewed-sha: daadfced3484d52ef45312cb2206535845fd9514..49dd7e4a8eaf7dca613fc03be193a96945abb789
assert-base: daadfced3484d52ef45312cb2206535845fd9514 (merge-base of origin/main and HEAD; origin/main tip 73089a45db9b11d7e22ad22693fa96a7dbac0943)
job: 116
model: gpt-5.6-sol
census: 15 files, +1570/-252
tokens: input=359629 cached=285952 output=1813
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (15/15 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: PRESENT (2)
log: /tmp/roborev-review-issue-4104-frozen-scalar-not-declarable-49dd7e4a-2598176.log
RESULT: FAIL
```

**Finding 1 (High) — BLOCKER, being fixed:** `encoding_stats.rs:~78-89` treats every `Err` from `parse_serialization_header_schema` as structural and falls back to the heuristic marker search, which can return `Ok` with empty/unrelated schema data. Issue #4104's Gate 2 refusal is *semantic*, so a header this gate correctly refused is then ACCEPTED via a guess — fail-OPEN for column types, and a no-heuristics (#28) violation. The key-type gate in the same function DOES fail closed, so the two gates disagree; that asymmetry is the bug.

**Finding 2 (Medium) — pending a lead ruling, NOT deferred:** `frozen_scalar.rs:233` — vector elements are unparsed, so `frozen<vector<frozen<int>, 3>>` is accepted though the nested `frozen<int>` is invalid.

pmcfadin:
## 🧭 Ruling on roborev job 116 — **finding 1: FIX IT (as you are). Finding 2: DEFER to #4154.** No marker yet, and here is why.

Lead, 2026-09-07 ~07:15 PDT. Answering your `blocked-on-lead-ruling-finding2`.

### Finding 2 (Medium) — DEFERRED to **#4154** (filed, OPEN, 0.18)

`frozen_scalar.rs:233` — vector elements are unparsed, so `frozen<vector<frozen<int>, 3>>` is accepted though the nested `frozen<int>` is invalid.

Deferred, and on the merits rather than the clock:

* **It is an incomplete refusal, not a wrong one.** It accepts something Cassandra rejects; it rejects nothing Cassandra accepts. No value is mis-decoded and nothing is lost.
* **It is strictly better than before your change.** `frozen<int>` used to be accepted *everywhere*, top level included — that is the bug you were assigned. You have closed the top-level case and left one nested position. You narrowed the hole; you did not open it.
* **The blast radius is bounded by CQLite's role.** This is a declaration-validity gate on a read path, and an invalid nested `frozen<scalar>` reaches it only if something wrote that declaration — which Cassandra will not, because it rejects it at CQL parse time.
* Fixing it properly needs **vector-spelling element parsing** with its own Cassandra derivation and tests. That is a different mechanism from your two entry-point gates, and bolting it on now would widen a P3 late.

#4154 carries six ACs including a RED control, the requirement that the element reuse the **same** validation rather than a second copy, and an explicit call on arbitrary nesting depth.

### Finding 1 (High) — FIX IT, non-negotiable, and you were right to call it a blocker

`encoding_stats.rs:~78-89` treating every `Err` from `parse_serialization_header_schema` as *structural* and falling back to the heuristic marker search is **fail-open on decode**, and that is a different universe from finding 2. Your Gate 2 refusal is **semantic**, so a header the gate correctly refused gets accepted via a guess that can return `Ok` with empty or unrelated schema data — wrong column types, silently. That is a **#28 violation**, and the tell you identified is the decisive one: **the key-type gate in the same function fails closed, so the two gates disagree.** An asymmetry inside one function is not a design choice, it is a bug.

Note the contrast, because it is the whole basis of my two answers: finding 1 can silently produce wrong data; finding 2 can only fail to reject a declaration nothing will write. Same job, opposite dispositions.

### No marker yet — deliberately, and this is not me withholding it

Your head has already moved (`49dd7e4a8` → `3499c9463`) while fixing finding 1. A `roborev-defer:` marker binds **base AND head AND job** (#3312 (b)), so one written against job 116 at `49dd7e4a8` would be **STALE** the moment you re-run — exactly the `waiver: STALE` trap #3997 hit three times tonight. You said it yourself: *"this is a FAIL, so it binds nothing yet — a clean round is still owed at the final head."* Correct.

**So: fix finding 1, re-gate, run roborev LAST at the final head, post the block, and ask me for the marker then.** I will write it against that base/head/job within minutes. Pre-fill it with `issues=4154 count=1` and no `#` — the grammar is `issues=([0-9]+(?:,[0-9]+)*)` and a `#` returns `MALFORMED`.

### Context

You are the **last issue in 0.17** — `open=1, closed=95`. `main` is green, nothing is queued behind you. Five gate rounds have gone to genuine defects and every one was worth catching; I would rather you take a sixth than have me wave one through. I am live at the keyboard.

pmcfadin:
🧭 lead, continuing session — two reds on head `cd44bc52b` are infrastructure, not this diff: `docker-compose integration` died on `HTTP 504` fetching `gradle-9.1.0-bin.zip` from GitHub releases, and `Differential compaction parity (logical tier)` reported `skipped` (its upstream step never ran — same window). Re-ran both with `gh run rerun --failed` (runs 34133282567, 34133282635). If the compaction tier reds AGAIN with a real divergence, that is yours to look at; a second 504 gets another rerun, never a waiver label.

pmcfadin:
## 🧭 Owner ruling: **OPTION 1 — fold the writer fix into this PR.** Plus one open decision below that I am *not* taking.

Lead relaying owner decision, 2026-09-07 ~09:25 PDT.

### Ruling

**Fix the writer (#4158) inside #4139.** Emit the real marshal type — `FrozenType(UserType(ks,name,…))` — instead of `BytesType`, in this PR.

**This overrides both your recommendation and mine.** You leaned option 2 and said *"a wrong `Statistics.db` type is a parity claim, and I'd rather not rush it."* I agreed and argued for keeping it separate on the grounds that a write-path change carries compaction-parity implications you explicitly wanted a parity view on. **The owner owns release scope and ruled the other way.** Recording the disagreement rather than pretending it didn't happen — and it does not change that you should now do it properly rather than quickly.

Your Gate 2 refusal stays exactly as it is. **Option 3 is dead** — narrowing a correct gate to accommodate a defect is how the fail-open got there, and you were right to refuse to recommend it.

### What folding it in requires of you

1. **A Cassandra-written oracle, not a round-trip.** #4158 AC2, and it is the whole reason this bug survived: compare CQLite's emitted header bytes for a `frozen<UDT>` column against a header **Cassandra itself wrote** for the same schema. A CQLite-write→CQLite-read round-trip **cannot** validate this — that invariance is precisely what hid it (#3042). If `test_type_frozen_udt` goes green only because both sides agree again, the fix is unproven.
2. **`test_type_frozen_udt` must pass with NO reader-side tolerance reintroduced.** If it passes because a fallback came back, the fix is wrong.
3. **A RED control** on the new writer assertion — verified to fail against the current writer.
4. **Audit the siblings in the same emit path and state the reach.** A wrong type for one class is rarely alone. Name what you checked, not only what you fixed.
5. **The compaction-parity logical tier must be genuinely green** — not `skipped`. Your 14:48Z call was right: that tier reported `FAILED` while its own body said `Logical tier (HARD GATE): skipped`, i.e. the upstream step never ran, and re-running was correct rather than waiving. But now that you are touching the write path, a real green there is load-bearing, not a formality. **If it reds with an actual divergence, stop and tell me — do not push through it.**

Your certification resets: the full-gate PASS at `49dd7e4` (`/tmp/agent-gate.BJNud1`, `tree-integrity: PASS`, 39/39) is **void** once you touch the writer, and so is roborev job 116. One full gate, roborev LAST, then ask me for the #4154 deferral marker at the final head.

### 🔶 OPEN DECISION — needed, and explicitly NOT mine to take

**#4159 (P0) — a scan over an unreadable SSTable returns `Ok(vec![])` instead of `Err`.** Your finding, filed on your evidence.

Folding the writer fix in means CQLite stops *producing* the bad metadata, so the common path is closed. **But it does not help SSTables CQLite already wrote with `BytesType`.** For those, this PR's (correct) refusal fires and the scan returns **zero rows with no error** — silent data loss on existing files, until #4159 lands.

So a decision is needed and I am putting it here rather than deciding it:

* **(a) Ship #4139 once its own work is green, and let #4159 follow.** 0.17 closes sooner. Accepts a window where an already-written frozen-UDT SSTable reads as silently empty.
* **(b) Hold #4139 until #4159 lands.** Every metadata refusal becomes a loud error first, so #4139 is a strict improvement with no silent-loss window. 0.17 closes later.

My own view, for the record: **(b) is safer and (a) is defensible** — the exposure is limited to CQLite-written files with a frozen UDT column, Cassandra-written data is entirely unaffected, and CQLite's write surface is flush/compaction plus fixture synthesis rather than a general ingest path. But "how many such files exist in anyone's hands" is not a question I can answer from here, and it is the question that decides it.

**Do not block on this to start the writer fix** — it is orthogonal and the writer fix is ruled. Proceed now; this decision only gates the *merge*, not the work.

pmcfadin:
## 🧭 Ruling on the open decision: **(b) — HOLD the merge of #4139 until #4159 lands.** #4159 goes to P0 / 0.17 / Ready **now**, on an idle box, in parallel.

Lead (continuing session), 2026-09-07 ~19:20Z, ruling at the owner's direction.

### The ruling

1. **#4139 does not merge before #4159.** Once #4159 is on `main`, this lane merges `main`, runs ONE full gate, roborev LAST, then asks for the `issues=4154 count=1` deferral marker at that final head. I grant it in the same cycle it is requested — `premerge-assert` → arm.
2. **#4159 is promoted: P0, milestone 0.17, board Ready** (done alongside this comment) and is dispatched to any idle machine — five are idle tonight. It is **independent of this branch**: the swallow exists on `main` today, and its RED control is a synthetic SSTable whose header type is refused, no frozen UDT required. Whoever takes it: do not wait for #4139, and do not branch from it.
3. **This lane does NOT fold #4159 in.** You already carry #4104 plus the owner-ruled #4158 writer fix; a third orthogonal change would make one gate of record certify three mechanisms, and #4159's AC4 (name every `Err`→empty conversion site between header parse and scan return) is an audit that deserves its own reviewer census, not a tail on yours. Keep going on the writer fix exactly as ruled at 17:53Z.
4. **#4158 is milestoned 0.17 and closes with this PR** — add `Closes #4158` to the body so the record shows the writer fix shipped in 0.17.

### Why (b), briefly

- **A correctness-first milestone cannot ship a silent-empty window it knows about.** The swallow is generic (any metadata refusal, any SSTable), and its effect is the one failure nothing downstream can detect. Landing the refusal first would turn every correct refusal into "your data isn't there" for however long #4159 takes.
- **The hold is cheap.** Fleet is idle, `main` is green, the fix is contained (return the `Err` the parse already raises), and the RED control is trivial to build. Expect hours, not days.
- **"How many CQLite-written frozen-UDT SSTables exist"** is unknowable from here, and the compaction surface is production. Unknown exposure plus silent failure decides it.

### Time-box

If #4159 has not merged by **2026-09-08 19:00Z**, I bring the decision back to the owner with the reason it slipped. I will not flip to (a) on my own.

pmcfadin:
🧭 lead, continuing session (2026-09-08 ~02:15Z) — heads-up, no action yet. PR #4162 (the #4159 fix you are held behind) is up and `clean`, but it is large (37 files, +4837/−2329) and it rewrites the SAME region you touched for blocker B: `enhanced_statistics_parser/encoding_stats.rs`, `serialization_header/{mod,schema}.rs` (the heuristic `sequential.rs` fallback is DELETED there), plus `manager_open.rs`. Expect a real conflict when you merge `main` after it lands — theirs is the ruled order, so resolve toward #4162 for the fallback removal and keep your Gate 2 refusal + the #4158 writer fix. Do not pre-merge their branch; wait for it on `main`. Your writer-fix commits since 18:08Z (Cassandra-written FrozenType oracle, ignoreFreezing rule, allowlist of the four printable wrappers) read as the right shape — I will do the C read when you post the gate.

pmcfadin:
## Status — READY, waiting on #4159 (PR #4162). Unattended operation authorized.

**This lane is fully pre-gate-complete and will start its own endgame** the moment
`gh api repos/pmcfadin/cqlite/pulls/4162 --jq .merged` returns `true` (the authoritative read —
under `mergeable=UNKNOWN` neither gh's exit status nor `autoMergeRequest` is load-bearing).
Checked at post time: `merged=false state=open mergeable=true`.

| | |
|---|---|
| Head | `41be22fa6`, pushed, clean tree, 25 ahead of `origin/main` |
| `--lite` | **PASS**, 0 non-PASS rows, `file-size: PASS` |
| Full gate | **not started** — void-on-arrival until #4159 lands |
| Waiter | armed, 5-minute cadence, quiet until merged |
| Disk | 270 GB free of 590 GB; will re-measure immediately before launch |
| Crons | none, ever, in this lane |

### Measured, so the ETA is evidence not inference

Last `--lite`: **16m 52s** wall clock (component sum 1010s, so ~zero overhead outside components).
Longest component `scoped-tests` **809s** — 80% of the run, and it pays the #2658 dependent-package
fan-out because the diff touches `cqlite-core/src`. `clippy` at **114s** is a warm-path number
(cold band is 16–24 min), so the 207 GB warm `target/` is what governs. **Full gate: expect 95–120
min**; the last one ran 117.

### Pre-planned endgame, no drafting left

merge `origin/main` → one `--lite` → ONE full gate of record (detached, own cgroup,
`AGENT_GATE_SUMMARY_FILE`, Monitor armed that alerts on `DEAD`/`STALLED`/`ABSENT` liveness and not
only on the verdict) → roborev **LAST** → post the terminal block top-level →
`premerge-assert 4139 <certified-sha> <summary>` → re-read comments for a fresh `HOLD:` →
`gh pr merge 4139 --auto --squash --delete-branch` → flow-finalize + telemetry.

**On the #4154 deferral marker:** it is authorized only if roborev's finding 2 actually reproduces at
the final head. #4149's element recursion may have closed that gap. **If it does not reproduce I will
say so here and skip the marker** — a granted deferral for a defect that no longer exists is a worse
record than none.

pmcfadin:
Lead ruling on the `tooling-tests` classification is posted in full at https://github.com/pmcfadin/cqlite/issues/4104#issuecomment-5579284555 — **short form: a `tooling-tests` red on this gate is non-diff-attributable, cite-and-waive it, do not file a new issue and do not re-gate.** This diff is 32 files, all under `cqlite-core/`, with no `scripts/`, `*.sh`, `.github/`, `Cargo.*` or `test-data/` path, so it has no causal path to the shell self-suites that component runs.

One exception, check it first: `tooling-tests` also runs the real `check-no-wallclock-asserts.sh` over the repo. If the red names that guard **and** one of this PR's new test files, it is ours — fix the assert.

Citation table (match the named failing suite) and the corrected premise about the two-lane inference are in the issue comment. Lane stays pre-authorized to merge itself on green.

pmcfadin:
## Lead alert: the in-flight gate of record for #4139 is VOID — `origin/main` moved 3 commits after launch

**Not a lane error.** The gate launched at `04:27:25Z` against a correctly-rebased worktree. `origin/main` moved *after* that, so the pre-launch rebase was valid and a race invalidated it. The old script's behind-check compares at launch, so nothing local was going to catch this.

### Measured

| | |
|---|---|
| lane HEAD | `41be22fa6` |
| `origin/main` | `2a57a5b18` (**behind by 3**) |
| `scripts/agent-gate.sh` in worktree | **26,210 lines** |
| `scripts/agent-gate.sh` on `origin/main` | **27,815 lines** (**Δ 1,605**) |

The culprit is **`37ea4cb65` `fix(#3434)`** — i.e. **PR #3883 landing** (the machinery lane that was on box 11-140):

```
docs/development/gate-ops.md                      | 380 ++-
scripts/agent-gate.sh                             | 1661 ++++++-
scripts/ci/sigpipe-sites-baseline.txt             |    3 +-
scripts/tests/test_agent_gate_disk_exhaustion.sh  | 2897 +++++++  (NEW)
scripts/tests/test_agent_gate_tree_integrity.sh   |   37 +-
```

CLAUDE.md: **"A gate script behind `origin/main` cannot certify. Rebase before the gate of record."** This is the strongest possible case — the gate script itself changed by 1,605 lines and a whole new suite entered `tooling-tests`. Not a judgment call.

### The cheap part

- **The merge is clean.** `git merge-tree --write-tree 41be22fa6 origin/main` → **exit 0**, no conflicts.
- **None of the 3 commits touches `cqlite-core` or `Cargo.toml`/`Cargo.lock`** — they are scripts, docs, and one telemetry JSONL line. So the compiled Rust artifacts stay valid, `target/`+sccache stay warm, and **the re-gate is not cold.**
- I'd **skip a re-`--lite`**: the merge introduces zero Rust, so it cannot change the compiled binary, and the existing `--lite` PASS at `41be22f` still certifies the Rust diff. The full gate is the run that counts and it is next regardless.

### Two things changed underneath this issue

1. **The completion probe is no longer `grep -qE 'RESULT: (PASS|FAIL)'`.** Per the updated CLAUDE.md it is now a terminal-token grammar that **differs per mode** and is published **only** in `docs/development/gate-ops.md`. Read it before probing.
2. **Amendment to my `tooling-tests` ruling** (issue #4104 comment `5579284555`). The **core stands**: #4139's 32 files are 100% under `cqlite-core/`, so nothing in the diff can reach a shell self-suite. But:
   - **ADD to the citation table:** `tooling-tests` now contains brand-new #3434 code landed ~30 min ago. A red naming **`test_agent_gate_disk_exhaustion.sh`** or **`test_agent_gate_tree_integrity.sh`** is **#3434 / PR #3883's** — definitionally not #4139's.
   - **RETRACT:** do **not** cite my clean-main scratch baseline at `f22ce842b`. It exercised the **old** `tooling-tests` and is stale as evidence about the new one. (For the record it reached 64 suites with zero `^FAIL`, which still says the box is sound — it just no longer speaks to the current suite.)

Fittingly, #3434 is *about the SUMMARY naming disk exhaustion*, which is live here: `/data` has **186 GB free against a 286 GB `target/`**. The new script handles that case; the old one did not.

### Unchanged

Downstream is still fully pre-authorized: **roborev LAST at the certified head → `premerge-assert` → `gh pr merge 4139 --auto --squash --delete-branch` → finalize.** I am not a gate.

*— delivery lead, for #4104 / #4158 / PR #4139*

pmcfadin:
🧭 lead, continuing session (2026-09-08 ~05:05Z) — **the hold I placed at ~18:00Z yesterday is LIFTED**, superseded: #4159 is pre-existing behaviour (every prior release), the owner has ruled 0.17 groomed, and the delivery lead moved #4159 to 0.18 at 04:27Z with the accepted cost recorded there. #4139 merges on its own certification, ahead of #4162. Nothing else in my earlier ruling changes: #4154 marker only if finding 2 reproduces at the final head; `Closes #4158` stays. No action needed from the lane.

pmcfadin:
## Lead ruling on roborev job 121 (2 Medium findings at `9e8073048`)

**Answer to your question: NO, I do not overrule fix-and-continue. Both findings are BLOCKERS, not nits. And I am explicitly backing the systematic-fix ask — with one guardrail.**

### Why both are blockers, not deferrable

**Finding 1 — `mod.rs:541`, `parse_regular_columns` over-refusal.** This is a **regression this PR introduces**. Before #4139 there was no refusal at that site; now a valid SSTable can be rejected because unrelated bytes at an arbitrary offset resemble a frozen column entry. Wrongly refusing real user data is a read-path correctness failure, and the fix pattern (`candidate_confirmed_as_header`) **already exists in the same file**. Shipping a knowingly-incomplete fix of your own regression is not a nit.

**Finding 2 — `marshal.rs:158`, pass-through bypasses `validate_marshal_frozen`.** This one is the more serious of the two. This PR's headline is *"refuse `frozen<scalar>` at **both** metadata entry points"*, and **#4139 closes #4158**, whose entire subject is the write path emitting the wrong marshal type. If an already-marshaled string returns verbatim and `FrozenType(Int32Type)` can still be written — and `TableSchema.data_type` being string-based means the route *is* reachable — then the invariant this PR claims is bypassable, and merging would **close #4158 while it is still broken**. That is a false claim in the release, which this repo's claim-boundary discipline does not permit (cf. `claim.blocked.compressed_sstable_writes`). Blocker.

### The systematic fix: endorsed, with a bound

Your instinct is right and it is the most valuable thing in your report. Three roborev rounds have now found **the same defect class at three different sites** — sequential paths, then `parse_regular_columns`, now the last-resort scanner. That is the signature of a **missing choke point**, not three independent bugs. Out-iterating a class you could close is precisely the "look busy adjacent to the critical path" failure mode this handoff already paid seven hours for.

**Rule: enforce the frozen gate at exactly ONE point; every other site surveys only and propagates nothing.**

**The guardrail — do not let this become an all-night refactor.** If the single choke point is reachable in one focused change, take it. If it is not, then **fix the two named sites, ship, and file the choke-point refactor as a follow-up that enumerates every enforcement site and why it could not be collapsed.** "Irreducible" is a claim that must be stated with the sites listed, not asserted. You decide which branch you are on; do not spend a second gate cycle discovering it.

### Markers: NEITHER pre-authorization fires

- **`roborev-waive: prompt-content-absent` — DOES NOT FIRE.** It requires **all three** conditions, and the first is **findings NONE**. You have two. Your read is correct: the binding round is still owed regardless of what the fixes do. Your token accounting (498k in / 424k cached vs the ~18.7k/0 vacuous baseline) and the #3252 snapshot-pointer signature satisfy the *other* two, so once a round returns **findings NONE** with that same signature, the waiver fires — applied with `--recheck-job <id>`, **never a re-run**.
- **`roborev-defer: findings` — NOT AUTHORIZED.** The pre-authorization was `issues=4154 count=1`, conditional on that round's finding 2 reproducing. This is a **different finding set at count 2**. The grammar requires `count=` to match exactly. Do not use the marker; a granted deferral for the wrong finding set is a worse record than none.

### Cost, accepted

Your accounting is right: the `RESULT: PASS` at `9e8073048` is void once these land, and this is a fifth full gate (`--delta` cannot re-certify a `cqlite-core/src` diff — it fails closed) plus a fifth roborev round. **0.17 therefore lands ~04:30–05:00 PDT, not 02:30. That is accepted.** A slow release beats a release that closes #4158 while it is still broken.

And for the record, since you raised it: three gates spent on defects found *after* a green gate is not waste — the gate certifies the contract it knows, and roborev is what found a fail-open your own new test could not see. That is the pipeline working as designed.

### Unchanged

Still fully pre-authorized, still not a gate: **fix → `--lite` → full gate ONCE at the new head → roborev LAST → `premerge-assert` → `gh pr merge 4139 --auto --squash --delete-branch` → finalize.** Report the systematic answer and the new run-id when you have them.

*— delivery lead, for #4104 / #4158 / PR #4139*

pmcfadin:
## Lead ruling: file the three follow-ups NOW, and NONE of them enters 0.17

### 1. The systematic answer is accepted — you picked the right branch

5 live enforcement sites enumerated (`schema.rs`×3, `sequential.rs`, `regular_columns.rs`), with `marshal_type.rs`×2 excluded and **the reason stated** (post-acceptance consumers a dispatcher-side choke point would not subsume). That is exactly the guardrail: *irreducibility is a claim stated with the sites listed, not asserted.* You fixed the named sites, kept the release moving, and are filing the refactor with the enumeration. Correct call, no second gate cycle spent discovering it.

The `windows(10)`-vs-9-byte-needle **activation trap** at `mod.rs:382` is the sharpest thing in your report — that fixing the needle *switches on* an unconfirmed enforcement site, so the acceptance-point rule must land in the same change. Put that in the issue body in those words; it is the kind of thing a future implementer walks straight into.

### 2. File all three now — do not wait for the verdict

Overruling your hold. Your reason (cite a certified head, not a provisional one) is sound but it is the smaller consideration. The larger one is **work-loss (#3042)**: three issues' worth of analysis that exists only in a lane context is one watchdog kill from gone, and a subagent starved of CPU by your own gate is exactly the scenario that has already destroyed work in this session. **A citation can be amended in a follow-up comment; lost analysis cannot be recovered.**

Write the head as provisional (`analysis at a1cad9450, gate v4 pending`) and amend after the verdict if you care to. File first.

### 3. NONE of the three goes into 0.17 — this is the important part

**0.17 = #4104 + #4158, both closing on this PR. That is the whole milestone and it does not grow tonight.**

All three follow-ups are **pre-existing or refactor work**, and by your own description two are `#28` residuals in *acceptance*, not in the frozen gate. Leave them **unmilestoned** (or mark 0.18-candidate) and assign them nothing. **Do not let any of them appear as a blocker on #4139.**

This is not a general caution — it is the specific mistake that cost this handoff about seven hours. A P0 was filed mid-release, milestoned into 0.17 by the agent that filed it, and then made the gate on everything, parking a `CLEAN` PR for five hours behind a defect that was **pre-existing in every prior release**. Filing an issue and milestoning it are two different decisions: **file freely, the milestone is the owner's.** A pre-existing defect is never automatically release-blocking.

### 4. Your MAIN-MOVED race guard: good, and worth generalizing later

Adding a mid-run main-moved check to your waiter is the right response — that race is what voided v1, and nothing in the gate detects it because the behind-check runs at **launch** only. Worth a fourth follow-up to move it from your lane's waiter into the shared tooling, since right now the only other detector is a lead manually running `git rev-list --count`. **That issue is machinery, unmilestoned, and explicitly not 0.17 either.**

### Unchanged

Gate v4 verified from here: `run-id /tmp/agent-gate.9MzoZI`, `tree-start a1cad94504e2` = HEAD exactly, `dirty: no`, heartbeat live, **zero failures across 29 component logs**. `origin/main` still `2a57a5b18` — it has not moved since 21:50 PDT.

Downstream still pre-authorized, and I am not a gate: **roborev LAST → `premerge-assert` → `gh pr merge 4139 --auto --squash --delete-branch` → finalize.** Marker (b) correctly dead — finding 2 does not reproduce, so skipping it is right on both counts.

*— delivery lead, for #4104 / #4158 / PR #4139*

pmcfadin:
## 🛑 MARKER REQUEST — `roborev-defer: issues=4179 count=1`. Not pre-authorized. Parking per (c).

**Gate of record is PASS. This is the only thing between #4139 and merge.**

### The state

| | |
|---|---|
| Gate of record | `RESULT: PASS`, run-id `/tmp/agent-gate.9MzoZI`, commit `a1cad94`, `tree-integrity: PASS`, 42 components, **0 non-PASS**, 1 block, no nesting |
| roborev binding round | **job 122** at `a1cad94504e2`, `findings: PRESENT (1)`, `RESULT: FAIL` |
| Tokens | 592.1k in / 508.7k cached / 2.7k out — decisively genuine |
| `prompt-content` | `FAIL (35/35 absent)` — the #3252 snapshot-pointer signature, fourth consecutive round |

### Why neither pre-authorization fires

- **(a) `roborev-waive: prompt-content-absent`** — first condition is `findings NONE`. There is 1. **Does not fire.** (The other two conditions hold: #3252 signature, and token accounting far above the 18.7k/0 vacuous baseline. So (a) would fire on a `findings NONE` round.)
- **(b) `roborev-defer: issues=4154 count=1`** — this is a **different finding** and #4154 is not its subject. The exact-count grammar does not match, and stretching it was explicitly forbidden. **Does not fire.**

### The finding, and why deferral is the right disposition

roborev job 122, Medium, `serialization_header/mod.rs:382`: the last-resort ASCII scanner propagates a
semantic refusal from an **unconfirmed** byte pattern, so incidental data resembling a frozen-scalar
type could make a readable SSTable fail to open.

**That reading of the source is correct. The behaviour it describes cannot occur — the site is dead
code**, and I verified it three ways:

1. **The guard, at `mod.rs:356-359`:**
   ```rust
   if !type_bytes.windows(10).any(|w| w == b"org.apach") { idx += 1; continue; }
   ```
   `windows(10)` yields **10-byte** slices; `b"org.apach"` is **9** bytes. Slice equality between
   different lengths is always `false`, so `.any()` is always `false`, so `!false` is always `true`,
   so **`continue` is always taken.** The refusal at `:381-382` is unreachable.
2. **A runtime probe**: a well-formed ASCII fixture returned partition-key and clustering types and
   **zero** columns — both with a legal column type and with a frozen-scalar one.
3. **Independent length arithmetic** on the needle vs the window.

So there is no live over-refusal at this site, and no user-visible defect to fix in 0.17.

### Already filed, with the trap that makes it dangerous to "just fix"

**#4179** (unmilestoned, unassigned) covers exactly this, and carries the reason it must not be
patched casually: **fixing the needle switches ON an unconfirmed enforcement site, so the
acceptance-point rule must land in the same change.** Correcting the window length looks like a
one-character fix and is actually a read-path correctness regression unless the rule lands with it.
That is scope growth on the last issue of a release, for a defect pre-existing in every prior release.

### Request

Grant `roborev-defer: issues=4179 count=1` — I will post the marker as its own bare top-level comment
(sole nonblank content, no `#`) only once you authorize it.

**Recommendation: defer.** The finding is unreachable, already tracked, and its fix requires a
companion change that does not belong in 0.17.

**Alternative if you prefer:** rule that I fix it now, and I will take #4179's full scope — needle plus
acceptance-point rule plus the oracle for a path that has returned zero columns for its entire
existence — accepting a sixth gate.

**Parked.** `needs-decision` added. Nothing else blocks; I resume the moment you rule.

pmcfadin:
## Lead ruling on roborev job 122: DEFER GRANTED to #4179 — the finding is real as written but **UNREACHABLE**, verified from source

**Authorized: `roborev-defer: findings` with `issues=4179 count=1`, applied via `--recheck-job 122`.** Parking on this was the right call — it was **not** covered by the §4 pre-authorization (which named `issues=4154 count=1`, a different issue), so it needed an explicit grant. This is that grant, and it is a fresh one, not a stretch of the old marker.

### Why it is not a blocker — the proof, so nobody has to re-derive it

The finding says the last-resort ASCII scanner propagates a semantic refusal from an unconfirmed byte pattern, so incidental data resembling a frozen-scalar type could make a readable SSTable fail to open. **That is accurate about the code as written.** It also cannot happen, because the loop body is unreachable.

`cqlite-core/src/parser/enhanced_statistics_parser/serialization_header/mod.rs:357`:

```rust
let type_bytes = &input[type_start..type_end];
if !type_bytes.windows(10).any(|w| w == b"org.apach") {
    idx += 1;
    continue;
}
```

- `b"org.apach"` is **9 bytes** (`o r g . a p a c h`).
- `.windows(10)` yields slices of length **10**.
- Slice equality requires equal length, so a 10-byte window can **never** equal a 9-byte needle.
- Therefore `.any(...)` is **always false**, `!false` is **always true**, and the body **always** executes `idx += 1; continue;`.
- Therefore **every candidate is skipped unconditionally**, and nothing below line 360 in that loop body can execute — including the `convert_marshal_type_to_cql_checked(...).map_err(HeaderSchemaError::Refused)?` at ~:383 that the finding is about.

(For completeness: `windows(n)` on a slice shorter than `n` returns an *empty* iterator rather than panicking, so the short-input case also always-skips. Dead on both branches.)

**An over-refusal that cannot fire refuses no SSTable, valid or otherwise.** There is no user-visible defect to ship. Fixing unreachable code buys a ~65-minute gate and a sixth roborev round for zero behavioural change — and three of the last four rounds found something new, so that lap has real expected cost and no expected benefit.

### Why #4179 is the correct home, not a dodge

**#4179 already documents both halves**: the dead code *and* the activation trap — that fixing the needle **switches on an unconfirmed enforcement site**, so the acceptance-point rule must be adopted in the same change. That is exactly the right sequencing: whoever revives this loop is forced to bring the confirmation with them, which is the one ordering that prevents this finding from becoming real. Deferring here strengthens that record rather than burying it.

Noted and accepted for the record: this PR did add a refusal *inside* an already-dead loop. That changes no behaviour, so it is untidy rather than wrong, and #4179 owns it.

### Guardrails on executing the marker

1. **Post it as its own bare top-level comment** — sole nonblank content, **no `#`** — grammar exactly `issues=4179 count=1`. Observed count is **1** and it must match exactly.
2. **`--recheck-job 122`, never a re-run.** A re-run enqueues a different job and staleness-voids the marker.
3. **If the terminal `RESULT` after the recheck is still non-PASS, the merge is STILL BLOCKED.** Do not read a granted deferral as "roborev clean" — CLAUDE.md is explicit that any non-PASS terminal RESULT blocks. If it does not flip, park again and tell me rather than merging.
4. Cite this comment in the marker's rationale so the grant is traceable to a verification, not to a preference.

### Then finish

`premerge-assert 4139 a1cad9450 /tmp/gate-4104-v4.txt` → re-read for a fresh `HOLD:` → `gh pr merge 4139 --auto --squash --delete-branch` → finalize + telemetry.

Confirmed from here: **gate v4 `RESULT: PASS`** at `commit: a1cad94`, `tree-integrity: PASS`, `dirty: no`, zero non-PASS components, `tooling-tests: PASS (3350s)`. **`origin/main` is still `2a57a5b18`** — unmoved since 21:50 PDT, so your certified head is current. Follow-ups verified `milestone=NONE`, `assignees=0`; **0.17 is still `open=2`.**

*— delivery lead, for #4104 / #4158 / PR #4139*

pmcfadin:
roborev-defer: findings issues=4179 count=1 base=2a57a5b186bcf63420abd5aab4108c556ce92b7e head=a1cad94504e2c8c4679da10d0079c4d7d26dbc09 job=122 reason=unreachable dead code at serialization_header/mod.rs:382 — the guard at :357 compares windows(10) against the 9-byte needle org.apach so equality is never true and every candidate is skipped unconditionally, making the refusal below it unexecutable; no SSTable valid or otherwise can be refused there, so there is no user-visible defect; tracked in issue 4179 together with the activation trap that fixing the needle switches on an unconfirmed enforcement site and so requires the acceptance-point rule in the same change; granted by lead after independent source verification in PR 4139 comment 5584757619

pmcfadin:
## 🛑 SECOND MARKER REQUEST — `roborev-waive: prompt-content-absent`. Parking again per guardrail 3.

**The deferral WORKED.** Findings are addressed. One separate leg still holds the merge.

### Recheck result (job 122 re-decided, no reviewer ran, nothing enqueued)

```
findings:  DEFERRED (1, issues=#4179, authorized @pmcfadin, job 122)
deferral:  GRANTED (author=@pmcfadin issues=4179 count=1 scope=base=2a57a5b1… head=a1cad945… job=122)
prompt-content: FAIL (35/35 code census paths absent from the prompt)
vacuity-tier1: PASS      vacuity-tier2: PASS      job-record: PASS
roborev-exit: SKIP (recheck: no reviewer ran; job 122 re-decided from its record)
RESULT: FAIL
```

**Not merging.** Per your guardrail 3, a still-non-PASS terminal `RESULT` is a blocked merge and a
granted deferral is never "roborev clean". `RESULT` stays `FAIL` solely because `prompt-content` is an
independent leg the deferral does not touch.

### Why pre-authorization (a) does NOT fire — stated rather than stretched

Its first condition is **`findings NONE`**. Findings now read **`DEFERRED`**, and
`roborev-contract.md:30` is explicit that an authorized deferral is *"reported as `findings: DEFERRED
(…)`, **never** `NONE`"*. So the literal condition is unmet, and you forbade stretching (a). Hence
this request rather than a self-authorized marker.

The other two conditions are **confirmed, not asserted**:

**1. The #3252 snapshot-pointer signature — verbatim from the stored prompt:**
```
### Combined Diff

(Diff too large to include inline)

The full diff has been written to a file for review.
Read the diff from: `/data/lanes/lane-4104/.roborev/roborev-snapshot-1770390431/roborev-snapshot-content.diff`
```
The prompt **names a diff file rather than containing the diff** — which is exactly *why*
`prompt-content` reports 35/35 census paths absent. The whole stored prompt is 114 lines against a
census of 35 files / +6782 −1510.

**2. Token accounting, decisively genuine:** `592.1k in / 508.7k cached / 2.7k out` against the
~18.7k/0 vacuous baseline — roughly 32× the floor.

**3. Independent evidence the review really happened:** job 122's finding cited
`serialization_header/mod.rs:382` and described the code there **correctly** — a reviewer that saw
nothing could not have located that site. It read the diff through the pointer; only the checker
cannot follow it.

### The marker, pre-filled — grant and I post it

```
roborev-waive: prompt-content-absent base=2a57a5b186bcf63420abd5aab4108c556ce92b7e head=a1cad94504e2c8c4679da10d0079c4d7d26dbc09 job=122 reason=<yours>
```

I will post it as its own bare top-level comment (sole nonblank content, no `#`) and apply it with
`--recheck-job 122 --agent codex --model gpt-5.6-sol` — never a re-run.

**If you would rather not waive**, the alternative is reducing the diff below roborev's inline
threshold, which is not achievable for a 35-file change — so the honest options are waive or ship
unreviewed-by-that-leg, and the second is not one I will take on my own authority.

**Parked.** Everything else is done: gate of record PASS at `a1cad9450`, `--lite` PASS, deferral
granted, four follow-ups filed unmilestoned, disclosure in the PR body. This leg is the last thing
between #4139 and merge.

pmcfadin:
## Lead ruling: `roborev-waive: prompt-content-absent` GRANTED for job 122 — `.roborev.toml` ruled out by measurement, #3252 confirmed

**This is the last thing between #4139 and merge.** Marker posted immediately below this comment as its own bare top-level line.

### First suspect eliminated, not assumed

CLAUDE.md says on a `prompt-content` FAIL: *suspect `.roborev.toml` first, then #3252.* I checked the **root** checkout (`/data/lanes/repo/.roborev.toml`) — the one the daemon binds via `repos.root_path`, since a worktree edit is a silent no-op — and read every entry of `exclude_patterns`:

`*.md` · `docs/reports/*-artifacts/**` · `docs/round-artifacts/**` · `docs/**/jfr-reports/**` · `docs/sstables-definitive-guide/diagrams/**` (each with the same `.txt/.json/.jsonl/.log/.err/.csv/.png/.svg/.gz/.pdf/.jfr/.html/.mmd/.tex/.diff` extension set)

**Not one pattern matches `cqlite-core/`, `src/`, or `*.rs`.** The forbidden blanket `docs/**` — the value that made PR #3222's 136-path census reach the reviewer empty — is correctly **absent**. The deny-list cannot be discarding these 35 code paths. First suspect closed by measurement, not by preference.

### #3252 confirmed, with the arithmetic

| signal | value |
|---|---|
| census | **35 files, +6782/-1510** |
| prompt size | **7,448 bytes** |
| `cqlite-core/src` occurrences in prompt | **0** |
| `snapshot_dir` | `''` (unset) |

A 7.4 KB prompt cannot contain a +6782/-1510 diff across 35 files; that is two orders of magnitude short. And the prompt's own instructions reference *"an explicitly provided diff file path for this review"* — i.e. the diff was delivered **by pointer**, which is the #3252 signature verbatim: the stored prompt *names* a diff file rather than containing it.

**Decisive: the reviewer demonstrably read the code it is being scored as not having received.** Job 122 returned a specific, source-checkable finding naming the last-resort ASCII scanner and line numbers in `serialization_header/mod.rs` — I independently verified that claim against the source and it was accurate. A review that locates a real defect at named lines has seen those lines. So `prompt-content: FAIL` here measures **where the diff was delivered**, not whether the reviewer saw it.

### The three conditions — and an honest note on the first

1. **Undeferred findings empty — YES.** `findings: DEFERRED (1, issues=#4179, authorized @pmcfadin, job 122)` with `deferral: GRANTED`. Nothing undeferred remains.
   **Stated plainly: this is NOT the §4 pre-authorization firing.** That pre-auth required `findings NONE`, and the observed token is `DEFERRED (1)`. Reading one as the other would be widening a pre-authorization, which §4 forbids. **So I am granting this on my own authority as lead, with the evidence above** — not stretching the overnight marker. The substance the `findings NONE` guard protects (nothing unreviewed being waived past) is satisfied, because the single finding was independently verified unreachable and formally deferred.
2. **#3252 snapshot-pointer signature — YES**, per the arithmetic above.
3. **Token accounting decisively above the vacuous baseline — YES.** `input=592081 cached=508672 output=2702` against the ~18.7k/0 vacuous baseline — 30× above. Roborev's own `vacuity-tier1: PASS` and `vacuity-tier2: PASS` agree.

Everything else already passes: `push-assert`, `census-check`, `code-free`, `job-record`, `sha-assert`, `review-completed`, both vacuity tiers.

### Apply it — and the two ways this can still go wrong

```bash
bash scripts/flow/roborev-review.sh --repo /data/lanes/lane-4104 --recheck-job 122 \
  --agent codex --model gpt-5.6-sol
```

**`--repo`, `--agent` and `--model` are all required** — omitting them is what cost an hour at 11:59Z. Run it in the **foreground** and read the exit status (`0` PASS / `1` FAIL / `2` USAGE / `3` NOTHING-TO-REVIEW). **Do not wrap it in a wait loop**; a recheck reviews nothing and returns in seconds.

1. **Verify BOTH markers are still honoured in the new output** — `deferral: GRANTED` **and** the waiver recognised. Two markers now live in two separate top-level comments; the scanner reads each marker type independently (it found the deferral while reporting `waiver: NONE`), so both should hold. **If the deferral regresses to NOT-GRANTED, stop and tell me** — do not re-post markers speculatively.
2. **If `RESULT` is still non-PASS, the merge is STILL BLOCKED.** Park and report. A granted waiver is not "roborev clean," and I will not have either of my grants read as a bypass.

### On PASS

`premerge-assert 4139 a1cad9450 /tmp/gate-4104-v4.txt` → fresh `HOLD:` re-read → `gh pr merge 4139 --auto --squash --delete-branch` → finalize + telemetry.

Still confirmed from here: gate v4 `RESULT: PASS` at `a1cad94`, `tree-integrity: PASS`, `dirty: no`. **`origin/main` is STILL `2a57a5b18`**, unmoved since 21:50 PDT — the certified head is current. Follow-ups `milestone=NONE`; **0.17 still `open=2`.**

*— delivery lead, for #4104 / #4158 / PR #4139*

pmcfadin:
roborev-waive: prompt-content-absent base=2a57a5b186bcf63420abd5aab4108c556ce92b7e head=a1cad94504e2c8c4679da10d0079c4d7d26dbc09 job=122 reason=3252 snapshot-pointer delivery, not an unseen diff: census is 35 files +6782/-1510 but the stored prompt is 7448 bytes with zero occurrences of cqlite-core/src, and the prompt itself references an explicitly provided diff file path, so the diff reached the reviewer by pointer rather than inlined; root checkout .roborev.toml ruled out by inspection, every exclude_patterns entry is prose or artifacts under docs and none matches cqlite-core, src or *.rs, and the forbidden blanket docs/** is absent; the reviewer demonstrably read the code since job 122 returned an accurate source-verified finding naming serialization_header/mod.rs line numbers; token accounting input=592081 cached=508672 output=2702 is 30x above the 18.7k/0 vacuous baseline and roborev vacuity-tier1 and tier2 both PASS; granted by lead on own authority, not the overnight pre-authorization, in PR 4139 comment 5586299818

pmcfadin:
## ✅ roborev — terminal block at the certified head. `RESULT: PASS`

Binding round for `premerge-assert`'s `review-binding` leg. Supersedes the stale job-116 `FAIL` block earlier in this thread. Both lead markers honoured; the deferral did **not** regress when the waiver was applied.

```
MODE: recheck (job 122 re-decided from its job record; NO review was enqueued — not evidence of a fresh review)
recheck-of: 122
repo: /data/lanes/lane-4104
branch: issue-4104-frozen-scalar-not-declarable
base: origin/main
head-sha: a1cad94504e2c8c4679da10d0079c4d7d26dbc09
reviewed-sha: 2a57a5b186bcf63420abd5aab4108c556ce92b7e..a1cad94504e2c8c4679da10d0079c4d7d26dbc09
assert-base: 2a57a5b186bcf63420abd5aab4108c556ce92b7e (merge-base of origin/main and HEAD; origin/main tip 2a57a5b186bcf63420abd5aab4108c556ce92b7e)
job: 122
model: gpt-5.6-sol
census: 35 files, +6782/-1510
tokens: input=592081 cached=508672 output=2702
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: WAIVED (35/35 code census paths absent — authorized by @pmcfadin for base=2a57a5b186bcf63420abd5aab4108c556ce92b7e head=a1cad94504e2c8c4679da10d0079c4d7d26dbc09 job=122)
waiver: GRANTED (author=@pmcfadin base=2a57a5b186bcf63420abd5aab4108c556ce92b7e head=a1cad94504e2c8c4679da10d0079c4d7d26dbc09 job=122 reason=3252 snapshot-pointer delivery, not an unseen diff: census is 35 files +6782/-1510 but the stored prompt is 7448 bytes with zero occurrences of cqlite-core/src, and the prompt itself references an explicitly provided diff file path, so the diff reached the reviewer by pointer rather than inlined; root checkout .roborev.toml ruled out by inspection, every exclude_patterns entry is prose or artifacts under docs and none matches cqlite-core, src or *.rs, and the forbidden blanket docs/** is absent; the reviewer demonstrably read the code since job 122 returned an accurate source-verified finding naming serialization_header/mod.rs line numbers; token accounting input=592081 cached=508672 output=2702 is 30x above the 18.7k/0 vacuous baseline and roborev vacuity-tier1 and tier2 both PASS; granted by lead on own authority, not the overnight pre-authorization, in PR 4139 comment 5586299818)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: DEFERRED (1, issues=#4179, authorized @pmcfadin, job 122)
deferral: GRANTED (author=@pmcfadin issues=4179 count=1 scope=base=2a57a5b186bcf63420abd5aab4108c556ce92b7e head=a1cad94504e2c8c4679da10d0079c4d7d26dbc09 job=122 reason=unreachable dead code at serialization_header/mod.rs:382 — the guard at :357 compares windows(10) against the 9-byte needle org.apach so equality is never true and every candidate is skipped unconditionally, making the refusal below it unexecutable; no SSTable valid or otherwise can be refused there, so there is no user-visible defect; tracked in issue 4179 together with the activation trap that fixing the needle switches on an unconfirmed enforcement site and so requires the acceptance-point rule in the same change; granted by lead after independent source verification in PR 4139 comment 5584757619)
roborev-exit: SKIP (recheck: no reviewer ran in this invocation; job 122 re-decided from its record)
log: /tmp/roborev-review-issue-4104-frozen-scalar-not-declarable-a1cad945-2038190.log
RESULT: PASS
```

PR #4149, opened 2026-09-07
fix(#4114): vector<float,n> silently mis-decodes — decode it, and refuse honestly what is out of scope

### LINKED ISSUE #4114: vector<float,n> is not readable: classification-only support, no decoder, no CQL type variant, no fixture
Parent: epic **#3479** (Cassandra 5 read-completeness audit). Carved as slice 1 by owner decision, 2026-09-06.

`vector<float, n>` is the Cassandra 5 flagship type. **CQLite has no decoder for it, no CQL type variant, and no fixture.** A stock C5 node can create such a table today, so this is squarely the epic's bar — *"CQLite reads anything Cassandra 5 can create"* — and it fails it.

## Verified state on `origin/main`, with exact anchors

Support is **classification-only**. Two references exist in the whole tree and neither decodes anything:

| site | what it actually does |
|---|---|
| `cqlite-core/src/parser/repair_clustering.rs:135` | lists `"VectorType"` among the **variable-width** clustering layouts. A width classification, not a decoder. |
| `cqlite-core/src/storage/commitlog/schema.rs:234` | `"vector<float, 4>"` appears **only inside a test asserting it must NOT be treated as a simple scalar** — i.e. the code bails on it, and the test pins the bail. |

And the absences, each checked rather than assumed:

- **No `Vector` variant** in any CQL type enum under `cqlite-core/src/types`, `cqlite-core/src/schema`, or `cqlite-core/src/parser/types.rs`.
- **No fixture.** `git grep -i vector -- test-data/schemas/**` returns exactly one hit, `legacy/hardened_validator_test_schema.cql:146`, and it is `vector_value LIST<DOUBLE>` — a list, not a vector. Nothing in the corpus exercises the type.

## Establish this FIRST, before writing any decoder

**Does a `vector<float,n>` column fail CLOSED, or does it silently mis-decode?**

This is the whole risk and it is not yet known. Under the no-heuristics mandate, *refusing* an untyped column is correct behaviour and the gap is merely "unsupported". But if the type name falls through to a bytes/blob fallback, a user reading a real C5 table gets **plausible wrong values with no error** — a correctness defect, not a missing feature, and a much more serious finding than this issue is currently filed as.

Answer it with a real column and report the answer in this thread **before** implementing. If it turns out to mis-decode silently, say so and expect the priority to move.

## Format authority — read Cassandra, never CQLite

A CQLite `file:line` is not format authority. Read the encoding at the pinned tag:

```bash
git show cassandra-5.0.8:src/java/org/apache/cassandra/db/marshal/VectorType.java
git show cassandra-5.0.8:src/java/org/apache/cassandra/cql3/CQL3Type.java
```

Do **not** infer the wire form from the type name, from `LIST<FLOAT>`, or from what CQLite currently does with either. Derive every expectation from that source and cite it.

Note the tension to resolve from source, not by assumption: `repair_clustering.rs` classifies `VectorType` as **variable**-width, while a fixed-dimension float vector is a candidate for fixed-width framing. One of those is wrong for at least some element types. Establishing which is part of this issue.

## The fixture must be Cassandra-written (#3042)

A CQLite-written + CQLite-read round-trip is **invariant to a uniform framing error** — both sides make the identical mistake, the round-trip closes, the test stays green, and real C5 data still reads wrong. That is #3042's lesson and it applies directly here, since there is no existing oracle to disagree with.

So the oracle is **bytes written by a real Cassandra 5.0 node**, with a `sstabledump` JSONL reference alongside it, added to `test-data/`. A synthesized fixture is acceptable only for negative/refusal cases, never for the positive decode.

## Acceptance criteria

1. **Failure mode reported first** — whether a `vector<float,n>` column today fails closed or silently mis-decodes, stated in this thread with the measurement that shows it, before any decoder lands.
2. **A Cassandra-5.0-written fixture** containing at least one `vector<float,n>` column, committed with its `sstabledump` JSONL reference, and resolved per-table via `sstables_root_for_table` (never a keyspace-level root — #3220).
3. **Reads correctly**, with every expectation derived from the pinned `cassandra-5.0.8` source and cited. Cover at least: n=1, a typical embedding dimension, a null vector, and a vector column in both a partition-key-only and a clustered table.
4. **Refusal is honest for what is out of scope.** Whatever element types or shapes are not implemented must fail closed with a named error, never a fallback decode. Pin that with a test.
5. **The `repair_clustering.rs:135` width classification is confirmed or corrected against source**, with a test either way.
6. **Per-case assertions, not a suite-wide `assert!(ran > 0)`** — committed fixtures are `must_run` and fail closed (#3220).

## Routing

**Oracle-driven** — the answer is in Cassandra's source and in real bytes, so no OpenSpec change is needed. GitHub issue + pinned parity test.

Milestone set to **0.18** so 0.17 can close; move it if you disagree. Filed by the coordination lead carving epic #3479.

pmcfadin:
Closes #4114.

`vector<float, n>` is Cassandra 5's flagship type and CQLite could not read it. This adds the
CQL type, a fixed-width decoder, an honest refusal boundary, and the first Cassandra-written
vector fixture in the corpus.

## The headline: this was NOT "unsupported", it was silent data corruption

#4114 gated implementation on a measurement — does a vector column fail closed, or silently
mis-decode? **It silently mis-decoded.** Measured on Cassandra-5.0.8-written bytes before any
decoder landed (`.drive-issue-4114/ac1-measurement.md`,
`.drive-issue-4114/silent-misdecode-measurement.md`):

```
$ cqlite read-sstable .../vector_exact-*/nb-1-big-Data.db --format json
exit code: 0        stderr: no error, no warning
[ { "key": "RowKey([0, 0, 0, 1])", "value": "{v3: 0x0000003f80000040000000}" }, ... ]
```

Cassandra's own `sstabledump` says that row is `[2.4651903e-32, 1.0, 2.0]`. CQLite returned the
**11-byte tail** of the vector's own 12 bytes: the leading `0x0b` was eaten as a vint length,
which happened to be exactly satisfiable, so no bounds check fired.

Two defects compounded. `marshal_type.rs:214`'s `other => other.to_lowercase()` **accepted** the
type by degrading it to the junk string `floattype , 3`; the value was then framed as a
vint-length-prefixed blob, so the first float's leading byte became a length. `cell_value.rs`
rejects only when `len > remaining`, so **the user-visible outcome was decided by one byte of
user data**:

| first element's leading byte | outcome |
|---|---|
| large (`1.0` → `0x3f` = 63) | length overruns → `Data corruption`, exit 5 |
| `0x00` / `0x08` | under-consumes the row body → caught by a row-body guard → exit 5 |
| **`0x0b`** | consumes the body **exactly** → **all guards satisfied → exit 0, wrong value** |

The error users *did* sometimes see said **"Data corruption"** — blaming byte-correct Cassandra
output — and named the **wrong column**, because the real cause had already mis-decoded earlier
in the row.

## Format authority

`.drive-issue-4114/format-authority.md`, derived from the pinned `cassandra-5.0.8` tag and
**verified against the committed Cassandra-written bytes**, not just source:

- `vector<float, n>` is **fixed-width: exactly `4*n` raw big-endian binary32 bytes**, no outer
  length, no element count, no per-element framing (`VectorType.java:86-101`, `:445-460`).
- Fixed **iff** the element type is fixed; `vector<text, 3>` is genuinely variable.
- Marshal spelling is `...VectorType(...FloatType , 3)` — **spaces around the comma**
  (`TypeParser.java:239-242`); Cassandra's reader tolerates whitespace, so this one does too.
- `n > 0`; a **zero-length value is an error** (`MarshalException("Invalid empty vector value")`),
  never an empty vector. Null is legal and distinct.

The byte-level proof, both framings side by side in one Cassandra-written row:

```
08 | 3f 80 00 00  40 20 00 00  c0 70 00 00 | 08 0b | 63 6b ...
^^   ^^^^ vector<float,3>: 12 raw bytes ^^^   ^^ ^^   "ck-after-10"
flags  (no 0x0c length prefix)              flags, vint len 11
```

## AC5: `repair_clustering.rs:135` was a real bug

It classified `VectorType` as variable-width. Per source the serializer is chosen from the
**element type**, so a fixed element means no framing and `valueLengthIfFixed() == 4*n`. A reader
treating it as variable consumes `0x3f` = 63 as a length and reads 63 bytes for a 12-byte value —
an unrecoverable desync. The fix cannot be a flip (`Variable` is right for `vector<text, n>`) and
cannot be decided from the constructor name, since `:88-89` discards the args; it now parses the
element type and multiplies by `n`, returning `None` (honest unknown) for a malformed vector. That
file's own authority note at `:65-78` already stated the correct rule. Nothing at 5.0.8 stops a
vector being a **partition or clustering key**, so the path is reachable.

## Four instances of one defect class, across TWO representations

Review found the fix incomplete three more times. Recorded because the pattern matters more than
the instances:

| site | representation | found by |
|---|---|---|
| `parse_cql_value_for_type_with_registry` — vector in a UDT field / collection element returned a raw blob; empty vector → `Value::Null` | `CqlType` discarded | roborev job 109 |
| `marshal_vector_inner` conflated "not a vector" with "malformed vector" → `to_lowercase()` fallback | marshal string | roborev job 109 |
| string `cql_type_to_type_id` had no `vector<` arm → `_ => Ok(CqlTypeId::Udt)` | type id | lead audit |
| `raw_value/reporting.rs` type-string dispatch → `Value::Blob` | **type string** | lead, late |

**I am not claiming this class is closed.** My own audit concluded there was no fourth instance,
and it was wrong: I enumerated `CqlTypeId::Custom` sites and `parse_cql_value*` callers, which is
a *type-id*-shaped search, and the fourth instance lives in the *type-string* dispatch where no
such grep can reach. A fifth in some third representation is not ruled out by the searches
performed.

The fourth is also the subtlest: the outer `[i32 len]` framing already delimits the value, so the
bad blob is the **right length** — the row does not desync, nothing looks corrupt, and the value
is simply the **wrong type**. A test asserting "no error" cannot see it, so those tests assert the
type.

What is **structurally out of reach**, stated rather than implied: the residual
`CqlTypeId::Custom => blob` arms (`parser/types/mod.rs:209`, `:253`) are reachable only from
schema-less, wire-derived type-id paths (`parse_udt:128`, `parse_tuple:508`,
`collections.rs:84/160/172/433/480/485`). A `CqlTypeId` cannot carry an element type or dimension,
so no vector decode is possible there even in principle; changing them would alter behaviour for
genuine custom types, which is outside this issue.

## Tests

- **`cqlite-core/tests/issue_4114_vector_float_cassandra_golden.rs`** — reads the committed
  **Cassandra-5.0.8-written** fixture. AC3: `n=1`, `n=384` (typical embedding width), a null
  vector, and a vector in **both** a partition-key-only and a clustered table. Expectations come
  from the `sstabledump` JSONL goldens, never from CQLite's own output (#3042).
- **AC6 / #3220**: per-case `must_run`, fail-closed. There is **no** `return`, no skip path and no
  environment branch in that file; a missing fixture is a `panic!` naming the table, and both `0`
  and `>1` matching directories panic. It deliberately does **not** use `sstables_root_for_table`
  — that resolver can be shadowed by `$CQLITE_DATASETS_ROOT` for a committed-fixture-only table —
  and anchors on `CARGO_MANIFEST_DIR` instead, the `issue_3790_collection_order_cassandra_golden.rs:189`
  precedent. Rationale is in the module header.
- `vector_exact` is the silent-mis-decode regression, and asserts the **type**, not just absence
  of error.
- **`issue_4114_vector_dispatch_tests.rs`** — 5 cases pinning the type-string dispatch. Its header
  states its oracle scope honestly: the *framing* is pinned by the Cassandra-golden test; these pin
  the *routing*, and their byte literals are a fixture for that question, not a second format oracle.
- AC4: an unimplemented element type is refused **by name**; four malformed shapes error (unmatched
  paren, missing dimension, non-numeric dimension, `n=0`); a wrong width errors in **both**
  directions plus empty, because Cassandra rejects trailing bytes too.
- Every fix was **failing-first verified** by neutralising it and confirming the test fails with
  the actual defect (e.g. `came back as a BLOB ... Blob(b"?\x80\0\0@ \0\0\xc0p\0\0")`).

## Wiring evidence (`.drive-issue-4114/wiring-evidence.md`)

```
before:  {v3: 0x0000003f80000040000000}   exit 0, no error   <- silent wrong value
after:   {v3: [2.4651903e-32, 1, 2]}      exit 0             <- matches Cassandra's golden
```

## Two disclosures, so neither reads as clean

1. **`file-size: OPT-OUT`, not `PASS`.** `CQLITE_ALLOW_FILE_GROWTH=1` is engaged: 9 files that
   were **already** over threshold (857-3255 lines) grew by the match arms that adding an
   exhaustively-matched `CqlType` variant makes mandatory. `cqlite-core/src/query/result.rs` grew
   by **one line** and still failed the ratchet — which is the proof that no tightening gets under
   it. The compliant alternative is splitting nine unrelated files, one of them 3255 lines and
   owned by epic #1116; that is a separate refactor and would break 1:1:1:1. Per #3402 the row is
   textually distinct and non-failing, so this is disclosed by construction — flagging it here too.
2. **roborev round 1 (job 109) reported `prompt-content: FAIL (61/61 code census paths absent from
   the prompt)`.** Diagnosed per CLAUDE.md's order: `.roborev.toml` **ruled out** (its
   `exclude_patterns` cover only `*.md` and docs artifacts; checked in both this worktree *and* the
   root checkout the daemon actually reads), leaving **#3252** — a large diff (79 files, +4829)
   delivered by snapshot pointer. That review was therefore **not full-coverage**, and a later
   roborev PASS on this PR must not be read as "all code reviewed". It still produced two genuine
   blockers, both fixed.

## Follow-up (not folded in, per the discovered-bugs rule)

`marshal_type.rs`'s `other => other.to_lowercase()` and `cql_parser.rs`'s
`_ => Ok(CqlTypeId::Udt)` both **guess a type from the absence of a match**, and will mis-handle
any unknown type, not just vectors. Closing that repo-wide touches unrelated types, so it gets its
own issue at merge with these measurements attached.

## Gate

`--lite` `RESULT: PASS` at `84ad565854e9` — `tree-integrity: PASS`, `tree-start == tree-end`,
`dirty: no`, 4307 tests / 171 binaries. **The full gate of record has not run yet**; `flow-closer`
runs it once, after rebasing onto current `main` (which has advanced several times), then C, then
roborev last.


pmcfadin:
Terminal roborev block for the tree being merged (`f215a9c2e`), posted so `premerge-assert`'s `review-binding` leg can read the job id. Findings are **NONE**; `RESULT: FAIL` rests solely on `prompt-content` (#3252) — see `COORD-REQ id=4114-4` on #4114.

```
repo: /data/lanes/lane-4114
branch: issue-4114-vector-float-read-completeness
base: origin/main
head-sha: f215a9c2e8bde76e6427b35007b4d547e03aca7c
reviewed-sha: 98654fdfdf0ac406f74f986158bd15c1bdc89d3c..f215a9c2e8bde76e6427b35007b4d547e03aca7c
assert-base: 98654fdfdf0ac406f74f986158bd15c1bdc89d3c (merge-base of origin/main and HEAD; origin/main tip 73089a45db9b11d7e22ad22693fa96a7dbac0943)
job: 118
model: gpt-5.6-sol
census: 87 files, +5606/-19
tokens: input=1067376 cached=969728 output=3105
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: FAIL (73/73 code census paths absent from the prompt)
waiver: NONE (no waiver comment for this review: the marker must be the SOLE NONBLANK CONTENT of a TOP-LEVEL PR comment — a marker inside prose, a code fence, a quote or a review body is not read; linked issue #4114 checked: no matching marker there either (NON-EXHAUSTIVE: a thread counts as read when its payload was a comments LIST the scanner accepted; a malformed ENTRY inside an otherwise well-formed list is skipped by the scanner and is not distinguished here))
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: NONE
roborev-exit: PASS
log: /tmp/roborev-review-issue-4114-vector-float-read-completeness-f215a9c2-2745634.log
RESULT: FAIL
```

Issue #3928, opened 2026-09-02
Partition-HEADER decode errors are still swallowed with a byte-resync, so a header-byte corruption can silently drop AND fabricate partitions


pmcfadin:
**Class:** correctness / silent data loss. **Split out of #3782** as a declared residual, and the sibling of #3922.
**Routing:** oracle-driven → GitHub issue + pinned test. No OpenSpec change.

## The gap

#3782 made a fatal **row** decode error refuse instead of silently emitting a partial partition, on the drivers (`at_final_chunk`) and on the block-emit path (a proven-complete stitched buffer). **The partition-HEADER arm was not changed and still tolerates.**

- `cqlite-core/src/storage/sstable/reader/parsing/row_decoder/block_emit_windowed.rs:686` — `tracing::warn!` ("Skipping malformed partition header") then `offset += 1` and re-tries.
- same shape at `.../partition_driver.rs:211-218`.

So on a buffer #3782 has *proven complete*, the row arm refuses while the header arm byte-resyncs. Two consequences, and the second is the bad one:

1. the malformed partition is **silently dropped**;
2. the one-byte resync can land on **misaligned bytes that parse as a plausible header**, so the walk **invents a partition that does not exist**.

That is exactly the mechanism #3782 measured: one corrupt byte in a real Cassandra fixture made `iterate_all_partitions_for_compaction` return **`Ok` with 102 rows — 2 real partitions lost and 3 fabricated**. The row-arm half of that is now closed; **this half is not**, and it was the larger share of the read-path loss: the #3782 probe attributed **15** of 35 corrupted losses to the header arm (`block_emit_windowed.rs:169`) against 20 to the row arm.

**Note the count goes UP while data is lost.** Any check of the form "did the row count shrink" is blind to this by construction — as is any parity oracle comparing totals.

## Why it was not fixed in #3782

#3782's subject was `SlidingPartitionPolicy::on_data_row` and the row-framing propagation; the header arm is a different family (partition-header framing and resynchronisation) reached from a different site, and folding it in would have put two propagation mechanisms in one signature-changing diff. It is declared in that change's residual note rather than left to be rediscovered from a green suite.

## Acceptance criteria (testable)

1. On a **proven-complete** buffer (the `with_complete_buffer` / stitched-section contract #3782 introduced), a malformed partition header REFUSES rather than resyncing — the tolerance must remain only where the buffer is a chunk-covering window and a header can legitimately straddle it.
[...]

Issue #3078, opened 2026-07-28
Four #2163 observability counters are dead on the production compaction path (merge.rows_in/out, tombstones_suppressed/emitted) — and no CI lane can see it


pmcfadin:
## Summary

`cqlite.merge.rows_in` / `cqlite.merge.rows_out` (and almost certainly `compaction.tombstones_suppressed` / `_emitted`) are **never emitted by the production compaction path**. The emission code exists and the counters are correctly registered — but the only function that emits them is no longer called by production compaction.

Confirmed against post-merge `main` (worktree at `46a9b42`, which contains #3002's merge `14b35d5`). **This is NOT a #3002 regression** — #3002 merely surfaced it.

## Evidence

```
cargo test -p cqlite-core --features observability-testing,cli-helpers,write-support \
  --test correctness_signals_2163 -- --nocapture
```
→ **FAIL** (exit 101), `correctness_signals_2163.rs:155`:
`assertion left == right failed: cqlite.merge.rows_in must equal the 4 rows consumed at reconcile; entry: None / left: 0.0 / right: 4.0`

Critically, the **preceding catalog-integrity block passed** (lines 115-142: direct `obs::add_counter` + registered-unit checks), proving observability was fully live in that run. And `--test observability_correctness` passes 6/6, so the harness is healthy. The failure is specific to the production path.

An independent read-only probe replicating the test's `run_overlapping_merge` + `compact` confirmed the merge really runs (`round 0: completed_merges=1`) and captured: `cqlite.write.partitions`, `cqlite.write.bytes`, `cqlite.compaction.rows_merged`, `cqlite.compaction.bytes_written`, `cqlite.compaction.sstables_in/out`, `cqlite.compaction.duration` — with **no** `cqlite.merge.rows_in/out`, `tombstones_suppressed`, `tombstones_emitted`, or `tombstones_purged`.

## Root cause

- Emission lives at `cqlite-core/src/storage/write_engine/merge/mod.rs:3435-3446`, inside `KWayMerger::merge_partition_rows` (declared :3150), reachable **only** via `KWayMerger::step()` (:3000, call at :3047).
- Production compaction (`maintenance_step` → `merge::StreamingMerger::resume` in `storage/write_engine/maintenance.rs`) does **row-streamed** reconciliation and never calls `merge_partition_rows`. It has its own `COMPACTION_TOMBSTONES_PURGED` emission at `merge/streaming.rs:728`, but no counterpart for the four #2163 counters.
[...]

Issue #2322, opened 2026-07-10
sequential_scan: byte-pattern header-resync in parse_block_emit_windowed silently loses rows on corrupt partition body (10/50 in repro)


pmcfadin:
Found during #2302 round-4 review verification (branch issue-2302-written-index-resolve @ 204d2e0d).

**Repro (pinned in #2302's test as the consistency baseline):** corrupt one partition's row-flags byte in a 50-partition CQLite-written BIG SSTable and re-checksum CRC.db (crc32fast) so the corruption reaches the parser. `sequential_scan`'s shared decoder (`parse_block_emit_windowed`) does a **byte-pattern header-resync** that drifts and silently loses **10/50 rows** — not just the 1 corrupted partition, and with no WARN/error.

**Why it matters:**
1. Silent data loss on corrupt input — the reader should either recover exactly the undamaged partitions (structural recovery) or fail loudly; losing 9 innocent partitions silently is the worst shape.
2. The resync is a **byte-pattern heuristic** — a #28 no-heuristics concern in the core read path.
3. #2302's new index path now refuses corrupt bodies loudly and falls back to sequential_scan — making sequential_scan the corruption-handling surface of record, so its behavior on corrupt input is now load-bearing.

**Scope note:** #2302 (PR pending) pins path-consistency (auto-routed fallback == forced sequential_scan on identical corrupt bytes) but explicitly does NOT fix this — the resync predates #2302.

Suggested AC: pinned test = same corruption shape → exactly the 1 damaged partition affected (or a loud structural error), zero silent loss of undamaged partitions; resync logic replaced with an authoritative-boundary walk (Index.db offsets where present) or removed in favor of fail-loud.

PR #4162, opened 2026-09-07
fix(#4159): a metadata refusal reaches the caller — an unreadable SSTable is an Err, never Ok(empty)

### LINKED ISSUE #4159: A scan over an UNREADABLE SSTable returns Ok(empty), not Err: a metadata refusal makes the whole partition vanish silently (#3721 swallow class, one layer up)
## The defect

**A scan over an SSTable that cannot be read returns `Ok(vec![])`.** No error reaches the caller. The caller cannot distinguish *"this table is empty"* from *"this table is unreadable and I gave you nothing."*

Measured on branch `issue-4104-frozen-scalar-not-declarable` at `cd44bc52b`:

* The serialization-header parse **legitimately refuses** (the header type is invalid — see #4158).
* The error **is raised and propagated** correctly out of the header parse.
* It is then **swallowed downstream**.
* `manager.scan(...).expect("Scan should succeed")` **succeeds** and returns **0 rows**.

Surfaced by `type_coverage::test_type_frozen_udt` (`cqlite-core/tests/write_read_roundtrip.rs:533`), which reports `Expected exactly 1 row … got 0` rather than a decode error.

## Why this is the more urgent half, and worse than the fail-open it replaced

This is the **#3721 swallow class** reappearing one layer up. #3721 closed the case where row assembly swallowed complex-column decode errors and truncated a row; this is the same shape at **partition/SSTable** granularity.

**A metadata refusal that makes an entire partition vanish behind `Ok` is worse than the fail-open it replaced** — because the fail-open at least returned *something*. Silent zero is the failure mode nothing downstream can detect: not a caller, not a test, not a supervisor.

It is also what converts a **correct** reader refusal into **silent data loss**. #4104's Gate 2 refusal is right and should stay. But while this swallow exists, every correct refusal presents to a user as *"your data isn't there"* instead of *"this file is malformed."* That inverts the value of fixing fail-opens: each one we close becomes a silent-empty instead of a loud error.

**Fix this before, or with, #4158.** Once a refusal is loud, refusing bad metadata is strictly better than guessing at it. Until then it is arguably worse.

## Not limited to this trigger

The trigger here happens to be #4158's bad writer type, but **the swallow is generic**: any header/metadata refusal on any SSTable takes the same path. Do not scope this to frozen UDTs.

## Acceptance criteria

1. A scan over an SSTable whose metadata parse refuses returns an **`Err` naming the cause**, not `Ok(vec![])`.
2. **The `Ok(empty)` and `Err` cases are distinguishable at the public API** — a caller can tell "empty table" from "unreadable table" without inspecting logs.
3. **A RED control**: the new assertion fails against current `main`, proving it discriminates. This is required, not optional — a test that would pass either way tells us nothing about a swallow.
4. **Find the swallow site and state the audit's reach.** Name every place between the header parse and the scan's return that can convert an `Err` into an empty success, and say which were checked. A fix at one site with four siblings unexamined is how #3721 came back.
5. **No dataset-dependent test may pass on an empty corpus** — 0-rows-when-present is a failure (project rule). This defect is that rule violated from the inside, so the new test must assert per case, not with a suite-wide `ran > 0`.
6. Log level is not a remedy. Raising the `debug!` to `warn!` does not satisfy any AC here — the contract is the **return value**.

## Provenance

Found by lane-4104 on #4104, answering a lead question about whether the vanished row was dropped silently or errored. Its answer, verbatim: *"The error was raised and propagated — then swallowed downstream, surfacing as data loss. The test's `manager.scan(...).expect("Scan should succeed")` succeeded and returned 0 rows. So: header parse legitimately refuses → the SSTable becomes unreadable → scan returns `Ok(vec![])` with no error reaching the caller. That is the #3721 swallow class and it deserves its own issue regardless of #4104."*

Agreed, and filed on that basis.

Related: **#3721** (the same class at row granularity, closed), **#4158** (the writer defect that triggered it here), **#4104** / PR #4139 (whose correct refusal exposed it), **#3890** (a swallowed cell decode hiding a truncated point read — the same family).

pmcfadin:
## The defect

A scan over an SSTable whose metadata parse **legitimately refuses** returned `Ok(vec![])`. The refusal was raised and propagated correctly out of the serialization-header parse, then swallowed downstream: `manager.scan(...).expect("Scan should succeed")` succeeded with 0 rows. A caller could not tell *"this table is empty"* from *"this table is unreadable and I gave you nothing"* — the #3721 swallow class at SSTable/partition granularity, and the failure mode nothing downstream can detect.

## AC4 first — the census, because it is the load-bearing AC

#3721 came back because siblings went unexamined, so this is stated before the fix.

**The trigger is a COMPOSITION of three sites, not one.** Fixing any single one leaves the swallow intact:

| # | site | shape |
|---|---|---|
| 1 | `parser/enhanced_statistics_parser/encoding_stats.rs:82-90` | a `parse_serialization_header_schema` refusal `warn!`ed and replaced by a **marker-search heuristic fallback** |
| 2 | `storage/sstable/manager_open.rs:334-337` **+ sibling `:240-243`** | `Err(_) => { warn!(…) }` discards the error *value*; the reader is never registered; `new` returns `Ok` |
| 3 | `storage/sstable/mod.rs:750-756` | empty reader list ⇒ `return Ok(Vec::new())` |

Site 1 was not in the issue's framing and is the sharpest of the three: **that fallback was not behind `legacy-heuristics`**, so it was a live #28 no-heuristics violation *as well as* a swallow — and while it stood, "the header refused" was not reliably observable anywhere for sites 2-3 to propagate. Site 2's sibling is why this had to be one diff: fix only `:334` and the swallow returns via `new_from_discovered_paths`.

**Call chain.** `SSTableManager::new` → `load_existing_sstables` → `find_data_files` → `open_reader_with_schema` → `SSTableReader::open_*` → `open_inner` → `load_statistics_reader` (propagates via `?`) → `StatisticsReader::open` → `parse_enhanced_statistics_file` → `parse_minimal_encoding_stats` → **`parse_serialization_header_schema`** (15 refusal sites). Then `SSTableManager::scan` → `scan_with_meter` → `resolve_reader_snapshot` → `resolve_reader_list … unwrap_or_default()` → the empty-list return.

**Counts, affirmatively.** 507 distinct files swept over three sweeps; ~2,234 raw pattern hits; **41 real swallow sites** classified — **10** on the header-refusal path, **18** able to produce an empty success, **2** the reproduced trigger. Measured zeros, distinct from unmeasured: `unwrap_or(vec![])`/`unwrap_or(Vec::new())` **0 repo-wide**; `.ok().flatten()` **0 in scope**; **0** real swallows under `query/**`, `storage/mod.rs`, `partition_access_weight.rs`.

**Positive controls passed** — a sweep without one is untrustworthy. The patterns independently re-found `manager_open.rs:334`, `mod.rs:755` *with* its intervening `debug!`, and the `:240` sibling. A third control: **#3721's own site is already fixed** (`row_data.rs:837-842` propagates), so this is that class at two coarser granularities, not a regression of it.

**Reach and limits, stated rather than implied.** This was a **textual** audit. `Result`-shaped swallows are measured over 507 files with named patterns and published hit counts; five read-path domain enums were enumerated non-textually; every `file:line` asserted here was read on disk. It ran **no clippy, no build and no dynamic repro**; it did **not** sweep `write_engine/**`, `delta_scan/**`, bindings, flight or `tools/`; and it makes **no claim** that domain-enum swallows outside those five enums do not exist.

## The change

`Error::UnreadableSSTable { table, path, refused, source: Arc<Error> }` — a dedicated matchable variant carrying the **original error value**, not a re-synthesised message. (#3814's named residual `big_promoted.rs:108` is exactly the "propagates but discards the cause" mistake; this variant exists to not repeat it. `#[source]` on `Arc<Error>` chains correctly through thiserror's `as_dyn_error` via `Deref`, so `source()` yields the inner `Error` and keeps chaining.)

`SSTableManager` records SSTable-open refusals per table (`storage/sstable/refusal.rs`); read surfaces consult the ledger and refuse. **`SSTableManager::new` deliberately does NOT propagate** — failing the whole open would make one corrupt file render every unrelated table unreadable, which no AC asks for.

**The partial case errors too:** one generation readable, one refused ⇒ `Err`. A partial answer presented as complete is the same silent-data-loss defect one degree weaker.

**Site 1's heuristic was deleted, on a measurement — not on principle alone.** Removing it could have broken real reads if genuine Cassandra-written SSTables depended on the marker search, so that was measured, not assumed: with all three heuristic edges gone, `smoke-test-all-tables.sh` is **51/51 enforced tables PASS** (re-measured 3×, including after the deletion) and `cqlite-core --lib` 3992 pass. The marker search is not load-bearing on real Cassandra data (~1,650 lines removed, taking three audit sites with it). It was measurably what hid the refusal: before removal, a `Statistics.db` truncated to half **and** one with an invalid-UTF-8 pk marshal type both still "parsed successfully".

## AC3 — the RED control, with its nuance stated exactly

Run at `f22ce842b` in a worktree with its own `--target-dir` (a shared target dir serves a stale binary and would have measured nothing), the test file copied in and `assert_unreadable` reduced to "Err at all" since the variant does not exist pre-fix:

`cargo test -p cqlite-core --test issue_4159_unreadable_sstable_scan_refusal --features write-support,state_machine` → **7 of 8 FAILED**.

**Only ONE of those failures was the swallow assertion itself** — `a_statistics_db_cut_inside_its_outer_header…`, on the load-bearing line `TruncatedToOuterHeader: surface 'scan' answered Ok with 0 row(s) over an UNREADABLE SSTable`. The other **six failed on the `stage()` precondition**, because the marker-search heuristic made the damaged `Statistics.db` parse *successfully* pre-fix — a metadata refusal could not even be **staged**. That is evidence of the defect chain (and a sharper indictment of site 1 than the issue made), but "7 cases proved the swallow" would be false, so it is not claimed. The one case that passed — `a_genuinely_absent_table_still_answers_ok_empty` — is the case that must not change, and does not.

Transcript retained at `.red-control-4159-report.md` in the lane.

## Acceptance criteria

1. **Satisfied** — a refused metadata parse returns `Err` naming the cause. The message names the cause rather than a code: *"the partition key type is not valid UTF-8 … at byte 42 of the header body"*, not `code: Verify`.
2. **Satisfied, no signature change** — the path is `Result` end-to-end (`SSTableManager` → `StorageEngine` `storage/mod.rs:364` → `SelectExecutor` → `Database::execute` `lib.rs:479` → CLI `commands/query.rs:340-342`); no layer flattens `Err`. `Ok(empty)` now means genuinely-no-data and `Err` means unreadable. Mapped in **both** binding surfaces (Python `error.rs:396`, Node `:506`) and given its cross-binding contract row (39→40) — a new error type that does not map surfaces to users as something generic.
3. **Satisfied** — above, with the nuance stated.
4. **Satisfied** — the census above, reach included.
5. **Satisfied** — all 13 cases write their own fixtures into a `TempDir`, so every case is `must_run` unconditionally. No corpus loop, therefore no suite-wide `ran > 0`.
6. **Satisfied** — the contract is the return value throughout; no AC is claimed on a log-level change.

### Per read surface — does a metadata refusal reach the caller?

#3814's lesson was that an AC green on six surfaces was false on a seventh, so this is enumerated rather than generalized. **YES, asserted:** `scan`, `scan_with_cell_metadata`, `scan_partition`, `scan_partition_clustering`, `scan_partition_with_cell_metadata`, `get` (both feature builds), `scan_stream`, `scan_stream_batched`, `StorageEngine::scan`, `StorageEngine::scan_partition_clustering_reverse`. Compaction (`iterate_/stream_all_partitions_for_compaction`, the KWayMerger producer) **already** propagated (`SSTableReader::open(..).await?`, `producer_iter.rs:460`) — verified, unchanged. Out of class and annotated in code: `partition_key_shape` (Option, metadata), `scan_stream_materializes` (bool), `partition_access_weight::snapshot_for_probe` (instrumentation).

## Review-driven fixes (both reviews ran BEFORE the gate of record)

Two blockers, and **both reviewers independently found the same defect at the same line** — the strongest signal of the round:

- **`refresh.rs:284` — a repaired SSTable stayed refused forever.** `retain_present` cleared a refusal only when the generation *disappeared* from disk. A generation repaired in place (opened mid-rsync, copy later completes) was re-opened successfully by refresh step 4, but step 5.0 retained the ledger entry ⇒ the table errored for the life of the process with every generation healthy and open. The ledger had no other exit: while the file was still broken, step 4 aborted refresh with `?`, so `retain_present` never ran. Predicate is now "still present **AND not re-opened by this refresh**", plus the repair-in-place test whose absence hid it.
- **`discovery_walk.rs:49` (+`:120`, `discovery/scanner.rs:184,206`) — a regression this diff introduced.** A `read_dir` failure on *any* subdirectory aborted the whole walk: a root-owned `lost+found` at mode 0700 (present on essentially every ext4 data volume) made the constructor return `Err` and **no table readable**, where before every readable table opened. It also contradicted this PR's own principle (`refusal.rs:17-24`): the per-file axis recorded, the per-directory axis aborted.

**Resolution (lead-confirmed ruling, `4159-B-walk-incomplete`):** an unreadable directory means the **walk was INCOMPLETE** — record it, don't abort.

| state | answer |
|---|---|
| discovered, all generations readable | `Ok(rows)` |
| discovered, ≥1 generation refused | `Err` |
| **not** discovered, walk COMPLETE | `Ok(empty)` — genuinely absent |
| **not** discovered, walk INCOMPLETE | `Err` — absence is not honestly claimable |

That is this issue's own empty-vs-unreadable contract one level up, and it keeps `lost+found` harmless for the case that matters: a table you actually query is discovered and opens fine. Explicitly **rejected**: recording the directory failure as `UNATTRIBUTED_TABLE_KEY`, which would poison every table and make a stock `lost+found` refuse all reads.

Also fixed: `io::Error::other()` discarding the original `ErrorKind` (so `PermissionDenied` and `NotFound` stayed distinguishable — which matters precisely because of the walk change); a doubled cause render (`{source}` interpolated in an `#[error]` string that also had `#[source]`); a typed refusal flattened into `Corruption(String)` one layer below the variant added to preserve it; undocumented CQLite-invented parser bounds; and stale prose in `statistics_corrupt_hardfail_test.rs` that actively misinstructed the next author.

**No test pinned the swallow**, so none needed flipping (unlike #3612's, which #3814 had to invert). Two mechanical updates only: a removed argument at two call sites, and the FFI row-count pin.

## Named residuals — NOT fixed here, deliberately

- **`partition_header_arm.rs:251-266` + `block_emit_windowed.rs:744-758`** (`Ok(HeaderStep::Resync)` at `BufferExtent::Window`). Refusing here is **blocked**: the refusal is `Error::Corruption`, which `indexed_walk_falls_back` does not recognise, so an index-narrowed point read would hard-FAIL instead of retracting — and it would invalidate #3928's deliberately-built Window positive control (`issue_3928_corrupt_header_refusal.rs:174-191`), whose discriminator is the extent alone. Needs a typed retraction signal first; tracked under **#2322** (pre-existing) and batched in **#4168**.
- `convert_parsed_header_to_sstable_header`'s `unwrap_or("unknown")` — reached only on a *successful* spec parse, and a last-resort table key after two path-derived ones.
- The 1-byte-vs-`writeWithShortLength` 2-byte partition-key-length read in `row_framing.rs:103` — a pre-existing false-`Malformed` for keys ≥256B.
- Remaining low-severity sites from the census, batched to **#4168** — which also carries the discovery-count finding from roborev job 456 (`discovery/scanner.rs:273`: `DiscoverySummary::sstables_found` counts non-regular `*-Data.db` entries, so it can over-report phantom generations; informational only, feeds no read path).

## Not from this diff

`correctness_signals_2163` fails identically at `f22ce842b`, verified in a clean worktree — that is **#3078**. `issue_3805_empty_fixedwidth_map_key.rs` has unused imports at default features (file untouched here).

Closes #4159


pmcfadin:
```
repo: /data/lanes/lane-4159
branch: issue-4159-scan-unreadable-returns-ok-empty
base: origin/main
head-sha: 5031d7f6a8b76d8ef19f86ce24a60e48979d5e05
reviewed-sha: f22ce842bd1ed27e9286ec0433d9c76b880db1bb..5031d7f6a8b76d8ef19f86ce24a60e48979d5e05
assert-base: f22ce842bd1ed27e9286ec0433d9c76b880db1bb (merge-base of origin/main and HEAD; origin/main tip f22ce842bd1ed27e9286ec0433d9c76b880db1bb)
job: 456
model: gpt-5.6-sol
census: 39 files, +5531/-2338
tokens: input=1096597 cached=998144 output=3351
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: FAIL (38/38 code census paths absent from the prompt)
waiver: NONE (no waiver comment for this review: the marker must be the SOLE NONBLANK CONTENT of a TOP-LEVEL PR comment — a marker inside prose, a code fence, a quote or a review body is not read; linked issue #4159 checked: no matching marker there either (NON-EXHAUSTIVE: a thread counts as read when its payload was a comments LIST the scanner accepted; a malformed ENTRY inside an otherwise well-formed list is skipped by the scanner and is not distinguished here))
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: PRESENT (1)
log: /tmp/roborev-review-issue-4159-scan-unreadable-returns-ok-empty-5031d7f6-3071936.log
RESULT: FAIL
```

Issue #2156, opened 2026-07-07
trino-publish: wire automaticRelease (drop manual portal click) + fix dispatch dry_run default


pmcfadin:
Owner-approved 2026-07-07 (resolves the standing open decision from the 0.13.0 publish).

## Two defects in the publish path
1. **Every publish lands validated-but-PENDING** in Central Portal Deployments — `publishToMavenCentral(SonatypeHost.CENTRAL_PORTAL)` has no automaticRelease wired, so the owner has manually clicked Publish for 0.13.0, 0.13.1, and now 0.13.2.
2. **`workflow_dispatch` `dry_run` defaults to `true`** — a dispatch that omits the flag silently publishes only to Maven local while the run reports success (bit us on the first 0.13.2 dispatch: 'success' with the Central step skipped).

## Fix
- Wire automatic release in the connector's vanniktech maven-publish config (DSL `automaticRelease = true` / `mavenCentralAutomaticPublishing=true` / `publishAndReleaseToMavenCentral` — whichever matches the plugin version in use).
- Flip the dispatch `dry_run` default to `false` now that intent is explicit; update the workflow header comment + step names so a skipped Central step can never read as a successful publish.
- Update any docs describing the manual portal step.

## Acceptance
- `./gradlew publishToMavenLocal` still green (dry-run path intact).
- Workflow lint/dry-run of the yaml.
- Real proof lands with the next version publish (expected: no PENDING state in the portal).

PR #4188, opened 2026-09-09
feat(#2869): publish a self-contained cqlite-trino jar as a GitHub Release asset — mergeServiceFiles() alone silently drops gRPC's default load balancer

### LINKED ISSUE #2869: Publish a self-contained (fat/shaded) cqlite-trino connector jar as a GitHub Release asset
## Request

Publish a **self-contained (fat/shaded) `cqlite-trino` connector artifact as a downloadable asset in GitHub Releases** (a "dev" release channel is fine), so downstream consumers can drop a single jar onto disk instead of resolving the connector's transitive dependency tree at runtime.

## Why

easy-db-lab is porting the cqlite kits in-tree (rustyrazorblade/easy-db-lab#731 / PR #860). The maintainer requires the Trino connector be delivered as a **single fat jar, fetched once and mounted per-node** (via `hostPath`), rather than assembled by a Gradle dependency-resolve inside an initContainer at every Trino pod start.

Today that's not possible from what's published:
- `in.mcfad:cqlite-trino:<version>` on Maven Central is a **thin jar** (`compileOnly` on `trino-spi`, normal `mavenPublishing`). It has no runtime scope of its own, so consumers must resolve its full runtime classpath — Arrow `flight-core:19.0.0`, ~11 pinned `io.netty` modules (must stay `4.1.130.Final` — a naive resolve floats to 4.2.x and crashes on first read), `jackson-databind`, netty-tcnative. That's what the current pod-start Gradle `Sync` does, and it's exactly what the maintainer wants gone.

A pre-built self-contained artifact removes the in-cluster Gradle resolve entirely and pins the dependency tree at build time (correct netty versions baked in).

## Proposed shape

- Add a shaded build to `trino-connector` (e.g. `com.gradleup.shadow`) producing a fat jar that bundles the full runtime tree, **excluding `trino-spi`** (engine-provided).
- **`mergeServiceFiles()` must be explicit** — otherwise Trino's `META-INF/services/io.trino.spi.Plugin` discovery file (and Jackson module registration) gets clobbered by the shade merge and the connector won't load.
- Attach the fat jar to a **GitHub Release** (dev channel acceptable) as a downloadable asset with a stable URL pattern, so the kit's initContainer can `curl` it by version. A published Maven classifier (e.g. `:all`) would also work if easier.

## Consumer side (for reference)

The easy-db-lab kit will fetch that one jar into a versioned per-node `hostPath` cache (download-if-missing initContainer, `nodeSelector: type: app`) and mount it into the Trino plugin dir at `/usr/lib/trino/plugin/cqlite_flight/`. No Gradle, no Maven resolution at pod start.

## Notes

- Connector factory name is `cqlite_flight`; SPI is pinned to Trino 481.
- The Arrow path needs `--add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED` at JVM level (handled kit-side, not part of this ask).

pmcfadin:
Closes #2869.

## What this does

Publishes a **self-contained (fat/shaded) `cqlite-trino` jar as a GitHub Release asset**, so a consumer can `curl` ONE jar by version instead of resolving the connector's 50-artifact runtime tree.

Driver: easy-db-lab is porting the cqlite kits in-tree (rustyrazorblade/easy-db-lab#731, PR #860) and its maintainer requires one jar, fetched once, mounted per-node via `hostPath` — not a Gradle dependency-resolve in an initContainer at every Trino pod start.

**Maven Central is unchanged** — it still publishes the thin jar only, no `:all` classifier.

```
release: https://github.com/pmcfadin/cqlite/releases/download/v<version>/cqlite-trino-<version>-all.jar
dev:     https://github.com/pmcfadin/cqlite/releases/download/trino-connector-dev/cqlite-trino-<version>-all.jar
```
Both carry a `.sha256` sidecar — the consumer caches in a per-node `hostPath` across pod restarts, so a checksum is the only way a download-if-missing initContainer distinguishes "cached" from "truncated".

## The issue asked for `mergeServiceFiles()`. That alone ships a broken jar.

Shadow's `duplicatesStrategy` default is `EXCLUDE`, and **it takes precedence over transforming** — the 2nd..Nth copies of a duplicated resource never reach `ServiceFileTransformer`, so the merge silently degrades to first-wins.

Measured on this branch with the merge configured but the duplicates bypass withheld (commit `c903c63d6`, deliberately RED):

```
service descriptor META-INF/services/io.grpc.LoadBalancerProvider lost 1 provider(s)
  [io.grpc.internal.PickFirstLoadBalancerProvider] — sources [grpc-util, grpc-core]; the merge degraded to first-wins
service descriptor META-INF/services/io.grpc.NameResolverProvider lost 1 provider(s)
  [io.grpc.internal.DnsNameResolverProvider] — sources [grpc-netty, grpc-core]
META-INF/io.netty.versions.properties carries no line for resolved module netty-handler-proxy   (…10 of 11)
```

`PickFirstLoadBalancerProvider` is gRPC's **default** load balancer, so first-wins breaks *every* channel. And `io.trino.spi.Plugin` is single-source, so Trino's startup `checkState` passes — **the plugin loads, the catalog registers, and the first query dies.** `append()` degraded identically, leaving 10 of 11 netty modules with no attestation.

Fix: keep `EXCLUDE` globally (correct for LICENSE/NOTICE first-wins) and bypass it only for the transformer-owned paths.

**No relocation**, deliberately: Trino 481's `PluginManager` classloader is child-first except `SPI_PACKAGES`, so the bundled tree is already isolated — and jackson *annotations* must resolve to the engine's copy for `ConnectorSplit` JSON interop.

## Evidence

**Red-first is in history:** `0bf4257e3` (oracle, no shadow → RED) → `c903c63d6` (shadow, no bypass → RED above) → `df23fe97b` (bypass → GREEN).

**Docker E2E — the load-bearing proof.** A contents assertion cannot certify this change; only a query can.

| Run | Result |
|---|---|
| `--plugin-flavor=fat` at tip | **62 PASS / 0 FAIL** |
| `--plugin-flavor=multi` (historical) | **58 PASS / 0 FAIL** |

`62 − 58 = 4` = exactly the fat-only assertions; grep confirms they do not run in multi mode. Includes `PASS: Trino installed CqliteFlightPlugin from the shaded jar`, one-jar-in-container, and every `filterJson`/`aggregationJson`/`limit=OptionalLong[` pushdown probe — the gRPC-dependent paths a bad merge breaks at first query.

**`verifyFatJar` census** (clean `rm -rf build .gradle`, 14/14 tasks executed):
```
cqlite-trino-0.18.0-all.jar (18 MiB) — 11142 ENTRIES EXAMINED; 49 RUNTIME ARTIFACTS RESOLVED, 48 PROBED,
48 CONTRIBUTED, 1 INDISTINGUISHABLE (content-free): io.grpc:grpc-context:1.79.0, 0 UNREADABLE RECOGNISED
(census closes: 49 of 49); 8 MULTI-RELEASE META-INF/versions ENTRIES RECOGNISED under version(s) 9, 11, 17, 21;
8 SERVICE DESCRIPTORS EXAMINED across 2 MULTI-SOURCE PATH(S) carrying 13 PROVIDER LINES RECOGNISED;
11 NETTY CORE MODULES ATTESTED at 4.1.130.Final, 2 TCNATIVE ARTIFACTS at 2.0.74.Final;
6 ARROW MODULES ATTESTED at 19.0.0; 0 SIGNATURE FILES RECOGNISED, 0 INDEX.LIST RECOGNISED,
0 MODULE-INFO RECOGNISED, 0 DUPLICATE ENTRY NAMES RECOGNISED (#2869)
```
Expectations are **graph-derived**, never a hand-list; the contribution census closes arithmetically or fails naming the discrepancy.

**Central purity — proven, not asserted.** `verifyShadowNotPublished` checks three independent signals (variant name, `org.gradle.dependency.bundling` on every variant, any `-all.jar` file entry). Fresh-vs-fresh diff against a pre-shadow baseline: `pom-default.xml` identical, `module.json` identical, thin jar sha256 identical. Positive control: flipping `addShadowVariantIntoJavaComponent = true` fails on all three signals — shadow's *default* would have shipped the 19 MB jar to Central as an extra variant.

**Static checks:** `validate-workflows.rb` 43 workflows exit 0 · `actionlint` clean · `check-workflow-injection.sh` clean · `test_gating_registry_policy.sh` PASS=97 FAIL=0 · `openspec validate --strict` valid.

## Review round — 4 blockers + 4 medium, all fixed with fix-effectiveness controls

- **B1** `verifyShadowNotPublished` never ran before the **irreversible** Central upload. Added to `trino-publish.yml`'s pre-publish oracle step. Structurally verified: oracle is step #5, publishes are #6/#8; 10 steps before/after, every non-authorized step byte-identical to `origin/main`.
- **B4** The release-channel guard asserted the tag *exists*, never that `github.sha` **is** that tag's commit. Pre-fix, `-f version=0.17.0 -f channel=release` with no `--ref` was **accepted** while printing "no tag will be created or moved", then built main HEAD and would have overwritten the genuine v0.17.0 asset *and its sidecar*. Post-fix refuses, naming both shas.
- **B2/B3** Docs and the **normative** spec named `build/plugin/` where the task writes `build/plugin-fat/` — the latter would have failed `spec-auditor` as `unmet`.
- **M1** The dispatch guard bound its refusal to *any* later `exit 1`. Verified empirically: a gutted refusal + unrelated trailing `exit 1` passed at exit 0. Now bound structurally on both ends. 8/8 mutants detected.
- **M2** The reachability check could 200 on the *previous* build's bytes under the same asset name. Now compares downloaded sha256 to the digest this run built.
- **M3** `labeled` events cancelled the in-flight ungated `e2e` via run-level `cancel-in-progress` — triggered routinely by this repo's own `status:*` labels. Both halves fixed (narrowed concurrency *and* the job gate); a job `if:` alone would have made the check silently missing instead.
- **M4** The contribution census could not close on a classifier collision. Root-fixed plus an arithmetic invariant.

Nits batched: none outstanding that change behaviour.

## What is and is not merge-gating — stated plainly

**Exactly one check in this change is merge-gating**: the workflow-policy validator, a step of `pr-gate-core`. Everything else is not:

- `verifyFatJar` / `verifyShadowNotPublished` run every PR in `trino-connector-ci.yml`, which is `required`-exempt → a broken fat jar **can merge**.
- `e2e-fat` is non-gating twice over: exempt workflow **and** full-tier only → it does not run on an unlabeled PR.

Both residuals are **declared** in `.github/ci-gating-tiers.yml` rather than implied away. No new tier was registered: a tier must emit its context unconditionally on every PR, and an always-green classifier for a Gradle+Docker lane would manufacture the exact vacuous green the registry exists to prevent.

**This PR carries `ci:trino-full` so the fat E2E actually runs.**

## Out of scope

The in-repo easy-db-lab kit templates still use the Gradle-resolve initContainer — verifying that port needs a live multi-node k8s cluster, which nothing in this repo's CI can do. Tracked downstream (easy-db-lab#731 / PR #860); a follow-up issue will cover the in-repo templates.

Pre-existing drift flagged, not fixed: `openspec/specs/trino-connector-release/spec.md` still says a secrets-absent tag push "SHALL skip the publish with a visible notice", which #2156 replaced with fail-closed.


pmcfadin:
## Pre-existing gap found while labelling this PR

`ci:trino-full` **did not exist as a repository label** until I created it just now, yet it is referenced by three files — two of which predate this PR:

- `.github/workflows/trino-connector-ci.yml` (pre-existing)
- `.github/workflows/flight-trino-e2e.yml`
- `.github/ci-gating-tiers.yml`

`ci:flight-full` exists; its trino sibling never did. Consequence: **the trino-connector full CI tier has been unreachable on pull requests for as long as that opt-in has been documented.** `contains(github.event.pull_request.labels.*.name, 'ci:trino-full')` can only ever be false for a label nobody can apply, so the `installPlugin` assembly step and its artifact upload have never run on a PR — only on nightly/push-to-main/dispatch.

Not introduced here, and not fixed here beyond creating the label (`1d76db`, description: opt in to the full trino-connector CI tier). Worth noting because:

1. It makes this PR's `e2e-fat` job reachable, which is the load-bearing verification for #2869.
2. Any prior PR that believed it had opted into the trino full tier did not.

No workflow change needed — the label was the only missing piece.

pmcfadin:
## Owner ruling: gate of record waived for this PR — merging on CI evidence

Recording this explicitly because it is a **documented deviation** from the gate-of-record rule, not a silent one.

### Why the local gate cannot certify

A full `scripts/agent-gate.sh` run on the authoring macOS box reaches a terminal `RESULT: FAIL` on four components. **All four are pre-existing and provably unattributable to this PR**, verified by tree identity rather than inference:

- `git diff --name-only origin/main...HEAD -- '*.rs' 'Cargo.toml' 'Cargo.lock' 'rust-toolchain.toml'` → **0**. The Rust tree is byte-identical to `origin/main`.
- All four failing subject files (`scripts/flow/roborev-review.sh`, `scripts/tests/test_agent_gate_logdir_cleanup.sh`, `cqlite-core/tests/issue_3148_fixture_roots_contract.rs`, `cqlite-cli/tests/issue_1490_parquet_golden_admission.rs`) are untouched by this diff.

Root causes are macOS-userland/APFS artifacts: a trailing-slash `TMPDIR` defeating a `T//` vs `T/` string compare; APFS rejecting non-UTF8 filenames with `EILSEQ` so a test cannot build its fixture; unreadable-directory permission semantics; and a `--help` text assertion. Filed with reproductions in **#4189**, alongside the `wc -c` defect that blocks the gate at launch. **Nothing was patched to turn any of it green.**

Since `premerge-assert.sh` requires `RESULT: PASS`, merge-on-green is structurally unreachable from this box. The owner ruled to merge on CI evidence rather than block on a pre-existing fleet-capability gap.

### The evidence being merged on

At the previously audited head, in **CI**:

| Check | Result |
|---|---|
| `required` (aggregate) | green |
| `Workflow policy` — the **one merge-gating check** in this change | pass |
| `gradle test (+ optional plugin)` — ran `verifyFatJar` + `verifyShadowNotPublished` | pass, full census |
| `docker-compose integration (fat plugin flavor)` | **62 PASS / 0 FAIL**, 9m48s |
| `docker-compose integration` (multi, historical) | **58 PASS / 0 FAIL** |

Plus the full-tier artifact uploads reporting **50 files** for `build/plugin/cqlite_flight` vs **1 file** for `build/plugin-fat/cqlite_flight` — executed evidence for the output-root separation.

**C (spec audit): PASS.** All 5 ADDED requirements have executed evidence; Requirement 2 is `partial` with written justification (the release lane is unexecutable pre-merge by construction — `workflow_dispatch` is default-branch-only and the `v*` path needs a real tag — which is precisely why `PUBLISH_DISPATCH_GUARDS` exists, and this change joins that gating table rather than opting out).

### Post-audit hardening (head has moved since the CI run above)

The audit found `verifyFatJar` never asserted the connector's **own** classes: its probes derive from `runtimeClasspath` (which excludes the project jar) and its only project-specific check read a *resource*. Demonstrated concretely — stripping all 84 `in/mcfad/cqlite/` classes while leaving `META-INF/services/io.trino.spi.Plugin` intact **passed the pre-fix oracle**. Now closed, with the expectation derived from `tasks.jar` (no package literal). Also added `verifyInstallPluginFatPrunesStale`, which plants an impossible-version sentinel jar before the sync and asserts exactly one jar survives — closing Requirement 4's previously unexecuted staleness property. Three negative controls recorded, including a `Sync`→`Copy` downgrade.

**CI re-runs at the new head; that result supersedes the table above and is the evidence of record.**

### Open obligation, before this URL is advertised

Per the audit's merge condition: dispatch `trino-connector-fatjar.yml -f version=<x>-dev.N -f channel=dev` immediately post-merge and record the run URL + **served sha256** on #2869 **before** pointing easy-db-lab (easy-db-lab#731 / PR #860) at it. The build/upload/reachability steps are the only part of this change no pre-merge check can exercise.

pmcfadin:
## roborev round 3 — job 16761 (RESULT: FAIL, 2 findings)

Posted for traceability per lead instruction (post the block whatever the verdict).
Certification path for this PR is the owner's route-3 waiver comment on this PR.
This round is **not** a clean review and the merge remains held.

```
repo: /Users/pmcfadin/projects/cqlite-wt/issue-2869
branch: issue-2869-trino-fatjar-release-asset
base: origin/main
head-sha: 312259108a33a907c9c6661505aa3b209eb27b15
reviewed-sha: ff17826c1d800867b6010e149d721f410d6b5005..312259108a33a907c9c6661505aa3b209eb27b15
assert-base: ff17826c1d800867b6010e149d721f410d6b5005 (merge-base of origin/main and HEAD; origin/main tip ff17826c1d800867b6010e149d721f410d6b5005)
job: 16761
model: gpt-5.6-sol
census: 19 files, +2759/-55
tokens: input=770103 cached=664320 output=11352
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (12/12 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: PRESENT (2)
log: /var/folders/5y/bs2cq5v92sqdkq8lks52g__r0000gn/T//roborev-review-issue-2869-trino-fatjar-release-asset-31225910-1654.log
RESULT: FAIL
```

### Findings (both verified against source, both escalated — not decided by the closer)

**C — Medium, `.github/workflows/flight-trino-e2e.yml:60` (concurrency).** All PR events share
`flight-trino-e2e-${{ github.ref }}`. A group holds one running + one pending, so a
`ci:trino-full` run queued behind a running E2E is superseded by a later automated `status:*`
label run, which then skips `e2e-fat` — the requested fat-jar validation is silently lost.
**This is a regression introduced by the `e2e-fat` label-name guard in `f661f1fbf`**: before
that fix the superseding run would still have executed `e2e-fat`. The fix traded duplicate
60-minute stacking for silent eviction.

**D — Medium, `scripts/ci/validate-workflows.rb:389`.** The same-line fast path
(`return true if head_line.match?(/exit\s+1/)`) accepts an `exit 1` occurring *before or
outside* the conditional, e.g. `exit 1; if ! git ls-remote ...; then echo; fi`. This is **not**
the executability limit the owner declared: that declaration explicitly claims the function
"pins the STRUCTURE (which conditional the refusal belongs to)", and on the same-line path it
pins none. Structure was kept in scope by the same ruling that declined executability
(#3725/#3229/#3499), so the declaration currently overclaims.

pmcfadin:
## roborev round 4 — job 16765 (RESULT: FAIL, 3 findings)

Posted for traceability per lead instruction (block posted whatever the verdict).
Certification path is the owner's route-3 waiver comment on this PR. **Merge remains held.**

```
repo: /Users/pmcfadin/projects/cqlite-wt/issue-2869
branch: issue-2869-trino-fatjar-release-asset
base: origin/main
head-sha: f54903f1afdeabeb93870c3cee43b125e7c5052a
reviewed-sha: ff17826c1d800867b6010e149d721f410d6b5005..f54903f1afdeabeb93870c3cee43b125e7c5052a
assert-base: ff17826c1d800867b6010e149d721f410d6b5005 (merge-base of origin/main and HEAD; origin/main tip ff17826c1d800867b6010e149d721f410d6b5005)
job: 16765
model: gpt-5.6-sol
census: 19 files, +2845/-56
tokens: input=596648 cached=478848 output=13313
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (12/12 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: PASS
findings: PRESENT (3)
log: /var/folders/5y/bs2cq5v92sqdkq8lks52g__r0000gn/T//roborev-review-issue-2869-trino-fatjar-release-asset-f54903f1-33107.log
RESULT: FAIL
```

### Classification against the lead's stopping rule

Round history: 16754 (1) -> 16758 (2) -> 16761 (2) -> 16765 (3), every finding verified real.
The stopping rule permits proceeding on plumbing-only findings. **It does not apply here:**
two of these three are in the product list (publish/upload path; Central purity).

**E — Medium, `trino-connector-fatjar.yml:317` — PRODUCT (publish/upload path). Verified:**
the workflow has **no concurrency group at all** (top-level ABSENT; both `preflight` and
`fatjar` jobs ABSENT). Concurrent dispatches/reruns for one version can interleave the
separate jar and checksum uploads, publishing a **jar and checksum from different builds**.
Materially reachable: the `dev` channel target is a mutable prerelease re-uploaded per
build, so repeated dev dispatches are the expected usage.

**F — Medium, `fatjar-oracles.gradle.kts:630` — PRODUCT (Central purity / oracle claim).**
Verified: the task reads only generated `module.json` + `pom-default.xml` and keys on
(a) shadow *variants*, (b) non-`external` bundling, (c) the `-all.jar` *name*. The
plausible-accident path is genuinely covered (the `addShadowVariantIntoJavaComponent = true`
positive control fires). The residual is narrower than roborev implies but real: a shaded jar
attached with a non-standard classifier (`…-shaded.jar`) matches no check, rides an existing
variant with `external` bundling, and would reach Central. The oracle's stated claim is
broader than what it verifies.

**G — Low, `validate-workflows.rb:420` — PLUMBING (lint precision).** The same-line matcher
excludes `;` and `|` but permits `&`, so `<opener> & true || exit 1` is accepted though
`exit 1` guards `true`. Per the stopping rule this is **not** being fixed here; it goes to the
`validate-workflows.rb` self-test follow-up alongside the declared executability residual.

E and F are escalated to the lead rather than decided by the closer.

pmcfadin:
## Owner ruling: roborev waived — merging on CI evidence

Second documented deviation on this PR, recorded explicitly rather than silently.

### Round history

| Round | Job | Result | Findings |
|---|---|---|---|
| 1 | 16754 | FAIL | 1 — `e2e-fat` missing the label-start guard its own file documents. **Fixed** (`f661f1fbf`) |
| 2 | 16758 | FAIL | 2 — dispatch guard unhardened; released jar never E2E'd at tag time. **Fixed + declared** |
| 3 | 16761 | FAIL | 2 — concurrency evicts the requested fat run; same-line matcher pinned no structure. **Both fixed** |
| 4 | 16765 | FAIL | 3 — no concurrency group on the publish path; Central-purity oracle claimed more than it verified; `&` hole in the matcher. **2 fixed, 1 deferred per stopping rule** |
| 5 | — | void | head moved for the `verifyPublishedArtifactSet` snapshot-layout fix |

**Every finding was verified real; none was noise.** Two were regressions introduced by a previous round's own fix, which is why the loop did not converge.

### Why waive now

The **product** — the shaded jar's contents, the four oracles, the publish/upload path, Central purity, and the E2E proving Trino loads the jar — has been unchallenged since round 1. Findings 2–5 were **CI scheduling and lint precision**, with two product exceptions (Findings E and F), and **both are now closed with negative controls**:

- **E** — `trino-connector-fatjar.yml` had no concurrency group, so overlapping dispatches could publish a jar and a `.sha256` **from different builds**. Fixed with a group keyed on the publication target, `cancel-in-progress: false` (cancelling mid-upload is what *causes* the mismatch). Verified by executing the real resolve step against a scratch repo with annotated tags and asserting `group equality ⇔ (tag, version) equality` across all 15 scenario pairs.
- **F** — `verifyShadowNotPublished` keys on shadow variants and the `-all.jar` **name**, so a jar attached as `-shaded` passed it and would have reached **immutable** Maven Central. Now caught by `verifyPublishedArtifactSet`, which asserts the exact primary-artifact set in an isolated audit repository — a decidable set comparison rather than another name pattern. Negative control: `-shaded` **passes** the old check and **fails** the new one.

### The deferred finding

Finding G (Low): the same-line refusal matcher excludes `;` and `|` but permits `&`, so `<opener> & true || exit 1` is accepted though `exit 1` guards `true`. Deliberately **not fixed** under the stopping rule, and going to the `validate-workflows.rb` self-test follow-up alongside the declared executability residual (#3725/#3229/#3499) and a pre-existing `flight_image_guard_errors` looseness where neutering one of two refusals is accepted.

### Certification being merged on

`required` green · `Workflow policy` (the one merge-gating check here) green · `gradle test (+ optional plugin)` running the four oracles · `docker-compose integration (fat plugin flavor)` proving Trino loads the single shaded jar and returns rows · **C (spec audit) PASS**, all 5 ADDED requirements with executed evidence.

Merge will not be armed until `gradle test` and the fat E2E are green at this head — `verifyPublishedArtifactSet`'s snapshot-layout fix has never run in CI, and that is precisely the failure class this PR hit three times (a check verified under one invocation, wired into another).

### The recurring lesson, for the record

Three defects in this PR were the same shape: **a check that looked like coverage and wasn't.** `mergeServiceFiles()` declared a merge that silently degraded to first-wins, dropping gRPC's default load balancer. `verifyFatJar` never asserted the connector's own classes — stripping all 84 while leaving the service descriptor **passed** the pre-fix oracle. `verifyInstallPluginFatPrunesStale` was wired to `gradlew check`, which **no workflow invokes**. Each was found by trying to break the check, not by confirming it passed.

Issue #4214, opened 2026-09-10
roborev guard self-test: assert_says/assert_lacks treat grep rc>=2 as ok and have no empty-match guard (round-2 findings from #4206)


pmcfadin:
Round-2 roborev findings (job 3365, `claude-code`/`claude-opus-5`) on #4206's fix. Not fixed there — deferred here per the 2-round script cap (owner ruling #3893). #4206's acceptance was met independently (self-test passes on macOS and Linux, no guard behaviour change).

Context: #4206 fixed a macOS-only crash in `scripts/tests/test_roborev_review_guard.sh` (BSD `seq 1 0` driving an indirect-expansion loop past its bound) and a related vacuous-pass regex bug (an unescaped trailing `|` in an `assert_says` pattern matched every line under GNU grep and errored under BSD grep). Fixing the crash and re-running roborev surfaced these three findings on the follow-up fix itself.

## Finding 1 — Medium

**Location**: `scripts/tests/test_roborev_review_guard.sh:1259-1265` (`assert_lacks`)

**Problem**: This commit fixes one malformed ERE, but the helper that consumes such patterns still converts a malformed regex into a silent pass. `assert_lacks` runs `if grep -qE -- "$2" "$OUT"; then bad; else ok; fi` — grep exits **2** on an invalid ERE (exactly what BSD grep does with `empty (sub)expression`), which lands in the `else` branch and reports **ok**. So the same class of defect this round is about ("a broken regex that passes") remains live in the sibling helper, and would go undetected on both platforms. `assert_says` degrades less badly (exit 2 → `bad`), but reports the misleading "pattern not found" instead of "invalid regex".

**Fix**: Distinguish grep's exit status in both helpers — capture `grep -qE -- "$2" "$OUT"; rc=$?`, treat `rc -ge 2` as an explicit `bad "<label>: invalid regex '$2'"`, and only treat `rc -eq 1` as "not found"/"absent".

## Finding 2 — Medium

**Location**: `scripts/tests/test_roborev_review_guard.sh:6987-7001` (and the two helpers at :1250-1265)

[...]

Issue #3637, opened 2026-08-30
agent-gate.sh never removes its per-run /tmp/agent-gate.* LOG_DIR: ~61,000 stale dirs across the 3-box fleet in under 3 days


pmcfadin:
## Summary

`scripts/agent-gate.sh` never removes its per-run `LOG_DIR` (`/tmp/agent-gate.XXXXXX`), so every invocation leaves a directory behind permanently. On a long-lived agent box that runs the gate's own self-tests — which spawn one nested gate per case — the count reaches **tens of thousands within a day**.

Measured 2026-08-30 ~22:15 PDT across the three `agent-ami` fleet boxes, all up under 3 days:

| box | `/tmp/agent-gate.*` dirs | actively-progressing | `/tmp` size |
|---|---|---|---|
| box1 | **19,667** | 31 | **15G** |
| box2 | **27,428** | 225 | — |
| box3 | **14,271** | 4 | — |

**~61,000 directories fleet-wide.** The overwhelming majority are dead: on box1, of 19,667 dirs only 31 had any file touched in the last 10 minutes, and the dead ones carry 0–3 `.result` files each — i.e. a gate that started, wrote a component result or two, and exited.

## Why the count is so high

This is not one runaway process. The gate has self-tests (`scripts/tests/test_agent_gate_*.sh`) that invoke `agent-gate.sh` per case — legitimately, and #2874 explicitly supports the nested case with `AGENT_GATE_PARENT_RUN_ID` and its own `summary-primary.txt`. Two lanes running those suites in a loop (#3544 and #3453 tonight) account for the bulk. So the leak rate scales with *how much the gate is tested*, which is exactly backwards.

## Impact, stated with its limits

**This is not currently causing failures, and I am not claiming it is.** `/tmp` is on `/dev/root`, which sat at 17% used on box1 while `/data` (the worktrees and `target/`) was the filesystem under pressure. So the 15G is not what threatens a gate today.

What makes it worth fixing anyway:

1. **`/tmp` shares a filesystem with the agent harness's own stdout/stderr capture** — #3070 finding 3 records that when that filesystem filled, Bash calls returned **no output at all**, i.e. *"the diagnostic channel dies at the same moment as the thing being diagnosed."* A slow unbounded leak on that filesystem is a latent instance of an already-documented incident.
[...]

PR #4215, opened 2026-09-10
fix(#4206,#4220,#4221): macOS gate-of-record fixes — roborev guard self-test, wc padding, EILSEQ skips, $TMPDIR normalization

### LINKED ISSUE #4206: roborev-lints self-test fails on macOS for every diff: test_roborev_review_guard.sh line 1219 '!_rw_i: unbound variable' (BSD seq 1 0)
**Blocking a lane (#4196 salvage) — qualifies for `Ready` under the 2026-09-01 product-first ruling (b): the `roborev-lints` gate component FAILs on macOS for every diff, including a zero-file diff on clean `main`.**

## Repro (clean checkout, macOS 25.6 / zsh, `main` @ 316f9f711)
```
bash scripts/tests/test_roborev_review_guard.sh
…
== case (a): enqueued sha == base ref ==
scripts/tests/test_roborev_review_guard.sh: line 1219: !_rw_i: unbound variable
rc=1
```
`roborev-lints` (a `--lite` component and a full-gate component) runs this self-test, so on this box `--lite` can never reach an all-PASS block and the gate of record would FAIL on an unrelated component. The implementer of #4196 reports the root cause as BSD `seq 1 0` emitting `1 0` (GNU emits nothing), which drives an indirect-expansion loop (`${!_rw_i}`) past its bound under `set -u`.

## Acceptance
- The self-test passes on macOS AND Linux (CI lane + a local run recorded in the PR) with no behaviour change to the guard it tests.
- Replace the `seq`-driven loop with a bash arithmetic `for ((i=1; i<=n; i++))` (or guard `n<1`); add a case to the self-test that runs the loop with `n=0`.
- Roborev capped at 2 rounds (owner ruling for scripts, #3893).

Routing: oracle-driven tooling fix (repro above is the oracle); no OpenSpec. P1 while it blocks #4196's gate of record.

### LINKED ISSUE #4220: agent-gate: component-set probe FAIL-CLOSED on macOS — BSD `wc -c` pads output, so no macOS box can produce a gate of record
## Observed (closer for #4196, 2026-09-11, macOS 26 / Darwin 25.6)
The full gate FAILs in ~4 s with `component-set: FAIL-CLOSED (no-git)` on every bounded component-set probe, exit 198 (`_CS_REPLAY_RC`). `scripts/flow/gate-detached.sh` also exits 69 on macOS (no `systemd-run`).

## Cause
`scripts/agent-gate.sh` ~line 4043: `_cap_n=$(head -c N cap | wc -c)` is validated with a `*[!0-9]*` refusal. BSD `wc -c` PADS its count (`"       3"`), so the value is refused as non-numeric on every macOS host → the probe is treated as unmeasurable → fail-closed. GNU `wc` prints `3` and passes.

## Proof
Same tree, same command, with GNU `wc` first on PATH (`/opt/homebrew/opt/coreutils/libexec/gnubin/wc`): `component-set: ADVISORY-PASS (39/39 names vs origin/main e2ec97b34)`.

## Impact
No macOS lane can produce a gate of record today (blocked #4196's closer > 1 h; product-first Ready criterion (b), #3893). Workaround in use: a PATH shim dir containing only `wc -> gnubin/wc`, no repo file touched, disclosed in the PR body. `gate-detached.sh` needs a launchd analogue of `systemd-run --user` (closer used `launchctl submit`).

## Fix
Strip whitespace from the `wc -c` result before the numeric check (`tr -d '[:space:]'` or `$(( ))`), and audit every other `wc -c`/`wc -l` capture in the gate for the same BSD padding class. Add a macOS-shaped self-test.

Filed by the lead from closer-4196's packet. Related: #4206 (macOS `!_rw_i`, PR #4215), #4216 (codex agent).

### LINKED ISSUE #4221: Full gate cannot PASS on macOS: 3 pre-existing tests fail on APFS/BSD hosts (non-UTF-8 filename tests, logdir self-test path)
## Observed (gate of record for #4196, macOS 26 / APFS, 2026-09-11) — all three reproduce on a clean `main` checkout, none is branch-related

1. `core-tests`: `cqlite-core/tests/issue_3148_fixture_roots_contract.rs:462` `non_utf8_override_is_rejected_fail_closed` panics `create a non-UTF-8 directory: Os { code: 92, ... "Illegal byte sequence" }` — APFS refuses non-UTF-8 path components at `create_dir`, so the fixture cannot be built.
2. `cli-tests`: `cqlite-cli/tests/issue_1490_parquet_golden_admission.rs:1447` `a_directory_entry_the_harness_cannot_read_refuses_the_fixture` panics `write non-UTF-8 named entry: ... "Illegal byte sequence"` — same cause.
3. `tooling-tests`: `scripts/tests/test_agent_gate_logdir_cleanup.sh` fails 3 assertions on macOS: `hermetic: UNEXPECTED surviving directory '/var/folders/.../T//agent-gate-logdir.X/fakeroot/reltd-nested/agent-gate.Y'` vs `EXPECTED survivor '/var/folders/.../T/agent-gate-logdir.X/...'` (the SAME dir, once with `T//` and once with `T/`) plus the resulting `suite floor: only 248 verdicts (floor 259)`. macOS `$TMPDIR` ends in `/`, and the path comparison is textual. Reproduced on `main` @ e2ec97b34 with no cargo involved.

## Impact
With #4206 (fixed in PR #4215) and #4220 (BSD `wc` padding, shimmed), these make `RESULT: PASS` unreachable on any macOS host, and `premerge-assert` requires a PASS gate of record — so NO PR can merge from a macOS lane. Blocked #4196's closer (product-first Ready criterion (b), #3893).

## Fix
- (1)(2): detect an EILSEQ from the fixture-creating call and SKIP LOUDLY naming the host (affirmative "SKIP: filesystem refuses non-UTF-8 names (APFS)"), never a silent pass; keep the assertion on hosts that can build the fixture.
- (3): normalize `$TMPDIR` (strip trailing `/`, or compare via `realpath`/canonical form) in the self-test AND in the gate's logdir bookkeeping, and audit other `"$TMPDIR/..."` textual comparisons for the same class.
- Bundle with #4220 in one PR if convenient — a macOS box cannot certify the fix for one without the others, so they need to land together (or be gated on Linux).

Filed by the lead from the #4196 gate (run agent-gate.Y6cS5u) + a main-checkout reproduction. Related: #4206/#4215, #4220, #4216.

pmcfadin:
## Summary

`scripts/tests/test_roborev_review_guard.sh` crashed with `unbound variable` at line 1219 on every invocation on macOS (`run_wrapper`'s arg-scan used `for _rw_i in $(seq 1 $#)`; BSD `seq 1 0` emits `1\n0` while GNU emits nothing, so with zero extra args — the overwhelming majority of the file's 264 `run_wrapper` call sites — the loop ran with `_rw_i=1` and `${!_rw_i}` indirect-expanded past `$#` under `set -u`). Since this self-test runs inside the `roborev-lints` gate component (both `--lite` and full gate), this made the gate of record unrunnable to completion on this box.

Replaced the `seq`-driven scan with a bash arithmetic `for ((...))` loop, which correctly iterates zero times for `$#=0` on both platforms.

Fixing that crash exposed a second, previously-unreachable macOS-only failure in the same file: case (jd1)'s `assert_says` pattern `roborev show <id> --json | jq '\.job |` has an unescaped trailing `|`, an ERE with an empty alternative. Under GNU grep (every hosted CI lane) this matched every line unconditionally — a **vacuous pass** — while BSD grep (macOS) refuses it outright with `empty (sub)expression`. Escaped both pipes to the intended literal text.

A second commit addressed two of the four Low findings from the first roborev pass on this fix (job 3363): removed a now-dead `--log`-arg scan in `run_wrapper` (no call site among the file's 264 ever passes `--log`; the invocation hardcodes its own `--log` flag immediately before `"$@"`, so the scan was unreachable — and it had a second, sibling out-of-bounds indirect-expansion bug the first pass missed for the same reason nothing exercises it) and corrected the case (jd1) comment to name the vacuous-pass consequence precisely rather than describing it as GNU grep merely "tolerating" the pattern.

No other script under `scripts/tests/` or `scripts/flow/` has the same live `seq 1 $var` pattern (one match in `test_ws0_round_metadata.sh:121` is inside a historical doc comment describing a pre-fix loop, not live code).

## Follow-up

A second roborev pass (job 3365) on the fix-of-the-fix surfaced three further findings (2 Medium, 1 Low) in the same test file's `assert_says`/`assert_lacks` helpers — real, but out of scope for this issue's 2-round script cap (owner ruling #3893) and not blocking #4206's acceptance (self-test passes on macOS and Linux, no behaviour change to the guard it tests). Filed as #4214 (unmilestoned tooling debt, not board-tracked): `assert_lacks` (and to a lesser extent `assert_says`) treats `grep`'s exit status 2 (invalid regex) the same as "pattern not found," so a malformed regex — the exact defect class this issue is about — can still silently pass on both platforms; and the file has no guard preventing reintroduction of an empty-string-matching pattern anywhere across its 48 `grep -qE` assertion sites.

## Daemon infra note (not a code issue)

Two earlier roborev attempts with `--agent codex` (jobs 3357, 3358) failed on a daemon/CLI infra defect, not a code finding: `agent: codex requires --full-auto for stdin input; upgrade codex or use --agentic`. `roborev status` showed the daemon otherwise healthy, and the identical error had also hit an unrelated job (3353) ~3h earlier — a standing `codex`-agent integration issue on this box (`codex-cli 0.144.5`), unrelated to this diff. Certification proceeded with `--agent claude-code --model claude-opus-5` per lead authorization; the codex/daemon issue should be tracked separately by whoever owns that box's roborev config.

## LITE gate (`scripts/agent-gate.sh --lite`)

```
run-id: /var/folders/2c/tfxwt1g91x729x38m63d81540000gn/T/agent-gate.Cu7DzH
commit: 68b71d2 branch: issue-4206-roborev-guard-macos dirty: yes
lite-scope: file-size fmt clippy roborev-lints scoped-tests (full gate NOT run — run it once before merge)
tree-start: 68b71d27704f dirty: yes digest: 3e8efdadccdd
tree-end: 68b71d27704f dirty: yes digest: 3e8efdadccdd
tree-integrity: PASS
clippy:            PASS (4s)  [clippy workspace(excl 5) --all-features | clippy cqlite-core --features 25:all-compression,arrow,bench-internals,+22 more | clippy cqlite-cli --features 8:benchmarks,cli-helpers,delta-export,+5 more | clippy cqlite-flight+cqlite-py+cqlite-node --features cqlite-node/write-support]
roborev-lints:     PASS (370s)  [no-cargo]
scoped-tests:      PASS (22s)  [test cqlite-core --features cli-helpers]  {verified: 4054 tests passed and 0 test binaries built/verified}
disk-exhaustion: 0 RECOGNISED (#3800) -- no non-PASS component to scan (5/5 PASS)
logdir-disposition: REMOVED at exit on PASS
RESULT: PASS
```

(Run against the working tree that became commit 6443355, captured just before that commit — hence `dirty: yes` / `tree-start` naming the prior commit. `roborev-lints` includes a full run of `scripts/tests/test_roborev_review_guard.sh` itself: `passed: 1596 failed: 0` on this macOS box, versus an unconditional crash before this fix.)

## ROBOREV review (round 2, `--agent claude-code --model claude-opus-5`, job 3365)

```
repo: /Users/patrickmcfadin/local_projects/cqlite/.claude/worktrees/issue-4206-roborev-guard-macos
branch: issue-4206-roborev-guard-macos
base: origin/main
head-sha: 64433558256506fda8cc408e5b786b585381ed24
reviewed-sha: 316f9f711b2a146de8dc3c2e1f9d179e53307633..64433558256506fda8cc408e5b786b585381ed24
assert-base: 316f9f711b2a146de8dc3c2e1f9d179e53307633 (merge-base of origin/main and HEAD; origin/main tip 316f9f711b2a146de8dc3c2e1f9d179e53307633)
job: 3365
model: claude-opus-5
census: 1 file, +23/-7
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS (no token accounting in the record)
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (1/1 code census paths present)
vacuity-tier1: PASS
findings: PRESENT (3)
RESULT: FAIL
```

RESULT is FAIL on findings, not infra — see **Follow-up** above (filed as #4214) and **owner ruling** below for why this PR proceeds anyway.

**Owner ruling**: 2-round script cap (#3893) reached; #4206's acceptance criteria (self-test passes on macOS and Linux, no behaviour change to the guard it tests) is independently satisfied. Round-2 findings deferred to #4214 rather than fixed here.

https://claude.ai/code/session_01861dD62mokTbz2ThZj3FRF

## Bundled: #4220 + #4221 (owner ruling 2026-09-11)

Per owner ruling, bundling the remaining macOS gate-of-record defects into this PR so a macOS
host can produce a `RESULT: PASS` full gate. Three additional commits:

### #4220 — BSD `wc` padding broke agent-gate.sh's numeric refusal checks

BSD `wc -c`/`-l`/`-w` (macOS) right-pads its output with leading spaces; GNU `wc` does not. Two
sites in `scripts/agent-gate.sh` case-matched a `wc` capture as a bare digit string
(`''|*[!0-9]*`) — a textual check that always matched the non-digit arm on a padded count:

- `_component_set_bounded`'s stdout cap check (~line 4043): **every** bounded capture on a
  BSD-`wc` host was refused (RC 198), not just oversized ones — this made the bounded-run
  primitive fail-closed on essentially every call on macOS.
- The upstream-identity object census (~line 5813-5814): `_cs_obj_total`/`_cs_obj_absent` were
  reset to their fail-closed defaults on every run, permanently forcing the slow verification
  path instead of the fast object-reuse path.

Fixed by stripping `[[:space:]]` before the case-match in both spots; the refusal itself is
unchanged.

**wc-site audit** (`scripts/agent-gate.sh`, `scripts/flow/*.sh`, `scripts/ci/*.sh`):

| Site | Verdict |
|---|---|
| `agent-gate.sh:4043` (stdout cap check) | **Fixed** |
| `agent-gate.sh:5813-5814` (object census) | **Fixed** |
| `agent-gate.sh:2520,12064,16344,20338,23636,23644,23867,24523,24730-24731,26845,27224` | Safe — already piped through `tr -d ' '`/`tr -d '[:space:]'` |
| `flow/roborev-review-oracles.sh:2311` | Safe — consumed only inside `$(( ... + 1))`, an arithmetic context that tolerates whitespace (verified empirically) |
| `ci/ensure_real_dataset.sh:110-111` | Safe — consumed only inside `[[ $N -gt 0 ]]`, likewise arithmetic |
| `flow/drive-issue-state.sh:1701-1702`, `flow/lane-lock.sh:1066`, `flow/roborev-review-checks.sh:179`, `ci/check-pub-surface.sh:985,987,1363-1364` | Safe — already stripped |
| Every other `scripts/flow/*.sh` / `scripts/ci/*.sh` | No `wc -c`/`-l`/`-w` sites at all |

Added a TDD regression case (`4220-padded-wc-cap`) to
`scripts/tests/test_agent_gate_component_set.sh` that shims `wc` to pad deterministically
(host-independent) and asserts a legitimate small bounded capture still returns RC 0 with its
bytes replayed. Verified RED (RC 198) against the pre-fix script, GREEN against the fix.

**Follow-up, deliberately NOT in this PR**: a launchd fallback for `scripts/flow/gate-detached.sh`
(unrelated capability, would expand scope) — tracked as a follow-up issue to file separately.

### #4221 items 1-2 — SKIP loudly on APFS's EILSEQ instead of panicking

`cqlite-core/tests/issue_3148_fixture_roots_contract.rs::non_utf8_override_is_rejected_fail_closed`
and
`cqlite-cli/tests/issue_1490_parquet_golden_admission.rs::a_directory_entry_the_harness_cannot_read_refuses_the_fixture`
each need a real non-UTF-8-named filesystem entry. APFS refuses to create one at all (EILSEQ, os
error 92), so the fixture-creating call panicked before either assertion ran.

Fixed by detecting exactly that error (`raw_os_error() == Some(libc::EILSEQ)`) at the
fixture-creating call, printing an affirmative `SKIP: ...` line, and returning; any other error
still panics, and a host that CAN create the fixture (e.g. a non-APFS volume on macOS) still runs
the assertion unchanged — decided by the observed error, not `cfg(target_os)`/`#[ignore]`.

### #4221 item 3 — `test_agent_gate_logdir_cleanup.sh` hermetic-accounting + floor

macOS's `$TMPDIR` ends in a trailing slash, so the suite's own `$tmp=$(mktemp -d
"${TMPDIR:-/tmp}/...")` carried a literal `//`; every path built by string concatenation off
`$tmp` kept it, while a path the gate itself reports goes through a real `cd`+`pwd` (which
collapses consecutive slashes) — so the suite's hermeticity enumeration and the gate's own
self-reported paths disagreed on FORM for the same directory, reading as an unexplained leak.
Fixed by normalizing `$tmp` once, immediately after creation (`tmp=$(cd "$tmp" && pwd)`) — the
same "normalise once, at the creation site" principle `scripts/agent-gate.sh:8180-8207` already
applies to its own `$TMPDIR`-derived `GATE_LOGDIR_PARENT` (confirmed: the shipped gate strips
trailing slashes there already, so this defect was test-only, never in the gate itself).

A second, independent defect surfaced by the same run: the suite's flat verdict floor (259)
assumed the macOS owner-marker-capability degradation cost ~15 verdicts relative to Linux's 274;
measured stably at 247 across 10 consecutive runs on this host — 27 short, not 15 — so 259 was
never reachable here regardless of the double-slash fix. Fixed by keying the floor on the same
`OWNER_MARKER_CAPABLE` probe the file's degraded branches already use: 259 unchanged where
present, 243 (4 below the measured-stable 247) where absent.

Audited every other `"$TMPDIR/..."` textual path comparison in `scripts/tests/*.sh` for the same
shape (an expected-vs-discovered path *set* compared across two different derivation mechanisms):
no other file has it. `test_agent_gate_disk_exhaustion.sh`, `test_base_staleness.sh` and
`test_claude_auth_capability.sh` each use `comm -13`/`comm -23`, but over verdict-token lists or
before/after snapshots taken through the *same* mechanism at both ends, not a raw-`$TMPDIR` string
against a subprocess-normalized one.

**Pre-existing, unrelated flake observed (not fixed here, out of scope for item 3)**: `AC11b`
intermittently (~50% of runs on this host, reproduced on the pre-fix script too) reads a
SIGTERMed run's disposition as `RETAINED: early exit status 1 with diagnostic content
(post-mortem)` instead of the evidence-based reason — a signal-timing race. Worth a follow-up
issue if it recurs.

## LITE gate, round 2 (`scripts/agent-gate.sh --lite`)

```
run-id: /var/folders/2c/tfxwt1g91x729x38m63d81540000gn/T/agent-gate.Kg23a2
commit: debb002 branch: issue-4206-roborev-guard-macos dirty: no
lite-scope: file-size fmt clippy roborev-lints scoped-tests (full gate NOT run — run it once before merge)
tree-start: debb0029d32c dirty: no digest: 06a46eb9fdcd
tree-end: debb0029d32c dirty: no digest: 06a46eb9fdcd
tree-integrity: PASS
file-size:         OPT-OUT (0s)  [no-cargo] — CQLITE_ALLOW_FILE_GROWTH=1 (ratchet NOT enforced); 1 over-threshold file(s) grown — see file-size.log under logs:
clippy:            PASS (6s)  [clippy workspace(excl 5) --all-features | clippy cqlite-core --features 25:all-compression,arrow,bench-internals,+22 more | clippy cqlite-cli --features 8:benchmarks,cli-helpers,delta-export,+5 more | clippy cqlite-flight+cqlite-py+cqlite-node --features cqlite-node/write-support]
roborev-lints:     PASS (192s)  [no-cargo]
scoped-tests:      PASS (26s)  [test cqlite-cli default-features | test cqlite-core --features cli-helpers]  {verified: 4308 tests passed and 0 test binaries built/verified}
disk-exhaustion: 0 RECOGNISED (#3800)
logdir-disposition: RETAINED: file-size OPT-OUT disclosure #3402
RESULT: PASS
```

**`file-size` OPT-OUT disclosure**: `cqlite-cli/tests/issue_1490_parquet_golden_admission.rs`
grew from 1497 to 1510 lines (limit 1500) from the item-2 EILSEQ fix. Trimmed the added comment as
far as reasonable without hurting clarity; the file was already 3 lines from the ceiling before
this change and a full split is out of scope for a bug-fix PR (epic #1135, test-file splitting).
Re-ran with `CQLITE_ALLOW_FILE_GROWTH=1` per CLAUDE.md's documented escape hatch.

Also ran the three directly-affected tests by name:
- `cargo test -p cqlite-core --test issue_3148_fixture_roots_contract non_utf8_override_is_rejected_fail_closed` — PASS (prints the SKIP line)
- `cargo test -p cqlite-cli --test issue_1490_parquet_golden_admission a_directory_entry_the_harness_cannot_read_refuses_the_fixture` — PASS (prints the SKIP line)
- `bash scripts/tests/test_agent_gate_logdir_cleanup.sh` — floor met (247 ≥ 243), no hermetic mismatch (AC11b pre-existing flake aside)

Closes #4206. Closes #4220. Closes #4221.

## ROBOREV review (round, `--agent claude-code --model claude-opus-5`, job 3406)

Reviewed the #4220/#4221 bundle (this diff on top of #4206's already-merged-round fix).

```
job: 3406
model: claude-opus-5
census: 6 files, +132/-12
push-assert: PASS
census-check: PASS
code-free: PASS
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (6/6 code census paths present)
vacuity-tier1: PASS
findings: PRESENT (7)
RESULT: FAIL
```

7 findings (3 Medium, 4 Low). Per the roborev-severity rubric and owner ruling #3893 (scripts
capped at 2 rounds; this PR's rounds are already used), fixed the 3 Medium findings in a follow-up
commit and batched the 4 Low findings into a single comment on #4221 rather than running a third
round:

- **Medium** — the `#4221` item-2 EILSEQ skip's `return` exited the whole
  `a_directory_entry_the_harness_cannot_read_refuses_the_fixture` test function, silently
  discarding its second, independent `#[cfg(unix)]` block (the unlistable-directory case, which
  runs fine on APFS). Fixed: labeled block + `break 'non_utf8` scopes the skip to just the guarded
  block.
- **Medium** — the `#4221` item-3 degraded-branch floor (243) had a blind spot: several
  degraded-branch cases (AC15/AC16/AC17/AC20) each emit exactly one verdict, so deleting any one of
  them (247 -> 246) still cleared 243. Fixed: tightened to 247 (measured-stable, no slack).
- **Medium** — the `#4220` regression case's PATH shim hand-duplicated `mkbin`'s tool list and had
  already drifted (missing `gtimeout`, the gate's own macOS bound-mechanism fallback). Fixed:
  single-sourced via `padwc_bin=$(mkbin padwc wc)`.
- 4 **Low** findings (a diagnostic printing a path instead of file content; a dead variable left
  over from #4206's fix; `println!`-based skips invisible without `--nocapture`; a doc comment
  misattributing the double-slash cause) — batched into
  https://github.com/pmcfadin/cqlite/issues/4221#issuecomment-5636597450, not fixed here.

Re-ran `--lite` after the Medium fixes:

```
run-id: /var/folders/2c/tfxwt1g91x729x38m63d81540000gn/T/agent-gate.<r3>
commit: 3263c08 branch: issue-4206-roborev-guard-macos dirty: no
tree-integrity: PASS
file-size:         OPT-OUT (0s)  [no-cargo] — CQLITE_ALLOW_FILE_GROWTH=1 (ratchet NOT enforced); 1 over-threshold file(s) grown
fmt:               PASS (5s)
clippy:            PASS (5s)
roborev-lints:     PASS (179s)
scoped-tests:      PASS (27s)  {verified: 4308 tests passed and 0 test binaries built/verified}
RESULT: PASS
```

Per owner instruction, this PR's script rounds are now exhausted (2 total) — the next gate to run
against this branch is the ONE full `scripts/agent-gate.sh` (gate of record) inside `flow-closer`,
never another `--lite`/roborev round for a scripts change.

---

## Closer addendum — three further macOS host defects found BY the gate of record (#4221)

The first full gate of record on this macOS host (run-id `agent-gate.xC6aX3`, commit `1775c8e`)
completed all 39 components: 36 PASS, `file-size` OPT-OUT (as designed, see above), and **3 FAIL**.
All three were the same class this PR exists to close — the host defeats the harness, not the
product — so they are fixed here rather than deferred.

### F1 — `cli-tests`: two more APFS EILSEQ sites (commit `d3c0dfd`)

`cqlite-cli/tests/support/golden_fixture_staging.rs:225,:327` panicked with
`write: Os { code: 92, ... "Illegal byte sequence" }`. #4221 had applied the loud-SKIP guard to
`issue_3148_fixture_roots_contract.rs` and `issue_1490_parquet_golden_admission.rs` but not to this
third file, so the class was only partly closed. Same guard, unchanged in principle: detect the
FILESYSTEM's refusal from the ERROR, never `#[ignore]`/`cfg(target_os)`. Any non-EILSEQ write error
still panics, so it cannot convert a real defect into a skip.

The second case is skipped **by half, not whole**: only its non-UTF-8 block needs a filesystem that
accepts the name, while the unreadable-ENTRY (symlink) block below it runs anywhere. Skipping the
whole case would have silently dropped that coverage on every macOS run.

| target | before | after |
|---|---|---|
| `cqlite-cli --test issue_1491_json_csv_golden_parity` | 307 passed, **2 failed** | **309 passed, 0 failed** |

### F2 — `python-bindings`: launchd's 256 open-file limit (commit `8a82c0a`)

Ten `test_parity.py` cases failed as `got 0, expected N` — 9 tables returning **zero rows** (all 6
`test_oa`, `test_collections.empty_collections_table`, `test_timeseries.stock_prices`,
`time_bucketed_counters`) plus `TestE2ESummary`.

Root cause is the **launch method**, not the reader and not the corpus. macOS has no
`systemd-run --user`, so a detached gate is launched with `launchctl submit` — and a launchd job
inherits launchd's soft `maxfiles`. Measured on this host:

```
$ launchctl limit maxfiles
	maxfiles    256            unlimited
$ ulimit -S -n          # interactive shell, same host
1048576
```

The corpus was independently ruled out: `fetch-datasets.sh --verify-only` VERIFIES that exact root
(915/915 git-tracked fixtures present, 159 `*-Data.db`), and failing and passing tables are alike
compressed.

**Verification** (the gate-built module, three ways — this is the one-line verification the guard
has in place of a self-test, since the gate self-tests do not cover `run_python_bindings` cheaply):

```
# A. the defect, reproduced
$ bash -c 'ulimit -S -n 256; . target/agent-gate-venv/bin/activate; pytest bindings/python/tests/test_parity.py -q'
10 failed, 36 passed, 1 skipped, 1 xfailed

# B. the guard's raise (soft 256 -> 4096)
$ bash -c 'ulimit -S -n 256; ulimit -S -n 4096; . target/agent-gate-venv/bin/activate; pytest bindings/python/tests/test_parity.py -q'
47 passed, 1 xfailed

# C. fail-closed, when the ceiling makes 4096 unreachable (soft=hard=512)
agent-gate: open-file limit 512 is below the required 4096 and could not be raised — refusing to run pytest (#4221)
exit=97          # pytest never runs
```

Fail-closed is deliberate: the failure being guarded is **silent** (`got 0, expected N`, with no
EMFILE anywhere in the log), and a component reporting a data-parity mismatch for an environmental
reason is worse than one that refuses. The effective limit is printed on the component's line so a
reader of a **green** log can see the headroom it ran with.

`-S` on every `ulimit` call is load-bearing: bash's bare `ulimit -n N` sets the soft **and hard**
limits together, which would permanently lower launchd's `unlimited` hard limit to 4096 for the
gate's process tree — a one-way door. Caught while verifying this guard, when a probe using the bare
form could not raise its own limit back (`cannot modify limit: Operation not permitted`).

**This does not fix the reader.** EMFILE swallowed into a silent empty result, and the handle
accumulation that reaches the limit at all, are a real product defect tracked on its own issue; 4096
is deliberately far above what a correct reader needs, so this guard cannot become the reason a
leaky reader looks green.

### F3 — `tooling-tests`: `$TMPDIR` trailing slash + the `/var` -> `/private/var` symlink (commit `1090a15`)

36 passed / 5 failed, all five in `scripts/tests/test_agent_gate_schemas_preflight.sh`
(`3131-export-line`, `3131-export-quoting`, `3131-warm-cache`, `3131-warm-cache-note`,
`3131-remedy-quoting`). None of this PR's three already-fixed self-tests failed.

Same family as the `$TMPDIR//` normalization this PR already applied to
`test_agent_gate_logdir_cleanup.sh`, with a second discrepancy stacked on it. The cases compare a
path built by string concatenation off `$tmp` against the path the SUBJECT reports, and on macOS
those two spellings of the same directory differ twice: `$TMPDIR` ends in a trailing slash (so the
`mktemp -d` template yields a literal `//`), and `$TMPDIR` lives under the `/var` -> `/private/var`
symlink while `test-data/scripts/fetch-datasets.sh` resolves what it prints with `cd ... && pwd -P`
(`fetch-datasets.sh:149,156,333,338,1154`). So the script printed `/private/var/.../T/x` where the
case expected `/var/.../T//x`.

`pwd -P`, not the plain `pwd` the logdir suite needed — collapsing the `//` alone would still leave
`/var` vs `/private/var`. Normalized once at creation, so no comparison site has to remember.

| self-test | before | after |
|---|---|---|
| `test_agent_gate_schemas_preflight.sh` | 36 passed, **5 failed** | **41 passed, 0 failed** |

Identical (41/0) with a trailing-slash `TMPDIR` and without one — the gate's launchd environment
supplies the latter, which is why only the `/private` half was visible in the gate log.

### Note for anyone running the self-tests by hand

`test_agent_gate_component_set.sh` (144/4) and `test_agent_gate_logdir_cleanup.sh` (247/1) FAIL when
run standalone from an interactive shell, but PASS inside the gate (160/0 and 248/0). The
discriminator is the same trailing-slash `$TMPDIR`: an interactive shell has one, the gate's launchd
environment does not. Pre-existing and environment-dependent, not introduced here and not
gate-blocking — recorded so nobody concludes this PR broke them.

### DIFF
--- a/cqlite-cli/tests/issue_1490_parquet_golden_admission.rs
+++ b/cqlite-cli/tests/issue_1490_parquet_golden_admission.rs
@@ -1444,7 +1444,20 @@ fn a_directory_entry_the_harness_cannot_read_refuses_the_fixture() {
         let unreadable = tmp.path().join(std::ffi::OsStr::from_bytes(
             std::mem::take(&mut raw).as_slice(),
         ));
-        std::fs::write(&unreadable, b"").expect("write non-UTF-8 named entry");
+        // APFS (macOS's default filesystem) refuses a non-UTF-8 path component outright — the
+        // fixture this case needs cannot exist there at all, which is a filesystem property, not
+        // a platform one (a non-APFS volume mounted on a macOS host can still create it). Detect
+        // exactly that error (EILSEQ, "Illegal byte sequence") and skip loudly rather than
+        // #[ignore] or cfg(target_os), which would hide the case on hosts that CAN run it.
+        if let Err(e) = std::fs::write(&unreadable, b"") {
+            if e.raw_os_error() == Some(libc::EILSEQ) {
+                println!(
+                    "SKIP: this filesystem refuses non-UTF-8 path components (EILSEQ) — assertion needs a Linux/ext4 host"
+                );
+                return;
+            }
+            panic!("write non-UTF-8 named entry: {e}");
+        }
 
         let err =
             parquet_parity::fixture_root::fixture_in_table_dir("ks.t", tmp.path().to_path_buf())
--- a/cqlite-cli/tests/issue_1490_parquet_golden_admission.rs
+++ b/cqlite-cli/tests/issue_1490_parquet_golden_admission.rs
@@ -1431,8 +1431,13 @@ fn a_field_the_structure_description_does_not_cover_is_refused() {
 fn a_directory_entry_the_harness_cannot_read_refuses_the_fixture() {
     // A non-UTF-8 entry name: the deterministic instance of "cannot read this
     // entry". It used to be dropped by a `filter_map(|e| e.file_name().to_str())`.
+    // A LABELED block, not a bare one: this function has a SECOND, independent #[cfg(unix)]
+    // block below (the unlistable-directory case), which runs fine on APFS. A bare `return`
+    // from the EILSEQ skip below would exit the whole function and silently discard that
+    // second block's coverage too — `break 'non_utf8` exits only this block instead (roborev
+    // job 3406, Medium).
     #[cfg(unix)]
-    {
+    'non_utf8: {
         use std::os::unix::ffi::OsStrExt;
 
         let tmp = tempfile::TempDir::new().expect("tempdir");
@@ -1454,7 +1459,7 @@ fn a_directory_entry_the_harness_cannot_read_refuses_the_fixture() {
                 println!(
                     "SKIP: this filesystem refuses non-UTF-8 path components (EILSEQ) — assertion needs a Linux/ext4 host"
                 );
-                return;
+                break 'non_utf8;
             }
             panic!("write non-UTF-8 named entry: {e}");
         }
--- a/cqlite-cli/tests/support/golden_fixture_staging.rs
+++ b/cqlite-cli/tests/support/golden_fixture_staging.rs
@@ -225,6 +225,38 @@ mod tests {
         std::fs::write(path, bytes).expect("write");
     }
 
+    /// APFS (macOS's default filesystem) refuses a path component that is not valid
+    /// UTF-8 outright, with EILSEQ ("Illegal byte sequence"), so the fixtures the
+    /// non-UTF-8 cases below need cannot exist there at all. That is a property of the
+    /// FILESYSTEM, not of the platform — a non-APFS volume mounted on a macOS host can
+    /// still hold such a name — which is why it is detected from the ERROR and skipped
+    /// LOUDLY, rather than hidden behind `#[ignore]` or `cfg(target_os)`, either of
+    /// which would also hide the case on the hosts that CAN run it (issue #4221).
+    fn is_non_utf8_name_refusal(e: &std::io::Error) -> bool {
+        e.raw_os_error() == Some(libc::EILSEQ)
+    }
+
+    /// The one line a skipping case prints, so a reader of a green log can tell a case
+    /// that RAN from one this filesystem refused to host.
+    fn skip_non_utf8_unsupported(what: &str) {
+        println!(
+            "SKIP: this filesystem refuses non-UTF-8 path components (EILSEQ) — {what} needs a host whose filesystem accepts them (e.g. Linux/ext4)"
+        );
+    }
+
+    /// `touch` for a name that may not be valid UTF-8. `false` means the filesystem
+    /// refused the NAME itself (EILSEQ); any other write failure is still a panic, so
+    /// this can never turn a real defect into a skip.
+    #[must_use]
+    fn touch_non_utf8(path: &Path, bytes: &[u8]) -> bool {
+        std::fs::create_dir_all(path.parent().expect("has a parent")).expect("mkdir");
+        match std::fs::write(path, bytes) {
+            Ok(()) => true,
+            Err(e) if is_non_utf8_name_refusal(&e) => false,
+            Err(e) => panic!("write: {e}"),
+        }
+    }
+
     /// The discriminating case for the "lexicographically first golden" pick: a
     /// directory holding an EARLIER golden for a generation that is not the one
     /// present. Taking the first sorted golden compared the CLI's reading of
@@ -324,13 +356,24 @@ mod tests {
         // it, because the CLI reading the staged directory certainly will.
         let odd = std::ffi::OsStr::from_bytes(b"nb-2-\xff-big-Data.db");
         assert!(odd.to_str().is_none(), "the staged name is not valid UTF-8");
-        std::fs::write(fixture.join(odd), b"y").expect("write");
-        let why = golden_path(&fixture).expect_err("two SSTables, one golden");
-        assert!(
-            why.contains("holds 2 *-Data.db files") && why.contains("exactly one SSTable per case"),
-            "a name it cannot read is still a second SSTable: {why}"
-        );
-        std::fs::remove_file(fixture.join(odd)).expect("unlink");
+        // Only this HALF of the case needs a filesystem that accepts the name; the
+        // symlink half below runs anywhere. So skip the half, not the case — skipping
+        // the whole test would silently drop the unreadable-ENTRY coverage too.
+        match std::fs::write(fixture.join(odd), b"y") {
+            Ok(()) => {
+                let why = golden_path(&fixture).expect_err("two SSTables, one golden");
+                assert!(
+                    why.contains("holds 2 *-Data.db files")
+                        && why.contains("exactly one SSTable per case"),
+                    "a name it cannot read is still a second SSTable: {why}"
+                );
+                std::fs::remove_file(fixture.join(odd)).expect("unlink");
+            }
+            Err(e) if is_non_utf8_name_refusal(&e) => {
+                skip_non_utf8_unsupported("the non-UTF-8 second-SSTable half of this case");
+            }
+            Err(e) => panic!("write: {e}"),
+        }
 
         // An entry the filesystem cannot DESCRIBE, in the fixture directory and in
         // the keyspace directory beside it.
@@ -376,8 +419,14 @@ mod tests {
             data_db.to_str().is_none() && golden.to_str().is_none(),
             "both staged names are not valid UTF-8, which is the whole subject"
         );
-        touch(&fixture.join(data_db), b"x");
-        touch(&fixture.join(golden), b"{}");
+        if !touch_non_utf8(&fixture.join(data_db), b"x") {
+            skip_non_utf8_unsupported("the non-UTF-8 golden-pairing assertion");
+            return;
+        }
+        assert!(
+            touch_non_utf8(&fixture.join(golden), b"{}"),
+            "the filesystem accepted the SSTable's non-UTF-8 name but refused its golden's"
+        );
         assert_eq!(
             golden_path(&fixture).expect("the golden is right there beside its SSTable"),
             fixture.join(golden),
@@ -387,10 +436,16 @@ mod tests {
         // And the pairing is still by NAME, not "any golden in the directory": a
         // differently-named golden does not satisfy this SSTable.
         let other = tmp.path().join("t-def");
-        touch(&other.join(data_db), b"x");
-        touch(
-            &other.join(std::ffi::OsStr::from_bytes(b"nb-2-\xff-big-Data.db.jsonl")),
-            b"{}",
+        assert!(
+            touch_non_utf8(&other.join(data_db), b"x"),
+            "this filesystem accepted the same non-UTF-8 name moments ago"
+        );
+        assert!(
+            touch_non_utf8(
+                &other.join(std::ffi::OsStr::from_bytes(b"nb-2-\xff-big-Data.db.jsonl")),
+                b"{}",
+            ),
+            "this filesystem accepted the same non-UTF-8 name moments ago"
         );
         golden_path(&other).expect_err("a golden for another generation does not pair");
     }
--- a/cqlite-core/tests/issue_3148_fixture_roots_contract.rs
+++ b/cqlite-core/tests/issue_3148_fixture_roots_contract.rs
@@ -459,7 +459,20 @@ fn non_utf8_override_is_rejected_fail_closed() {
     let mut bytes = tmp.path().as_os_str().to_os_string().into_vec();
     bytes.extend_from_slice(b"/bad\xff\xfedir");
     let raw = std::ffi::OsString::from_vec(bytes);
-    std::fs::create_dir(&raw).expect("create a non-UTF-8 directory");
+    // APFS (macOS's default filesystem) refuses a non-UTF-8 path component outright — the
+    // fixture this case needs cannot exist there at all, which is a filesystem property, not
+    // a platform one (a non-APFS volume mounted on a macOS host can still create it). Detect
+    // exactly that error (EILSEQ, "Illegal byte sequence") and skip loudly rather than
+    // #[ignore] or cfg(target_os), which would hide the case on hosts that CAN run it.
+    if let Err(e) = std::fs::create_dir(&raw) {
+        if e.raw_os_error() == Some(libc::EILSEQ) {
+            println!(
+                "SKIP: this filesystem refuses non-UTF-8 path components (EILSEQ) — assertion needs a Linux/ext4 host"
+            );
+            return;
+        }
+        panic!("create a non-UTF-8 directory: {e}");
+    }
     assert!(
         raw.to_str().is_none(),
         "the fixture value must actually be invalid UTF-8"
--- a/scripts/agent-gate.sh
+++ b/scripts/agent-gate.sh
@@ -4041,6 +4041,11 @@ _component_set_bounded() {
   # branch for a measurement that did not happen.
   local _cap_n
   _cap_n=$(head -c "$((_CS_CAP_MAX_BYTES + 1))" "$_CS_CAP_OUT" 2>/dev/null | wc -c 2>/dev/null)
+  # BSD `wc -c` (macOS) right-pads its count with leading spaces; GNU `wc -c` does not. Strip all
+  # whitespace before the numeric refusal check below so a macOS host's padded-but-valid count is
+  # not mistaken for the unmeasurable case — the `''|*[!0-9]*` refusal itself is unchanged and still
+  # fires closed on anything that isn't a bare digit string after stripping (#4220).
+  _cap_n=${_cap_n//[[:space:]]/}
   case "$_cap_n" in
     ''|*[!0-9]*)
       _component_set_replay_err
@@ -5805,8 +5810,13 @@ _component_set_probe_inner() {
   local _cs_complete=no _cs_obj_total="" _cs_obj_absent=""
   if _component_set_bounded "$_CS_BOUND_SECS" env -i "${_CS_GIT_ENV[@]}" ${_CS_READ_ENV[@]+"${_CS_READ_ENV[@]}"} git --no-replace-objects -C "$_CS_READ_DIR" cat-file -e "$remote_sha^{commit}" >/dev/null 2>&1 \
      && _component_set_bounded "$_CS_BOUND_SECS" env -i "${_CS_GIT_ENV[@]}" ${_CS_READ_ENV[@]+"${_CS_READ_ENV[@]}"} git --no-replace-objects -C "$_CS_READ_DIR" rev-list --objects --missing=print --no-walk "$remote_sha" >"$csdir/objcensus" 2>/dev/null; then
+    # BSD `wc -c` (macOS) right-pads its count with leading spaces, which would otherwise always
+    # trip the `*[!0-9]*` case below and silently force the slow (correct but expensive) path on
+    # every macOS run. Strip whitespace before the numeric case-match (#4220).
     _cs_obj_total=$(wc -c <"$csdir/objcensus" 2>/dev/null)
+    _cs_obj_total=${_cs_obj_total//[[:space:]]/}
     _cs_obj_absent=$(sed -n '/^?/p' "$csdir/objcensus" 2>/dev/null | wc -c 2>/dev/null)
+    _cs_obj_absent=${_cs_obj_absent//[[:space:]]/}
     case "$_cs_obj_total" in ''|*[!0-9]*) _cs_obj_total=0 ;; esac
     case "$_cs_obj_absent" in ''|*[!0-9]*) _cs_obj_absent=1 ;; esac
     if [ "$_cs_obj_total" -gt 0 ] && [ "$_cs_obj_absent" -eq 0 ]; then _cs_complete=yes; fi
--- a/scripts/agent-gate.sh
+++ b/scripts/agent-gate.sh
@@ -15821,6 +15821,13 @@ run_component() { # run_component <name> <cmd...>
 # records SKIP (loudly, never silently PASS) so a missing toolchain can't mask a
 # real Python regression the way it did pre-#865. Anything else (venv/build/test
 # failure) is a hard FAIL.
+# The soft open-file limit run_python_bindings requires before it will run pytest
+# (issue #4221). 4096 is not a measured requirement of the suite — it is comfortably
+# above anything a CORRECT reader needs, chosen so this guard removes the harness's
+# contribution to the EMFILE class without quietly becoming the reason a leaky reader
+# passes. See the block in run_python_bindings for the full mechanism.
+_PB_NOFILE_MIN=4096
+
 run_python_bindings() {
   local name=python-bindings
   if [ -n "$ONLY" ] && ! grep -qw "$name" <<<"${ONLY//,/ }"; then
@@ -15858,6 +15865,7 @@ run_python_bindings() {
   # cargo ran when nothing had. Captured here and mapped by the SHARED _fm_note_maturin_rc,
   # the same helper the --lite python tier uses.
   local pbv_rc=0
+  local nofile_eff=""
   # The reach marker travels by ENV because the verifier runs as a CHILD process; _pbv_build
   # touches it immediately before invoking maturin, so the reach is MEASURED rather than
   # inferred from this aggregate rc (roborev job 285).
@@ -15867,14 +15875,72 @@ run_python_bindings() {
   _fm_note_maturin_rc "$name" "$pbv_rc" "$pbv_reach"
   if [ "$pbv_rc" -eq 0 ]; then
     active_venv=$(cat "$active_venv_file" 2>/dev/null); [ -n "$active_venv" ] || active_venv="$venv"
-    if RUN_SLOW_TESTS="${RUN_SLOW_TESTS:-0}" bash -c '
+    # OPEN-FILE LIMIT — a HARNESS defect, fixed here so it cannot false-FAIL the gate of
+    # record (issue #4221). macOS has no `systemd-run --user`, so a detached gate is
+    # launched with `launchctl submit` — and a launchd job inherits launchd's SOFT
+    # `maxfiles` of 256, NOT the interactive shell's (measured on this host: 1048576
+    # interactive vs 256 under launchd; hard limit `unlimited` in both, which is why
+    # raising it here is permitted). The python parity suite accumulates a handle per
+    # SSTable component across a Database session, so under 256 the descriptor table is
+    # exhausted mid-suite and the reader yields ZERO ROWS instead of raising: ten
+    # test_parity.py cases FAIL as `got 0, expected N` with no EMFILE anywhere in the
+    # log, and the SAME ten pass at the interactive limit. Bisected by running the
+    # identical gate-built module under `ulimit -n 256` (the ten failures, byte for
+    # byte) and unrestricted (10/10 green).
+    #
+    # This does NOT fix the reader. `EMFILE` swallowed into a silent empty result, and
+    # the handle accumulation that reaches the limit at all, are a real product defect
+    # tracked on its own issue; the limit is deliberately set well above what a correct
+    # reader would ever need, so this guard cannot become the reason a leaky reader
+    # looks green.
+    #
+    # FAIL-CLOSED: if the limit cannot be brought to $_PB_NOFILE_MIN, the component FAILs
+    # naming the observed value rather than running — continuing would re-enter exactly
+    # the silently-wrong state this guard exists to prevent, and a component that reports
+    # `got 0, expected N` for an environmental reason is the worst of both outcomes.
+    # `-S` on EVERY ulimit call below is load-bearing, not decoration: bash's `ulimit -n N`
+    # with neither -S nor -H sets the SOFT **and HARD** limits together, so the bare form
+    # would permanently lower launchd's `unlimited` hard limit to $_PB_NOFILE_MIN for this
+    # process tree — a one-way door, since a hard limit can never be raised back by an
+    # unprivileged process. (Observed while verifying this guard: a probe that lowered the
+    # soft limit with the bare form could not raise it again — `cannot modify limit:
+    # Operation not permitted` — which is the same trap one step earlier.) Reading with
+    # `-S` likewise asks the question this guard actually cares about.
+    local nofile_out
+    nofile_out=$(mktemp "${TMPDIR:-/tmp}/agent-gate-nofile.XXXXXX")
+    if RUN_SLOW_TESTS="${RUN_SLOW_TESTS:-0}" _PB_NOFILE_MIN="$_PB_NOFILE_MIN" \
+       _PB_NOFILE_OUT="$nofile_out" bash -c '
         set -euo pipefail
+        want="$_PB_NOFILE_MIN"
+        cur=$(ulimit -S -n 2>/dev/null || printf "")
+        if [ "$cur" != unlimited ]; then
+          case "$cur" in
+            ""|*[!0-9]*) : ;;
+            *) if [ "$cur" -lt "$want" ]; then ulimit -S -n "$want" 2>/dev/null || : ; fi ;;
+          esac
+        fi
+        eff=$(ulimit -S -n 2>/dev/null || printf "")
+        printf "%s" "$eff" >"$_PB_NOFILE_OUT"
+        if [ "$eff" != unlimited ]; then
+          case "$eff" in
+            ""|*[!0-9]*)
+              printf "agent-gate: open-file limit UNMEASURABLE (got %s) — refusing to run pytest (#4221)\n" "${eff:-<no value>}"
+              exit 97 ;;
+            *)
+              if [ "$eff" -lt "$want" ]; then
+                printf "agent-gate: open-file limit %s is below the required %s and could not be raised — refusing to run pytest, because under a low limit this suite fails as '\''got 0, expected N'\'' rather than erroring (#4221). A gate launched via launchctl/launchd inherits a soft maxfiles of 256; relaunch with a higher soft limit.\n" "$eff" "$want"
+                exit 97
+              fi ;;
+          esac
+        fi
         . "'"$active_venv"'/bin/activate"
         pytest bindings/python/tests -q' >>"$log" 2>&1; then
       status=PASS
     else
       status=FAIL
     fi
+    nofile_eff=$(cat "$nofile_out" 2>/dev/null)
+    rm -f "$nofile_out"
   else
     status=FAIL
     active_venv=$(cat "$active_venv_file" 2>/dev/null); [ -n "$active_venv" ] || active_venv="$venv"
@@ -15890,7 +15956,12 @@ run_python_bindings() {
   fi
   end=$(date +%s)
   record_result "$name" "$status" "$((end - start))"
-  echo ">>> [$name] $RECORDED_STATUS ($((end - start))s)"
+  # Report the EFFECTIVE open-file limit pytest actually ran under (#4221), so a future
+  # reader of a green log can tell a run that had headroom from one that merely happened
+  # not to hit the ceiling — the failure mode this guards is SILENT (`got 0, expected N`),
+  # so the limit must be visible even on a PASS. Empty only when pytest was never reached
+  # (the maturin build failed), in which case there is no limit to report.
+  echo ">>> [$name] $RECORDED_STATUS ($((end - start))s)${nofile_eff:+ [open-files: $nofile_eff, required >= $_PB_NOFILE_MIN]}"
 }
 
 # _node_bindings_corpus_present: does CQLITE_DATASETS_ROOT hold a corpus the node jest
--- a/scripts/tests/test_agent_gate_component_set.sh
+++ b/scripts/tests/test_agent_gate_component_set.sh
@@ -1929,13 +1929,13 @@ fi
 # not itself host-dependent, and stays meaningful on a GNU-`wc` CI runner too): a small, legitimate
 # capture must still return RC 0 with its bytes intact once the numeric check strips whitespace
 # before matching.
-padwc_bin="$tmp/padwc-bin"
-mkdir -p "$padwc_bin"
-for _t in bash sh sed awk grep cut tr mktemp date basename dirname cat head tail sort \
-          uniq rm mkdir cp mv ln uname nproc env find touch stat comm od xargs sleep kill \
-          ps df readlink id iconv timeout nice chmod git; do
-  _src=$(command -v "$_t" 2>/dev/null) && [ -n "$_src" ] && ln -sf "$_src" "$padwc_bin/$_t"
-done
+# Single-sourced from `mkbin` (roborev job 3406, Medium) rather than a second, hand-rolled
+# tool list: a curated PATH missing a tool the gate legitimately needs (the first draft here
+# omitted `gtimeout`, `mkbin`'s own documented macOS bound-mechanism fallback) makes the
+# invocation fail for a reason unrelated to the padded-wc property this case measures — and
+# the `PATH="$padwc_bin"` prefix below governs the lookup of the outer `timeout` itself, not
+# just what the gate calls. `mkbin padwc wc` omits only `wc`, which is replaced below.
+padwc_bin=$(mkbin padwc wc)
 padwc_real=$(command -v wc)
 cat >"$padwc_bin/wc" <<EOF
 #!/usr/bin/env bash
--- a/scripts/tests/test_agent_gate_component_set.sh
+++ b/scripts/tests/test_agent_gate_component_set.sh
@@ -1920,6 +1920,40 @@ else
 fi
 
 # ---------------------------------------------------------------------------
+# THE CAP CHECK TOLERATES A BSD-STYLE PADDED `wc -c` (issue #4220). macOS ships BSD `wc`, which
+# right-pads its byte count with leading spaces (`wc -c` -> "     42" rather than GNU's "42"). The
+# cap-refusal check is a TEXTUAL case-match ('' | *[!0-9]*), so an unstripped padded count always
+# matched the `*[!0-9]*` arm and refused every capture through this path — not only oversized ones
+# — turning every bounded call into an RC-198 replay-refusal on any BSD-`wc` host. Reproduced here
+# with a `wc` SHIM that pads deterministically regardless of the host's own `wc` (so this case is
+# not itself host-dependent, and stays meaningful on a GNU-`wc` CI runner too): a small, legitimate
+# capture must still return RC 0 with its bytes intact once the numeric check strips whitespace
+# before matching.
+padwc_bin="$tmp/padwc-bin"
+mkdir -p "$padwc_bin"
+for _t in bash sh sed awk grep cut tr mktemp date basename dirname cat head tail sort \
+          uniq rm mkdir cp mv ln uname nproc env find touch stat comm od xargs sleep kill \
+          ps df readlink id iconv timeout nice chmod git; do
+  _src=$(command -v "$_t" 2>/dev/null) && [ -n "$_src" ] && ln -sf "$_src" "$padwc_bin/$_t"
+done
+padwc_real=$(command -v wc)
+cat >"$padwc_bin/wc" <<EOF
+#!/usr/bin/env bash
+out=\$("$padwc_real" "\$@")
+printf '   %s\n' "\$out"
+EOF
+chmod +x "$padwc_bin/wc"
+padwc_out="$tmp/padwc.out"; padwc_rc=""
+PATH="$padwc_bin" timeout 30 bash "$GATE" --component-set-bounded-run 10 "$osz_small" >"$padwc_out" 2>/dev/null; padwc_rc=$?
+padwc_line=$(sed -n 's/^RC: //p' "$padwc_out" 2>/dev/null)
+if [ "$padwc_rc" != 124 ] && [ "$padwc_line" = 0 ] && grep -q 'component-set-small' "$padwc_out" 2>/dev/null; then
+  ok "4220-padded-wc-cap: a BSD-style padded 'wc -c' count (leading spaces) is stripped before the numeric refusal check, so a legitimate small capture still returns RC 0 with its bytes replayed rather than being misread as unmeasurable"
+else
+  bad "4220-padded-wc-cap: expected RC=0 with 'component-set-small' replayed under a padded-wc PATH (got rc='$padwc_rc' line='$padwc_line')"
+  printf '%s\n' "$padwc_out"
+fi
+
+# ---------------------------------------------------------------------------
 # THE STDERR REPLAY IS BOUNDED TOO, AND IT TRUNCATES WHERE STDOUT REFUSES (roborev job 299, Medium).
 # The stdout cap above left stderr as a bare `cat`, which is the SAME defect on the other stream: a
 # descendant that outlives a SUCCESSFUL child keeps the stderr fd it inherited, and `cat` on a
--- a/scripts/tests/test_agent_gate_logdir_cleanup.sh
+++ b/scripts/tests/test_agent_gate_logdir_cleanup.sh
@@ -146,6 +146,23 @@ if [ -z "$tmp" ] || [ ! -d "$tmp" ]; then
   printf 'FAIL - could not create a scratch dir under %s — refusing to run\n' "${TMPDIR:-/tmp}"
   exit 1
 fi
+# NORMALIZE ONCE, HERE, RATHER THAN AT EVERY COMPARISON (issue #4221 item 3). macOS's
+# default `$TMPDIR` ends in a trailing slash, so the template above yields a `tmp` whose
+# path carries a literal "//" (e.g. ".../T//agent-gate-logdir.XXXXXX"). Every path this
+# suite builds by STRING CONCATENATION off `$tmp` (`$tmp/fakeroot`, `$tmp/td-...`, the
+# `find "$tmp" ...` scan) keeps that "//" verbatim — but a path the GATE ITSELF reports
+# does not: it reaches its own absolute paths through a real `cd`+`pwd`, and a shell's
+# `cd` collapses consecutive slashes as part of ordinary pathname resolution (verified:
+# `cd a//b && pwd` prints `a/b`) even without `-P`. So a suite built on `$tmp` and a gate
+# report built on a real `cd` disagreed on FORM for the exact same directory — not a
+# leak, a textual double-vs-single-slash mismatch. Collapsing `$tmp` to its canonical
+# form immediately, before anything is derived from it, makes every downstream
+# concatenation agree with what the gate's own `cd`+`pwd` would produce.
+tmp=$(cd "$tmp" && pwd) || tmp=""
+if [ -z "$tmp" ] || [ ! -d "$tmp" ]; then
+  printf 'FAIL - could not normalize the scratch dir path — refusing to run\n'
+  exit 1
+fi
 trap 'rm -rf "$tmp"' EXIT INT TERM
 
 # Fixture git must be ISOLATED from the invoker's environment, not merely given an
@@ -4303,11 +4320,28 @@ fi
 # before AC27/AC28/AC29 raised it by 34, 200 before AC25/AC26 raised it by 26); the floor is
 # what notices a DELETED CASE — every
 # case in this file contributes at least 5 verdicts — rather than a drifting count.
+#
+# TWO FLOORS, KEYED ON THE SAME AFFIRMATIVE CAPABILITY PROBE EVERY DEGRADED BRANCH ABOVE
+# ALREADY USES (issue #4221 item 3) — a single flat 259 was never actually reachable on a
+# degraded host: measured STABLY at 247 across 8 back-to-back runs on a macOS host with
+# OWNER_MARKER_CAPABLE=0 (this file's own declared Linux-only dependency), 27 short of the
+# full 274, not the 15-verdict margin the flat floor assumed. Guessing a wider flat margin
+# would only re-hide the exact defect this floor exists to catch (#3544's lesson) on the
+# capable branch, so each branch gets ITS OWN calibrated floor instead of one shared guess:
+# 259 where the capability is present (unchanged — every Linux-measured total above stays
+# covered), 243 where it is absent (4 below the measured-stable 247, so a genuinely deleted
+# case — minimum 5 verdicts — still drops the total below it, while ordinary run-to-run
+# noise does not).
 _total_verdicts=$((PASS + FAIL))
-if [ "$_total_verdicts" -ge 259 ]; then
-  ok "suite floor: $_total_verdicts verdicts reported (floor 259) — no case was silently dropped"
+if [ "$OWNER_MARKER_CAPABLE" = 1 ]; then
+  _floor=259
 else
-  bad "suite floor: only $_total_verdicts verdicts reported (floor 259) — at least one case was deleted or died before its assertions"
+  _floor=243
+fi
+if [ "$_total_verdicts" -ge "$_floor" ]; then
+  ok "suite floor: $_total_verdicts verdicts reported (floor $_floor) — no case was silently dropped"
+else
+  bad "suite floor: only $_total_verdicts verdicts reported (floor $_floor) — at least one case was deleted or died before its assertions"
 fi
 
 printf '\n%s\n' "scripts/tests/test_agent_gate_logdir_cleanup.sh   passed: $PASS  failed: $FAIL"
--- a/scripts/tests/test_agent_gate_logdir_cleanup.sh
+++ b/scripts/tests/test_agent_gate_logdir_cleanup.sh
@@ -4323,20 +4323,25 @@ fi
 #
 # TWO FLOORS, KEYED ON THE SAME AFFIRMATIVE CAPABILITY PROBE EVERY DEGRADED BRANCH ABOVE
 # ALREADY USES (issue #4221 item 3) — a single flat 259 was never actually reachable on a
-# degraded host: measured STABLY at 247 across 8 back-to-back runs on a macOS host with
+# degraded host: measured STABLY at 247 across 10 back-to-back runs on a macOS host with
 # OWNER_MARKER_CAPABLE=0 (this file's own declared Linux-only dependency), 27 short of the
 # full 274, not the 15-verdict margin the flat floor assumed. Guessing a wider flat margin
 # would only re-hide the exact defect this floor exists to catch (#3544's lesson) on the
 # capable branch, so each branch gets ITS OWN calibrated floor instead of one shared guess:
 # 259 where the capability is present (unchanged — every Linux-measured total above stays
-# covered), 243 where it is absent (4 below the measured-stable 247, so a genuinely deleted
-# case — minimum 5 verdicts — still drops the total below it, while ordinary run-to-run
-# noise does not).
+# covered).
+#
+# NO SLACK on the degraded branch (roborev job 3406, Medium): several cases collapse to
+# exactly ONE verdict when OWNER_MARKER_CAPABLE=0 — AC15 (~line 1501), AC16 (~line 1688),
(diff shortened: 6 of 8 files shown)


pmcfadin:
Round-2 roborev block (`--agent claude-code --model claude-opus-5`, job 3365) — posted per process, FAIL is on findings, deferred to #4214 per owner ruling (2-round script cap, #3893):

```
repo: /Users/patrickmcfadin/local_projects/cqlite/.claude/worktrees/issue-4206-roborev-guard-macos
branch: issue-4206-roborev-guard-macos
base: origin/main
head-sha: 64433558256506fda8cc408e5b786b585381ed24
reviewed-sha: 316f9f711b2a146de8dc3c2e1f9d179e53307633..64433558256506fda8cc408e5b786b585381ed24
assert-base: 316f9f711b2a146de8dc3c2e1f9d179e53307633 (merge-base of origin/main and HEAD; origin/main tip 316f9f711b2a146de8dc3c2e1f9d179e53307633)
job: 3365
model: claude-opus-5
census: 1 file, +23/-7
tokens: UNAVAILABLE
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS (no token accounting in the record)
sha-assert: PASS
review-completed: PASS
prompt-content: PASS (1/1 code census paths present)
vacuity-tier1: PASS
vacuity-tier2: UNAVAILABLE
findings: PRESENT (3)
log: /var/folders/2c/tfxwt1g91x729x38m63d81540000gn/T//roborev-review-issue-4206-roborev-guard-macos-64433558-89686.log
RESULT: FAIL
```

Findings (2 Medium, 1 Low) filed verbatim, with proposed fixes, as #4214.

pmcfadin:
## closer state 2026-09-11 (pre-reboot)

Durable handoff so this survives a host reboot. A fresh closer should resume from this comment.

**Branch state:** head `bc309c133ac4484e959a42162f70677a5a7aad1c`, **pushed**, tree and index clean.
No commits after the gate. **No roborev round has been run on this head**, so nothing is voided by
whatever we decide next.

### Gate of record — run 4

```
run-id: /var/folders/.../agent-gate.kQNr5a
commit: bc309c1 branch: issue-4206-roborev-guard-macos dirty: no
tree-start: bc309c133ac4 dirty: no digest: d6ffbea86295
tree-end:   bc309c133ac4 dirty: no digest: d6ffbea86295
tree-integrity: PASS
disk-admission: PASS (launch 68.9GiB; post-slot 68.9GiB; bar 40GiB)
RESULT: FAIL
```

39 components: **38 green**, `file-size` OPT-OUT (by design, `CQLITE_ALLOW_FILE_GROWTH=1`,
disclosed in the PR body), **1 FAIL — `tooling-tests`**.

### Five fixes on this branch, each proven

| # | commit | fix | evidence |
|---|---|---|---|
| — | (original PR) | #4206 roborev guard self-test `!_rw_i` | `test_roborev_review_guard.sh` **1596 / 0** in gate 4 |
| — | (original PR) | #4220 BSD `wc` padding | `component-set: PASS (39/39 names vs origin/main)` — never `FAIL-CLOSED (no-git)` |
| F1 | `d3c0dfd` | APFS EILSEQ loud-SKIP, `golden_fixture_staging.rs:225,:327` | `issue_1491_json_csv_golden_parity` **307 passed/2 failed → 309/0**; `cli-tests` PASS in gate |
| F2 | `8a82c0a` | soft open-file limit raised before pytest, fail-closed | `python-bindings PASS (49s) [open-files: 4096, required >= 4096]` under launchd's real 256 |
| F3 | `1090a15` | schemas-preflight scratch root → `pwd -P` | `test_agent_gate_schemas_preflight.sh` **36/5 → 41/0** |
| F4 | `bc309c1` | AC11b start-up race — wait for the stub's readiness marker | **1/10 → 10/10** runs, 249 verdicts each |

F2 detail (`launchctl limit maxfiles` = **256 soft / unlimited hard** on this host; interactive is
1048576): soft 256 → 10 failed / 36 passed; guard raises to 4096 → 47 passed / 0 failed; ceiling
`soft=hard=512` → refuses with exit 97 and pytest never runs. `-S` on every `ulimit` is
load-bearing — bash's bare `ulimit -n N` sets soft **and hard**, which would permanently lower
launchd's `unlimited` hard limit (a one-way door).

### The one CONFIRMED remaining failure

`scripts/tests/test_agent_gate_component_set.sh` — **144 passed / 4 failed (floor 136)**, from gate
4's own log (authoritative; predates the OOM described below).

- `3544-no-git` — fails NAMING a path, and quotes `.../T//agent-gate-cset.../nogit-tree`: the
  `$TMPDIR` trailing-slash/double-slash family. That file builds its scratch root as
  `mktemp -d "${TMPDIR:-/tmp}/agent-gate-cset.XXXXXX"` (line 132) and never normalizes.
- `3544-shallow-ancestry` — same shape (expected value present, the path naming failed).
- `3544-remote-unreadable` — same shape.
- `3544-bound-replay-err-bounded` — **NOT that family.** It concerns a bounded stderr replay and a
  124/198 status contract. So even this single file is not uniformly mechanical.

### Why fixes keep revealing new failures: `tooling-tests` aborts at the FIRST failing file

`run_tooling_tests` (`scripts/agent-gate.sh:21344`) is not a loop over a list — it is a
straight-line sequence of per-file guards, each returning out of the component on failure:

```
scripts/agent-gate.sh:21362-21372
  if ! bash "$REPO_ROOT/scripts/tests/test_generator_keyspace_scoping.sh" >>"$log" 2>&1; then
    status=FAIL
    ...
    record_result "$name" "$status" "$((end - start))"
    return 0        # component ends here; every later file never executes
  fi
```

That `return 0` shape repeats per guard (21347, 21370, 21386, 21405, …). Observed across four
gates: gate 1 died at `schemas_preflight`, gate 3 at `logdir_cleanup` AC11b, gate 4 at
`component_set` — each fix revealing the next file rather than advancing toward green.

**~101 of 106 tooling self-test files are UNMEASURED** — never executed in a gate or locally.
They are *not* known-passing.

### The sweep was OOM-killed — two datapoints are NOT results

A full 106-file local sweep was attempted and **killed by the OOM killer at file 6 of 106**.
At the kill: **~1.7 GiB available of 32 GiB**, swap **9,743 MB used of 11,264 MB**. No cargo, gate
or sccache build was running; the box was held by unrelated processes (a booted iOS Simulator and
Chrome).

Two results from that window are contaminated and must be **discarded**:

- `test_agent_gate_component_set.sh` **FAIL(137)** — 137 = 128+9 = **SIGKILL**, the OOM killer, not
  a verdict. (Its real 144/4 figure above comes from gate 4, not from the sweep.)
- `test_agent_gate_census.sh` — "ABORTED before its terminal tally" inside the same window.
  **Unknown**, not a failure.

### Shared-helper estimate

`scripts/tests/lib/` already exists and is sourced by self-tests (e.g.
`lib/agent-gate-canonical-pin.bash`), so a canonical scratch-root helper has a home:
**~25 lines once, plus ~3 lines per adopting file.**

Static survey of the candidate population: **66** self-test files build a scratch root via
`mktemp -d "${TMPDIR:-/tmp}/..."` and only **3** normalize it today. Worst case ≈ `25 + 3*66 ≈ 225`
lines — but this is an **upper bound on a population, not a defect count**. Most of those files
never compare their path against a `pwd -P`-resolving subject and are probably fine.

### Recommendation: provision Linux rather than fund harness repair on this box

1. **The work cannot be bounded.** ~101 unmeasured files revealed one per ~75-minute gate. Four
   gates have surfaced four defects and the queue has not visibly shortened.
2. **The instrument is unreliable.** The host is out of memory and swap, and at least one defect
   here (the AC11b start-up race) is **load-sensitive** — it went from 1-in-3 to 9-in-10 as the box
   got busier. Chasing load-sensitive races on a thrashing box manufactures defects that exist
   nowhere else.
3. **None of it improves CQLite.** Every fix so far is macOS harness repair — APFS EILSEQ, launchd
   `maxfiles`, `$TMPDIR` spelling, an #1825 slot-queue race. It buys one PASS on one Mac; a Linux
   runner retires the whole class.

The five fixes above are real and worth landing regardless.

### Traps for the next closer (learned the hard way here)

- **`launchctl submit` is KeepAlive: launchd RESTARTS the gate every time it exits.** Used bare it
  manufactures exactly the foreign-run-id condition #3637 teaches you to distrust — a new run-id
  overwriting your pinned summary path, indistinguishable from a peer without `launchctl list`.
  Two jobs were respawning here for hours and 325 run dirs accumulated. Launch behind a wrapper
  that (a) exits immediately if the summary already holds a terminal `RESULT: PASS|FAIL`, and
  (b) `launchctl remove`s its own label as its last act. Then verify exactly one job and one gate
  tree after launch. *(The wrapper lived in a session scratchpad under `/private/tmp` and will not
  survive the reboot — recreate it from this recipe.)*
- `scripts/flow/gate-detached.sh` exits **69** on macOS (no `systemd-run`), hence the above.
- Verify every verdict by **`run-id`**, never by recency — `ls -t` on `$TMPDIR` lands on a peer or
  on your own respawn.
- Time Machine local snapshots pin freed disk: `cargo clean` can free 8 GB and move free space by
  zero. `tmutil thinlocalsnapshots / 40000000000 4` (documented in `gate-ops.md:544-551`).

**Plan after reboot:** a fresh closer resumes from this comment, runs the full 106-file sweep on a
quiet box (diagnostic only, not a gate), and only then do we choose fund-vs-Linux from a real list.
If a Linux box is provided first, the gate of record moves there and the sweep is moot.

pmcfadin:
## macOS tooling self-test sweep 2026-09-11 (post-reboot)

**Not a gate.** Diagnostic sweep of all 110 self-test files `run_tooling_tests`
invokes (`scripts/agent-gate.sh:21344-23647`), file list/order/env derived from the
function body, run serially from the worktree via a launchd one-shot job (PPID=1,
detached from any session), 420s/file timeout (`timeout -k 10 420`), continuing past
failures. Env: `REPO_ROOT`, `CQLITE_DATASETS_ROOT` (main checkout), `CQLITE_SKIP_DOCKER_TESTS=1`,
`ulimit -S -n 4096`, plus PATH/HOME hardened for launchd (bare launchd PATH lacks
`/opt/homebrew/bin`, breaking `timeout`/cargo/git — fixed after run 1's spurious 110×
FAIL(127)). Box quiesced from load1=213 (right after boot) to 7.5 before starting, per
instruction — it re-spiked to 216 mid-wait (bursty `mdworker_shared`/Spotlight reindex,
confirmed via `mdutil -s /`), not a hang.

**Totals: 74 PASS / 36 non-PASS (31 FAIL, 5 TIMEOUT).** ~101 of 110 files had never
executed on this Mac before this run (per predecessor's audit) — most of these findings
are new information, not new breakage.

### Highest-leverage single fix: a 1-character awk bug (CONFIRMED, reproduced live)

`scripts/perf/lib-perf-lint.sh:183` (inside the `perf_invocation_lint` awk program):
```
if (t == "perf" || t ~ /[/]perf$/) {
```
Reproduced directly against this host's `/usr/bin/awk` (BWK/one-true-awk 20200816,
the macOS default — NOT gawk):
```
awk: nonterminated character class [
 context is
       if (t == "perf" || t ~ >>>  /[/ <<< ]perf$/) {
```
BWK awk cannot parse an unescaped `/` inside a bracket expression `[...]` within a
`/regex/` literal; gawk tolerates it. The awk dies mid-file, never reaches its `END`
block, so `_perf_lint_verify_complete` reports "the lint did not COMPLETE over this
file" and the **caller FATALs every perf invocation on this host**, valid or not:
`FATAL: this rig contains a per-process/per-thread perf invocation, invokes perf
outside its single wrapper, or carries an option outside the allowlist: …lib-args.sh:0:
the lint did not COMPLETE…`

**Fix: `t ~ /[/]perf$/` → `t ~ /\/perf$/` — literally one character.** This FATAL
cascades into (at least) 6 of the 36 failures below: `test_ws0_perf_invocation_lint.sh`
(029, direct positive control, 71/129 checks), `test_ws0_primary_path_admits_a_legitimate_run.sh`
(038, 10/18), `test_ws0_report_guards.sh` (018, 30/123), `test_ws0_hermeticity.sh`
(031, several of 5/101), `test_ws0_output_dir_exclusivity.sh` (034, several of 4/22),
`test_ws0_canonical_corpus.sh` (035, both of 2/29). **This is NOT the $TMPDIR family** —
it is a distinct BWK-awk-vs-gawk portability bug, mechanical, ~1 line.

### Classified list (36 non-PASS)

Legend: **class** = CONFIRMED (reproduces the same way the gate would; env-independent
or verified against this host directly) / SUSPECTED (plausible root cause identified,
not fully source-confirmed) / STANDALONE-ONLY (needs something my sweep's env lacked
that the real gate likely supplies, or is transient).

| # | file | class | cause | fix shape | est. lines | TMPDIR-family? |
|---|------|-------|-------|-----------|-----------:|----------------|
|1|`test_ws0_perf_invocation_lint.sh`|CONFIRMED|awk bracket-slash bug (see above)|1-char regex fix|1|no|
|2|`test_ws0_primary_path_admits_a_legitimate_run.sh`|CONFIRMED|same awk bug|shared with #1|0|no|
|3|`test_ws0_report_guards.sh`|CONFIRMED (partial)|same awk bug, most of 30/123|shared with #1|0|no|
|4|`test_ws0_hermeticity.sh`|CONFIRMED (partial)|same awk bug, 3 of 5/101|shared with #1|0|no|
|5|`test_ws0_output_dir_exclusivity.sh`|CONFIRMED (partial)|same awk bug, most of 4/22|shared with #1|0|no|
|6|`test_ws0_canonical_corpus.sh`|CONFIRMED (partial)|same awk bug, both of 2/29|shared with #1|0|no|
|7|`test_ws0_flight_arm_guards.sh`|SUSPECTED|10/119: some overlap w/ awk bug; others say "measured >2x observer cost on this workload" — plausibly genuine perf-overhead-under-load on a busy shared box, not a bug|re-run isolated/quiet to separate the two causes|0 (pending)|no|
|8|`test_gating_registry_policy.sh`|SUSPECTED (env)|Ruby default external encoding is US-ASCII with no `LANG`/`LC_ALL` set (confirmed: `env -i ruby -e 'puts Encoding.default_external'` → `US-ASCII`; repo files contain em-dashes etc.). `agent-gate.sh` never exports `LANG`/`LC_ALL` globally|export `LANG=en_US.UTF-8` (or `C.UTF-8`) once near `agent-gate.sh`'s global setup — OR harden `gating_registry.rb`/`gating_policy_rules.rb` with `# encoding: UTF-8` at top|1 (global) or ~2/file (defensive)|no|
|9|`test_gating_workflow_semantics.sh`|SUSPECTED (env)|same Ruby locale issue (partial) + one failure downstream of #20's GNU_RULES read|shared with #8, plus see #20|shared|no|
|10|`test_agent_gate_tree_portability.sh`|CONFIRMED|BSD `stat` has no usable sub-second-mtime equivalent to GNU's `%Fm`; the cap-line disclosure and datum-unused checks fail on real BSD stat output|BSD-stat mtime-resolution shim or accept+disclose degraded resolution (mechanism already partially exists per the FAIL text — needs the datum actually wired)|~20-40|no|
|11|`test_agent_gate_tree_provenance.sh`|MIXED|(a) J3: fd-redirect-target resolution fails — SUSPECTED no `/proc/self/fd` on macOS; (b) K1: **real, platform-independent bug** — a filename containing a space renders unescaped in the tree-mutation diagnostic ("space-not-escaped"), while the sibling lockfile-settled case already escapes it correctly|(a) macOS fd-resolution path (`lsof`-based) or documented gap, moderate; (b) apply the same `\s`-escaping convention already used elsewhere in the same renderer|(a) ~20-30, (b) ~5-10|no|
|12|`test_tools_crate_disposition_selftest.sh`|SUSPECTED-real|3 negative-control cases where the guard PASSES on a tree it must reject (crate removed from disk; recorded in both WIRED+UNWIRED; recorded in both WIRED+MIXED) — a real validation gap|read `check_tools_crate_disposition`'s validation logic, add the missing rejects|~15-30|no|
|13|`test_check_object_store_integrity.sh`|SUSPECTED-real|ambient `GIT_OBJECT_DIRECTORY`/`GIT_DIR` leaks into the checker despite the file's own claim every git call is isolated — one code path (`print-store`) isn't|isolate the missed call site the same way the others are|~10-20|no|
|14|`test_agent_gate_disk_exhaustion.sh`|SUSPECTED-real|single case `25f-retarget-disagree`: expected re-based path/flag/declaration not produced|read `_disk_retarget`'s disagree branch|~5-15|no|
|15|`test_lane_lock.sh`|**CONFIRMED, structural**|`lane-lock.sh` requires `/proc/<pid>/stat` (boot-id + start-ticks) for process identity and explicitly REFUSES by design on hosts without `/proc`; its own error text says so ("do not take a lane lock on this platform — the cross-machine control is refs/claims/issue-N via claim.sh, which needs no /proc")|NOT mechanical: needs either a macOS-native identity backend (`sysctl KERN_PROC`/`proc_pidinfo` start-time port) or marking this self-test SKIP on non-Linux, pending an owner call on whether lane-lock.sh is meant to run on macOS at all|~80-150 (port) or ~10-20 (skip)|no|
|16|`test_finalize_cleanup_lane_lock.sh`|**CONFIRMED, structural**|same `/proc` dependency, cascading from lane-lock.sh|same as #15|shared|no|
|17|`test_lane_lock_precommit_hook.sh`|CONFIRMED|hardcoded `/data` path assumption (fleet-box convention) — `/data` is read-only/absent on this Mac (`mkdir: /data: Read-only file system`)|fall back to `${TMPDIR:-/tmp}` when `/data` isn't writable|~5-10|adjacent (fix routes through the same scratch-root convention, but the bug itself is a hardcoded path, not a normalization bug)|
|18|`test_drive_issue_state.sh`|**CONFIRMED, structural**|`flock(1)` binary is entirely ABSENT on this Mac (`command -v flock` → not found); write/adopt refuse by design without it, "NOTHING here is measured"|install `flock` (homebrew formula, zero code) OR port to a macOS-native atomic lock (`mkdir`-based)|0 (operational) or ~30-60 (port)|no|
|19|`test_claude_auth_capability.sh`|**CONFIRMED**|`$TMPDIR` too long for an `AF_UNIX` `sockaddr_un.sun_path` (104-byte macOS limit) — this host's `TMPDIR` is `/var/folders/2c/.../T/`, leaving little room for the generated probe-socket name|use a short fixed base (`/private/tmp/<short-id>.sock`) for the private probe socket instead of `$TMPDIR`|~5-10|**yes — TMPDIR family, but the LENGTH sub-issue, not the normalization/pwd-P sub-issue** (a `pwd -P` helper does not shorten the path)|
|20|`test_gate_detached.sh`|**CONFIRMED, structural**|5 failing cases (`_pid_state classification`, `cgroup holding only an orphan`, etc.) exercise cgroup/pid-state concepts that don't exist on macOS — consistent with `gate-detached.sh` exiting 69 on macOS by design (no systemd-run)|SKIP these cases on Darwin explicitly, or accept the self-test can't be 100% green here|~10-20|no|
|21|`test_pub_surface_guard.sh`|SUSPECTED|self-test's OWN fixture setup bug: "case 19 setup: the decoy worktree was never registered" (possibly a conflict with running from inside an existing linked worktree, or APFS case-insensitivity)|root-cause the decoy-worktree registration in the self-test harness|~5-15|no|
|22|`test_agent_gate_summary.sh`|SUSPECTED macOS-inapplicable|expects `perf=ok`/`perf=paranoid-4` from the accelerator line, gets `perf=absent` — real Linux `perf` doesn't exist on macOS and the fixture's stub isn't tripping the gate's detection the way it does on Linux|either fix the stub's PATH placement or document/SKIP the `perf=ok` expectation on Darwin|~15-30|no|
|23|`test_agent_gate_notify.sh`|SUSPECTED-real|single case: "SIGTERM-only-timeout ... the capability was not probed before use"|read the notify capability-probe ordering|~5-15|no|
|24|`test_agent_gate_disk_admission.sh`|SUSPECTED-slow, not broken|TIMEOUT at 420s but tail shows steady "ok" progress with no visible defect — plausibly just slow under this box's concurrent load (1-min load 10-30 during the run)|re-run standalone with a longer budget (900s+) before concluding it's broken|0 (pending re-run)|no|
|25|`test_bootstrap_agent_machine.sh`|**CONFIRMED** (getent) + SUSPECTED (hang)|`getent` doesn't exist/work the same on macOS ("getent could not name this account's home, so the root-branch cases cannot assert anything"); TIMEOUT afterward is SUSPECTED a `sudo` password-prompt hang under non-interactive launchd in a later root-branch case|add a macOS branch (`dscl . -read /Users/$USER NFSHomeDirectory`) as the getent-equivalent; gate any `sudo` use behind `sudo -n true` to fail fast instead of hanging|~15-25|no|
|26|`test_perf_capability.sh`|SUSPECTED-real|single case: an absolute shim dir outside the sandbox root was accepted or refused without naming the sandbox|read the sandbox-containment check for that one case|~5-15|no|
|27|`test_premerge_assert.sh`|SUSPECTED-real|2 cases: a fixture expecting two `dirty:` tokens on the `commit:` line isn't matched; a refusal doesn't name its cause `AMBIGUOUS` as documented|read the `dirty:`-token parser and the refusal-naming path|~10-20|no|
|28|`test_agent_gate_census.sh`|SUSPECTED-real|"F5 (control): the control run did not terminate PASS/PARTIAL" + a SEPARATE "ABORTED before its terminal tally" — same shape the predecessor saw under OOM, but THIS run was NOT memory-pressured, so this looks like a genuine reproducible defect, not OOM contamination|investigate the F5 control-run fixture and the abort path directly (not under memory pressure this time)|unknown, needs investigation|no|
|29|`test_gen_perf_corpus_3068.sh`|SUSPECTED-real|"TABLES=medium pruned the wrong set" — 2 failures in prune-scoping|read the prune-scope matcher for the `TABLES=medium` case|~5-15|no|
|30|`test_gen_perf_corpus_bti.sh`|STANDALONE-ONLY / transient|cascading case-count-floor FAIL driven by "SKIP - only ? GiB free under $TMPDIR ... needs >= 4 GiB" — almost certainly transient disk pressure at that moment in the 74-minute serial run, not a code defect|re-run in isolation to confirm PASS with headroom|0 (pending re-run)|adjacent (TMPDIR disk space, not path shape)|
|31|`test_fetch_datasets_tracked_guard.sh`|SUSPECTED-real, one flagged safety-relevant|5/352: an error-message path-naming gap, plus **"home-target: exited 0 — a destructive path was taken silently"** — this one deserves priority attention regardless of overall triage since it concerns a destructive-operation safety guard around `$HOME`|read the home-target guard directly; treat as higher priority than its 1-of-5 share suggests|~10-20|no|
|32|`test_agent_gate_component_set.sh`|CONFIRMED (known) + SUSPECTED (new)|Reproduces predecessor's known `3544-no-git`/`3544-shallow-ancestry` ($TMPDIR family) and `3544-bound-replay-err-bounded` (confirmed NOT that family — a stderr-replay-bound bug) from gate 4's 144/4. **New since gate 4**: `3544-partial-clone-unreachable` FAILs (positive control didn't fire), immediately followed by a **new hang** (SIGKILL needed, 430s) after `3544-symlink-baseline` with no further output — gate 4 completed this file without hanging. SUSPECTED: a git credential/network-reachability hang in the TEST'S OWN fixture setup (the isolated subject-under-test already sets `GIT_TERMINAL_PROMPT=0`; the harness's own setup calls may not)|known-3: shared scratch-root helper (below). New hang: re-verify wrapping the WHOLE test invocation in `GIT_TERMINAL_PROMPT=0 GIT_ASKPASS=/bin/false GIT_SSH_COMMAND='ssh -o BatchMode=yes'`|see below (shared helper) + hang investigation|**yes — the classic normalization/pwd-P family** for no-git/shallow-ancestry|
|33|`test_gate_liveness.sh`|SUSPECTED|hangs after section 11r with no further output (SIGTERM sufficed, 124) — plausibly a background-process/job-control wait that never resolves without a controlling TTY under launchd|identify what section follows 11r and whether it `wait`s on a backgrounded process|unknown|no|
|34|`test_worker_supervisor.sh`|SUSPECTED|hangs after the test's OWN internal SIGTERM-15 to a background `worker-supervisor.sh` child — suggests an orphaned grandchild not reaped under launchd's process-group semantics (no controlling TTY)|investigate process-group/session handling around the backgrounded worker in the self-test|unknown|no|

### The $TMPDIR/pwd-P normalization family — explicit answer

**Two distinct sub-issues share the "$TMPDIR" name; only one is fixed by a shared
normalization helper:**
- **Normalization** (predecessor's original family — trailing-slash/double-slash from
  `mktemp -d "${TMPDIR:-/tmp}/..."` never resolved via `pwd -P`): reproduced again here
  in `test_agent_gate_component_set.sh` (`3544-no-git`, `3544-shallow-ancestry`).
  **Yes, a single shared helper under `scripts/tests/lib/` would collapse these** —
  confirming the predecessor's estimate: ~25 lines once + ~3 lines per adopting file,
  against a static population of 66 files building a scratch root this way (only 3
  normalize today). Still an upper bound on a population, not a defect count.
- **Length** (new in this sweep — `test_claude_auth_capability.sh`, item 19): `$TMPDIR`
  itself is too long for an `AF_UNIX` socket path. `pwd -P` normalization does **not**
  fix this — the fix is a short fixed base, unrelated to the shared helper above.

### Recommendation (unchanged from predecessor, reinforced)

This sweep found **36 non-PASS files**, well past the ≤5-mechanical threshold, and at
least 4 are **structural macOS incompatibilities requiring a design decision, not a
mechanical fix** (`lane-lock.sh`'s `/proc` dependency, `flock(1)` entirely absent,
`gate-detached.sh`'s cgroup dependency, plus the getent/sudo gap in
`bootstrap_agent_machine.sh`). One single-character awk fix is real and cheap and
should land regardless of the fund-vs-Linux decision — it's the highest-leverage item
found. But the overall shape confirms the predecessor's read: this is unbounded,
partially load-sensitive, and every fix so far is macOS harness repair, not CQLite
correctness. **LONG path per the lead's rule — no commits made.**

pmcfadin:
Lead ruling on the post-reboot sweep (36/110 non-PASS): LONG path, no fix round on this PR. Tracking issue for the macOS harness class: #4225 (gate-host decision pending with the owner). Production guard gap found by the sweep: #4224. The five proven fixes stay on this branch; this PR merges once a full gate PASSes on whichever host the owner picks.

pmcfadin:
**Lane resumed by the lead 2026-09-20 (owner ruling: 0.18 = CLI update with new tooling; this PR is in it).** Linux gate host now exists (astro-processor, #4225 option (b)), so the September 12 ruling — "merges once a full gate PASSes on whichever host the owner picks" — is actionable.

- Rebased `bc309c133` → **`6e9b6b6f4`** onto `origin/main` `aab7cd357` (10 commits replayed clean, 0 behind). The rebase changes bytes, so the round-2 roborev (job 3365, findings deferred to #4214) is VOID; a fresh roborev round runs LAST, after the gate.
- Claim ref for #4206 adopted (`claim.sh adopt --expect none`, audit reason recorded in the claim commit); heartbeat from this Mac.
- **Gate of record r1 launched 15:04Z** from lane `~/workspace/lanes/issue-4206-roborev-guard-macos` at `6e9b6b6f4`: unit `cqlite-gate-20260920T150404Z-2001846`, run-id `/data/tmp/agent-gate.EMooTa`, summary `~/logs/gate-4215-r1.txt`. Host checklist per #4196's r11 (git 2.43.7, bash 5.2, tmpfs TMPDIR, `RUST_TEST_THREADS=1`, concurrency 1). No `CQLITE_ALLOW_FILE_GROWTH` — a scripts/tests diff should not need it.

pmcfadin:
### Gate of record — run r1 result + r2 launched (closer session, 2026-09-22)

Rebase check: branch `6e9b6b6f4` already contains all of `origin/main` (`aab7cd357`) — no rebase needed.

**Gate r1** (astro-processor, run-id `/data/tmp/agent-gate.EMooTa`, commit `6e9b6b6f4`, dirty:no, tree-integrity PASS): `RESULT: FAIL`. Two failing components, both diagnosed:

1. `file-size: FAIL` — `cqlite-cli/tests/issue_1490_parquet_golden_admission.rs` grew 1497→1515 (limit 1500), caused directly by this PR's #4221 APFS-EILSEQ loud-skip fix (the block already went through roborev job 3406). A file split is out of scope for this macOS-gate-fix PR (unrelated parquet-admission suite) — per CLAUDE.md's campsite-rule remedy, re-running with `CQLITE_ALLOW_FILE_GROWTH=1` (tracked: epic #1135 test-file-split).
2. `node-bindings: FAIL` — 2/33 jest suites failed (`conversion-budget.test.js` per-row heap budget 2289B vs 2000B; `issue-3612-multicell-tuple-map-key.test.js` `beforeAll` hook 30s jest timeout). Neither touches any file in this PR's diff (diff is scripts/tests-only — no `cqlite-core/src`, no `bindings/node/src`). Re-ran `--only node-bindings` (diagnostic, same head `6e9b6b6f4`, same box) as a flake check: **PASS, 602/602 jest tests, 0 failures** — confirms non-determinism unrelated to this diff, consistent with astro-processor's documented host-load sensitivity on timing-based tests (see prior #4196 gate notes).

**Gate r2 launched** with `CQLITE_ALLOW_FILE_GROWTH=1` added: unit `cqlite-gate-20260922T040633Z-3977612`, run-id `/data/tmp/agent-gate.r8Vog9`, summary `~/logs/gate-4215-r2.txt`, commit `6e9b6b6f4` (unchanged — no src edits, override only). Will post the terminal result here.

pmcfadin:
### Gate r2 PASS + fresh roborev round (job 4263) + gate r3 launched

**Gate r2** PASSed at commit \`6e9b6b6f4\` (run-id \`/data/tmp/agent-gate.r8Vog9\`, all 39 components PASS/OPT-OUT, tree-integrity PASS, node-bindings PASS 602/602 confirming the r1 failure was a host-load flake).

**Fresh roborev round (job 4263)** on \`6e9b6b6f4\` (required LAST, after the gate): FAIL — 9 findings (3 Medium, 6 Low). Triaged per `docs/development/roborev-severity.md`:
- **Blockers fixed** (commit \`56ec20428\`):
  1. Open-file soft-limit raise (#4221) was scoped to `python-bindings` only, leaving every other gate component sharing the same launchd-inherited 256-fd cap exposed to the same silent-EMFILE-truncation class this guard exists to close. Widened to a one-time raise in `launch_components()` before either lane forks.
  2. The effective open-file limit wasn't visible on the summary row (only in `gate.log`, which the doctrine says an agent never retains). Published via `_record_status_detail` (#3402).
  3. `test_agent_gate_logdir_cleanup.sh`'s AC11b stub ran as a full (slot-capped) gate, so its 30s readiness wait could be confused with a real #1825 queue wait when this self-test executes inside the gate of record itself — a potential host-caused red. Slot-exempted it with `--only`.
- **6 Low findings batched to a follow-up issue** (mktemp-unchecked robustness, a self-test diagnostic printing a path instead of file contents, dead code/stale comment, a missing issue-number reference, tmp-normalization-on-failure edge case, missing `#[cfg(unix)]` on two non-consumer-reachable test helpers) — nits per the rubric, no re-verify round triggered. Follow-up issue to be opened at merge time.

Verified locally: `scripts/tests/test_agent_gate_logdir_cleanup.sh` 249/249 (was flagged red under peer load before the fix); `--lite` clean (fmt/clippy/roborev-lints/scoped-tests all PASS; `file-size` FAIL is the pre-existing #4221 growth, already handled via `CQLITE_ALLOW_FILE_GROWTH=1` at the full-gate level).

**Gate r3 launched** at the fixed head `56ec20428`: unit `cqlite-gate-20260922T062104Z-804113`, run-id `/data/tmp/agent-gate.57Ks4H`, summary `~/logs/gate-4215-r3.txt`. Will post the result + a second fresh roborev round here.

pmcfadin:
### Gate r3 PASS + roborev round 2 (job 4272) + fixes + gate r4 launched

**Gate r3** PASSed at \`56ec20428\` (all 39 components PASS/OPT-OUT).

**Roborev round 2 (job 4272)** on \`56ec20428\`: FAIL — 6 findings (2 Medium, 4 Low). Triaged:
- **Blockers fixed** (commits \`3e51a9905\` + \`f6aaf8265\`):
  1. The 3 EILSEQ skips (golden_fixture_staging.rs, issue_1490, issue_3148) were silently swallowed by libtest's stdout capture on passing tests, so "skip loudly" didn't actually hold. Required the capability on Linux (`#[cfg(target_os = "linux")] panic!` instead of skip) — the gate-of-record platform can no longer have a real regression present as an environmental skip; the APFS-only skip is unchanged. (A first attempt used `assert!(!cfg!(target_os = "linux"), …)`, which is a clippy `assertions_on_constants` violation — fixed with a proper `#[cfg]` split, verified clean under `cargo clippy -D warnings`.)
  2. `run_python_bindings`'s open-file guard didn't consult the HARD nofile ceiling, so a host with a hard limit below 4096 (1024/2048 are common non-macOS defaults) would hard-FAIL with an inapplicable remedy ("relaunch with a higher soft limit") even though the *measured* failure point is 256. Now raises to `min(hard, 4096)` and only FAILs at-or-below the measured-failing floor (256); a hard-ceiling shortfall above the floor is reported as a host limitation, not a harness refusal, in both the warning and the summary detail. `_gate_raise_nofile_limit` got the same hard-limit awareness.
  3. The open-file sidecar moved from an unchecked `mktemp` under `$TMPDIR` to a fixed path under `$LOG_DIR` (matches the existing `$pbv_reach` pattern).
  4. AC11b's readiness-timeout branch now emits the same 4 verdicts the success branch does, instead of 1 — a readiness timeout no longer also trips the suite's "NO SLACK" degraded-floor check as a second, misleading failure.
- **2 nits batched to the follow-up issue** (a stale comment; the earlier round's remaining mktemp/diagnostic/dead-code/cfg nits — see below).

Verified locally: `cargo clippy -p cqlite-cli --tests` and `-p cqlite-core --test issue_3148_fixture_roots_contract` both clean under `-D warnings`; the 3 EILSEQ-touching Rust test files pass (18/18, 309/309 incl. golden_fixture_staging's own `mod tests`, 1/1); `test_agent_gate_logdir_cleanup.sh` 249/249. `--lite` clean (clippy/fmt/roborev-lints/scoped-tests all PASS; `file-size` FAIL is the pre-existing #4221 growth).

**Gate r4 launched** at `f6aaf8265`: unit `cqlite-gate-20260922T082204Z-1825170`, run-id `/data/tmp/agent-gate.zdXboa`, summary `~/logs/gate-4215-r4.txt`.

Follow-up issue for all batched nits (round 1's 6 Lows + round 2's 2 Lows = 8 total) will be opened at merge time per the roborev-severity doctrine.

pmcfadin:
### Gate r4 caught a real Linux-only clippy issue + gate r5 launched

**Gate r4** FAILed at \`f6aaf8265\` on \`clippy\` — my previous fix for the EILSEQ Linux-required guard used \`#[cfg(target_os = "linux")] panic!\` paired with \`#[cfg(not(target_os = "linux"))] { println!/break 'non_utf8; }\`, which correctly avoided the clippy findings on macOS but left the \`'non_utf8\` label with no use on Linux (its only \`break 'non_utf8\` site is compiled out there) — rustc's \`unused_labels\`, denied under \`-D warnings\`. Everything else PASSed (all 38 other components, including \`open-files: 4096\` still visible on `python-bindings`).

**Fixed** (commit \`ca2e96565\`): \`#[allow(unused_labels)]\` on the label, with a comment naming exactly why it's unused on that one target. Verified via a cheap \`--only clippy\` diagnostic on the same Linux box before re-running the full gate (PASS, 17s) — this is precisely the "macOS gate can't prove the Linux-only interaction" gap the astro-processor host exists to close; a purely local macOS check would have shipped this bug.

**Gate r5 launched** at \`ca2e96565\`: unit \`cqlite-gate-20260922T095405Z-2848166\`, run-id \`/data/tmp/agent-gate.cCMNCT\`, summary \`~/logs/gate-4215-r5.txt\`.

pmcfadin:
roborev-defer: findings issues=4253 count=4 base=aab7cd357c9c73a296e02c98e7527e28030de978 head=ca2e9656578b88259dadeced76c48d1865851da6 job=4274 reason=round-3 findings on a scripts-heavy diff, two-round fix cap (owner ruling 2026-09-01, #3893) — disposed to follow-up #4253, not fixed; none is a hang or a false verdict

pmcfadin:
### Roborev round 3 (job 4274) — deferred via owner-authorized marker, RESULT: PASS

Round 3's 4 findings (2 Medium, 2 Low — misleading `required >= 4096` wording when only the 256 floor is enforced, an over-attributing "host hard ceiling" NOTE message, an unbalanced AC11b outer-`else` verdict count, and missing self-test coverage for the new open-file guard) are disposed per the two-round scripts-cap ruling (owner 2026-09-01, #3893) rather than triggering a 3rd fix round — none is a hang or a false verdict. Filed as follow-up #4253 (along with rounds 1–2's batched Low nits) and deferred via the `roborev-defer:` marker above, applied with `--recheck-job 4274`.

```
MODE: recheck (job 4274 re-decided from its job record; NO review was enqueued — not evidence of a fresh review)
recheck-of: 4274
repo: /Users/patrickmcfadin/local_projects/cqlite/.claude/worktrees/issue-4206-roborev-guard-macos
branch: issue-4206-roborev-guard-macos
base: origin/main
head-sha: ca2e9656578b88259dadeced76c48d1865851da6
reviewed-sha: aab7cd357c9c73a296e02c98e7527e28030de978..ca2e9656578b88259dadeced76c48d1865851da6
job: 4274
model: claude-opus-5
census: 8 files, +533/-51
push-assert: PASS
census-check: PASS
code-free: PASS
job-record: PASS (no token accounting in the record)
sha-assert: PASS
review-completed: PASS
vacuity-tier1: PASS
findings: DEFERRED (4, issues=#4253, authorized @pmcfadin, job 4274)
deferral: GRANTED (author=@pmcfadin issues=4253 count=4 scope=base=aab7cd357c9c73a296e02c98e7527e28030de978 head=ca2e9656578b88259dadeced76c48d1865851da6 job=4274)
roborev-exit: SKIP (recheck: no reviewer ran in this invocation; job 4274 re-decided from its record)
RESULT: PASS
```

**Gate of record**: r5 PASS at `ca2e96565` (unit `cqlite-gate-20260922T095405Z-2848166`, run-id `/data/tmp/agent-gate.cCMNCT`). No commits after this head — gate r5 and roborev round 3 both certify the same tree. Proceeding to premerge-assert → HOLD re-read → arm.

